PR #19, opened 2025-04-23
Add Github actions for PR validation tasks

mariomac:
Adds the following PR validation tasks:
* Lint, formatting and unit tests
* Integration tests for x86_64 and ARM64 architectures
* Kubernetes integration tests
* Licensing validation tests


marctc (approved):

marctc (commented):
Maybe in OpenTelemetry don't use Codecov and other tooling instead, but we can remove in another PR

mariomac:
@marctc I checked that other projects also use codecov, so I leaved it here.

PR #22, opened 2025-04-25
Generate and push docker images

mariomac:
Generating docker files for instrumenter and K8s cache images, to be pushed to `docker.io/otel/...`


.github/workflows/publish_dockerhub_main.yml
```diff
@@ -0,0 +1,43 @@
...
+          context: .
+          platforms: |-
+            "linux/amd64"
+            "linux/arm64"
+          tags: |-
+            "main"
+          push: true
+
+      - id: push-k8s-cache-to-dockerhub
+        uses: grafana/shared-workflows/actions/build-push-to-dockerhub@main
```
grcevski: Should we still be using these Grafana shared workflows? There might be something we can get from OTel instead?

mariomac: ouch! You are right. I forgot to modify that during the copy-paste. Good catch!

grcevski (approved):
LGTM!

.github/workflows/publish_dockerhub_main.yml
```diff
@@ -0,0 +1,53 @@
...
+
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    env:
+      IMAGE_NAME: autoinstrumentation-ebpf
```
MrAlias: This is verbose. Can we shorten this to the same name we plan to use for any binary build?

mariomac: Sure. I observed there are many `autoinstrumentation-*` images in DockerHub so I wanted to follow the same pattern. But I'm fine using something shorter as `ebpf-instrument` (same as the binary name). Is it Ok to you?

MrAlias: I'm in favor of `ebpf-instrument`.

PR #24, opened 2025-04-29
Enhance security of github actions

mariomac:
Ran [zizmor](https://github.com/woodruffw/zizmor) for each workflow file and fixed several complaints about potential security issues


rafaelroquetto (approved):

MrAlias (approved):
We should setup renovatebot to manage these pinned actions.

PR #46, opened 2025-05-14
Support k8s.job.name and k8s.cronjob.name attributes

cuichenli:
related to https://github.com/grafana/beyla/issues/1603


pkg/kubecache/meta/types.go
```diff
@@ -36,6 +36,15 @@ func ownersFrom(meta *metav1.ObjectMeta) []*informer.Owner {
...
 		}
+		if or.APIVersion == "batch/v1" && or.Kind == "Job" {
+			// we heuristically extract the CronJob name from the Job name
+			if idx := strings.LastIndexByte(or.Name, '-'); idx > 0 {
+				owners = append(owners, &informer.Owner{Kind: "CronJob", Name: or.Name[:idx]})
+				// we already have what we need for decoration and selection. Ignoring any other owner
+				// it might hypothetically have (it would be a rare case)
+				return owners
+			}
+		}
```
mariomac, line 47: Ii think you can avoid repeating the same code block if you just add another case to the condition in line 30:
```go
if (or.APIVersion == "apps/v1" && or.Kind == "ReplicaSet") ||
   (or.APIVersion == "batch/v1" && or.Kind == "Job") {
```

mariomac (changes requested):
Amazing!! I just have a minor suggestion about a code block.

Thanks for your contribution!!

pkg/kubecache/meta/types.go
cuichenli: i don't know, but if we do that, we will have another nested condition for specify the owner kind 
```golang
if (or.APIVersion == "apps/v1" && or.Kind == "ReplicaSet") ||
   (or.APIVersion == "batch/v1" && or.Kind == "Job") {
	if or.Kind == "Job" {
		....
	} else if or.Kind == "ReplicaSet" {
		...
	}
```

personally i feel this will add unnecessary complexity, but if you think that is better, i can surely update accordingly. thanks!

mariomac: Oh! you are right. I didn't notice the "subtle" difference. Then it's fine.

mariomac (approved):
🚢

cuichenli:
looks like one integration test is failing. but i checked some other prs and this failure also occured. is that a flaky test?

grcevski:
> looks like one integration test is failing. but i checked some other prs and this failure also occured. is that a flaky test?
It's a flaky test, I've kicked off a re-run.

cuichenli:
the previous code actually break one integration test, i have fixed it locally. but still i got failure on `TestMultiNodeTracingL7`. but i think this one is not related. lets see how it goes in the ci, i will try to fix this flaky test

cuichenli:
just realized the `TestMultiNodeTracingL7` has been removed in the main branch. but i am still failing at `TestMultiNodeTracing` locally.

PR #54, opened 2025-05-15
Remove prototype L7 tracer

rafaelroquetto:
This tracer was a prototype, is undocumented and unused. This also remove the tests, which should speed things up a little bit.


grcevski (approved):
LGTM!

PR #55, opened 2025-05-15
Fix offsets update workflow

### LINKED ISSUE #49: Fix offset generation action
This action is continually failing. For eample:

https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/15010282005/job/42177624917

MrAlias:
Fix #49 

Use the OTel bot as the author, committer, and signer.

https://github.com/open-telemetry/community/blob/main/assets.md#otelbot

### DIFF
--- a/.github/workflows/update-offsets.yml
+++ b/.github/workflows/update-offsets.yml
@@ -36,7 +36,7 @@ jobs:
         title: Automatic update of offsets.json
         body: The offsets have been updated by go-offsets-tracker
         base: main
-        branch: offset-content-auto-update
+        branch: otelbot/offset-content-auto-update
         labels: automated-pr
         delete-branch: true
         committer: otelbot <197425009+otelbot@users.noreply.github.com>
--- a/.github/workflows/update-offsets.yml
+++ b/.github/workflows/update-offsets.yml
@@ -24,6 +24,11 @@ jobs:
         check-latest: true
     - name: "Update offsets"
       run: make update-offsets
+    - uses: actions/create-github-app-token@v1
+      id: generate-token
+      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/update-offsets.yml
```diff
@@ -34,3 +39,7 @@ jobs:
         branch: offset-content-auto-update
```
trask: this will make it match the new branch protection I added
```suggestion
        branch: otelbot/offset-content-auto-update
```

MrAlias: Gah, lost in a merge. Thanks for catching.

mariomac:
Looks good! But I guess that then the code needs to be modified to use the format of the new Bot, otherwise the generated JSON will be incompatible with the current code.

mariomac (approved):
Looks good!

.github/workflows/update-offsets.yml
```diff
@@ -24,13 +24,22 @@ jobs:
...
         body: The offsets have been updated by go-offsets-tracker
         base: main
-        branch: offset-content-auto-update
+        branch: otelbot/offset-content-auto-update
         labels: automated-pr
         delete-branch: true
+        committer: otelbot <197425009+otelbot@users.noreply.github.com>
+        author: otelbot <197425009+otelbot@users.noreply.github.com>
+        token: ${{ steps.generate-token.outputs.token }}
+        sign-commits: true
```
trask, line 45: @MrAlias looks like `sign-commits` is causing it to fail

MrAlias: Correct. I need to submit a fix for this. Just haven't got to it yet.

PR #64, opened 2025-05-21
Start modifying the instrumenter for vendoring mode

mariomac:
This PR is open for discussion.

To integrate the BPF instrumenter within other components, such as the OTEL collector, I'm starting to experiment with some concepts to allow, not only running the instrumenter as a Go function, but allow to configuring and extending it from outside.

In the provided `examples/vendoring/vendoring.go` example file, I show how to attach a new exporter to the library. This would be useful to create a plugin within the OTEL collector that directly receives the `[]request.Span` objects from the instrument and is able to transform it into its internal OTEL representation for later transformation, filtering and export.

Many other channel overrides might be required.

Also, some types that were under an `internal/` folder now need to be part of the public API (e.g. request.Span). That explains how many files are changed (95% of then are just changes in the import paths)


pkg/instrumenter/beyla.go
```diff
@@ -1,4 +1,4 @@
-package components
+package instrumenter
```
marctc: shoud we rename this file too?

marctc (approved):

pkg/instrumenter/beyla.go
mariomac: Oh yeah! Still few "Beylas" leaked around the code 😅

marctc: yeah i can imagine...but we can rename here to obi :smile:

PR #73, opened 2025-05-22
Fix race condition in K8s informers cache

mariomac:
Since the Kubernetes event handler does not run with a defined order, using it to define the last status update of the stored entities was a bad Idea because they could be modified after the data is stored (and accessed e.g. for ordering).

This patch makes sure that the kubernetes entities are immutable, since the StatusTimeEpoch field is inferred from metadata before it is stored in the Informer storage.

Port of https://github.com/grafana/beyla/pull/1973


grcevski (approved):
LGTM!

PR #74, opened 2025-05-22
Make update offsets to work with limited otelbot permissions

trask:
The default token (`secrets.GITHUB_TOKEN`) has content write permission, so that can be used to push the branch.

That token could be used to create the PR also, but then workflows don't get triggered automatically which is annoying, and where otelbot comes in.

### DIFF
--- a/.github/workflows/update-offsets.yml
+++ b/.github/workflows/update-offsets.yml
@@ -6,7 +6,6 @@ on:
 
 permissions:
   contents: write  # Required for creating PRs
-  pull-requests: write  # Required for creating PRs
 
 jobs:
   UpdateOffsets:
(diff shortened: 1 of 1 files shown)


trask:
> Warning: update-offsets.yml:14: credential persistence through GitHub Actions artifacts: does not set persist-credentials: false

interesting...

.github/workflows/update-offsets.yml
```diff
@@ -15,7 +14,7 @@ jobs:
     - name: "Checkout repo"
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
       with:
-        persist-credentials: false
+        persist-credentials: true
```
trask, line 17: the zizmor checker seems to be ok as long as it's explicitly opted into (which makes sense to avoid people using the default not knowing better)

MrAlias (approved):

PR #75, opened 2025-05-22
[port] #1861: Ensure target_info and traces_target_info are consistent with the instrumentation state

marctc:
Original PR: https://github.com/grafana/beyla/pull/1861
cc @grcevski double check i didn't miss anything


grcevski (approved):
This looks good from a diff perspective, but I wonder if we can ensure we do the rest of the changes in this area together. The reason is that this PR would increase memory usage as we saw when people tried it with the OTel metric exporter, since now the PID was also a differentiator.

MrAlias:
@grcevski does this look good to go?

pkg/export/otel/common.go
```diff
@@ -88,6 +88,7 @@ func getResourceAttrs(hostID string, service *svc.Attrs) []attribute.KeyValue {
 		semconv.TelemetrySDKVersion(buildinfo.Version),
 		semconv.HostName(service.HostName),
 		semconv.HostID(hostID),
+		semconv.OSTypeKey.String("linux"),
```
MrAlias:
```suggestion
		semconv.OSTypeLinux,
```

https://github.com/open-telemetry/opentelemetry-go/blob/a85ae98dcedc0761078518a715dea53e519b4846/semconv/v1.32.0/attribute_group.go#L9266

grcevski (approved):
LGTM! Thanks Marc for combining these all into one PR!

PR #76, opened 2025-05-22
Prevent unnecessary K8s cache updates

mariomac:
Ported from https://github.com/grafana/beyla/pull/1977

To avoid submitting too many updates to the OBI instances, the K8s cache compares the old and new versions of each updated element that is received from the informer.

With the previous optimization, the StatusTimeEpoch field is always updated, meaning that the cmp.Equal function would always return false.

This PR also removes some updates to the StatusTimeEpoch field that should have been removed in a previous PR.


grcevski (approved):
LGTM!

PR #96, opened 2025-05-27
Temporarily restore regular expression usage in service selector

mariomac:
In order to smooth the transition from Beyla to OBI, this PR temporarily reverts the changes we did to replace regular expressions by globs.

The plan will be:
* Temporarily keep the discovery > services section with regular expressions in order to keep compatibility between configurations.
* Deprecate the discovery > services section, and warn users about its future removal.
* Create a proper discovery > instrument section that will use globs, which will be the "stable" service selection mechanism.

To facilitate the switch, this PR abstract either Regex or Glob matchers between an interface, to let the matchers easily exchange between one implementation or another, depending on the used configuration section.

### DIFF
--- a/pkg/beyla/config.go
+++ b/pkg/beyla/config.go
@@ -4,10 +4,10 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"regexp"
 	"time"
 
 	"github.com/caarlos0/env/v9"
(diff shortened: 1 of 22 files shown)


mariomac:
Yeah @grcevski , that was my idea

PR #118, opened 2025-06-05
bpf: parse mysql data in kernel space

mmat11:
This PR allows parsing mysql requests in kernel space. The benefits are that bigger queries/buffers are supported and the spans are enriched with additional data and errors based on the response.

Prepared statements support will be introduced shortly after this is approved/merged.

It has only been tested manually, let me know if additional integration/e2e testing can be done.

TODO:

- [ ] additional tests?


bpf/common/common.h
```diff
@@ -98,3 +107,66 @@ typedef struct redis_client_req {
...
+// else
+//     if (next byte = '#')
+//         string<1> sql state marker '#'
+//         string<5>sql state
+//         string<EOF> human-readable error message
+//     else
+//         string<EOF> human-readable error message
+struct mysql_response_err {
+    // https://mariadb.com/kb/en/mariadb-error-code-reference/
+    uint16_t error_code;
```
rafaelroquetto: nit: use u16, etc... instead of uint16_t as that's what we used across the code

bpf/common/common.h
```diff
@@ -98,3 +107,66 @@ typedef struct redis_client_req {
...
+    uint8_t sql_state[7];
+    char _pad;
+    uint8_t error_message[MYSQL_ERROR_MESSAGE_MAX];
+};
```
rafaelroquetto:
```suggestion
struct mysql_response_err {
    // https://mariadb.com/kb/en/mariadb-error-code-reference/
    u8 error_code;
    u8 sql_state[7];
    u8 _pad;
    u8 error_message[MYSQL_ERROR_MESSAGE_MAX];
};
```

grcevski (commented):
Thanks for this @mmat11!

I have one base question before we look more in detail here, what is the reason we'd want to parse the SQL protocol in eBPF? 

In a sense, the rule we've followed so far has been that we want to parse in C code only the protocols where we see necessary the need because of injecting trace context. In this case SQL is simply extracting data so we can report the event.

One main drawback of implementing this in C is that it's harder to test and implement additional logic. For example, one interesting aspect is the statement prepare. If we did this in userspace perhaps we can cache the preapred statement per pid/sql query and then find the sql when the statement execute is made. This would be harder in C.

Is the main reason that we have such short buffers for the TCP data and you'd like to extract more context, e.g. larger query text? I thought about this more and I wonder if we can do the following:

1. Look at some SQL like protocol detection, e.g. find out it's mysql, like you've done.
2. Introduce a variable buffer event as a type for the ring buffer. We can communicate from the userspace config as a volatile const what this max size is.
3. At the time of making the TCP event, we can create our usual event, but read more of the data now and store it after the current event. We tag the TCP event flags to say look for a follow-up event when you read it and then pull both to be processed.

I think we can reuse this approach more in general for other types of TCP events where we need larger buffers.

bpf/common/common.h
```diff
@@ -98,3 +107,66 @@ typedef struct redis_client_req {
...
+    uint8_t ssl;
+    char _pad2[7];
+    uint64_t start_monotime_ns;
+    uint64_t end_monotime_ns;
+    uint8_t command_id;
+    char _pad3[3];
+    uint32_t query_len;
+    uint8_t buf[MYSQL_QUERY_MAX];
+    tp_info_t tp;
+} mysql_request_event_t;
```
rafaelroquetto:
```suggestion

typedef struct mysql_request_event {
    u8 type; // Must be first
    u8 command_id;
    u8 ssl;
    u8 _pad;
    u32 query_len;
    u64 start_monotime_ns;
    u64 end_monotime_ns;
    connection_info_t conn_info;
    pid_info pid;
    tp_info_t tp;
    u8 buf[MYSQL_QUERY_MAX];
} mysql_request_event_t;
```

This optimises/eliminates padding, which saves us space. Don't hesitate to shuffle things around if `-Wpadded` complains, whenever there's an optimal way.

bpf/common/common.h
```diff
@@ -98,3 +107,66 @@ typedef struct redis_client_req {
...
+
+typedef struct mysql_response_event {
```
rafaelroquetto: same reasoning applies here

bpf/common/sql.h
```diff
@@ -0,0 +1,56 @@
+#pragma once
+
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_helpers.h>
+
+#define MAX_QUERY_OFFSET 4
```
rafaelroquetto: You couldn't have known that because it's not written anywhere (yet), but moving forward, we'd like to avoid macro constants in favour of `enum { k_max_query_offset = 4 };`.

bpf/common/sql.h
```diff
@@ -0,0 +1,56 @@
...
+
+static inline char lowercase(char c) {
```
rafaelroquetto: Prefer `__always_inline` - `inline` does not guarantee the compiler will actually inline the code, it just means it will apply `inline` semantics and then it may or may not actually inline it.

`__always_inline` will always inline because it's defined as:
```
#define __always_inline inline __attribute__((always_inline))
````

File: bpf/common/sql.h [hunk 7]
```diff
@@ -0,0 +1,56 @@
...
+    return (c >= 'A' && c <= 'Z') ? c + 32 : c;
+}
+
+static inline bool case_insensitive_strncmp(const char *s1, const char *s2, int n) {
```
rafaelroquetto: prefer `unsigned` types for ranges instead of `int` - i.e. here you may want to go for `u8 n` or `u16 n` depending on your use case.

bpf/common/sql.h
```diff
@@ -0,0 +1,56 @@
...
+}
+
+static inline bool is_sql_query_stmt(void *data) {
+    if (is_nullbyte(data)) {
+        return false;
+    }
+    char stmt[6];
+    if (bpf_probe_read(&stmt, 6, data)) {
+        return false;
+    }
```
rafaelroquetto: Here, there's no point on reading it twice (it just costs us more).

Instead, you could do something like this:

```
char stmt[] = "SELECT"; // SELECT here is a placeholder that means stmt can accomodate the largest word

// a few points here - notice the explicit != 0 comparsion - it's clearer than the implicit cast to bool as 0 actually means success
if (bpf_probe_read_user(&stmt, sizeof(stmt), data) != 0) {
    return false;
}
```
also notice the usage of `bpf_probe_read_user` (or _read_kernel if this is coming from kernel space instead) - `bpf_probe_read` has been deprecated since kernel 5.5

you could then explicit do
```
if (stmt[0] == '\0') { ... }
```
but your `case_insenstive_strncmp` will catch that case.

bpf/common/sql.h
```diff
@@ -0,0 +1,56 @@
...
+    return case_insensitive_strncmp(stmt, "SELECT", 6) ||
```
rafaelroquetto: I'd rename this to `strnicmp` (which is usual elsewhere) and you can do:

```
const u8 max_size = sizeof(stmt);
stricmp(stmt, ..., max_size);
```

bpf/common/sql.h
```diff
@@ -0,0 +1,56 @@
...
+           case_insensitive_strncmp(stmt, "DELETE", 6) ||
+           case_insensitive_strncmp(stmt, "CREATE", 6) ||
+           case_insensitive_strncmp(stmt, "DROP", 4) ||
+           case_insensitive_strncmp(stmt, "ALTER", 5) || case_insensitive_strncmp(stmt, "WITH", 4);
+}
+
+// Returns the index of the first character of the SQL query in the buffer.
+// Some SQL packets contain some flags which are not a part of the SQL query.
+// Returns -1 if the buffer doesn't contain an SQL query.
+static inline int is_sql_query(void *data) {
```
rafaelroquetto:
```suggestion
enum : u8 { k_max_query_offset = 4 };
enum : u8 { k_invalid_query_offset = 0xff };
static __always_inline u8 find_sql_query_offset(void *data) {
    /* ... */
     return k_invalid_query_offset;
}
```

bpf/generictracer/k_tracer_defs.h
```diff
@@ -14,6 +14,7 @@
 #include <generictracer/protocol_http.h>
 #include <generictracer/protocol_http2.h>
 #include <generictracer/protocol_tcp.h>
+#include <generictracer/protocol_mysql.h>
```
rafaelroquetto: nit: the headers are in alphabetic order

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,664 @@
...
+#include <common/tp_info.h>
+#include <common/trace_common.h>
+
+#include <generictracer/protocol_common.h>
+
+#include <maps/active_ssl_connections.h>
+
+// Every mysql command packet is prefixed by an header
+// https://mariadb.com/kb/en/0-packet/
+struct mysql_hdr {
```
rafaelroquetto, line 24: if you can, move the maps and the types to their own headers like we did in the `pid/` subdir - unfortunately it's an ongoing effort and we did not have time to clean up the rest of the code, but new code could already do that.

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,664 @@
...
+#define MYSQL_COM_STMT_PREPARE 0x16
+#define MYSQL_COM_STMT_EXECUTE 0x17
+// MySQL Server
+#define MYSQL_OK 0x0
+#define MYSQL_OK_EOF_DEPRECATED 0xfe
+#define MYSQL_ERR 0xff
+
+#define MYSQL_STATE_MARKER '#'
+
+static inline void mysql_store_state_data(connection_info_t *conn_info, const unsigned char *data) {
```
rafaelroquetto:
```suggestion
static inline void mysql_store_state_data(const connection_info_t *conn_info, const unsigned char *data) {
```

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,664 @@
...
+        bpf_probe_read(&new_state_data, MYSQL_HDR_PAYLOAD_LENGTH_SIZE, (void *)data);
+        bpf_probe_read(&new_state_data.sequence_id, MYSQL_HDR_SEQUENCE_ID_SIZE, (void *)(data + 3));
+        bpf_map_update_elem(&mysql_state, conn_info, &new_state_data, BPF_ANY);
+    } else {
+        bpf_probe_read(state_data, MYSQL_HDR_WITHOUT_COMMAND_SIZE, (void *)data);
+        bpf_map_update_elem(&mysql_state, conn_info, state_data, BPF_ANY);
+    }
+}
+
+static inline int mysql_parse_fixup_header(connection_info_t *conn_info,
```
rafaelroquetto:
```suggestion
static inline int mysql_parse_fixup_header(const connection_info_t *conn_info,
```

having said that, any chance you can split this function into smaller bits? I feel it's doing too many things (i.e. deleting data _sometimes_, and what not.

rafaelroquetto (commented):
Good stuff! Thank you so much for putting in the effort, this is much appreciated. This is a long PR with a lot of implicit knowledge, so all I did was a first pass in terms of style and other small things. Once those are addresses, we can iterate this to bring it to the finish line - also, if you have any questions, don't hesitate to bug me on Slack or elsewhere.

bpf/common/sql.h
mmat11: Thanks for the quick review! Out of curiosity, why no more macro constants?

bpf/generictracer/protocol_mysql.h
mmat11: do I need to make a mysql protocol directory? mixing these maps with other protocols' maps would be confusing imho

rafaelroquetto:
_If_ we decide to go down the parsing on userspace route, why not then use bpf_ringbuf_output()? In most cases when using bpf_ringbuf_reserve() we already copy the data anyway, so instead of doing

```
 evt *trace = bpf_ringbuf_reserve(&events, sizeof(evt), 0); 
__buitin_memcpy(trace, src, sizeof(*trace));
bpf_ringbuf_submit(trace, 0):
```

for the _mysql_ case, we could do:

```
bpf_ringbuf_output(events, trace, variable_trace_size, 0);
```

on userspace then I am assuming (for what I saw of cilium/ebpf source) that we'd have no issues reading this as a regular event as usual - so nothing changes. We could still have a "maximum size" flag as @grcevski is proposing, but IMHO that makes it simpler. So if we do decide to ship things to userspace for further processing, that would be my preference (assuming it works).

`bpf2go` is able to generate a struct with a trailing `data` field (it generates something like `Data[0] uint8`) - I am not sure if that can be directly accessed from Go, but if not, we can just do it manually (like we do in `ReinterpretCast`).

As for whether we should do that in first place, I think we should evaluate both options. 
Now, I am not sure what the best approach is here (ebps vs kernel space).

I am unsure what the best approach is here (user space vs ebpf). I'd _personally_ do things in ebpf only _provided that the complexity can be kept at bay_ as the less event overhead (less pressure in the ringbuffer, less allocation, less copies, etc...).

So what would be the objective criteria for that? And tangent to this question (because it's an issue regardless of the approach chosen) how are we planning to deal with incomplete buffers?

bpf/common/sql.h
rafaelroquetto: There are a few pitfalls. Macros can be undefined, redefined, they are not type safe, they are prone to text substitution, etc - enums are not pre-processed, but are rather type safe, can have an explicit underlying type (in this case u8), they are subject to implicit conversion rules, etc...

Not very relevant for ebpf, but enums also retain their symbolic names in debugging tools, which helps a lot

bpf/generictracer/protocol_mysql.h
rafaelroquetto: The structure we are _experimenting_ with is `tracer/{types,maps}` - so for now I'd just chuck it in there for consistency. We do need to make a pass sometime soon and finish tidying up the code - it's in my list, but it's low priority at the moment.

mmat11:
@grcevski @rafaelroquetto  I think we are discussing and trying to decide between two different topics:

1. Mysql protocol parsing in ebpf vs userspace
2. Variable length event buffers

---

1. Mysql protocol parsing in ebpf vs userspace

In my opinion, it boils down to personal preference and confidence in the implementation. Event processing in userspace might in theory be less performant, since in ebpf any unknown buffer or inconsistent data can be discarded early. (I don't have any data/numbers to back this up, though.)

About extensibility:
Any request and response is sent to userspace, so there are no technical limitations to supporting additional features such as prepared statements (with the known drawback of only knowing the statement IDs starting from the OBI process startup — but this limitation applies regardless of the implementation).

About complexity:
Given that we should keep the `is_mysql` ebpf function, what we save by implementing the rest in userspace is limited to the event sending functions. We still need the most complex parts in place, specifically, parsing the header and the most common response packet types, regardless of where we choose to implement it. And personally I also value the fact that the mysql-specific code is self contained and not mixed with other protocols, making it easier to understand the whole picture.

About testing:
While I agree that it's harder to test in general (specifically unit tests), I think most of the testing value comes from integration/e2e tests which would remain transparent to this implementation detail. The mysql (and other products) network protocols don't tend to change too often and what's worth testing I believe are the clients implementations and edge cases, which is better done in integration/e2es anyway.

---

2. Variable-length event buffers

This decision is independent of this PR, and both the "unknown TCP" path and the mysql handler could benefit from variable-length buffers. We can perhaps discuss this further in the next SIG meeting?

--- 

@rafaelroquetto Thanks again for the prompt review 🙏🏻 , I'll wait a decision on .1 before addressing the changes

grcevski:
Hi @mmat11, I think you bring a very valid point about the ability to filter in eBPF space, this would remove the overhead of sending the TCP packets for unknown protocols and make the use of the ring buffer more efficient. 

Let me summarise my thoughts here on the topic and I think I have a proposal for a way forward:

1. Personally, I think the decision between eBPF side parsing and user space parsing is not always personal preference. Where parsing at eBPF really matters is reducing the overall overhead of the protocol handling, e.g. being able to extract the useful information without passing the whole buffer to the user space. Another is the ability to parse protocols for correct handling, e.g. the HTTP2/gRPC `stream id`, without it we wouldn't be able to handle HTTP2/gRPC correctly. On the other hand, depending on the protocol it's much harder to parse and test in eBPF code than in Go user space code. For example, if we consider the complexity and the amount of changes that have been done over the course of years to the Kafka protocol, you'll see how developing that is much easier in userspace. Integration tests go only so far, unit tests are much better at testing handling 10 different protocol versions.
2. I don't have particular speed concerns on the Go userspace side, the Go compiled code is reasonably efficient for the kinds of operations we do, and it's very easy to extend the protocols, e.g. if we wanted to start parsing MongoDB we can easily use the bson library for Go. From my own benchmarking, the most expensive part if the transfer of data between the kernel and userspace.
3. Sending unknown TCP packets to userspace has its own merits, it's not always a bad thing. One example is the current backup path we have for parsing HTTP2/gRPC packets. Namely, HTTP2/gRPC clients tend to use keepalive aggressively, so it's very common that the connection is already established when OBI instruments the application. In this scenario the HTTP2 connect preamble is never seen and we use userspace parsing to detect the connection and establish the needed metadata. Without this backup path, no gRPC connections work in the OpenTelemetry DEMO for example. Also, we've had requests in the past to report unknown protocols in the traces, mostly for proprietary or esoteric protocols. While you can't see the operation, you can tell how long it took and what was the destination.
4. Having one TCP handler place is beneficial for context propagation code reuse, where we handle the detection of nested SQL/Redis/Kafka requests under HTTP for example.

Having said all of this, I think the filtering at eBPF space has a lot of merit, especially when users want this. Therefore I have a proposal that will allow us to do this, if you think it's sensible.

1. Start by adding an option to the config file that will allow us to enable eBPF side filtering. This will be passed through the eBPF side in `generictracer.go` as a constant. We'll have this off by default for now until we migrate enough of the protocol detection to eBPF side.
2. Implement protocol detectors in eBPF like you've done with MySQL, but keep it light especially for complex protocols like Kafka, because we can always finish the detection in userspace. These protocol detectors still use the TCP path, don't make new event types, but they tag the TCP event with sub-type hint to be taken in consideration in userspace.
3. Use the option in 1. to discard anything that's not detected.
4. Provide means to send variable buffer sizes, keeping the defaults as they are now. 8K is a lot for each request and users must consciously choose to enable/increase this. We've had users with 100s nested SQLs under a HTTP transaction, some ORMs do this commonly. Imagine what the overhead would be if we captured  8K for each SQL and they were running with 100 QPS? We had code in the past for capturing extra buffers, but it was removed. I'm going to try to find it. The main issue with variable buffers is that you need to capture one at request start and another on the response and you need to keep the request start around for the duration of the request. If we cannot make the types use variable buffers, what we used to do in the past is ship the larger buffer to userspace when the request start is seen, it would be held in an LRU until the full event arrives and then we use the extra data.

Thanks again for bringing this up and let me know what your thoughts are.

grcevski:
Hey Mattia, 

> Sounds good! Should we have this toggle per-protocol?

We can do that, we already have this functionality in user-space to filter out events we don't want, e.q. no SQL traces, but SQL metrics etc. Maybe we can leverage those toggles to remove the unwanted events after we've migrated most to eBPF side, if we manage to do it all. Perhaps we can leave adding this option as a last step, for now we can benefit by detecting something like MySQL in eBPF and then directing the userspace parsing directly.

> Given the fact events are different from each other, wouldn't it be more consistent to use different event types? (just a nit, I'm fine with doing both)

Sure, let's add separate event types for the TCP data type. I would just keep the data type the same and the code that handles the buffer acquisition and nesting for distributed traces. I'm suggesting this so that we minimise how much code we have to duplicate. If it's not possible to make all protocols easily fit inside the instruction limit, perhaps we can tail call to say MySQL handler and then tail call to the TCP event with custom type.

> By discard you mean let the "unknown TCP" fallback handle it?

Yeah let's keep the unknown TCP fall back as it is today. If we manage to migrate most protocol parsing to eBPF space we can decide to add the option to not send the TCP fallback. Kafka is a lot of parsing code, so we need to add C unit tests. HTTP2/gRPC is also a lot of work for the fall back code we use at the moment in userspace. Redis and MongoDB (which someone is adding at the moment) are simple enough.

mmat11:
hi @grcevski and @rafaelroquetto , I did as discussed here and offline, the PR needs a bit of polishing and perhaps I missed some cleanups like moving the maps in their files as suggested by Rafael, but it can be re-reviewed now

bpf/common/http_types.h
```diff
@@ -45,37 +42,25 @@
...
-    u32 req_len;
-    u32 resp_len;
-    unsigned char buf[K_TCP_MAX_LEN];
-    unsigned char rbuf[K_TCP_RES_LEN];
-    // we need this to filter traces from unsolicited processes that share the executable
-    // with other instrumented processes
-    pid_info pid;
-    tp_info_t tp;
-} tcp_req_t;
+enum protocol_type {
```
rafaelroquetto:
```suggestion
enum protocol_type : u8 {
```

bpf/common/common.h
```diff
@@ -98,3 +114,37 @@ typedef struct redis_client_req {
...
 } redis_client_req_t;
+
+// Here we track unknown TCP requests that are not HTTP, HTTP2 or gRPC
+typedef struct tcp_req {
+    u8 flags; // Must be fist we use it to tell what kind of packet we have on the ring buffer
+    u8 ssl;
+    u8 direction;
+    u8 has_large_buffers;
+    u8 _pad1[4];
+    enum protocol_type protocol_type;
```
rafaelroquetto:
```suggestion
    enum protocol_type protocol_type;
    u8 _pad1[3];
```

(see my comment on protocol_type)

bpf/common/common.h
```diff
@@ -98,3 +114,37 @@ typedef struct redis_client_req {
...
+    pid_info pid;
+    u8 _pad3[4];
+    tp_info_t tp;
+} tcp_req_t;
+
+typedef struct tcp_large_buffer {
+    u8 type; // Must be first
+    u8 direction;
+    u8 _pad[2];
+    u32 len;
```
rafaelroquetto, line 145:
```suggestion
    u16 len;
```

bpf/common/sql.h
```diff
@@ -0,0 +1,52 @@
...
+}
+
+static __always_inline bool stricmp(const char *s1, const char *s2, u8 n) {
+    for (int i = 0; i < n && s1[i] && s2[i]; i++) {
+        if (lowercase(s1[i]) != lowercase(s2[i])) {
+            return false;
+        }
+    }
+    return true;
+}
```
rafaelroquetto: nice! can we move this to a string_utils.h header of some sort? And if you would be willing to go the extra mile (but perhaps outside the scope of this PR), replace the usage of `bpf_memicmp` elsewhere in the code with this

bpf/common/sql.h
```diff
@@ -0,0 +1,52 @@
...
+
+static __always_inline bool is_sql_query_stmt(void *data) {
+    if (is_nullbyte(data)) {
+        return false;
+    }
+    char stmt[6];
+    if ((bpf_probe_read_kernel(&stmt, 6, data)) != 0) {
+        return false;
+    }
```
rafaelroquetto: The `is_nullbyte` definitely improves clarity, but unfortunately it's a bit expensive to use in this context, because it means we'd be calling `bpf_probe_read_kernel` twice - once for `is_nullbyte` and another for reading into `stmt`. (`bpf_probe_read_kernel does quite a few non-neglectable  checks and copies).

We could just do

```
char stmt[] = "SELECT"; // largest string we want to read
const size_t stmt_size = sizeof(stmt) - 1;

bpf_probe_read_kernel(&stmt, read_size, data)
```
`bpf_probe_read_kernel` will ultimately copy zeroes if it can't read, so you can just do the usual `stricmp` check as usual

bpf/common/sql.h
```diff
@@ -0,0 +1,52 @@
...
+    return stricmp(stmt, "SELECT", 6) || stricmp(stmt, "INSERT", 6) || stricmp(stmt, "UPDATE", 6) ||
+           stricmp(stmt, "DELETE", 6) || stricmp(stmt, "CREATE", 6) || stricmp(stmt, "DROP", 4) ||
+           stricmp(stmt, "ALTER", 5) || stricmp(stmt, "WITH", 4);
```
rafaelroquetto: I'd declare those as constants, you can define them toplevel in the file as
```
static const char alter_str[] = "ALTER";
```
and then you can do `stricmp(stmt, alter, sizeof(alter_str) - 1)` for instance (feel free to define a size constant as well)

bpf/common/sql.h
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// Returns the index of the first character of the SQL query in the buffer.
+// Some SQL packets contain some flags which are not a part of the SQL query.
+// Returns -1 if the buffer doesn't contain an SQL query.
+static __always_inline int is_sql_query(void *data) {
```
rafaelroquetto:
```suggestion
static __always_inline int find_sql_query(void *data) {
```

parts of our code tend to use the `find_` prefix when the function has the semantics you describe

bpf/common/sql.h
```diff
@@ -0,0 +1,52 @@
...
+    for (int i = 0; i < k_max_query_offset; i++) {
```
rafaelroquetto:
```suggestion
    for (u16 i = 0; i < k_max_query_offset; i++) {
```

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,492 @@
...
+    __uint(value_size, k_large_buf_max_size);
+    __uint(max_entries, 1);
+} mysql_large_buffers SEC(".maps");
+
+static __always_inline void mysql_store_state_data(const connection_info_t *conn_info,
+                                                   const unsigned char *data) {
+    struct mysql_state_data *state_data = bpf_map_lookup_elem(&mysql_state, conn_info);
+    if (state_data == NULL) {
+        struct mysql_state_data new_state_data = {};
+        bpf_probe_read_kernel(&new_state_data, k_mysql_hdr_without_command_size, (void *)data);
```
rafaelroquetto:
```suggestion
        bpf_probe_read_kernel(&new_state_data, k_mysql_hdr_without_command_size, (const void *)data);
```

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,492 @@
...
+        bpf_map_update_elem(&mysql_state, conn_info, &new_state_data, BPF_ANY);
+    } else {
+        bpf_probe_read_kernel(state_data, k_mysql_hdr_without_command_size, (void *)data);
```
rafaelroquetto:
```suggestion
        bpf_probe_read_kernel(state_data, k_mysql_hdr_without_command_size, (const void *)data);
```

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,492 @@
...
+                                                    struct mysql_hdr *hdr,
+                                                    const unsigned char *data,
+                                                    size_t data_len) {
+    struct mysql_state_data *state_data = bpf_map_lookup_elem(&mysql_state, conn_info);
+    if (state_data != NULL) {
+        bpf_dbg_printk("mysql_parse_fixup_header: state data found");
+
+        hdr->payload_length = state_data->payload_length;
+        hdr->sequence_id = state_data->sequence_id;
+        bpf_probe_read_kernel(&hdr->command_id, k_mysql_hdr_command_id_size, (void *)data);
```
rafaelroquetto:
```suggestion
        bpf_probe_read_kernel(&hdr->command_id, k_mysql_hdr_command_id_size, (const void *)data);
```

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,492 @@
...
+        }
+        bpf_probe_read_kernel(hdr, k_mysql_hdr_size, (void *)data);
+    }
+    return 0;
+}
+
+// This is an alternative version of mysql_parse_fixup_header that fills the buffer
+// without reading header fields.
+static __always_inline int mysql_read_fixup_buffer(const connection_info_t *conn_info,
+                                                   u8 *buf,
```
rafaelroquetto:
```suggestion
                                                   unsigned char *buf,
```

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,492 @@
...
+                                                   u32 *buf_len,
+                                                   const unsigned char *data,
+                                                   size_t data_len,
+                                                   bool delete_data) {
+    u8 offset = 0;
+    u8 buf_len_mask = mysql_buffer_size - 1; // mysql_buffer_size is guaranteed to be a power of 2
```
rafaelroquetto:
```suggestion
    const u8 buf_len_mask = mysql_buffer_size - 1; // mysql_buffer_size is guaranteed to be a power of 2
```

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,492 @@
...
+
+    struct mysql_state_data *state_data = bpf_map_lookup_elem(&mysql_state, conn_info);
+    if (state_data != NULL) {
+        bpf_dbg_printk("mysql_fill_buffer: state data found");
+
+        bpf_probe_read_kernel(buf, k_mysql_hdr_without_command_size, (void *)state_data);
```
rafaelroquetto:
```suggestion
        bpf_probe_read_kernel(buf, k_mysql_hdr_without_command_size, (const void *)state_data);
```

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,492 @@
...
+        if (data_len < k_mysql_hdr_size) {
+            bpf_dbg_printk("mysql_fill_buffer: data_len is too short: %d", data_len);
+            return -1;
+        }
+    }
+
+    *buf_len = (data_len + offset) > mysql_buffer_size ? mysql_buffer_size : (data_len + offset);
+    if (*buf_len == mysql_buffer_size) {
+        bpf_dbg_printk("WARN: mysql_fill_buffer: buffer is full, truncating data");
+    }
```
rafaelroquetto:
```suggestion
    *buf_len = data_len + offset;
    
    if (*buf_len > mysql_buffer_size) {
        *buf_len = mysql_buffer_size;
        bpf_dbg_printk("WARN: mysql_fill_buffer: buffer is full, truncating data");
    }
```

rationale: since we need the `if` anyway, at least we make a single conditional (otherwise we have two when debugging is enabled).

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,492 @@
...
+
+    bpf_probe_read_kernel(buf + offset, *buf_len & buf_len_mask, (void *)data);
```
rafaelroquetto:
```suggestion
    bpf_probe_read_kernel(buf + offset, *buf_len & buf_len_mask, (const void *)data);
```

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,492 @@
...
+
+    return *buf_len;
+}
+
+static __always_inline void mysql_send_large_buffer(
```
rafaelroquetto: can we also make the len here unsigned (i.e. use unsigned in general for lengths / sizes with the most appropriate size, e.g. u16 or u32 here), and const the buf when possible?

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,492 @@
...
+        bpf_dbg_printk("mysql_send_large_buffer: 4 bytes packet, storing state data");
+        mysql_store_state_data(&pid_conn->conn, u_buf);
+        return;
+    }
+    if (bytes_len < (k_mysql_hdr_size + 1)) {
+        bpf_dbg_printk("mysql_send_large_buffer: bytes_len is too short: %d", bytes_len);
+        return;
+    }
+
+    tcp_large_buffer_t *large_buf = bpf_map_lookup_elem(&mysql_large_buffers, &(u32){0});
```
rafaelroquetto: these unnamed temporaries are cool! :)

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,492 @@
...
+ */
+static bool is_eof_packet(u8 command_id, size_t data_len) {
+    return command_id == k_mysql_ok_eof_deprecated;
+}
+
+/**
+ * https://mariadb.com/kb/en/protocol-data-types/#length-encoded-integers
+ * returns the length of the encoded integer, or -1 if invalid
+ */
+static int parse_length_encoded_integer(const u8 *data, size_t data_len) {
```
rafaelroquetto:
```suggestion
static int parse_length_encoded_integer(const unsigned char *data, size_t data_len) {
```

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,492 @@
...
+            data_len - offset);
+        return false;
+    }
+
+    return (command_id == k_mysql_ok || command_id == k_mysql_ok_eof_deprecated);
+}
+
+static __always_inline bool
+is_last_packet_eof(struct mysql_hdr *hdr, const u8 *data, size_t data_len) {
+    const u8 eof;
```
rafaelroquetto:
```suggestion
    u8 eof;
```

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,492 @@
...
+ * 11, 13, 14, 15, 16, 17, 19, 21, 25
+ */
+#define CHECK_OK_PACKET_FROM_OFFSET(offset)                                                        \
+    ok = data + data_len - offset;                                                                 \
+    bpf_probe_read_kernel(&command_id, 1, ok + command_offset);                                    \
+    if (is_ok_packet(hdr, ok, offset, command_id)) {                                               \
+        return true;                                                                               \
+    }
+
+static bool is_last_packet_ok(struct mysql_hdr *hdr, const u8 *data, size_t data_len) {
```
rafaelroquetto:
```suggestion
static bool is_last_packet_ok(struct mysql_hdr *hdr, const unsigned char *data, size_t data_len) {
```

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,492 @@
...
+    u8 command_offset = mysql_command_offset(hdr);
+    CHECK_OK_PACKET_FROM_OFFSET(11);
+    CHECK_OK_PACKET_FROM_OFFSET(13);
+    CHECK_OK_PACKET_FROM_OFFSET(14);
+    CHECK_OK_PACKET_FROM_OFFSET(15);
+    CHECK_OK_PACKET_FROM_OFFSET(16);
+    CHECK_OK_PACKET_FROM_OFFSET(17);
+    CHECK_OK_PACKET_FROM_OFFSET(19);
+    CHECK_OK_PACKET_FROM_OFFSET(21);
+    CHECK_OK_PACKET_FROM_OFFSET(25);
```
rafaelroquetto: can we find a way to replace this with a bulk read? could we read `data` into a single large buffer and process it from there, to avoid all of these expensive 1 byte reads?

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,492 @@
...
+    return false;
+}
+
+static int parse_response_packet_type(struct mysql_hdr *hdr, const u8 *data, size_t data_len) {
```
rafaelroquetto: for instance, could we have a buffer here at this stage in which we read data, and all of the parsing functions operate on that buffer rather than a cascade of `bpf_probe_read_kernel`? Or at least read bigger chunks?

bpf/common/sql.h
mmat11: I'll create the header in this PR and replace it in another one, opened https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/195 so I won't forget

bpf/common/sql.h
mmat11: I removed the `is_nullbyte` function and just checked the first stmt byte after reading data, if that works for you

bpf/generictracer/protocol_mysql.h
mmat11: I changed `is_last_packet_ok()` to read 25 bytes from the end on the stack so we can avoid all the small `bpf_probe_read*`.
Do you mind if we do this other optimization in a separate PR? From what I see we need 2 separate chunks as sometimes we read from the start of the buffer and sometimes from the end, doing it on the stack would be ok but ideally I'd put these chunks in a map

bpf/common/sql.h
```diff
@@ -0,0 +1,59 @@
...
+enum stmt_lengths : u8 {
+    k_stmt_with_len = 4,
+    k_stmt_drop_len = 4,
+    k_stmt_alter_len = 5,
+    k_stmt_select_len = 6,
+    k_stmt_insert_len = 6,
+    k_stmt_update_len = 6,
+    k_stmt_delete_len = 6,
+    k_stmt_create_len = 6
+};
```
rafaelroquetto: you should really key these off the constants (e.g. k_stmt_with_len = sizeof(with_stmt) - 1) — it's less error-prone. If someone changes this later, say to support "COMMENT", they'll need to manually count characters. An off-by-one mistake might not be obvious and could be hard to track down.

bpf/common/sql.h
```diff
@@ -0,0 +1,59 @@
...
+
+static __always_inline bool is_sql_query_stmt(void *data) {
+    char stmt[6];
```
rafaelroquetto: same with these magic constants here

you could define a constant, e.g. `k_largest_statement_len = sizeof(create_stmt) - 1;` - it makes it more obvious down the road

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,490 @@
...
+
+#include <generictracer/protocol_common.h>
+#include <generictracer/k_tracer_tailcall.h>
+
+#include <maps/active_ssl_connections.h>
+
+// Every mysql command packet is prefixed by an header
+// https://mariadb.com/kb/en/0-packet/
+struct mysql_hdr {
+    u32 payload_length : 24;
```
rafaelroquetto: I realised there's an issue here - the MySQL protocol defines integers to always be little-endian - using bitfields may break that assumption on big endian architectures.

Also, there are more general issues that can come to bite: the alignment of bitfields is unspecified, the compiler may choose to add padding there, etc... we shouldn't be using it, instead, we should do

```
u8 payload_lenth[3];
```

and then have a helper that returns something like
```

	return ((u32) hdr->payload[0])
		| ((u32) hdr->payload[1] << 8)
		| ((u32) hdr->payload[2] << 16);
```

for an extended rationale, see the "Notes" section at https://cppreference.com/w/c/language/bit_field.html

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,490 @@
...
+    u8 sequence_id;
+    u8 command_id;
+
+    // Metadata
+    bool hdr_arrived; // Signals whether to skip or not the first 4 bytes in the current buffer as
+                      // they arrived in a previous packet.
+} __attribute__((packed));
```
rafaelroquetto: reading about the mysql header, it's best to actually define two structs whose size you can take with `sizeof()` instead of manually setting the size constants above, e.g.

```
struct mysql_hdr {
    u8 payload_length[3]; // little-endian 24-bit value
    u8 sequence_id;
};
// can now write sizeof(mysql_hdr) rather than k_mysql_hdr_size

struct mysql_command_packet { // usually client requests
    struct mysql_hdr hdr;
    u8 command_id;
};
// ditto for sizeof(mysql_command_packet)
```

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,490 @@
...
+
+static __always_inline u8 is_mysql(connection_info_t *conn_info,
+                                   const unsigned char *data,
+                                   u32 data_len,
+                                   u8 *packet_type,
+                                   enum protocol_type *protocol_type) {
+    u8 response_packet_type;
+    int packet_type_response;
+
+    if (data_len == k_mysql_hdr_without_command_size) {
```
rafaelroquetto: this would then read
```suggestion
    if (data_len == sizeof(struct mysql_hdr)) {
```

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,490 @@
...
+        bpf_dbg_printk("is_mysql: 4 bytes packet, storing state data");
+        mysql_store_state_data(conn_info, data);
+        return 0;
+    }
+
+    if (data_len < (k_mysql_hdr_size + 1)) {
```
rafaelroquetto, line 225:
```suggestion
    if (data_len < sizeof(struct mysql_command_packet) + 1) {
```

bpf/generictracer/protocol_mysql.h
rafaelroquetto: what happens if you get a subsequence packet that is also 4 bytes long? Imagine the first packet is the header, and then for the same connection info, you get a 4 byte payload (e.g. header + `COM_QUERY "SET"`), you will overwrite the header info here.
What you could do instead is:

1. check if you already have stored a header in the map
2. perhaps check if the sequence number is sane

bpf/generictracer/protocol_mysql.h
rafaelroquetto, line 225: then here, you should bail _only_ if you don't have any previously stored header

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,490 @@
...
+        bpf_dbg_printk("is_mysql: data_len is too short: %d", data_len);
+        return 0;
+    }
+
+    struct mysql_hdr hdr = {};
```
rafaelroquetto, line 230: you could then ditch the `hdr_arrived` metadata, and also this variable - at this point, either you have a previously stored header (what comes next is a payload) or you don't, and `data` contains the header + a payload chunk.

 I think what I am trying to say is something along the lines of:

```
    struct mysql_hdr *hdr = find_mysql_hdr(info);

    if (hdr == NULL && data_len < sizeof(*hdr)) {
        bpf_dbg_printk("not mysql");
        return 0;
    }

    // consume header if we don't have any yet
    if (hdr == NULL) {
        hdr = new_mysql_hdr(info);
        bpf_probe_read_kernel(hdr, ...);
        data_len -= sizeof(*hdr);
    }

    // nothing left to read
    if (data_len == 0) {
        return 0;
    }

   // ---> at  this stage, you have a header and some data to consume
```

as far as I know  MySQL protocol is strictly request->response - we should leverage that, so that once you've parsed a command (request), you remember that, and then you only try to parse a response if you have successfully parsed a request of that command of interest - otherwise, you can drop the orphan response packet and bail early, which is good.

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,490 @@
...
+    case k_mysql_com_stmt_execute:
+        // COM_STMT_EXECUTE packet structure:
+        // +------------+-------------+----------------------+
+        // | payload_len| sequence_id | command_id | stmt_id |
+        // +------------+-------------+----------------------+
+        // |    3B      |     1B      |     1B     | 4B      |
+        // +------------+-------------+----------------------+
+        *packet_type = PACKET_TYPE_REQUEST;
+        break;
+    default:
```
rafaelroquetto, line 272: so if we haven't found any request, are we interested on "orphan" responses?

bpf/generictracer/protocol_mysql.h
```diff
@@ -0,0 +1,490 @@
...
+        // We store this state data to be able to parse the full header later.
+        bpf_dbg_printk("mysql: 4 bytes packet, storing state data");
+        mysql_store_state_data(&args->pid_conn.conn, (const unsigned char *)args->u_buf);
+        return 0;
+    }
+
+    // Tail call back into generic TCP handler.
+    // Once the header is fixed up, we can use the generic TCP handling code
+    // in order to reuse all the common logic.
+    bpf_tail_call(ctx, &jump_table, k_tail_protocol_tcp);
```
rafaelroquetto, line 210: at this stage, we should have a header and a payload ready to be consumed, so couldn't we simply construct a specific event that contains the relevant bits of `tcp_req` plus the "raw buffer"?

otherwise, are we not sending two events, the `tcp_large_buffer` one and the `tcp_req` one? If that's the case, it means we are reading `u_buf` multiple times - once for the large buffer, and once as part of of the original TCP codepath (which reads K_TCP_MAX_LEN=256) and puts it in the ringbuffer...

If my understanding is correct (and it may as well not be), then we should instead really dispatch everything just once via a large buffer to userspace and skip the common TCP codepath altogether.

rafaelroquetto (commented):
This is a big PR with a lot of implicit knowledge, so I hope you don't mind me having to iterate this.

bpf/common/sql.h
mmat11: I'm adding some macros so we just need to define the new operation in the future

bpf/generictracer/protocol_mysql.h
mmat11: don't we only support amd64 and arm64?

regarding alignment, is that true for explicitly packed structures? I have `__attribute__((packed))` there

bpf/generictracer/protocol_mysql.h
mmat11: orphan responses will be dropped both later in the tcp codepath and also in userspace, I can add another check here if you prefer

bpf/generictracer/protocol_mysql.h
mmat11: It was agreed offline to reuse the tcp code path in order to have common code (such as traceparent) and edge cases already figured out (also see: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/118#issuecomment-2954306735) and maximize code reuse at the cost of one additional tail call.

On the "non-large buffer" part, given the tcp code path uses the reserve/commit API, the event we send is still fixed size so we just save a copy by not using the fixed buffer. I kept it in order to use it as fallback in the case the large buffer is not found in the LRU

bpf/generictracer/protocol_mysql.h
rafaelroquetto: > don't we only support amd64 and arm64?
we do, until we don't... but this is just one of the issues, overall bitfields are problematic - if you look at the "Notes" section of the link above, you will see we'd be relying on a lot of undefined and implementation defined behaviour, many of which are hard to spot and to enforce (the compiler will allow the code to compile, may not even emit a warning, but then the code misbehaves).

> regarding alignment, is that true for explicitly packed structures? I have __attribute__((packed)) there

here's an example of bitfields in a packed structure, in which the bitfieds themselves are not packed: https://godbolt.org/z/7zbaMrz9M

bpf/generictracer/protocol_mysql.h
rafaelroquetto: yeah, we should drop them early - there's no point on running all that code downstream just to drop them

bpf/generictracer/protocol_mysql.h
rafaelroquetto: These aren't mutually exclusive in my view - we could (and should) move the reusable part of the code to a common place that can be invoked by both codepaths. But in my opinion there's no upside whatsoever on doing multiple passes on `u_buf` and submitting duplicated information (both events) when it's perfectly feasible not to do so - otherwise this is premature pessimisation in a hotpath (`tcp_sendmsg` and `tcp_recvmsg` are invoked at a very high frequency). Separating things would also improve clarity on an otherwise dense code.

bpf/generictracer/protocol_mysql.h
mmat11: I don't see padding being added there 🤔 , the size is 5 because of the bitfield (3b => 1 byte) + the uint32 (4 bytes) and the offset is 1. If you remove the packed attribute, the size is 8 and the offset is 4 as it should be.
My bitfield size is also a multiple of 8 so there should be no issues doing that

bpf/generictracer/protocol_mysql.h
grcevski: What you are proposing @rafaelroquetto is not easy to do with our current way of the way the code is structured. How about we keep it as is and then we can follow-up with a PR to clean that part up. I feel like the changes are getting too large anyway. Nothing stops us to iterate on this, it doesn't have to be the final state.

pkg/components/ebpf/common/tcp_large_buffer.go
```diff
@@ -0,0 +1,61 @@
...
+	largeBufferKey struct {
+		traceID   [16]uint8
+		direction uint8
+	}
+	largeBuffer struct {
+		buf []byte
+	}
+)
+
+var largeBuffers = expirable.NewLRU[largeBufferKey, largeBuffer](1024, nil, 5*time.Minute)
```
grcevski: Most of the changes look good IMO now, I will do another review soon to see if there are other things I might have missed, but we should move these large buffers in the `EBPFParseContext` structure, rather than making them static in the file. This is in preparation for OBI being able to be vendored in the OTel collector in the future, where essentially we might expect the OBI component to be reloaded. We used to have these statics everywhere, now they are all removed.

grcevski:
> * [ ]  additional tests?

I believe we are good on integration tests for this PR, since we test MySQL and as long as it doesn't break things should be good. I'm going to spend some time today looking to see if I can tell why some of these tests are failing if you need help. 

One thing I would do is add unit tests for the new bits of the IF logic in the request handling, just to ensure we handle the path correctly.

Ideally we should add integration tests for the variable buffers, but that can follow in another PR, since it's not blocking the default use of the product.

bpf/generictracer/protocol_mysql.h
rafaelroquetto: What I meant is exactly that, sorry if the example was crappy - as a multiple of 8, your bitfield neatly aligns to the next byte, so there are no "spare" bits, which affects the size of the struct. If you pass `-Wpadded` to the example, you will see

> warning: padding struct 'struct Test' with 3 bits to align 'b' [-Wpadded]

Even if there are no alignment issues with your bitfield (which is not guaranteed), the usage is still probe to all of the other issues described above.

bpf/generictracer/protocol_mysql.h
rafaelroquetto: I don’t feel comfortable merging this as-is for two main reasons:

1. Duplicate events: We’re emitting both the large-buffer event and the original TCP event with a fixed payload.
2. Excessive bpf_probe_read_kernel(): We call `bpf_probe_read_kernel()` far more often than necessary

Originally, I understood the plan to be a single variable-buffer event into userspace—so I may be missing something.  

That said, if you guys prefer to move forward as-is and address it in subsequent PRs, I won’t block it.

rafaelroquetto: @mmat11, I spoke to @grcevski and he gave me the entire rationale on why it was decided to use two events, so it makes sense to me now.

rafaelroquetto:
@mmat11 another thing I realised is that sometimes the buffer we are reading may be userspace memory, not always kernel - if you suspect the failing tests could be related to this (and I don't mean to send you on the wrong path), perhaps we should consider using `bpf_probe_read()` after all (instead of `_read_kernel()`), as it will dispatch to the correct `_kernel` or `_user` variant.

bpf/generictracer/protocol_mysql.h
mmat11: I'll do it in this PR

bpf/generictracer/protocol_mysql.h
mmat11: @rafaelroquetto nice! I'll address the excessive usage of `bpf_probe_read` today so we can hopefully get this in

bpf/generictracer/protocol_mysql.h
mmat11: > check if you already have stored a header in the map

I anyway have to assume that no broken/manually crafted packets are part of this stream or we can't have any sufficient number of checks in place anyway

> perhaps check if the sequence number is sane

how would I do that if I got only an header? or by "sane" you mean a normal number (in that case, what should we define as sane? :D)

rafaelroquetto: > I anyway have to assume that no broken/manually crafted packets are part of this stream or we can't have any sufficient number of checks in place anyway

It doesn't have to be perfect (it would be difficult to do so as you noticed), but you can try to mitigate it as much as possible - we face the same issue elsewhere - for instance, for the IPC code (ebpf_ipc.h) we mitigate that through a checksum and other sanity checks on the size.

> how would I do that if I got only an header? or by "sane" you mean a normal number (in that case, what should we define as sane? :D)

if you only get a header you accept it (since you can infer anything) - but once you have one as an anchor, you can use it as a reference

bpf/generictracer/protocol_mysql.h
mmat11: I removed all the complex parts coming from parsing the response packet and added some checks on the sequence id, hopefully it's good enough now

pkg/components/ebpf/common/tcp_large_buffer.go
mmat11: Added the map in EBPFParseContext 👍🏻

bpf/generictracer/protocol_mysql.h
mmat11: > if you only get a header you accept it (since you can infer anything) - but once you have one as an anchor, you can use it as a reference

I don't understand what would this fix then, if we take into account these cases:

1. I get 4 bytes, they are a valid header -> I proceed to read the next packet 
2. I get 4 bytes, I then get 4 bytes again -> I should assume the first 4 bytes were a valid header else something is wrong in the connection stream and I'm getting split packets which are broken. If the second 4 bytes are a payload, this is handled, if they are a new header.. how should I know that? I need the command id which will come in a future packet (it's the fifth byte) and I can't do assumption on the first byte as the command IDs could overlap with the first byte of the payload_length field

on top of this, I think it's very difficult to have a request long exactly 8 bytes (so it gets split in 4 + 4), I was not able to reproduce this by using 0x03+SET (but I also didn't manually craft the packet since it's not in scope)

rafaelroquetto: I guess what I am saying is, what the present code does:

 1. you receive 4 bytes - `if (data_len == k_mysql_hdr_without_command_size)` hits and you consume it
 2. then you receive another 4 bytes - `if (data_len == k_mysql_hdr_without_command_size)` hits and you consume that too - but those 4 bytes were likely payload, not header, so you've effectively skipped over them - Instead you end up overwriting the previous mysql_state_data

unless I am misunderstanding the code, this seems like a bug

A safer approach might be:

1. On receiving 4 bytes, check if a header is already partially stored.
2. If not, treat it as a header.
3. If yes, treat it as payload.

one way to phrase this is https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/118/files#r2157916013

> on top of this, I think it's very difficult to have a request long exactly 8 bytes (

I agree it’s not trivial to hit exactly 8 bytes, but since TCP can fragment at any offset, even at 8 bytes, so the code needs to handle that possibility regardless. We've been bitten by all kinds of weird corner-cases before, so I'd rather guard against that.

grcevski:
So this is weird, I don't know why the latest failure :). It's a Go application which unless we impacted some userspace stuff shouldn't be failing. I looked at the failing test:

```
	// Eventually, Prometheus would make this query visible
	pq := prom.Client{HostPort: prometheusHostPort}
	var results []prom.Result
	test.Eventually(t, testTimeout, func(t require.TestingT) {
		var err error
		results, err = pq.Query(`http_server_request_duration_seconds_count{` +
			`http_request_method="GET",` +
			`http_response_status_code="404",` +
			`service_namespace="` + svcNs + `",` +
			`service_name="` + svcName + `",` +
			`server_port="` + serverPort + `",` +
			`http_route="/basic/:rnd",` +
			`url_path="` + path + `"}`)
		require.NoError(t, err)
		// check duration_count has 3 calls and all the arguments
		enoughPromResults(t, results)
		val := totalPromCount(t, results)
		assert.LessOrEqual(t, 3, val)
		if len(results) > 0 {
			res := results[0]
			addr := res.Metric["client_address"]
			assert.NotNil(t, addr)
			assert.NotNil(t, res.Metric["server_port"])
		}
	})
```

and the logs show printed trace like this:

```
HTTP 404 GET /basic/8129 [172.18.0.1 as 172.18.0.1:35624]->[172.18.0.18 as 172.18.0.18.initial-set:8080] contentLen:38B responseLen:0B svc=[initial-set/testserver go] traceparent=[00-875ab79e55cb53ce15ee0c87e39ae241-5da3f19300cacce9[0000000000000000]-01]
```
The code is correct, server port is 8080 (after the -> ), namespace and name are initial-set and testserver. I see 4 prints in the logs, so we got all calls to run...

test/integration/red_test_python_sql.go
```diff
@@ -73,13 +73,45 @@ func testREDMetricsForPythonSQLLibrary(t *testing.T, url, comm, namespace string
 	require.Empty(t, traces)
 }
 
+func testREDMetricsForPythonSQLLibraryError(t *testing.T, url, comm string) {
+	urlPath := "/error"
+
+	for i := 0; i < 4; i++ {
+		doHTTPGet(t, url+urlPath, 200)
```
grcevski: Great test, I checked the failures and it does what it's supposed to! :)) We just need to change the 200 here to 500, because I think based on the python code we get 500 on the application side:
```
autoinstrumenter-1  | 2025-06-27 13:06:26.6271626 (408.52µs[408.52µs]) SQLClient 0 SELECT nonexisting [172.18.0.6 as 172.18.0.6.integration-test:57262]->[172.18.0.3 as 172.18.0.3:5432] contentLen:31B responseLen:0B svc=[integration-test/python3.12 python] traceparent=[00-99987568e36f1efdda949fe64ec6e921-4038d4cbd21b27d8[3d7d0e9edc0e5c72]-00]
autoinstrumenter-1  | 2025-06-27 13:06:26.6271626 (1.249397ms[1.249397ms]) HTTP 500 GET /error [172.18.0.1 as 172.18.0.1:52812]->[172.18.0.6 as 172.18.0.6.integration-test:8080] contentLen:190B responseLen:174B svc=[integration-test/python3.12 python] traceparent=[00-99987568e36f1efdda949fe64ec6e921-3d7d0e9edc0e5c72[0000000000000000]-01]

```

mmat11: yeah this one is "intentionally wrong", I'm fixing it offline; good thing is that the PHP test is now fixed, I forgot to revert a single line and it broke the probe 🙈

bpf/generictracer/protocol_mysql.h
mmat11: I get your point, changed the logic

test/integration/red_test_python_sql.go
mmat11: I commented out this test because the docker-compose file it refers to uses Postgres. I'll uncomment when I do the same changes for Postgres

mmat11:
@rafaelroquetto I _think_ I did all the changes except the removal of `hdr_arrived`; I am using it in the `is_mysql` function too and removing that just to not use constants is a bit too messy. Could you please have another look?

Also, not sure the failing k8s test is because of the changes

grcevski:
It seems like we are hitting the instruction limit for the new mysql program on older kernels (5.15 in this case). The error is as below:
```
autoinstrumenter-1      | parent already had regs=80 stack=0 marks
autoinstrumenter-1      | ; if (lowercase(s1[i]) != lowercase(s2[i])) {
autoinstrumenter-1      | 5096: (bf) r2 = r1
autoinstrumenter-1      | 5097: (57) r2 &= 255
autoinstrumenter-1      | ; if (lowercase(s1[i]) != lowercase(s2[i])) {
autoinstrumenter-1      | 5098: (15) if r2 == 0x69 goto pc+1
autoinstrumenter-1      | 
autoinstrumenter-1      | from 5098 to 5100: R0=inv0 R1=inv(id=153650,smin_value=-128,smax_value=64) R2=inv105 R3=inv64 R6=inv0 R7=inv91 R8=map_value(id=0,off=0,ks=4,vs=88,imm=0) R9=inv100 R10=fp0 fp-16=????mmmm fp-48=??mmmmmm fp-88=??mmmmmm fp-128=mmmmmmmm fp-136=mmmmmmmm fp-144=mmmmmmmm fp-152=mmmmmmmm fp-160=mmmmmmmm fp-168=mmmmmmmm fp-192=????mmmm fp-200=mmmmmmmm fp-208=00000000 fp-216=mmmmmmmm fp-224=00000000 fp-232=mmmmmmmm fp-240=map_value fp-248=ctx fp-256=map_value fp-264=map_value fp-272=inv512
autoinstrumenter-1      | ; for (u16 i = 0; i < n && s1[i] && s2[i]; i++) {
autoinstrumenter-1      | 5100: (71) r3 = *(u8 *)(r10 -87)
autoinstrumenter-1      | 5101: (67) r3 <<= 56
autoinstrumenter-1      | 5102: (c7) r3 s>>= 56
autoinstrumenter-1      | ; for (u16 i = 0; i < n && s1[i] && s2[i]; i++) {
autoinstrumenter-1      | 5103: (15) if r3 == 0x0 goto pc+249
autoinstrumenter-1      | BPF program is too large. Processed 1000001 insn
autoinstrumenter-1      | processed 1000001 insns (limit 1000000) max_states_per_insn 35 total_states 53678 peak_states 2931 mark_read 183
```

5.15 optimiser is lot worse and it results in larger number of instructions than the new kernels. Let me look for obvious issues like I've done myself in the past.

bpf/common/common.h
```diff
@@ -52,6 +66,7 @@ typedef struct http_request_trace_t {
     pid_info pid;
 } http_request_trace;
 
+// TODO(matt): fix naming
```
rafaelroquetto, line 69: is this todo still actual? does it make sense to either fix it as part of this PR or remove it?

pkg/beyla/config.go
```diff
@@ -71,6 +71,9 @@ var DefaultConfig = Config{
 			Enabled: false,
 			MaxSize: 1000,
 		},
+		BufferSizes: config.EBPFBufferSizes{
+			MySQL: 1024,
```
grcevski: I'm just wondering if we should use a smaller value here or do some performance tests to see what the impact is? I would prefer to have the value set to 0 and be something we tell users to enable if they want detailed errors. I'm attaching here an example of what an ORM did for an application, and I have seen worse examples, where 100s SQL statements were executed for a given HTTP request. 
Screenshot from 2025-04-16 14-03-24

When we consider something like this, a buffer of 1K would cause heavy overhead on OBI side and likely we'll get blamed for slowing down their system, before something is noticed.

I have a proposal how to teach customers for larger buffers and let me know what you think. When we parse the error and it's too short of a buffer, we can put a message in the actual error, please increase your buffer size for more detailed errors. What do you think?

grcevski (approved):
This looks good to me now, I just have a concern about the default 1K buffer for MySQL. Given what I've seen in the past, this can end up being OBI problem.

pkg/beyla/config.go
mmat11: I made the large buffer default size 0 (disabled). The buffer is more useful for the request (queries), as the error fits in the 128B (or 256B) default response buffer. 
For a more precise warning feedback to customers we could do that directly in ebpf, but I would do this in a separate PR if that's ok for you

bpf/common/common.h
mmat11: yes the naming of the structs is inconsistent, can we keep it to remember fixing it?

mmat11:
> As discussed "offline", the only thing that we should do before merging this is to tail call into handle_buf_with_args - I gave it a shot on 5.15 with the optimisations reversed, and the verifier was happy and tests passed.

done

bpf/generictracer/k_tracer_defs.h
```diff
@@ -26,7 +27,14 @@ typedef struct recv_args {
     unsigned char iovec_ctx[sizeof(iovec_iter_ctx)];
 } recv_args_t;
 
-static __always_inline void handle_buf_with_args(void *ctx, call_protocol_args_t *args) {
+// k_tail_handle_buf_with_args
+SEC("kprobe")
+int obi_handle_buf_with_args(void *ctx) {
```
rafaelroquetto: This needs to be in the `.c` file (e.g. k_tracer.c), otherwise we duplicate the program if multiple translation units include the same header (at the moment we have a single translation unit but it's best to be explicit / future-proof - in the past we used to have these programs split into different ones and that could eventually happen again).
I'd also keep the prefix as `beyla_` just for the sake of consistency (no other function uses the obi_ prefix yet), and then once this is merged I'll open a PR renaming everything at once.

rafaelroquetto: (For reference, check out the "One Definition Rule" at https://en.cppreference.com/w/c/language/extern.html)

mmat11: done

bpf/common/common.h
rafaelroquetto: Of course - I thought I'd ask just in case

rafaelroquetto (approved):
Fantastico, grazie davvero per la pazienza e per tutto l’impegno che hai messo in questo.

grcevski:
Thanks for persisting with this @mmat11! It's really great work.

PR #134, opened 2025-06-10
Allow OBI to be reloaded as component

mariomac:
Port from https://github.com/grafana/beyla/pull/2017 and https://github.com/grafana/beyla/pull/2018

We were missing proper handling of the cancellation context in number of places which lead to OBI not properly shutting down sometimes and potentially not being able to reload as a component when we vendor it inside other components.

Additionally, statics inside OBI are problematic when it is reloaded as a component of another executable. Therefore, I moved as many as I could to be instance variables. The ones that still remain as static are the eBPF remapped maps and the shared ring buffer. I manually clean those up when the ctx.Done is received. Maybe we can come up with some object hung of the attacher in the future that will help resolve this.

### DIFF
--- a/pkg/components/appolly/appolly.go
+++ b/pkg/components/appolly/appolly.go
@@ -9,6 +9,8 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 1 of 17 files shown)


grcevski (approved):
LGTM! Thanks for porting this!

PR #138, opened 2025-06-10
Fix unknown gRPC status when indexing fails

grcevski:
If a gRPC request isn't captured by OBI from the very beginning, there might be indexed fields for which we haven't recorded in the OBI dynamic gRPC tables for the given connection. In these cases the information captured is partial. Namely, in this scenario we may fail to decode `grpc-status` and the current OBI code assumes this is successful, e.g. sets the gRPC status 0. 

However there's another gRPC field called `grpc-message` which will be set to non-empty string if a request fails. In this PR we try to do a bit better, when the request fails, but we don't know the reason we say error UNKNOWN (status code 2) instead of incorrectly marking it as unsuccessful. This error will eventually translate to proper OTel error request. So even thought we don't know the protocol exact error we'll at least be able to mark the gRPC request as failed.

Relates to Beyla issue https://github.com/grafana/beyla/issues/1777

### DIFF
--- a/pkg/components/ebpf/common/http2grpc_transform.go
+++ b/pkg/components/ebpf/common/http2grpc_transform.go
@@ -46,6 +46,7 @@ func byteFramer(data []uint8) *http2.Framer {
 // not all requests for a given stream specify the protocol, but one must
 // we remember if we see grpc mentioned and tag the rest of the streams for
 // a given connection as grpc. default assumes plain HTTP2
+// this is why we need the h2c cache
 func getOrInitH2Conn(activeGRPCConnections *lru.Cache[uint64, h2Connection], connID uint64) *h2Connection {
 	v, ok := activeGRPCConnections.Get(connID)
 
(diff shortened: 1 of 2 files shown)


pkg/components/ebpf/common/http2grpc_transform.go
```diff
@@ -346,7 +358,7 @@ func http2FromBuffers(parseContext *EBPFParseContext, event *BPFHTTP2Info) (requ
 			method, path, contentType, ok := readMetaFrame(parseContext, connID, framer, ff)
 
 			if path == "" {
-				path = "."
+				path = "*"
```
grcevski, line 361: @mariomac  I believe this was accidentally changed when reverting the Paths from Glob to Regex. The Path here is the route, unknown, which we mark with asterisk.

pkg/components/ebpf/common/http2grpc_transform_test.go
```diff
@@ -256,7 +291,7 @@ func TestDynamicTableUpdates(t *testing.T) {
 	s, ignore, _ = http2FromBuffers(parseContext, &info)
 	assert.False(t, ignore)
 	assert.Equal(t, "POST", s.Method)
-	assert.Equal(t, ".", s.Path) // this value is the same I just changed the first character from r to p
+	assert.Equal(t, "*", s.Path) // this value is the same I just changed the first character from r to p
```
grcevski, line 294: same fix for accidentally changing our unknown path symbol from * to .

pkg/components/ebpf/common/http2grpc_transform.go
mariomac: Oh!!! Good finding

mariomac (approved):

PR #139, opened 2025-06-10
Sync all eBPF fixes from Beyla to OBI

grcevski:
This PR brings all eBPF changes that were done in the meantime in Beyla, related to various fixes for discovered bugs. We are syncing the codebase so we can start making all eBPF code changes upstream in OBI and then vendor in Beyla.

I'll follow-up with the two test changes for Rails and NGinx


mariomac (approved):
AMAZING!! Thanks for driving this

rafaelroquetto (approved):
Gnarly!!!! Thanks!

PR #149, opened 2025-06-12
Make BPF-generated objects public

mariomac:
To smooth vendoring transition, BPF-generated object types are now public, this would allow using them directly from Beyla instead of having to import also the packages of OBI that currently used the private objects.


grcevski (approved):
LGTM! Excellent

PR #152, opened 2025-06-12
Use injection for NodeJS request correlation

rafaelroquetto:
Relying on async ids and ordering has proved to be a fragile approach, as we don't have access to all of the async ID chains from the C code, and the ordering can be unpredictable.

This PR replaces that approach. Instead, we tap into the nodejs inspector to insert an "agent" that is able to correlate requests at the JavaScript level (thanks to the AsyncLocalStore) - we use it to correlate the file descriptors of the incoming requests with the respective descriptors of the outgoing requests.

In order to pass data back to the ebpf layer, the agent sets up a tiny IPC client/server on an ephemeral port and writes the 20 byte payload that contains a marker, event type, size and the two file descriptors plus a 8-bit CRC - which are then captured in tcp_sendmsg and used to resolve the trace parent information for the requests.


bpf/common/trace_common.h
```diff
@@ -108,14 +108,29 @@ find_nginx_parent_trace(const pid_connection_info_t *p_conn, u16 orig_dport) {
...
+    populate_ephemeral_info(&client_part, &p_conn->conn, orig_dport, p_conn->pid, FD_CLIENT);
+    fd_info_t *fd_info = fd_info_for_conn(&client_part);
+
+    if (!fd_info) {
+        return NULL;
+    }
+
     const u64 pid_tgid = bpf_get_current_pid_tgid();
-    const s32 *node_parent_request_fd = bpf_map_lookup_elem(&active_nodejs_ids, &pid_tgid);
+    const u64 client_key = (pid_tgid << 32) | fd_info->fd;
```
grcevski, line 122: What's the intention for the tgid to be used as a key? I think it might need to be the PID included here, rather than just the thread id? I guess if the tgid is unique then it would work, but I thought it isn't.

pkg/components/nodejs/fdextractor.js
```diff
@@ -0,0 +1,164 @@
...
+	}
+	return orig.serverEmit.call(this, event, ...args);
+};
+
+function correlate(incomingFd, outFd, socket) {
+	if (incomingFd < 0 || outFd < 0 || incomingFd === outFd) {
+		return Promise.resolve();
+	}
+
+	const addr = socket.remoteAddress || 'unknown';
```
grcevski: I think these lines should go under the debug_enabled.

pkg/components/nodejs/fdextractor.js
```diff
@@ -0,0 +1,164 @@
...
+	buf.writeUInt8(0, 6);
+	buf.writeUInt8(0, 7);
+	buf.writeUInt32BE(incomingFd, 8);
+	buf.writeUInt32BE(outFd, 12);
+	buf.writeUInt32BE(0, 16);
+
+	const crc = crc8(buf.slice(0, -1))
+	buf.writeUInt8(crc, 19);
+
+	if (ipcClient && ipcClient.writable) {
```
grcevski: Nit: There are some formatting issues, it might be good to run some sort of js formatter on this.

pkg/components/nodejs/fdextractor.js
```diff
@@ -0,0 +1,164 @@
...
+		});
+	}
+
+	return result;
+};
+
+net.Socket.prototype.write = function(data, ...rest) {
+	const doWrite = () => orig.socketWrite.apply(this, [data, ...rest]);
+
+	// skip ipc writes
```
grcevski: Can you please explain this check a bit more?

pkg/components/nodejs/fdextractor.js
```diff
@@ -0,0 +1,164 @@
...
+	console.log(`[ipc] server listening`);
+});
+
+let ipcClient;
+
+ipcServer.on('listening', () => {
+	const { port, address } = ipcServer.address();
+	ipcClient = net.connect(port, address, () => {
+		ipcClient.setNoDelay(true)
+		if (debug_enabled)
```
grcevski: Curlies here please.

pkg/components/nodejs/fdextractor.js
```diff
@@ -0,0 +1,164 @@
...
+
+ipcServer.on('connection', socket => {
+	socket.on('data', data => {
+		if (!debug_enabled || data.length < 20) return;
+		const marker = data.readUInt32BE(0);
+		const evType = data.readUInt8(4);
+		const len = data.readUInt8(5);
+		const inFd   = data.readUInt32BE(8);
+		const outFd  = data.readUInt32BE(12);
+		console.log(
```
grcevski: Should we put the whole socke.on within debug_enabled? It would make sense to not even hook if we don't have a reason to add a hook here?

pkg/components/nodejs/injector.go
```diff
@@ -0,0 +1,287 @@
...
+			} else {
+				b, _ := json.MarshalIndent(parsed, "", "  ")
+				i.log.Debug("[event]", "payload", string(b))
+			}
+		}
+	}()
+
+	idCount := 0
+
+	send := func(method string, params map[string]any) (map[string]any, error) {
```
grcevski: Can we move this out to a separate function and avoid the nocyc lint?

pkg/components/nodejs/nodejs.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if err := os.Chmod(file.Name(), 0o444); err != nil {
+		cleanup()
+		return nil, errors.New("failed to chmod file")
+	}
+
+	return cleanup, nil
+}
+
+func (i *NodeInjector) Enabled() bool {
```
grcevski, line 59: I think we only need this if traces are enabled? Correct? Metrics don't care about the internal context propagation.

grcevski (commented):
Awesome stuff! I left a few comments.

bpf/common/trace_common.h
rafaelroquetto: There are two kinds of "tids", which often get confused - the "POSIX thread id" (e.g. returned by the likes of `pthread_self()`) and the actual tid in the kernel (returned by `gettid()` or the "tid" bytes of the `bpf_get_current_pid_tgid()` - these are unique within the system (and they are basically pids, or rather,  pids are basically tids of the process main thread). So the intention was to really have a unique key, although I think it doesn't really matter since file descriptors are process-wide.

pkg/components/nodejs/fdextractor.js
rafaelroquetto: Done - nice catch

pkg/components/nodejs/fdextractor.js
rafaelroquetto: Will do

pkg/components/nodejs/fdextractor.js
rafaelroquetto: Done

pkg/components/nodejs/fdextractor.js
rafaelroquetto: Adding the hook puts the stream in "flow mode", which causes it to start pulling data from the buffer. I don't know what would happen if we let the buffer overflow by removing this.

The alternative would be to call `stream.resume()` to set it into flow mode anyway, and then when there's no handler, it will just drop the data. (https://nodejs.org/api/stream.html#two-reading-modes) - looking at the code though in this case it appears that they just set the handler to be a "noop" function - so pretty much what this is when debug_enabled is false.

grcevski: ok, cool. We should keep it then, but put a comment and nicely wrap the whole inner part with if (!debug_enabled), so that the intent is clear, we just need the hook to drain the stream.

pkg/components/nodejs/fdextractor.js
rafaelroquetto: this is just skipping trying to correlate writes coming from the ipc or stdin/stdout "sockets", as we are not interested in those - instead of letting `correlate()` be called a few lines down, we just call the original socket write (e.g. `write()`) from the get go.

pkg/components/nodejs/fdextractor.js
rafaelroquetto: Done with `npx prettier --write fdextractor.js`

pkg/components/nodejs/injector.go
rafaelroquetto: Done, looks better indeed!

pkg/components/nodejs/nodejs.go
rafaelroquetto: Nice catch, fixed!

grcevski (approved):
LGTM! Thanks for iterating on this!

PR #161, opened 2025-06-14
feat: support JSON-RPC 2.0 over http in golang's uprobe

titaneric:
I try to capture http request body by introducing new hook for `net/http.(*body).Read` function.
The body struct is initialized in the [readTransfer  in readRequest](https://github.com/golang/go/blob/8dd7d2111b8622dac4b0127fa1d26da3c1c4c274/src/net/http/request.go#L1158).

[readTransfer](https://github.com/golang/go/blob/8dd7d2111b8622dac4b0127fa1d26da3c1c4c274/src/net/http/transfer.go#L561-L583) prepare the body reader, and body's `io.Reader` implementation is [here](https://github.com/golang/go/blob/8dd7d2111b8622dac4b0127fa1d26da3c1c4c274/src/net/http/transfer.go#L811-L838).

Note that I capture partial http request body and retrieve the `Content-Type` in bpf, and send them into userspace.

<details>

<summary>Test JSON-RPC code:</summary>

```go
package main

import (
    "io"
    "log"
    "net"
    "net/http"
    "net/rpc"
    "net/rpc/jsonrpc"
)

// Args defines the arguments for the RPC methods.
type Args struct {
    A, B int
}

// Arith provides methods for arithmetic operations.
type Arith struct{}

// Multiply multiplies two numbers and returns the result.
func (t *Arith) Multiply(args *Args, reply *int) error {
    *reply = args.A * args.B
    return nil
}

// ReadWriteCloserWrapper wraps an io.Reader and io.Writer to implement io.ReadWriteCloser.
type ReadWriteCloserWrapper struct {
    io.Reader
    io.Writer
}
// Close is a no-op to satisfy the io.ReadWriteCloser interface.
func (w *ReadWriteCloserWrapper) Close() error {
    return nil
}
func main() {
    // Register the Arith service.
    arith := new(Arith)
    rpc.Register(arith)

    // Set up an HTTP handler for JSON-RPC.
    http.HandleFunc("/jsonrpc", func(w http.ResponseWriter, r *http.Request) {
        if r.Method != http.MethodPost {
            http.Error(w, "Only POST method is allowed", http.StatusMethodNotAllowed)
            return
        }
	// Wrap the request body and response writer in a ReadWriteCloser.
        conn := &ReadWriteCloserWrapper{Reader: r.Body, Writer: w}
        // Serve the request using JSON-RPC codec.
        rpc.ServeCodec(jsonrpc.NewServerCodec(conn))
    })

    // Start the HTTP server.
    go func() {
        log.Println("JSON-RPC server is listening on HTTP port 8080...")
        if err := http.ListenAndServe(":8080", nil); err != nil {
            log.Fatalf("Error starting HTTP server: %v", err)
        }
    }()

    // Optionally, start a TCP server as well.
    listener, err := net.Listen("tcp", ":1234")
    if err != nil {
        log.Fatalf("Error starting TCP server: %v", err)
    }
    log.Println("JSON-RPC server is listening on TCP port 1234...")

    for {
        conn, err := listener.Accept()
        if err != nil {
            log.Printf("Connection error: %v", err)
            continue
        }
        // Serve the connection using JSON-RPC.
        go rpc.ServeCodec(jsonrpc.NewServerCodec(conn))
    }
}

```

</details>

Sample request

```bash
curl -X POST http://localhost:8080/jsonrpc \
           -H "Content-Type: application/json" \
           -d '{
             "jsonrpc": "2.0",
             "method": "Arith.Multiply",
             "params": [{"A": 7, "B": 8}],
             "id": 1
           }'
```

Output:

```bash
make dev
sudo -E ./bin/ebpf-instrument --config ~/beyla-config.yaml
```

```
time=2025-06-15T12:19:40.682Z level=DEBUG msg="found relevant function for instrumentation" component=goexec.instructions function=net/http.(*conn).serve offsets="{Start:2521664 Returns:[2522012 2522356 2522372 2522388 2523316 2523332 2523412 2523428 2523452 2523860 2524116 2524228 2524444 2524564 2525716 2525856]}"
time=2025-06-15T12:19:40.682Z level=DEBUG msg="found relevant function for instrumentation" component=goexec.instructions function=net/http.(*body).Read offsets="{Start:2567744 Returns:[2568020 2568076 2568104]}"
time=2025-06-15T12:19:40.682Z level=DEBUG msg="found relevant function for instrumentation" component=goexec.instructions function=net/http.serverHandler.ServeHTTP offsets="{Start:2581872 Returns:[2582068]}"
time=2025-06-15T12:19:40.682Z level=DEBUG msg="Can't read all offsets from DWARF info. Checking in prefetched database" component=goexec.structMemberOffsets
time=2025-06-15T12:19:40.683Z level=DEBUG msg="Go version detected" component=goexec.structMemberOffsets version=1.24.2
...
2025-06-15 12:19:53.615121953 (504.832µs[494.916µs]) HTTP 200 Arith.M /jsonrpc [::1 as ::1:38966]->[::1 as ::1:8080] contentLen:140B responseLen:0B svc=[server go] traceparent=[00-dea07b178064c9d2d08b9856b20bdb8c-b2bf279cc2145e9f[0000000000000000]-01]
time=2025-06-15T12:19:53.925Z level=DEBUG msg="storing new metric label set" component=prom.Expirer labelValues="[Arith.M 200 /* titan:54337 server ::1 8080 server ]"
time=2025-06-15T12:19:53.925Z level=DEBUG msg="storing new metric label set" component=prom.Expirer labelValues="[Arith.M 200 /* titan:54337 server ::1 8080 server ]"
time=2025-06-15T12:19:53.925Z level=DEBUG msg="storing new metric label set" component=prom.Expirer labelValues="[Arith.M 200 /* titan:54337 server ::1 8080 server ]"
```

```yaml
open_port: 1234,8080
trace_printer: text
log_level: DEBUG
ebpf:
  bpf_debug: true
  enable_context_propagation: true
  track_request_headers: true
  traffic_control_l7_context_propagation: true
  protocol_debug_print: true
routes:
    unmatched: heuristic
prometheus_export:
  port: 9090
  path: /metrics
otel_traces_export:
  endpoint: http://localhost:4318


```

I hope that present direction is correct for the implementation. I would like to hear more discussion and feedback, and I would continue the work for response body capturing, error code handling, etc.

P.S. The patch is came from https://github.com/grafana/beyla/pull/2008


titaneric:
Ohh, I would check it in the old kernel version. I develop this in relatively new version.
```
❯ uname -a
Linux titan 6.11.0-26-generic #26~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Apr 17 19:41:24 UTC 2 aarch64 aarch64 aarch64 GNU/Linux
```

titaneric:
I fix it in kernel 5.15, and it also works on newer kernel such as 6.11 as well!

```
time=2025-06-17T15:57:48.258Z level=DEBUG msg="submitting traces on timeout" component=ringbuf.Tracer len=1
time=2025-06-17T15:57:48.260Z level=DEBUG msg="storing new metric label set" component=prom.Expirer labelValues=[go]
2025-06-17 15:57:47.61735747 (1.114756ms[1.102506ms]) HTTP 200 Arith.M /jsonrpc [192.168.106.5 as 192.168.106.5:60376]->[192.168.106.6 as 192.168.106.6:8080] contentLen:140B responseLen:0B svc=[server go] traceparent=[00-dc8cf5f20469a22fa08442e737a60e05-2e0d40b8ea61b0c6[0000000000000000]-01]
```

bpf/common/tc_common.h
```diff
@@ -131,3 +131,31 @@ static __always_inline u32 sk_msg_local_port(struct sk_msg_md *ctx) {
 
     return data;
 }
+
+// find the needle in the haystack, return the position of the first occurrence, return -1 if not found
+static __always_inline u32 bpf_memstr(const char *haystack,
+                                      int haystack_len,
+                                      const char *needle,
+                                      int needle_len) {
```
rafaelroquetto:
```suggestion
static __always_inline u32 bpf_memstr(const unsigned char *haystack,
                                      u32 haystack_len,
                                      const unsigned char *needle,
                                      u32 needle_len) {
```

rationale: prefer unsigned (size_t or u32 in this case) for lengths

bpf/common/tc_common.h
```diff
@@ -131,3 +131,31 @@ static __always_inline u32 sk_msg_local_port(struct sk_msg_md *ctx) {
...
+    if (needle_len == 0 || haystack_len < needle_len) {
+        return INVALID_POS;
+    }
+    for (int i = 0; i <= haystack_len - needle_len; i++) {
```
rafaelroquetto:
```suggestion
    for (u32 i = 0; i <= haystack_len - needle_len; i++) {
```

bpf/common/tc_common.h
```diff
@@ -131,3 +131,31 @@ static __always_inline u32 sk_msg_local_port(struct sk_msg_md *ctx) {
...
+        int found = 1;
```
rafaelroquetto:
```suggestion
        u8 found = 1;
```

bpf/common/tc_common.h
```diff
@@ -131,3 +131,31 @@ static __always_inline u32 sk_msg_local_port(struct sk_msg_md *ctx) {
...
+#pragma unroll
+        // max needle length
+        for (int j = 0; j < MAX_NEEDLE_LEN; j++) {
```
rafaelroquetto:
```suggestion
        for (u8 j = 0; j < MAX_NEEDLE_LEN; j++) {
```

bpf/common/tc_common.h
```diff
@@ -131,3 +131,31 @@ static __always_inline u32 sk_msg_local_port(struct sk_msg_md *ctx) {
...
+            if (j >= needle_len) {
+                break;
+            }
+            if (haystack[i + j] != needle[j]) {
+                found = 0;
+                break;
+            }
+        }
+        if (found) {
+            return (u32)i;
```
rafaelroquetto:
```suggestion
            return i;
```

bpf/gotracer/go_nethttp.c
```diff
@@ -262,6 +273,19 @@ int beyla_uprobe_http2Server_processHeaders(struct pt_regs *ctx) {
     return 0;
 }
 
+static __always_inline void update_traceparent(server_http_func_invocation_t *inv,
+                                               u8 *header_start) {
```
rafaelroquetto:
```suggestion
                                               const unsigned char *header_start) {
```

you will need to fix `decode_go_traceparent` to take a `const` buf as well

bpf/gotracer/go_nethttp.c
```diff
@@ -262,6 +273,19 @@ int beyla_uprobe_http2Server_processHeaders(struct pt_regs *ctx) {
...
+    decode_go_traceparent(header_start, inv->tp.trace_id, inv->tp.parent_id, &inv->tp.flags);
+    bpf_dbg_printk("Found traceparent in header %s", header_start);
+}
+
+static __always_inline void update_content_type(server_http_func_invocation_t *inv,
+                                                u8 *header_start) {
```
rafaelroquetto:
```suggestion
                                                const unsgined char *header_start) {
```

bpf/gotracer/go_nethttp.c
```diff
@@ -270,22 +294,51 @@ int beyla_uprobe_readContinuedLineSliceReturns(struct pt_regs *ctx) {
...
     u8 *buf = (u8 *)GO_PARAM1(ctx);
 
-    if (len >= (W3C_KEY_LENGTH + W3C_VAL_LENGTH + 2)) {
-        u8 temp[W3C_KEY_LENGTH + W3C_VAL_LENGTH + 2];
-        bpf_probe_read(temp, sizeof(temp), buf);
-        bpf_dbg_printk("goroutine_addr %lx", goroutine_addr);
-        go_addr_key_t g_key = {};
-        go_addr_key_from_id(&g_key, goroutine_addr);
+    // avoid variable-length buffer reads to work on older kernels such as 5.15 or earlier
+    u8 temp[HTTP_HEADER_MAX_LEN] = {};
```
rafaelroquetto: so, instead of u8, we tend to use `unsigned char` (and sometimes char). Using u8 to represent arrays of bytes in C (and C++) is frowned upon because `uint8_t` is not guaranteed to be a character type (even though in practice it often is). Casts from a pointer to any object type to a character type are allowed well defined to be the lowest byte of the object - otherwise it's undefined behaviour (and may break strict aliasing rules).

For reference: https://reviews.llvm.org/D114718 and [ISO C11 §6.5/7](https://port70.net/~nsz/c/c11/n1570.html#6.5p7)


Also, it's best to use a map for this rather than put this on the stack - this is a big array for ebpf standard and will contribute to a stack overflow in the future.

Instead, you can use a map:
```
struct {
    __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
    __type(key, u32);
    __type(value, unsigned char[HTTP_HEADER_MAX_LEN]);
    __uint(max_entries, 1);
} temp_mem_scratch SEC(".maps");

static __always_inline unsigned char *temp_mem() {
    const u32 zero = 0;
    return bpf_map_lookup_elem(&temp_mem_scratch, &zero);
}
```

bpf/gotracer/go_nethttp.c
```diff
@@ -270,22 +294,51 @@ int beyla_uprobe_readContinuedLineSliceReturns(struct pt_regs *ctx) {
...
+    u64 safe_len = len > HTTP_HEADER_MAX_LEN ? HTTP_HEADER_MAX_LEN : len;
```
rafaelroquetto:
```suggestion
    const u32 safe_len = len > HTTP_HEADER_MAX_LEN ? HTTP_HEADER_MAX_LEN : len;
```

bpf/gotracer/go_nethttp.c
```diff
@@ -67,6 +68,8 @@ typedef struct server_http_func_invocation {
     u8 method[METHOD_MAX_LEN];
     u8 path[PATH_MAX_LEN];
     u8 _pad[5];
+    u64 body_addr; // pointer to the body buffer
+    u8 content_type[HTTP_CONTENT_TYPE_MAX_LEN];
```
rafaelroquetto:
```suggestion
    unsigned char content_type[HTTP_CONTENT_TYPE_MAX_LEN];
```

I see you are just sticking to what is already in the code (and rightfully so) - but if you feel like updating method and path as well, that would be a bonus (but don't feel compelled to).

A lot of the following suggestions I have down below relate to this change - but take this as an optional one, someone can address all of this inconsistencies in a separate PR.

bpf/gotracer/go_nethttp.c
```diff
@@ -270,22 +294,51 @@ int beyla_uprobe_readContinuedLineSliceReturns(struct pt_regs *ctx) {
...
-        if (existing) {
-            if (!bpf_memicmp((const char *)temp, "traceparent: ", W3C_KEY_LENGTH + 2)) {
-                server_http_func_invocation_t inv = {};
-                decode_go_traceparent(
-                    temp + W3C_KEY_LENGTH + 2, inv.tp.trace_id, inv.tp.parent_id, &inv.tp.flags);
-                bpf_dbg_printk("Found traceparent in header %s", temp);
-                bpf_map_update_elem(&ongoing_http_server_requests, &g_key, &inv, BPF_ANY);
-            }
+#pragma unroll
+    for (int i = 0; i < HTTP_HEADER_MAX_LEN; i++) {
```
rafaelroquetto:
```suggestion
    for (u8 i = 0; i < HTTP_HEADER_MAX_LEN; i++) {
```

bpf/gotracer/go_nethttp.c
```diff
@@ -270,22 +294,51 @@ int beyla_uprobe_readContinuedLineSliceReturns(struct pt_regs *ctx) {
...
+        if (i < safe_len) {
+            bpf_probe_read(&temp[i], 1, buf + i);
```
rafaelroquetto: We should try to read everything at once here. Reading byte by byte here is a no go as `bpf_probe_read_user()` does a lot of book-keeping under the hood (such as handling page-faults, permissions and expensive atomic operations) every time it's invoked.

Instead, you can zero-initialise temp and read it all at once:

```
unsigned char temp[HTTP_HEADER_MAX_LEN] = {};
if (bpf_probe_read_user(&temp, sizeof(temp), buf) != 0) { /* error handling */ }; 
```

bpf/gotracer/go_nethttp.c
```diff
@@ -270,22 +294,51 @@ int beyla_uprobe_readContinuedLineSliceReturns(struct pt_regs *ctx) {
...
+    }
+
+    bpf_dbg_printk("goroutine_addr %lx", goroutine_addr);
+    go_addr_key_t g_key = {};
+    go_addr_key_from_id(&g_key, goroutine_addr);
+
+    int w3c_value_start = W3C_KEY_LENGTH + 2; // "traceparent: "
+    int w3c_header_length = w3c_value_start + W3C_VAL_LENGTH;
+    int content_type_value_start = CONTENT_TYPE_KEY_LEN + 2; // "content-type: "
+    int content_type_header_length = content_type_value_start + HTTP_CONTENT_TYPE_MAX_LEN;
```
rafaelroquetto: use `const` when possible and unsigned (e.g. u8 or u32) when feasible

bpf/gotracer/go_nethttp.c
```diff
@@ -270,22 +294,51 @@ int beyla_uprobe_readContinuedLineSliceReturns(struct pt_regs *ctx) {
...
+
+    connection_info_t *existing = bpf_map_lookup_elem(&ongoing_server_connections, &g_key);
+    if (!existing) {
+        return 0;
+    }
```
rafaelroquetto, line 365: I'd do this before reading `temp` - it gives us the opportunity to bail early and not wasting time copying data from user space

bpf/gotracer/go_nethttp.c
```diff
@@ -270,22 +294,51 @@ int beyla_uprobe_readContinuedLineSliceReturns(struct pt_regs *ctx) {
...
+    server_http_func_invocation_t *inv = bpf_map_lookup_elem(&ongoing_http_server_requests, &g_key);
+
+    if (safe_len >= w3c_header_length &&
+        !bpf_memicmp((const char *)temp, "traceparent: ", w3c_value_start)) {
+        u8 *traceparent_start = temp + w3c_value_start;
```
rafaelroquetto:
```suggestion
        bpf_memicmp(temp, "traceparent: ", w3c_value_start) == 0)  {
        unsigned char *traceparent_start = temp + w3c_value_start;
```
having said this, instead of doing

```
int w3c_value_start = W3C_KEY_LENGTH + 2; // "traceparent: "
```
and then spelling the string literal `"traceparent: "` here , which is very error-prone, it's best to define a constant, e.g.

```
const char traceparent[] = "traceparent: ";
/* ... */
if (... && bpf_memicmp(temp, (const unsigned char*) traceparent, sizeof(traceparent) - 1) == 0)
```


and can we break this if into functions so that it reads something like:

```
if (is_traceparent(temp, safe_len)) {
   update_inv_traceparent(temp, safe_len, inv);
} else if (is_content_len(...) {
 /* ... */
}
```

bpf/gotracer/go_nethttp.c
```diff
@@ -270,22 +294,51 @@ int beyla_uprobe_readContinuedLineSliceReturns(struct pt_regs *ctx) {
...
+        if (inv) {
+            update_traceparent(inv, traceparent_start);
+        } else {
+            server_http_func_invocation_t minimal_inv = {};
+            update_traceparent(&minimal_inv, traceparent_start);
+            bpf_map_update_elem(&ongoing_http_server_requests, &g_key, &minimal_inv, BPF_ANY);
+        }
+    } else if (safe_len >= content_type_header_length &&
+               !bpf_memicmp((const char *)temp, "content-type: ", content_type_value_start)) {
```
rafaelroquetto: ditto

bpf/gotracer/go_nethttp.c
```diff
@@ -1166,6 +1219,72 @@ int beyla_uprobe_netFdRead(struct pt_regs *ctx) {
...
+        bpf_map_lookup_elem(&ongoing_http_server_requests, &g_key);
+    if (!invocation) {
+        bpf_dbg_printk("can't find invocation info for server call");
+        return 0;
+    }
+    // content-type is set in invocation in ServeHTTP
+    bpf_dbg_printk("n is %d", n);
+    bpf_dbg_printk("content type is %s", invocation->content_type);
+
+    char body_buf[HTTP_BODY_MAX_LEN] = {};
```
rafaelroquetto: we should also us a map instead of putting this on the stack, as suggested above

bpf/gotracer/go_nethttp.c
```diff
@@ -1166,6 +1219,72 @@ int beyla_uprobe_netFdRead(struct pt_regs *ctx) {
...
+    u64 n = (u64)GO_PARAM1(ctx);
+
+    server_http_func_invocation_t *invocation =
+        bpf_map_lookup_elem(&ongoing_http_server_requests, &g_key);
+    if (!invocation) {
+        bpf_dbg_printk("can't find invocation info for server call");
+        return 0;
+    }
+    // content-type is set in invocation in ServeHTTP
+    bpf_dbg_printk("n is %d", n);
```
rafaelroquetto:
```suggestion
    bpf_dbg_printk("n is %llu", n);
```

bpf/gotracer/go_nethttp.c
```diff
@@ -1166,6 +1219,72 @@ int beyla_uprobe_netFdRead(struct pt_regs *ctx) {
...
+    bpf_dbg_printk("content type is %s", invocation->content_type);
+
+    char body_buf[HTTP_BODY_MAX_LEN] = {};
+    if (n > 0 && invocation->body_addr) {
```
rafaelroquetto:
```suggestion
    if (n <= 0 || !invocation->body_addr) {
      return 0;
    }
```

rationale: 
 - [F.56: Avoid unnecessary condition nesting](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f56-avoid-unnecessary-condition-nesting)
 - https://www.kernel.org/doc/html/next/process/coding-style.html

rafaelroquetto (commented):
A few comments - thank you again for sticking with this - it's great work! I hope you don't mind my multiple passes, it's a big PR and it's difficult to catch everything in one go.

rafaelroquetto:
@titaneric do you think it would be possible to add an integration test as well? Similar to the example you mentioned in the review description. Happy to help with that as well if you prefer to share the burden.

titaneric:
Yes! I would like to add the integration test as well.

bpf/gotracer/go_nethttp.c
titaneric: Very instresting! Thank you for pointing out the possible stack overflow issue. I would refactor this to bpf map instead.

titaneric:
I have hit verfifier error in kernel 5.15 in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/161/commits/2976d5e0e3edb23d23b28867def7a70aa8181bcc . Should I resolve it first or I change everything you suggest in newer kernel and then go back to resolve it?

titaneric:
Thank you! The verifier error on kernel 5.15 is attached as follows.
[beyla-ebpf-debug.log](https://github.com/user-attachments/files/20800810/beyla-ebpf-debug.log)

rafaelroquetto:
Thanks @titaneric - yeah it seems we will have to tailcall. The fact that it didn't happen earlier is sheer luck - the compiler is generating a number of instructions that sits close to the allowed limit - so any changes in the code end up triggering it.

If you need a reference implementation, have a look at:
- https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/3425fe3c036bd7698b535de1fb25bcf945e984e6/bpf/tpinjector/tpinjector.c#L485
- https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/3425fe3c036bd7698b535de1fb25bcf945e984e6/pkg/components/ebpf/tpinjector/tpinjector.go#L59

If you need help with this, let me know and I can try and have a go at it tomorrow.

titaneric:
It does not work by simply adopting the tail code https://github.com/titaneric/opentelemetry-ebpf-instrumentation/commit/8b5e12f1ce5406d3a7922fc69cde7523b41a30a0 , I might need some time to investigate it in the tomorrow.

bpf/gotracer/go_nethttp.c
```diff
@@ -1166,6 +1219,72 @@ int beyla_uprobe_netFdRead(struct pt_regs *ctx) {
...
+        if (is_json_content_type((void *)invocation->content_type,
+                                 sizeof(invocation->content_type))) {
+            if (read_go_str_n(
+                    "http body", (void *)invocation->body_addr, n, body_buf, sizeof(body_buf))) {
+                bpf_dbg_printk("body is %s", body_buf);
+                if (is_jsonrpc2_body((const unsigned char *)body_buf, sizeof(body_buf))) {
```
rafaelroquetto: you can try to tail here instead - i.e. 

```
if (!is_json_content_type(...)) {
   return ...
 }

if (!read_go_str_n()) {
   return ...
}}

// all good, tail

bpf_tailcall_....
````

this will split the program complexity into two different bpf programs

File: bpf/gotracer/go_nethttp.c (hunk 15 above)
Comment (titaneric) [reply bpf/gotracer/go_nethttp.c]: I hope https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/161/commits/3848820d1d16a9762698120938255b1995582998 could meet your need!

Comment (rafaelroquetto) [reply bpf/gotracer/go_nethttp.c]: @titaneric absolutely! thank you so much for this, looks great!

File: bpf/gotracer/go_nethttp.c (hunk 17 above)
Comment (titaneric) [reply bpf/gotracer/go_nethttp.c]: It seems that it's not as trivial as header map does. I hit the bound check in https://github.com/titaneric/opentelemetry-ebpf-instrumentation/commit/9d0b223017f085c4ac9bb2b409362aaf61329a72.

[body-mem-bpf-map.log](https://github.com/user-attachments/files/20841418/body-mem-bpf-map.log)

Comment (rafaelroquetto) [reply bpf/gotracer/go_nethttp.c]: Based on the logs, it seems the inner loop is suffering a off-by-one error (it shouldn't be **off=64**)



```
1176: (bf) r5 = r3                    ; R3_w=map_value(map=temp_body_mem_s,ks=4,vs=64,off=62) R5_w=map_value(map=temp_body_mem_s,ks=4,vs=64,off=62)
1177: (0f) r5 += r4                   ; R4_w=2 R5_w=map_value(map=temp_body_mem_s,ks=4,vs=64,off=64)
1178: (71) r4 = *(u8 *)(r5 +0)
invalid access to map value, value_size=64 off=64 size=1
```
when unrolling the loop, the verifier is not able to track the `if (j>=needle_len)` and breadk ... I guess... so we need to explicitly verify that `[i + j]` is within bounds - can you see if the following works?
```
#pragma unroll
for (u8 j = 0; j < MAX_NEEDLE_LEN; j++) {
    if (j >= needle_len)
        break;
    // ensure i + j < haystack_len
    if (i + j >= haystack_len) {
        found = 0;
        break;
    }
    if (haystack[i + j] != needle[j]) {
        found = 0;
        break;
    }
```

**Comment by titaneric** [comment]
Thank you for your information, I would try it in the tomorrow!
Also, I have tried the tail call and allocate body memory from bpf map in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/161/commits/b3dc7d4e5054717b0bfe49d027aa0ad2f80bfebe. It works in kernel 6.11 but not in 5.15 (max instruction limit, again).
[bpf-tail-code.log](https://github.com/user-attachments/files/20842384/bpf-tail-code.log)

File: bpf/gotracer/go_nethttp.c (hunk 17 above)
Comment (titaneric) [reply bpf/gotracer/go_nethttp.c]: resolved in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/161/commits/2e75fda72f42e8545a4396ff06883ccaf7427648

File: bpf/gotracer/go_nethttp.c (hunk 20 above)
Comment (titaneric) [reply bpf/gotracer/go_nethttp.c]: resolved in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/161/commits/b3dc7d4e5054717b0bfe49d027aa0ad2f80bfebe

File: bpf/gotracer/go_nethttp.c (hunk 10 above)
Comment (titaneric) [reply bpf/gotracer/go_nethttp.c]: resolved in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/161/commits/06fc0d87194ef60ced5aa2c8dbe91dd7409b303a

File: bpf/gotracer/go_nethttp.c [hunk 21]
```diff
@@ -117,6 +147,18 @@ int beyla_uprobe_ServeHTTP(struct pt_regs *ctx) {
         // TODO: if context propagation is supported, overwrite the header value in the map with the
         // new span context and the same thread id.
 
+        // get content-type from readContinuedLineSlice
+        if (header_inv && header_inv->content_type[0]) {
+            bpf_dbg_printk("Found content type in ongoing request: %s", header_inv->content_type);
+            // __builtin_memcpy(invocation.content_type,
```
titaneric: Should we copy the previously set content type into new invocation?

titaneric:
I run the [mcp-go server example](https://github.com/mark3labs/mcp-go/blob/8f5b048218f6d044c3322f16b8cb0b08e10bf5d0/examples/custom_context/main.go), and it works!

```bash
go run server.go -t http
```

with  [mcp-go client example](https://github.com/mark3labs/mcp-go/blob/8f5b048218f6d044c3322f16b8cb0b08e10bf5d0/examples/simple_client/main.go)

```bash
go run client.go --http http://localhost:8080/mcp
```

Result log: 
[mcp-server.log](https://github.com/user-attachments/files/20852139/mcp-server.log)

Partial result:
```
2025-06-22 10:06:38.62210638 (4.441337ms[4.024504ms]) HTTP 200 initial /mcp [127.0.0.1 as 127.0.0.1:36896]->[127.0.0.1 as 127.0.0.1:8080] contentLen:177B responseLen:0B svc=[server go] traceparent=[00-9c70703095a7979a788673a8cdf70c73-c6dfbb050b503ca4[0000000000000000]-01]
2025-06-22 10:06:38.62210638 (644.613µs[637.693µs]) HTTP 202 notific /mcp [127.0.0.1 as 127.0.0.1:36896]->[127.0.0.1 as 127.0.0.1:8080] contentLen:66B responseLen:0B svc=[server go] traceparent=[00-1d02d1b36f1bfb82b72ddeb969dd9953-2ea1990ceb7144c7[0000000000000000]-01]
2025-06-22 10:06:38.62210638 (1.711144ms[1.69105ms]) HTTP 200 tools/l /mcp [127.0.0.1 as 127.0.0.1:36896]->[127.0.0.1 as 127.0.0.1:8080] contentLen:58B responseLen:0B svc=[server go] traceparent=[00-32505e9e074d9e312ed0026e638703fa-138d73580aa4cb49[0000000000000000]-01]
2025-06-22 10:06:38.62210638 (431.006µs[427.963µs]) HTTP 200 resourc /mcp [127.0.0.1 as 127.0.0.1:36896]->[127.0.0.1 as 127.0.0.1:8080] contentLen:62B responseLen:0B svc=[server go] traceparent=[00-0f0c9828e70b90469d5e0517c6734e66-a2fc7d723c9249b1[0000000000000000]-01]
time=2025-06-22T10:06:39.568Z level=DEBUG msg="storing new metric label set" component=prom.Expirer labelValues=[go]
time=2025-06-22T10:06:39.570Z level=DEBUG msg="storing new metric label set" component=prom.Expirer labelValues="[initial 200 /* titan:378356 server 127.0.0.1 8080 server ]"
time=2025-06-22T10:06:39.573Z level=DEBUG msg="storing new metric label set" component=prom.Expirer labelValues="[initial 200 /* titan:378356 server 127.0.0.1 8080 server ]"
time=2025-06-22T10:06:39.573Z level=DEBUG msg="storing new metric label set" component=prom.Expirer labelValues="[initial 200 /* titan:378356 server 127.0.0.1 8080 server ]"
time=2025-06-22T10:06:39.573Z level=DEBUG msg="storing new metric label set" component=prom.Expirer labelValues="[notific 202 /* titan:378356 server 127.0.0.1 8080 server ]"
time=2025-06-22T10:06:39.573Z level=DEBUG msg="storing new metric label set" component=prom.Expirer labelValues="[notific 202 /* titan:378356 server 127.0.0.1 8080 server ]"
time=2025-06-22T10:06:39.573Z level=DEBUG msg="storing new metric label set" component=prom.Expirer labelValues="[notific 202 /* titan:378356 server 127.0.0.1 8080 server ]"
time=2025-06-22T10:06:39.573Z level=DEBUG msg="storing new metric label set" component=prom.Expirer labelValues="[tools/l 200 /* titan:378356 server 127.0.0.1 8080 server ]"
time=2025-06-22T10:06:39.573Z level=DEBUG msg="storing new metric label set" component=prom.Expirer labelValues="[tools/l 200 /* titan:378356 server 127.0.0.1 8080 server ]"
time=2025-06-22T10:06:39.573Z level=DEBUG msg="storing new metric label set" component=prom.Expirer labelValues="[tools/l 200 /* titan:378356 server 127.0.0.1 8080 server ]"
time=2025-06-22T10:06:39.573Z level=DEBUG msg="storing new metric label set" component=prom.Expirer labelValues="[resourc 200 /* titan:378356 server 127.0.0.1 8080 server ]"
time=2025-06-22T10:06:39.573Z level=DEBUG msg="storing new metric label set" component=prom.Expirer labelValues="[resourc 200 /* titan:378356 server 127.0.0.1 8080 server ]"
time=2025-06-22T10:06:39.573Z level=DEBUG msg="storing new metric label set" component=prom.Expirer labelValues="[resourc 200 /* titan:378356 server 127.0.0.1 8080 server ]"
```

bpf/gotracer/go_nethttp.c
```diff
@@ -63,10 +67,13 @@ typedef struct server_http_func_invocation {
...
+    u64 body_addr; // pointer to the body buffer
     tp_info_t tp;
-    unsigned char method[METHOD_MAX_LEN];
-    unsigned char path[PATH_MAX_LEN];
-    u8 _pad[5];
+    unsigned char content_type[HTTP_CONTENT_TYPE_MAX_LEN];
+    u8 method[METHOD_MAX_LEN];
+    u8 path[PATH_MAX_LEN];
+    u8 json_content_type;
+    u8 _pad[12];
```
rafaelroquetto:
```suggestion
    u8 _pad[4];
```

bpf/gotracer/go_nethttp.c
```diff
@@ -76,6 +83,29 @@ struct {
...
+    __type(key, u32);
+    __type(value, unsigned char[HTTP_HEADER_MAX_LEN]);
+    __uint(max_entries, 1);
+} temp_header_mem_store SEC(".maps");
+
+static __always_inline unsigned char *temp_header_mem() {
+    const u32 zero = 0;
+    return bpf_map_lookup_elem(&temp_header_mem_store, &zero);
+}
+struct {
```
rafaelroquetto, line 101: nit: missing empty line above

bpf/gotracer/go_nethttp.c
rafaelroquetto: Only if strictly necessary - is it being used elsewhere?

bpf/gotracer/go_nethttp.c
```diff
@@ -262,31 +304,94 @@ int beyla_uprobe_http2Server_processHeaders(struct pt_regs *ctx) {
...
-            if (!bpf_memicmp((const char *)temp, "traceparent: ", W3C_KEY_LENGTH + 2)) {
-                server_http_func_invocation_t inv = {};
-                decode_go_traceparent(
-                    temp + W3C_KEY_LENGTH + 2, inv.tp.trace_id, inv.tp.parent_id, &inv.tp.flags);
-                bpf_dbg_printk("Found traceparent in header %s", temp);
-                bpf_map_update_elem(&ongoing_http_server_requests, &g_key, &inv, BPF_ANY);
-            }
-        }
+    const char traceparent[] = "traceparent: ";
+    const char content_type[] = "content-type: ";
```
rafaelroquetto: if possible, you should move these to `static const char ....[] = " ... ";` at file scope level, just so that they don't occupy the already small stack space.

rafaelroquetto: I forgot to mention, if not necessarily, we should remove the commented out code (or uncommented them if they turn out to be necessary after all).

bpf/gotracer/go_nethttp.c
```diff
@@ -1181,6 +1283,102 @@ int beyla_uprobe_netFdRead(struct pt_regs *ctx) {
...
+    if (!body_buf) {
+        return 0;
+    }
+
+    const u32 safe_len = n > HTTP_BODY_MAX_LEN ? HTTP_BODY_MAX_LEN : n;
+    if (!read_go_str_n("http body", (void *)invocation->body_addr, n, body_buf, safe_len)) {
+        bpf_dbg_printk("failed to read body, n=%llu, body_addr=%llx", n, invocation->body_addr);
+        return 0;
+    }
+    // bpf_dbg_printk("body is %s", body_buf);
```
titaneric, line 1340: I leave the commented code here fo future debug purpose

titaneric:
I have one question about the http method length. Current maximum length is 7 (`OPTIONS` length), but for the jsonrpc method, it's not enough and it's usually get cut off. Should we increase the length for it? I understand we need to re-organize the field position in certain struct, and I would update it if we agree for that.

rafaelroquetto:
> I have one question about the http method length. Current maximum length is 7 (OPTIONS length), but for the jsonrpc method, it's not enough and it's usually get cut off. Should we increase the length for it? I understand we need to re-organize the field position in certain struct, and I would update it if we agree for that.

@titaneric yes, we can do it - how many bites do we need?

rafaelroquetto:
I am not sure why the tests are failing, do they succeed locally?

Btw, it may be worth it also adding your tests to `docker-compose-multiexec-host.yml `

titaneric:
Let me dig into the test issue

titaneric:
> yes, we can do it - how many bytes do we need?

Maybe keeping the length to match with 16 bytes? I define the maximum length of JSON-RPC method in the header file.

grcevski (approved):
This is some great work here @titaneric and thank you so much! Once tests pass we should be able to merge this.

titaneric:
I would try to check the integration test locally. If it runs without error, I would ping you to trigger the action

rafaelroquetto:
@titaneric sounds good! I'll keep an eye here

titaneric:
k8s test still fail, I would take a look

titaneric:
I ran the k8s integration test locally, this was what I got
image

I suspect the error came from the beyla's config, heuristic unmatching policy.
I added the new `/jsonrpc` path to `pkg/components/transform/route/cluster_test.go`
image

This was what I got
image

May I remove the `unmatched: heristic` in `test/integration/k8s/manifests/06-beyla-daemonset-multi-node.yml`?

titaneric:
I think the k8s integration test error came from the original k8s manifests since I have similar error in the main branch

titaneric:
I think the otel collect is not correctly initailized in the k8s [integration test](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/15931741853/job/44945875685?pr=161).
```
=== RUN   TestMultiNodeTracing
=== RUN   TestMultiNodeTracing/Beyla_is_able_to_generate_distributed_traces_go->go_jsonrpc->python->ruby
=== RUN   TestMultiNodeTracing/Beyla_is_able_to_generate_distributed_traces_go->go_jsonrpc->python->ruby/it_sends_connected_traces_for_all_services
    eventually.go:81: 
        	Error Trace:	/home/runner/work/opentelemetry-ebpf-instrumentation/opentelemetry-ebpf-instrumentation/test/integration/k8s/daemonset_multi_node/k8s_daemonset_multi_node_traces_test.go:74
        	            				/home/runner/work/opentelemetry-ebpf-instrumentation/opentelemetry-ebpf-instrumentation/vendor/github.com/mariomac/guara/pkg/test/eventually.go:57
        	            				/opt/hostedtoolcache/go/1.24.4/x64/src/runtime/asm_amd64.s:1700
        	Error:      	"[]" should have 1 item(s), but has 0
        
--- FAIL: TestMultiNodeTracing (180.00s)
    --- FAIL: TestMultiNodeTracing/Beyla_is_able_to_generate_distributed_traces_go->go_jsonrpc->python->ruby (180.00s)
        --- FAIL: TestMultiNodeTracing/Beyla_is_able_to_generate_distributed_traces_go->go_jsonrpc->python->ruby/it_sends_connected_traces_for_all_services (180.00s)
FAIL
2025/06/27 18:13:47 INFO exporting cluster logs component=kube.Kind directory=../../../../testoutput/kind
2025/06/27 18:13:49 INFO exported cluster logs component=kube.Kind out="Exporting logs for cluster \"test-kind-cluster-otel-multi\" to:\n../../../../testoutput/kind"
2025/06/27 18:13:49 ERROR dumping prometheus metrics component=kube.Kind error="querying prometheus: Get \"http://localhost:39090/api/v1/query?query=%7B__name__%21=%22%22%7D\": read tcp 127.0.0.1:42864->127.0.0.1:39090: read: connection reset by peer"
```

I have the similar error in my local side, and I suspect the error is came from the [kind cluster config](test/integration/k8s/manifests/00-kind-multi-node.yml), the worker node does NOT correctly mount the local configs into worker node's host path. I exec into kind's worker node, and do not find any files under `/configs` path.
```yaml
  - role: worker
    labels:
      cluster.x-k8s.io/cluster-name: beyla-k8s-test-cluster
    kubeadmConfigPatches:
      - |
        kind: JoinConfiguration
        nodeRegistration:
          kubeletExtraArgs:
            node-labels: "deployment/zone=other-progs"  
    extraMounts:
      # configuration files that need to be mounted in the host
      - hostPath: ../../configs
        containerPath: /configs
      # testoutput folder to store coverage data
      - hostPath: ../../../../testoutput
        containerPath: /testoutput
```

That cause the otel collector config, which is store in configs PVC (mount host path `/configs`), failed to be read by the otel collector, initalize failed.

titaneric:
Also, the image used  in`utestserver`  [deployment](test/integration/k8s/manifests/05-uninstrumented-few-services.yml), seems that it does not work in arm64. The pod failed immediately and show "execuation format error", which would cause its trace failed to be retrieved in the test.

> This way the code will first match /jsonrpc and then go to the route heuristic.

This trick do work, thank you! I would get back to the integration test in the tomorrow.

titaneric:
I think the integration test is a little flakey. It pass in my local environment... May need to re-trigger the test again, would you @rafaelroquetto ?

titaneric:
> Apologies @titaneric for the delayed response. It has been a long holiday weekend here in Canada and we are all coming back to work today. I've retriggered the test, let's see how it fares and then take it from there.

That's ok! However, the test still failed. There is some errors in the jsonrpc's trace and metrics.
It seems that the test result is limited. I could not easily check the dumped metrics and traces from it. I would dig deeper and try to re-produce the error on my side.

titaneric:
Yes, I could download it. However, unlike k8s integration test, the docker compose integration test don't have the dumped metrics and tracee for debug purpose.

test/integration/red_test.go
```diff
@@ -222,6 +237,60 @@ func testSpanMetricsForHTTPLibrary(t *testing.T, svcName, svcNs string) {
...
+func testSpanMetricsForJSONRPCHTTP(t *testing.T, svcName, svcNs string) {
+	pq := prom.Client{HostPort: prometheusHostPort}
+	var results []prom.Result
+
+	expectedSpanName := "Arith.M /jsonrpc"
+
+	// Test span metrics
+	test.Eventually(t, testTimeout, func(t require.TestingT) {
+		var err error
+		results, err = pq.Query(`traces_span_metrics_duration_count{` +
```
rafaelroquetto: So @titaneric , I managed to debug your test, and the problem lies on the name of this metric. Try this:
```suggestion
		results, err = pq.Query(`traces_span_metrics_duration_seconds_count{` +
```

rafaelroquetto: I also rebased your branch (locally of course) on top of the latest `main`. That and the comment above did the trick. Once you update your branch, I will retrigger the test and we should be good to go ;)

rafaelroquetto: For future reference, to debug the test, I manually launched the docker compose file, triggered a request with curl, and checked the results in the prometheus ui:
```
OTEL_EBPF_EXECUTABLE_PATH=testserver docker compose -f test/integration/docker-compose.yml up

curl -H "Content-Type: application/json" -d @body/formated.json  http://localhost:8088/jsonrpc

firefox http://localhost:9090
```

rafaelroquetto:
@titaneric you can replace bpf_memicmp with stricmp in `common/strings.h `

test/integration/red_test.go
titaneric: Thank you for your advice, it's very helpful!

rafaelroquetto:
Thank you @titaneric !

PR #171, opened 2025-06-16
Fix beyla_genfiles.go

mariomac:
When we try to run `make generate`inside a dockerfile, we get this error:

```
2.827 error waiting for child process: failed to start program: exec: "docker": executable file not found in $PATH
```

This seems to be caused by `beyla_genfiles.go` go:Generate header.

>that was just so if people did not want to call go run beyla_genfiles , but instead decided to do go generate ./...  - then it would in turn go run beyla_genfiles to make the generation process more "go-esque"
>I think it can be removed at this stage

### DIFF
--- a/cmd/beyla-genfiles/beyla_genfiles.go
+++ b/cmd/beyla-genfiles/beyla_genfiles.go
@@ -1,5 +1,3 @@
-//go:generate go run beyla_genfiles.go
-
 package main
 
 import (


rafaelroquetto (changes requested):
Actually, looking at the target:
```
.PHONY: generate
generate: export BPF_CLANG := $(CLANG)
generate: export BPF_CFLAGS := $(CFLAGS)
generate: export BPF2GO := $(BPF2GO)
generate: bpf2go
	@echo "### Generating files..."
	@BEYLA_GENFILES_RUN_LOCALLY=1 go generate cmd/beyla-genfiles/beyla_genfiles.go

```
I was wondering why `@BEYLA_GENFILES_RUN_LOCALLY=1` is not being picked up. This should be preventing `beyla_genfiles` to invoke docker:

```
	if cfg.RunLocally {
		runLocally(wd)
	} else {
		runInContainer(wd)
	}
```
But then, looking at the sources:
```
	RunLocally      bool   `env:"OTEL_EBPF_GENFILES_RUN_LOCALLY"      envDefault:"false"`
```

so instead of removing the go:generate line, we should update the Makefile to have 

```
OTEL_EBPF_GENFILES_RUN_LOCALLY=1
```
instead of
```
BEYLA_GENFILES_RUN_LOCALLY=1
```

There are also other variables that need to be updated (such as BEYLA_GENFILES_GEN_IMG)

PR #172, opened 2025-06-16
Fix genfiles environment vars in Makefile

rafaelroquetto:
The Makefiles were using the old `BEYLA_GENFILES` prefix, which was causing the build to break.


grcevski (approved):
LGTM!

PR #173, opened 2025-06-16
New OTel span metrics format and port pid handling

grcevski:
OBI can generate span metrics directly which enables tools that rely on span metrics to render service graphs, without having to go through traces, which if are sampled will produce unreliable data. The format that Beyla/OBI initially used was based on the original donation of Tempo span metrics generator. However, since that donation the OTel community has changed the format and this PR fixes that support. Port of https://github.com/grafana/beyla/pull/1989

Additionally, the current OBI implementation added the PID to the target_info which can significantly increase cardinality. This PR ports the fixes related to multiple PIDs per container from Beyla. Port of https://github.com/grafana/beyla/pull/1960 

Also makes host info metrics explicitly enabled. Port from https://github.com/grafana/beyla/pull/1990

### DIFF
--- a/pkg/beyla/config.go
+++ b/pkg/beyla/config.go
@@ -39,7 +39,7 @@ const (
 )
 
 const (
-	defaultMetricsTTL = 70 * time.Minute
+	defaultMetricsTTL = 5 * time.Minute
 )
 
 var DefaultConfig = Config{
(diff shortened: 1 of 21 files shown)


pkg/beyla/config.go
```diff
@@ -39,7 +39,7 @@ const (
 )
 
 const (
-	defaultMetricsTTL = 70 * time.Minute
+	defaultMetricsTTL = 5 * time.Minute
```
grcevski, line 42: Expiration works now correctly, we don't need the insanely high interval anymore

rafaelroquetto (approved):

PR #176, opened 2025-06-17
Make AppendKubeMetadata public

marctc:
Required for refactors in Beyla


grcevski:
The test failure is on a new test, we need @rafaelroquetto to take a look at this. If it persists please add t.Skip to the test.

PR #177, opened 2025-06-17
Add route ignore logic back to routes

marctc:
This logic was somehow removed when donated Beyla.
This PR brings it back in order to proverly vendorise this in Beyla.


mariomac:
Ah wait! Sorry for not remembering this before.

We decided to deprecate/remove this option because:
1- it was redundant with current attribute filtering
2- it was not compliant with OTEL standards (e.g. using regexpes instead globs)

I think we can close this PR and not need to add it.

PR #181, opened 2025-06-17
Stop go HTTP context propagation on invalid headers

grcevski:
We add our context propagation code for HTTP in Go on `header.writeSubset`, where we look for sufficient buffer in the writer to write our headers. The code relies on using the header pointer as a key to distinguish between multiple headers. However, according to the Go HTTP code, the writeSubset function can be called on `nil` header, effectively potentially concurrently modifying the same data structure between client and server calls, which may result in a crash of the application. 

I've not been able to reproduce it, but a user has reported this following stack trace from running the Dex OpenID server:
image

The crash happens after writeSubset is invoked on HTTP server response, where we don't actually write anything, or at least our intention is not to write anything. The code tests for a header setup by a client request and it also ensures that the goroutine that setup the header is handling a client request. So the only possibility for this code to do anything is that an outgoing client request is sharing the exact same header structure as the outgoing server response, and the client and server use the exact same goroutine. This is impossible, unless it's a special condition which seems to exist in the code where the header value is `nil`.

This PR adds a header and buffer nil check and also adds additional safety net to avoid doing anything if for some reason we weren't able to correctly resolve the offsets of the Go application.

### DIFF
--- a/bpf/gotracer/go_nethttp.c
+++ b/bpf/gotracer/go_nethttp.c
@@ -592,9 +592,15 @@ int beyla_uprobe_writeSubset(struct pt_regs *ctx) {
 
     void *header_addr = GO_PARAM1(ctx);
     void *io_writer_addr = GO_PARAM3(ctx);
-    off_table_t *ot = get_offsets_table();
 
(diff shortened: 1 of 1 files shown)


mariomac (approved):
LGTM!

PR #186, opened 2025-06-18
Enable more clang-tidy diagnostics

rafaelroquetto:
Enable a new class of clang-tidy diagnostics (clang-analyzer). One of the diagnostics (dead store) already pointed out a couple of places in which we were reading/writing variables that were not being used.

The `-clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling` has been disabled as it implicates `__builtin_memcpy` and it's all over the place at this stage (and there's no ebpf alternative, in a regular scenario we'd replace that with `memcpy_s`).


grcevski (approved):
LGTM!

PR #188, opened 2025-06-18
bpf: use unix.ByteSliceToString for debug logs instead of unsafe

mmat11:
I was originally working on a branch without https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/commit/08d9fca59d9bad25b249bce5acd443a7019f7e40 and I realized later I fixed something already fixed in main; This changes nothing except using a sys/unix function instead of `readString`

### DIFF
--- a/pkg/components/ebpf/logger/logger.go
+++ b/pkg/components/ebpf/logger/logger.go
@@ -1,14 +1,13 @@
 package logger
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"io"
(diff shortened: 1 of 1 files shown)


rafaelroquetto (approved):
Nice - thanks for catching that. I'm approving it, but I suspect there might be other places in the code with a similar approach - feel free to amend this PR if you find them and feel like it.

mmat11:
thanks!

PR #190, opened 2025-06-19
Remove unused mountinfo files

marctc:
This files are not being used. After reducing the privileges to run Beyla, we don't need to
mount anything, so this is not needed.


mariomac (approved):
Good catch!

PR #192, opened 2025-06-19
Cleanup API for extra resource attributes

mariomac:
Simplifies the way extra resource attributes can be passed.


rafaelroquetto (approved):
🚀🚀🚀

PR #194, opened 2025-06-19
Add process minimum age to filter out short lived processes if needed

corneliu-calciu:
Added minimum process age to be considered for discovery.

The new process detection logic will ignore a new process until its age reach a minimum value. As an example we may set a 3 seconds threshold as part of the monitoring configuration.
Any process will be considered as suitable for monitoring once is passing that configurable limit.
The noise of short lived processes can be filtered out.

Configuration example:
discovery:
exclude_services:
- k8s_namespace: (istio-system|kube-node-lease)
services:
- k8s_namespace: (acmefitness)
min_process_age: 5s

Moved the RP from https://github.com/grafana/beyla/pull/2066

### DIFF
--- a/pkg/components/discover/watcher_proc.go
+++ b/pkg/components/discover/watcher_proc.go
@@ -48,6 +48,12 @@ type ProcessAttrs struct {
 	metadata       map[string]string
 	podLabels      map[string]string
 	podAnnotations map[string]string
+	// Name of the process
+	processName string
+	// Status of the process
+	processStatus string
+	// Age of the process
+	processAge time.Duration
 }
 
 func wplog() *slog.Logger {
@@ -96,7 +102,7 @@ type pollAccounter struct {
 	// same process might appear several times
 	pidPorts map[pidPort]ProcessAttrs
 	// injectable function
-	listProcesses func(bool) (map[PID]ProcessAttrs, error)
+	listProcesses func(bool, time.Duration) (map[PID]ProcessAttrs, error)
 	// injectable function
 	executableReady func(PID) (string, bool)
 	// injectable function to load the bpf program
@@ -133,7 +139,7 @@ func (pa *pollAccounter) run(ctx context.Context) {
 	go pa.watchForProcessEvents(ctx, log, bpfWatchEvents)
 
 	for {
-		procs, err := pa.listProcesses(pa.portFetchRequired())
+		procs, err := pa.listProcesses(pa.portFetchRequired(), pa.cfg.Discovery.MinProcessAge)
 		if err != nil {
 			log.Warn("can't get system processes", "error", err)
 		} else {
@@ -211,6 +217,7 @@ func (pa *pollAccounter) watchForProcessEvents(ctx context.Context, log *slog.Lo
 // snapshot compares the current processes with the status of the previous poll
 // and forwards a list of process creation/deletion events
 func (pa *pollAccounter) snapshot(fetchedProcs map[PID]ProcessAttrs) []Event[ProcessAttrs] {
+	log := wplog()
 	var events []Event[ProcessAttrs]
 	currentPidPorts := make(map[pidPort]ProcessAttrs, len(fetchedProcs))
 	reportedProcs := map[PID]struct{}{}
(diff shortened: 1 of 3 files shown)


pkg/components/discover/watcher_proc_test.go
```diff
@@ -145,7 +145,7 @@ func TestProcessNotReady(t *testing.T) {
 		},
 	}
 
-	procs, err := acc.listProcesses(true)
+	procs, err := acc.listProcesses(true, 0)
```
grcevski: Can you please add a test in this file that ensures that young processes are not found?

pkg/services/criteria.go
```diff
@@ -91,6 +91,10 @@ type DiscoveryConfig struct {
 
 	// Disables instrumentation of services which are already instrumented
 	ExcludeOTelInstrumentedServices bool `yaml:"exclude_otel_instrumented_services" env:"OTEL_EBPF_EXCLUDE_OTEL_INSTRUMENTED_SERVICES"`
+
+	// Min process age to be considered for discovery.
+	//nolint:undoc
+	MinProcessAge time.Duration `yaml:"min_process_age" env:"BEYLA_MIN_PROCESS_AGE"`
```
grcevski: Can we please prefix the new environment variable with OTEL_EBPF instead of BEYLA_? When OBI is built inside Beyla we'll do the substitution, so essentially in beyla both BEYLA_MIN_PROCESS_AGE and OTEL_EBPF_MIN_PROCESS_AGE will be accepted.

grcevski: Do you mind adding a reasonable default in `config.go` for the process age, perhaps something like a 5 seconds?

grcevski (commented):
This is looking good already, I left a few comments that I think we should address. Thanks for working on this!!!

pkg/components/discover/watcher_proc.go
```diff
@@ -48,6 +48,12 @@ type ProcessAttrs struct {
 	metadata       map[string]string
 	podLabels      map[string]string
 	podAnnotations map[string]string
+	// Name of the process
```
grcevski: These process attributes get passed around, if we need them just to report the values in a logger, maybe we shouldn't be storing them?

pkg/components/discover/watcher_proc.go
```diff
@@ -325,7 +335,7 @@ func (pa *pollAccounter) checkNewProcessNotification(pid PID, reportedProcs, not
 
 // fetchProcessConnections returns a map with the PIDs of all the running processes as a key,
 // and the open ports for the given process as a value
-func fetchProcessPorts(scanPorts bool) (map[PID]ProcessAttrs, error) {
+func fetchProcessPorts(scanPorts bool, minProcessAge time.Duration) (map[PID]ProcessAttrs, error) {
```
grcevski: This is the most important point which I think @mariomac was concerned about. Namely, when `scanPorts` is true, we must accept processes regardless of their age.

This is because scanning ports is an expensive operation on disk, it takes a lot of reading of files etc. We don't want to do this scan every time we poll, so this part is asynchronous. We run a BPF watcher program for new ports being opened, once we get an event from the BPF side, we turn a flag on to scan for new ports. If we ignore the process because it's young on the scan we do based on the async event, we won't have a chance to find it again unless someone opens another port and sets the flag to scan for ports.

pkg/components/discover/watcher_proc_test.go
corneliu-calciu: Done. Added new test at the end of the file checking three cases.

pkg/services/criteria.go
corneliu-calciu: Done

corneliu-calciu: Done.

pkg/components/discover/watcher_proc.go
corneliu-calciu: Done.

pkg/components/discover/watcher_proc.go
corneliu-calciu: Done. If scanPorts is true will not consider the process age.

corneliu-calciu:
> Thanks @corneliu-calciu for the reply!
> 
> > This is exactly the scenario I want to address. The Beyla PODs are already running. A new process will be created on a node.
> > Now we can specify what is the minimum life span needed to accept and instrument the process.
> > If we have a large number of processes created and destroyed within a 1-2 seconds interval are just noise, especially on a > Kubernetes environment where we expect long lived micro-services.
> 
> But imagine a new long-lived process is created (e.g. a web server). Beyla gets the notification of a new process/port, then before deciding whether it has to be instrumented, checks the age of the process, and since it's very recent, the process will be discarded. So our new web server is never instrumented.
> 
> Am I right, or am I missing something?

Hi Mario

The process scanning is a periodical activity. Basically the same process will be ignored in the first 5 seconds, and once will pass the minimum age will be handled.

Best Regards
Corneliu

grcevski:
> Beyla gets the notification of a new process/port, then before deciding whether it has to be instrumented, checks the age of the 

So as you noticed, there's a problem with the new port event, since those are async and reported by eBPF, but @corneliu-calciu added a check to accept the process regardless of the age if the event was port open.

PR #197, opened 2025-06-19
Ensure we support more than one pid per container, in pids to pods

grcevski:
OBI can match processes by various attributes, namely, they can be matched by their "process" attributes, such as executable name or ports they open. OBI can also match processes by extended attributes in certain environments like Kubernetes, for example a Kubernetes namespace.

To perform the matching of these extended attributes, like Kubernetes metadata attributes, we must match the process information (the process ID - PID) to the Kubernetes pod metadata. To make things more complicated, the pod metadata may not be present at the time we discover the process, which means we'll discover a process, discard it based on the existing selection criteria, but then we need to report it as found again once we get Kubernetes metadata for it, so that we can match it by the Kubernetes selection criteria.

To perform this matching step, OBI keeps a mapping of container information to process ID. At a time a process is discovered, we read it's container ID and record it as seen container. When Kubernetes pod information is received, we pull all the container information for the pod and match it to the previously discovered PID container infromation. This is known as the PID to POD functionality in OBI.

The problem is that the map we kept container to PID information assumed that there will only be one PID per pod that we want to instrument, e.g. a simple string -> PID map is used for the `processByContainer` mapping datastructure. This is insufficient, since there are deployment scenarios where multiple processes can be found in a container.

This PR extends the map to contain an array of PIDs instead of single PID mapping.

Relates to https://github.com/grafana/beyla/issues/2046

### DIFF
--- a/pkg/components/discover/watcher_kube.go
+++ b/pkg/components/discover/watcher_kube.go
@@ -30,7 +30,7 @@ type watcherKubeEnricher struct {
 	// cached system objects
 	mt                 sync.RWMutex
 	containerByPID     map[PID]container.Info
-	processByContainer map[string]ProcessAttrs
+	processByContainer map[string][]ProcessAttrs
 
 	podsInfoCh chan Event[*informer.ObjectMeta]
 	output     *msg.Queue[[]Event[ProcessAttrs]]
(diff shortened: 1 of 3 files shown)


rafaelroquetto (approved):
Groovy!

PR #198, opened 2025-06-19
Reorganise ebpf file structure - part II

rafaelroquetto:
This is a follow up of https://github.com/grafana/beyla/pull/1791 - see that PR's decription for context.

There's still plenty to do and other PRs will follow from time to time.

The present PR solely targets the maps in `generictracer`.

### DIFF
--- a/bpf/generictracer/iovec_len.h
+++ b/bpf/generictracer/iovec_len.h
@@ -1,0 +1,3 @@
+#pragma once
+
+enum { k_iovec_max_len = 512 };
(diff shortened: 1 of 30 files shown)


rafaelroquetto:
As discussed offline, those ebpf constants are actually enums, so all good here.

For anyone who come across this in the future, the extended rationale is that a constant defined as an enumerator, unlike a macro, cannot be undefined or redefined, is type safe, but unlike a `static const T t = ...` constant, it works pretty much like an rvalue, i.e. you can't take its memory address and has no storage - pretty much like a macro define.

grcevski (approved):

PR #202, opened 2025-06-20
Fix reading of TCP event buffer

rafaelroquetto:
When reading `u_buf` into the TCP request/response buffers, we were passing `K_TCP_MAX_LEN`/ `K_TCP_RES_LEN`, which is the maximum buffer size allowed in the ring buffer - that means  that, in architectures that allow it (e.g. x64), we were just reading past `u_buf` limits (i.e. it would continue to read as long as it does not hit the userspace memory space).

Not only we are sending garbage into the buffer after `bytes_len`, but we could be sending data that shouldn't make to the userspace there to begin with.

Instead we should clamp and read at most `bytes_len` (i.e. the effective size of `u_buf`) into the ring buffer event.


mariomac (approved):
LGTM!

grcevski:
Looks like the change didn't work, SQL tests are failing which is where this would have mattered.

grcevski:
The 5.15 test looks like a problem with the collector starting, I re-launched it. 

I'm curious, what was the problem with the previous change? It seems like you force pushed, so I can't see the before and after, looks similar enough :)

rafaelroquetto:
@grcevski I reverted the `bpf_probe_read_kernel()` to `bpf_probe_read()` - it indeed dispatches to user or kernel internally

PR #219, opened 2025-06-25
Fix tcp redis error parsing

NimrodAvni78:
* fix redis errors which were not parsed correctly
* add additional error attributes and status to redis spans
* fix trailing space in redis statements


test/integration/test_utils.go
```diff
@@ -22,6 +27,19 @@ import (
...
+}
+
+type TestCase struct {
+	Route     string
+	Subpath   string
+	Comm      string
+	Namespace string
+	Spans     []TestCaseSpan
+}
+
```
mariomac: Since this looks very focused on the Python+Redis test (and there are actually other local testCase structs), what do you think about moving `TestCase` and `TestCaseSpan` to the `red_test_python_redis.go` file?

NimrodAvni78: iv'e added it to test_utils since i think its kind of generic, doing an api call and expecting some spans to appear
it can also be reused in other tests instead of kind of doing a lot of assertions manually using the struct to declaratively say what you expect

but i can move it to the `red_test_python_redis.go` if you feel its more suited to be there

test/oats/redis/yaml/oats_go_redis.yaml
```diff
@@ -15,7 +15,7 @@ expected:
             db.operation.name: set
             db.system.name: redis
             server.port: 6379
-            db.query.text: "set beyla rocks "
+            db.query.text: "set beyla rocks"
```
marctc: nit: i think obi can rock here too?

NimrodAvni78: i didnt change to go redis application to match python, just the assertion since i removed the trailing space
i can change it to match

test/integration/test_utils.go
mariomac: I'd move it the closer to where it is used, as we can always move it here. I'd also add few comments giving some context about the structure.

test/oats/redis/yaml/oats_go_redis.yaml
mariomac: Both OBI and Beyla rock 😄

NimrodAvni78:
@mariomac @marctc addressed comments

grcevski (approved):
Amazing work!

PR #220, opened 2025-06-25
Add YAML marshaller for services.GlobAttr, PortEnum and RegexpAttr

mariomac:
This introduces a breaking change at API level: `services.NewGlob` and `services.NewPathRegexp` now receive directly a string (they were always invoked with `regexp/glob.MustCompile` as they are just internal utility functions).
Also `services.NewPathRegexp` has been renamed to `NewRegexp` for coherence.


grcevski (approved):
LGTM!

PR #224, opened 2025-06-26
Remove trace internal metrics

grcevski:
The trace internal metrics currently produce a single trace span of the collector API export which provides no value on top of the existing internal metrics we have. The exported span is confusing for end users because they see OBI as a service, producing spans they don't understand and eventually they would have to pay for.


marctc (approved):
Nice!

PR #241, opened 2025-07-02
Add Initial support for mongodb

NimrodAvni78:
The purpose of this pr is to support calls, creating spans and metrics from them
the current scope of this pr:

* use the unknown TCP path and try to parse mongo packets from there (parsing by [mongodb wire format](https://www.mongodb.com/docs/manual/reference/mongodb-wire-protocol/))
    * initially support only `OP_MSG` commands, which are the most common in modern MongoDB versions (v5.1+)
    * support for `OP_COMPRESSED` and legacy operations like `OP_QUERY`, `OP_GET_MORE`, `OP_INSERT`, `OP_UPDATE`, `OP_DELETE` can be added later

* build spans with relevant fields from [mongodb semantic conventions](https://opentelemetry.io/docs/specs/semconv/database/mongodb/)
    * operation name is the first key in the `OP_MSG` bson message request
    * collection name is usually the value of the first key in the `OP_MSG` bson message request (in most non admin commands)
    * db name is the `$db` field in the `OP_MSG` bson message request
    * status is determined by the `ok` field in the `OP_MSG` bson message response
    * errors are determined by the `errmsg` and `code` field in the `OP_MSG` bson message response

open questions:
* what commands are interesting to us? this is the list of [mongo db commands](https://www.mongodb.com/docs/manual/reference/command/), are only `find`, `insert`, `update`, `delete` interesting to us?
    * do we care about other types of commands, aggreagtion? health checks? admin commands? do we want it to be configurable by the user?
* for certain some of the mongodb requests won't fully fit in the buffers we send to user space, even after integrating with [large buffer solution implemented here](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/118),
  in uncompressed payloads we can theoretically handle this by implementing an incomplete bson message parser, which will parse the data it has and work by that (omitting things like the db namespace, or even the ok field in response if it doesn't exist in the buffer)
* currently the rbuf on the event only contains the header (first 16 bytes), im guessing they are written in different calls but are part of the same packet, so for now we always parse
responses as success, we might want to implement some packet reassembly logic to handle this
* do we want to support requests and responses that span across multiple packets? mongodb wire protocol allows this and signals it with headers and flags like responseTo, moreToCome and exhaustAllowed
right now its only reqeust / response in the same event


grcevski:
This is great @NimrodAvni78!

Let me answer what I think about the open questions:

> what commands are interesting to us? this is the list of [mongo db commands](https://www.mongodb.com/docs/manual/reference/command/), are only find, insert, update, delete interesting to us?
do we care about other types of commands, aggreagtion? health checks? admin commands? do we want it to be configurable by the user?

I think that list is good to start with, if it's anything else maybe we can just put a `*` for the command and let users drive the need further?

> for certain some of the mongodb requests won't fully fit in the buffers we send to user space, even after integrating with https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/118,
in uncompressed payloads we can theoretically handle this by implementing an incomplete bson message parser, which will parse the data it has and work by that (omitting things like the db namespace, or even the ok field in response if it doesn't exist in the buffer)

Yeah that makes sense, we can start with simple use of the bson parser and extend further. if it can't be parsed because of incomplete error message we can report `*` again and document that users should use larger buffers.

> currently the rbuf on the event only contains the header (first 16 bytes), im guessing they are written in different calls but are part of the same packet, so for now we always parse
responses as success, we might want to implement some packet reassembly logic to handle this

That makes sense. I think this is good to start with, we can make the TCP packets delay the send like we do for HTTP and capture more of the response in the future. I'm OK with having just success for now.

> do we want to support requests and responses that span across multiple packets? mongodb wire protocol allows this and signals it with headers and flags like responseTo, moreToCome and exhaustAllowed
right now its only reqeust / response in the same event

Sound great for a first iteration as it is, we can extend later and use some sort of userspace cache if we need to.

grcevski:
> i mean i can get the command either way from the first bson key, i can extract the collection name for known commands like we said, and for the rest i can just omit the collection name, beside healthcheck commands which we know which are quite spammy, how does that sound?

That sounds great!

> as for the rest of the comments, your saying for first iteration start by not changing any kernel space and just going down unknown tcp path, adding * for operation if we can't fully parse the message, also failing to parse response status and error. and address integration with large buffers and support multi packet requests in a later pr?

Yeah, I think we can get something going with simply using the userspace and falling back to * when we can't deal with it. Once the large buffers are there we can make it better and iterate from there.

> you mean creating a different event type in kernel space?

So for the request we already keep on appending the buffers, until the TCP traffic switches the direction. At the moment of switch, we finish the TCP request and call it a day with whatever was in that buffer. To be able to pile up more of the response we need to keep on receiving and not terminate the TCP event just yet, or send additional buffers and delay the userspace processing of the TCP event to ensure we see more of the response. For HTTP we delay making the HTTP response on seeing `200 OK` (or something else) we keep the request "unsent" until we see TCP close or connection reuse etc. It's quite complicated and it's hard to get right, especially with TLS. So, let's go with the simpler approach and maybe we can figure something special for MongoDB a bit later.

Maybe as you say the trick would be to send additional event, delay in userspace and refresh the events as new response buffers arrive. But let's follow-up with that, easier to review smaller PRs :).

pkg/components/ebpf/common/mongo_detect_transform.go
```diff
@@ -0,0 +1,530 @@
...
+	}
+	if pendingRequest == nil {
+		return nil, false, errors.New("no MongoDB request or response found in the message")
+	}
+	requests.Add(key, pendingRequest)
+	if !moreToCome && !isRequest {
+		requests.Remove(key)
+		// If this is a response and there are no more sections to come, we can finalize the request
+		return pendingRequest, false, nil
+	}
```
rafaelroquetto: what do you think of
```suggestion
	if !moreToCome && !isRequest {
		// If this is a response and there are no more sections to come, we can finalize the request
		return pendingRequest, false, nil
	}
	requests.Add(key, pendingRequest)
```

to avoid the penalty of adding the key just to end up removing it anyway?

pkg/components/ebpf/common/mongo_detect_transform.go
```diff
@@ -0,0 +1,530 @@
...
+			requestID: header.ResponseTo,
+		}
+		time = endTime
+	} else {
+		key = MongoRequestKey{
+			connInfo:  connInfo,
+			requestID: header.RequestID,
+		}
+		time = startTime
+	}
```
rafaelroquetto: would it make sense to wrap this into a function that could be called like:
```
key := makeRequestKey(header)
```

in order to improve clarity?

pkg/components/ebpf/common/mongo_detect_transform.go
```diff
@@ -0,0 +1,530 @@
...
+	inFlightRequest, ok := requests.Get(key)
+	if !ok && !isRequest {
```
rafaelroquetto: completely minor nit, but when trying to understand the code, it may be better to phrase it as `isResponse` rather than `!isRequest` as it reads much more naturally (took me a while to understand that `!isRequest` actually meant you were handling a response) - feel free to ignore this comment.

pkg/components/ebpf/common/mongo_detect_transform.go
```diff
@@ -0,0 +1,530 @@
...
+		// TODO (mongo) currently the response is only the header, since the client sends only the first 16 bytes at first,
+		// we need to fix the tcp path to send the response body as well
+		// for now we just dont add response section
+		requests.Remove(key)
+		// If this is a response and there are no more sections to come, we can finalize the request
+		return inFlightRequest, false, nil
+	}
+	pendingRequest, moreToCome, err = parseMongoMessage(buf, *header, time, isRequest, inFlightRequest)
+	if err != nil {
+		return nil, false, fmt.Errorf("failed to parse MongoDB response: %w", err)
```
rafaelroquetto, line 149: do we need to `requests.Remove(key)` here as well? I am thinking of the case in which you have an `inFlightRequest` (which started at a previous iteration) but then for some reason  `parseMongoMessage` returns an error, but I am not sure

pkg/components/ebpf/common/mongo_detect_transform.go
```diff
@@ -0,0 +1,530 @@
...
+		}
+	} else {
+		if pendingRequest == nil {
+			return nil, errors.New("MongoDB response received but no pending request found")
+		}
+		if pendingRequest.ResponseSections != nil {
+			pendingRequest.ResponseSections = append(pendingRequest.ResponseSections, sections...)
+		} else {
+			pendingRequest.ResponseSections = sections
+		}
```
rafaelroquetto: I think go allows writing
```suggestion
		pendingRequest.ResponseSections = append(pendingRequest.ResponseSections, sections...)
```
i.e. it deals with the nil case already

pkg/components/ebpf/common/mongo_detect_transform.go
```diff
@@ -0,0 +1,530 @@
...
+				RequestSections: sections,
+				StartTime:       time,
+				Flags:           flagBits,
+			}
+		} else {
+			pendingRequest.RequestSections = append(pendingRequest.RequestSections, sections...)
+			pendingRequest.Flags = flagBits
+			if pendingRequest.StartTime > time {
+				pendingRequest.StartTime = time
+			} else if pendingRequest.EndTime < time {
```
rafaelroquetto: non-rhetorical question: what happens when  StartTime > time > EndTime? I mean, it shouldn't happen, I am just wondering if we should ditch the `else if` for a standalone `if` (in which case we'd end up with StarTime = EndTime = time)

rafaelroquetto (approved):
Good stuff! I've left a bunch of minor and subjective comments, feel free to adopt those you think are worth it.

But from my end, LGTM - perhaps we should wait for a second +1 from @grcevski as he's been more on top of this before merging.

pkg/components/ebpf/common/mongo_detect_transform.go
```diff
@@ -0,0 +1,530 @@
...
+		} else {
+			pendingRequest.ResponseSections = sections
+		}
+		pendingRequest.Flags = flagBits
+	}
+	return pendingRequest, nil
+}
+
+func parseOpMessage(buf []uint8, time int64, isRequest bool, pendingRequest *MongoRequestValue) (*MongoRequestValue, bool, error) {
+	// MONGODB_OP_MSG packet structure:
```
marctc: nit: add this comment before the function definition?

pkg/components/ebpf/common/mongo_detect_transform.go
NimrodAvni78: we might but i think we shouldn't remove it because of some false positives (maybe some packet that looks like a mongodb header with the same request id but it isnt? and the response is about to arrive in another packet)
anyway i dont think it really matters since the lru will remove it eventually

pkg/components/ebpf/common/mongo_detect_transform.go
NimrodAvni78: the end time condition shouldnt even be in the request path here but in the response, so i moved it

NimrodAvni78:
@marctc @rafaelroquetto addressed comments

pkg/components/ebpf/common/mongo_detect_transform.go
rafaelroquetto: Right - could you add a comment explaining this for future reference?

grcevski (approved):
This is really great!

PR #242, opened 2025-07-02
Action to create generator image + Update it to Cilium 0.19.0

mariomac:
This PR addresses some issues:
1. We are currently using a `grafana/beyla-ebpf-generator` image so after this we will be able to use our own `obi-generator` image.
2. There is a circular dependency if we want to update Cilium version with the current means: if we update the cilium version in the generator image, OBI builds will be broken until we update it in the `go.mod`. Same if we first update cilium in the `go.mod`.

Now, instead of generating and pushing the image on each push to main (which is unnecessary because the image is not updated so often), we need to manually publish a new image. Allowing to override the tags and cilium version in the triggering GH Action.

After this PR is merged, we can replace the current `grafana...` artifact by an OTEL image, and update the Cilium version.

### DIFF
--- a/generator.Dockerfile
+++ b/generator.Dockerfile
@@ -12,7 +12,7 @@ FROM base AS dist
 
 WORKDIR /src
 
-ARG EBPF_VER
+ENV EBPF_VER=v0.19.0
 
 RUN apk add clang llvm19 wget
 RUN apk cache purge
--- a/.github/workflows/generator-image.yml
+++ b/.github/workflows/generator-image.yml
@@ -1,0 +1,98 @@
+name: Publish OBI Docker Generator Image
+on:
+  workflow_dispatch:
+    inputs:
+      override_image_tag:
+        type: string
+        description: Override image tag (empty for current timestamp)
+
+# Set restrictive permissions at workflow level
+permissions:
+  contents: read
+
+env:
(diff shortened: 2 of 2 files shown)


generator.Dockerfile
```diff
@@ -12,7 +12,7 @@ FROM base AS dist
 
 WORKDIR /src
 
-ARG EBPF_VER
+ENV EBPF_VER=v0.19.0
```
rafaelroquetto, line 15: The reason this was originally an ARG is so that this doesn't drift from the `CILIUM_EBPF_VER` in the `Makefile`, avoiding having to remember to update this in two places.

I don't have a strong opinion about this though, just thought I'd mention it - feel free to change it back or keep it as is.

rafaelroquetto (approved):

generator.Dockerfile
mariomac: The problem now is that I want to update Cilium version. If I update the version `go.mod` (and fix the code breaking changes), then the code would be generated with the old `bpf2go` version (and there is no guarantee that this would work).

This way we first update EBPF_VER, then publish the new generator image under a new tag, then we can update it in Go mod.

I think we can add a script in the image generation action to prevent that the EBPF_VER is never left to a smaller version than the version in go.mod.

PR #244, opened 2025-07-02
cleanup: replace some `bpf_printk` with `bpf_dbg_printk`

Andreagit97:
While running the code, I noticed some `bpf_printk` that could be leftovers since they seem to be called frequently or at least at each bpf prog invocation. If this is not the case and they are intentional, feel free to close this one :)


grcevski (approved):
Great find! I was maybe worried the tp_injector ones may not work, because the dbg_printk looks up the pid, but I guess it's not a banned helper.

grcevski:
Thanks so much for your contribution @Andreagit97 !

PR #253, opened 2025-07-03
cleanup: rename socket filter prog in netolly

Andreagit97:
If you run together netolly and appolly, you will see 2 EBPF programs with the same name `beyla_socket__http_filter`.  This is not ideal if you want to debug/benchmark EBPF programs. Moreover, this is not an HTTP-specific code, so it could make sense to change its name.


grcevski (approved):
Excellent!

PR #256, opened 2025-07-04
mysql: support prepared statements, refactor event handler

mmat11:


mariomac (commented):
Amazing!! Would it be possible to add some pepared statements into this test component: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/test/integration/components/sqlclient/gosqlite.go and add some test cases for it?

Another question. How is the prepared statement data accessed? Usually prepared statements are instantiated during the program startup. Wouldn't OBI be able to instrument the PreparedStatement if OBI starts after the instrumented process is already up and running? In understand this would be an unavoidable limitation

mmat11:
> Would it be possible to add some pepared statements into this test component: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/test/integration/components/sqlclient/gosqlite.go and add some test cases for it?

this handles specifically mysql packets, not sqlite
I saw there are some integration tests at https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/test/integration/red_test_python_sql.go but all or almost all of them use the postgres specific version of the docker compose. I was thinking of replicating the test with the mysql version of those, maybe this is the right PR for adding that coverage

> Another question. How is the prepared statement data accessed? Usually prepared statements are instantiated during the program startup. Wouldn't OBI be able to instrument the PreparedStatement if OBI starts after the instrumented process is already up and running? In understand this would be an unavoidable known limitation

Yes this is a known limitation, if we can't see the `PREPARE` command, we can't store the query which will be executed with `EXECUTE`

grcevski:
Epic! I really like this!

I took a look at the test failure, seems to be a problem related to parsing somehow, and we got less events than expected:
```
autoinstrumenter-1  | time=2025-07-04T11:20:54.057Z level=WARN msg="MySQL command ID unhandled" commandID=1
```
>  I was thinking of replicating the test with the mysql version of those, maybe this is the right PR for adding that coverage

Yes, please :). We already have MySQL test, so maybe just extend the program to prepare a statement and then use it. I would create a new HTTP API in the wrapper program, so first we call the regular direct query API and then once that test passes we start calling the prepared one. This way we'll ensure OBI has instrumented the program before we do the prepare, so we can see it for the cache you added.

Great stuff!

pkg/components/ebpf/common/tcp_detect_transform.go
```diff
@@ -148,6 +120,118 @@ func ReadTCPRequestIntoSpan(parseCtx *EBPFParseContext, cfg *config.EBPFTracer,
...
+			connInfo: event.ConnInfo,
+			stmtID:   stmtID,
+		})
+		if !found {
+			slog.Debug("MySQL EXECUTE command with unknown statement ID", "stmtID", stmtID)
+			return span, errFallback
+		}
+	case "QUERY":
+	default:
+		slog.Warn("MySQL command ID unhandled", "commandID", requestBuffer[sqlprune.MySQLHdrSize])
```
NimrodAvni78, line 239: should we warn on comm = "QUERY"?

pkg/components/ebpf/common/tcp_detect_transform.go
```diff
@@ -54,47 +39,34 @@ func ReadTCPRequestIntoSpan(parseCtx *EBPFParseContext, cfg *config.EBPFTracer,
 	// We might know already the protocol for this event
 	switch event.ProtocolType {
 	case ProtocolTypeMySQL: // MySQL
-		if len(requestBuffer) < sqlprune.MySQLHdrSize+1 {
-			slog.Warn("MySQL request too short, falling back to generic handler", "len", len(requestBuffer))
+		span, err := handleMySQL(parseCtx, event, requestBuffer, responseBuffer)
+		if errors.Is(err, errFallback) {
+			slog.Warn("MySQL: falling back to generic handler")
```
NimrodAvni78: i think also here this is not a warning, its a normal behavior IMO

pkg/components/ebpf/common/tcp_detect_transform.go
```diff
@@ -148,6 +120,118 @@ func ReadTCPRequestIntoSpan(parseCtx *EBPFParseContext, cfg *config.EBPFTracer,
...
+		stmt = string(requestBuffer[sqlprune.MySQLHdrSize+1:])
+
+		parseCtx.mysqlPreparedStatements.Add(mysqlPreparedStatementsKey{
+			connInfo: event.ConnInfo,
+			stmtID:   stmtID,
+		}, stmt)
+
+		return span, errIgnore
+	case "STMT_EXECUTE":
+		if sqlError != nil {
```
NimrodAvni78: if execute fails isn't this like if a query fails? and thats a usecase we dont want to ignore

mmat11: nice catch, copy paste gone wrong

mmat11:
Closing, will open multiple separate PRs

PR #258, opened 2025-07-04
Start consolidating integration tests and removing duplicates

mariomac:


test/integration/suites_test.go
```diff
@@ -177,21 +170,6 @@ func TestSuite_GRPCExportKProbes(t *testing.T) {
 	require.NoError(t, compose.Close())
 }
 
-// Same as Test suite, but searching the executable by port instead of executable name
-func TestSuite_OpenPort(t *testing.T) {
```
mariomac, line 181: OPEN_PORT is already used in many other tests. No need to run a specific test suite only to make sure open_port works

grcevski (approved):
Nice!

Issue #63, opened 2025-05-20
Support manual instrumentation


einepeople:
Support registering Beyla as global tracer provider, so that it can support manual instrumentation and link manually created spans with auto-generated traces.
Such functionality exists in `opentelemetry-go-instrumentation` since late 2023.

Without manual spans support every user request generates two completely disconnected traces - one with manually defined (e.g. business logic oriented) spans, and one with auto-generated spans (which only describes app's external communication). 

Reference: 
* Issue with description: https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/352
* MR with impl:  https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/523

P.S. there is a [similar issue](https://github.com/grafana/beyla/issues/1936) opened for `grafana/beyla`. I heard the community should now contribute to this repo and not to the `beyla`. If that's not the case and this is a duplicate issue, feel free to close it.

Issue #245, opened 2025-07-02
POC Probe implementation and machinery integration with opentelemetry-go-instrumentation ("Steel Thread")


damemi:
To further the design of a [Custom Probe API](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1105), we first need to understand the needs of projects that intend to adopt Probes.

### Current Overlap

The main overlap between OBI and [Go-Auto](https://github.com/open-telemetry/opentelemetry-go-instrumentation) is:

* **OBI `Tracer` = Go-Auto `Probe`**
  The Go-Auto `Probe` (the "Probe API") is conceptually equivalent to OBI’s `Tracer`.
  Currently, the Probe API is responsible for defining instrumentation libraries (e.g., uprobes, function symbols, offsets, and config options), managing their lifecycle (e.g., loading eBPF programs, maps), and handling emitted events. In Go-Auto, this is defined as a Go interface.

* **OBI `Instrumenter` = Go-Auto `Manager`**
  The `Manager` in Go-Auto plays a role similar to the OBI `Instrumenter`. It is the core orchestration component, instantiated via [`auto.NewInstrumentation()`](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/48361bc4d9b19f671081efffc4678e36fde3c404/instrumentation.go#L43), and manages probe lifecycles and runtime behavior.

**tl;dr**
`obi.Tracer` = `auto.Probe`
`obi.Instrumenter`  = `auto.Manager`

Our aim is to unify these overlapping concepts and remove redundant implementations across OBI and Go-Auto.

---

## Proposal

I propose refactoring the Go-Auto architecture to shift most lifecycle responsibilities from `Probe` to `Manager`. Under this model:

* **Probes become declarative:**
  The Probe API should focus on describing instrumentation (what to trace, where, and how) without dealing with lifecycle management.

* **Managers control execution:**
  The `Manager` becomes the sole runtime controller, responsible for initializing, loading, and managing instrumentations.

This change will simplify and stabilize the Probe API, allowing developers to define portable, reusable Probes. Any eBPF tool built with OpenTelemetry standards could import and use these Probes without needing to fork or customize core logic, provided they conform to the Probe API.

## Benefits of This Approach

1. **Reduced API Surface Area:** By stripping lifecycle logic from Probes, we make the API easier to maintain and less prone to breaking changes.
[...]

PR #260, opened 2025-07-04
Support for manual spans in Go

grcevski:
This PR adds initial support for adding manual spans in Go with various SDK approaches. Supports the simple tracer approach as well as the new Go Auto SDK as done in OTel Go Auto Instrumentation.

This PR vendored (manually, e.g. copy paste) the code from Go Auto Instrumentation until we figure out a plan to add the required features for OBI to be able to directly use Go Auto Instrumentation. Once we replace this with direct vendoring from OTel Go Auto, the tests should simply pass, ensuring we've done the vendoring correctly (https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/245).

There are couple of differences in how this is implemented compared to Go Auto's version, mainly related to two areas:
1. We don't set the Auto SDK boolean because it will require one more use of CAP_SYS_ADMIN (related to bpf_probe_write_user), which we are trying to phase out. I worked around this by probing directly into the Start and ReportError options and by using an `itab` pointer comparison. The type comparison was a bit tricky because `go.opentelemetry.io/otel/trace.attributeOption` is a composite type and loaded into an anonymous memory region on program load. However, matching on the internal type information worked.
2. We don't read the Tracer name and options because the existing code in Go Auto is using map read, which doesn't work for the latest version of Go, e.g. we haven't written support for parsing swiss maps. I see two approaches going forward
  a) Add swiss map support and use that
  b) Uprobe on `provider.Tracer` and use the approach for interface type matching to pull out the information.

Fixing 2. should be done in a separate PR, I think this is large enough as it stands.

TODO:
 - [x] Unit tests for the new itab code
 - [x] Integration tests for Go 1.24+ and latest Go OTel SDK, using both Auto and simple tracer.
 - [x] Integration tests for Go 1.17 (earliest supported in OBI) with old Go OTel SDK.

CC @MrAlias It might be easier to review this as a PR rather than my original branch. It this is sensible, maybe we can discuss if we want to port some of the ideas to Go Auto as well, e.g. support older SDKs?

Closes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/63


grcevski:
I modified slightly the example in Go Auto and tried a few things, this is what the outcome looks like:
```
var tracer = otel.Tracer("trace-example")

func inner(id int) {
	ctx := context.Background()
	ts := y2k.Add(10 * time.Microsecond)

	t := tracer

	opts := []trace.SpanStartOption{
		trace.WithAttributes(
			attribute.String("user", "user"+strconv.Itoa(id)),
			attribute.Bool("admin", true),
		),
		trace.WithTimestamp(y2k.Add(500 * time.Microsecond)),
		trace.WithSpanKind(trace.SpanKindServer),
	}

	_, span := t.Start(ctx, fmt.Sprintf("sig_inner %d", id), opts...)
	// span.SetAttributes(
	// 	attribute.String("user", "user"+strconv.Itoa(id)),
	// 	attribute.Bool("admin", true),
	// )

	if id == 2 {
		span.SetName("changed name")
	}
	defer span.End(trace.WithTimestamp(ts.Add(100 * time.Microsecond)))
}

func handler(w http.ResponseWriter, r *http.Request) {
	ctx := context.Background()
	ts := y2k.Add(10 * time.Microsecond)

	provider := sdk.TracerProvider()
	t := provider.Tracer(
		"main",
		trace.WithInstrumentationVersion("v0.0.1"),
		trace.WithSchemaURL("https://some_schema"),
	)

	_, span := t.Start(ctx, "sig", trace.WithTimestamp(ts))
	inner(1)
	inner(2)

	span.SetStatus(codes.Error, "application error")
	span.RecordError(
		fmt.Errorf("some unknown error"),
		trace.WithTimestamp(y2k.Add(2*time.Second)),
		trace.WithStackTrace(true),
		trace.WithAttributes(attribute.Int("impact", 11)),
	)

	defer span.End(trace.WithTimestamp(ts.Add(100 * time.Microsecond)))
}
```
In here I mix use of `sdk.TracerProvider()` and `tracer = otel.Tracer("trace-example")`. It will become easier to digest once I add tests.

image

Makefile
```diff
@@ -144,8 +144,8 @@ lint: prereqs
 
 .PHONY: update-offsets
 update-offsets: prereqs
-	@echo "### Updating pkg/internal/goexec/offsets.json"
-	$(GO_OFFSETS_TRACKER) -i configs/offsets/tracker_input.json pkg/internal/goexec/offsets.json
+	@echo "### Updating pkg/components/goexec/offsets.json"
```
rafaelroquetto, line 147: groovy!

bpf/common/common.h
```diff
@@ -146,3 +149,75 @@ typedef struct tcp_large_buffer {
...
+typedef struct go_otel_key_value {
+    struct go_string key;
+    go_otel_attr_value_t value;
+} go_otel_key_value_t;
+
+#define OTEL_ATTRIBUTE_KEY_MAX_LEN (32)
+#define OTEL_ATTRIBUTE_VALUE_MAX_LEN (128)
+#define OTEL_ATTRUBUTE_MAX_COUNT (16)
+
+typedef struct otel_attirbute {
```
rafaelroquetto, line 198:
```suggestion
typedef struct otel_attribute {
```

bpf/gotracer/go_nethttp.c
```diff
@@ -294,7 +294,7 @@ int beyla_uprobe_readContinuedLineSliceReturns(struct pt_regs *ctx) {
 }
 
 SEC("uprobe/ServeHTTP_ret")
-int beyla_uprobe_ServeHTTPReturns(struct pt_regs *ctx) {
+int serve_http_returns(struct pt_regs *ctx) {
```
rafaelroquetto, line 398: so, the `beyla_` prefix was intended to "namespace" the program when looking at it from `bpftool` - I'd keep it for now, I am planning to raise a PR renaming those to `obi_` all at once.

grcevski: Yeah I know I was making it consistent, we need one change to rename them all.

bpf/common/common.h
grcevski: :) it's vendored directly from Go Auto, I want to keep as little as possible changes.

PR #268, opened 2025-07-07
NodeJS correlation fixes

rafaelroquetto:
We noticed that under heavy load (i.e. hundreds of parallel requests per second), some traces context propagation would break - namely, because `fd_to_connection` map was being exhausted, and to a lesser extent, because there were no hard guarantees that the `ebpf_ipc` code would run in time, before the corresponding outgoing client call would start.

This PR addresses that by:

- increasing the `fd_to_connection` map size to a reasonable size
- replacing the entire `ebpf_ipc` mechanism with a more straightforward and synchronous approach, that relies on a `uprobe` to `uv_fs_access` to pass the socket correlation data to the ebpf layer.


bpf/maps/fd_to_connection.h
```diff
@@ -11,6 +11,6 @@ struct {
     __uint(type, BPF_MAP_TYPE_LRU_HASH);
     __type(key, fd_key); // key: fd
     __type(value, connection_info_t);
-    __uint(max_entries, 1024);
+    __uint(max_entries, 20000);
```
grcevski: Can we please use the constant MAX_SHARED_REQUESTS here? Or MAX_REQUESTS, depending which one is closer. I think Shared is 30K.

PR #274, opened 2025-07-08
chore: improve config feature detection error message

skl:
In the case that both network and application features are not enabled via config, this PR improves the error message by providing the user with examples.

### DIFF
--- a/pkg/beyla/config.go
+++ b/pkg/beyla/config.go
@@ -262,7 +262,11 @@ func (c *Config) Validate() error {
 	}
 
 	if !c.Enabled(FeatureNetO11y) && !c.Enabled(FeatureAppO11y) {
-		return ConfigError("missing to enable application discovery or network metrics. Check documentation")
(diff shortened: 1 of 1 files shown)


pkg/beyla/config.go
```diff
@@ -262,7 +262,11 @@ func (c *Config) Validate() error {
 	}
 
 	if !c.Enabled(FeatureNetO11y) && !c.Enabled(FeatureAppO11y) {
-		return ConfigError("missing to enable application discovery or network metrics. Check documentation")
+		return ConfigError("at least one of 'network' or 'application' features must be enabled. " +
+			"Enable Otel export features using the 'OTEL_EBPF_OTEL_METRIC_FEATURES=network,application' environment variable " +
```
grcevski: small nit, we should write `Otel` as `OTel` or use the full word, `OpenTelemetry`.

grcevski (approved):

pkg/beyla/config.go
skl: np, fixed in 00b7acd7

PR #276, opened 2025-07-08
WIP: Introduce granular (service selector) export modes

rafaelroquetto:
Introduce a per service-selector export mode. This allows individual selectors to specify whether they want to:

- export metrics only (`ExportMetrics`)
- export traces only (`ExportTraces`)
- export both metrics and traces (`ExportMetricsAndTraces`)
- export nothing (`ExportDisabled`)
- use the default behaviour (`ExportUnset`)

The selection criteria are applied in order of appearance. For example, consider the following sample configuration for discovering a service called `planet-service` which lives in the `default` namespace:
```
discovery:
  services:
    - k8s_namespace: default              #(1)
      export: metrics
    - k8s_deployment_name: planet-service #(2)
      export: traces
```

Because selectors are processed in sequential order of appearance (selector _2_ appears after selector _1_), the resulting configuration will only enable _traces_ export for `planet-service`. Whilst this PR only adds a new `export` attribute, this paves the road for future attributes to be added and allows for configurations to be stacked. For instance, we can decide to export metrics and traces for the entire default namespace, except for `planet-service`, which exports metrics only:

```
discovery:
  services:
    - k8s_namespace: default              #(1)
      export: metrics_and_traces
    - k8s_deployment_name: planet-service #(2)
      export: metrics
```

Omitting the `export` attribute falls back to the conventional behaviour that relies on variables such as `OTEL_EXPORTER_OTLP_ENDPOINT`, i.e. they are orthogonal. For instance, if a service selector specifies `export: traces` but no trace endpoint is defined, then obviously no traces will be generated.

*This PR is best reviewed on a per-commit basis*

TODO

- [ ] finish hooking other exporters
- [ ] tests


pkg/services/attr_regex.go
```diff
@@ -76,6 +76,10 @@ type RegexSelector struct {
 
 	// Restrict the discovery to processes which are running inside a container
 	ContainersOnly bool `yaml:"containers_only"`
+
+	// Configures what to export. Allowed values are 'metrics', 'traces',
+	// 'metrics_and_traces' and 'disabled'
+	ExportMode ExportMode `yaml:"exports"`
```
mariomac: I would make this an array with `metrics`, `traces` or empty for disabled (despite probably there is no reason to add a service selector to disable any export).

If nothing is defined, it would default to `exports: ["metrics", "traces"]`, and to disable it, an explicit empty list should be set: `exports: []`.

pkg/services/criteria.go
```diff
@@ -113,6 +113,72 @@ func (c *DiscoveryConfig) Validate() error {
...
+
+	for i, v := range textForMode {
+		if v == s {
+			*e = ExportMode(i)
+			return nil
+		}
+	}
+
+	return fmt.Errorf("invalid value for export_mode: %q", s)
+}
```
mariomac: Maybe the `maps.MappedBits` tool is helpful here:

```go
	const (
		Metrics = 1 << iota
		Traces
	)
	stringValues := map[string]Bits{
		"metrics": Metrics,
		"traces":  Traces,
	}
	userSelected := []string{"metrics"}
	canExport := MappedBits(userSelected, stringValues)

	assert.True(t, canExport.Has(Metrics))
	assert.False(t, canExport.Has(Traces))
```

pkg/services/attr_regex.go
rafaelroquetto: I like the idea of the array, but would that allow me to distinguish between "disabled" vs "default"? Default means "inherit whatever criteria has been previously matched", whereas disabled literally means "none of those, turn it off" (for the attribute in question only, not for the entire selector).

> If nothing is defined, it would default to exports: ["metrics", "traces"]

That unfortunately does not work for what I am trying to accomplish. We need to be able to inherit the previously matched criteria. In the future, there will be more settings beyond `export`, such as `sampler`, etc...

So imagine this (assume all services in the `default` namespace):

```
discovery:
  services:
    - k8s_namespace: default
      export: metrics
      sampler: 99.99
    - k8s_deployment_name: jupiter-service  #inherits sampler = 99.99 but exports only traces
      export: traces
     - k8s_deployment_name: saturn-service  #overrides sampler but exports only metrics
      sampler: 88.88
      - k8s_deployment_name: uranus-service  #disable metrics export, but enable debug
      export: disabled
      debug: true
```

If we defaulted the empty array to `[metrics, traces]`, `saturn-service` would break, as the semantics we are trying to implement would expect it to inherit `metrics` from the default namespace selector.

Likewise, perhaps we want a service to have debug enabled for uranus-service (this doesn't exist, it's just for illustration purposes) but not export any metrics - we can't simply put it in the exclude list.

In a nutshell, this tries to implement a very  flexible mechanism (and not policy) to allow selectors to override one another based on ordering. This PR started with `exports` but there will be more options.

pkg/services/criteria.go
rafaelroquetto: Nice - I will give it a shot. I went out to check the docs and they reminded me of someone, I just can't remember who exactly ;)

pkg/services/attr_regex.go
mariomac: I guess not defining anything will leave the value as `nil` so you can assume it as undefined (default), while an explicit empty array would mean "disabled".

Despite it might look cumbersome, it is how usually defaults are explicitly emptied in YAML.

The reason I'm in favor of an array is because it looks more aligned with other similar configuration options, while I'm also thinking in the future, when we might want to add logs support to Beyla. What would be the accepted values here? `metrics`, `traces`, `metrics_and_traces`, `traces_and_metrics`, `metrics_and_logs`, `logs_and_metrics`, `traces_and_logs`, `metrics_and_traces_and_logs`...

rafaelroquetto: Ah that's great then, I did not know.  The array will then look much better then so thanks for this, I will implement the changes (and let you know if I run into trouble with them) - will just finish fixing the tests first and will go back and switch to the array

pkg/services/criteria.go
rafaelroquetto: @mariomac I've updated the code and the unit tests (and a single k8s test) to reflect the array usage - _it's so much better_, so thanks for the suggestions.

I've implemented it in the most obvious way - next, I will see if I can shoehorn `maps.MappedBits` or some sort of flags - I haven't looked into it yet, but just in case, do you see a way forward with `maps.MappedBits` that would allow me to represent the `nil` case (i.e. unset)? And would that mean that the `yaml` / config struct type needs to be a `[]string` rather than `ExportModes`?

rafaelroquetto: As discussed offline, will keep the current implementation to leverage the automatic deserialisation from YAML

pkg/components/discover/matcher.go
```diff
@@ -104,51 +104,80 @@ func (m *Matcher) filter(events []Event[ProcessAttrs]) []Event[ProcessMatch] {
...
-			return Event[ProcessMatch]{
-				Type: EventCreated,
-				Obj:  ProcessMatch{Criteria: m.Criteria[i], Process: proc},
-			}, true
-		}
+
+	processMatch := m.matchCriteria(obj, proc)
+
+	if processMatch != nil {
+		m.ProcessHistory[obj.pid] = *processMatch
```
mariomac: Nit: I'd restrict the `processMatch` lifespan, to avoid being alive together with `procMatch` below:
```go
if processMatch := m.matchCriteria(obj, proc); processMatch != nil {
```

rafaelroquetto: nice catch, will fix

rafaelroquetto:
@mariomac should be good to go!

PR #279, opened 2025-07-09
change opentelemetry.sdk.name from beyla to opentelemetry-ebpf-instrumentation

### LINKED ISSUE #261: change telemetry.sdk.name from beyla to obi

NimrodAvni78:
Resolves #261


marctc:
Doesn't make sense here to have the full namw?

grcevski:
> Doesn't make sense here to have the full name?

Oh that's a good point, why don't we discuss this at the SIG call? Maybe it's better if we have otel-ebpf-instrumentation in the attribute. I think `obi` is great for the binary.

PR #288, opened 2025-07-10
Fix panic due invalid UTF-8 labels

marctc:
Add sanitization step in Prometheus metric generation to avoid make OBI panic when a label with invalid UTF8 is present. 
Fixes https://github.com/grafana/beyla/issues/1867

### DIFF
--- a/pkg/export/prom/prom.go
+++ b/pkg/export/prom/prom.go
@@ -8,6 +8,7 @@ import (
 	"slices"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"github.com/prometheus/client_golang/prometheus"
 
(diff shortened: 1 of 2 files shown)


mariomac:
Looking at the related issue: it seems the panic is due to trying to pass a clearly invalid value to the prom label: 

`deb.debian.or 1498318199  0     0     100644  828       `\n\x1f\x8b\b`

Maybe we should fix the fact that we are passing a wrong value instead of just silently ignore the bug.

marctc:
I thought that users might want to review what's being instrumented and which faulty labels are trying to ingest instead of sanitizing. But I can totally change it to try sanize instead of detecting.

mariomac:
@marctc I'm afraid that, if we just log errors but don't panic, most users won't realize the error message until they maybe realize that some metrics/traces are missing.

mariomac (approved):
👏🏻

Issue #282, opened 2025-07-09
Audit for the `Beyla` name


MrAlias:
Ensure we do not have the name at all. Or, at least not public facing while we update internals.

MrAlias:
All external attributes, names, and metrics are all renamed. Only internal symbols remain.

PR #289, opened 2025-07-10
Rename pkg/beyla to pkg/obi

mariomac:
Partially addresses: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/282


marctc (approved):
:broom:

Issue #285, opened 2025-07-09
Running random TLS traffic causes MySQL warning messages


grcevski:
While testing a simple HTTPS client program in Java (which we cannot handle correctly) I noticed that (perhaps not surprising) that the MySQL code tries to handle this request. The issue is that when we have libssl request, those are TCP requests are being ignored (as much as we can) from the regular paths, since they are identified as encrypted traffic which we cannot parse. 

Java doesn't use libssl and I've been trying to write a small agent to help us there. So while doing this, I wrote a simple Java program that makes an outgoing HTTPS request in a loop and I noticed that when instrumented with OBI we now see the following log messages:
```
time=2025-07-09T17:23:39.310-04:00 level=WARN msg="MySQL command ID unhandled" commandID=203
time=2025-07-09T17:23:39.312-04:00 level=WARN msg="MySQL command ID unhandled" commandID=1
time=2025-07-09T17:23:39.358-04:00 level=WARN msg="MySQL command ID unhandled" commandID=85
time=2025-07-09T17:23:39.397-04:00 level=WARN msg="MySQL command ID unhandled" commandID=199
```

I think these probably shouldn't be WARNs.

Here's my simple program:
```
import org.apache.http.HttpResponse;
import org.apache.http.client.HttpClient;
import org.apache.http.client.methods.HttpGet;
import org.apache.http.impl.client.HttpClients;

import java.io.BufferedReader;
import java.io.IOException;
import java.io.InputStreamReader;

public class ApacheHttpsExample {
    private static void run() throws IOException {
        String url = "https://www.grafana.com";
        HttpClient httpClient = HttpClients.createDefault();

        HttpGet httpGet = new HttpGet(url);

        HttpResponse response = httpClient.execute(httpGet);

        BufferedReader reader = new BufferedReader(new InputStreamReader(response.getEntity().getContent()));
        StringBuilder result = new StringBuilder();
        String line;

        while ((line = reader.readLine()) != null) {
            result.append(line);
        }

        System.out.println(result.toString().length());
    }
    public static void main(String[] args) throws IOException, InterruptedException {
        while (true) {
            run();
            Thread.sleep(2_000);
        }
    }
}
```

I'm running OBI as:
```
sudo OTEL_EBPF_EXECUTABLE_PATH=java OTEL_EBPF_TRACE_PRINTER=text OTEL_EBPF_BPF_DEBUG=1 bin/ebpf-instrument
```

[...]

PR #290, opened 2025-07-10
bpf: correctly initialize classification protocol

mmat11:
Resolves https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/285

/cc @grcevski looks like the issue was the protocol type initialization, I thought it would be always initialized to zero and cached for the connection lifetime but that seems not to be the case. I noticed `make_protocol_args()` initializes the protocol type value to 1 without the mysql classifier code doing anything (and doesn't keep it for the connection lifetime). 

```
time=2025-07-10T14:15:47.383+02:00 level=DEBUG msg="GREPME protocol_type 0" component=BPFLogger pid=192519 comm=Finalizer
time=2025-07-10T14:15:47.383+02:00 level=DEBUG msg="is_mysql: unhandled mysql command_id: 35" component=BPFLogger pid=192519 comm=Finalizer
...snip...
time=2025-07-10T14:15:47.391+02:00 level=DEBUG msg="GREPME protocol_type 1" component=BPFLogger pid=192519 comm=ApacheHttpsExam
```

I changed it to always initialize it to `unknown`; We should probably change it (now or in the future) to keep it cached for the connection lifetime, wdyt?
I also modified the classification code as the `execute` command check is too shallow and could misclassify packets too easily.
I tested it with the java snippet and it seems to fix the issue

### DIFF
--- a/bpf/generictracer/k_tracer_defs.h
+++ b/bpf/generictracer/k_tracer_defs.h
@@ -36,6 +36,7 @@ make_protocol_args(void *u_buf, int bytes_len, u8 ssl, u8 direction, u16 orig_dp
     args->direction = direction;
     args->orig_dport = orig_dport;
     args->u_buf = (u64)u_buf;
+    args->protocol_type = k_protocol_type_unknown;
 
     return args;
 }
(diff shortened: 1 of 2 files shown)


grcevski (approved):
Excellent! Thanks for the quick turnaround @mmat11! This looks good. I'll give it a try myself as soon as this is merged.

rafaelroquetto (approved):

PR #316, opened 2025-07-17
OATS: replace maven by maven wrapper

mariomac:
Some OATS dockerfiles manually install maven, and this fails from time to time because the URL changes.

We are changing to Maven wrapper to hide the installation complexity and ensure stability in the mid term.


grcevski (approved):
Thanks! Much better fix.

PR #318, opened 2025-07-17
bpf: implement large buffer action, improve mysql protocol inference

mmat11:
- Add large buffer `action`: init/append
- Remove mysql tail call program: it was essentially useless
- Add more sanity checks and better header parsing based on payload lengths
- Add large buffers userspace rwlock, add tests for different actions

### DIFF
--- a/bpf/generictracer/k_tracer.c
+++ b/bpf/generictracer/k_tracer.c
@@ -945,18 +945,18 @@ int obi_handle_buf_with_args(void *ctx) {
         if (has_preface(args->small_buf, args->bytes_len) && args->bytes_len > MIN_HTTP2_SIZE) {
             args->u_buf = args->u_buf + MIN_HTTP2_SIZE;
         }
-    }
-
-    http2_conn_info_data_t *h2g = bpf_map_lookup_elem(&ongoing_http2_connections, &args->pid_conn);
-    if (h2g && (http2_flag_ssl(h2g->flags) == args->ssl)) {
-        bpf_tail_call(ctx, &jump_table, k_tail_protocol_http2);
     } else if (is_mysql(&args->pid_conn.conn,
                         (const unsigned char *)args->u_buf,
                         args->bytes_len,
                         &args->packet_type,
                         &args->protocol_type)) {
         bpf_dbg_printk("Found mysql connection");
-        bpf_tail_call(ctx, &jump_table, k_tail_protocol_mysql);
+        bpf_tail_call(ctx, &jump_table, k_tail_protocol_tcp);
+    }
+
+    http2_conn_info_data_t *h2g = bpf_map_lookup_elem(&ongoing_http2_connections, &args->pid_conn);
+    if (h2g && (http2_flag_ssl(h2g->flags) == args->ssl)) {
+        bpf_tail_call(ctx, &jump_table, k_tail_protocol_http2);
     } else { // large request tracking and generic TCP
         http_info_t *info = bpf_map_lookup_elem(&ongoing_http, &args->pid_conn);
 
--- a/bpf/generictracer/k_tracer.c
+++ b/bpf/generictracer/k_tracer.c
@@ -945,6 +945,11 @@ int obi_handle_buf_with_args(void *ctx) {
         if (has_preface(args->small_buf, args->bytes_len) && args->bytes_len > MIN_HTTP2_SIZE) {
             args->u_buf = args->u_buf + MIN_HTTP2_SIZE;
         }
+    }
+
+    http2_conn_info_data_t *h2g = bpf_map_lookup_elem(&ongoing_http2_connections, &args->pid_conn);
+    if (h2g && (http2_flag_ssl(h2g->flags) == args->ssl)) {
+        bpf_tail_call(ctx, &jump_table, k_tail_protocol_http2);
     } else if (is_mysql(&args->pid_conn.conn,
                         (const unsigned char *)args->u_buf,
                         args->bytes_len,
@@ -952,11 +957,6 @@ int obi_handle_buf_with_args(void *ctx) {
                         &args->protocol_type)) {
         bpf_dbg_printk("Found mysql connection");
         bpf_tail_call(ctx, &jump_table, k_tail_protocol_tcp);
-    }
-
-    http2_conn_info_data_t *h2g = bpf_map_lookup_elem(&ongoing_http2_connections, &args->pid_conn);
-    if (h2g && (http2_flag_ssl(h2g->flags) == args->ssl)) {
-        bpf_tail_call(ctx, &jump_table, k_tail_protocol_http2);
     } else { // large request tracking and generic TCP
         http_info_t *info = bpf_map_lookup_elem(&ongoing_http, &args->pid_conn);
 
--- a/bpf/generictracer/protocol_mysql.h
+++ b/bpf/generictracer/protocol_mysql.h
@@ -209,17 +209,12 @@ static __always_inline u8 is_mysql(connection_info_t *conn_info,
         return 0;
     }
 
-    if (data_len < (k_mysql_hdr_size + 1)) {
-        bpf_dbg_printk("is_mysql: data_len is too short: %d", data_len);
-        return 0;
-    }
-
     struct mysql_hdr hdr = {};
     if (mysql_parse_fixup_header(conn_info, &hdr, data, data_len) != 0) {
         bpf_dbg_printk("is_mysql: failed to parse mysql header");
         return 0;
     }
-    u32 payload_len = mysql_payload_length(hdr.payload_length);
+    const u32 payload_len = mysql_payload_length(hdr.payload_length);
 
     if (payload_len > k_mysql_payload_length_max) {
         bpf_dbg_printk("is_mysql: payload length is too large: %d", payload_len);
--- a/bpf/generictracer/protocol_mysql.h
+++ b/bpf/generictracer/protocol_mysql.h
@@ -56,6 +56,9 @@ enum {
     // Large buffer
     k_large_buf_max_size = 1 << 14, // 16K
     k_large_buf_max_size_mask = k_large_buf_max_size - 1,
+
+    // Sanity checks
+    k_mysql_payload_length_max = 1 << 13, // 8K
 };
 
 struct {
@@ -67,6 +70,12 @@ struct {
 
 SCRATCH_MEM_SIZED(mysql_large_buffers, k_large_buf_max_size);
 
+// This function is used to store the MySQL header if it comes in split packets
+// from double send.
+// Given the fact that we need to store this for the duration of the full request
+// (split in potentially multiple packets), we will **not** process or preserve
+// any actual payloads that are exactly 4 bytes long — they are intentionally
+// dropped in favor of state storage.
 static __always_inline int mysql_store_state_data(const connection_info_t *conn_info,
                                                   const unsigned char *data,
                                                   size_t data_len) {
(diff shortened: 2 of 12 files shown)


bpf/generictracer/k_tracer.c
```diff
@@ -943,18 +943,18 @@ int obi_handle_buf_with_args(void *ctx) {
...
                         (const unsigned char *)args->u_buf,
                         args->bytes_len,
                         &args->packet_type,
                         &args->protocol_type)) {
         bpf_dbg_printk("Found mysql connection");
-        bpf_tail_call(ctx, &jump_table, k_tail_protocol_mysql);
+        bpf_tail_call(ctx, &jump_table, k_tail_protocol_tcp);
+    }
+
+    http2_conn_info_data_t *h2g = bpf_map_lookup_elem(&ongoing_http2_connections, &args->pid_conn);
```
rafaelroquetto: I could be wrong here (@grcevski can probably confirm or deny), but as far as I remember, we need to check for ongoing http2 connections before mysql to prevent some random buffer from being miscategorised - I think this was recently introduced due to some obscure SSL bugs.

bpf/generictracer/protocol_mysql.h
```diff
@@ -232,9 +217,15 @@ static __always_inline u8 is_mysql(connection_info_t *conn_info,
         bpf_dbg_printk("is_mysql: failed to parse mysql header");
         return 0;
     }
+    u32 payload_len = mysql_payload_length(hdr.payload_length);
```
rafaelroquetto:
```suggestion
    const u32 payload_len = mysql_payload_length(hdr.payload_length);
```

pkg/components/ebpf/common/tcp_large_buffer.go
```diff
@@ -19,6 +19,9 @@ type (
 )
 
 func setTCPLargeBuffer(parseCtx *EBPFParseContext, record *ringbuf.Record) (request.Span, bool, error) {
+	parseCtx.largeBuffersMU.Lock()
+	defer parseCtx.largeBuffersMU.Unlock()
+
 	hdrSize := uint32(unsafe.Sizeof(TCPLargeBufferHeader{}))
 
 	event, err := ReinterpretCast[TCPLargeBufferHeader](record.RawSample[:hdrSize])
```
rafaelroquetto: Revisiting this, we don't really need [:hdrSize] here

pkg/components/ebpf/common/tcp_large_buffer.go
```diff
@@ -19,6 +19,9 @@ type (
 )
 
 func setTCPLargeBuffer(parseCtx *EBPFParseContext, record *ringbuf.Record) (request.Span, bool, error) {
```
rafaelroquetto: Since this now unconditionally appends a buffer to `key`, I'd rename this to `appendTCPLargeBuffer`

pkg/components/ebpf/common/tcp_large_buffer.go
```diff
@@ -27,14 +30,27 @@ func setTCPLargeBuffer(parseCtx *EBPFParseContext, record *ringbuf.Record) (requ
...
 		traceID:   event.Tp.TraceId,
 		spanID:    event.Tp.SpanId,
 		direction: event.Direction,
 	}
 
-	copiedBuffer := make([]byte, event.Len)
+	switch event.Action {
+	case 1: // LargeBufActionAppend
+		// If this is an append action, we need to check if the buffer already exists
+		if lb, ok := parseCtx.largeBuffers.Get(key); ok {
```
rafaelroquetto: This can be very inefficient, specially at this codepath, for a few reasons:

1. we incur a hash lookup
2. a buffer realloc when appending newBuffer to `lb.buf`
3. a dealloc / hash removal
4. once the switch exits,  we do another alloc (`copiedBuffer`)
5. and yet another copy (newBuffer -> copiedBuffer)
6. and then we readd it to to `parseCtx.largeBuffers`
7. all of which whilst holding a mutex - which puts considerable pressure on the ring buffer, since this function will ultimately block the ring buffer read loop until it returns

Instead, we should lock and release the mutex ASAP and keep the number of allocs/copies to the bare minimum required.

The proper way to go about this I'm afraid will require a little bit more refactoring - we need to change `largeBuffer` to be an array of buffers to which we can simply append a new buffer to, and bail, i.e. the canonical "scatter/gather" pattern - analogous to Linux / [POSIX iovecs](https://man7.org/linux/man-pages/man2/readv.2.html) - but simpler, as we are just appending.

So basically something like:

```
parseCtx.largeBuffersMU.Lock()
largeBuffer := getLargeBuffer(parseCtx.largeBuffers, key) // returns a pointer to a large buffer, creating it if it does not exist
largeBuffer.append(newBuffer)
parseCtx.largeBuffersMU.Unlock()
```
for the existing `getTCPLargeBuffer`, whose semantics is to simply return a large buffer if it exists, removing it from the cache, we can rename it to a more appropriate name: `extractTCPLargeBuffer`

usage:

```
parseCtx.largeBuffersMRU.Lock()
largeBuffer := extractLargeBuffer(parseCtx.largeBuffers, key)
parseCtx.largeBuffersMRU.Unlock()

if largeBuffer != nil { ... } // or simply return an empty one
```

because `struct largeBuffer` now carries more than one buffer, to facilitate reading it we can just implement a proper API on top of it - for instance, by having a method `.Bytes(offset, length)` that returns a contiguous byte slice on that range and a convenience method:

```
func (ln *largeBuffer) ContiguousBuffer() {
   return lb.Bytes(0, lb.Size())
}
```

pkg/components/ebpf/common/tcp_large_buffer.go
```diff
@@ -44,6 +60,9 @@ func setTCPLargeBuffer(parseCtx *EBPFParseContext, record *ringbuf.Record) (requ
 }
 
 func getTCPLargeBuffer(parseCtx *EBPFParseContext, traceID [16]uint8, spanID [8]uint8, direction uint8) ([]byte, bool) {
+	parseCtx.largeBuffersMU.RLock()
```
rafaelroquetto: out of curiosity though, where does the concurrency requiring the mutex stem from?

pkg/components/ebpf/generictracer/generictracer.go
```diff
@@ -147,45 +147,17 @@ func (p *Tracer) Load() (*ebpf.CollectionSpec, error) {
...
-		},
-		{
-			index: 6,
-			prog:  p.bpfObjects.ObiProtocolMysql,
-		},
-		{
-			index: 7,
-			prog:  p.bpfObjects.ObiHandleBufWithArgs,
-		},
+	for i, prog := range []*ebpf.Program{
```
rafaelroquetto, line 150: nice!

pkg/config/ebpf_tracer.go
```diff
@@ -92,7 +92,7 @@ type EBPFTracer struct {
 	RedisDBCache RedisDBCacheConfig `yaml:"redis_db_cache"`
 
 	// Limit max data buffer size per protocol.
-	BufferSizes EBPFBufferSizes `yaml:"buffer_sizes" env:"OTEL_EBPF_BPF_BUFFER_SIZES"`
+	BufferSizes EBPFBufferSizes `yaml:"buffer_sizes"`
```
rafaelroquetto, line 87: non-rhetorical question: what prompted the removal of the env var?

rafaelroquetto (changes requested):

pkg/config/ebpf_tracer.go
mmat11: this is a struct containing other fields, the inner fields' env vars can be used to set it; tldr it was unused

pkg/components/ebpf/common/tcp_large_buffer.go
mmat11: I added those while debugging for peace of mind; given the fact all the code after the ringbuffer read is synchronous, I don't think we need it since we are going to read/write always at one time (correct me if I'm wrong), so I can probably remove the mutex directly

pkg/config/ebpf_tracer.go
rafaelroquetto: ah got it! thanks for the clarifications :)

bpf/generictracer/k_tracer.c
mmat11: there's an additional check `else if (is_http2_or_grpc` on line 931, after the rebase I thought it would make more sense to check for different protocols in the first `if` block, I can change it if it's incorrect here (not sure it belongs in the second one.. maybe put an additional block at the end?)

pkg/components/ebpf/common/tcp_large_buffer.go
mmat11: I removed the locks, as written in the other discussion, as long as we run the handlers synchronously there's no need for them. As for hash lookups and buffer copies, I'm afraid we can't get rid of them, if I try to use the slice as it comes from ringbuf.Record, I see wrong data at extraction time (I suppose it gets wrongfully collected?)

Anyway, I don't understand what an array of buffers would accomplish as the only operation we wouldn't be using is `append(...)` (which reallocates only when necessary) and we would still need to copy the buffer

rafaelroquetto: I see what you mean. I think the garbage that you are seeing comes from the fact that we do this: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/1416e1f99840dc7ee8f205700401073b68272885/pkg/components/ebpf/common/ringbuf.go#L157

i.e. the `record` buffers will be overwritten the next time we read from the ring buffer into it (unfortunately the cilium/ebpf Go API does not give us access to the actual ring buffer buffer, so this is already a copy in itself).

So yeah, you are right - in this case we will need to copy anyway so we might as well just do the `append` as you are doing. Sorry, I completely missed that.

The only thing I think we should still do (and correct me if I am wrong), is to avoid `copiedBuffer` in the `LargeBufActionAppend` case - could we simply do:

```
lb.buf = append(lb.buf, newBuffer...)
```
since `buf` is actually implicitly shared and bail? (thus avoiding the second copy altogether as well as the need for newLen)?

bpf/generictracer/k_tracer.c
grcevski: If we've detected it's HTTP2 we need to proceed with that, not check mysql, unless we can guarantee that mySQL detection will never trigger on binary HTTP2 frame, whcih I think it's not possible.

What's the reasoning behind moving MySQL above the map lookup for http2? Did you experience a problem with mysql detection?

grcevski: Btw, the mysql detection message is still present, but less frequent now:

```
time=2025-07-18T12:45:10.022-04:00 level=WARN msg="MySQL command ID unhandled" commandID=23
```

I can reproduce it by just instrumenting all executables on my system with some random k8s kind cluster, like this:
```
 sudo OTEL_EBPF_EXECUTABLE_PATH=. OTEL_EBPF_TRACE_PRINTER=text OTEL_EBPF_BPF_DEBUG=1 bin/ebpf-instrument

```

pkg/components/ebpf/common/tcp_large_buffer.go
mmat11: > The only thing I think we should still do (and correct me if I am wrong), is to avoid copiedBuffer in the LargeBufActionAppend case - could we simply do:

I tried but I believe we still need to copy the new part of the buffer

```
GREPME SET (append) ---> [9 0 0 0 3 82 79 76 76 66 65 67 75]
GREPME SET (init) ---> [7 0 0 1 0 0 0 0 0 0 0]
GREPME GET <--- [131 114 58 65 14 149 53 0 0 0 0 0 0]
GREPME GET <--- [7 0 0 1 0 0 0 0 0 0 0]
```

I changed it to only copy the newer portion of the buffer

bpf/generictracer/k_tracer.c
mmat11: > What's the reasoning behind moving MySQL above the map lookup for http2? Did you experience a problem with mysql detection?

No reasoning, I just rebased and thought it would be its place, I moved it back down

> Btw, the mysql detection message is still present, but less frequent now:

I think I got the first fix wrong 🤦🏻‍♂️ , I was missing a return in the switch case, fixed here https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/318/commits/fe43e38187ea14c66499f9ffb3ce406a17327f27

pkg/components/ebpf/common/tcp_large_buffer.go
rafaelroquetto: My wild guess is that:

```
lb.buf = append(lb.buf, newBuffer...)
```

will break if `buf` is reallocated - because then the local `lb.buf` points to the right slice, but the one inside the LRU map still points to the original `lb.buf` - can you try changing the map so that it stores a pointer to `largeBuffer`, i.e.

```
expirable.LRU[largeBufferKey, *largeBuffer]
```
?

I am hoping then we can do `lb.buf = append(lb.buf, record.RawSample[hdrSize:]...)` and avoid the key reinsertion and remove the intermediate `newBuffer` copy.

mmat11: > My wild guess is that:
> lb.buf = append(lb.buf, newBuffer...)
> will break if buf is reallocated - because then the local lb.buf points to the right slice, but the one inside the LRU map still points to the original lb.buf

mmm I don't think this can happen, I'm replacing the map value in any case since I'm assigning `lb.buf` to `newBuffer`

rafaelroquetto: Well, there's something off here nonetheless - unless I am missing something obvious, `lb.buf = append(lb.buf, record.RawSample[hdrSize:]...)` should in theory work. If not, we need to understand why that is. My suggestion remains to changing the LRU to carry a pointer, so we can do:

```
lb := getLargeBuffer(key) // this returns a pointer, creating one if key doesn't exist on the LRU map
lb.buf = append(lb.buf, record.RawSample[hdrSize:]...) // 2
```

so that line `//2` will copy-allocate only once

mmat11: done

pkg/components/ebpf/common/tcp_large_buffer.go
```diff
@@ -18,32 +18,36 @@ type (
...
 		direction: event.Direction,
 	}
 
-	copiedBuffer := make([]byte, event.Len)
-	copy(copiedBuffer, newBuffer)
-	parseCtx.largeBuffers.Add(key, largeBuffer{
-		buf: copiedBuffer,
-	})
+	lb, ok := parseCtx.largeBuffers.Get(key)
+	if ok && event.Action == 1 {
```
rafaelroquetto, line 36: small nit: we should probably define a constant/enum for `LaegeBufActionAppend` and the rest.

Also, I'd have an explicitly else-if for LargeBufActionInit  and a catch all else that fails - just in case the constants go out of sync with the ebpf side - then we will see an error message during runtime, rather than the buffer silently being replaced when it shouldn't

rafaelroquetto (approved):
Awesome, thank you so much for your changes. I've left a comment at your discretion.

pkg/components/ebpf/common/tcp_large_buffer.go
mmat11: sounds good, do you mind if I do these 2 changes in the next PR? so we can avoid a CI run

rafaelroquetto: Sure, no problem!

grcevski (approved):
Great!

PR #321, opened 2025-07-17
Replace module name with `go.opentelemetry.io/obi`

MrAlias:
https://github.com/open-telemetry/opentelemetry-go-vanityurls/pull/55


mariomac:
@MrAlias To fix the _Check third-party licenses_ check, you just have to  run:
```
make update-licenses       
```
And push the changes in `third_party_licenses.csv`.

BTW: this file was a legal requirement in Beyla bundles. We left this file and check here because "it won't hurt". If you think it's unnecessary and/or annoying, we can remove the file and the related check.

MrAlias:
Thanks for the help. I think I've updated successfully, but it might be problematic given `go.opentelemetry.io/obi` may need a commit before a version can be found.

PR #323, opened 2025-07-17
Markdown lint

MrAlias:
- Add a project configuration for markdown linting (taken from go.opentelemetry.io/otel)
- Fix linting

Follow-up will be to add automated testing and lint fixing in the CI system and Makefile.


grcevski (approved):
Nice!

PR #330, opened 2025-07-18
Unify all module Go ver to 1.24

MrAlias:
We require the latest version of Go in our CONTRIBUTING.md. This unifies our modules to the latest major release of Go and fixes `go mod tidy` issues.


test/integration/components/testserver_1.17/go.mod
```diff
@@ -1,11 +1,13 @@
 module github.com/open-telemetry/opentelemetry-ebpf-instrumentation/testserver_1.17
 
-go 1.17
+go 1.24
```
grcevski: We need to keep this one at 1.17, otherwise the alternative is to completely remove the test. This was added to ensure we still work with the oldest version of Go we technically support. We've broken this in the past so the test was added.

grcevski (commented):
Looks good except this one Dockerfile which I think should stay at 1.17, it was done on purpose.

rafaelroquetto (commented):
non-rhetorical question: should we be using the latest patch release of 1.24.x rather than going with 1.24.0?

rafaelroquetto (approved):
Thanks!

PR #331, opened 2025-07-18
Do not fail cleanup when empty

MrAlias:
If there are no docker containers or images the `docker rm` and `docker rmi` commands become syntactically invalid. Only run the commands if there are containers or images to remove.

For example: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/16375613470/job/46278493579?pr=330


grcevski (approved):
Nice, thanks for fixing this. I've hit it myself :).

PR #332, opened 2025-07-18
Close the elf file when attacher is done with it

grcevski:
We never closed the ELF file opened during binary file inspection, which can cause for a large number of files to remain opened when OBI is running. This can especially become a problem for short lived binaries that we keep on detecting, instrumenting etc.

### DIFF
--- a/pkg/components/discover/attacher.go
+++ b/pkg/components/discover/attacher.go
@@ -118,7 +118,7 @@ func (ta *TraceAttacher) attacherLoop(_ context.Context) (swarm.RunFunc, error) 
 							}
 
 							if instr.Obj.FileInfo.ELF != nil {
-								instr.Obj.FileInfo.ELF.Close()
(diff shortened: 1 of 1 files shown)


pkg/components/discover/attacher.go
```diff
@@ -116,6 +116,10 @@ func (ta *TraceAttacher) attacherLoop(_ context.Context) (swarm.RunFunc, error)
 							if ok := ta.getTracer(&instr.Obj); ok {
 								ta.OutputTracerEvents.Send(Event[*ebpf.Instrumentable]{Type: EventCreated, Obj: &instr.Obj})
 							}
+
+							if instr.Obj.FileInfo.ELF != nil {
+								instr.Obj.FileInfo.ELF.Close()
```
MrAlias: Makes it clear we are ignoring the returned error.


```suggestion
								_ = instr.Obj.FileInfo.ELF.Close()
```

mariomac (approved):

PR #333, opened 2025-07-18
Update tool management

MrAlias:
Allow renovatebot to manage tool versioning. This ensures the repository does not contain a dependency on a vulnerable releases of code.

Additionally:

- Upgrade golangci-lint to `v2`
  - Fix linting associated with upgrade
- Upgrade go-licences to `v2` to use the latest fixes to the package
- Update the Makefile to handle tooling independently
  - Use Make tooling to identify when tools need to be rebuilt (based on file changes)
  - Use target dependencies to only build tooling when needed (no longer all or nothing)


rafaelroquetto (approved):
Nice!

Issue #283, opened 2025-07-09
[TracesHostInfo] Only produce metric for instances generating application metrics


jcreixell:
The `traces_host_info` metric should only be produced for hosts generating application telemetry (specifically, spanmetrics and servicegraph), and stop being produced when the host no longer produces such metrics.

At the moment, the prometheus implementation behaves correctly in the sense that is wrapped in an [expirer](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/1d909944bc3f7e73a193ec2050e69b26b069d06d/pkg/export/prom/prom.go#L573-L576) and only [updated](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/1d909944bc3f7e73a193ec2050e69b26b069d06d/pkg/export/prom/prom.go#L823) when consuming a span.

The OTEL implementation does not follow this pattern and always [exposes](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/1d909944bc3f7e73a193ec2050e69b26b069d06d/pkg/export/otel/metrics.go#L739) the metric, regardless on whether spans have been consumed or not.

PR #348, opened 2025-07-22
Initialize host info metric on first span

jcreixell:
- The metric `traces_host_info` metric should only be emitted for application hosts (i.e. hosts that emit traces).
  - This was the case for the prometheus exporter, but not for OTEL.
  - Fixes an issue where once the host info metric was initialized, it was never expired (even though it is wrapped in an `Expirer`) due to the fact that expirers only remove outdated entries as part of the process of recording new ones. Since the host won't go away (as opposed to service instances), the metric will always be recorded on every processed span.
  - Fixes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/283

### DIFF
--- a/pkg/export/otel/metrics_test.go
+++ b/pkg/export/otel/metrics_test.go
@@ -27,7 +27,6 @@ import (
 	"go.opentelemetry.io/obi/pkg/export/instrumentations"
 	"go.opentelemetry.io/obi/pkg/export/otel/otelcfg"
 	"go.opentelemetry.io/obi/pkg/pipe/msg"
-	"go.opentelemetry.io/obi/pkg/pipe/swarm"
 	"go.opentelemetry.io/obi/test/collector"
 )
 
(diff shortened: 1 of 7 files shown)


mariomac:
Thanks for your contribution!! Would it be possible to add a unit or integration test for this feature?

jcreixell:
I have now fixed the expiration for the prometheus case and am ready to start working on proper testing (it's been a hectic lately but will try to have something ready by the end of the week, can also pair to accelerate things).

grcevski:
I think unit test would be easier, for integration test we need maybe a test where on an API call we kill a service and then check after a while that the host metrics are gone. If we want to try that, maybe the easiest would be to add an API /shutdown to testserver (the Go program) and call it at the end of the other tests, the TestREDMetrics in suites_test.go. At this point we should eventually see host info disappear.

pkg/export/otel/metrics_test.go
```diff
@@ -511,6 +511,41 @@ func TestAppMetrics_ResourceAttributes(t *testing.T) {
...
+	assert.Len(t, res, 2)
+
+	found := false
+	for _, r := range res {
+		if r.Name == "traces_host_info" {
+			found = true
+			break
+		}
+	}
+	assert.True(t, found, "traces_host_info metric not found in the exported metrics")
```
grcevski: This is great, do you think it's possible to refactor the function so that we can directly call it and then ensure the call to host info delete happened?

grcevski (commented):
This looks good! I just wonder if it's possible to make sure the If condition on expiry truly does what we expect it should do.

mariomac:
@jcreixell good testing!! Once you address Nikola's comment and you remove the draft status, I'll do the final review and eventual merge. Thank you a lot!

jcreixell:
yes, addressing the comments now

jcreixell:
I still need to look into how easy would be it to test the prometheus part, but the OTEL part should be good to go

PR #350, opened 2025-07-22
Don't delete target info on expiry if the service is still instrumented

grcevski:
We recently changed the design of the handling of target_info to be setup when we first instrument a service and removed when the service is shutdown. This allows customers to implement liveness like graphs on the metrics data. However, I had put code in the metrics expiration to delete the target info when the metrics reporter for a service is expunged from the reporter pool.

This code was necessary because of the following scenarion:
1. A service has been generating metrics all the time and the metrics export buffer is full.
2. A service is now terminated, and we delete the target info based on the process remove hook.
3. Since there might be metrics that haven't fully been exported, the target info can be brought back, since it will be restored by the delayed exported events.

However, the way the code is setup today has one potential problem. A dormant service that doesn't export any data in 5 minutes (the default TTL interval) will be kicked out of the metrics pool, causing the function to run and remove the target info metric. It's not a big deal, since the metric will be restored as soon as new data is generated in the future, but it defeats the purpose of keeping target_info related to the process lifecycle so that dashboards don't end up with gaps.

With this PR I'm relying on the pid tracker to check if we should really remove the target info. If the service has been removed, then we remove it once more after all buffered metrics have been serialised. If the reporter is expunged because of inactivity, and the service is really alive, we don't delete the target info, relying on the process termination hook to do it when the process terminates.

### DIFF
--- a/pkg/export/otel/metrics.go
+++ b/pkg/export/otel/metrics.go
@@ -425,8 +425,10 @@ func newMetricsReporter(
 			llog.Debug("evicting metrics reporter from cache")
 			v.cleanupAllMetricsInstances()
 
-			mr.deleteTracesTargetInfo(v)
-			mr.deleteTargetInfo(v)
+			if !mr.pidTracker.ServiceLive(id) {
+				mr.deleteTracesTargetInfo(v)
+				mr.deleteTargetInfo(v)
+			}
 
 			go func() {
 				if err := v.provider.ForceFlush(ctx); err != nil {
--- a/pkg/export/otel/pid_tracker.go
+++ b/pkg/export/otel/pid_tracker.go
@@ -64,6 +64,15 @@ func (p *PidServiceTracker) RemovePID(pid int32) (bool, svc.UID) {
 	return false, svc.UID{}
 }
 
+func (p *PidServiceTracker) ServiceLive(uid svc.UID) bool {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+
+	_, exists := p.servicePIDs[uid]
+
+	return exists
+}
+
 func (p *PidServiceTracker) IsTrackingServerService(n svc.ServiceNameNamespace) bool {
 	_, ok := p.names[n]
 	return ok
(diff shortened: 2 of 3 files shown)


pkg/export/otel/metrics.go
```diff
@@ -425,8 +425,10 @@ func newMetricsReporter(
 			llog.Debug("evicting metrics reporter from cache")
 			v.cleanupAllMetricsInstances()
 
-			mr.deleteTracesTargetInfo(v)
-			mr.deleteTargetInfo(v)
+			if !mr.pidTracker.ServiceLive(id) {
```
MrAlias, line 428: Is this running the hot path? Is this added lock contention going to be a performance problem?

pkg/export/otel/metrics.go
```diff
@@ -425,8 +425,10 @@ func newMetricsReporter(
...
+				mr.deleteTracesTargetInfo(v)
+				mr.deleteTargetInfo(v)
+			}
```
mariomac, line 431: Do you think we might end up having multiple dangling `target_infos` for services that don't exist anymore?

I imagine:
- We stop reporting metrics for an existing service
- When the cache evicts, this function is executed but the target_infos are not removed because the process still exist
- When the process exists, the `deleteTargetInfo` function is not executed.

Maybe we'd need to create some kind of "cleanupResourceMetrics" function that is invoked either here, or after a signal (e.g. on resource removal)

pkg/export/otel/metrics.go
grcevski: No, it's very rare path. It is only triggered when a service stops sending telemetry for more than 5 minutes.

pkg/export/otel/metrics.go
grcevski: We execute the delete of the target info when we receive the process termination event. https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/35639ce473a02767d938174c9a83e8c284f0a624/pkg/export/otel/metrics.go#L1244.

So if a service stops sending telemetry, we don't remove the target info, it should get cleaned up when the service eventually terminates.

mariomac (approved):

PR #369, opened 2025-07-24
bpf: classify postgres in ebpf, add large buffers and errors support

mmat11:
- Classify postgres connections in ebpf
- Add large buffers support for postgres
- Parse postgres errors

### DIFF
--- a/bpf/bpfcore/utils.h
+++ b/bpf/bpfcore/utils.h
@@ -59,14 +59,8 @@
                  : "+r"(VAR)                                                                       \
                  : [max] "i"(UMAX))
 
-// Check if a number is a power of 2.
-#define IS_POW2(n) ((n) != 0 && (((n) & ((n) - 1)) == 0))
-
-// Return from function if the number is NOT a power of 2
-#define RET_IF_NOT_POW2(n, ret)                                                                    \
-    do {                                                                                           \
-        if (!IS_POW2(n))                                                                           \
-            return (ret);                                                                          \
-    } while (0)
+static __always_inline bool is_pow2(u32 n) {
+    return n != 0UL && (n & (n - 1)) == 0UL;
+}
 
 #endif /* __UTILS_H__ */
--- a/bpf/bpfcore/utils.h
+++ b/bpf/bpfcore/utils.h
@@ -59,4 +59,14 @@
                  : "+r"(VAR)                                                                       \
                  : [max] "i"(UMAX))
 
+// Check if a number is a power of 2.
+#define IS_POW2(n) ((n) != 0 && (((n) & ((n) - 1)) == 0))
+
+// Return from function if the number is NOT a power of 2
+#define RET_IF_NOT_POW2(n, ret)                                                                    \
+    do {                                                                                           \
+        if (!IS_POW2(n))                                                                           \
+            return (ret);                                                                          \
+    } while (0)
+
 #endif /* __UTILS_H__ */
--- a/bpf/common/common.h
+++ b/bpf/common/common.h
@@ -39,13 +39,7 @@
 #define HTTP_CONTENT_TYPE_MAX_LEN 16
 
 volatile const u32 mysql_buffer_size = 0;
-
-enum {
-    k_mysql_query_max = 8192,
-    k_mysql_query_max_mask = k_mysql_query_max - 1,
-    k_mysql_error_message_max = 512,
-    k_mysql_error_message_max_mask = k_mysql_error_message_max - 1
-};
+volatile const u32 postgres_buffer_size = 0;
 
 enum large_buf_action : u8 {
     k_large_buf_action_init = 0,
--- a/bpf/generictracer/protocol_mysql.h
+++ b/bpf/generictracer/protocol_mysql.h
@@ -54,8 +54,8 @@ enum {
     k_mysql_com_stmt_execute = 0x17,
 
     // Large buffer
-    k_large_buf_max_size = 1 << 14, // 16K
-    k_large_buf_max_size_mask = k_large_buf_max_size - 1,
+    k_mysql_large_buf_max_size = 1 << 14, // 16K
+    k_mysql_large_buf_max_size_mask = k_mysql_large_buf_max_size - 1,
 
     // Sanity checks
     k_mysql_payload_length_max = 1 << 13, // 8K
@@ -68,7 +68,7 @@ struct {
     __uint(max_entries, MAX_CONCURRENT_REQUESTS);
 } mysql_state SEC(".maps");
 
-SCRATCH_MEM_SIZED(mysql_large_buffers, k_large_buf_max_size);
+SCRATCH_MEM_SIZED(mysql_large_buffers, k_mysql_large_buf_max_size);
 
 // This function is used to store the MySQL header if it comes in split packets
 // from double send.
@@ -185,7 +185,7 @@ static __always_inline int mysql_send_large_buffer(tcp_req_t *req,
     req->has_large_buffers = true;
     bpf_ringbuf_output(&events,
                        large_buf,
-                       (sizeof(tcp_large_buffer_t) + written) & k_large_buf_max_size_mask,
+                       (sizeof(tcp_large_buffer_t) + written) & k_mysql_large_buf_max_size_mask,
                        get_flags());
     return 0;
 }
@@ -202,8 +202,12 @@ static __always_inline u32 mysql_command_offset(struct mysql_hdr *hdr) {
 static __always_inline u8 is_mysql(connection_info_t *conn_info,
                                    const unsigned char *data,
                                    u32 data_len,
-                                   u8 *packet_type,
                                    enum protocol_type *protocol_type) {
+    if (*protocol_type != k_protocol_type_mysql && *protocol_type != k_protocol_type_unknown) {
+        // Already classified, not mysql.
+        return 0;
+    }
+
     if (mysql_store_state_data(conn_info, data, (size_t)data_len) < 0) {
         bpf_dbg_printk("is_mysql: 4 bytes packet, storing state data");
         return 0;
@@ -246,7 +250,6 @@ static __always_inline u8 is_mysql(connection_info_t *conn_info,
                 "is_mysql: COM_QUERY or COM_PREPARE found, but buf doesn't contain a sql query");
             return 0;
         }
-        *packet_type = PACKET_TYPE_REQUEST;
         break;
     case k_mysql_com_stmt_execute:
         // COM_STMT_EXECUTE packet structure:
@@ -259,7 +262,6 @@ static __always_inline u8 is_mysql(connection_info_t *conn_info,
             // Already identified, mark this as a request.
             // NOTE: Trying to classify the connection based on this command
             // would be unreliable, as the check is too shallow.
-            *packet_type = PACKET_TYPE_REQUEST;
             break;
         }
         return 0;
@@ -270,7 +272,6 @@ static __always_inline u8 is_mysql(connection_info_t *conn_info,
             // If the request came in split packets, the sequence ID will be 2 (hdr->hdr_arrived == false) or 3 (hdr->hdr_arrived == true).
             bpf_dbg_printk("is_mysql: already identified as MySQL protocol");
             if ((hdr.sequence_id == 1 && !hdr.hdr_arrived) || hdr.sequence_id > 1) {
-                *packet_type = PACKET_TYPE_RESPONSE;
                 break;
             }
             bpf_dbg_printk(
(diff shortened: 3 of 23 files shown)


bpf/common/common.h
```diff
@@ -39,13 +39,7 @@
...
 
 volatile const u32 mysql_buffer_size = 0;
-
-enum {
-    k_mysql_query_max = 8192,
-    k_mysql_query_max_mask = k_mysql_query_max - 1,
-    k_mysql_error_message_max = 512,
-    k_mysql_error_message_max_mask = k_mysql_error_message_max - 1
-};
+volatile const u32 postgres_buffer_size = 0;
```
grcevski, line 42: I think we likely need one buffer size option for all protocols. What do you think? I think it might be an overkill to ask folks to configure each protocol?

bpf/generictracer/protocol_mysql.h
```diff
@@ -54,8 +54,8 @@ enum {
     k_mysql_com_stmt_execute = 0x17,
 
     // Large buffer
-    k_large_buf_max_size = 1 << 14, // 16K
-    k_large_buf_max_size_mask = k_large_buf_max_size - 1,
+    k_mysql_large_buf_max_size = 1 << 14, // 16K
```
grcevski, line 58: I guess we don't need these changes if we simply use one buffer size for all.

pkg/components/sqlprune/postgres.go
```diff
@@ -0,0 +1,103 @@
...
+// 'S' 'ERROR' '\0'                                             // Severity
+// 'C' '23505' '\0'                                             // SQLSTATE code (unique violation)
+// 'M' 'duplicate key value violates unique constraint "mytable_pkey"' '\0' // Error message
+// 'D' 'Key (id)=(1) already exists.' '\0'                      // Detail
+// 'H' 'Perhaps you need to add a conditional statement?' '\0'  // Hint
+// 'F' 'nbtinsert.c' '\0'                                       // Source file
+// 'L' '402' '\0'                                               // Source line number
+// 'R' '_bt_check_unique' '\0'                                  // Routine
+// '\0'                                                         // End of message
+func parsePostgresError(buf []uint8) *request.SQLError {
```
grcevski, line 45: Do you mind adding tests for this file?

grcevski (commented):
This is awesome, thanks! I just left a comment to discuss if we should bother with configuring specific buffer sizes for each protocol. Perhaps it makes sense, I'm just wondering if it was considered and there's a need for this level of detail.

bpf/common/common.h
mmat11: The rationale for having it configurable per protocol was that if the workload has heavy traffic across multiple protocols (http, mysql, postgres, etc.), increasing the buffer for all of them could generate significant load, while enabling it for a specific protocol is a safer choice. Also, since the default is 0, it effectively acts as a toggle to enable or disable the feature based on the protocol.

pkg/components/sqlprune/postgres.go
mmat11: I added tests for this to the generic `SQLParseError` function (see `pkg/components/sqlprune/sqlparser_test.go`)

bpf/generictracer/protocol_postgres.h
```diff
@@ -0,0 +1,156 @@
...
+#include <generictracer/protocol_common.h>
+#include <generictracer/k_tracer_tailcall.h>
+
+#include <generictracer/maps/protocol_cache.h>
+
+#include <maps/active_ssl_connections.h>
+
+struct postgres_hdr {
+    uint8_t message_type;
+    uint32_t message_len;
```
rafaelroquetto:
```suggestion
    u8 message_type;
    u32 message_len;
```

rationale: consistency with the rest of the code

bpf/generictracer/protocol_postgres.h
```diff
@@ -0,0 +1,156 @@
...
+} __attribute__((packed));
```
rafaelroquetto: don't pack the struct - by doing this, accessing message_len is UB because it's not longer aligned to its type (u32) - it also incurs the performance penalty for the misaligned reads (or a plain fault in some archs)

bpf/generictracer/protocol_postgres.h
```diff
@@ -0,0 +1,156 @@
...
+// The return value is used to control the flow for this specific protocol.
+// -1: wait additional data; 0: continue, regardless of errors.
+static __always_inline int postgres_send_large_buffer(tcp_req_t *req,
+                                                      pid_connection_info_t *pid_conn,
+                                                      const void *u_buf,
+                                                      u32 bytes_len,
+                                                      u8 direction,
+                                                      enum large_buf_action action) {
+    const u8 buf_len_mask =
+        postgres_buffer_size - 1; // postgres_buffer_size is guaranteed to be a power of 2
```
rafaelroquetto: can we sanity check that here as well?

bpf/generictracer/protocol_postgres.h
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    large_buf->type = EVENT_TCP_LARGE_BUFFER;
+    large_buf->direction = direction;
+    large_buf->action = action;
+    __builtin_memcpy((void *)&large_buf->tp, (void *)&req->tp, sizeof(tp_info_t));
+
+    large_buf->len = bytes_len;
+    if (large_buf->len >= mysql_buffer_size) {
+        large_buf->len = mysql_buffer_size;
```
rafaelroquetto: I believe this is a bug:

```suggestion
    if (large_buf->len >= postgres_buffer_size) {
        large_buf->len = postgres_buffer_size;
```

bpf/generictracer/protocol_postgres.h
```diff
@@ -0,0 +1,156 @@
...
+                       get_flags());
+    return 0;
+}
+
+static __always_inline void postgres_parse_hdr(struct postgres_hdr *hdr,
+                                               const unsigned char *data) {
+    bpf_probe_read(&hdr->message_type, 1, data);
+    u32 message_len_le;
+    bpf_probe_read(&message_len_le, 4, data + 1);
+    hdr->message_len = bpf_ntohl(message_len_le);
```
rafaelroquetto: Just like we did for mysql, we should do a single 5 byte `bpf_probe_read` call here. Don't pack the struct as later we'll be having lots of misaligned reads of `message_len` and other issues.

something like

```
// disclaimer: untested
u8 header[5];
bpf_probe_read(header, sizeof(header), data)

u32 len;
__builtin_memcpy(&len, header + 1, sizeof(len));

struct header hdr = {0};
hdr.type = header[0];
hdr.message_len = bpf_ntohl(len);

return hdr;
```
I'd also just return the header as a value (if the verifier doesn't give you trouble) as that allows for const correctness (and the compiler will apply NRVO anyway).

bpf/generictracer/protocol_postgres.h
```diff
@@ -0,0 +1,156 @@
...
+            includes_known_command = true;
+            break;
+        default:
+            bpf_dbg_printk("postgres_detect: unhandled message type 0x%x", hdr.message_type);
+            return 0;
+        }
+    }
+
+    if (message_size != data_len) {
+        bpf_dbg_printk("is_postgres: message length mismatch: message_size=%d data_len=%d",
```
rafaelroquetto:
```suggestion
        bpf_dbg_printk("is_postgres: message length mismatch: message_size=%u data_len=%u",
```

bpf/generictracer/protocol_postgres.h
```diff
@@ -0,0 +1,156 @@
...
+        bpf_dbg_printk("is_postgres: data_len is too short: %d", data_len);
+        return 0;
+    }
+
+    size_t message_size = 0;
+    struct postgres_hdr hdr;
+    int i;
+    bool includes_known_command = false;
+
+    for (i = 0; i < k_pg_messages_in_packet_max; i++) {
```
rafaelroquetto:
```suggestion
    bool includes_known_command = false;

    for (u8 i = 0; i < k_pg_messages_in_packet_max; i++) {
```

bpf/generictracer/protocol_postgres.h
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        postgres_parse_hdr(&hdr, data + message_size);
+
+        message_size += hdr.message_len + 1;
+        if (hdr.message_len == 0) {
+            break;
+        }
+
+        switch (hdr.message_type) {
```
rafaelroquetto, line 137: non-rhetorical question: what happens if, say, the for loop iterates 2 times? Assume on the first time you parsed the header, so you end up with a given `message_type` - will this not override the `message_type` on the second iteration?

bpf/generictracer/protocol_postgres.h
```diff
@@ -0,0 +1,156 @@
...
+
+    if (!includes_known_command) {
+        bpf_dbg_printk("is_postgres: no known command found");
+        return 0;
+    }
+
+    *protocol_type = k_protocol_type_postgres;
+    bpf_map_update_elem(&protocol_cache, conn_info, protocol_type, BPF_ANY);
+
+    bpf_dbg_printk("is_postgres: postgres! message_type=%d", hdr.message_type);
```
rafaelroquetto:
```suggestion
    bpf_dbg_printk("is_postgres: postgres! message_type=%u", hdr.message_type);
```

pkg/components/sqlprune/postgres.go
grcevski: I was thinking maybe making some simple tests in postgres_test.go and just running various edge cases, maybe it's all covered, but it's hard to tell with the other test. These days any Gen AI tool will give you a good start if you  point it to this code.

bpf/common/common.h
grcevski: ok, cool. I can see this being important, it wasn't obvious since both are SQL, but I can imagine you want less bytes for specific protocol.

pkg/components/sqlprune/postgres.go
mmat11: Done

bpf/generictracer/protocol_postgres.h
mmat11: Done

bpf/generictracer/protocol_postgres.h
mmat11: Done

bpf/generictracer/protocol_postgres.h
mmat11: Yes, copypaste gone wrong

bpf/generictracer/protocol_postgres.h
mmat11: Done

bpf/generictracer/protocol_postgres.h
mmat11: Done

bpf/generictracer/protocol_postgres.h
mmat11: Done

bpf/generictracer/protocol_postgres.h
mmat11: We made it this way because postgres packets can contain more than one message so it's expected we overwrite the hdr; the important bit is that we add the size of each message while iterating so we can later check the total size of the packet against the calculated size of all the messages

bpf/generictracer/protocol_postgres.h
mmat11: Done

bpf/generictracer/protocol_postgres.h
mmat11: Done

bpf/bpfcore/utils.h
```diff
@@ -59,4 +59,14 @@
                  : "+r"(VAR)                                                                       \
                  : [max] "i"(UMAX))
 
+// Check if a number is a power of 2.
+#define IS_POW2(n) ((n) != 0 && (((n) & ((n) - 1)) == 0))
```
rafaelroquetto: IMHO we should avoid macros, even if that means slightly longer code, as they can harbour obscure bugs. For instance, in this case it evaluates n 3 times. For example, if you call `RET_IF_NOT_POW2(i++, -1)`, you’ll get unexpected behaviour (and undefined behaviour from modifying i twice without a sequence point).

Another issue is that bitwise manipulations are not well defined for signed integers with negative values - we should restrict this to unsigned values only - so a type safer alternative would just be turning this into a function:

```
static __always_inline bool is_pow2_u(unsigned long n) { // or u64 or whatever the most appropriate type here is
    return n != 0UL && (n & (n - 1)) == 0UL;
}
```

bpf/bpfcore/utils.h
```diff
@@ -59,4 +59,14 @@
...
+
+// Return from function if the number is NOT a power of 2
+#define RET_IF_NOT_POW2(n, ret)                                                                    \
```
rafaelroquetto: in the same spirit, can you please get rid of this and just spell in place
```
if (!is_pow2_u(foo)) {
   return
}
```

this doesn't save us a lot of lines, and is a very restrictive macro (i.e. can't be used in contexts returning `void`, overall it's just an abstraction that obscures otherwise trivial code - a plain `if (..) return` is clearer for a reader unfamiliar with the codebase.

bpf/bpfcore/utils.h
mmat11: Done

bpf/bpfcore/utils.h
mmat11: Done

rafaelroquetto (approved):
Thank you - that was quick! :)

PR #372, opened 2025-07-25
Add obi.version and obi.revision as resource attributes in network metrics

mariomac:
This only affects OTEL exports ATM.

Since Prometheus resource metrics (target_info) might require extra coordination effort between te different exporters (network and application), that attribute will go into another PR.

### DIFF
--- a/pkg/buildinfo/buildinfo.go
+++ b/pkg/buildinfo/buildinfo.go
@@ -4,6 +4,7 @@
 package buildinfo
 
 // Version and Revision variables are overridden at build time with Git repository information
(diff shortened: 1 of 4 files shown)


skl (approved):

grcevski (commented):
Do we not need a change in `func RecordGetters(name attr.Name) (attributes.Getter[*Record, attribute.KeyValue], bool)` ?

mariomac:
@grcevski we don't need it because `RecordGetters` applies only to metric-level attributes, while resource-level attributes (the ones that go in `target_info` are fixed)

PR #373, opened 2025-07-25
Add metric for instrumentation error

marctc:
Add metric to track different errors that could happen when trying to instrument processes.

Fixes https://github.com/grafana/beyla/issues/1881


rafaelroquetto (approved):
No bats, no bugs — Ozzy would approve. So do I. 🦇✅

grcevski (changes requested):
Can we please add an integration or unit test to prove this works? One way to test this would be to make a configuration file with insufficient permissions. For example we can attempt to run OBI without privileged in one of the docker test files. This should cause failure that we should be able to identify.

grcevski (approved):
Sweet!

PR #375, opened 2025-07-25
Executable isn't ready if there's no path

grcevski:
If we scan the /proc/ file system for new processes, we might get unlucky and scan the process as it's being created, which might cause us to skip discovery of a process. This PR changes the condition of marking a process ready, if there's no path we say the process isn't ready.


rafaelroquetto (approved):

NimrodAvni78 (approved):

mariomac (approved):
Good catch!! I bet this will fix some flakiness in tests (and hence in prod)

PR #381, opened 2025-07-28
Refactor OTEL Metrics exporter

mariomac:
In preparation for a fix in service graph metrics, this PR reworks the OTEL exporter to avoid that multiple metrics export nodes have their own OTEL metrics exporter (more connections, more resources, more API calls...).

* The OTEL service graph metrics generation has been moved to another node. The cause is that in the future fix, service graph metrics will require their own resource attributes, incompatible with rest of RED metrics.
* The OTEL exporter creation has been unified into a factory patter that will return the same exporter for all the metrics exporters (network, RED, service graph, internal...). Similar to what we currently do for the Prometheus Exporter.

This PR mostly just moves some types and definitions to a common "otelcfg" package to avoid cyclic dependencies.

### DIFF
--- a/pkg/components/pipe/instrumenter.go
+++ b/pkg/components/pipe/instrumenter.go
@@ -114,6 +114,14 @@ func newGraphBuilder(
 		processEventsCh,
 	), swarm.WithID("OTELMetricsExport"))
 
+	swi.Add(otel.ReportSvcGraphMetrics(
+		ctxInfo,
+		&config.Metrics,
+		selectorCfg,
+		exportableSpans,
+		processEventsCh,
+	), swarm.WithID("OTELSvcGraphMetricsExport"))
+
 	swi.Add(otel.TracesReceiver(
 		ctxInfo, config.Traces, config.Metrics.SpanMetricsEnabled(), selectorCfg, exportableSpans,
 	), swarm.WithID("OTELTracesReceiver"))
--- a/pkg/export/otel/metrics_svc_graph.go
+++ b/pkg/export/otel/metrics_svc_graph.go
@@ -8,6 +8,8 @@ import (
 	"fmt"
 	"log/slog"
 
+	"go.opentelemetry.io/obi/pkg/export/otel/otelcfg"
+
 	"go.opentelemetry.io/otel/attribute"
 	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
 	"go.opentelemetry.io/otel/sdk/resource"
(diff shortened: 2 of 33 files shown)


pkg/export/otel/otelcfg/exporter.go
skl, line 1: Possible to add unit tests?

pkg/export/otel/metrics_svc_graph.go
skl, line 1: Possible to add unit tests?

pkg/export/otel/otelcfg/exporter.go
mariomac: Good point. As an accessory type, most of the behavior is already tested, especially in integration tests.

But I'll add a unit test for the Instance method, to verify that it does not introduce any race condition.

pkg/export/otel/metrics_svc_graph.go
mariomac: Done!

rafaelroquetto (approved):

pkg/components/pipe/instrumenter.go
```diff
@@ -114,6 +114,14 @@ func newGraphBuilder(
 		processEventsCh,
 	), swarm.WithID("OTELMetricsExport"))
 
+	swi.Add(otel.ReportSvcGraphMetrics(
```
grcevski, line 117: Nice, this is much cleaner design!

pkg/export/otel/metrics_svc_graph.go
```diff
@@ -0,0 +1,467 @@
...
+		attributes.Field[*request.Span, attribute.KeyValue]{
+			ExposedName: string(attr.HostID.OTEL()),
+			Get: func(_ *request.Span) attribute.KeyValue {
+				return semconv.HostID(mr.hostID)
+			},
+		})
+}
+
+func (mr *SvcGraphMetricsReporter) serviceGraphAttributes() []attributes.Field[*request.Span, attribute.KeyValue] {
+	return attributes.OpenTelemetryGetters(
```
grcevski, line 324: OK, so just to confirm, this PR doesn't actually add the k8s_cluster_name as an attribute to the service graph and it doesn't prefix the not default fields, as described [here](https://pkg.go.dev/github.com/open-telemetry/opentelemetry-collector-contrib/connector/servicegraphconnector#section-readme), with `client_` and `server_`? I guess we need to follow-up with another PR on that after this is merged?

grcevski (approved):

pkg/export/otel/metrics_svc_graph.go
mariomac: You are right. This is just the refactor to allow a later, smaller PR fixing the issue. I preferred this way to avoid having an even bigger and disperse PR

PR #384, opened 2025-07-28
Don't filter out pod update events

rafaelroquetto:
The criteria used to filtering out POD events is not sufficient, causing it to filter out important POD events such as when the POD container changes.
It also prevents users from instrumenting headless services, a decision that should potentially be asserted via configuration, rather than hardcoded.

This fixes an important and silent bug which prevents OBI from instrumenting services that have been killed/restarted but end up in a different container (but same pod).

### DIFF
--- a/pkg/kubecache/envtest/integration_test.go
+++ b/pkg/kubecache/envtest/integration_test.go
@@ -281,64 +281,6 @@ func TestAsynchronousStartup(t *testing.T) {
 	assert.LessOrEqual(t, int32(createdPods), cl3.syncSignalOnMessage.Load())
 }
 
-func TestIgnoreHeadlessServices(t *testing.T) {
-	svcClient := serviceClient{
-		Address:  fmt.Sprintf("127.0.0.1:%d", freePort),
-		Messages: make(chan *informer.Event, 10),
-	}
-	test.Eventually(t, timeout, func(t require.TestingT) {
-		svcClient.Start(ctx, t, 0)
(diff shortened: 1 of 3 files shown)


rafaelroquetto:
@mariomac could you please have a look at this PR? I'm assuming the code being removed was there for a good reason in first place, so I'd really appreciate if you could confirm this approach is safe.

grcevski (approved):
Thanks Rafael! I think this makes sense. Let's see if Mario knows maybe why we needed those checks, perhaps there's some extra overhead, but I couldn't see extra events in the experiments I did.

The headless services, I think they can be clients too or maybe even servers, with pod IPs instead of service ips. I think we should let them through and have people opt them out if needed in discovery.

mariomac:
I think the `headlessService` verification can be removed. It is a leaked check from having two separate informers in the past (one for network metrics and another for app metrics). When we merged both informers, this code was left there but I did not consider that it could affect Pods. Good catch!!

The `unchanged` verification is a performance optimization to avoid sending the same message multiple times from the external cache to the OBI instance, because the message encoding/decoding puts a lot of pressure in CPU, Memory and Network on big clusters.

Instead of removing it, could we just improve it to detect changes in the Pod containers?

grcevski (approved):

grcevski:
@NimrodAvni78 if you have a chance please try this fix to see if it resolves the issue you are seeing, thanks a bunch!

PR #385, opened 2025-07-29
ignore mongo server spans

NimrodAvni78:
we dont have this event type, unless we want to create it, currently we missclassify mongo server spans and client spans

### DIFF
--- a/pkg/components/ebpf/common/tcp_detect_transform.go
+++ b/pkg/components/ebpf/common/tcp_detect_transform.go
@@ -116,18 +116,19 @@ func ReadTCPRequestIntoSpan(parseCtx *EBPFParseContext, cfg *config.EBPFTracer, 
 			return TCPToFastCGIToSpan(event, op, uri, status), false, nil
 		}
 	}
-
-	var mongoRequest *MongoRequestValue
-	var moreToCome bool
-	_, _, err = ProcessMongoEvent(requestBuffer, int64(event.StartMonotimeNs), int64(event.EndMonotimeNs), event.ConnInfo, parseCtx.mongoRequestCache)
-	if err == nil {
-		mongoRequest, moreToCome, err = ProcessMongoEvent(responseBuffer, int64(event.StartMonotimeNs), int64(event.EndMonotimeNs), event.ConnInfo, parseCtx.mongoRequestCache)
-	}
-	if err == nil && !moreToCome && mongoRequest != nil {
-		mongoInfo, err := getMongoInfo(mongoRequest)
+	if event.Direction != 0 {
+		var mongoRequest *MongoRequestValue
+		var moreToCome bool
+		_, _, err = ProcessMongoEvent(requestBuffer, int64(event.StartMonotimeNs), int64(event.EndMonotimeNs), event.ConnInfo, parseCtx.mongoRequestCache)
 		if err == nil {
-			mongoSpan := TCPToMongoToSpan(event, mongoInfo)
-			return mongoSpan, false, nil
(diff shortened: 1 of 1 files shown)


pkg/components/ebpf/common/tcp_detect_transform.go
```diff
@@ -116,18 +116,19 @@ func ReadTCPRequestIntoSpan(parseCtx *EBPFParseContext, cfg *config.EBPFTracer,
...
-			mongoSpan := TCPToMongoToSpan(event, mongoInfo)
-			return mongoSpan, false, nil
+			mongoRequest, moreToCome, err = ProcessMongoEvent(responseBuffer, int64(event.StartMonotimeNs), int64(event.EndMonotimeNs), event.ConnInfo, parseCtx.mongoRequestCache)
+		}
+		if err == nil && !moreToCome && mongoRequest != nil {
+			mongoInfo, err := getMongoInfo(mongoRequest)
+			if err == nil {
+				mongoSpan := TCPToMongoToSpan(event, mongoInfo)
+				return mongoSpan, false, nil
+			}
```
mariomac: Is it possible to replace excessive nesting by some early exits on error? Something like:
```go
_, _, err = ProcessMongoEvent(requestBuffer, int64(event.StartMonotimeNs), int64(event.EndMonotimeNs), event.ConnInfo, parseCtx.mongoRequestCache)
if err != nil {
	return nil, nil, err
}
mongoRequest, moreToCome, err := ProcessMongoEvent(responseBuffer, int64(event.StartMonotimeNs), int64(event.EndMonotimeNs), event.ConnInfo, parseCtx.mongoRequestCache)
if err != nil {
	return nil, nil, err
}
if !moreToCome && mongoRequest != nil {
	mongoInfo, err := getMongoInfo(mongoRequest)
	if err == nil {
		mongoSpan := TCPToMongoToSpan(event, mongoInfo)
		return mongoSpan, false, nil
	}
```

NimrodAvni78: can't really early exit from this function since if mongo inference fails there are more protocols to be inferred, i can move it to a different function and implement this

rafaelroquetto: I think moving to a different function is the way to go IMHO.

pkg/components/ebpf/common/tcp_detect_transform.go
```diff
@@ -116,18 +116,19 @@ func ReadTCPRequestIntoSpan(parseCtx *EBPFParseContext, cfg *config.EBPFTracer,
...
-
-	var mongoRequest *MongoRequestValue
-	var moreToCome bool
-	_, _, err = ProcessMongoEvent(requestBuffer, int64(event.StartMonotimeNs), int64(event.EndMonotimeNs), event.ConnInfo, parseCtx.mongoRequestCache)
-	if err == nil {
-		mongoRequest, moreToCome, err = ProcessMongoEvent(responseBuffer, int64(event.StartMonotimeNs), int64(event.EndMonotimeNs), event.ConnInfo, parseCtx.mongoRequestCache)
-	}
-	if err == nil && !moreToCome && mongoRequest != nil {
-		mongoInfo, err := getMongoInfo(mongoRequest)
+	if event.Direction != 0 {
```
rafaelroquetto: Maybe outside of the scope of this PR, but eventually we should really replace these magic numbers with proper meaningful constants. Just mentioning this here because it caught my attention, feel free to ignore.

pkg/components/ebpf/common/tcp_detect_transform.go
NimrodAvni78: @rafaelroquetto @mariomac done

mariomac (approved):

PR #386, opened 2025-07-29
postgres: support prepared statements

mmat11:


bpf/generictracer/protocol_postgres.h
```diff
@@ -140,10 +139,8 @@ static __always_inline u8 is_postgres(connection_info_t *conn_info,
         case k_pg_msg_bind:
         case k_pg_msg_execute:
             includes_known_command = true;
-            break;
```
rafaelroquetto, line 139: don't fall-through, it makes for really fragile code - we should probably enable `-Wimplicit-fallthrough` (not as a part of this PR, I can do this later I guess)

pkg/components/ebpf/common/sql_detect_postgres.go
```diff
@@ -173,8 +185,47 @@ func postgresPreparedStatements(b []byte) (string, string, string) {
...
+		ptr      = 0
+	)
+
+	for ptr < len(buf) {
+		if ptr+sqlprune.PostgresHdrSize > len(buf) {
+			return nil, errors.New("remaining buffer too short for message header")
+		}
+
+		msgType := sqlprune.SQLParseCommandID(request.DBPostgres, buf[ptr:])
+		size := int(binary.BigEndian.Uint32(buf[ptr+1 : ptr+5]))
```
rafaelroquetto: Just confirming, is the original message header value a `uint32` or an `int32`? https://www.postgresql.org/docs/current/protocol-message-formats.html implies it's a signed value but I am not sure I am looking at the right place

pkg/components/ebpf/common/sql_detect_postgres.go
```diff
@@ -173,8 +185,47 @@ func postgresPreparedStatements(b []byte) (string, string, string) {
...
+type postgresMessage struct {
+	typ  string
+	size int
+	data []byte
+}
+
+func messagesFromBuffer(buf []byte) ([]postgresMessage, error) {
 	var (
+		messages = make([]postgresMessage, 0)
+		ptr      = 0
```
rafaelroquetto: nit: I'd call this `offset`, `pos`, or anything that is not `ptr` - I was under the impression this was a pointer looking at the code down below.

Having said that, another way to go about this is to simply advance `buf` directly and do without this variable altogether - which ends up being a bit more idiomatic, e.g.

```
for len(buf) > 0 {
    if len(buf) < sqlprune.PostgresHdrSize {
        /* ... */
    }
   
   /* ... */
   buf = buf[sqlprune.PostgresHdrSize + 1:]
   /* ... */
}
````

File: pkg/components/ebpf/common/sql_detect_postgres.go [hunk 4]
```diff
@@ -173,8 +185,47 @@ func postgresPreparedStatements(b []byte) (string, string, string) {
...
+		// but it does include itself (the 4-byte length field).
+		fixedSize := size - sqlprune.PostgresHdrSize + 1
+		if fixedSize > len(buf) {
+			return nil, errors.New("remaining buffer too short for message data")
+		}
+
+		data := buf[ptr : ptr+fixedSize]
+		ptr += fixedSize
+
+		messages = append(messages, postgresMessage{typ: msgType, size: fixedSize, data: data})
```
rafaelroquetto: instead of doing this and returning `messages`, since you loop over `messages` in the calling code, it's best to rewrite this as
```
type parseContext struct {
   hasSpan bool
   spanCommand string
   /* ... */
}

ctx := parseContext{}

for ... {
   /* ... */
   handlePostgresMessage(message, ctx) // ctx is passed as a pointer here

  if ctx.hasSpan {
      return ctx // or whatever makes sense
  }
```
right here. Then the caller can do without the loop, see my next comment.

pkg/components/ebpf/common/sql_detect_postgres.go
```diff
@@ -188,22 +239,86 @@ func handlePostgres(_ *EBPFParseContext, event *TCPRequestInfo, requestBuffer, r
 		return span, errFallback
 	}
 
-	sqlCommand := sqlprune.SQLParseCommandID(request.DBPostgres, requestBuffer)
+	requestMessages, err := messagesFromBuffer(requestBuffer)
```
rafaelroquetto: then here `messagesFromBuffer` would return `ctx` or whatever is required (and also an error).

The rationale behind this is mostly to avoid allocating `[]messages` (which from experience ends up with quite a lot of overhead in such a hotpath, and each individual `postgressMessage` inside, I reckon they'd all escape to the heap.

It has the added bonus that we are forced to encapsulate the body of the `for` loop below into a nice function, which clears things up a bit.

pkg/components/ebpf/common/sql_detect_postgres.go
mmat11: it should be `int32`

bpf/generictracer/protocol_postgres.h
mmat11: should I repeat for every command

```
            includes_known_command = true;
            break;
```

for consistency?

rafaelroquetto: No, the fallthrough such as 
```
case A:
case B:
case C:
   x = 1;
   break;
```

is fine and usual

the problem (and what -Wimplicit-fallthrough will complain about) is
```
case A:
  x = 0
  // fallthrough
case B:
  y = 1;
  break;
```

mmat11: done 👍🏻

pkg/components/ebpf/common/sql_detect_postgres.go
mmat11: done

pkg/components/ebpf/common/sql_detect_postgres.go
mmat11: I tried doing it like that but it looked a bit messy and hard to follow to me. I did it with an iterator instead, let me know if you like it

pkg/components/ebpf/common/sql_detect_postgres.go
```diff
@@ -173,8 +185,47 @@ func postgresPreparedStatements(b []byte) (string, string, string) {
...
+	fixedSize := size - sqlprune.PostgresHdrSize + 1
+	if len(it.buf) < fixedSize {
+		it.err = errors.New("remaining buffer too short for message data")
+		return nil
+	}
+
+	data := it.buf[:fixedSize]
+	it.buf = it.buf[fixedSize:]
+
+	return &postgresMessage{typ: msgType, size: fixedSize, data: data}
```
rafaelroquetto: So, here we escape to the heap (i.e. we allocate), which is what we are trying to avoid. To accomplish this, you simply return `postgresMessage` by value.

rafaelroquetto (commented):
Thank you for the changes @mmat11! Nice work with the iterator - there's just one final adjustment I think we should make - instead of returning a pointer in `next()`, return a value - this will ensure nothing is being allocated. With the present code, there are 2 allocs per operation (according to benchmarks).

Semantically speaking, instead of using `nil` to symbolise the end iterator, you can use a sentinel value, a function `it.isEof()`, or even return a bool - whatever is the clearest. (C++ iterators tend to return a end value, so one writes `if (it == container.end())` for reference/inspiration)

We should then add an extra test to ensure there are no allocations happening:

```
func TestMessageIteratorNoAllocs(t *testing.T) {
	buf := func() []byte {
		// First message: type 'Q', length 11, data "SELECT\x00"
		// Second message: type 'Q', length 11, data "COMMIT\x00"
		b := []byte{'Q', 0, 0, 0, 11}
		b = append(b, append([]byte("SELECT"), 0)...)
		b = append(b, 'Q', 0, 0, 0, 11)
		b = append(b, append([]byte("COMMIT"), 0)...)
		return b
	}()

	allocs := testing.AllocsPerRun(1000, func() {
		it := &postgresMessageIterator{buf: buf}

		for {
			msg := it.next()

			if msg.size == <end sentinel value> {
				break;
			}
		}
	})

	if allocs != 0 {
		t.Errorf("MessageIterator allocated %v allocs per run; want 0", allocs)
	}
}
```

Finally, do we really need `fixedSize`, or can we just infer the size from `it.buf`, since you do:

```
data := it.buf[:fixedSize]
```

rafaelroquetto (approved):
Good stuff, thanks a lot for this!

mmat11:
> Good stuff, thanks a lot for this!

I still need to fix OATS, working on it

PR #390, opened 2025-07-30
Add internal metrics for avoided services

marctc:
Add metric to track those services detected by OBI but we don't instrument because they are already instrumented with the collector.

Fixes https://github.com/grafana/beyla/issues/1882

### DIFF
--- a/test/integration/internal_metrics_test.go
+++ b/test/integration/internal_metrics_test.go
@@ -124,8 +124,8 @@ func checkAvoidedServicesMetrics(t *testing.T) {
 		}
 
 		// Assert specific values for service_name and service_namespace
-		assert.Equal(t, "example-service", labelMap["service_name"], "service_name label should be 'example-service'")
-		assert.Equal(t, "default", labelMap["service_namespace"], "service_namespace label should be 'default'")
+		assert.Equal(t, "rolldice", labelMap["service_name"], "service_name label should be 'rolldice'")
+		assert.Equal(t, "integration-test", labelMap["service_namespace"], "service_namespace label should be 'integration-test'")
 		assert.NotEmpty(t, labelMap["telemetry_type"], "telemetry_type label should not be empty")
 		assert.Condition(t, func() bool {
 			return labelMap["telemetry_type"] == "metrics" || labelMap["telemetry_type"] == "traces"


pkg/components/imetrics/iprom.go
```diff
@@ -79,6 +80,10 @@ func NewPrometheusReporter(cfg *PrometheusConfig, manager *connector.PrometheusM
 			Name: attr.VendorPrefix + "_instrumentation_errors_total",
 			Help: "Total number of instrumentation errors by process and error type",
 		}, []string{"process_name", "error_type"}),
+		avoidedServices: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: attr.VendorPrefix + "_avoided_services",
+			Help: "Services avoided by Beyla due to existing OpenTelemetry instrumentation",
```
grcevski: Mention of Beyla in the description. Let's avoid the product name or use the vendor prefix if needed.

pkg/components/imetrics/iprom.go
```diff
@@ -160,3 +167,20 @@ func (p *PrometheusReporter) UninstrumentProcess(processName string) {
 func (p *PrometheusReporter) InstrumentationError(processName string, errorType string) {
 	p.instrumentationErrors.WithLabelValues(processName, errorType).Inc()
 }
+
+func (p *PrometheusReporter) recordAvoidedService(serviceName, serviceNamespace, serviceInstanceID, serviceVersion, telemetryType string) {
+	// Use empty string for service_version if not provided, consistent with other Prometheus metrics
+	version := serviceVersion
+	if version == "" {
```
grcevski: This looks like a mistake?

pkg/export/otel/metrics_internal.go
```diff
@@ -106,6 +108,14 @@ func NewInternalMetricsReporter(ctx context.Context, ctxInfo *global.ContextInfo
 		return nil, err
 	}
 
+	avoidedServices, err := meter.Int64Gauge(
+		attr.VendorPrefix+".avoided.services",
+		instrument.WithDescription("Services avoided by Beyla due to existing OpenTelemetry instrumentation"),
```
grcevski: another mention of Beyla.

pkg/export/otel/metrics_internal.go
```diff
@@ -175,3 +186,27 @@ func (p *InternalMetricsReporter) InstrumentationError(processName, errorType st
...
+func (p *InternalMetricsReporter) recordAvoidedService(serviceName, serviceNamespace, serviceInstanceID, serviceVersion, telemetryType string) {
+	attrs := []attribute.KeyValue{
+		semconv.ServiceName(serviceName),
+		semconv.ServiceNamespace(serviceNamespace),
+		semconv.ServiceInstanceID(serviceInstanceID),
+		attribute.String("telemetry_type", telemetryType),
+	}
+	
+	// Only add service version if it's not empty
+	if serviceVersion != "" {
```
grcevski: Are you sure this works? We always set the labels in certain order, if the values don't match, I wonder if we'll end up storing a wrong value for certain label?

test/integration/internal_metrics_test.go
```diff
@@ -0,0 +1,178 @@
...
+	io_prometheus_client "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"go.opentelemetry.io/obi/test/integration/components/docker"
+	"go.opentelemetry.io/obi/test/integration/components/prom"
+)
+
+func TestInstrumentationErrors(t *testing.T) {
```
grcevski, line 24: Can we please add a unit test or an integration test that ensures the behaviour or empty version is correct?

grcevski (changes requested):

pkg/export/otel/metrics_internal.go
marctc: I don't think we have service version for other services so we can remove it from here

pkg/components/ebpf/common/pids.go
```diff
@@ -215,16 +218,37 @@ func (pf *IdentityPidsFilter) Filter(inputSpans []request.Span) []request.Span {
...
-func checkIfExportsOTelSpanMetrics(svc *svc.Attrs, span *request.Span) {
+func (pf *PIDsFilter) checkIfExportsOTelSpanMetrics(svc *svc.Attrs, span *request.Span) {
 	if span.IsExportTracesSpan() {
 		svc.SetExportsOTelMetricsSpan()
+		pf.reportAvoidedService(svc, "metrics_span")
+	}
+}
+
+// reportAvoidedService calls the appropriate internal metrics method based on telemetry type
+func (pf *PIDsFilter) reportAvoidedService(svc *svc.Attrs, telemetryType string) {
```
grcevski, line 239: Do we need to check somewhere here that internal metrics are enabled so that we don't call this function if they are not on?

The reason this is very important is that this code will now run for every span that comes through. 

It also raises the question, do we want to count every time a span is processed? Seems very excessive for something that's only per service. I'd much rather have this check done when we detect that we have otel enabled service and then never again.

grcevski (commented):
This looks great now, the remaining concern I have is that this will invariably slow down significantly the span processing loop and we should make an effort to report the metrics only once per service once we discover that it is exporting otel.

PR #392, opened 2025-07-30
Fix missing namespace discovery from env variables

grcevski:
We were correctly setting the namespace from environment variables for Kubernetes, but not when the process isn't running in k8s we were just parsing the service name, not namespace.

### DIFF
--- a/pkg/components/exec/file.go
+++ b/pkg/components/exec/file.go
@@ -29,9 +29,10 @@ type FileInfo struct {
 }
 
 const (
-	envServiceName   = "OTEL_SERVICE_NAME"
-	envResourceAttrs = "OTEL_RESOURCE_ATTRIBUTES"
-	serviceNameKey   = "service.name"
+	envServiceName      = "OTEL_SERVICE_NAME"
+	envResourceAttrs    = "OTEL_RESOURCE_ATTRIBUTES"
+	serviceNameKey      = "service.name"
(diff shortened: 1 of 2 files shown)


pkg/components/exec/file.go
```diff
@@ -75,24 +76,32 @@ func FindExecELF(p *services.ProcessInfo, svcID svc.Attrs, k8sEnabled bool) (*Fi
...
-		// If Kubernetes is enabled we use the K8S metadata as the source of truth
-		if !k8sEnabled {
-			file.Service.UID.Name = svcName
-		}
+	setServiceEnvVariables(&file.Service, envVars, k8sEnabled)
+
+	return &file, nil
+}
+
+func setServiceEnvVariables(service *svc.Attrs, envVars map[string]string, k8sEnabled bool) {
```
MrAlias: Can we use return values instead of having the function perform a side-effect on the parameter?


```suggestion
func setServiceEnvVariables(service *svc.Attrs, envVars map[string]string, k8sEnabled bool) *svc.Attrs {
```

This way the function signature clearly communicates what it accepts and what value it will produce.

grcevski: Good idea! I'll make the change.

grcevski: Done!

mariomac (approved):
LGTM!

PR #393, opened 2025-07-30
Fix missing pid info and prometheus collections

grcevski:
There are two separate issues with our CUDA kernel launch monitoring:

1. We recently changed all kprobes related code to share the same PID filter, the common PID filter, but the GPU/CUDA monitoring isn't tested because we don't have NVIDIA runners. The change to use the correct PID filter exposed a bug in the CUDA event creation that we weren't populating the PID info, so the filter can correctly remove unnecessary events. This is the first fix.
2. There were 2 series attributes (grid size and block size) which were missing to be included in the prometheus constructor.

### DIFF
--- a/pkg/components/ebpf/gpuevent/gpuevent.go
+++ b/pkg/components/ebpf/gpuevent/gpuevent.go
@@ -266,6 +266,11 @@ func (p *Tracer) readGPUMallocIntoSpan(record *ringbuf.Record) (request.Span, bo
 	return request.Span{
 		Type:          request.EventTypeGPUMalloc,
 		ContentLength: int64(event.Size),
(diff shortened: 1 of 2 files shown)


pkg/components/ebpf/gpuevent/gpuevent.go
```diff
@@ -275,21 +280,26 @@ func (p *Tracer) readGPUKernelLaunchIntoSpan(record *ringbuf.Record) (request.Sp
 		return request.Span{}, true, err
 	}
 
-	// Log the GPU Kernel Launch event
```
grcevski, line 278: I made this debug option, it's too verbose for normal usage.

MrAlias (approved):

PR #394, opened 2025-07-31
Enable misc-unused-parameters checks

rafaelroquetto:
Ensure clang-tidy notifies us of unused parameters. This requires us to explicitly note unused parameters that cannot be removed (such as kprobe signatures), but it also caught a few instances in which such parameters should no longer be there. It also uncovered a few other warnings that were being shadowed.

On a follow up PR I intend to enable `-Werror -Wunused-parameter` - I am not doing this now because a bunch of those are passed to `bpf_dbg_printk`, which expands to nothing on  non-debug builds, causing a few parameters to only be unused in a few instances (e.g. they need to be tagged with `[[maybe_unused]]`).

### DIFF
--- a/bpf/.clang-tidy
+++ b/bpf/.clang-tidy
@@ -12,7 +12,6 @@ Checks:
   - '-bugprone-narrowing-conversions' #TODO this needs to be checked/enabled
   - 'clang-analyzer-*'
   - '-clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling'
-  - '-misc-unused-parameters'
   - '-misc-misplaced-const'
   - '-misc-include-cleaner'
   - '-readability-else-after-return'
--- a/bpf/generictracer/protocol_postgres.h
+++ b/bpf/generictracer/protocol_postgres.h
@@ -50,12 +50,8 @@ SCRATCH_MEM_SIZED(postgres_large_buffers, k_pg_large_buf_max_size);
 // Emit a large buffer event for Postgres protocol.
 // The return value is used to control the flow for this specific protocol.
 // -1: wait additional data; 0: continue, regardless of errors.
-static __always_inline int postgres_send_large_buffer(tcp_req_t *req,
-                                                      pid_connection_info_t *pid_conn,
-                                                      const void *u_buf,
-                                                      u32 bytes_len,
-                                                      u8 direction,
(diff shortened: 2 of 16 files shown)


bpf/generictracer/protocol_postgres.h
```diff
@@ -50,12 +50,8 @@ SCRATCH_MEM_SIZED(postgres_large_buffers, k_pg_large_buf_max_size);
 // Emit a large buffer event for Postgres protocol.
 // The return value is used to control the flow for this specific protocol.
 // -1: wait additional data; 0: continue, regardless of errors.
-static __always_inline int postgres_send_large_buffer(tcp_req_t *req,
-                                                      pid_connection_info_t *pid_conn,
-                                                      const void *u_buf,
-                                                      u32 bytes_len,
-                                                      u8 direction,
-                                                      enum large_buf_action action) {
+static __always_inline int postgres_send_large_buffer(
```
rafaelroquetto, line 53: disclaimer: this was clang-format's doing after I removed `pid_conn` which was unused

bpf/gotracer/go_nethttp.c
```diff
@@ -930,12 +929,7 @@ int obi_uprobe_http2WriteHeaders(struct pt_regs *ctx) {
 
     bpf_dbg_printk("=== uprobe/proc http2WriteHeaders === ");
 
-    setup_http2_client_conn(goroutine_addr,
-                            cc_ptr,
-                            (u32)stream_id,
-                            _cc_tconn_pos,
-                            _cc_next_stream_id_pos,
-                            _cc_framer_pos);
+    setup_http2_client_conn(goroutine_addr, cc_ptr, (u32)stream_id, _cc_tconn_pos, _cc_framer_pos);
```
rafaelroquetto, line 932: ditto

bpf/.clang-tidy
```diff
@@ -12,7 +12,6 @@ Checks:
   - '-bugprone-narrowing-conversions' #TODO this needs to be checked/enabled
   - 'clang-analyzer-*'
   - '-clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling'
-  - '-misc-unused-parameters'
```
rafaelroquetto, line 15: this is actually enabling the check (I am removing a mask that was explicitly disabling it)

PR #399, opened 2025-07-31
[Feature] GPU/CUDA memory copy events

grcevski:
This PR finalises the base GPU CUDA instrumentation for OBI. The work done here will be presented at KubeCon India 2025 and later at SRECon EU 2025 (in Dublin).

The code for this PR was left in a [branch in my personal repo](https://github.com/grcevski/ebpf-autoinstrument/commits/gpu_context/) before OBI was donated and it was never merged. Essentially, what we currently have in OBI is tracking CUDA kernel launches and CUDA memory allocations. This PR extends that support with CUDA memory copies from and to the GPU, which are more important than the memory allocations. LLMs typically allocate all the GPU memory at load times, so the actual memory movement while running the LLM workload is actually more represented by the memory moves ON and OFF GPU.

## So what's this GPU support about?

Two recently published papers raise the concern of "bubbles" in the GPU kernel execution when running LLM models. With the baseline models, it's really difficult to achieve full GPU saturation while running inference. In fact, as the research papers point out, the typical utilisation for high end GPU models can be as low as 40%.

1. From Stanford: https://hazyresearch.stanford.edu/blog/2025-05-27-no-bubbles
2. From CMU: https://zhihaojia.medium.com/compiling-llms-into-a-megakernel-a-path-to-low-latency-inference-cf7840913c17 

## How does OBI fit into this picture?

If we take GPU metrics directly from an NVIDIA GPU (in my case a very low end model - 3090), running Alibaba Qwen3 with the default Pytorch mode, we observe the following GPU utilisation:

Screenshot from 2025-07-30 19-46-50

As we can see, my low end GPU is only 60% saturated and we can't really tell why. Am I not pushing sufficient work? Is there's something inherently wrong with the baseline model, which is preventing me to use all of my GPU capacity.

So if we now fire up OBI on this model, with it's model agnostic/framework agnostic CUDA GPU instrumentation, we see the following result:

Screenshot from 2025-07-30 19-37-39

If we dissect the dashboards created from these metrics we can notice a few things:
1. We see 100s of kernels being launched a second.
2. We see see memory copy from the device to the host in great amounts, which is very unexpected, we are not getting a lot of data back from the GPU with the query I supplied.
3. We see the GPU grid sizes for computation all under 1K, which way below the theoretical limit of 65K even for one dimension.

All of this data points to very inefficient GPU work scheduling, and this is where the tools mentioned in the two papers above come into play. In my example, I used Mirage from CMU to optimise my Qwen3 Pytorch workload. My optimised model with Mirage has twice as good latency on average!

We can now use OBI again to verify the outcome of the optimisation. Did it do what we expected, or there's something else at play?

Screenshot from 2025-07-30 19-38-29

We notice that few metrics have dramatically changed:
1. There are only handful of kernel launches and actually less than 1 per second, which suggests I can push my workload now much higher.
2. Memory copies are dominated to in between device copies, which is what we want to see.
3. GPU is at 100% utilisation when active!
4. The Grid sizes are 3 times larger.

We can now confirm that we are efficiently using our GPU and it confirms the improvements in the inference latency -  4x improved at the 95% and 2x at the 50%.

### DIFF
--- a/bpf/gpuevent/gpuevent.c
+++ b/bpf/gpuevent/gpuevent.c
@@ -20,9 +20,11 @@ char LICENSE[] SEC("license") = "Dual MIT/GPL";
 
 const gpu_kernel_launch_t *unused_gpu __attribute__((unused));
 const gpu_malloc_t *unused_gpu1 __attribute__((unused));
+const gpu_memcpy_t *unused_gpu2 __attribute__((unused));
 
 #define EVENT_GPU_KERNEL_LAUNCH 1
 #define EVENT_GPU_MALLOC 2
+#define EVENT_GPU_MEMCPY 3
 
 struct {
     __uint(type, BPF_MAP_TYPE_RINGBUF);
@@ -120,6 +122,31 @@ int BPF_KPROBE(handle_cuda_malloc, void **devPtr, size_t size) {
     e->flags = EVENT_GPU_MALLOC;
     task_pid(&e->pid_info);
     e->size = (u64)size;
+
+    bpf_ringbuf_submit(e, 0);
+    return 0;
+}
+
+SEC("uprobe/cudaMemcpyAsync")
+int BPF_KPROBE(handle_cuda_memcpy, void *dst, void *src, size_t size, u8 kind) {
+    u64 id = bpf_get_current_pid_tgid();
+
+    if (!valid_pid(id)) {
+        return 0;
+    }
+
+    bpf_dbg_printk("=== cudaMemcpyAsync %llx ===", id);
+
+    gpu_memcpy_t *e = bpf_ringbuf_reserve(&rb, sizeof(*e), 0);
+    if (!e) {
+        bpf_dbg_printk("Failed to allocate ringbuf entry");
+        return 0;
+    }
+
+    e->flags = EVENT_GPU_MEMCPY;
+    task_pid(&e->pid_info);
+    e->size = (u64)size;
+    e->kind = kind;
 
     bpf_ringbuf_submit(e, 0);
     return 0;
--- a/bpf/gpuevent/gpuevent.c
+++ b/bpf/gpuevent/gpuevent.c
@@ -124,7 +124,7 @@ int BPF_KPROBE(handle_cuda_malloc, void **devPtr, size_t size) {
 
     e->flags = EVENT_GPU_MALLOC;
     task_pid(&e->pid_info);
-    e->size = (u64)size;
+    e->size = (s64)size;
 
     bpf_ringbuf_submit(e, 0);
     return 0;
@@ -152,7 +152,7 @@ int BPF_KPROBE(handle_cuda_memcpy, void *dst, void *src, size_t size, u8 kind) {
 
     e->flags = EVENT_GPU_MEMCPY;
     task_pid(&e->pid_info);
-    e->size = (u64)size;
+    e->size = (s64)size;
     e->kind = kind;
 
     bpf_ringbuf_submit(e, 0);
--- a/bpf/gpuevent/gpuevent.c
+++ b/bpf/gpuevent/gpuevent.c
@@ -105,6 +105,9 @@ int BPF_KPROBE(handle_cuda_launch,
 
 SEC("uprobe/cudaMalloc")
 int BPF_KPROBE(handle_cuda_malloc, void **devPtr, size_t size) {
+    (void)ctx;
+    (void)devPtr;
+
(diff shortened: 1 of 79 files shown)


pkg/components/exec/file.go
```diff
@@ -29,9 +29,10 @@ type FileInfo struct {
 }
 
 const (
-	envServiceName   = "OTEL_SERVICE_NAME"
```
grcevski: Ignore this change and file_test.go, I'll merge with main once that PR is merged. I needed it in my branch for local testing and it ended up here.

pkg/app/request/metric_attributes.go
```diff
@@ -183,3 +183,22 @@ func PeerAsClient(span *Span) string {
 func CudaKernel(val string) attribute.KeyValue {
 	return attribute.Key(attr.CudaKernelName).String(val)
 }
+
+func CudaMemcpyName(val int) string {
+	switch val {
+	case 0:
```
rafaelroquetto: I don't know much (or anything really) about CUDA, should these magic numbers be an enum?

pkg/components/ebpf/gpuevent/gpuevent.go
```diff
@@ -274,6 +284,27 @@ func (p *Tracer) readGPUMallocIntoSpan(record *ringbuf.Record) (request.Span, bo
 	}, false, nil
 }
 
+func (p *Tracer) readGPUMemcpyIntoSpan(record *ringbuf.Record) (request.Span, bool, error) {
+	var event GPUMemcpyInfo
+	if err := binary.Read(bytes.NewReader(record.RawSample), binary.LittleEndian, &event); err != nil {
```
rafaelroquetto: Ideally, we should use `ReinterpretCast` here, which replaced `binaryRead` as it was ultra-slow (because it relies on reflection).

But I see how you are just being consistent with the rest of the file - so up to you if you want to extend this PR and fix it altogether, or we can do it in a separate PR.

rafaelroquetto (approved):
Cuda this be any better? :) Looks great!

pkg/app/request/metric_attributes.go
grcevski: Good point! I think we should make them enum.

pkg/components/ebpf/gpuevent/gpuevent.go
grcevski: Good catch, I will change that to make it work like the other code.

bpf/gpuevent/gpuevent.c
```diff
@@ -127,3 +129,32 @@ int BPF_KPROBE(handle_cuda_malloc, void **devPtr, size_t size) {
...
+    (void)dst;
+    (void)src;
+
+    u64 id = bpf_get_current_pid_tgid();
+
+    if (!valid_pid(id)) {
+        return 0;
+    }
+
+    bpf_dbg_printk("=== cudaMemcpyAsync %llx ===", id);
```
skl: Maybe worth adding `kind` to the log?

pkg/components/ebpf/gpuevent/gpuevent.go
```diff
@@ -274,6 +284,27 @@ func (p *Tracer) readGPUMallocIntoSpan(record *ringbuf.Record) (request.Span, bo
...
+		return request.Span{}, true, err
+	}
+
+	// Log the GPU Kernel Launch event
+	p.log.Debug("GPU Memcpy", "event", event)
+
+	return request.Span{
+		Type:          request.EventTypeGPUMemcpy,
+		ContentLength: int64(event.Size),
```
skl: Assuming memory < 9 exabytes 😄 (unit64 --> int64)

bpf/gpuevent/gpuevent.c
grcevski: good idea.

pkg/components/ebpf/gpuevent/gpuevent.go
grcevski: I suppose I can look into changing content length to be uint64, I don't think it will affect anything

bpf/gpuevent/gpuevent.c
grcevski: done

pkg/components/ebpf/gpuevent/gpuevent.go
grcevski: done. I ended up respecting the size type from eBPF, int64 so we don't do these conversions.

PR #401, opened 2025-08-01
Fix package imports' ordering

mariomac:
The previous goimports-based package ordering was not behaving as expected, as many automatically-added imports (e.g. by the IDE) weren't properly ordered if they were inserted in their own group (separated by a line from the rest).

Replacing goimports by `gci` in the golanci-lint, we get a more consistent order of imports in 4 separate groups:
        - standard packages
        - generic 3rd party packages
        - OpenTelemetry packages
        - Local OBI packages

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,19 +7,21 @@ issues:
 
 formatters:
   enable:
+    - gci
     - gofumpt
-    - goimports
 
   settings:
     gofmt:
       # simplify code: gofmt with `-s` option, true by default
       simplify: true
 
-    goimports:
-      # put imports beginning with prefix after 3rd-party packages;
-      # it's a comma-separated list of prefixes
-      local-prefixes:
-        - go.opentelemetry.io/obi
+    gci:
+      sections:
+        - standard
+        - default
+        - prefix(github.com/open-telemetry)
+        - prefix(go.opentelemetry.io)
+        - localmodule
 
 linters:
   exclusions:
--- a/configs/offsets/otelsdk/inspect.go
+++ b/configs/offsets/otelsdk/inspect.go
@@ -7,7 +7,6 @@ import (
 	"context"
 
 	"go.opentelemetry.io/otel"
-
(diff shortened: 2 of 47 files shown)


rafaelroquetto (commented):
Thanks for cleaning this up! :soap:

MrAlias:
The common standard for Go is what we have using `goimports` where there are only three import sections (standard lib, 3rd-party, and local). I think keeping that standard is going to better align with most development environments and the rest of OTel repositories.

mariomac:
@MrAlias yeah, I'm able to configure my IDE to follow that convention but since we can't make sure that all the contributors have it, from time to time some imports get unordered. That's why I added the check. However it seems that `goimports` as configured in `golangci-lint` doesn't enforce correctly the imports order.

On the other side, I'm happy removing the `otel` sections to follow the common OTEL standard of `std - 3rd party - local`

.golangci.yml
```diff
@@ -7,19 +7,21 @@ issues:
 
 formatters:
   enable:
+    - gci
     - gofumpt
-    - goimports
 
   settings:
     gofmt:
```
MrAlias, line 14: Setting `extra-rules: true` for the `gofumpt` formater will enforce imports using `goimports` when golangci-lint is run

```suggestion
    gofumpt:
      extra-rules: true
    gofmt:
```

https://golangci-lint.run/usage/formatters/#gofumpt

MrAlias:
> @MrAlias yeah, I'm able to configure my IDE to follow that convention but since we can't make sure that all the contributors have it, from time to time some imports get unordered. That's why I added the check. However it seems that `goimports` as configured in `golangci-lint` doesn't enforce correctly the imports order.
> 
> On the other side, I'm happy removing the `otel` sections to follow the common OTEL standard of `std - 3rd party - local`

The issue with using an alternate import formatting approach than the standard here is that it is almost certainly likely that developers will not be using `gci`. This may work for the small group that are using this formater, but the larger majority that use the defaults from the Go toolchain are going to now fight with the CI system.

PR #402, opened 2025-08-01
Optimize ExportModes

mariomac:
`ExportModes.CanExportMetrics` is invoked for each Span. I slightly optimized it by changing its internal representation by a bits map.

Now, instead of having to iterate a short slice and perform 1-2 string comparisons for each span, it just checks a bit in an integer.

The YAML marshaling/unmarshaling remained the same, as well as the rest of the public API.

### DIFF
--- a/pkg/services/export.go
+++ b/pkg/services/export.go
@@ -1,0 +1,105 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package services
+
+import (
+	"fmt"
+
+	"gopkg.in/yaml.v3"
+
+	"go.opentelemetry.io/obi/pkg/components/helpers/maps"
+)
+
+const (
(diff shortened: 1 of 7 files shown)


pkg/services/export.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+type ExportModes struct {
+	items maps.Bits
+}
+
+func (modes *ExportModes) canExport(mode maps.Bits) bool {
+	if modes == nil {
+		return true
+	}
```
mariomac, line 31: we can even get rid of this null check (and save an indirection) if we don't use pointers and, instead of internally storing a bit for each type of signal we can export, we negate the bits and store the bit for the signal we can't export, so the undefined, zero-value for ExportModes would allow exporting all the signals.

But probably we don't need this level of optimization

grcevski (approved):
Nice! LGTM!

PR #409, opened 2025-08-05
Fix env var substitution for OpenTelemetry Operator

marctc:
We support OTEL env variable replacement but we don't have support for variables
of the format $(var), which is how OpenTelemetry Operator works.

Fixes https://github.com/grafana/beyla/issues/1856


grcevski (approved):
LGTM!

PR #410, opened 2025-08-05
bpf: infer packet type based on server port

mmat11:
The current large buffer implementation, saves buffers per _direction_. In userspace we then try to parse request/responses
based on packet data.
This changes aims to infer packet type (request/response) and do more precise parsing once the buffers get to userspace.

### DIFF
--- a/pkg/components/ebpf/generictracer/generictracer.go
+++ b/pkg/components/ebpf/generictracer/generictracer.go
@@ -304,10 +304,6 @@ func (p *Tracer) KProbes() map[string]ebpfcommon.ProbeDesc {
 			Start:    p.bpfObjects.ObiKprobeUnixStreamSendmsg,
 			End:      p.bpfObjects.ObiKretprobeUnixStreamSendmsg,
 		},
-		"inet_csk_accept": {
-			Required: true,
-			Start:    p.bpfObjects.ObiKprobeInetCskAccept,
-		},
 		"inet_csk_listen_stop": {
 			Required: true,
 			Start:    p.bpfObjects.ObiKprobeInetCskListenStop,
--- a/bpf/generictracer/k_tracer.c
+++ b/bpf/generictracer/k_tracer.c
@@ -117,12 +117,9 @@ int BPF_KRETPROBE(obi_kretprobe_sys_accept4, s32 fd) {
     (void)ctx;
 
     u64 id = bpf_get_current_pid_tgid();
-
     if (!valid_pid(id)) {
         return 0;
     }
-
-    //bpf_dbg_printk("=== accept 4 ret id=%d ===", id);
 
     bpf_dbg_printk("=== accept 4 ret id=%d, fd=%d ===", id, fd);
 
@@ -137,6 +134,11 @@ int BPF_KRETPROBE(obi_kretprobe_sys_accept4, s32 fd) {
         bpf_dbg_printk("No accept sock info %d", id);
         goto cleanup;
     }
+
+    struct socket *sock = (struct socket *)args->addr;
+    struct sock *sk = BPF_CORE_READ(sock, sk);
+    struct sock_port_ns np = sock_port_ns_from_sk(sk);
+    bpf_map_update_elem(&listening_ports, &np, &(bool){true}, BPF_ANY);
 
     ssl_pid_connection_info_t info = {};
 
@@ -1050,20 +1052,6 @@ int obi_handle_buf_with_args(void *ctx) {
         }
     }
 
-    return 0;
-}
-
-SEC("kprobe/inet_csk_accept")
-int BPF_KPROBE(obi_kprobe_inet_csk_accept, struct sock *sk) {
-    (void)ctx;
-
-    u64 id = bpf_get_current_pid_tgid();
-    (void)id;
-
-    bpf_dbg_printk("=== inet_csk_accept id=%d ===", id);
-
-    struct sock_port_ns np = sock_port_ns_from_sk(sk);
-    bpf_map_update_elem(&listening_ports, &np, &(bool){true}, BPF_ANY);
     return 0;
 }
 


bpf/generictracer/protocol_mysql.h
```diff
@@ -190,7 +190,8 @@ static __always_inline int mysql_send_large_buffer(tcp_req_t *req,
     req->has_large_buffers = true;
     bpf_ringbuf_output(&events,
                        large_buf,
-                       (sizeof(tcp_large_buffer_t) + written) & k_mysql_large_buf_max_size_mask,
+                       (sizeof(tcp_large_buffer_t) + sizeof(void *) + written) &
```
mmat11: Note for the reviewer: I had to add `sizeof(void *)` here because the sizeof() the type in C and in Go don't match because of the `buf` field. This was an issue only in the case the written data was less than 8 bytes, practically almost never

bpf/common/accepting_port.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
```
rafaelroquetto: don't forget to add the copyright headers

bpf/common/accepting_port.h
```diff
@@ -0,0 +1,23 @@
...
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_core_read.h>
+
+struct accepting_port {
+    u32 netns;
+    u16 port;
+    u8 _pad[2];
+};
+
+static __always_inline struct accepting_port accepting_port_from_skc(struct sock_common *skc) {
```
rafaelroquetto:
```suggestion
static __always_inline struct accepting_port accepting_port_from_skc(const struct sock_common *skc) {
```

if possible

bpf/common/accepting_port.h
```diff
@@ -0,0 +1,23 @@
...
+    struct accepting_port ap = {
+        .port = BPF_CORE_READ(skc, skc_num),
+        .netns = BPF_CORE_READ(skc, skc_net.net, ns.inum),
+    };
+    return ap;
+}
+
+static __always_inline struct accepting_port accepting_port_from_sk(struct sock *sk) {
```
rafaelroquetto: ditto

bpf/generictracer/iter_tcp.c
```diff
@@ -0,0 +1,28 @@
...
+#include <bpfcore/bpf_core_read.h>
+#include <bpfcore/bpf_helpers.h>
+
+#include <common/accepting_port.h>
+
+#include <generictracer/maps/accepting_ports.h>
+
+SEC("iter/tcp")
+int obi_iter_tcp(struct bpf_iter__tcp *ctx) {
+    struct seq_file *seq = ctx->meta->seq;
```
rafaelroquetto: nit: move close to usage site around line 25

rationale: improves readability by limiting scope

bpf/generictracer/iter_tcp.c
```diff
@@ -0,0 +1,28 @@
...
+    struct sock_common *skc = ctx->sk_common;
+    if (!skc) {
+        return 0;
+    }
+
+    unsigned char skc_state = BPF_CORE_READ(skc, skc_state);
```
rafaelroquetto:
```suggestion
    const unsigned char skc_state = BPF_CORE_READ(skc, skc_state);
```

bpf/common/accepting_port.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_core_read.h>
+
+struct accepting_port {
```
rafaelroquetto: could we rename this to `struct sock_port_ns`? this code could in theory be applied to any socket. The following function could therefore be called `sock_port_ns_from_skc(...)` as it's what it does (as well as this filename)

bpf/generictracer/iter_tcp.c
```diff
@@ -0,0 +1,28 @@
...
+    if (skc_state != TCP_LISTEN) {
+        return 0;
+    }
+
+    struct accepting_port ap = accepting_port_from_skc(skc);
+    bpf_map_update_elem(&accepting_ports, &ap, &(bool){true}, BPF_ANY);
```
rafaelroquetto: Could we rename this map to `listening_ports` ? Unlike "accepting", "listening" is common jargon (as exemplified above in the `TCP_LISTEN` constant), and conveys best the nature of this value, unlike "accepting port".

bpf/generictracer/maps/accepting_ports.h
```diff
@@ -0,0 +1,14 @@
...
+
+#include <common/accepting_port.h>
+#include <common/map_sizing.h>
+
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, struct accepting_port);
+    __type(value, bool);
+    __uint(max_entries, MAX_CONCURRENT_REQUESTS);
+} accepting_ports SEC(".maps");
```
rafaelroquetto:
```suggestion
} listening_ports SEC(".maps");
```

bpf/generictracer/protocol_common.h
```diff
@@ -20,6 +22,32 @@
 
 volatile const s32 capture_header_buffer = 0;
 
+static __always_inline bool is_accepting(const u16 port, const u32 netns) {
+    struct accepting_port ap = {
```
rafaelroquetto:
```suggestion
    const struct accepting_port ap = {
```

pkg/components/ebpf/generictracer/generictracer.go
```diff
@@ -526,6 +553,42 @@ func (p *Tracer) watchForMisclassifedEvents(ctx context.Context) {
...
+		_, err := rd.Read(buf)
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return fmt.Errorf("read iterator: %w", err)
+		}
+	}
+
+	for _, line := range strings.Split(unix.ByteSliceToString(buf), "\n") {
```
rafaelroquetto: Tip: use pkg/components/split/splititerator.go here instead to avoid allocations caused by `strings.Split()`

Having said that, perhaps instead of using `bpf_seq_printf`, we should use `bpf_seq_write` and simply write a 64-bit payload containing the port and the namespace (i.e. basically the accepting_port struct). Then you don't need the split loop, instead just a debug print loop iterating on `buf`.

bpf/generictracer/protocol_mysql.h
rafaelroquetto: shouldn't we be writing `max(sizeof(void*), written)` here? otherwise, if written is actually 9, we'll be writing  17 instead.

pkg/components/ebpf/instrumenter.go
```diff
@@ -432,6 +433,28 @@ func (i *instrumenter) tracepoint(funcName string, programs ebpfcommon.ProbeDesc
...
+		})
+		if err != nil {
+			if i.metrics != nil {
+				i.metrics.InstrumentationError(i.processName, imetrics.InstrumentationErrorAttachingIter)
+			}
+			return fmt.Errorf("attaching iterator %s: %w", name, err)
+		}
+		iter.Link = lnk
+
+		p.(*generictracer.Tracer).IterLinks[name] = lnk
```
rafaelroquetto: This doesn't look right,  the instrumenter shouldn't/cannot have a dependency on a tracer. Are you sure you need `name` here altogether? IMHO `Iters()` should return `[]ebpfcommon.Iter`, not a map - it's then up to the tracer implementation to deal with its iterators afterwards.

pkg/components/ebpf/tracer.go
```diff
@@ -102,6 +102,9 @@ type Tracer interface {
 	// SockOps returns a list of programs that need to be loaded as a
 	// BPF_PROG_TYPE_SOCK_OPS eBPF programs
 	SockOps() []ebpfcommon.SockOps
+	// Iters returns a list of programs that need to be loaded as a
+	// BPF_PROG_TYPE_TRACING with BPF_TRACE_ITER attach type
+	Iters() map[string]ebpfcommon.Iter
```
rafaelroquetto:
```suggestion
	Iters() []*ebpfcommon.Iter
```

You don't techinically need name here as far as I understand the code, but if you must, embed it as a "desc" variable or soemthing like that inside the `ebpfcommon.Iter` type itself.

pkg/components/ebpf/generictracer/generictracer.go
```diff
@@ -46,6 +50,8 @@ type Tracer struct {
 	ingressFilters   map[ifaces.Interface]*netlink.BpfFilter
 	instrumentedLibs ebpfcommon.InstrumentedLibsT
 	libsMux          sync.Mutex
+
+	IterLinks map[string]link.Link // TODO: tmp solution, add helpers to tracer interface?
```
rafaelroquetto: here you could simply keep an explicit pointer to your iterator, not a map, then see below.

pkg/components/ebpf/generictracer/generictracer.go
```diff
@@ -382,6 +397,14 @@ func (p *Tracer) SockMsgs() []ebpfcommon.SockMsg { return nil }
 
 func (p *Tracer) SockOps() []ebpfcommon.SockOps { return nil }
 
+func (p *Tracer) Iters() map[string]ebpfcommon.Iter {
+	return map[string]ebpfcommon.Iter{
+		"obi_iter_tcp": {
+			Program: p.bpfObjects.ObiIterTcp,
+		},
+	}
+}
```
rafaelroquetto:
```suggestion
func (p *Tracer) Iters() []*ebpfcommon.Iter {
	return []*ebpfcommon.Iter{
	    p.obiIterTcp
	}
}
```

and then here you just return that

pkg/components/ebpf/generictracer/generictracer.go
```diff
@@ -437,6 +460,10 @@ func (p *Tracer) Run(ctx context.Context, ebpfEventContext *ebpfcommon.EBPFEvent
 	go p.lookForTimeouts(ctx, timeoutTicker, eventsChan)
 	defer timeoutTicker.Stop()
 
+	if err := p.runIterator("obi_iter_tcp"); err != nil {
```
rafaelroquetto: finally, you could do something like:
```suggestion
	if err := p.runIterator(p.obiIterTcp); err != nil {
```

pkg/components/ebpf/generictracer/generictracer.go
```diff
@@ -526,6 +553,42 @@ func (p *Tracer) watchForMisclassifedEvents(ctx context.Context) {
 	}
 }
 
+func (p *Tracer) runIterator(name string) error {
```
rafaelroquetto:
```suggestion
func (p *Tracer) runIterator(iter *ebpfcommon.Iter) error {
```

pkg/components/ebpf/generictracer/generictracer.go
```diff
@@ -526,6 +553,42 @@ func (p *Tracer) watchForMisclassifedEvents(ctx context.Context) {
...
+	p.log.Debug("Running iterator", "name", name)
```
rafaelroquetto:
```suggestion
	p.log.Debug("Running iterator", "name", iter.name)
```

pkg/components/ebpf/generictracer/generictracer.go
```diff
@@ -526,6 +553,42 @@ func (p *Tracer) watchForMisclassifedEvents(ctx context.Context) {
...
+
+	lnk, ok := p.IterLinks[name]
+	if !ok {
+		return fmt.Errorf("iterator %s not found", name)
+	}
```
rafaelroquetto:
```suggestion
```

bpf/generictracer/iter_tcp.c
```diff
@@ -0,0 +1,28 @@
...
+
+    BPF_SEQ_PRINTF(seq, "Adding accepting port:%d, netns:%d\n", ap.port, ap.netns);
```
rafaelroquetto: how can I visualise this message? Wouldn't that require bpffs?

grcevski:
> Good point, I remember doing some experiments locally on these attach points and they were the only ones called exactly once, but I don't remember having tried it with `accept4`, so I will test it tomorrow

It might actually be better to use what you have here, since your probe on accept is a krpobe, not kretprobe, but we need the file descriptor for other purposes in accept4. kretprobes are less reliable, so maybe your approach is better, we just need to move the code if possible from the accept4 here.

bpf/generictracer/iter_tcp.c
rafaelroquetto: Nevermind, I understand this now (and see it in the logs).

bpf/common/accepting_port.h
mmat11: In the C files those are very inconsistent (some files have the full apache license, some others nothing, some the otel copyright + spdx identifier), should I open an issue to also track these files when running the license checker? (see: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/Makefile#L361)

bpf/generictracer/protocol_mysql.h
mmat11: yes I think you are right

pkg/components/ebpf/generictracer/generictracer.go
mmat11: I could do that, but my idea was to have a generic `runIterator()` function we could run from userspace without caring about what would be written in the iterator file. If we go this road we need to have N functions for the future iterators which take into account the structure of the type that is written into the file.. and I don't think it's worth it given the fact this will run once at process startup.
We could optimize this if wel'll need an iterator which needs to run periodically (and in that case we should implement a `resetIterator()` or similar)

rafaelroquetto: I see what you mean, but I am not sure I understand the generic part of the argument - isn't  `runIterator()` private to generictracer already? And in this case, all this function does is to print a debug message, right? Honest question, how are you planning to generalise this?

I don't really see a problem of having N functions for future iterators - if each function does something different anyway, then it's best they are different and encapsulating their own functionality. Or how would a generalised version of `runIterator()` work?

mmat11: It's already generalised as it is now, the only "contract" with the ebpf program is that one iteration will write a line ended by '\n'. Most changes we will do is just move the function out of `generictracer` to some higher level struct accessible by other tracers.

If we save "typed data" instead, we have to repeat this function for every iterator - which is also fine - for now that we have one.

> if each function does something different anyway

I would avoid that unless we need it (eg. for every iterated object we need to proactively do something in userspace). The common parts of all iterators should be: get the link -> open the iter file associated to the link -> consume it till EOF

rafaelroquetto: > The common parts of all iterators should be: get the link -> open the iter file associated to the link -> consume it till EOF

Ok, understood. How would that be generalised, though? Would we pass a callback to handle each line? I'm just trying to visualise how this will actually look. Please don't get me wrong, it's just that at the current state we have _neither a full generalised "API" nor a simple/lean alternative, but rather a mix of both_ .

I have reservations on settling on a string/line-based API as it enforces parsing/serialisation/deserialisation onto the handler. For instance, check out this probe: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/1e58b846b3421568478bb8bb695e8ed31d8e4c48/bpf/generictracer/nodejs.c#L19 

In the example above, unfortunately here there's no way out, numbers need to be encoded as strings, but imposes some decoding/handling back to integral types that wouldn't be the case (or desirable) if there were another way, massively simplified the code.

I think this warrants further discussion that goes beyond the scope of this PR, but perhaps we don't have to choose between "no abstraction" / raw code vs "one size fits all line based framing". Instead, we should simply implement the `io.Reader` interface. Then user-space gets the best of both worlds whilst being able to rely on everything that plugs into `io.Reader` - for instance, here one could use `bufio.Reader()` and then call `reader.ReadString("\n")` to read lines, or call `Read()` directly to read bytes, etc: _full flexibility!_  IMHO this is the way to go if the plan is to generalise, as we don't have to do manual string splitting ourselves, and it provides all the mechanism without imposing any policy on the programmer.

rafaelroquetto: It just occurred to me that `ebpf.Iter` already implements the Reader interface (it was right there under my nose in the `Read()` call in the code).

Couldn't we simply leverage that together with `bufio`? it already handles internal buffering anyway, so this function becomes trivial, in either case with `ReadString('\n')` to read the lines, or the raw `Read()` to read an 8 bit payload.

pkg/components/ebpf/instrumenter.go
mmat11: The issue is that here we need to work with the link _after_ it has been loaded so we need some way for the instrumenter to store the link to be used by the tracer. I added another function to the interface

rafaelroquetto: But if iter is a pointer, why doesn't `iter.Link = lnk` suffice?

mmat11: because `Iters()` will return new pointers every time it's called, so we need to set something in the tracer anyway

rafaelroquetto: I mean, it doesn't have to. Just store the iterators in the generic tracer instance, i.e. 

```
func (p *Tracer) Iters() []*ebpfcommon.Iter() {
   return p.Iters
}
```

this keeps the semantics straightforward and simplifies it by doing without `UpdateIteratorLink` altogether.

mmat11: ok I tried to do this and it's effectively less ugly than what I did earlier (but inconsistent with the rest of the methods :D) so if it looks good to you, it's also good to me

pkg/components/ebpf/generictracer/generictracer.go
```diff
@@ -382,6 +396,18 @@ func (p *Tracer) SockMsgs() []ebpfcommon.SockMsg { return nil }
 
 func (p *Tracer) SockOps() []ebpfcommon.SockOps { return nil }
 
+func (p *Tracer) Iters() []*ebpfcommon.Iter {
+	if len(p.iters) == 0 {
```
rafaelroquetto, line 396: cool, lazy-initialisation! :)

rafaelroquetto (dismissed):
LGTM - thank you so much for the changes!

grcevski:
> > It might actually be better to use what you have here, since your probe on accept is a krpobe, not kretprobe, but we need the file descriptor for other purposes in accept4. kretprobes are less reliable, so maybe your approach is better, we just need to move the code if possible from the accept4 here.
> 
> @grcevski I had a look and from what I could see, this would be mostly an optimization, right? Do you mind if I do this in a follow-up PR?

Yeah that's fine, but I'd like to avoid adding an additional kprobe if the work in `SEC("kprobe/inet_csk_accept")` can be done in `sys_accept4` return probe. Kprobes cost always, the less we have them the better. The point is just do what you need to do in that handler.

rafaelroquetto (commented):
Oh, I think Nikola has a point here, I missed that - @mmat11 do you think it's possible to use a single kprobe (sys_accept4) as he's proposing? Might be best to do it on this PR, there's no point on having an intermediate "commit" that is bound to be refactored anyway.

mmat11:
@grcevski @rafaelroquetto I looked into it and it might no be as immediate as I thought it would be. The reason is that if we want to merge them, we need both the _fd_ and the _sk_ args, while `accept`/`accept4` have the _fd_ , `inet_csk_accept` doesn't (and the other way around, the _sk_ is not present in `accept`/`accept4` args - I guess I could get it from `sock_args_t` but I am not sure on who populates the `active_accept_args` map and if the data is _always_ available).
Given all of this, I would keep the 2 probes distinct - unless there's an alternative way to get _sk_ in the accept functions (?)

grcevski:
> @grcevski @rafaelroquetto I looked into it and it might no be as immediate as I thought it would be. The reason is that if we want to merge them, we need both the _fd_ and the _sk_ args, while `accept`/`accept4` have the _fd_ , `inet_csk_accept` doesn't (and the other way around, the _sk_ is not present in `accept`/`accept4` args - I guess I could get it from `sock_args_t` but I am not sure on who populates the `active_accept_args` map and if the data is _always_ available). Given all of this, I would keep the 2 probes distinct - unless there's an alternative way to get _sk_ in the accept functions (?)

If you look at how we use sock_args_t, you can easily extract the sk* from there:

```
static __always_inline bool parse_accept_socket_info(sock_args_t *args, connection_info_t *info) {
    struct sock *s;

    struct socket *sock = (struct socket *)(args->addr);
    BPF_CORE_READ_INTO(&s, sock, sk);

    return parse_sock_info(s, info);
}
```

That map is setup by security_socket_accept, so you should have the same information.

mmat11:
> That map is setup by security_socket_accept, so you should have the same information.

Then it should work fine as `security_socket_accept` is called just before `inet_csk_accept` 👍🏻, thanks. I'll use the current accept4 program then.

PR #417, opened 2025-08-06
[Route Classifier] Remove additional valid chars config option

jcreixell:
Adds a lookup table for performance reasons.

Benchmark results over 60s:

```
> table lookup

BenchmarkClusterURLWithoutCache-20      21180284              3417 ns/op            2000 B/op         59 allocs/op

> loop 

BenchmarkClusterURLWithoutCache-20      20406266              3387 ns/op            2000 B/op         59 allocs/op

> OR chain

BenchmarkClusterURLWithoutCache-20      21681082              3326 ns/op            2000 B/op         59 allocs/op
```

### DIFF
--- a/pkg/components/transform/route/clusterurl/cluster.go
+++ b/pkg/components/transform/route/clusterurl/cluster.go
@@ -15,9 +15,10 @@ import (
 )
 
 type ClusterURLClassifier struct {
-	classifier *structs.GibberishData
-	cache      *lru.Cache[string, bool]
-	cfg        *Config
+	classifier     *structs.GibberishData
+	cache          *lru.Cache[string, bool]
+	cfg            *Config
+	validCharTable [256]bool
 }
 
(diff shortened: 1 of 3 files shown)


pkg/components/transform/route/clusterurl/cluster.go
```diff
@@ -142,7 +142,7 @@ func (csf *ClusterURLClassifier) isValid(c byte) bool {
 		return true
 	}
 
-	for _, ac := range DefaultConfig().AdditionalValidChars {
+	for _, ac := range csf.cfg.AdditionalValidChars {
```
grcevski: Looking at this now, I think we should remove this option. Adding another nested loop here is going to be very slow.

grcevski (approved):
Interesting, well then we can bring back the loop :)

jcreixell:
@grcevski this is ready to go, feel free to merge if you are happy with it

PR #421, opened 2025-08-07
Add license headers to all C files

MrAlias:
Include C files (`*.c` and `*.h`) in the license-header-check make target to validate this in the CI and locally.


titaneric:
It's hard to believe that the rpc test failed again. I would take a look in these days

MrAlias:
> should we exclude the `bpf/bpfcore` directory?

If we are going to keep a copy of this in the repo and modify it, we need to add our copyright to the files as well as preserve what already exists.

mmat11:
> If we are going to keep a copy of this in the repo and modify it, we need to add our copyright to the files as well as preserve what already exists.

Except `utils.h` which should probably live somewhere else, I don't think any other header has been modified, so it's similar to the exception we make for `vendor/` - correct me if I am wrong

rafaelroquetto:
Just an addendum, some headers are auto-generated (such as vmlinux) and they've all been formatted thru clang-format, in case this is relevant.

MrAlias:
> > If we are going to keep a copy of this in the repo and modify it, we need to add our copyright to the files as well as preserve what already exists.
> 
> Except `utils.h` which should probably live somewhere else, I don't think any other header has been modified, so it's similar to the exception we make for `vendor/` - correct me if I am wrong

Gotcha. If that is the case I can look into moving `utils.h` and excluding that dir.

grcevski (approved):
LGTM!

PR #424, opened 2025-08-08
Stop using local `vendor` directory

MrAlias:
The only modified package discovered was for `golang.org/x/net/http2/hpack`. As far as I could tell, this was not used. It is removed along with all the other vendored code and its offsets are also removed.


MrAlias:
Working on test failure.

grcevski (approved):
LGTM!

Issue #420, opened 2025-08-07
TestYAMLMarshal_Exports/all_values fails intermittently


grcevski:
Here's an example of a CI job that fails:
https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/16792942991/job/47557948676?pr=419

It seems that maybe the checks are flaky because of the order the map is serialised.

Maybe if it's not easy to fix we can do a PR to disable this test.

CC @NimrodAvni78

PR #438, opened 2025-08-11
Fix yaml serialization test

grcevski:
The test TestYAMLMarshal_Exports frequently fails because the array serialisation can be unpredictable.

Fixes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/420

### DIFF
--- a/pkg/services/export_test.go
+++ b/pkg/services/export_test.go
@@ -45,7 +45,7 @@ func TestYAMLMarshal_Exports(t *testing.T) {
 		var exports struct {
 			Exports []string `yaml:"exports"`
 		}
-		yaml.Unmarshal(yamlOut, &exports)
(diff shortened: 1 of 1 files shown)


pkg/services/export_test.go
```diff
@@ -41,7 +41,12 @@ func TestYAMLMarshal_Exports(t *testing.T) {
 			Exports: ExportModes{blockSignal: ^(blockMetrics | blockTraces)},
 		})
 		require.NoError(t, err)
-		assert.YAMLEq(t, `exports: ["metrics", "traces"]`, string(yamlOut))
+
+		var exports struct {
+			Exports []string `yaml:"exports"`
+		}
+		yaml.Unmarshal(yamlOut, &exports)
```
MrAlias:
```suggestion
		require.NoError(t, yaml.Unmarshal(yamlOut, &exports))
```

PR #439, opened 2025-08-11
Fix number of other span metrics bugs

grcevski:
While looking at the span metrics generation code I found couple of other issues:

1. Prometheus export and OTel export were inconsistent on what they reported on AnySpanMetricsEnabled(). Graph wasn't in the list for OTel and it should've been, otherwise if someone enabled only graph metrics we wouldn't enable traces_target_info.
2. When OBI generates span metrics it also adds an attribute on any generated traces `span.metrics.skip=true` so the Collector can easily be configured to avoid double generating span metrics. However, the code currently checked only for OTel metrics exports, if Prometheus metrics exports the spans weren't marked with the same attribute.

Most of the code changes are tests to catch the edge cases discovered.

### DIFF
--- a/pkg/export/otel/otelcfg/config_metrics.go
+++ b/pkg/export/otel/otelcfg/config_metrics.go
@@ -121,7 +121,7 @@ func (m *MetricsConfig) EndpointEnabled() bool {
 }
 
 func (m *MetricsConfig) AnySpanMetricsEnabled() bool {
-	return m.SpanMetricsEnabled() || m.SpanMetricsSizesEnabled()
+	return m.SpanMetricsEnabled() || m.SpanMetricsSizesEnabled() || m.ServiceGraphMetricsEnabled()
 }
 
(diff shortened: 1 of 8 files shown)


pkg/export/otel/otelcfg/config_metrics.go
```diff
@@ -121,7 +121,7 @@ func (m *MetricsConfig) EndpointEnabled() bool {
 }
 
 func (m *MetricsConfig) AnySpanMetricsEnabled() bool {
-	return m.SpanMetricsEnabled() || m.SpanMetricsSizesEnabled()
+	return m.SpanMetricsEnabled() || m.SpanMetricsSizesEnabled() || m.ServiceGraphMetricsEnabled()
```
grcevski, line 124: This was recently introduced bug with the refactoring to make spangraphs generated separately. I've added a test now to ensure this doesn't happen anymore.

PR #444, opened 2025-08-12
Fix parent-based sampling

marctc:
The sampling functionality was orginally ported from tracing SDK but we missed the part of fetching and using
trace info from parent traces to use for sampling.

To better review this PR, look at the second commit, as the first one is just moving the tests that now should live in `tracesgen_test.go`


abhiroop93:
@marctc  May I ask why this was closed and not merged? This issue still persists right? The parentbased_traceidratio flag still does not work as expected, and if that is the case, the sampling will always be disjoint.

marctc:
The approach didn't work and it was wrong. I don't know if the issue still persists but i'd say so.

abhiroop93:
Understood. From what I see, the issue still persists, can you please have a look again at this bug? It is causing all services to sample 100% since it believes it's parent was sampled. Or is there any other approach we can take going forward? @marctc

PR #464, opened 2025-08-14
kafka: add 2.8 and 4.0 integration tests, increase apiVersion coverage, fix span names

mmat11:
- Fix kafka span name to follow opentelemetry semantic conventions (https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#span-name)
- Increase apiVersion coverage for _fetch_ and _produce_ requests
- Add some missing attributes for `EventTypeKafkaClient`
- Add some kafka integration tests: python, java (2.8.0), java (4.0.0)


test/integration/components/javakafka/Dockerfile
```diff
@@ -17,10 +17,10 @@ ENV JAVA_HOME /usr/lib64/graalvm/graalvm22-ce-java17
 WORKDIR /build
 
 # Copy the source code into the image for building
-COPY test/integration/components/javakafka/.mvn .mvn/
```
rafaelroquetto: out of curiosity, what was the issue here? I see some of our other tests also use full paths relative to the repository root, are they also affected?

mmat11: I don't think they are affected if the compose files are not complaining; I had to make the docker-compose contexts uniform for both oats and integration tests since this dockerfile is used by both

rafaelroquetto: I think there's something going on here though. Our tests tend to reference the full path `test/integration/components/...` instead - and if you look [here](https://github.com/coralogix/opentelemetry-ebpf-instrumentation/blob/52fb6a02fa4bd0c01e99a5ad12b5cd12cee805dd/test/oats/kafka/docker-compose-beyla-java-kafka.yml#L38), in the very same file we have a reference to another Dockerfile that uses a relative path to the repo root. So I would like to understand why we need to do this for this particular Dockerfile and if there's no other way to keep it consistent.

mmat11: > I think there's something going on here though. Our tests tend to reference the full path test/integration/components/... instead

Not all of them, it's a mix right now and it's inconsistent either way 🤷🏻‍♂️, there are Dockerfiles which use `.` as context and some others the root directory, changing the context or the path would still keep it inconsistent.

I am reverting it to use the full path and change the docker-compose integration test file, but it will still be inconsistent -- we would need to refactor all the docker/compose files

rafaelroquetto (approved):

PR #468, opened 2025-08-15
[chore,bpf] drop duplicate function

florianl:
Remove d_print_http_connection_info as duplicate of dbg_print_http_connection_info.


marctc (changes requested):
There are changes to github action

marctc (commented):
There's a GH action failing. Let me fix in another PR

rafaelroquetto:
Hey @florian! They are not equivalent - if you look closer, one uses `bpf_dbg_printk` whilst the other uses `bpf_d_printk`. `bpf_d_printk` does not send the logs to userspace, but rather only to `trace_pipe`, the reason is that we don't want to be that noisy in those points.

PR #469, opened 2025-08-15
Fix zizmor

marctc:


.github/workflows/check_gh_actions_security.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - name: Install the latest version of uv
         uses: astral-sh/setup-uv@d9e0f98d3fc6adb07d1e3d37f3043649ddad06a1 # v6.5.0
         with:
-          enable-cache: true
+          enable-cache: false
```
skl, line 28: I was wondering what the context of this change was, but then I saw the following message on another PR, so I assume it is to address this:

```
check_gh_actions_security.yml:27: runtime artifacts potentially vulnerable to a cache poisoning attack: opt-in for caching here
```

skl (approved):

PR #476, opened 2025-08-18
trace export internal metrics + BPF internal metrics

NimrodAvni78:
improve internal metrics
fix traces internal metrics not being reported
report bpf metrics as internal metrics

### DIFF
--- a/pkg/export/otel/metrics_internal.go
+++ b/pkg/export/otel/metrics_internal.go
@@ -35,6 +35,9 @@ type InternalMetricsReporter struct {
 	instrumentationErrors instrument.Int64Counter
 	avoidedServices       instrument.Int64Gauge
 	buildInfo             instrument.Int64Gauge
+	bpfProbeLatencies     instrument.Float64Histogram
+	bpfMapEntries         instrument.Int64Gauge
+	bpfMapMaxEntries      instrument.Int64Gauge
 }
 
 func imlog() *slog.Logger {
@@ -126,6 +129,45 @@ func NewInternalMetricsReporter(ctx context.Context, ctxInfo *global.ContextInfo
 		return nil, err
 	}
 
+	// TODO should it be ebpf like the others, or bpf like the original one?
+	bpfProbeLatencies, err := meter.Float64Histogram(
+		attr.VendorPrefix+".bpf.probe.latency_seconds",
+		instrument.WithDescription("Latency of the eBPF probe in seconds"),
+		instrument.WithUnit("1"),
+		instrument.WithExplicitBucketBoundaries(
+			0.0000001,
+			0.0000005,
+			0.000001,
+			0.000002,
+			0.000005,
+			0.00001,
+			0.00002,
+			0.00005,
+			0.0001,
+			0.0002,
+			0.0005,
+			0.001,
+			0.002,
+			0.005),
+	)
+	if err != nil {
+		return nil, err
+	}
+	bpfMapEntries, err := meter.Int64Gauge(
+		attr.VendorPrefix+".bpf.map.entries_total",
+		instrument.WithDescription("Number of entries in the eBPF map"),
+	)
+	if err != nil {
+		return nil, err
+	}
+	bpfMapMaxEntries, err := meter.Int64Gauge(
+		attr.VendorPrefix+".bpf.map.max.entries_total",
+		instrument.WithDescription("Max number of entries in the eBPF map"),
+	)
+	if err != nil {
+		return nil, err
+	}
+
 	return &InternalMetricsReporter{
 		ctx:                   ctx,
 		tracerFlushes:         tracerFlushes,
@@ -137,6 +179,9 @@ func NewInternalMetricsReporter(ctx context.Context, ctxInfo *global.ContextInfo
 		instrumentationErrors: instrumentationErrors,
 		avoidedServices:       avoidedServices,
 		buildInfo:             buildInfo,
+		bpfProbeLatencies:     bpfProbeLatencies,
+		bpfMapEntries:         bpfMapEntries,
+		bpfMapMaxEntries:      bpfMapMaxEntries,
 	}, nil
 }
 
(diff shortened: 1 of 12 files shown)


pkg/export/otel/metrics_internal.go
```diff
@@ -126,6 +129,45 @@ func NewInternalMetricsReporter(ctx context.Context, ctxInfo *global.ContextInfo
 		return nil, err
 	}
 
+	// TODO should it be ebpf like the others, or bpf like the original one?
```
marctc, line 134: I don't understand this commet, what do you mean like the others?

pkg/export/otel/metrics_internal.go
```diff
@@ -126,6 +129,45 @@ func NewInternalMetricsReporter(ctx context.Context, ctxInfo *global.ContextInfo
...
+	bpfProbeLatencies, err := meter.Float64Histogram(
+		attr.VendorPrefix+".bpf.probe.latency_seconds",
+		instrument.WithDescription("Latency of the eBPF probe in seconds"),
+		instrument.WithUnit("1"),
+		instrument.WithExplicitBucketBoundaries(
+			0.0000001,
```
marctc: Is possible to try reuse the buckets definition here from the one `imetrics.go`?

pkg/export/otel/traces.go
```diff
@@ -153,8 +154,22 @@ func (tr *tracesOTELReceiver) provideLoop(ctx context.Context) {
 	}
 }
 
+// instrumentTracesExporter checks whether the context is configured to report internal metrics and,
```
marctc, line 166: Why do we need this here? We use to have but I think we removed not long ago

pkg/export/prom/prom_bpf.go
```diff
@@ -92,28 +106,94 @@ func newBPFCollector(ctxInfo *global.ContextInfo, cfg *PrometheusConfig) *BPFCol
...
+	}
+}
+
 func (bc *BPFCollector) reportMetrics(ctx context.Context) {
 	go bc.promConnect.StartHTTP(ctx)
 }
 
+func (bc *BPFCollector) collectInternalMetrics(ctx context.Context) {
+	// TODO configurable?
+	ticker := time.NewTicker(15 * time.Second)
```
marctc: Why do we need a ticker here? Is not possible to keep how it is now and fetch those values when /internal/metrics is requested?

marctc (commented):
Added some comments

pkg/export/prom/prom_bpf.go
NimrodAvni78: i want to enable it also for otel metrics, which are sent with push, and don't depend on prometheus scraping

pkg/export/otel/traces.go
NimrodAvni78: idk why it was removed, it reports the `obi.otel.trace.exports` and `obi.otel.trace.exports.errors` metrics, their metric counterparts are reported but it seems traces was forgotten

pkg/export/otel/metrics_internal.go
NimrodAvni78: the original metric is `attr.VendorPrefix+".bpf.probe.latency_seconds"`
but i see internal metrics with `ebpf` prefix like:
```
	tracerFlushes, err := meter.Float64Histogram(
		attr.VendorPrefix+".ebpf.tracer.flushes",
		instrument.WithDescription("Length of the groups of traces flushed from the eBPF tracer to the next pipeline stage"),
		instrument.WithUnit("1"),
	)
```
so wondered what convention to keep

pkg/export/otel/metrics_internal.go
NimrodAvni78: yes my bad, will fix

pkg/export/otel/metrics_internal.go
marctc: mmmh i see. Both imply breaking changes, but I guess bpf metrics are less used so we can rename them.

NimrodAvni78: i dont think its a breaking change, its just what to decide when using the internal metrics reporter

marctc: I mean changing metric names could be also considered breaking changes, but here it's fine.

pkg/export/otel/traces.go
marctc: Isn't this to instrument OBI itself? Do we need it for this PR tho?

pkg/export/prom/prom_bpf.go
marctc: Ok, in that case maybe we should make that configurable. WDYT?

NimrodAvni78: yeah ill make it configurable

pkg/export/otel/traces.go
NimrodAvni78: yeah its to instrument OBI
i basically combined both adding bpf metrics and fix the trace export metrics, which the docs say its reported and its actually not :)

pkg/export/prom/prom_bpf.go
NimrodAvni78: @marctc done

pkg/export/otel/traces.go
marctc: IMO we shouldn't bring this back. So far it only brought confusion to users as they expect by default with avoid self instrument and default excluded services not to see internal beyla spans. We could selectively enable/disable this, but I think its out of the scope of this PR.

rafaelroquetto: Just to add a bit more context, as I believe this warrants further discussion - by default, we do not self instrument. This is implemented via the "default exclude services" mechanism in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/84165422efaef1e6c4f385b99fd7ed4ebd5d2e69/pkg/obi/config.go#L165

I personally feel that when we are not self-instrumenting, then nothing that would semantically fit in this category should happen. So if we decide to add a flag to enable this, it needs to take the above into account.

In my experience it shoud all be disabled by default - otherwise, as Marc said, customers start to scratch their heads wondering what the hell are these OBI spans that I didn't even choose to instrument flowing around (happened a lot).

NimrodAvni78: yeah but this is done only when adding the internalmetrics configuration, which is disabled by default
and it doesn't export service metrics and spans like we do for other services, its metrics relevant to obi itself

NimrodAvni78:
@rafaelroquetto validated, after setting `OTEL_EBPF_INTERNAL_METRICS_EXPORTER: otel`
I don't see any obi spans, just internal metrics

PR #479, opened 2025-08-19
flag to enable / disable node injector

NimrodAvni78:


pkg/components/nodejs/nodejs.go
```diff
@@ -11,23 +11,27 @@ import (
 
 	"go.opentelemetry.io/obi/pkg/components/ebpf"
 	"go.opentelemetry.io/obi/pkg/components/svc"
-	"go.opentelemetry.io/obi/pkg/obi"
 )
 
+type NodeInjectorConfig struct {
+	// LogConfig enables the logging of the configuration on startup.
```
mmat11: the comment is outdated

pkg/components/nodejs/nodejs.go
```diff
@@ -11,23 +11,27 @@ import (
...
+func NewNodeInjector(cfg *NodeInjectorConfig) *NodeInjector {
 	return &NodeInjector{
 		cfg: cfg,
 		log: slog.With("component", "nodejs.Injector"),
 	}
 }
 
 func (i *NodeInjector) Enabled() bool {
-	return i.cfg.Traces.Enabled() || i.cfg.TracePrinter.Enabled()
+	return i.cfg.Enabled
```
rafaelroquetto: should we still check if Traces or TracePrinter is enabled in addition to that? We need to double check if the injector will run if those are disabled, because we don't want to inject the agent if there are no traces (which I believe would happen if metrics are enabled).

pkg/components/nodejs/nodejs.go
NimrodAvni78: oops

pkg/components/nodejs/nodejs.go
NimrodAvni78: so something like 
`i.cfg.NodeInjector.Enabled && (i.cfg.Traces.Enabled() || i.cfg.TracePrinter.Enabled())`?

mmat11: IMHO, unless I am missing something, the traceprinter shouldn't factor into this

rafaelroquetto: Unfortunately we need to check for both Traces.Enabled() and TracePrinter.Enabled() as they are orthogonal to each other.

Traces.Enabled() is true when a trace export endpoint is defined (e.g.  `OTEL_EXPORTER_OTLP_ENDPOINT` or `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`). One can run OBI without setting those but with setting `OTEL_EBPF_TRACE_PRINTER=text` - in which case only TracePrinter.Enabled() == true and we'd want to see the trace context propagation in action (thus injecting the node agent).

Therefore, the final condition should be what @NimrodAvni78 is proposing:
```
i.cfg.NodeInjector.Enabled && (i.cfg.Traces.Enabled() || i.cfg.TracePrinter.Enabled())
```

NimrodAvni78: pushed

pkg/components/nodejs/nodejs.go
```diff
@@ -6,14 +6,18 @@ package nodejs
...
 	"log/slog"
 	"syscall"
 
 	"go.opentelemetry.io/obi/pkg/components/ebpf"
 	"go.opentelemetry.io/obi/pkg/components/svc"
-	"go.opentelemetry.io/obi/pkg/obi"
 )
 
+type NodeInjectorConfig struct {
+	Enabled bool `yaml:"enabled" env:"OTEL_EBPF_NODE_INJECTOR_ENABLED"`
```
rafaelroquetto:
```suggestion
	Enabled bool `yaml:"enabled" env:"OTEL_EBPF_NODEJS_ENABLED"`
```

Just a suggestion - it may make it easier for the end user/documentation reader to understand what this does.

pkg/obi/config.go
```diff
@@ -233,6 +237,8 @@ type Config struct {
 
 	// LogConfig enables the logging of the configuration on startup.
 	LogConfig bool `yaml:"log_config" env:"OTEL_EBPF_LOG_CONFIG"`
+
+	NodeInjector nodejs.NodeInjectorConfig `yaml:"node_injector"`
```
rafaelroquetto:
```suggestion
	NodeInjector nodejs.NodeInjectorConfig `yaml:"nodejs"`
```
ditto

rafaelroquetto (approved):
LGTM! Thanks for spotting this and fixing it!

pkg/obi/config.go
```diff
@@ -252,6 +257,10 @@ type HostIDConfig struct {
 	FetchTimeout time.Duration `yaml:"fetch_timeout" env:"OTEL_EBPF_HOST_ID_FETCH_TIMEOUT"`
 }
 
+type NodeJsConfig struct {
```
MrAlias: nit: Can we name this using the capitalized initialism?

```suggestion
type NodeJSConfig struct {
```

NimrodAvni78: yep

NimrodAvni78: done :)

pkg/obi/config.go
```diff
@@ -233,6 +236,8 @@ type Config struct {
 
 	// LogConfig enables the logging of the configuration on startup.
 	LogConfig bool `yaml:"log_config" env:"OTEL_EBPF_LOG_CONFIG"`
+
+	NodeJs NodeJSConfig `yaml:"nodejs"`
```
MrAlias: Sorry missed this in my last review ...

```suggestion
	NodeJS NodeJSConfig `yaml:"nodejs"`
```

NimrodAvni78: done :)

NimrodAvni78:
@MrAlias think we need to rerun the integration tests

NimrodAvni78:
nice, looks like we can merge

Issue #218, opened 2025-06-25
Prometheus metrics export is missing the SDK version in target_info


grcevski:
The OTel metrics export correctly report the SDK version used:
```
semconv.TelemetrySDKVersion(buildinfo.Version)
```

However the same field is missing when exporting Prometheus metrics in the target info.

nikhilbhatia08:
Hey @grcevski I would like to pick this issue and try it out. Thanks!

rafaelroquetto:
Assigned to you @nikhilbhatia08

PR #492, opened 2025-08-22
fix: Prometheus metrics export is missing the SDK version in target_info

nikhilbhatia08:
Fixes - [issue](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/218)

Added telemetryVersionKey in labelNamesTargetInfo for `telemetry.sdk.version` to be in the target_info while metrics export

### DIFF
--- a/pkg/export/prom/prom.go
+++ b/pkg/export/prom/prom.go
@@ -84,6 +84,7 @@ const (
 	sourceKey            = "source"
 	telemetryLanguageKey = "telemetry_sdk_language"
 	telemetrySDKKey      = "telemetry_sdk_name"
+	telemetryVersionKey  = "telemetry_sdk_version"
 
 	clientKey          = "client"
 	clientNamespaceKey = "client_service_namespace"
@@ -1024,6 +1025,7 @@ func labelNamesTargetInfo(kubeEnabled bool, extraMetadataLabelNames []attr.Name)
 		serviceJobKey,
 		telemetryLanguageKey,
 		telemetrySDKKey,
+		telemetryVersionKey,
 		sourceKey,
 		osTypeKey,
 	}


pkg/export/prom/prom.go
```diff
@@ -1024,6 +1025,7 @@ func labelNamesTargetInfo(kubeEnabled bool, extraMetadataLabelNames []attr.Name)
 		serviceJobKey,
 		telemetryLanguageKey,
 		telemetrySDKKey,
+		telemetryVersionKey,
```
mariomac, line 1028: In addition to the attribute name, you need to add the attribute value in the `labelValuesTargetInfo` function.

Also, would it be possible to add a test case in `prom_test.go` verifying the correctness of that value?

Thanks for your contribution!

nikhilbhatia08: Yeah sure, Thanks!

nikhilbhatia08: Sorry @mariomac I could not understand why and where would we need to add the attribute value in `labelValuesTargetInfo` ?

nikhilbhatia08: And also I think we need to add the test in `common_test.go` as well right ?

mariomac: `labelNamesTargetInfo` just defines half of each attribute: they attribute name.

For the attribute value, you need to set it in  `labelValuesTargetInfo` following the same order it was defined in `labelNamesTargetInfo`.

This exotic way of defining them is because the Prometheus library first defines the attribute names when the metric is created, and later you need to pass the attribute values on each datapoint value update.

The unit test could go in `prom_test.go`, following the same structure of other tests (or even just updating an existing test, if it applies).

PR #501, opened 2025-08-25
Try to fix flaky multiprocess tests

mariomac:


marctc (approved):
LGTM! BTW I also recently open a PR to fix flakiness, but no sure if same as here: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/463

PR #502, opened 2025-08-25
Fix possible memory leak on OTEL metrics exporter

mariomac:
We detected that recently, OTEL metrics' `mr.reporters.For` invocations starting to generate memory that is never disposed.

### DIFF
--- a/test/integration/traces_test.go
+++ b/test/integration/traces_test.go
@@ -873,45 +873,51 @@ func ensureTracesMatch(t *testing.T, urlPath string) {
 
 	// Ensure all 5 traces have proper full chain Java -> Node
 	for _, trace := range multipleTraces {
-		// Check the information of the java parent span
-		res := trace.FindByOperationName("GET /"+urlPath, "server")
-		require.Len(t, res, 1)
-		parent := res[0]
-		require.NotEmpty(t, parent.TraceID)
-		traceID := parent.TraceID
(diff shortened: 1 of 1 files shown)


rafaelroquetto:
I've given this a spin and it seems the problem is still there:

image

The base and last heap profiles:
[f2.pb.gz](https://github.com/user-attachments/files/21976860/f2.pb.gz)
[f3.pb.gz](https://github.com/user-attachments/files/21976862/f3.pb.gz)

The way I tested this was to launch a toy (go) service listening on port 5002 and then:

```
while true; do   ./service-c &   pid=$!;   sleep 1;   kill -9 $pid; done
```
to keep restarting the service.

OBI command line:

```
sudo OTEL_EBPF_PROFILE_PORT=6060 OTEL_EBPF_OPEN_PORT=5002 OTEL_EBPF_TRACE_PRINTER=text OTEL_EBPF_LOG_LEVEL=debug OTEL_EXPORTER_OTLP_ENDPOINT='http://localhost:4318' bin/ebpf-instrument
```

And finally I captured two profiles (the baseline, and one after some time had elapsed) with

```
 http://localhost:6060/debug/pprof/heap -o /tmp/f2.pb.gz
 ```
 
 the diff generated with
 
 ```
go tool pprof -http=:5000 -diff_base /tmp/f2.pb.gz /tmp/f3.pb.gz
```

PR #507, opened 2025-08-26
Add option to remove IPs from metrics to reduce cardinality

marctc:
Currently when OBI is not able to resolve IPs (either because are internal IPs or external from LBs) it's adding those IPs as labels in RED metrics and service graph metrics, leading to some cardinality explosions. Adding a flag to enable/disable this behaviour, making disabled it by default.

Fixes https://github.com/grafana/beyla/issues/2145


mariomac:
I think we are overloading the Attributes Getter logic with some extra responsibilites, adding some coupling and increasing the API complexity.

I think it would be simpler if, instead, we add an extra node to the pipeline, something like:

```go
func UnresolvedIPFilter(mc *MetricsConfig,  input, output *msg.Queue[[]*request.Span]) {
	return func(ctx context.Context) (swarm.RunFunc, error) {
		if !mc.DropUnresolvedIPS {
			return swarm.Bypass(input, output)
		}
		in := input.Subscribe()
		return func(_ context.Context) {
			defer output.Close()
			for spans := range in {
				// for each span:
				// parse src/dst Hostname
				// if it belongs to an unresolved host: drop the span (or just leave the field as empty)
}
```

This would add the feature without needed to load the rest of the logic.

marctc:
> I think we are overloading the Attributes Getter logic with some extra responsibilites, adding some coupling and increasing the API complexity.
> 
> I think it would be simpler if, instead, we add an extra node to the pipeline, something like:
> 
> ```go
> func UnresolvedIPFilter(mc *MetricsConfig,  input, output *msg.Queue[[]*request.Span]) {
> 	return func(ctx context.Context) (swarm.RunFunc, error) {
> 		if !mc.DropUnresolvedIPS {
> 			return swarm.Bypass(input, output)
> 		}
> 		in := input.Subscribe()
> 		return func(_ context.Context) {
> 			defer output.Close()
> 			for spans := range in {
> 				// for each span:
> 				// parse src/dst Hostname
> 				// if it belongs to an unresolved host: drop the span (or just leave the field as empty)
> }
> ```
> 
> This would add the feature without needed to load the rest of the logic.

But this would remove Ips from any span, removing them from any OTEL span. AFAIK we can't cleanup labels on already defined Prom metric.

mariomac (approved):

PR #512, opened 2025-08-27
Feature/improved kafka parsing

NimrodAvni78:
- refactor kafka parsing, trying to seperate code and make it more clear
- add support for fetching the topic name for kafka fetch request +v13 where only topic UUID is sent, we do this by also listening to MetadataResponse packets and saving a hashmap, this will work only when OBI starts before the kafka consumer or when the kafka consumer refreshes metadata in case of rebalance
- for metadata response to work fully, we need to add support for kafka large buffers, since the response packet is only 128 bytes, and before the topics theres a list of brokers, which in large clusters will be way more then 128 bytes
- added some tests


MrAlias:
Based on https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/464

rafaelroquetto (approved):

rafaelroquetto:
@MrAlias @NimrodAvni78 does it mean #464 is superseded?

PR #513, opened 2025-08-27
Make use of OTEL/OBI generator image for Protoc generation

mariomac:
We were relying on a personal repo image. After `obi-generator:0.2.1` image, now we can use it to generate protoc code too.


pkg/components/otelsdk/sdk_inject.go
```diff
@@ -5,7 +5,7 @@
 
 package otelsdk
 
-//go:generate wget --quiet -N -O grafana-opentelemetry-java.jar https://github.com/grafana/grafana-opentelemetry-java/releases/download/v2.13.2.1/grafana-opentelemetry-java.jar
+//go:generate curl -L -sS -o grafana-opentelemetry-java.jar https://github.com/grafana/grafana-opentelemetry-java/releases/download/v2.13.2.1/grafana-opentelemetry-java.jar
```
rafaelroquetto: This undoes https://github.com/grafana/beyla/pull/1714 - afaik curl does not have the equivalent of `wget` -N, so we'd either need to do find it or revert back to using `wget`.

mariomac: Ok, then I will patch the protoc image to install `wget`

mariomac: When this is merged we can publish generator 0.2.1 and then revert back this change: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/516/files

Issue #505, opened 2025-08-26
K8s: client_service_namespace is not overriden by env or annotations


mariomac:
`client_service_namespace`  value is set using reverse DNS in K8s, basically associating a stored IP with a bunch of K8s metadata and extracting the namespace/name from the `OTEL_...` env vars, labels/annotations, or Kube metadata (as done in `serviceNameNamespaceOwnerID`).

However, we observed in some deployments where the namespace set via `OTEL_RESOURCE_ATTRIBUTES` differs from the K8s namespace, that the namespace seems to be reported correctly in most attributes but in client_service_namespace is still the K8s namespace.

mariomac:
Update: I found that also the `client` attribute fails, as well as `server` and `server_service_namespace`.

Summarizing:
* For client-side service graph metrics, `server` and `server_service_namespace` is stuck in the K8s namespace, while it should override them via annotations or env vars
* For server-side service graph metrics, `client` and `client_service_namespace` are wrong.

mariomac:
This PR will fix it in OBI: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/522

PR #522, opened 2025-08-28
Fix service graph metrics client and client_service_namespace

mariomac:
Addresses: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/505

When a service overrides their OTEL service name and namespace via environment variables, the client and client_service_namespace attributes across different metrics might be contradictory:

* For metadata where we know the process PID or container name, it correctly fetched the service name/namespace from env vars.
* For metadata where we only know the remote IP (e.g. Peer IP), it could fetch the service name/namespace from kubernetes metadata but not from the container env vars.

### DIFF
--- a/test/integration/k8s/common/k8s_common.go
+++ b/test/integration/k8s/common/k8s_common.go
@@ -30,4 +30,5 @@ var (
 type Pinger struct {
 	PodName   string
 	TargetURL string
+	Env       map[string]string
 }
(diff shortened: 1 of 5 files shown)


debug.Dockerfile
```diff
@@ -0,0 +1,44 @@
+FROM golang:1.25-alpine AS builder
```
mariomac: After this PR is merged https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/521

I will rebase to remove this file from this PR.

PR #525, opened 2025-08-28
feat(ci): shard integration tests

skl:


mmat11:
just an idea, would `gotestsum` + `ci-matrix` (see: https://github.com/gotestyourself/gotestsum/pull/258) be a good solution? 

as far as I understood, that tool is able to split jobs automatically based on previous durations so we won't need to maintain shards and balance them ourselves

skl:
Oh nice @mmat11, thanks! I'll take a look at that next week!

skl:
Closing in favour of:

- #535

PR #528, opened 2025-08-29
Add information message when injecting the NodeJS agent

rafaelroquetto:
The NodeJS agent injection is done via the debugger, which, when activated, prints the following message:
```
Debugger listening on ws://127.0.0.1:9229/579497dd-efb5-4813-b8e7-f5f46d7553ca
For help, see: https://nodejs.org/en/docs/inspector
Debugger attached.
```

Customers sometimes wonder what this is about and why their debug port is being opened, which can come across as a major red flag. This PR introduces an informative message to help contextualise this. The output now looks like this:

```
Debugger listening on ws://127.0.0.1:9229/67ef359d-92ec-419d-8af7-d5c3579d2b36
For help, see: https://nodejs.org/en/docs/inspector
Debugger attached.
OpenTelemetry eBPF Instrumentation has injected instrumentation via the NodeJS debugger
The debugger will be deactivated again and closed
```

### DIFF
--- a/pkg/components/nodejs/fdextractor.js
+++ b/pkg/components/nodejs/fdextractor.js
@@ -20,6 +20,9 @@ const { AsyncLocalStorage } = require('async_hooks');
 
 const debug_enabled = false;
 
+console.log('OpenTelemetry eBPF Instrumentation has injected instrumentation via the NodeJS debugger');
+console.log('The debugger will be deactivated again and closed');
+
 // ALS store holds only incomingFd
 const als = new AsyncLocalStorage();
 


pkg/components/nodejs/fdextractor.js
```diff
@@ -20,6 +20,9 @@ const { AsyncLocalStorage } = require('async_hooks');
 
 const debug_enabled = false;
 
+console.log('OpenTelemetry eBPF Instrumentation has injected instrumentation via the NodeJS debugger');
+console.log('The debugger will be deactivated again and closed');
```
skl, line 24: Deactivated when? Might be worth being more specific, is it immediately deactivated like a one-off probe?

rafaelroquetto: Yes - like an one off. I did not want to say "has been deactivated" because the way it works is that we issue a command for the debugger to terminate - but we can't tell if it fails (it should never fail, but famoust last words)

mariomac (approved):

PR #529, opened 2025-08-29
[WIP] Refactor NetO11y tracer

rafaelroquetto:
_This PR is best reviewed on a per-commit basis_


rafaelroquetto:
@skl do you by any chance know what's going on with the tests? It started to fail after I rebased.

skl:
@rafaelroquetto here you go:

- https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/540

pkg/export/otel/expirer.go
```diff
@@ -85,7 +85,7 @@ func (ex *Expirer[Record, Metric, ValType]) ForRecord(r Record, extraAttrs ...at
 	}
 	recordAttrs, attrValues := ex.recordAttributes(r, extraAttrs...)
 	return ex.metric, ex.entries.GetOrCreate(attrValues, func() attribute.Set {
-		ex.log.With("labelValues", attrValues).Debug("storing new metric label set")
+		// ex.log.With("labelValues", attrValues).Debug("storing new metric label set")
```
rafaelroquetto: We are doing I/O on a hotpath, which gets really expensive. I've left it commented out because I am not sure if this is useful for debugging purposes. Happy to nuke it.

mariomac: I find it useful for debugging, but I would replace it by:
```
ex.log.Debug("storing new metric label", "labelValues", attrValues)
```
This is equivalent to `With(...)` but we avoid creating a new  `slog.Logger` instance on each invocation.

If debug logs are disabled, I don't think this should affect performance as it is only invoked on each new combination of metric labels.

rafaelroquetto: Good idea, I've fixed it.

PR #544, opened 2025-09-03
Fix PostgreSQL error parsing

rafaelroquetto:
- Ensure the error parsing does not overrun the buffer.
- Take the reported payload length into account
- Fix unit tests


rafaelroquetto:
(cc @mmat11 @NimrodAvni78 )

mariomac (approved):

PR #545, opened 2025-09-03
Limit cardinality of Span Metrics

mariomac:
Some services might report a high number of routes, which affects also to the cardinality of the span name.

While it is OK for traces, this can dangerously increases to cardinality explosion in the case of spanmetrics.

This PR:

* Creates a new node in the pipeline that keeps track of how many different routes do we have per service. When a given service has a route cardinality higher than the configurable `OTEL_EBPF_METRIC_SPAN_NAMES_LIMIT` variable, the span metrics' name changes to `AGGREGATED`. The names of the traces would remain unaltered.
* Hashicorps' expirable LRU cache is untestable, as it doesn't allow mocking the clock and leaks a goroutine that causes new `synctest.Test` to panic. So a new expirable LRU implementation has been provided.
* The `instrumentation.go` pipeline build has been slightly reorganized for more clarity. Check the updated `pipeline-map.md` mermaid graph for an updated view of the pipeline.


mariomac:
package synctest requires Go 1.25: will wait to this PR to be merged https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/546

pkg/components/helpers/cache/expirable_lru.go
```diff
@@ -0,0 +1,106 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package cache
+
+import (
+	"container/list"
+	"time"
+)
+
+// ExpirableLRU cache. It is not safe for concurrent access.
```
mmat11, line 11: question: does this add some new functionality in comparison to `"github.com/hashicorp/golang-lru/v2/expirable"`?

mariomac: Not in terms of functionality but I found some problems in hashicorp's expirable:

* You can't test it properly (unless you run slow tests) because it does not allow mocking the clock.
* Go 1.25's `synctest` package should allow testing it, but each `expirable.LRU` instance is leaking a goroutine that cannot be destroyed and `synctest.Test` panics in consequence.

This expirable LRU is simpler and easier to test with `synctest.Test`, and more lightweight because it is designed to run everything from the same goroutine, where we control when we want to expire entries.

mmat11: Ok I see, thanks!

pkg/app/request/span.go
```diff
@@ -175,6 +175,10 @@ type Span struct {
 	DBNamespace    string         `json:"-"`
 	SQLCommand     string         `json:"-"`
 	SQLError       *SQLError      `json:"-"`
+
+	// OverrideTracename is set under some conditions, like spanmetrics reaching the maximum
+	// cardinality for trace names
```
MrAlias:
```suggestion
	// OverrideTraceName is set under some conditions, like spanmetrics reaching the maximum
	// cardinality for trace names.
```

pkg/components/helpers/cache/expirable_lru.go
```diff
@@ -0,0 +1,106 @@
...
+	evictCallBack func(K, V)
+}
+
+type expirableEntry[K comparable, V any] struct {
+	key        K
+	value      V
+	lastAccess time.Time
+}
+
+type ExpirableCacheOpt[K comparable, V any] func(*ExpirableLRU[K, V])
```
MrAlias, line 26:
```suggestion
// ExpirableCacheOpt defines a functional option for configuring an ExpirableLRU cache.
type ExpirableCacheOpt[K comparable, V any] func(*ExpirableLRU[K, V])
```

pkg/components/helpers/cache/expirable_lru.go
```diff
@@ -0,0 +1,106 @@
...
+
+func WithEvictCallBack[K comparable, V any](cb func(K, V)) ExpirableCacheOpt[K, V] {
```
MrAlias, line 30:
```suggestion
// WithEvictCallBack sets a callback function to be called whenever an entry is evicted
// from the [ExpirableLRU] cache.
func WithEvictCallBack[K comparable, V any](cb func(K, V)) ExpirableCacheOpt[K, V] {
```

pkg/components/helpers/cache/expirable_lru.go
```diff
@@ -0,0 +1,106 @@
...
+	return func(c *ExpirableLRU[K, V]) {
+		c.evictCallBack = cb
+	}
+}
+
+// NewExpirableLRU creates a new Cache whose entries older than the provided TTL are removed
```
MrAlias:
```suggestion
// NewExpirableLRU creates a new [ExpirableLRU] whose entries older than the provided TTL are removed
```

pkg/components/helpers/cache/expirable_lru_test.go
```diff
@@ -0,0 +1,178 @@
...
+	lru := NewExpirableLRU[string, string](testTTL)
+
+	lru.Put("foo", "bar")
+	v, ok := lru.Get("foo")
+	assert.True(t, ok)
+	assert.Equal(t, "bar", v)
+	assert.Equal(t, 1, lru.Len())
+
+	_, ok = lru.Get("baz")
+	assert.False(t, ok)
```
MrAlias: nit: we should ensure the empty value is also returned

```suggestion
	v, ok = lru.Get("baz")
	assert.False(t, ok)
	assert.Empty(t, v)
```

pkg/components/helpers/cache/expirable_lru_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+		v, ok = cache.Get("key2")
+		require.True(t, ok, "Expected to find key2")
+		assert.Equal(t, "value2", v)
+		time.Sleep(2 * time.Minute)
+
+		v, ok = cache.Get("key3")
+		require.True(t, ok, "Expected to find key3")
+		assert.Equal(t, "value3", v)
+		time.Sleep(2 * time.Minute)
```
MrAlias, line 71:
```suggestion
		// Advance age by two minute:
		//   - "key1": 6 minutes old
		//   - "key2": 4 minutes old
		//   - "key3": 2 minutes old
		time.Sleep(2 * time.Minute)
```

pkg/components/helpers/cache/expirable_lru_test.go
```diff
@@ -0,0 +1,178 @@
...
+		_, ok = cache.Get("key1")
+		assert.False(t, ok, "Expected key1 to be expired")
+		v, ok = cache.Get("key2")
+		require.True(t, ok, "Expected to find key2")
+		assert.Equal(t, "value2", v)
+		v, ok = cache.Get("key3")
+		require.True(t, ok, "Expected to find key3")
+		assert.Equal(t, "value3", v)
+
+		time.Sleep(2 * time.Minute)
```
MrAlias, line 91:
```suggestion
		// Advance age by two minute:
		//   - "key1": removed (expired)
		//   - "key2": 6 minutes old
		//   - "key3": 4 minutes old
		time.Sleep(2 * time.Minute)
```

pkg/components/helpers/cache/expirable_lru_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+		// add key 4
+		cache.Put("key4", "value4")
+		time.Sleep(4 * time.Minute)
```
MrAlias, line 100:
```suggestion
		// Advance age by four minute:
		//   - "key1": removed (expired)
		//   - "key2": 10 minutes old
		//   - "key3": 8 minutes old
		//   - "key4": 2 minutes old
		time.Sleep(4 * time.Minute)
```

pkg/components/helpers/cache/expirable_lru_test.go
```diff
@@ -0,0 +1,178 @@
...
+		assert.False(t, ok, "Expected key1 to be expired")
+		_, ok = cache.Get("key2")
+		assert.False(t, ok, "Expected key2 to be expired")
+		_, ok = cache.Get("key3")
+		assert.False(t, ok, "Expected key3 to be expired")
+		v, ok = cache.Get("key4")
+		require.True(t, ok, "Expected to find key4")
+		assert.Equal(t, "value4", v)
+
+		time.Sleep(4 * time.Minute)
```
MrAlias, line 121:
```suggestion
		// Advance age by four minute:
		//   - "key1": removed (expired)
		//   - "key2": removed (expired)
		//   - "key3": removed (expired)
		//   - "key4": 6 minutes old
		time.Sleep(4 * time.Minute)
```

pkg/components/helpers/cache/expirable_lru_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+		// a re-added key should not expire
+		cache.Put("key1", "value1")
+		time.Sleep(2 * time.Minute)
```
MrAlias, line 130:
```suggestion
		// Advance age by two minute:
		//   - "key1": two minutes old
		//   - "key2": removed (expired)
		//   - "key3": removed (expired)
		//   - "key4": 10 minutes old
		time.Sleep(2 * time.Minute)
```

pkg/components/helpers/cache/expirable_lru_test.go
```diff
@@ -0,0 +1,178 @@
...
+import (
+	"testing"
+	"testing/synctest"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const testTTL = 5 * time.Minute
```
MrAlias: Renaming to the value this contains helps understanding where it is referenced below.

```suggestion
const fiveMin = 5 * time.Minute
```

MrAlias (approved):
Looks good overall. Minor feedback.

PR #552, opened 2025-09-04
Fix Queue mechanism

mariomac:
As long as the pipeline in `instrumenter.go` is growing, we have found some cases where the data stopped flowing across the internal queues.

In complex scenarios where queues are forked to different nodes which could either bypass or subscribe to the original queue, some race condition caused that messages were never delivered.

Current version of the pipeline works correctly, but future changes and addition could manifest again this behavior, and they are somewhat difficult to diagnose.

This PR fixes the race conditions but also adds more flexibility to the queue mechanism:
* To simplify, all the queues without subscriptors become non-blocking. We just have to make sure that each `subscribe()` method is invoked during the swarm instancing.
* Now we allow both bypassing and subscribing to the same queue (previously they were excluding operations).
    1. When you subscribe to a bypassing queue, you get the subscription to the last queue in the bypassing chain.
    2. When you bypass a queue with subscriptors, the subscriptors are forwarded to the last queue in the bypassing chain.


grcevski (approved):
LGTM!

PR #555, opened 2025-09-04
Optimise process age calculation

rafaelroquetto:
Running `process.NewProcess(pid)` for every new process in the system has proven to be very expensive. It does a lot of I/O and line parsing:

image

We don't need to instantiate a full process object (and all that comes with it: env parsing, boot time, all kinds of statistics). Instead, we can parse the process start time right away with little to none abstraction overhead to calculate its age.


rafaelroquetto:
@grcevski good idea! will do

grcevski (approved):
LGTM!

mariomac:
Clever!!

PR #558, opened 2025-09-05
Fix race condition in unit tests

mariomac:
The pipeline unit tests failed with race conditions beause the atributes' selector `Normalize` method was invoked concurrently with the `otelcfg.GetFilteredAttributesByPrefix()` function.

`Normalize()` must be only invoked once, when we are sure the configuration is properly loaded. So I have moved it to a special section at the very beginning of the `instrumenter.Run` function and removed all the redundant invocations around the codebase.

There was already a similar issue with `Normalize` and `Matching`, that was initially resolved by adding a global mutex. Now there is no need for this synchronization mechanism as we are sure that `Normalize` won't be invoked concurrently with any other component.

This PR also reorganizes some functions introduced in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/555 to let OBI unit tests compile in Non-linux hosts.

### DIFF
--- a/pkg/components/discover/watcher_proc.go
+++ b/pkg/components/discover/watcher_proc.go
@@ -19,7 +19,6 @@ import (
 	"github.com/shirou/gopsutil/v3/net"
 	"github.com/shirou/gopsutil/v3/process"
 	"github.com/tklauser/go-sysconf"
-	"golang.org/x/sys/unix"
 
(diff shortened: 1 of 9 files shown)


grcevski (approved):
LGTM! I see you fixed the bsd build issue too :)

PR #562, opened 2025-09-06
Create a single instance of target info instruments

rafaelroquetto:
The instruments for `target_info` and `traces_target_info` were being created per each new metrics instance / reporter pool. That caused a memory explosion in the otel metrics exporter whenever there was some process churn: every time a new process was being created, we'd retrieve a reporter pool for it just to be able to create `target_info` and `traces_target_info`, which was heavy-handed. The graph below demonstrates it:

image

This PR addresses this by moving the `target_info` and `traces_target_info` instruments to the scope of the MetricsReporter - they are now just created once. The issue pretty much disappeared from the profiles:
image


mariomac (approved):
Amazing!

PR #571, opened 2025-09-08
ForEachInput pattern wrapper [Open to discussion]

mariomac:
This PR is mainly a request for discussions/opinions. We can eventually merge it or discard it depending on whether you prefer to adopt the wrapper or stay splicit.

Most of our processing loops within each pipeline node follow the same pattern:

```go
	debug("starting node")
	for {
		select {
		case <-ctx.Done():
			debug("context done, stopping node")
			return
		case obj, ok := <-inputCh:
			if !ok {
				debug("input channel closed, stopping node")
				return
			}
			// process and forward obj
		}
	}
```

Sometimes we forget to handle `<-ctx.Done` or we ignore the scenario where inputCh is closed prematurely, causing the node to get blocked when we try to stop Beyla.

Also, the complete pattern adds some verbosity to each node.

To alleviate the mentioned issues, the `ForEachInput` helper function just wraps the above code pattern and _un-verboses_ the nodes while handling the stop conditions automatically. It is used in a functional manner:

```go
swarms.ForEachInput(ctx, inputCh, debugFunc, func(obj []request.Span) {
   process(obj)
   output.Send(obj)
})


grcevski (approved):
LGTM! I like the idea, it doesn't prevent us from doing anything custom and it removes a lot of boilerplate.

PR #572, opened 2025-09-08
Java routes harvester

grcevski:
This PR introduces a way for OBI to capture actual HTTP routes from instrumented Java processes and produce an accurate route information for a given path. Until now, we were relying on generating routes based on the heuristic generator, which is not 100% accurate and can cause cardinality explosion in certain scenarios. Namely, well named paths like folder structure can trip up the heuristic route generator.

This PR does the following:
1. Sets up a common infrastructure for route harvesting.
2. Implements a single language route harvester for Java. (More will follow, I think we should do Go next).
3. Since Java routes can be broken up how they are constructed, I made a new partial matcher which helps compose the final route.

TODO:
- [x] Integration tests (we have a Java testcase, but unfortunately none of it's paths today have route parameters. I need to add couple of more examples in the test and write a testcase for it). I'm currently thinking the easiest to add might be an extension to the Java Kafka OATS test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,4 @@ test/integration/components/gosql/qosql
 pkg/components/otelsdk/grafana-opentelemetry-java.jar
 opentelemetry-ebpf-instrumentation.sln
 obi.sln
+test/integration/components/javakafka/target/*
--- a/test/integration/components/javatestserver/src/main/java/de/fstab/demo/greeting/GreetingController.java
+++ b/test/integration/components/javatestserver/src/main/java/de/fstab/demo/greeting/GreetingController.java
@@ -3,7 +3,9 @@ package de.fstab.demo.greeting;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.client.RestClient;
+
 import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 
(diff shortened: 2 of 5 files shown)


pkg/components/transform/route/harvesters/java.go
```diff
@@ -0,0 +1,148 @@
...
+	"net/url"
+	"regexp"
+	"sort"
+	"strings"
+	"unicode"
+
+	"github.com/grafana/jvmtools/jvm"
+)
+
+type JavaRouteHarvester struct {
```
mariomac: Nit: outside this package, we will refer to this as `harvesters.JavaRouteHarvester`. Maybe you can just use `harvesters.JavaRoute` or `harvesters.JavaRoutes`.

Maybe also shortening the package name to `harvest`?

pkg/components/transform/route/harvesters/java.go
```diff
@@ -0,0 +1,148 @@
...
+			}
+		}
+		return "{" + b.String() + "}"
+	})
+}
+
+func (h *JavaRouteHarvester) sortRoutes(routes []string) []string {
+	sort.Slice(routes, func(i, j int) bool {
+		hasParamsI := strings.Contains(routes[i], "{")
+		hasParamsJ := strings.Contains(routes[j], "{")
```
mariomac: Should you also take parameters like `/foo/:userID` into consideration?

pkg/components/transform/route/harvesters/java.go
```diff
@@ -0,0 +1,148 @@
...
+	if pos := strings.Index(line, jvmAnnotationDelimiter); pos > 0 {
+		routes = h.parseAndAdd(routes, line, pos, len(jvmAnnotationDelimiter))
+	}
+
+	return routes
+}
+
+var jvmAttachFunc func(pid int, argv []string, logger *slog.Logger) (io.ReadCloser, error) = jvm.Jattach
+
+func (h *JavaRouteHarvester) ExtractRoutes(pid int32) (*RouteHarvesterResult, error) {
```
mariomac: I'm afraid this process might cause some slow startup in a big node. Is there any possibility to know if a given PID points to the same instance of a container or pod so we avoid extracting the same routes multiple times?

Just thinking out loud. We can leave it as-is and try to cache this process only if we really verify a performance problem.

pkg/components/transform/route/part_matcher.go
```diff
@@ -0,0 +1,134 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package route
+
+import (
+	"regexp"
+	"strings"
+)
+
+// PartialRouteMatcher allows matching a given URL path towards a set of framework-like provided
```
mariomac, line 11: Do we need another RouteMatcher instance? The way I was initially thinking the implementation was just to get the harvested pattern strings as a slice, then append it to the user-configured list of routes that is passed to the `NewMatcher(routes []string)` function.

pkg/components/transform/route/harvesters/java.go
grcevski: Those are good ideas, I'll rename.

pkg/components/transform/route/harvesters/java.go
grcevski: The PartialRouteMatcher does support `/foo/:userID`, but Java annotations only use the `{userID}` format, so I think it should work.

pkg/components/transform/route/harvesters/java.go
grcevski: I don't think it's possible, even if it's the same instance of the executable, nothing stops it from loading different Java classes and having different paths. Java is very dynamic. I just realised I have a bug, because if there are multiple Java processes from the same executable (e.g. same path to java) we won't harvest those paths. 

It will be better for Go and binaries.

pkg/components/transform/route/part_matcher.go
grcevski: I attempted that approach but it won't work for Java, here's an example of that they do:

```
@Path("/persons")
public class PersonResource {
    @GET
    @Path("/name/{name}")
    public List<Person> getPersonsByName(@PathParam("name") String name) {
        List<Person> people = new ArrayList<Person>() {
        };
        people.add(new Person(1L, "Hello", 1));
        return people;
    }
....
```
When we ask the JVM for symbols it will give us two paths, one the root `/persons` another the internal `/name/{name}`. The problem is that the actual path is something like `/persons/name/{name}`. I initially tried to generate all combinations, but I abandoned that idea, it simply will explode in memory usage on a large app.

I do agree though with your idea, I think we may need to use both. In a sense, I think the new matcher is a superset of the original matcher, although the original is faster I think. Once we add per service route definitions, I think we'll need a way to use both.

mariomac (approved):
Thanks for the clarifications! LGTM (once the tests get fixed)

pkg/components/discover/attacher.go
```diff
@@ -77,6 +81,7 @@ func (ta *TraceAttacher) attacherLoop(_ context.Context) (swarm.RunFunc, error)
 	ta.processInstances = maps.MultiCounter[uint64]{}
 	ta.beylaPID = os.Getpid()
 	ta.EbpfEventContext.CommonPIDsFilter = ebpfcommon.CommonPIDsFilter(&ta.Cfg.Discovery, ta.Metrics)
+	ta.routeHarvester = *harvest.NewRouteHarvester()
```
rafaelroquetto: Not sure if this matters (and this is clearly unimportant, I am mostly approaching this from a curiosity standpoint): why not have `routeHarvester` be a pointer, so that we don't need to copy the object?

rafaelroquetto (approved):
My comment is minor, don't worry about it - I am mostly being curious. Looks great to me!

pkg/components/discover/attacher.go
grcevski: Good point, I'll make it a pointer.

PR #578, opened 2025-09-09
distinguish flags for codecov reports

mariomac:
Codecov metrics decreased since the parallelization. I suspect that the cause is all the tests within the same matrix being submitted under the same flag.

This provides different coverage flags for Codecov, expecting that instead of overwriting coverage data, it merges.


grcevski (approved):
LGTM!

PR #587, opened 2025-09-10
Per service routes

grcevski:
This PR extends the recently added support for harvesting executable/process specific routes, by allowing users to specify custom route definitions per service. Since different usecases require filtering of incoming and outgoing paths, I've added two sections where the users can define the route patterns.

Namely users can now specify routes as part of the discovery section. For example, if we wanted to instrument a service that's listening on port 8085 and we wanted a custom route for the server calls it receives, we can write OBI config like this:

```
discovery:
  services:
    - namespace: multi-k
      name: java-serv
      open_ports: 8085
      routes:
        incoming:
          - /greeting123/{something}
    - namespace: multi-k
      name: node-serv
      open_ports: 3030
```

In this case the route defined `/greeting123/{something}` is specific for `java-serv`, it has no impact on the paths received by the `node-serv`.

Similarly, one can define outgoing route definitions for HTTP client calls, for example, here's a modified example from above, by putting a `/*` outgoing route definition for the service, e.g. the service is making high-cardinality outgoing calls and we are not interested in them.

```
discovery:
  services:
    - namespace: multi-k
      name: java-serv
      open_ports: 8085
      routes:
        incoming:
          - /greeting123/{something}
        outgoing:
          - /*
    - namespace: multi-k
      name: node-serv
      open_ports: 3030
```

This setting will not impact any outgoing calls made by the `node-serv` service.

The priority for route matching is as follows:

1. We look at any defined global route patterns first.
2. Next we consider the custom manually specified route patterns per service.
3. Finally, we consider any harvested routes from the actual executable.

### DIFF
--- a/pkg/components/discover/typer_test.go
+++ b/pkg/components/discover/typer_test.go
@@ -1,3 +1,6 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
 package discover
 
 import (
@@ -5,12 +8,9 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
 	"go.opentelemetry.io/obi/pkg/services"
 )
-
-type dummyProcess struct {
-	Pid int32
-}
 
 type dummyCriterion struct {
 	name      string


pkg/app/request/span_test.go
```diff
@@ -613,3 +613,15 @@ func TestResponseBodyLength(t *testing.T) {
 		})
 	}
 }
+
+func TestIsHTTPSpan(t *testing.T) {
+	spanHTTP := &Span{Type: EventTypeHTTP}
```
rafaelroquetto, line 618: Should we test every event type here, or it doesn't really matter?

rafaelroquetto (approved):

pkg/app/request/span_test.go
grcevski: yeah I guess so, I was just being lazy, but it won't help with adding new event types, so there's no perfect test for this. I think the most important is that we check http and httpclient.

rafaelroquetto: Makes sense

PR #596, opened 2025-09-11
Feature/improve exclude otel services

NimrodAvni78:
like we talked about in the sig meeting
- try to detect otel protocol from env
- if its for sure not GRPC we move on
- if it can be GRPC, we try to extract the port from the endpoint specified in envvars
- if no envvar exist, check for default port specified in config
- check if it equals the span.PeerPort

### DIFF
--- a/pkg/app/request/span.go
+++ b/pkg/app/request/span.go
@@ -628,6 +628,14 @@ func (s *Span) isTracesExportURL() bool {
 	}
 }
 
+func (s *Span) sendsOnDefaultGrpcOtelPort(defaultOtlpGRPCPort int) bool {
+	otlpPort, ok := s.portFromEndpointEnvVar(envOTLPEndpoint)
+	if ok {
+		return otlpPort == s.PeerPort
+	}
+	return s.PeerPort == defaultOtlpGRPCPort
+}
+
 func (s *Span) sendsTracesOnGrpcOtelPort(defaultOtlpGRPCPort int) bool {
 	otlpTracesProtocol, ok := s.Service.EnvVars[envOTLPTracesProtocol]
 	if ok && otlpTracesProtocol != otlpGrpcProtocol {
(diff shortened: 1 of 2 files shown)


pkg/services/criteria.go
```diff
@@ -96,6 +96,10 @@ type DiscoveryConfig struct {
 	// Disables instrumentation of services which are already instrumented
 	ExcludeOTelInstrumentedServices bool `yaml:"exclude_otel_instrumented_services" env:"OTEL_EBPF_EXCLUDE_OTEL_INSTRUMENTED_SERVICES"`
 
+	// DefaultOtlpGRPCPort specifies the default OTLP gRPC port (4317) to fallback on when missing environment variables on service, for
+	// checking for grpc export requests, defaults to 4317
+	DefaultOtlpGRPCPort int `yaml:"default_otlp_grpc_port" env:"OTEL_DEFAULT_OTLP_GRPC_PORT"`
```
rafaelroquetto:
```suggestion
	DefaultOtlpGRPCPort int `yaml:"default_otlp_grpc_port" env:"OTEL_EBPF_DEFAULT_OTLP_GRPC_PORT"`
```

pkg/app/request/span.go
```diff
@@ -613,22 +624,96 @@ func (s *Span) isTracesExportURL() bool {
...
+	}
+	otlpProtocol, ok := s.Service.EnvVars[envOTLPProtocol]
+	if ok && otlpProtocol != otlpGrpcProtocol {
+		return false
+	}
+	otlpTracesPort, ok := s.portFromEndpointEnvVar(envOTLPTracesEndpoint)
+	if ok {
+		return otlpTracesPort == s.PeerPort
+	}
+	otlpPort, ok := s.portFromEndpointEnvVar(envOTLPEndpoint)
```
grcevski: small nit that these last couple of lines can be combined for both metrics and traces.

NimrodAvni78:
@grcevski @rafaelroquetto done :)

Issue #595, opened 2025-09-11
Add support for MongoDB for Go


grcevski:
The initial support for MongoDB right now works only for kprobe based instrumentations. I think to complete the MongoDB support we should add some uprobes for popular Go MongoDB packages, perhaps there's only one important one, I haven't done research.

PR #601, opened 2025-09-11
MongoDB support for Go programs

grcevski:
This PR introduces MongoDB support for Go programs. It piggybacks on the existing MongoDB support for the userspace, but we capture the information from the eBPF space using uprobes.

TODO:
- [x] Unit tests
- [x] Integration tests
- [x] Tests with the lowest supported mongodb-driver.

Closes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/595

### DIFF
--- a/bpf/common/sockaddr.h
+++ b/bpf/common/sockaddr.h
@@ -11,9 +11,8 @@
 #include <common/http_types.h>
 #include <common/protocol_defs.h>
 
-typedef struct accept_args {
+typedef struct sock_args {
     u64 addr; // linux sock or socket address
-    u64 accept_time;
     int fd;
     u32 __pad;
 } sock_args_t;
--- a/bpf/generictracer/k_tracer.c
+++ b/bpf/generictracer/k_tracer.c
@@ -55,7 +55,6 @@ int BPF_KPROBE(obi_kprobe_security_socket_accept, struct socket *sock, struct so
     sock_args_t args = {};
 
     args.addr = addr;
-    args.accept_time = bpf_ktime_get_ns();
 
     // The socket->sock is not valid until accept finishes, therefore
     // we don't extract ->sock here, we remember the address of socket
@@ -216,7 +215,6 @@ int BPF_KPROBE(obi_kprobe_tcp_connect, struct sock *sk) {
 
     if (args) {
         args->addr = addr;
-        args->accept_time = bpf_ktime_get_ns();
     }
 
     return 0;
--- a/internal/tools/go.mod
+++ b/internal/tools/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/google/go-licenses/v2 v2.0.0
 	github.com/grafana/go-offsets-tracker v0.1.7
 	github.com/onsi/ginkgo/v2 v2.25.3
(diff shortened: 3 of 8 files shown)


bpf/gotracer/go_mongo.c
```diff
@@ -0,0 +1,150 @@
...
+        .start_monotime_ns = bpf_ktime_get_ns(),
+    };
+
+    if (!read_go_str("name",
+                     coll_ptr,
+                     go_offset_of(ot, (go_offset){.v = _mongo_conn_name_pos}),
+                     &req.coll,
+                     sizeof(req.coll))) {
+        bpf_dbg_printk("can't read mongodb Collection.name");
+        goto done;
```
rafaelroquetto:
```suggestion
        return 0;
```

bpf/gotracer/go_mongo.c
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    go_addr_key_t g_key = {};
+    go_addr_key_from_id(&g_key, goroutine_addr);
+
+    client_trace_parent(goroutine_addr, &req.tp);
+
+    bpf_map_update_elem(&ongoing_mongo_requests, &g_key, &req, BPF_ANY);
+
+done:
```
rafaelroquetto:
```suggestion
```

bpf/gotracer/go_mongo.c
```diff
@@ -0,0 +1,150 @@
...
+    go_addr_key_t g_key = {};
+    go_addr_key_from_id(&g_key, goroutine_addr);
+
+    mongo_go_client_req_t *req = bpf_map_lookup_elem(&ongoing_mongo_requests, &g_key);
+
+    if (!req) {
+        client_trace_parent(goroutine_addr, &fresh_req.tp);
+        req = &fresh_req;
+    }
+
```
rafaelroquetto: Just a suggestion, feel free to ignore.

You can wrap this behaviour into its own function so that the call site looks like:

```
mongo_go_client_req_t *req = mongo_request(&g_key);
```

rationale: clarity.

Which begs the question - is it possible that the code that inserts on the very same map in `obi_uprobe_mongo_coll_op` clobbers an already_existing request (because it's inserted with BPF_ANY)? If so, then such a function could come in handy.

bpf/gotracer/go_mongo.c
```diff
@@ -24,40 +24,107 @@
 
 #include <logger/bpf_dbg.h>
 
-SEC("uprobe/op_coll_op")
-int obi_uprobe_mongo_coll_op(struct pt_regs *ctx) {
+#define MONGO_OP_DEF(name, str)                                                                    \
+    const char name[] = str;                                                                       \
+    const u32 name##_size = sizeof(name) - 1;
```
rafaelroquetto:
```suggestion
#define MONGO_OP_DEF(name, str)                                                                    \
    static const char name[] = str;                                                                       \
    static const u32 name##_size = sizeof(name) - 1;
```

bpf/gotracer/go_mongo.c
grcevski: yeah, you know I don't like to allocate stack inside helper functions and return it, I think it's less clear when the bpf verifier freaks out about the stack limit.

PR #602, opened 2025-09-12
bpf: fallback to bounded loop if bpf_loop is not available

mmat11:
- Based on runtime kernel, load either a bpf_loop-enabled tail call or a bounded loop tail call*
- Continue with ebpf collection loading if sockops load fails

*I couldn't make this decision in a single program via BTF at runtime because of a possible loader bug. Loading an unused subprog with static linkage:
```
int prog(...) {
    ...

    if (bpf_core_enum_value_exists(enum bpf_func_id, BPF_FUNC_loop)) {
        res = myfunc();
    } else {
        res = myfunc__legacy();
    }

    ...
}

static int culprit(u32 index, void *data) {
    ...
}

static __always_inline int myfunc() {
    struct callback_ctx data = {.buf = buf, .pos = 0};
    bpf_loop(nr_loops, culprit, &data, 0);
    return 0;
}

static __always_inline int myfunc__legacy() {
    ...
    return 0;
}
```
would make func_info from .BTF.ext inconsistent:
```
field ObiProtocolHttp: program obi_protocol_http: load program: invalid argument: number of funcs in func_info doesn't match number of subprogs (1 line(s) omitted)
```
and I believe it has to do with the way instructions are marshaled in the loader; also these type of functions *must* be subprogs with static linkage: https://elixir.bootlin.com/linux/v6.16.3/source/kernel/bpf/verifier.c#L16708, and I found no other way around that

### DIFF
--- a/bpf/common/trace_common.h
+++ b/bpf/common/trace_common.h
@@ -56,10 +56,6 @@ static __always_inline void trace_key_from_pid_tid(trace_key_t *t_key) {
 }
 
 static int tp_match(u32 index, void *data) {
-    if (!k_bpf_traceparent_enabled) {
-        return 0;
-    }
-
     if (index >= (TRACE_BUF_SIZE - TRACE_PARENT_HEADER_LEN)) {
         return 1;
     }
@@ -88,6 +84,27 @@ static __always_inline unsigned char *bpf_strstr_tp_loop(unsigned char *buf, int
 
     if (data.pos) {
         return (data.pos > (TRACE_BUF_SIZE - TRACE_PARENT_HEADER_LEN)) ? NULL : &(buf[data.pos]);
+    }
+
+    return NULL;
+}
+
+static __always_inline unsigned char *bpf_strstr_tp_loop__legacy(unsigned char *buf, int buf_len) {
+    (void)buf_len;
+
+    if (!k_bpf_traceparent_enabled) {
+        return NULL;
+    }
+
+    const u16 k_besteffort_max_loops = 450; // Limited best-effort search to stay within insns limit
+    for (u16 i = 0; i + TRACE_PARENT_HEADER_LEN < k_besteffort_max_loops; i++) {
+        if (i != 0 && buf[i - 1] != '\n') {
+            continue;
+        }
+
+        if (is_traceparent(&buf[i])) {
+            return &buf[i];
+        }
     }
 
     return NULL;
--- a/bpf/common/trace_common.h
+++ b/bpf/common/trace_common.h
@@ -71,9 +71,9 @@ static int tp_match(u32 index, void *data) {
     return 0;
 }
 
-static __always_inline unsigned char *bpf_strstr_tp_loop(unsigned char *buf, int buf_len) {
+static __always_inline u16 bpf_strstr_tp_loop(unsigned char *buf, u16 buf_len) {
     if (!k_bpf_traceparent_enabled) {
-        return NULL;
+        return 0;
     }
 
     struct callback_ctx data = {.buf = buf, .pos = 0};
@@ -83,31 +83,29 @@ static __always_inline unsigned char *bpf_strstr_tp_loop(unsigned char *buf, int
     bpf_loop(nr_loops, tp_match, &data, 0);
 
     if (data.pos) {
-        return (data.pos > (TRACE_BUF_SIZE - TRACE_PARENT_HEADER_LEN)) ? NULL : &(buf[data.pos]);
+        return (data.pos > (TRACE_BUF_SIZE - TRACE_PARENT_HEADER_LEN)) ? 0 : data.pos;
     }
 
-    return NULL;
+    return 0;
 }
 
-static __always_inline unsigned char *bpf_strstr_tp_loop__legacy(unsigned char *buf, int buf_len) {
+static __always_inline u16 bpf_strstr_tp_loop__legacy(unsigned char *buf, u16 buf_len) {
     (void)buf_len;
 
     if (!k_bpf_traceparent_enabled) {
-        return NULL;
+        return 0;
     }
 
-    const u16 k_besteffort_max_loops = 450; // Limited best-effort search to stay within insns limit
-    for (u16 i = 0; i + TRACE_PARENT_HEADER_LEN < k_besteffort_max_loops; i++) {
-        if (i != 0 && buf[i - 1] != '\n') {
-            continue;
-        }
+    // Limited best-effort search to stay within insns limit
+    const u16 k_besteffort_max_loops = 350;
 
+    for (u16 i = 0; i < k_besteffort_max_loops; i++) {
         if (is_traceparent(&buf[i])) {
-            return &buf[i];
+            return i;
         }
     }
 
-    return NULL;
+    return 0;
 }
 
 static __always_inline const tp_info_pid_t *
--- a/bpf/common/trace_common.h
+++ b/bpf/common/trace_common.h
@@ -71,9 +71,9 @@ static int tp_match(u32 index, void *data) {
     return 0;
 }
 
-static __always_inline u16 bpf_strstr_tp_loop(unsigned char *buf, u16 buf_len) {
+static __always_inline unsigned char *bpf_strstr_tp_loop(unsigned char *buf, const u16 buf_len) {
     if (!k_bpf_traceparent_enabled) {
-        return 0;
+        return NULL;
     }
 
     struct callback_ctx data = {.buf = buf, .pos = 0};
@@ -83,17 +83,18 @@ static __always_inline u16 bpf_strstr_tp_loop(unsigned char *buf, u16 buf_len) {
     bpf_loop(nr_loops, tp_match, &data, 0);
 
     if (data.pos) {
-        return (data.pos > (TRACE_BUF_SIZE - TRACE_PARENT_HEADER_LEN)) ? 0 : data.pos;
+        return (data.pos > (TRACE_BUF_SIZE - TRACE_PARENT_HEADER_LEN)) ? NULL : &buf[data.pos];
     }
 
-    return 0;
+    return NULL;
 }
 
-static __always_inline u16 bpf_strstr_tp_loop__legacy(unsigned char *buf, u16 buf_len) {
+static __always_inline unsigned char *bpf_strstr_tp_loop__legacy(unsigned char *buf,
+                                                                 const u16 buf_len) {
     (void)buf_len;
 
     if (!k_bpf_traceparent_enabled) {
-        return 0;
+        return NULL;
     }
 
     // Limited best-effort search to stay within insns limit
@@ -101,11 +102,11 @@ static __always_inline u16 bpf_strstr_tp_loop__legacy(unsigned char *buf, u16 bu
 
     for (u16 i = 0; i < k_besteffort_max_loops; i++) {
         if (is_traceparent(&buf[i])) {
-            return i;
+            return &buf[i];
         }
     }
 
-    return 0;
+    return NULL;
 }
 
 static __always_inline const tp_info_pid_t *
--- a/bpf/generictracer/protocol_http.h
+++ b/bpf/generictracer/protocol_http.h
@@ -13,6 +13,8 @@
 #include <common/trace_common.h>
 
 #include <generictracer/maps/http_info_mem.h>
+
+#include <generictracer/k_tracer_tailcall.h>
 #include <generictracer/protocol_common.h>
 
 #include <maps/accepted_connections.h>
@@ -41,16 +43,14 @@ static __always_inline u32 trace_type_from_meta(http_connection_metadata_t *meta
     return TRACE_TYPE_SERVER;
 }
 
-static __always_inline void
-http_get_or_create_trace_info(http_connection_metadata_t *meta,
-                              u32 pid,
-                              connection_info_t *conn,
-                              void *u_buf,
-                              int bytes_len,
-                              s32 capture_header_buffer,
-                              u8 ssl,
-                              u16 orig_dport,
-                              unsigned char *(*tp_loop_fn)(unsigned char *, int)) {
+static __always_inline void http_get_or_create_trace_info(http_connection_metadata_t *meta,
+                                                          u32 pid,
+                                                          connection_info_t *conn,
+                                                          void *u_buf,
+                                                          int bytes_len,
+                                                          u8 ssl,
+                                                          u16 orig_dport,
+                                                          u16 (*tp_loop_fn)(unsigned char *, u16)) {
     //TODO use make_key
     egress_key_t e_key = {
         .d_port = conn->d_port,
@@ -146,13 +146,15 @@ http_get_or_create_trace_info(http_connection_metadata_t *meta,
 
         unsigned char *buf = tp_char_buf();
         if (buf) {
-            int buf_len = bytes_len;
-            bpf_clamp_umax(buf_len, TRACE_BUF_SIZE - 1);
-
+            u16 buf_len = bytes_len & (TRACE_BUF_SIZE - 1);
+            __builtin_memset(buf, 0, TRACE_BUF_SIZE);
             bpf_probe_read(buf, buf_len, u_buf);
 
-            unsigned char *res = tp_loop_fn(buf, buf_len);
-            if (res) {
+            u16 idx = tp_loop_fn(buf, buf_len);
+
+            if (idx > 0 && idx < (TRACE_BUF_SIZE - TRACE_PARENT_HEADER_LEN)) {
+                unsigned char *res = &buf[idx];
+
                 bpf_dbg_printk("Found traceparent in headers [%s] overriding what was before", res);
                 unsigned char *t_id = extract_trace_id(res);
                 unsigned char *s_id = extract_span_id(res);
@@ -168,8 +170,7 @@ http_get_or_create_trace_info(http_connection_metadata_t *meta,
                 bpf_dbg_printk("new tp: %s", tp_buf);
 #endif
             } else {
-                bpf_dbg_printk("No additional traceparent in headers, using what was made before",
-                               res);
+                bpf_dbg_printk("No additional traceparent in headers, using what was made before");
             }
         } else {
             return;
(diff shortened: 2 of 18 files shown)


mmat11:
> @mmat11 the k8s tests failing are on me, but hopefully I've got a PR coming this arvo that fixes it.
> 
> If it's not a big ask, can you break this PR into smaller ones? I saw you cleaned up a few things that would go well on their own PRs.

Done, also fixed tests; The failing ones should be flaky

pkg/components/ebpf/common/common.go
```diff
@@ -310,6 +312,26 @@ func SupportsEBPFLoops(log *slog.Logger, overrideKernelVersion bool) bool {
 	return kernelMajor > 5 || (kernelMajor == 5 && kernelMinor >= 17)
 }
 
+func FixupSpec(spec *ebpf.CollectionSpec, overrideKernelVersion bool) {
+	if !SupportsEBPFLoops(ptlog(), overrideKernelVersion) {
```
grcevski, line 318: Magic!

grcevski (commented):
I really like this change, very cool stuff @mmat11!

I'm good to approve, except do you think it would be possible to use the loop override in one of our tests to prove that it works? Maybe a clone of MultiProcessTest with the config option to use the legacy loop? I think all headers are small enough for the parsing to work.

mmat11:
> do you think it would be possible to use the loop override in one of our tests to prove that it works? Maybe a clone of MultiProcessTest with the config option to use the legacy loop?

I think there's currently no way to explicitly force the use of the legacy function (you can only force enabling it via config). I'll check if it's possible to test it in the 5.15 VM tests

mmat11:
@grcevski I enabled the subtests here https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/602/commits/4c50b2d62027df83f696b3d63336b5013f3f70a1 but now the chance flaky tests fail are doubled 😄 , the 5.15 test needs to be restarted

bpf/common/trace_common.h
```diff
@@ -56,10 +56,6 @@ static __always_inline void trace_key_from_pid_tid(trace_key_t *t_key) {
 }
 
 static int tp_match(u32 index, void *data) {
-    if (!k_bpf_traceparent_enabled) {
```
rafaelroquetto, line 59: Good catch! Instead of removing this, perhaps we should simply return 1 to end the loop asap.

bpf/generictracer/protocol_http.h
```diff
@@ -492,3 +490,17 @@ int obi_protocol_http(void *ctx) {
 
     return 0;
 }
+
+// k_tail_protocol_http
+SEC("kprobe/http")
+static int obi_protocol_http(void *ctx) {
```
rafaelroquetto: Are these `static` here intended? Otherwise, they may be the culprit on why you were hitting an instruction limit. _It shouldn't be the case_ as these functions are only being invoked via tailcalls, but might be worth it making sure.

pkg/components/ebpf/common/common.go
```diff
@@ -305,6 +306,7 @@ func SupportsContextPropagationWithProbe(log *slog.Logger) bool {
 }
 
 func SupportsEBPFLoops(log *slog.Logger, overrideKernelVersion bool) bool {
+	// TODO(matt): probe for BPF_FUNC_loop instead of relying on kernel version?
```
rafaelroquetto: I think we can remove this comment, it's more like a private note - and I think you are spot on, it's not possible to do at the ebpf level.

pkg/components/ebpf/common/common.go
```diff
@@ -313,6 +315,26 @@ func SupportsEBPFLoops(log *slog.Logger, overrideKernelVersion bool) bool {
 	return kernelMajor > 5 || (kernelMajor == 5 && kernelMinor >= 17)
 }
 
+func FixupSpec(spec *ebpf.CollectionSpec, overrideKernelVersion bool) {
+	if !SupportsEBPFLoops(ptlog(), overrideKernelVersion) {
+		// Hack: instead of redefining bpf2go generated struct for mutually exclusive conditional programs,
```
rafaelroquetto, line 319: Not hack, but a great idea :)

pkg/components/ebpf/instrumenter.go
```diff
@@ -20,7 +20,7 @@ import (
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/link"
-	v2 "github.com/containers/common/pkg/cgroupv2"
+	"github.com/containers/common/pkg/cgroupv2"
```
rafaelroquetto: These seem unrelated to the PR

pkg/components/ebpf/tcmanager/tcmanager.go
```diff
@@ -93,7 +93,7 @@ func (b *TCBackend) UnmarshalText(text []byte) error {
 		return nil
 	}
 
-	return fmt.Errorf("invalid TCBakend value: '%s'", text)
+	return fmt.Errorf("invalid TCBackend value: '%s'", text)
```
rafaelroquetto: ditto

rafaelroquetto (commented):
Really cool!

Regarding the static callback_fn and the fixup you added, that's about as clean as it gets.

The only alternative I can think of would be splitting the HTTP programs into their own object file or eBPF program and loading them separately. But that's essentially what your fixup already achieves, so this looks solid.

Are we missing any tests or good to go?

bpf/common/trace_common.h
mmat11: this was a redundant check, the only caller of this function already checks for `k_bpf_traceparent_enabled` so returning early would be redundant as well

bpf/generictracer/protocol_http.h
mmat11: mmm good catch, it's probably something I forgot to revert while doing experiments, I'll revert it back

pkg/components/ebpf/common/common.go
mmat11: I noticed there's an exported function for checking if an helper is available, see: https://github.com/cilium/ebpf/blob/main/features/prog.go#L241

I'll do that in another PR and remove the comment here

pkg/components/ebpf/tcmanager/tcmanager.go
mmat11: should I open another PR or can we sneak these typos in this one?

mmat11:
> Are we missing any tests or good to go?

@rafaelroquetto I believe this codepath is tested in integration tests on 5.15 VM (see https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/602/commits/bc365089ba71f2a2bc4034eec27c0024454157ed)

bpf/common/trace_common.h
rafaelroquetto: Ah perfect!

pkg/components/ebpf/tcmanager/tcmanager.go
rafaelroquetto: Yes please, that would be super!

rafaelroquetto (approved):
LGTM! 
I'd just wait for @grcevski to see if he has any further comments before merging it.

mmat11:
> LGTM! I'd just wait for @grcevski to see if he has any further comments before merging it.

I had to split it with tail calls unfortunately. And one wasn't enough (for some reason the 5.15 ci kernel is very picky while it works fine for me locally on AL2 5.10), so I'm testing it with 2

mmat11:
2 tail calls seems to fix the instructions limit issue. The test is now failing for another reason (the rust test found 2 spans instead of 1, https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/17861937621/job/50794720006?pr=602), investigating why

bpf/common/trace_common.h
```diff
@@ -75,9 +71,9 @@ static int tp_match(u32 index, void *data) {
     return 0;
 }
 
-static __always_inline unsigned char *bpf_strstr_tp_loop(unsigned char *buf, int buf_len) {
+static __always_inline u16 bpf_strstr_tp_loop(unsigned char *buf, u16 buf_len) {
```
rafaelroquetto: Are the semantics now correct? In the previous implementation, NULL  represented "not found" - replacing that with 0 doesn't work because 0 can be a valid position.

bpf/generictracer/protocol_http.h
```diff
@@ -144,13 +146,15 @@ static __always_inline void http_get_or_create_trace_info(http_connection_metada
 
         unsigned char *buf = tp_char_buf();
         if (buf) {
-            int buf_len = bytes_len;
-            bpf_clamp_umax(buf_len, TRACE_BUF_SIZE - 1);
-
+            u16 buf_len = bytes_len & (TRACE_BUF_SIZE - 1);
```
```suggestion
            const u16 buf_len = bytes_len & (TRACE_BUF_SIZE - 1);
```

bpf/generictracer/protocol_http.h
```diff
@@ -144,13 +146,15 @@ static __always_inline void http_get_or_create_trace_info(http_connection_metada
...
+            __builtin_memset(buf, 0, TRACE_BUF_SIZE);
```
```suggestion
            bpf_memset(buf, 0, TRACE_BUF_SIZE);
```

it's highly optimised

bpf/generictracer/protocol_http.h
```diff
@@ -144,13 +146,15 @@ static __always_inline void http_get_or_create_trace_info(http_connection_metada
...
             bpf_probe_read(buf, buf_len, u_buf);
-            unsigned char *res = bpf_strstr_tp_loop(buf, buf_len);
 
-            if (res) {
+            u16 idx = tp_loop_fn(buf, buf_len);
+
+            if (idx > 0 && idx < (TRACE_BUF_SIZE - TRACE_PARENT_HEADER_LEN)) {
```
rafaelroquetto: As pointed out on a previous comment, this is not strictly correct as pos 0 is a valid position.

bpf/common/trace_common.h
mmat11: I was trying to reduce states as much as possible; 0 should be invalid in this case because in practice headers portion never start at index 0, correct me if I am wrong

mmat11: Also, I had to do this because adding a check inside the loop (or even outside) would create too many instructions.. there's something fishy with the 5.15 CI kernel

rafaelroquetto: So there's no way of making this work with NULL and the right semantics?

In general terms, whilst unlikely, a big HTTP header can span multiple packets and therefore multiple buffers. I don't think our code handles that gracefully at the moment, but that can change in the future. I think we should make the function robust no matter what, unless that is not possible.

bpf/generictracer/protocol_http.h
mmat11: I get build_bug(), probably because of the size?

bpf/common/trace_common.h
mmat11: I can make one last tentative without checking the lenght, since that's the reason I added the memset. If the CI doesn't agree, I need to look into alternatives

rafaelroquetto: Sounds good. Another thought that popped up in my mind though: if this is about reducing the number of instructions, then I think adding (yet another) tail call may be more robust/safer - my reasoning is that, if these kinds of statements are triggering the maximum instruction count, it will be just a matter of short time until it happens again (or even a different compiler version generating different asm code). It does not even need to be related to this part of the code, but any other section in this codepath that eventually gets expanded will hit the instruction limit again.

mmat11: I also thought about it and the issue is that it's hard to split the `http_get_or_create_trace_info` function (essentially the second tail call only calls this function). I think we would need to refactor that if we get into similar issues

bpf/generictracer/protocol_http.h
mmat11: I reverted the logic

bpf/generictracer/protocol_http.h
mmat11: I replaced memset with a map_lookup from a zeroed buffer

bpf/common/trace_common.h
mmat11: I reverted the logic

bpf/generictracer/protocol_http.h
```diff
@@ -401,13 +402,100 @@ static __always_inline void handle_http_response(unsigned char *small_buf,
...
+                                                         http_info_t *info,
+                                                         http_connection_metadata_t *meta) {
     (void)ctx;
 
+    if (meta) {
+        u32 type = trace_type_from_meta(meta);
+        tp_info_pid_t *tp_p = trace_info_for_connection(&args->pid_conn.conn, type);
+        if (tp_p) {
+            info->tp = tp_p->tp;
+            if (args->self_ref_parent_id) {
```
grcevski, line 421: Hey @mmat11, this self_reference parent id seems unrelated to this change, does it make sense to have this as a separate PR? Or maybe I misunderstand the code?

mmat11: this is needed because of the newly added tail calls, the ID needs to get resolved at the entrypoint and processed at the end so by storing it in args I didn't change the current logic

grcevski: Ah I'm sorry I think I get it, it's related to the tail call.

bpf/generictracer/protocol_http.h
rafaelroquetto: So the size is indeed the problem - bpf_memset will only work on 512 byte chunks.

`cheap_bzero` is not that cheap - it actually calls into the kernel twice - one for a map lookup, and one for a probe read.

The map lookup does quite a bit https://github.com/torvalds/linux/blob/d0ca0df179c4b21e2a6c4a4fb637aa8fa14575cb/kernel/bpf/syscall.c#L1660

`bpf_probe_read` also involves copying from kernel into the buffer.


Instead, you can do something like this:

```
            _Static_assert(TRACE_BUF_SIZE == 1024, "TRACE_BUF_SIZE has changed, please adjust the following memset code");

            bpf_memset(buf, 0, TRACE_BUF_SIZE / 2);
            bpf_memset(buf + 512, 0, TRACE_BUF_SIZE / 2);
```

not the prettiest either, but all inline - I tried it and at least on my kernel no instruction limit was hit.

rafaelroquetto: Even better (seems to work, I will try on 5.x)

```
static __always_inline void cheap_bzero(unsigned char *buf, u16 size) {
  if (!__builtin_constant_p(size))
    __throw_build_bug();

  if (size > 512) {
    bpf_memset(buf, 0, 512);
    cheap_bzero(buf + 512, size - 512);

  } else {
    bpf_memset(buf, 0, size);
  }
}
```

rafaelroquetto: I tried on 5.15 and the ebpf programs loaded.

mmat11: > cheap_bzero is not that cheap - it actually calls into the kernel twice - one for a map lookup, and one for a probe read.

well, it's cheap wrt total instructions

I'll test bpf_memset in the CI

mmat11: both solutions don't work for me, just including "bpfcore/bpf_builtins.h" breaks the build:

```
../../../../bpf/bpfcore/bpf_builtins.h:151:2: error: A call to built-in function 'abort' is not supported.
../../../../bpf/bpfcore/bpf_builtins.h:151:2: error: A call to built-in function 'abort' is not supported.
../../../../bpf/bpfcore/bpf_builtins.h:276:2: error: A call to built-in function 'abort' is not supported.
  276 |         __throw_build_bug();
      |         ^
```

I also tried updating the headers with the same result

mmat11: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/18005635430/job/51225110062?pr=602

grcevski: I think we've always had issues with memset, I don't think we use it anywhere. The only thing that might work is `__builtin_memset`

grcevski: I think we should revert the latest changes and we can create a follow-up PR to improve the performance if it's an issue. I'd like to see bpftop analysis that the previous approach was too slow.

mmat11: I reverted already, I believe the current CI failures are due to temporary issues

grcevski: yeah there's some instability with dockerhub, I've restarted them.

rafaelroquetto: >  I'd like to see bpftop analysis that the previous approach was too slow.

I'd be surprised if it was faster, since it does two helper calls, one of each does internal copies. The magnitude being compared here is bpf_memset vs __builtin_memset, which is already considered slow:
```
        /* __builtin_memset() is crappy slow since it cannot
		 * make any assumptions about alignment & underlying
		 * efficient unaligned access on the target we're
		 * running.
		 */
```

The commit in Cilium has more context https://github.com/cilium/cilium/commit/9c85721783447e10e359f723a1ee7d465c704ccc

And I'd expect it to still be faster than cheap_bzero() - there are no calls into the kernel contexts, no copies nor helpers involved.


@mmat11 the error you are seeing is legit and on purpose: it means we are hitting a codepath we shouldn't - they are a guard.  That's because bpf_builtins redefine __builtin_memcpy and friends to that error, to force us to use the optimised versions.

Have a look here: https://github.com/coralogix/opentelemetry-ebpf-instrumentation/pull/39 - this builds locally for me. I've disabled the redefines as it breaks most of our code, we can tackle re-enabling and porting to the recommended functions into a separate PR.

mmat11: I understand, so basically it's the rest of the code that's triggering the guard. I started another CI with the patch

rafaelroquetto: That's right. I believe the cilium authors (this code is widely used in cilium itself) wanted to avoid __builtin_memset/cpy/move at all costs - so the code redefines these "symbols" to a function that fails the compilation on purpose.

rafaelroquetto: Btw I tested it on 5.10 and 5.15 (running locally, the github workers are toast) and it passes for me.

mmat11: I wanted to make sure it also passes on CI vms, my local 5.10 AL2 kernel had no issues while the CI 5.15 was failing because of instruction limit

bpf/generictracer/protocol_http.h
```diff
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_builtins.h>
 #include <bpfcore/bpf_helpers.h>
```
rafaelroquetto, line 8:
```suggestion
#include <bpfcore/bpf_helpers.h>
#include <bpfcore/bpf_builtins.h>
```

this should fix clang-tidy

mmat11: I included it in bpf_builtins.h. We should eventually upgrade these headers from mainstream once we stop using `__builtin_*` functions so it'll hopefully work without our manual changes

rafaelroquetto: sounds good! thanks for fixing it!

grcevski:
If CI fails again I think we need to merge again with main so that we pick up the disk size fix.

PR #621, opened 2025-09-15
Route harvester timeout

grcevski:
This PR adds a timeout around the call to harvest routes from executables/processes. We can have a zombie process or a process that's struggling to respond, we shouldn't permanently block the pipeline when this happens. I set the default timeout to be 10 seconds.


pkg/components/transform/route/harvest/harvester.go
```diff
@@ -30,29 +36,79 @@ type RouteHarvesterResult struct {
...
-	if fileInfo.Service.SDKLanguage == svc.InstrumentableJava {
-		if _, ok := h.disabled[svc.InstrumentableJava]; !ok {
-			return h.java.ExtractRoutes(fileInfo.Pid)
+	// Create a context with timeout
+	ctx, cancel := context.WithTimeout(context.Background(), h.timeout)
+	defer cancel()
+
+	// Channel to receive the result
+	resultChan := make(chan *RouteHarvesterResult, 1)
+	errorChan := make(chan error, 1)
```
MrAlias: nit: no need for two chan

```suggestion
	// Channel to receive the result
	type result struct {
		r *RouteHarvesterResult
		err error
	}
	resultChan := make(chan result, 1)
```

pkg/components/transform/route/harvest/harvester_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"go.opentelemetry.io/obi/pkg/components/exec"
+	"go.opentelemetry.io/obi/pkg/components/svc"
+)
+
+// successfulExtractRoutes simulates a successful route extraction
+func successfulExtractRoutes(_ int32) (*RouteHarvesterResult, error) {
```
MrAlias: nit

```suggestion
func successfulExtractRoutes(int32) (*RouteHarvesterResult, error) {
```

Similar below.

pkg/components/transform/route/harvest/harvester.go
grcevski: great suggestion, done!

pkg/components/transform/route/harvest/harvester_test.go
grcevski: nice catch, thanks!

Issue #396, opened 2025-07-31
HTTP header and body extraction


NimrodAvni78:
We wanted to open this issue to discuss the design of technical implementation of header and body extraction for http

Requirements:

- attach http request and or response headers to spans as attributes
    - needs to be opt-in
    - some glob pattern to specify which headers to add to spans
    - needs to follow otel semantic conventions
        - `http.(request|response).header.<key>`
    - for some reason semantic convention says the value of the attribute should be a is a list of values instead of a single one, where a header can appear multiple times, don’t think thats a real case?
- attach http request and or response body to a span
    - customer should limit the size of the payload to extract
    - i’m pretty sure there are no otel semantic conventions for this
        - we can invent our own and try to push them to otel, `http.(request|response).body.payload`
    - we can also populate the `http.(request|response).body.size` attributes
    - needs to be opt-in
    - maybe we should concider adding sampling to this, maybe we dont want every span to have the payload since it might add significant cost, or is normal span sampling good enough
    - obfuscation, we must allow users to remove sensitive parts from their request / response payloads, and replace them with some filler (`***`), we might want to have multiple strategies for obfuscating payloads
        - **json obfuscation:** we can detect the payload is json relying on the content-type header:
        for example, lets look at this example json
            
            ```jsx
            {"root":{"username": "myusername", "age": 51, "password": "mypassword"}}
            ```
            
            we can have multiple strategies of how to obfuscate jsons
            
            - **json_values**: obfuscate all json values
                
                ```jsx
                {"root":{"username": "***", "age": "***", "password": "***"}}
                ```
                
            - **json_path**: obfuscate all json values matching a specific jsonPath
                - jsonPath: `$.root.password`
                
                ```jsx
                {"root":{"name": "myusername", "age": 51, "password": "***"}}
                ```
                
            - json_value_pattern: obfuscate all json values matching some glob pattern
                - pattern: `my*`
            
            ```jsx
            {"root":{"name": "***", "age": 51, "password": "***"}}
            ```
            
            - note, this might be too cpu intensive to be worth implementing
        - Plain Text obfuscation: for plain text payloads we can just remove the entire payload and replace it with `***` using the `plaintext` obfuscation

this is an example of how a client can configure this:

```yaml
payload_extraction:
  http:
    max_payload_size: 8192 # maximum size of http packtes to send
    truncate: true # if http request size is bugger the max_payload_size, should we truncate it to fit or ignore the payload extraction for this request
[...]

PR #631, opened 2025-09-16
DRAFT: bpf: use configurable large buffers for HTTP requests

mmat11:
Context: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/396


pkg/config/ebpf_tracer.go
```diff
@@ -102,24 +102,27 @@ type EBPFTracer struct {
...
+	HTTP     uint32 `yaml:"http" env:"OTEL_EBPF_BPF_BUFFER_SIZE_HTTP"`
+	MySQL    uint32 `yaml:"mysql" env:"OTEL_EBPF_BPF_BUFFER_SIZE_MYSQL"`
 	Postgres uint32 `yaml:"postgres" env:"OTEL_EBPF_BPF_BUFFER_SIZE_POSTGRES"`
 }
 
 func (c *EBPFTracer) Validate() error {
 	// TODO(matt): validate all the existing attributes
 
+	switch c.BufferSizes.HTTP {
+	case 0, 128, 256, 512, 1024, 2048, 4096, 8192:
```
rafaelroquetto: Instead of hardcoding those values, we should generalise https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/f3dd462c1392aa433ffb014876230557b4ed77d1/pkg/components/netolly/ebpf/sock_tracer.go#L55

Usage: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/f3dd462c1392aa433ffb014876230557b4ed77d1/pkg/components/netolly/ebpf/sock_tracer.go#L109

bpf/generictracer/protocol_http.h
```diff
@@ -401,6 +408,52 @@ static __always_inline void handle_http_response(unsigned char *small_buf,
...
+    tcp_large_buffer_t *large_buf = (tcp_large_buffer_t *)http_large_buffers_mem();
+    if (!large_buf) {
+        bpf_dbg_printk("http_send_large_buffer: failed to reserve space for HTTP large buffer");
+        return -1;
+    }
+
+    large_buf->type = EVENT_TCP_LARGE_BUFFER;
+    large_buf->packet_type = packet_type;
+    large_buf->action = action;
+    __builtin_memcpy((void *)&large_buf->tp, (void *)&req->tp, sizeof(tp_info_t));
```
```suggestion
    large_buf->tp = req->tp;
```

bpf/generictracer/protocol_http.h
```diff
@@ -401,6 +408,52 @@ static __always_inline void handle_http_response(unsigned char *small_buf,
...
+    }
+
+    if (!is_pow2(http_buffer_size)) {
+        bpf_dbg_printk("http_send_large_buffer: bug: http_buffer_size is not a power of 2");
+        return -1;
+    }
+    const u32 buf_len_mask = http_buffer_size - 1;
+
+    tcp_large_buffer_t *large_buf = (tcp_large_buffer_t *)http_large_buffers_mem();
+    if (!large_buf) {
```
rafaelroquetto, line 422:
```suggestion

    if (!large_buf) {
```

bpf/generictracer/protocol_http.h
```diff
@@ -401,6 +408,52 @@ static __always_inline void handle_http_response(unsigned char *small_buf,
...
+        return -1;
+    }
+
+    large_buf->type = EVENT_TCP_LARGE_BUFFER;
+    large_buf->packet_type = packet_type;
+    large_buf->action = action;
+    __builtin_memcpy((void *)&large_buf->tp, (void *)&req->tp, sizeof(tp_info_t));
+
+    large_buf->len = bytes_len;
+    if (large_buf->len >= http_buffer_size) {
```
rafaelroquetto, line 433:
```suggestion

    if (large_buf->len >= http_buffer_size) {
```

bpf/generictracer/protocol_http.h
```diff
@@ -401,6 +408,52 @@ static __always_inline void handle_http_response(unsigned char *small_buf,
...
+        large_buf->len = http_buffer_size;
+        bpf_dbg_printk("WARN: http_send_large_buffer: buffer is full, truncating data");
```
rafaelroquetto, line 435: I wonder if we should drop the buffer here instead. How do we deal with truncated data?

bpf/generictracer/iovec_len.h
```diff
@@ -3,4 +3,4 @@
 
 #pragma once
 
-enum { k_iovec_max_len = 512 };
+enum { k_iovec_max_len = 8192 };
```
rafaelroquetto, line 6: I wonder if we can calculate this from userspace, to ensure it is compatible to any large buffer we have (I am guessing the largest one).

bpf/generictracer/protocol_http.h
```diff
@@ -21,6 +23,11 @@
 
 volatile const u32 high_request_volume;
 
+enum {
+    k_http_large_buf_max_size = 1 << 14, // 16K
+    k_http_large_buf_max_size_mask = k_http_large_buf_max_size - 1,
+};
```
rafaelroquetto: I think you don't need these. Instead, you can use `SCRATCH_MEM_NAMED(http_large_buffers, tcp_large_buffer_t)`

Then, in userspace, you effectively set the value size of this map to be that of the http buffer (if > 0) like you for `http_buffer_size`.

We do a similar thing here: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/f3dd462c1392aa433ffb014876230557b4ed77d1/pkg/components/netolly/ebpf/sock_tracer.go#L111

But instead of `.MaxEntries` you can use `.ValueSize`

bpf/generictracer/protocol_http.h
```diff
@@ -401,6 +408,52 @@ static __always_inline void handle_http_response(unsigned char *small_buf,
...
+
+    bpf_probe_read(large_buf->buf, large_buf->len & buf_len_mask, u_buf);
+
+    u32 total_size = sizeof(tcp_large_buffer_t);
+    total_size += large_buf->len > sizeof(void *) ? large_buf->len : sizeof(void *);
+
+    req->has_large_buffers = true;
+
+    bpf_ringbuf_output(
+        &events, large_buf, total_size & k_http_large_buf_max_size_mask, get_flags());
```
rafaelroquetto: You may not need the mask here (in my tests I did not, but I did not test with older kernels).

If we can remove it, then we can do without those `k_http_large_buf...` constants altogether and rely on a single source of truth - way less error prone.

pkg/components/ebpf/common/httpfltr_transform.go
```diff
@@ -75,15 +76,31 @@ func ReadHTTPInfoIntoSpan(record *ringbuf.Record, filter ServiceFilter) (request
...
+
+	if event.HasLargeBuffers == 1 {
+		b, ok := extractTCPLargeBuffer(parseCtx, event.Tp.TraceId, event.Tp.SpanId, packetTypeRequest)
+		if !ok {
+			slog.Debug("missing large buffer for HTTP request", "traceID", event.Tp.TraceId, "spanID", event.Tp.SpanId)
+			return request.Span{}, false, nil
+		}
+		requestBuffer = b
+
+		// TODO: response?
```
rafaelroquetto: Does that mean we are not using/consuming the buffers being set for the response? If that's the case, then we should remove that part from the ebpf code as well (i.e. not send any response buffers) until this is done - otherwise we are just wasting processing.

pkg/components/ebpf/common/httpfltr_transform.go
```diff
@@ -92,34 +109,34 @@ func HTTPInfoEventToSpan(event *BPFHTTPInfo) (request.Span, bool, error) {
...
 		parsedHost = true
 
 		if bufPort >= 0 {
 			result.Host = bufHost
 			result.ConnInfo.D_port = uint16(bufPort)
 		}
 	}
-	result.URL = event.url()
-	result.Method = event.method()
+	result.URL = httpURLFromBuf(requestBuffer)
```
rafaelroquetto, line 116: For each of these extraction methods, we are always allocating a new string no matter what e.g. `buf := string(req)` - which copies the underlying buffer, just to discard it sometimes when we do not find what we are looking for.

Instead we should strive to only allocate when we really need to return a non empty-value. This predates this PR but since we are touching these, it's a good opportunity to fix them.

rafaelroquetto (commented):
Good stuff!

Just a bunch of minor comments. The more pressing of them IMHO relates to streamlining the buffer sizes - currently we have a lot of them, and it's not clear to the naked eyes / future maintainer how (and if) they relate.

My recommendation is to have the three sizes (HTTP, MySQL and Postgres) as the source of their respective truths.

It's then possible to adjust the respective map value sizes at load time to match those, and also to be a bit more lax on the accepted values: instead of hardcoding them, we can constrain them to the required alignment and adjust the final value - this is less error prone because everything is enforced as a rule, rather than explicit comparisons, analogous to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/f3dd462c1392aa433ffb014876230557b4ed77d1/pkg/obi/network_cfg.go#L97

bpf/generictracer/protocol_http.h
mmat11: We don't -- we just truncate it and it's up to the handler to deal with it. I was thinking to introduce some new tag for this.
Imho we shouldn't drop it, there could be cases in troubleshooting in which it might be useful to see a span with truncated data rather than nothing

bpf/generictracer/iovec_len.h
mmat11: I tried already, I don't think we can without further refactoring, this value needs to be known at compile time

bpf/generictracer/protocol_http.h
mmat11: I could do this, but we don't save much per cpu, I wonder if it's worth adding this complication for a few kbytes

bpf/generictracer/protocol_http.h
mmat11: I will test it without mask, I just put it out of habit :D

pkg/components/ebpf/common/httpfltr_transform.go
mmat11: it's not being used as of now, we do need it shortly though - I have graphql support in WIP state which needs this

rafaelroquetto: I'd suggest moving it as part of that PR then

bpf/generictracer/protocol_http.h
rafaelroquetto: It's not only that, but the main point is to streamline all of these constants. Currently we have `http_buffer_size` being set from userspace, and some defined in kernel space, and they are _not_ orthogonal to each other - so its easy for someone to break the code by adjusting `k_http_large_buf_max_size` to a size that ends up being less than the max allowed `http_buffer_size` for instance - they need to know to change it in both places.

On the other hand, userspace can calculate the "effective http buffer size" from the original `BufferSizes.HTTP` value (e.g. rounding it to the next power of 2) and ensure both the map value size (the scratch mem) and `http_buffer_size` remain compatible (e.g. they could be the same when the buffer size is > 0, otherwise the map value size can be 1 to pass the verifier)

bpf/generictracer/iovec_len.h
rafaelroquetto: All good!

bpf/generictracer/protocol_http.h
rafaelroquetto: "Provide mechanism, not policy" they said! Very good point, thanks for clarifying.

bpf/generictracer/protocol_http.h
mmat11: Ok I am getting confused :D, the sizes specified there act as a "guard" for the maximum buffer size the event can assume regardless of config options. The event real size will always be dynamic depending on the payload, so if we get http events with a 20 bytes payload, we send 20 bytes + "event size" via the ringbuffer, not the max size (16k). I refactored everything large buffer related into `common/large_buffers.h` into a single constant (instead of per-protocol constants) and added a comment to make it clear. Let me know if this is good enough

pkg/components/ebpf/common/httpfltr_transform.go
mmat11: Removed

pkg/config/ebpf_tracer.go
mmat11: I don't understand how should I generalize, you mean making a function to check if it's 0 or a power of 2 for validation?

pkg/components/ebpf/common/httpfltr_transform.go
mmat11: I'm refactoring this code in the next PR, can we address this there?

rafaelroquetto: Yes no worries

bpf/common/large_buffers.h
```diff
@@ -0,0 +1,24 @@
...
+    // This value represents a pessimistic guard for the maximum size
+    // a large buffer event can take into the ring buffer.
+    // The actual size is "event size + payload". Since the payload
+    // is guaranteed to be a power of 2, we take the next power of 2
+    // of the maximum payload size as a guard.
+    k_large_buf_max_size = 1 << 14, // 16K
+    k_large_buf_max_size_mask = k_large_buf_max_size - 1,
+};
+
+SCRATCH_MEM_SIZED(http_large_buffers, k_large_buf_max_size);
```
rafaelroquetto, line 26: Awesome!

pkg/config/ebpf_tracer.go
rafaelroquetto: I am suggesting to scrap this from Validate() altogether, and instead let the user specify any buffer size - we then adjust this value to the next power of 2 or next sane value:

```
// ensures http_buffer_size is a power of 2 of at least os.Pagesize()
m["http_buffer_size"] = effectiveRingBufferSize(p.cfg.EBPF.BufferSizes.HTTP)
```

(obviously we should likely rename `effectiveRingBufferSize()` to something like `effectiveBufferSize()`).

The generalisation comment was in reference to generalising `effectiveRingBufferSize()` for this purpose.

mmat11: I miss the point, doing so would let users specify a value and have a different effective value being loaded (eg. user specifies size=2050, effective size would be 4096, which is ~double that size)

rafaelroquetto: That's right and we should document that - perhaps even print it in the logs.

rafaelroquetto: This got me wondering - does the buffer need to be a power of 2, or just aligned?

pkg/components/ebpf/common/httpfltr_transform.go
```diff
@@ -75,15 +76,29 @@ func ReadHTTPInfoIntoSpan(record *ringbuf.Record, filter ServiceFilter) (request
...
-	parsedHost := false
+	var (
+		bufHost       string
+		bufPort       int
+		parsedHost    bool
+		requestBuffer []byte
+	)
+
+	if event.HasLargeBuffers == 1 {
+		b, ok := extractTCPLargeBuffer(parseCtx, event.Tp.TraceId, event.Tp.SpanId, packetTypeRequest)
```
grcevski, line 93: Actually, can we please change this to fallback to the event.Buf[:] if we fail to extract the buffer? No need to reject the span. I am thinking this is the reason the integration tests are failing.

pkg/config/ebpf_tracer.go
mmat11: Mmm I don't think it needs to, after all once we set the values from userspace they are treated as constants IIRC - that's just what we agreed on during the first iteration, I'll do some testing to see if I can make it more flexible

pkg/components/ebpf/common/httpfltr_transform.go
mmat11: Yes I can do that, I am not sure the failing test is related though, the same test just failed in my other PR -> https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/18014863416/job/51257072591 😞

grcevski: Yeah possibly not, it passed on the latest re-run, I'll take a look what's up. Maybe just make the safe fix anyway so that we get the most spans, better something than nothing.

mmat11:
@rafaelroquetto @grcevski I removed the pow2 limitation so now the size can be anything from 0 to 8192. I also added fallback to the small buffer that was copied previously

rafaelroquetto (approved):

mmat11:
> Argh, I was about to hit merge and we have conflicts now. Sorry @mmat11, would it be possible to resolve them? I don't know which merge caused it.

Done, CI is green too ✅

PR #633, opened 2025-09-16
Increase grpc max recv size

rafaelroquetto:
The neto11y tests are generating more metrics, which are causing the collector ingest to overflow and reject them.


grcevski (approved):
LGTM!

PR #635, opened 2025-09-16
Improve route harvesting

grcevski:
The Java route harvester was parsing the output from the JVM with a `scanner` which is by default limited to 64K. On larger Java applications this buffer limit may not be enough for all symbols, so instead of extending the scanner to another limit, I re-wrote the code to directly parse the output buffer.

I also improved the loop that handles the partial Java routes by reducing the allocations required and I added an extra test for the edge case where the optimisation may not work:

Prior benchmark results:

```
goos: linux
goarch: amd64
pkg: go.opentelemetry.io/obi/pkg/components/transform/route
cpu: 12th Gen Intel(R) Core(TM) i7-1280P
BenchmarkPartialRouteMatcherSimple-20     	 1177230	       964.7 ns/op	     736 B/op	      20 allocs/op
BenchmarkPartialRouteMatcherComplex-20    	  355447	      3185 ns/op	    2040 B/op	      32 allocs/op
```

New benchmark results:

```
goos: linux
goarch: amd64
pkg: go.opentelemetry.io/obi/pkg/components/transform/route
cpu: 12th Gen Intel(R) Core(TM) i7-1280P
BenchmarkPartialRouteMatcherSimple-20     	 1369430	       833.6 ns/op	     496 B/op	      11 allocs/op
BenchmarkPartialRouteMatcherComplex-20    	  454360	      2606 ns/op	    1016 B/op	      12 allocs/op
```


mariomac (approved):
LGTM!!

PR #643, opened 2025-09-17
use durations for bpf_metric_scrape_interval

mariomac:
For consistency with the rest of configuration options that use `time.Duration` instead of integers, the `bpf_metric_scrape_interval_seconds` has been renamed to `bpf_metric_scrape_interval` and accepting any duration string.


grcevski (approved):
LGTM!

PR #648, opened 2025-09-17
fix: panic in clusterurl

skl:
Investigating the following error:

```
panic: runtime error: index out of range [9] with length 9
```

After fixing the panic, I found some of the existing test cases to be inconsistent, and added some missing test cases for `ClusterURL`. Discussed with @grcevski and agreed to update the existing test cases to focus on consistency.

TL;DR The `p` slice was shrunk to strip characters, then access was attempted on an index that no longer existed.


grcevski (approved):
LGTM!

Issue #656, opened 2025-09-19
K8s namespace and deployment type information lost when previous pod exits


paweljw:
I'm observing a somewhat similar situation to https://github.com/grafana/beyla/issues/1228 in a test opentelemetry-demo deployment + Beyla. 

Rough timeline:

- Beyla starts (privileged container)
- otel demo starts, amongst other things a `valkey-cart` container
- initially traces from this container result in traces with `service.name == "otel-demo/Deployment/valkey-cart"`
- after quite a long time (I believe around 1-2 days) of uninterrupted running from both processes, Beyla appeared to "lose" the tagging information, and spans were sent with `service.name == "valkey-cart"`
- this persisted until the otel-demo deployment was fully restarted.

Extra context is that this test deployment was a little "troubled" initially, so valkey-cart was crash-looping and was redeployed multiple times. In Beyla logs, I can see this situation (limited to valkey-cart PIDs for clarity):

- PID 786043 instrumented: Sep 16 at 11:54:34 <- stable valkey pid, tracked for 2 days
- PID 825135 ended:        Sep 16 at 15:49:28 <- first "lingering" stuck pod dies
- PID 912010 ended:        Sep 17 at 02:05:51
- PID 913843 ended:        Sep 17 at 02:17:19
- PID 1000051 ended:       Sep 17 at 11:02:35
- PID 786043 ended:        Sep 18 at 12:54:28 <- full otel demo restart

Tracking this change, it appears that the "cutoff" happens at exactly 15:49 on Sep 16th, coinciding with the first "lingering" PID seen as ended:

<img width="1290" height="343" alt="Image" src="https://github.com/user-attachments/assets/dd045386-415d-471a-a16d-0a7a55daaa64" />

From a (rather uneducated) reading of the code, this appears to be a suspect:

https://github.com/grafana/opentelemetry-ebpf-instrumentation/blob/df9d1603b2bf061d059e2c213025cfdfde92fd23/pkg/components/kube/store.go#L229-L239

Reading previous pull requests, it appears that a similar bit of code was previously removed to fix a bug:

https://github.com/grafana/beyla/pull/1156/files

Perhaps reintroducing this behavior is a regression?

Please let me know if I can provide any more information that would be helpful. Sudden service name changes are quite detrimental to building e.g. meaningful alerting on top of them without workarounds. 🙏

grcevski:
[...]

PR #661, opened 2025-09-19
Fix target_info tracking issues

grcevski:
This PR resolves multiple edge cases with our tracking of target_info:

1. In certain scenarios we can receive multiple events for a given process, for example a single container which has multiple instrumented processes, but they are built with different programming languages. One example of such container is the load-generator in the OTel DEMO. It has a python, nodejs and generic (c++/rust) application, all within the same container. Until now, in these scenarios we'd leave stale target_info information for 2 out of the 3 processes, since the labels are recorded as last one wins. The code now checks to see if we are receiving duplicate labels for the same service and correctly removes the old info before creating new information.
2. Another edge-case has to do with delayed container creation. A customer may configure OBI to instrument all Kubernetes services that are listening on a certain port, e.g. 8080. In this case, we might match the process information before we receive the Kubernetes metadata. Since we have no metadata yet, we'll record the process in target_info as "naked" -- undecorated process. Later once the Kubernetes metadata arrives, the spans will start reporting service names based on the updated information, but target_info will never change. I've added a subscriber to the Kubernetes store metadata in the process event Kubernetes generator, which pushes a new process created event once the updated metadata arrives. Since I've also fixed 1., this nicely updates the information and the target_info is now aligned with what's reported with the spans.
3. Finally, I found that the store.go code also had the assumption that there can only be one PID per container. What this would do is that if we launched 3 processes, we'll update the information in the store based on the last process. However, if that last process terminates (but the other 2 are alive), we'll remove the metadata from the store. I believe this caused #656, but I'm investigating a bit more.

Relates to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/656

These error conditions are impossible to test reliably with integration tests, so I relied heavily in this PR on unit tests, simulating delayed information.

### DIFF
--- a/pkg/components/kube/store.go
+++ b/pkg/components/kube/store.go
@@ -197,19 +197,13 @@ func (s *Store) cacheResourceMetadata(meta *informer.ObjectMeta) *CachedObjMeta 
 func (s *Store) On(event *informer.Event) error {
 	switch event.Type {
 	case informer.EventType_CREATED:
-		// go func() {
-		// 	time.Sleep(10 * time.Second)
 		s.addObjectMeta(event.Resource)
-		s.Notify(event)
-		// }()
 	case informer.EventType_UPDATED:
 		s.updateObjectMeta(event.Resource)
-		s.Notify(event)
 	case informer.EventType_DELETED:
 		s.deleteObjectMeta(event.Resource)
-		s.Notify(event)
 	}
-	//s.Notify(event)
+	s.Notify(event)
 	return nil
 }
 
--- a/pkg/transform/k8s.go
+++ b/pkg/transform/k8s.go
@@ -25,9 +25,7 @@ import (
 	"go.opentelemetry.io/obi/pkg/pipe/swarm"
 )
 
-var (
-	containerInfoForPID = container.InfoForPID
-)
+var containerInfoForPID = container.InfoForPID
 
 func klog() *slog.Logger {
 	return slog.With("component", "transform.KubernetesDecorator")
--- a/pkg/components/kube/store_test.go
+++ b/pkg/components/kube/store_test.go
@@ -721,7 +721,7 @@ func TestStore_MultiPID_SameContainerAndNamespace(t *testing.T) {
 			setupPIDs:   []uint32{2001, 2002, 2003, 2004},
 			containerID: "container456",
 			pidNS:       6000,
-			operations: func(t *testing.T, store *Store, pids []uint32) {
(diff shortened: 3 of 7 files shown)


pkg/components/kube/store.go
```diff
@@ -82,15 +82,15 @@ type Store struct {
 
 	metadataNotifier meta.Notifier
 
-	containerIDs map[string]*container.Info
+	containerIDs maps.Map2[string, uint32, *container.Info]
```
grcevski, line 85: Fixes for 3)

pkg/export/otel/metrics.go
```diff
@@ -1024,6 +1039,27 @@ func (mr *MetricsReporter) onProcessEvent(pe *exec.ProcessEvent) {
 	mr.log.Debug("Received new process event", "event type", pe.Type, "pid", pe.File.Pid, "attrs", pe.File.Service.UID)
 
 	if pe.Type == exec.ProcessEventCreated {
+		uid := pe.File.Service.UID
```
grcevski, line 1042: Fixes for 1)

pkg/export/prom/prom.go
```diff
@@ -1154,28 +1162,68 @@ func (r *metricsReporter) disassociatePIDFromService(pid int32) (bool, svc.UID)
...
+	r.deleteTargetInfoMetric(service)
+	r.deleteTracesTargetInfoMetric(service)
+}
+
+func (r *metricsReporter) deleteTargetInfos(uid svc.UID, service *svc.Attrs) {
+	r.deleteEventMetrics(r.origService(uid, service))
+}
+
+func (r *metricsReporter) handleProcessEvent(pe exec.ProcessEvent, log *slog.Logger) {
+	log.Debug("Received new process event", "event type", pe.Type, "pid", pe.File.Pid, "attrs", pe.File.Service.UID)
```
grcevski, line 1180: Fixes for 1)

pkg/transform/k8s.go
```diff
@@ -159,12 +164,123 @@ func (md *metadataDecorator) nodeLoop(ctx context.Context) {
...
+	defer t.missedPodsMux.Unlock()
+
+	m, ok := t.missedPods[containerID]
+
+	return m, ok
+}
+
+func (md *procEventMetadataDecorator) ID() string { return "unique-proc-event-metadata-decorator-id" }
+
+func (md *procEventMetadataDecorator) On(event *informer.Event) error {
```
grcevski, line 263: Fixes for 2)

PR #685, opened 2025-09-25
ci: VM integration tests speedup

skl:
This PR speeds up and simplifies the VM integration tests CI workflow:

1. Combine three workflow files into one, with more matrix parameters.
2. Pre-compile the Go integration tests before entering the qemu VM.
3. Additional sharding per test was introduced to increase the number of parallel VMs.
4. Drop the 6.10 workflow (see "Research" below).
5. Reduce Docker command output verbosity.

> [!NOTE]
> The tests below were done by removing the workflow path filter, which was restored in 7d8360f5e.

## New workflow timings

Total workflow: **18 minutes** ([link](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/18175407029?pr=685), commit: 6c8617c1b)

Screenshot 2025-10-01 at 22 18 46

## Baseline timings

Phase | Kernel 5.15 | Kernel 6.10 | Difference
-- | -- | -- | --
Docker generate | ~72s (15:17:07 → 15:18:19) | ~70s (15:17:23 → 15:18:33) | +2s (+3%)
VM image build | 29s (15:18:19 → 15:18:48) | 27s (15:18:33 → 15:19:00) | +2s (+7%)
VM execution | 4910s (81.8 min) | 2541s (42.4 min) | +2369s (+93%)
Cleanup | <1s | <1s | ~0s
Total workflow | 83.5 minutes | 44.0 minutes | +39.5 min (+90%)

### Research

1. Kernel 5.15 workflow takes twice as long as the 6.10 workflow
1. The majority of the time spent seems to be in the test compilation phase. As this is done from within the qemu VM, it would be faster to pre-compile the go integration tests first, and mount them into the VM.
1. The VM resources do not match the GitHub Action Runners ([spec](https://docs.github.com/en/actions/how-tos/write-workflows/choose-where-workflows-run/choose-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories)). The number of vCPUs should be reduced, and we can allocate more RAM.
1. Kernel 6.10 workflow no longer required, as GitHub Runners are on 6.11, so the normal integration test workflow covers the "newer" kernel branch of tests.
1. Much of the test logs are taken up by verbose Docker command output, which is not useful (such as progress indicators). The Docker commands should be tuned for a non-interactive environment.


grcevski (approved):
Looks great. LGTM!

For a moment I thought no VM tests ran, but it's the filter and I see that you ran them without the filter before. Good stuff!

PR #688, opened 2025-09-26
Better unresolved span metrics

grcevski:
To avoid cardinality explosion we had implemented a catch all "unresolved" for any IP target found for service graph metrics. However, there were some edge cases where this might not work properly.

1. This "unresolved" service is a single node without a namespace. This means that any tool that wants to print a service graph would essentially have edges to this one "singular" node, making the service graphs hard to interpret. For example, all incoming web traffic to a service will be the "unresolved" node, but also all outgoing cloud services will point to the same now "incoming" node.
2. The Prometheus export was not using the new attribute helpers, so the cardinality fixes didn't fix Prometheus metric exports. I've modified the code now to use the attributes instead of hardcoded constants.
3. When we were tracing an outgoing client call, the target host information is typically present. We pass this information and it's typically available in traces. However the name resolver code didn't use this information, so we'd end up with IP addresses and then "unresolved" for cases where we actually know the target information.
4. The nameresolver only worked in K8s mode, which is suboptimal. Namely, we'd mark the names of our services as unresolved when we actually have a service name. For example, let's consider an HTTP server call for which we don't know the names, e.g. it's running in a docker container. When we report the OTel HTTP metrics the service name would be something like "java" which is also what's in "target_info". However, since we never considered the name in name_resolver, we would call the "client" "unresolved" and the server "unresolved", while it should be "unresolved" -> "java".

I made a few fixes to address the issues above:
1. I split the "unresolved" into 3 parts: generic unresolved, incoming and outgoing. Incoming unresolved clients are now called "incoming", the outgoing "hosts" are called "outgoing". If we can't find the service name we set "unresolved". I'm also matching the namespace of the other "known" party to the "unresolved" calls. This way we will not bundle all namespaces to a single "incoming" and "outgoing" nodes.
2. I fixes the Prom service graphs to use the regular attributes.
3. I added the logic to try the host information embedded on the span for client calls. For server calls we don't know what the client is encoding in the "Host" header, so better not try to resolve this.
4. I made name resolver run even if there are no sources, so that it can at least set the "host" and "service name" information.

### DIFF
--- a/pkg/app/request/metric_attributes.go
+++ b/pkg/app/request/metric_attributes.go
@@ -136,12 +136,20 @@ func SpanPeer(span *Span) string {
 	return span.Peer
 }
 
-func HTTPClientHost(span *Span) string {
+func HostFromSchemeHost(span *Span) string {
 	if strings.Index(span.Statement, SchemeHostSeparator) > 0 {
 		schemeHost := strings.Split(span.Statement, SchemeHostSeparator)
 		if schemeHost[1] != "" {
 			return schemeHost[1]
 		}
+	}
+
+	return ""
+}
+
+func HTTPClientHost(span *Span) string {
+	if host := HostFromSchemeHost(span); host != "" {
+		return host
 	}
 
 	return HostAsServer(span)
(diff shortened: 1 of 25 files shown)


grcevski:
I also made a small fix for Python SSL. Apparently in SSL 3.3 they added SSL_write_ex2 which is what Python started using. When I wrote the integration test I noticed that we were missing the SSL calls.

mariomac (approved):

PR #693, opened 2025-09-26
Add lag metric to kubernetes cache

mariomac:
We observed that in some high-load scenarios, some informer metadata might take too long to get received by the OBI instances. This PR adds two metrics, both in client and cache side, that indicates how long it takes for OBI to get notified and receive about a metadata change since this event actually happens at Kubernetes.


skl (approved):

grcevski (approved):
LGTM!

PR #709, opened 2025-09-29
Prevent pipeline from blocking

rafaelroquetto:
The map tracer pipeline node has been removed, but this has been left behind, preventing the pipeline from cleanly shutting down.


grcevski (approved):
LGTM!

skl (approved):
ty!

PR #723, opened 2025-10-01
Update bpf metric scrape interval

rafaelroquetto:
Gathering ebpf metrics is expensive:

1. it involves multiple system calls when iterating maps
2. it needs to create ebpf.Program objects

On a large cluster with hundreds of ebpf programs, that can take a toll. This PR aims to establish saner defaults.

Below are the CPU and memory flame graphs for reference:

image

image

### DIFF
--- a/pkg/obi/config.go
+++ b/pkg/obi/config.go
@@ -132,7 +132,7 @@ var DefaultConfig = Config{
 			Port: 0, // disabled by default
 			Path: "/internal/metrics",
 		},
-		BpfMetricScrapeInterval: 15 * time.Second,
+		BpfMetricScrapeInterval: 30 * time.Minute,
 	},
 	Attributes: Attributes{
 		InstanceID: traces.InstanceIDConfig{


pkg/obi/config.go
```diff
@@ -132,7 +132,7 @@ var DefaultConfig = Config{
 			Port: 0, // disabled by default
 			Path: "/internal/metrics",
 		},
-		BpfMetricScrapeInterval: 15 * time.Second,
+		BpfMetricScrapeInterval: 30 * time.Minute,
```
mariomac, line 135: I think this is equivalent to disable it. These metrics can be useful especially for debugging purposes, so having to wait 30 minutes for each debug session might be unrealistic.

How about changing this scraping interval to some kind of exponential backoff?

We can add two properties: `InitScrapeInterval` and `MaxScrapeInterval`.

So first metrics are collected after 1 minute, next after 2 minute, next after 4 minutes... and ends up being collected every 30 minutes.

rafaelroquetto:
The performance impact needs to be further investigated.

PR #726, opened 2025-10-01
Few extra code comments to BPF

mariomac:
⚠️⚠️⚠️ Please triple-check the correctness of the assertions in the code canges ⚠️⚠️⚠️

As the BPF code grows in size and complexity, every time it's more difficult to understand about the context of the different pieces. Even for someone involved in the OBI source code (🙋🏻).

A simple unit of business logic business implemented in BPF consists of multiple small programs that are triggered independently on certain system events. To understand the status of a process (especially during debugging) we need to clearly understand the different BPF maps that are defined: not only what do they store but also when the information is updated or retrieved, and under which assumptions.

### DIFF
--- a/bpf/bpfcore/bpf_helpers.h
+++ b/bpf/bpfcore/bpf_helpers.h
@@ -13,7 +13,7 @@
  */
 #include "bpf_helper_defs.h"
 
-#define __uint(name, val) int(*name)[val]
+#define __uint(name, val) int (*name)[val]
 #define __type(name, val) typeof(val) *name
 #define __array(name, val) typeof(val) *name[]
 
--- a/bpf/bpfcore/bpf_tracing.h
+++ b/bpf/bpfcore/bpf_tracing.h
@@ -518,8 +518,8 @@ struct pt_regs;
                 ___bpf_ctx_cast(args));                                                            \
         _Pragma("GCC diagnostic pop")                                                              \
     }                                                                                              \
-    static                                                                                         \
-        __attribute__((always_inline)) typeof(name(0)) ____##name(unsigned long long *ctx, ##args)
+    static __attribute__((always_inline)) typeof(name(0)) ____##name(unsigned long long *ctx,      \
+                                                                     ##args)
 
 struct pt_regs;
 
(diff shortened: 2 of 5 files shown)


bpf/bpfcore/bpf_helpers.h
```diff
@@ -13,7 +13,7 @@
  */
 #include "bpf_helper_defs.h"
 
-#define __uint(name, val) int(*name)[val]
+#define __uint(name, val) int (*name)[val]
```
mariomac:
```suggestion
#define __uint(name, val) int(*name)[val]
```

bpf/bpfcore/bpf_tracing.h
```diff
@@ -518,8 +518,8 @@ struct pt_regs;
                 ___bpf_ctx_cast(args));                                                            \
         _Pragma("GCC diagnostic pop")                                                              \
     }                                                                                              \
-    static                                                                                         \
-        __attribute__((always_inline)) typeof(name(0)) ____##name(unsigned long long *ctx, ##args)
+    static __attribute__((always_inline)) typeof(name(0)) ____##name(unsigned long long *ctx,      \
```
mariomac:
```suggestion
    static                                                                                         \
        __attribute__((always_inline)) typeof(name(0)) ____##name(unsigned long long *ctx, ##args)
```

bpf/bpfcore/bpf_tracing.h
```diff
@@ -520,6 +520,7 @@ struct pt_regs;
     }                                                                                              \
     static                                                                                         \
         __attribute__((always_inline)) typeof(name(0)) ____##name(unsigned long long *ctx, ##args)
+                                                                     ##args)
```
mariomac:
```suggestion
```

bpf/generictracer/protocol_tcp.h
```diff
@@ -265,6 +265,11 @@ SEC("kprobe/tcp")
 int obi_protocol_tcp(void *ctx) {
     (void)ctx;
 
+    // it assumes that the actual protocol_args have been previously set
+    // from another BPF function.
+    // If that's not the case, the connection details might be empty.
+    // If the same thread manages multiple connections at the same thread,
+    // the retrieved data might end up being wrong.
```
rafaelroquetto: I think we are actually fine here. This is part of a tail-call chain, so the current thread is currently inside the kernel (or blocked waiting for the kernel to complete) before it can service another connection.

PR #730, opened 2025-10-01
Introduce Queue.TrySend

rafaelroquetto:
When pushing objects into a queue, `Queue.Send()` will panic if the queue is full. This is done to expose programming logic errors (such as cyclic dependencies).

There are scenarios though in which panicking may not be desirable: for instance, when too many events are pushing into the queue, causing it to fill and then block. In this case, the `TrySend()` function returns `false` and gives the caller the ability to handle this scenario without crashing the application.

This PR uses this function to prevent the network observability tracer from crashing / panicking when there are too many flows (due to a bad sampling configuration) that the pipeline can handle: in this case, the flows are simply dropped without any repercussion.

A log message is printed once (to avoid flooding the logs) to warn the user.

### DIFF
--- a/pkg/components/netolly/flow/tracer_ringbuf.go
+++ b/pkg/components/netolly/flow/tracer_ringbuf.go
@@ -74,6 +74,8 @@ func (m *RingBufTracer) TraceLoop(out *msg.Queue[[]*ebpf.Record]) swarm.RunFunc 
 		// event throughput and the memory pressure
 		flows := make([]*ebpf.Record, 0, 1048576)
 
+		printLog := true
+
 		for {
 			select {
 			case <-ctx.Done():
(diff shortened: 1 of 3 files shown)


pkg/components/netolly/flow/tracer_ringbuf.go
```diff
@@ -74,14 +74,25 @@ func (m *RingBufTracer) TraceLoop(out *msg.Queue[[]*ebpf.Record]) swarm.RunFunc
...
 				return
 			case <-ticker.C:
 				if len(flows) > 0 {
-					out.Send(append([]*ebpf.Record(nil), flows...))
+					// this will drop the flows if the queue is full / stalled
+					// when that happens, it is safe to assume that no
+					// important information is lost, as this pretty much
+					// works as a fallback sampling mechanism
+					if !out.TrySend(append([]*ebpf.Record(nil), flows...)) && printLog {
+						// TODO: perhaps print drop stats?
```
grcevski, line 91: Internal metrics for dropped flows?

rafaelroquetto: That's a good idea. I will handle this on a separate PR if this gets merged.

rafaelroquetto:
Not needed for now due to #732. I might reopen in the future.

PR #735, opened 2025-10-03
Allow overriding Network Export channel

mariomac:
Analogously to the `OverrideAppExportChannel` option in the context info, we are providing an analogous queue to allow providing more network data exporters.


skl (commented):
It's hard to tell if this is covered by unit tests already as the coverage report is broken, I'll push a separate PR (#736) to fix the coverage reporting from unit tests.

Is it a lot of work to a add a unit test for the new net export queue override?

grcevski (approved):
LGTM!

PR #737, opened 2025-10-03
Report failed connects

grcevski:
This PR adds probe and logic to track failed connections for our kprobes based instrumentation. Since our kprobes based instrumentation rely on various levels of protocol detection (eBPF or user-space), if we fail to connect we are oblivious to something bad happening, unless the parent transaction fails.

I've added two separate mechanisms for detection of failed connects:

1. A kprobe on `sock_def_error_report`, which will trigger every time a TCP socket gets into an error state.
2. The approach one doesn't work with proxies, essentially the proxy allows the connection but it never gets established, although it will get ports and IP address. For this kind of disconnect, I added logic in `tcp_close` to detect it and report the event.


mariomac (approved):
This is great!

bpf/common/sockaddr.h
```diff
@@ -134,3 +136,16 @@ parse_sockaddr_info(u32 pid, struct sockaddr *addr, connection_info_part_t *info
 
     return false;
 }
+
+static __always_inline bool is_socket_never_connected(struct sock *sk) {
```
mmat11: Is this a best-effort result? I am not sure those 2 states always represent an unconnected socket, or if there are more states in which the socket could be unconnected.

I remember doing something like this in the past:
```
    struct tcp_sock *tp = (struct tcp_sock *)sk;
    u64 bytes_sent      = BPF_CORE_READ(tp, bytes_sent);
    u64 bytes_received  = BPF_CORE_READ(tp, bytes_received);
    // if bytes_sent == 0 && bytes_received == 0 ==> unconnected
```
but this is also not perfect as there might be some bytes sent and the connection would still be not established. 🤔

If this has worked for you locally/in the tests we can leave it as is and re-evaluate in the future if we happen to hit edge cases, etc.

bpf/generictracer/k_tracer.c
```diff
@@ -568,6 +599,38 @@ int BPF_KPROBE(obi_kprobe_tcp_close, struct sock *sk, long timeout) {
     return 0;
 }
 
+SEC("kprobe/sk_error_report")
```
mmat11: I don't see this function in older kernels (5.8/5.10), should we have a runtime-selected alternative? (I'm also not sure on what we could use alternatively)

mmat11: Oh nevermind, I see the function is being attached to `sock_def_error_report`, so we should just rename the probe names 👍🏻

bpf/common/connection_info.h
```diff
@@ -18,6 +18,7 @@ enum protocol_type : u8 {
     k_protocol_type_unknown = 0,
     k_protocol_type_mysql = 1,
     k_protocol_type_postgres = 2,
+    k_protocol_type_connect_failed = 3,
```
mmat11: this is more of a metadata than an application protocol, perhaps is better to keep the field separated in `tcp_req_t`?

mmat11 (approved):
Other than a couple comments, lgtm, great stuff!

bpf/common/connection_info.h
grcevski: Good point, I'll change that.

bpf/common/sockaddr.h
grcevski: Yeah it's sort of best effort. I like your suggestion though, I'll add that logic and see how it works. I had a previous version with TCP_CLOSE too, but I found it gave me false positives.

bpf/generictracer/k_tracer.c
grcevski: Good catch, I forgot to rename the helper in the code, so it's confusing.

bpf/common/sockaddr.h
grcevski: I added your suggestion as a secondary check if the simple flags check doesn't work, thanks again!

Issue #449, opened 2025-08-13
Evaluate what we can move internal before initial release


MrAlias:
Try to minimize the Go API and move what does not need to be exported internal.

Example of dependence to use in audit: https://github.com/search?q=repo%3Agrafana%2Fbeyla%20go.opentelemetry.io%2Fobi&type=code

MrAlias:
From within [Beyla](https://github.com/grafana/beyla)

```console
> grep -Rn 'go.opentelemetry.io/obi' | grep -v '^\(vendor\|go.mod\|third_party_licenses.csv\|.golangci.yml\|Makefile\|README.md\)' | sed -n 's/.*"\(go.opentelemetry.io\/obi[^"]*\)"/\1/p' | sort -u
grep: .git/index: binary file matches
go.opentelemetry.io/obi/pkg/app/request
go.opentelemetry.io/obi/pkg/buildinfo
go.opentelemetry.io/obi/pkg/components/connector
go.opentelemetry.io/obi/pkg/components/discover
go.opentelemetry.io/obi/pkg/components/ebpf
go.opentelemetry.io/obi/pkg/components/ebpf/common
go.opentelemetry.io/obi/pkg/components/ebpf/tcmanager
go.opentelemetry.io/obi/pkg/components/exec
go.opentelemetry.io/obi/pkg/components/helpers
go.opentelemetry.io/obi/pkg/components/imetrics
go.opentelemetry.io/obi/pkg/components/kube
go.opentelemetry.io/obi/pkg/components/netolly/agent
go.opentelemetry.io/obi/pkg/components/netolly/flow
go.opentelemetry.io/obi/pkg/components/netolly/transform/cidr
go.opentelemetry.io/obi/pkg/components/pipe
go.opentelemetry.io/obi/pkg/components/pipe/global
go.opentelemetry.io/obi/pkg/components/svc
go.opentelemetry.io/obi/pkg/components/traces
go.opentelemetry.io/obi/pkg/export/attributes
go.opentelemetry.io/obi/pkg/export/attributes/names
go.opentelemetry.io/obi/pkg/export/debug
go.opentelemetry.io/obi/pkg/export/expire
go.opentelemetry.io/obi/pkg/export/instrumentations
go.opentelemetry.io/obi/pkg/export/otel
go.opentelemetry.io/obi/pkg/export/otel/metric
go.opentelemetry.io/obi/pkg/export/otel/metric/api/metric
[...]

PR #751, opened 2025-10-09
Hiding the visibility of some functions and types

mariomac:
Partially handles #449  :

*  Moving `pkg/components/traces` to `pkg/traces` and `pkg/internal/traces`
* Moving `pkg/components/netolly` to  `pkg/netolly` and `pkg/internal/netolly`

### DIFF
--- a/pkg/components/pipe/instrumenter_test.go
+++ b/pkg/components/pipe/instrumenter_test.go
@@ -25,6 +25,7 @@ import (
 	"go.opentelemetry.io/obi/pkg/components/kube"
 	"go.opentelemetry.io/obi/pkg/components/pipe/global"
 	"go.opentelemetry.io/obi/pkg/components/svc"
+	"go.opentelemetry.io/obi/pkg/config"
 	"go.opentelemetry.io/obi/pkg/export/attributes"
 	attr "go.opentelemetry.io/obi/pkg/export/attributes/names"
 	"go.opentelemetry.io/obi/pkg/export/instrumentations"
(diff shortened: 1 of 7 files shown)


pkg/traces/tracescfg/tracescfg.go
```diff
@@ -0,0 +1,16 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package tracescfg
```
mmat11: nit: maybe we can merge this package with `traces` and put this in `config.go` to reduce noise?

pkg/traces/tracescfg/tracescfg.go
```diff
@@ -0,0 +1,16 @@
...
+
+// InstanceIDConfig configures how Beyla will get the Instance ID of the traces/metrics
```
mmat11:
```suggestion
// InstanceIDConfig configures how OBI will get the Instance ID of the traces/metrics
```

pkg/traces/tracescfg/tracescfg.go
```diff
@@ -0,0 +1,16 @@
...
+// from the current hostname + the instrumented process PID
+type InstanceIDConfig struct {
+	// HostnameDNSResolution is true if Beyla uses the DNS to resolve the local hostname or
```
mmat11:
```suggestion
	// HostnameDNSResolution is true if OBI uses the DNS to resolve the local hostname or
```

pkg/traces/tracescfg/tracescfg.go
```diff
@@ -0,0 +1,16 @@
...
+	// false if it uses the local hostname.
+	HostnameDNSResolution bool `yaml:"dns" env:"OTEL_EBPF_HOSTNAME_DNS_RESOLUTION"`
+	// OverrideHostname can be optionally set to avoid resolving any hostname and using this
+	// value. Beyla will anyway attach the process ID to the given hostname for composing
```
mmat11:
```suggestion
	// value. OBI will anyway attach the process ID to the given hostname for composing
```

pkg/traces/tracescfg/tracescfg.go
mariomac: Good observation!

Issue #141, opened 2025-06-11
Post processing of Http events to more specific protocols


NimrodAvni78:
Many protocols are built on top of http, like the ones used for graphql, elasticseach, s3, sqs and more

What Im suggesting trying to narrow down a Http span into a more specific type of span (like those above)
this can be done with simple post processing of http events in userspace.
it should be configurable, because this will need to rely on some heuristics like:
- port numbers (9200 is probably elastic)
- route conventions ( /.../_search is probably an elastic search query)
- common headers (x-amz-content-sha256 will probably be included in requests to aws s3 / sqs)
and maybe more

let me know if you have any thoughts regarding specific concerns in the implementations, because i think it can definitely bring a lot of value.

NimrodAvni78:
Closing this, already implemented in many post-processing of http events like #827 #801 #1210 and more

PR #755, opened 2025-10-09
http: add support for GraphQL attributes

mmat11:
part of https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/141

### DIFF
--- a/pkg/app/request/span.go
+++ b/pkg/app/request/span.go
@@ -247,7 +247,7 @@ func spanAttributes(s *Span) SpanAttributes {
 		}
 		return attrs
 	case EventTypeHTTPClient:
-		attrs := SpanAttributes{
+		return SpanAttributes{
 			"method":     s.Method,
 			"status":     strconv.Itoa(s.Status),
 			"url":        s.Path,
@@ -255,7 +255,6 @@ func spanAttributes(s *Span) SpanAttributes {
 			"serverAddr": SpanHost(s),
 			"serverPort": strconv.Itoa(s.HostPort),
 		}
-		return attrs
 	case EventTypeGRPC:
 		return SpanAttributes{
 			"method":     s.Path,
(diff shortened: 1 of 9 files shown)


pkg/app/request/span.go
```diff
@@ -76,6 +76,11 @@ const (
 	DBMySQL
 )
 
+const (
+	HTTPSubtypeNone    = 0 // http
+	HTTPSubtypeGraphQL = 1 // http + graphql
+)
+
```
mariomac, line 83: For more safety, I'd create our own type:

```golang
type Subtype int
const (
    SubtypeNone = HTTPSubType(0) // generic: http, grpc, etc...
    SubtypeHTTPGraphQL = HTTPSubType(1) // http + graphql
)
```

Then also change the type of `span->Subtype` to this.

bpf/generictracer/k_tracer.c
```diff
@@ -1140,6 +1140,13 @@ int obi_handle_buf_with_args(void *ctx) {
                         }
                     }
                 }
+
+                http_send_large_buffer(info,
+                                       (void *)args->u_buf,
+                                       args->bytes_len,
+                                       args->packet_type,
+                                       args->direction,
+                                       k_large_buf_action_append);
```
mariomac, line 1149: Just to know (and support/document accordingly). Do you think that a current OBI user might get a performance penalty after upgrading to this change? e.g. in HTTP requests whose subtype can't be identified at BPF side.

mmat11: I don't think so, `http_send_large_buffer` always checks if large buffers are enabled at the start:

```
    if (http_buffer_size == 0) {
        return 0;
    }
```

so a current user with buffers disabled would see no change in terms of performance

pkg/app/request/span.go
mmat11: Mmm I think this would imply converting to this new type all the other subtypes (db, gpu, etc.). Regarding the GPU one, it's not even clear what it needs to be because it uses event.Kind (which could change regardless).
You mind if I tackle this in a followup PR (once I better understand the gpu event)?

mariomac: Ok! No worries, it was not something super important. We can do a later refactor/improvement.

PR #759, opened 2025-10-10
[chore] apply modernize

florianl:
This is an automated change done by:
```
go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -fix -test ./...
```

### DIFF
--- a/pkg/internal/appolly/appolly.go
+++ b/pkg/internal/appolly/appolly.go
@@ -153,7 +153,7 @@ func (i *Instrumenter) instrumentedEventLoop(ctx context.Context, processEvents 
 			log.Debug("running tracer for new process",
 				"inode", pt.FileInfo.Ino, "pid", pt.FileInfo.Pid, "exec", pt.FileInfo.CmdExePath)
 			if pt.Tracer != nil {
-				&{i tracersWg}.Go(func() {
+				i.tracersWg.Go(func() {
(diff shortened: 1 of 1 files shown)


pkg/components/discover/watcher_proc_test.go
```diff
@@ -233,7 +233,7 @@ func TestPortsFetchRequired(t *testing.T) {
 	// we send port that's not in our port range
 	eventsChan <- watcher.Event{Type: watcher.NewPort, Payload: 8090}
 	// 5 seconds should be enough to have the channel send something
-	for i := 0; i < 5; i++ {
+	for range 5 {
```
mariomac, line 236: Didn't know this form of `range`. Nice!

pkg/components/helpers/container/container.go
```diff
@@ -61,7 +61,7 @@ func InfoForPID(pid uint32) (Info, error) {
 		return Info{}, fmt.Errorf("reading %s: %w", cgroupFile, err)
 	}
 
-	for _, cgroupEntry := range bytes.Split(cgroupBytes, []byte{'\n'}) {
+	for cgroupEntry := range bytes.SplitSeq(cgroupBytes, []byte{'\n'}) {
```
mariomac, line 64: This is a nice performance optimization. I've been looking for replacing some array generation by sequence, but I was not aware that `Split` already provides a sequence-based form.

PR #767, opened 2025-10-13
Handle receiving large HTTPS payloads

grcevski:
So I encountered this particular issue which caused a second nested call in distributed traces to not be correctly nested, but I think the problem may manifest itself differently.

Essentially, HTTPS requests are handled slightly differently than HTTP. We delay the HTTP requests until we see a reversal in the TCP traffic (or TCP close), so that we can capture the full request time (with responding large bytes), but for HTTPS we could never find a reliable way to know when the request finishes. Sometimes we never see the reversed traffic, so if we don't terminiate the request on the status like (e.g. 200 OK) we may never record the request. SSL_close or SSL_shutdown are not reliably called by frameworks.

This is the difference in the http_info->delayed flag handling.

Because we terminate the HTTPS request right away, we end up cleaning the HTTP info right away, along with the SSL to connection relationship. This causes 2 problems:
1. The connection information is not there and we end up starting from scratch for the follow-up data, but often with incorrect connection information.
2. We end up making TCP requests for the follow-up packets, since we have no way to know anymore that this was HTTPS request. This now with combination of wrong connection information will break context propagation, and may send TCP traffic to be handled by userspace, which can result in unpredictable events being parsed.

I'm attempting a change that leaves the HTTPS request behind and marks it as submitted if it wasn't delayed. This is effectively the same as finishing the request and deleting it, except we keep the metadata alive to know this was valid HTTP request.

### DIFF
--- a/bpf/generictracer/ssl_defs.h
+++ b/bpf/generictracer/ssl_defs.h
@@ -38,10 +38,6 @@ static __always_inline void cleanup_ssl_trace_info(http_info_t *info, void *ssl)
             delete_server_trace(&ssl_info->p_conn, &t_key);
         }
     }
-
-    // We can't delete this here. What if it's a large SSL request and we keep on
-    // receiving data after we've seen 200 OK?
-    //bpf_map_delete_elem(&ssl_to_conn, &ssl);
 }
 
 static __always_inline void


bpf/common/ssl_helpers.h
```diff
@@ -73,6 +73,6 @@ connect_ssl_to_connection(u64 id, pid_connection_info_t *conn, u8 direction, u16
     set_active_ssl_connection(&ssl_conn, ssl);
 }
 
-static __always_inline void *is_ssl_connection(pid_connection_info_t *conn) {
-    return bpf_map_lookup_elem(&active_ssl_connections, conn);
+static __always_inline u64 *is_ssl_connection(pid_connection_info_t *conn) {
```
grcevski, line 76: Having this as void * caused a nasty bug, we were getting technically a pointer to the u64, but we were later querying with & of the result of this call, instead of the value directly.

bpf/generictracer/k_send_receive.h
```diff
@@ -27,3 +27,16 @@ static __always_inline void ensure_sent_event(u64 id, u64 *sock_p) {
         finish_possible_delayed_http_request(&s_args->p_conn);
     }
 }
+
+static __always_inline void force_sent_event(u64 id, u64 *sock_p) {
```
grcevski, line 31: I added this function because I found that if we managed to connect to a server, but we never get the response, we'll not record a failed request, e.g. the server dies. This is now used by TCP close to force push the failed event.

bpf/generictracer/protocol_http.h
```diff
@@ -257,8 +263,28 @@ static __always_inline void finish_http(http_info_t *info, pid_connection_info_t
...
+static __always_inline void force_finish_http(http_info_t *info, pid_connection_info_t *pid_conn) {
+    if (info->submitted) {
+        return;
+    }
+
+    if (!high_request_volume) {
+        if (!http_info_complete(info)) {
+            info->resp_len = 0;
+            info->end_monotime_ns = bpf_ktime_get_ns();
+            info->status = 409;
```
grcevski: I'm setting the status as 499 when we haven't yet completed the request but we are pushing another. This is typically caused by a client timeout and retry and NGINX sets 499 (although it's not officially set)

bpf/generictracer/protocol_http.h
```diff
@@ -257,8 +263,28 @@ static __always_inline void finish_http(http_info_t *info, pid_connection_info_t
...
+static __always_inline void force_finish_http(http_info_t *info, pid_connection_info_t *pid_conn) {
+    if (info->submitted) {
+        return;
+    }
+
+    if (!high_request_volume) {
+        if (!http_info_complete(info)) {
+            info->resp_len = 0;
+            info->end_monotime_ns = bpf_ktime_get_ns();
+            info->status = 499;
```
grcevski, line 283: Setting the status as 499 just like NGINX when we have a prior request on the same connection waiting uncompleted, most likely cause is client timeout and retry.

bpf/logger/bpf_dbg.h
```diff
@@ -31,7 +31,7 @@ typedef struct log_info {
 
 struct {
     __uint(type, BPF_MAP_TYPE_RINGBUF);
-    __uint(max_entries, 1 << 12);
+    __uint(max_entries, 1 << 15);
```
grcevski, line 34: I increased the debugger ringbuffer, because I ended up having to debug with the stdout logs from tests and a lot of messages were missing with 4K. It shouldn't increase the overall footprint, since this is loaded only with OTEL_EBPF_BPF_DEBUG=1

test/integration/docker-compose-multiexec-host.yml
```diff
@@ -146,7 +146,6 @@ services:
       OTEL_EBPF_BPF_DISABLE_BLACK_BOX_CP: "${OTEL_EBPF_BPF_DISABLE_BLACK_BOX_CP}"
       OTEL_EBPF_BPF_CONTEXT_PROPAGATION: "${OTEL_EBPF_BPF_CONTEXT_PROPAGATION}"
       OTEL_EBPF_BPF_TRACK_REQUEST_HEADERS: "${OTEL_EBPF_BPF_TRACK_REQUEST_HEADERS}"
-      OTEL_EBPF_BPF_HIGH_REQUEST_VOLUME: 1
```
grcevski, line 149: I removed this because it's a source of flakiness, it shouldn't have been added to the test that looks for correct nesting of transactions. in this high request mode we terminate requests immediately, meaning we can still be receiving packets and the cleanup of the connection information will cause the correlation logic to fail.

bpf/generictracer/k_send_receive.h
```diff
@@ -27,3 +27,16 @@ static __always_inline void ensure_sent_event(u64 id, u64 *sock_p) {
...
+    send_args_t *s_args = (send_args_t *)bpf_map_lookup_elem(&active_send_args, &id);
+    if (s_args) {
+        bpf_dbg_printk("Checking if we need to finish the request per thread id");
+        force_finish_possible_delayed_http_request(&s_args->p_conn);
```
mmat11, line 35: does this need a `return`?

bpf/generictracer/k_tracer.c
```diff
@@ -413,7 +413,7 @@ int BPF_KPROBE(obi_kprobe_tcp_sendmsg, struct sock *sk, struct msghdr *msg, size
             return 0;
         }
 
-        tcp_send_ssl_check(id, ssl, &s_args.p_conn, orig_dport);
+        tcp_send_ssl_check(id, (void *)(*ssl), &s_args.p_conn, orig_dport);
```
mmat11, line 429: question: wouldn't it be enough to just cast to void without dereferencing?

bpf/generictracer/k_tracer.c
```diff
@@ -482,7 +482,7 @@ int BPF_KPROBE(obi_kprobe_tcp_rate_check_app_limited, struct sock *sk) {
             }
         } else {
             make_inactive_sk_buffer(&s_args.p_conn.conn);
-            tcp_send_ssl_check(id, ssl, &s_args.p_conn, orig_dport);
+            tcp_send_ssl_check(id, (void *)(*ssl), &s_args.p_conn, orig_dport);
```
mmat11, line 511: same here

bpf/generictracer/ssl_defs.h
```diff
@@ -39,7 +39,9 @@ static __always_inline void cleanup_ssl_trace_info(http_info_t *info, void *ssl)
         }
     }
 
-    bpf_map_delete_elem(&ssl_to_conn, &ssl);
+    // We can't delete this here. What if it's a large SSL request and we keep on
```
mmat11: does this comment need to stay for future reference or is this a leftover?

bpf/generictracer/k_send_receive.h
grcevski: good point, we are trying to clean-up everything possible. If we find the same connection in both places, the second will be no-op, but perhaps if I treat it as else we might miss certain cleanup. I'd like to leave it as is for now.

bpf/generictracer/k_tracer.c
grcevski: I think that was my original bug to be honest. The tcp_send_ssl_check does &ssl for the lookup. We pull the ssl from a map, at that point it's a pointer to the address where the SSL is, so tcp_send_ssl_check should lookup with `ssl` not with `&ssl`, but the helper is used elsewhere. So I need to derefence it to get the actual ssl * (which is stored as u64 in the map) and then the lookup will happen with & of that value.

bpf/generictracer/k_send_receive.h
mmat11: sounds good!

bpf/generictracer/ssl_defs.h
grcevski: good point, let me remove it. it will also run the tests one more time to see if passing was a fluke :)

bpf/generictracer/k_tracer.c
mmat11: thanks for the explanation!

PR #768, opened 2025-10-13
Integrate go-playground/validator for improved configuration validation

pinoOgni:
Hi everyone.

This PR is the first one in introducing a more standardized configuration validation using the go-playground/validator pkg which is a consistent and declarative way to validate structs using tags so it's easier to read, extend and test it.

Instead of creating a single large PR, I broke it down into smaller parts to gather feedback from the community and evaluate if it makes sense to continue. 

In this first part, I've implemented the validation for the `EBPFTracer` and `NetworkConfig` structs.

I have some doubts about some reference values ​​used in some validations.
Finally, I had to change a bit the default configuration taking into account the new validation constraints.

Thanks, 
Pino

### DIFF
--- a/pkg/config/ebpf_tracer.go
+++ b/pkg/config/ebpf_tracer.go
@@ -14,8 +14,8 @@ import (
 type ContextPropagationMode uint8
 
 type RedisDBCacheConfig struct {
-	Enabled bool `yaml:"enabled" env:"OTEL_EBPF_BPF_REDIS_DB_CACHE_ENABLED"`
-	MaxSize int  `yaml:"max_size" env:"OTEL_EBPF_BPF_REDIS_DB_CACHE_MAX_SIZE"`
+	Enabled bool `yaml:"enabled" env:"OTEL_EBPF_BPF_REDIS_DB_CACHE_ENABLED" validate:"boolean"`
+	MaxSize int  `yaml:"max_size" env:"OTEL_EBPF_BPF_REDIS_DB_CACHE_MAX_SIZE" validate:"gt=0"`
 }
 
 const (
@@ -25,12 +25,10 @@ const (
 	ContextPropagationDisabled
 )
 
-const bufferSizeMax = 8192
-
 // EBPFTracer configuration for eBPF programs
 type EBPFTracer struct {
 	// Enables logging of eBPF program events
-	BpfDebug bool `yaml:"bpf_debug" env:"OTEL_EBPF_BPF_DEBUG"`
+	BpfDebug bool `yaml:"bpf_debug" env:"OTEL_EBPF_BPF_DEBUG" validate:"boolean"`
 
 	// WakeupLen specifies how many messages need to be accumulated in the eBPF ringbuffer
 	// before sending a wakeup request.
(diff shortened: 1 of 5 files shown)


pkg/config/ebpf_tracer.go
```diff
@@ -114,70 +112,28 @@ type EBPFTracer struct {
 // Max: 8192 bytes.
 // Default: 0 (disabled).
 type EBPFBufferSizes struct {
-	HTTP     uint32 `yaml:"http" env:"OTEL_EBPF_BPF_BUFFER_SIZE_HTTP"`
-	MySQL    uint32 `yaml:"mysql" env:"OTEL_EBPF_BPF_BUFFER_SIZE_MYSQL"`
-	Postgres uint32 `yaml:"postgres" env:"OTEL_EBPF_BPF_BUFFER_SIZE_POSTGRES"`
+	HTTP     uint32 `yaml:"http" env:"OTEL_EBPF_BPF_BUFFER_SIZE_HTTP" validate:"lt=8192"`
```
NimrodAvni78: i think this only accepts powers of 2, right @mmat11 ?
in that case will it be correct to create a custom validator or something?

pkg/obi/network_cfg.go
```diff
@@ -22,106 +22,113 @@
...
+	Print bool `yaml:"print_flows" env:"OTEL_EBPF_NETWORK_PRINT_FLOWS" validate:"boolean"`
 
 	// CIDRs list, to be set as the "src.cidr" and "dst.cidr"
 	// attribute as a function of the source and destination IP addresses.
 	// If an IP does not match any address here, the attributes won't be set.
 	// If an IP matches multiple CIDR definitions, the flow will be decorated with the
 	// narrowest CIDR. By this reason, you can safely add a 0.0.0.0/0 entry to group there
 	// all the traffic that does not match any of the other CIDRs.
-	CIDRs cidr.Definitions `yaml:"cidrs" env:"OTEL_EBPF_NETWORK_CIDRS" envSeparator:","`
+	CIDRs cidr.Definitions `yaml:"cidrs" env:"OTEL_EBPF_NETWORK_CIDRS" envSeparator:"," validate:"omitempty,dive,cidr"`
```
NimrodAvni78, line 131: just a question for these enum style values, when i add a new option i need to add it both in the code and in the validation, is there some way to take all valid enum values dynamically?

pkg/obi/network_cfg.go
```diff
@@ -134,10 +141,33 @@ var DefaultNetworkConfig = NetworkConfig{
 	Deduper:            flowdef.DeduperFirstCome,
 	Direction:          "both",
 	ListenInterfaces:   "watch",
-	ListenPollPeriod:   10 * time.Second,
+	// ListenPollPeriod:   10 * time.Second,
```
NimrodAvni78: why did you comment this out?

pkg/config/ebpf_tracer.go
mmat11: This is no longer the case, that restriction has been lifted in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/631

pkg/obi/network_cfg.go
pinoOgni: I hope I understood the question. If for "enum style values" you mean like "dive, cidr" and so on, the only way that I know of to add new ones is to create a custom tag and register it. For example the `reverseDnsCacheLen` tag.

pkg/obi/network_cfg.go
pinoOgni: Because the `ListenPollPeriod` is used only if `ListenInterfaces` is `poll`. In the old configuration the type was `watch` so I commented `ListenPollPeriod` to pass the validation. Technically the default configuration was wrong but the control was done later in the `monitorMode()` function, but imho it's wrong to write `ListenPollPeriod` even if the `ListInterfaces` is `watch`. I used the same approach for `ReverseDNS` struct where the default value for `Type` is `none` and  I commented `CacheLen` and `CacheTTL` because they only apply if `Type` is `local` or `ebpf`.

pkg/config/ebpf_tracer.go
```diff
@@ -14,8 +14,8 @@ import (
 type ContextPropagationMode uint8
 
 type RedisDBCacheConfig struct {
-	Enabled bool `yaml:"enabled" env:"OTEL_EBPF_BPF_REDIS_DB_CACHE_ENABLED"`
-	MaxSize int  `yaml:"max_size" env:"OTEL_EBPF_BPF_REDIS_DB_CACHE_MAX_SIZE"`
+	Enabled bool `yaml:"enabled" env:"OTEL_EBPF_BPF_REDIS_DB_CACHE_ENABLED" validate:"boolean"`
```
mariomac, line 17: Isn't `validate:"boolean"` redundant if the type of the variable is already `bool`? If you try to set something different in the YAML or the env variables, I would expect that the marshaller raises an error before the validation is invoked.

pkg/config/ebpf_tracer.go
```diff
@@ -14,8 +14,8 @@ import (
...
+	MaxSize int  `yaml:"max_size" env:"OTEL_EBPF_BPF_REDIS_DB_CACHE_MAX_SIZE" validate:"required_if=Enabled true,gte=0"`
```
mariomac: Question, if you set `enabled: true` and leave this property un set, how would it behave?
1. Validation raises an error because `max_size` is not set in the configuration.
2. Everything is OK because `max_size` is defaulted to 1000, so the validation will count this value as set.

mariomac: Also, shouldn't the validation be `gt` instead of `gte`? Sounds strange that a max size is zero.

pkg/config/ebpf_tracer.go
mariomac: I think it should be `lte`.

pkg/obi/network_cfg.go
mariomac: I think we still should leave the value uncommented. So if the user sets `ListenInterfaces` to `poll`, the default poll period would apply without forcing the user to set a value.

pkg/obi/network_cfg.go
```diff
@@ -134,10 +141,18 @@ var DefaultNetworkConfig = NetworkConfig{
...
 	ReverseDNS: flow.ReverseDNS{
-		Type:     flow.ReverseDNSNone,
-		CacheLen: 256,
-		CacheTTL: time.Hour,
+		Type: flow.ReverseDNSNone,
+		// CacheLen: 256,
+		// CacheTTL: time.Hour,
```
mariomac: Same as before. We should uncomment the values even if they are not useful in the default configuration, so if the users set a reverse DNS type, they don't need to configure the cache but leave the default values.

pkg/config/ebpf_tracer.go
pinoOgni: Yes, it's redundant because the field is `bool`. I wrote it just for consistency and to make the field's semantic intent explicit. The validation is self-documented and consistent even if the filed type changes in the future (because the definition of the tag is `This validates that a string value can successfully be parsed into a boolean with strconv.ParseBool`).
If you want I can remove it.

pkg/obi/network_cfg.go
pinoOgni: Yes, I forgot to uncomment it in the new push. Now the validation for `ListenPollPeriod` is only `gte=0`

pkg/obi/network_cfg.go
pinoOgni: Same as before, I forgot to uncomment it in the new push.

pkg/config/ebpf_tracer.go
pinoOgni: > Question, if you set `enabled: true` and leave this property un set, how would it behave?
> 
> 1. Validation raises an error because `max_size` is not set in the configuration.
> 2. Everything is OK because `max_size` is defaulted to 1000, so the validation will count this value as set.

Yes, without the default config we would be in the first case `Key: 'User.MaxSize' Error:Field validation for 'MaxSize' failed on the 'required_if' tag` but we have the default config, so yes, everything is fine.

> Also, shouldn't the validation be `gt` instead of `gte`? Sounds strange that a max size is zero.

Yep. Anyway I think we should decide which values are strange. For example it's strange to me that some duration like `CacheActiveTimeout` can be 0.

mariomac: I think that users should not be forced to set `max_size`. This looks to me as a fine-tuning parameter that most users should not touch (or don't have to know what to set). So I would disable that validation. If `enabled: true`, max_size will default to 1000 so users don't have to set it.

pinoOgni: fine by me

PR #770, opened 2025-10-16
bpf: make http split buffers big enough to store large payloads

mmat11:
This is a follow-up for https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/750 , this PR copies a bigger portion of the buffer in a per-cpu internal pinned map shared between tp-injector and generictracer. Since the data is no longer on the stack, the size can be increased to 8k which should cover most use cases.

Given the fact there could be CPU migration between the two programs, the old buffer is kept as fallback and the CPU id is passed as a guard

### DIFF
--- a/bpf/generictracer/k_tracer.c
+++ b/bpf/generictracer/k_tracer.c
@@ -369,8 +369,8 @@ int BPF_KPROBE(obi_kprobe_tcp_sendmsg, struct sock *sk, struct msghdr *msg, size
                         if (m_buf) {
                             u32 cpu_id = bpf_get_smp_processor_id();
                             if (m_buf->cpu_id != cpu_id) {
-                                bpf_dbg_printk(
-                                    "cpu id mismatch, using stack-allocated fallback buffer");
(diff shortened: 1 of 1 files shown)


bpf/common/msg_buffer.h
```diff
@@ -4,14 +4,34 @@
...
 
 #include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_helpers.h>
 
 #include <common/http_buf_size.h>
+#include <common/pin_internal.h>
+
+enum {
+    k_msg_buffer_size_max = 8192,
+    k_msg_buffer_size_max_mask = k_msg_buffer_size_max - 1,
```
mariomac, line 14: Non-C-coder question: why do we define an enum for defining a constant? Instead of a `#define ...`?

mmat11: Full explanation here https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/118#discussion_r2129028261

mariomac: Amazing! Thanks for pointing to the original reason 👏🏻

PR #786, opened 2025-10-21
Max transaction time

grcevski:
In certain scenarios, for example an HTTP started background task, we can get an infinite trace, e.g. trace that may last for hours or even days. One example is the OTel DEMO load generator. A thread just keeps on sending client requests in a loop.

These traces cause problem for databases and for rendering, essentially they are massive and not useful.

This PR adds a maximum transaction time option which specifies what we expect to be a reasonable time a transaction may run. I've set the default to 5 minutes. A parent child relationship that spans more than 5 minutes is deemed unacceptable and the child gets a new transaction.

### DIFF
--- a/test/integration/traces_test.go
+++ b/test/integration/traces_test.go
@@ -999,13 +999,9 @@ func testHTTPTracesNestedCallsTooLong(t *testing.T) {
 	// Run a request, since we have a single app, we should see always all requests
 	doHTTPGet(t, "http://localhost:7773/slow", 200)
 
-	// Give time for all spans to make it to jeager, we want to ensure the children
-	// also make it and are not nested properly
-	time.Sleep(5 * time.Second)
-
 	var trace jaeger.Trace
 	test.Eventually(t, testTimeout, func(t require.TestingT) {
-		resp, err := http.Get(jaegerQueryURL + "?service=python-self&operation=GET%20%2Fslow")
+		resp, err := http.Get(jaegerQueryURL + "?service=python-self&operation=GET%20%2Fsmoke1")
 		require.NoError(t, err)
 		if resp == nil {
 			return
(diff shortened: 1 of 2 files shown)


test/integration/traces_test.go
```diff
@@ -993,6 +993,65 @@ func testNestedHTTPSTracesKProbes(t *testing.T) {
...
 
+func testHTTPTracesNestedCallsTooLong(t *testing.T) {
+	var parentID string
+
+	// Run a request, since we have a single app, we should see always all requests
+	doHTTPGet(t, "http://localhost:7773/slow", 200)
+
+	// Give time for all spans to make it to jeager, we want to ensure the children
+	// also make it and are not nested properly
+	time.Sleep(5 * time.Second)
```
mariomac: This will slow down the test. Can we query inside the `Eventually` clause below that also the child spans are in jaeger, before we verify that they are disconnected?

test/integration/traces_test.go
```diff
@@ -993,6 +993,65 @@ func testNestedHTTPSTracesKProbes(t *testing.T) {
 	assert.Empty(t, sd, sd.String())
 }
 
+func testHTTPTracesNestedCallsTooLong(t *testing.T) {
+	var parentID string
+
+	// Run a request, since we have a single app, we should see always all requests
+	doHTTPGet(t, "http://localhost:7773/slow", 200)
+
+	// Give time for all spans to make it to jeager, we want to ensure the children
```
mariomac:
```suggestion
	// Give time for all spans to make it to Jaeger, we want to ensure the children
```
(nit)

test/integration/traces_test.go
grcevski: Good idea!

PR #793, opened 2025-10-23
DNS tracing 

grcevski:
This PR adds tracking of DNS requests in OBI. It watches for traffic on the well known DNS ports and parses the request and response. From this data we extract information about which addresses were requested and what we received as response, as  well as how long did it take.

At the same time, OBI now can also report UDP requests that have been rejected, e.g. failed completely.

The PR is too large at this point, so I'm proposing the PR with the functionality disabled. 

At the moment the DNS data is shown only in traces, because it was little effort to do, but it's not on by default and needs to be explicitly enabled. I think we want to keep it like this, because the information can be very verbose, but I'm open for discussion.

I'm going to follow-up with a PR to implement the OpenTelemetry metrics for DNS (which will be on by default) and another PR to make a DNS trace example.


pkg/ebpf/common/dns_request_transform.go
```diff
@@ -0,0 +1,135 @@
...
+	return func(_ dnsparser.DNSId, span *request.Span) {
+		// final status is -1, which means we never received a response
+		if span.Status == -1 {
+			span.Status = int(dnsparser.RCodeRefused)
+			spansChan.Send(filter.Filter([]request.Span{*span}))
+		}
+	}
+}
+
+func ReadDNSEventIntoSpan(parseCtx *EBPFParseContext, record *ringbuf.Record) (request.Span, bool, error) {
```
mariomac: I think this method can be private, so we keep the public API as smallest as possible

bpf/common/common.h
```diff
@@ -239,3 +239,22 @@ typedef struct mongo_go_client_req {
     connection_info_t conn;
     tp_info_t tp;
 } mongo_go_client_req_t;
+
+#define DNS_MAX_LEN 516
```
mmat11: nit: use enum constant

bpf/common/common.h
```diff
@@ -239,3 +239,22 @@ typedef struct mongo_go_client_req {
...
+
+typedef struct dns_req {
+    u8 flags; // Must be fist we use it to tell what kind of packet we have on the ring buffer
```
mmat11:
```suggestion
    u8 flags; // Must be first we use it to tell what kind of packet we have on the ring buffer
```

bpf/common/common.h
```diff
@@ -239,3 +239,22 @@ typedef struct mongo_go_client_req {
...
+    u8 p_type;
```
mmat11, line 249: I think this is unused

bpf/common/dns.h
```diff
@@ -0,0 +1,167 @@
...
+
+#include <maps/sock_pids.h>
+
+#include <pid/types/pid_info.h>
+
+#define DNS_QR_QUERY 0
+#define DNS_QR_RESP 1
+
+// https://datatracker.ietf.org/doc/html/rfc1035#section-4.1.1
+union dnsflags {
```
mmat11, line 28: question: shouldn't this always be in network byte order?

bpf/common/dns.h
```diff
@@ -0,0 +1,167 @@
...
+    struct {
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+        __u8 rcode : 4;  // response code
```
mmat11: nit: use `u8` instead of `__u8` (applies everywhere), for consistency

bpf/common/dns.h
```diff
@@ -0,0 +1,167 @@
...
+#include <common/trace_util.h>
+
+#include <generictracer/k_tracer_defs.h>
+#include <generictracer/protocol_tcp.h>
+
+#include <maps/sock_pids.h>
+
+#include <pid/types/pid_info.h>
+
+#define DNS_QR_QUERY 0
```
mmat11: nit: use enums

bpf/common/connection_info.h
```diff
@@ -123,6 +131,13 @@ static __always_inline void swap_connection_info_order(connection_info_t *info)
     __builtin_memcpy(info->d_addr, tmp_addr, sizeof(info->d_addr));
 }
 
+static __always_inline void copy_connection_info(connection_info_t *dst, connection_info_t *src) {
```
mmat11: can't we use memcpy directly on `connection_info_t`?

bpf/common/common.h
```diff
@@ -239,3 +239,22 @@ typedef struct mongo_go_client_req {
...
+    u8 dns_q;
+    u8 _pad1[1];
+    u32 len;
+    pid_connection_info_t p_conn;
```
mmat11: perhaps we can use `connection_info_t` as pid information is redundant since we already send `pid_info`?

bpf/common/common.h
```diff
@@ -239,3 +239,22 @@ typedef struct mongo_go_client_req {
...
+    u16 id;
+    u8 _pad2[6];
+    tp_info_t tp;
+    u64 ts;
+    // we need this to filter traces from unsolicited processes that share the executable
+    // with other instrumented processes
+    pid_info pid;
+    unsigned char buf[DNS_MAX_LEN];
```
mmat11: how's the average packet payload size looking like? I wonder if it would be better for this kind of event to use dynamic-sized buffers?
maybe we can start with the hardcoded version and evaluate in the future?

bpf/common/dns.h
```diff
@@ -0,0 +1,167 @@
...
+            req->ts = bpf_ktime_get_ns();
+            req->tp.ts = bpf_ktime_get_ns();
+            __builtin_memcpy(&req->pid, &conn_pid->p_info, sizeof(pid_info));
+
+            trace_key_t t_key = {0};
+            trace_key_from_pid_tid_with_p_key(&t_key, &conn_pid->p_key, conn_pid->id);
+
+            u8 found = find_trace_for_client_request_with_t_key(
+                &p_conn, orig_dport, &t_key, conn_pid->id, &req->tp);
+            bpf_dbg_printk("Looking up client trace info, found %d", found);
```
mmat11:
```suggestion
            bpf_dbg_printk("handle_dns: looking up client trace info, found %d", found);
```

bpf/generictracer/k_tracer.c
```diff
@@ -191,6 +209,24 @@ int BPF_KPROBE(obi_kprobe_tcp_connect, struct sock *sk) {
     return 0;
 }
 
+// Used by connect so that we can grab the sock details
```
mmat11: Is this comment valid?

bpf/tpinjector/tpinjector.c
```diff
@@ -251,7 +251,7 @@ static __always_inline u8 protocol_detector(struct sk_msg_md *msg,
     }
 
     if (is_http_request_buf((const unsigned char *)msg_ptr)) {
-        bpf_dbg_printk("Setting up request to be extended");
+        //bpf_dbg_printk("Setting up request to be extended");
```
mmat11: leftover?

pkg/ebpf/common/dnsparser/dns.go
```diff
@@ -0,0 +1,90 @@
...
+import "fmt"
+
+type DNSId struct {
+	HostPID uint32
+	ID      uint16
+}
+
+type Type uint16
+
+const (
```
mmat11, line 15: these constants are already defined in `golang.org/x/net/dns/dnsmessage`

bpf/common/common.h
grcevski: It's the packet type, I'm not using it at the moment, but we may want to export it for metrics when we do that work. I'd prefer to keep it.

bpf/common/dns.h
grcevski: Good point, I think I can use ntohs to convert them!

bpf/common/common.h
grcevski: I just searches for various examples and 512 seemed like an overkill to be honest. I'd keep it for now constant, unless we encounter something very big in the future.

bpf/common/connection_info.h
grcevski: I think this is unused now, I wrote it when I struggled with some verifier issue and I forgot to remove it. Thanks for catching it.

bpf/common/common.h
grcevski: good idea.

bpf/common/common.h
```diff
@@ -47,6 +47,10 @@ enum large_buf_action : u8 {
     k_large_buf_action_append = 1,
 };
 
+enum {
+    k_dns_max_len = 516,
```
mmat11, line 51: I think this is `512`?

mmat11 (approved):
lgtm!

bpf/common/common.h
grcevski: I had extra 4 bytes for the padding, why not take them :)

grcevski:
I looked at the test failure, I don't get it, we see the correct output, all transactions are correlated correctly:
```
obi-1                   | 2025-10-29 15:32:17.102933217 (7.050664ms[7.050664ms]) HTTPClient(subType=0) 403 GET /users(/users) [172.18.0.4 as python-service.multi-k:56704]->[172.18.0.11 as utestserver:3040] contentLen:152B responseLen:0B svc=[multi-k/python-service python] traceparent=[00-f4ed642f5ea40d29c9ea682093508d57-9db418bb298ea95f[2e37b92858428e79]-01]
obi-1                   | 2025-10-29 15:32:17.102933217 (6.892726ms[6.540728ms]) HTTP(subType=0) 403 GET /users(/users) [172.18.0.4 as 172.18.0.4:56704]->[172.18.0.11 as rails-service.multi-k:3040] contentLen:152B responseLen:5459B svc=[multi-k/rails-service ruby] traceparent=[00-f4ed642f5ea40d29c9ea682093508d57-be54b48d99564eb9[9db418bb298ea95f]-01]
obi-1                   | 2025-10-29 15:32:17.102933217 (11.00823ms[10.85235ms]) HTTP(subType=0) 200 GET /tracemetoo(/tracemetoo) [172.18.0.18 as 172.18.0.18:60232]->[172.18.0.4 as python-service.multi-k:7773] contentLen:108B responseLen:157B svc=[multi-k/python-service python] traceparent=[00-f4ed642f5ea40d29c9ea682093508d57-2e37b92858428e79[169b7c1a8372b3be]-01]
obi-1                   | 2025-10-29 15:32:17.102933217 (13.728148ms[13.728148ms]) HTTPClient(subType=0) 200 GET /tracemetoo(/tracemetoo) [172.18.0.18 as testserver.initial-set:60232]->[172.18.0.4 as pytestserver:7773:7773] contentLen:0B responseLen:5B svc=[initial-set/testserver go] traceparent=[00-f4ed642f5ea40d29c9ea682093508d57-169b7c1a8372b3be[658d681a286ccb97]-01]
obi-1                   | 2025-10-29 15:32:17.102933217 (14.47405ms[14.454554ms]) HTTP(subType=0) 200 Arith.T /jsonrpc(/jsonrpc) [172.18.0.18 as 172.18.0.18:33990]->[172.18.0.18 as testserver.initial-set:8088] contentLen:74B responseLen:0B svc=[initial-set/testserver go] traceparent=[00-f4ed642f5ea40d29c9ea682093508d57-658d681a286ccb97[d807ddd730c54975]-01]
obi-1                   | 2025-10-29 15:32:17.102933217 (17.209135ms[17.209135ms]) HTTPClient(subType=0) 200 POST /jsonrpc(/jsonrpc) [172.18.0.18 as testserver.initial-set:33990]->[172.18.0.18 as testserver:8088:8088] contentLen:74B responseLen:33B svc=[initial-set/testserver go] traceparent=[00-f4ed642f5ea40d29c9ea682093508d57-d807ddd730c54975[d075d0a20188870d]-01]
obi-1                   | 2025-10-29 15:32:17.102933217 (17.5443ms[17.46911ms]) HTTP(subType=0) 200 GET /gotracemetoo(/gotracemetoo) [172.18.0.2 as 172.18.0.2:36520]->[172.18.0.18 as testserver.initial-set:8080] contentLen:0B responseLen:0B svc=[initial-set/testserver go] traceparent=[00-f4ed642f5ea40d29c9ea682093508d57-d075d0a20188870d[5b1fda9401e276ab]-01]
obi-1                   | 2025-10-29 15:32:17.102933217 (18.796274ms[18.796274ms]) HTTPClient(subType=0) 200 GET /gotracemetoo(/gotracemetoo) [172.18.0.2 as nodejs-service.multi-k:36520]->[172.18.0.18 as testserver:8080] contentLen:77B responseLen:0B svc=[multi-k/nodejs-service nodejs] traceparent=[00-f4ed642f5ea40d29c9ea682093508d57-5b1fda9401e276ab[4689bd94cd58d096]-01]
obi-1                   | 2025-10-29 15:32:17.102933217 (24.530902ms[24.530902ms]) HTTPClient(subType=0) 200 GET /traceme(/traceme) [172.18.0.10 as java-service.multi-k:36106]->[172.18.0.2 as ntestserver:3030] contentLen:236B responseLen:0B svc=[multi-k/java-service java] traceparent=[00-f4ed642f5ea40d29c9ea682093508d57-38d30e04afd529a8[5b15ad5a954b9209]-01]
obi-1                   | 2025-10-29 15:32:17.102933217 (233.198822ms[231.41876ms]) HTTP(subType=0) 200 GET /jtraceA(/jtraceA) [172.18.0.1 as 172.18.0.1:39274]->[172.18.0.10 as java-service.multi-k:8085] contentLen:192B responseLen:115B svc=[multi-k/java-service java] traceparent=[00-f4ed642f5ea40d29c9ea682093508d57-5b15ad5a954b9209[0000000000000000]-01]
obi-1                   | 2025-10-29 15:32:17.102933217 (23.135994ms[23.135994ms]) HTTP(subType=0) 200 GET /traceme(/traceme) [172.18.0.10 as 172.18.0.10:36106]->[172.18.0.2 as nodejs-service.multi-k:3030] contentLen:236B responseLen:229B svc=[multi-k/nodejs-service nodejs] traceparent=[00-f4ed642f5ea40d29c9ea682093508d57-4689bd94cd58d096[38d30e04afd529a8]-01]

```
It's like it's not in `jaeger`, maybe because of our trace batching etc.

grcevski:
Oh I know, I need to follow-up with a test fix. We need to reload the trace from jaeger every time in the eventually loop. When it fetches the first time, not all trace components would be there. This is the source of the flakiness.

bpf/common/dns.h
```diff
@@ -0,0 +1,153 @@
...
+
+#include <maps/sock_pids.h>
+
+#include <pid/types/pid_info.h>
+
+enum dns_qr_type : u8 { k_dns_qr_query = 0, k_dns_qr_resp = 1 };
+
+// https://datatracker.ietf.org/doc/html/rfc1035#section-4.1.1
+union dnsflags {
+    struct {
```
rafaelroquetto, line 29: bitfields are a dangerous idea - there are a few problems here

1. the C standard does not guarantee the ordering or packing of the bitfields:

> "The order of allocation of bit-fields within a unit (high-order to low-order or low-order to high-order) is implementation-defined."
(§6.7.2.1 paragraph 11)

i.e. it's undefined behaviour - note that this is orthogonal as to whether the target architecture is big-endian or little-endian - the compiler may choose to pack these fields in a LSB order regardless, and then it's broken.

There's more:

> "If enough space remains, a bit-field that immediately follows another bit-field in a structure shall be packed into adjacent bits of the same unit. If insufficient space remains, whether a bit-field that does not fit entirely within the unit is put into the next unit or overlaps adjacent units is implementation-defined."
(§6.7.2.1 paragraph 12)

So the packing and alignment is also implementation defined.

Instead, I'd like to suggest we simply implement the DNS header layout as per the spect and use bitwise operations - this is safest way (and the canonical way in these scenarios):

```
struct dnsheader {
    uint16_t id;
    uint16_t flags;  // in network byte order
    uint16_t qdcount;
    uint16_t ancount;
    uint16_t nscount;
    uint16_t arcount;
};

static __always_inline uint8_t dns_qr(uint16_t f)     { return (f >> 15) & 0x1; }
static __always_inline uint8_t dns_opcode(uint16_t f) { return (f >> 11) & 0xF; }
static __always_inline uint8_t dns_aa(uint16_t f)     { return (f >> 10) & 0x1; }
static __always_inline uint8_t dns_tc(uint16_t f)     { return (f >> 9)  & 0x1; }
static __always_inline uint8_t dns_rd(uint16_t f)     { return (f >> 8)  & 0x1; }
static __always_inline uint8_t dns_ra(uint16_t f)     { return (f >> 7)  & 0x1; }
static __always_inline uint8_t dns_z(uint16_t f)      { return (f >> 4)  & 0x7; }
static __always_inline uint8_t dns_rcode(uint16_t f)  { return  f        & 0xF; }
```

bpf/generictracer/k_tracer.c
```diff
@@ -926,7 +969,15 @@ int obi_socket__http_filter(struct __sk_buff *skb) {
     protocol_info_t tcp = {};
     connection_info_t conn = {};
 
-    if (!read_sk_buff(skb, &tcp, &conn)) {
+    u8 success = read_sk_buff(skb, &tcp, &conn);
+
+    if (is_dns(&conn)) {
```
rafaelroquetto, line 974: Perhaps we should update the comment at the top of this function to cite that it's not only a fallback path anymore, but that it's the entry point for tracking DNS requests.

bpf/common/dns.h
```diff
@@ -0,0 +1,153 @@
...
+        // DNS is after the TCP header and the 2 bytes of the length of the DNS packet
+        dns_off += 2;
+        break;
+    default:
+        return 0;
+    }
+
+    union dnsflags flags;
+    bpf_skb_load_bytes(skb, dns_off + offsetof(struct dnshdr, flags), &flags.flags, sizeof(u16));
+    flags.flags = bpf_ntohs(flags.flags); // Convert from network to host byte order
```
rafaelroquetto, line 99: This then becomes simply:
```suggestion
    struct dnshdr hdr = {};
    bpf_skb_load_bytes(skb, dns_off, &hdr, sizeof(hdr));
```

bpf/common/dns.h
```diff
@@ -0,0 +1,153 @@
...
+
+    u8 qr = flags.qr;
```
rafaelroquetto, line 101:
```suggestion
    const u8 qr = dns_qr(hdr.flags);
```

bpf/common/dns.h
```diff
@@ -0,0 +1,153 @@
...
+    if (qr == k_dns_qr_query || qr == k_dns_qr_resp) {
+        u16 id = 0;
+        bpf_skb_load_bytes(skb, dns_off + offsetof(struct dnshdr, id), &id, sizeof(u16));
```
rafaelroquetto, line 104:
```suggestion
```

bpf/common/dns.h
```diff
@@ -0,0 +1,153 @@
...
+        u16 orig_dport = conn->d_port;
```
rafaelroquetto, line 105:
```suggestion
        const u16 orig_dport = conn->d_port;
```

bpf/common/dns.h
```diff
@@ -0,0 +1,153 @@
...
+        dns_req_t *req = bpf_ringbuf_reserve(&events, sizeof(dns_req_t), 0);
+
+        if (req) {
+            __builtin_memcpy(&req->conn, conn, sizeof(connection_info_t));
+
+            req->flags = EVENT_DNS_REQUEST;
+            req->p_type = skb->pkt_type;
+            req->len = skb->len;
+            req->dns_q = qr;
+            req->id = id;
```
rafaelroquetto, line 128:
```suggestion
            req->id = hdr.id;
```

bpf/common/dns.h
```diff
@@ -0,0 +1,153 @@
...
+            req->ts = bpf_ktime_get_ns();
+            req->tp.ts = bpf_ktime_get_ns();
+            __builtin_memcpy(&req->pid, &conn_pid->p_info, sizeof(pid_info));
```
rafaelroquetto, line 131:
```suggestion
            req->pid = conn_pid->p_info;
            __builtin_memcpy(&req->pid, &conn_pid->p_info, sizeof(pid_info));
```

this is just a nit - the `=` reads better IMHO (and is less error prone as the size is implicitly derived)

bpf/common/dns.h
```diff
@@ -0,0 +1,153 @@
...
+
+            trace_key_t t_key = {0};
+            trace_key_from_pid_tid_with_p_key(&t_key, &conn_pid->p_key, conn_pid->id);
+
+            u8 found = find_trace_for_client_request_with_t_key(
```
rafaelroquetto, line 136:
```suggestion
            const u8 found = find_trace_for_client_request_with_t_key(
```

bpf/common/dns.h
```diff
@@ -0,0 +1,153 @@
...
+        if (req) {
+            __builtin_memcpy(&req->conn, conn, sizeof(connection_info_t));
+
+            req->flags = EVENT_DNS_REQUEST;
+            req->p_type = skb->pkt_type;
+            req->len = skb->len;
+            req->dns_q = qr;
+            req->id = id;
+            req->ts = bpf_ktime_get_ns();
+            req->tp.ts = bpf_ktime_get_ns();
```
rafaelroquetto, line 130: I don't believe this is intended, but if it is ignore this comment.

Calling bpf_ktime_get_ns() twice may cause `req->ts != req->tp.ts`.

```suggestion
            const u64 ts = bpf_ktime_get_ns();
            req->ts = ts;
            req->tp.ts = ts;
```

pkg/ebpf/common/dnsparser/dns.go
```diff
@@ -0,0 +1,90 @@
...
+	RCodeFormatError:    "FormatError",
+	RCodeServerFailure:  "ServerFailure",
+	RCodeNameError:      "NameError",
+	RCodeNotImplemented: "NotImplemented",
+	RCodeRefused:        "Refused",
+}
+
+// String implements fmt.Stringer.String.
+func (t Type) String() string {
+	if n, ok := typeNames[t]; ok {
```
rafaelroquetto, line 78: this is just a nit - feel free to ignore.

I've seen map lookups appear a lot when profiling the network code in these kind of events. I am not sure this is a problem here, but if this is a high throughput codepath, you may want to change this to a `swtich t { ... }` statement and return directly.

pkg/ebpf/common/dnsparser/dns.go
```diff
@@ -0,0 +1,90 @@
...
+		return n
+	}
+	return fmt.Sprintf("%d", t)
+}
+
+// String implements fmt.Stringer.String.
+func (r RCode) String() string {
+	if n, ok := rCodeNames[r]; ok {
```
rafaelroquetto, line 86: same here

rafaelroquetto (changes requested):
Mostly nits - the "request changes" refer exclusively to the bitfield usage and the obscure bugs it may harbour.

PR #798, opened 2025-10-24
ci: run main jobs from main repo only, check go mod tidy diffs

mmat11:
- Run push jobs from this repository only
- Check go mod tidy is updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -449,3 +449,8 @@ check-clean-work-tree:
 		echo 'Working tree is not clean, did you forget to run "make"?' \
 		exit 1; \
 	fi
+
+.PHONY: check-go-mod
+check-go-mod:
+	go mod tidy
+	git diff -s --exit-code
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -449,3 +449,8 @@ check-clean-work-tree:
 		echo 'Working tree is not clean, did you forget to run "make"?' \
 		exit 1; \
 	fi
+
+.PHONY: check-go-mod
+check-go-mod:
+	go mod tidy
```
MrAlias, line 455: There are many `go.mod` files in the repository. Something like the follow is needed to update all go mod files.

```suggestion
	for d in $( find -type f -name 'go.mod' -exec dirname {} \; ); do (cd $d && go mod tidy); done
```

In other repositories we have this:

https://github.com/open-telemetry/opentelemetry-go/blob/e71de160de98a3fc8d97f957b88963b30f811977/Makefile#L209-L215

And use a separate check to validate a clean repository:

https://github.com/open-telemetry/opentelemetry-go/blob/e71de160de98a3fc8d97f957b88963b30f811977/Makefile#L260-L268

I think the `git diff` approach here is better, but we likely want to split these two targets up so they can be used independently.

PR #801, opened 2025-10-24
Add support for Elasticsearch client span

pinoOgni:
This PR introduces support for Elasticsearch client spans as part of #141.

In this initial implementation, support has been added for the `_search` operations. Additional operations will be covered in future PRs.

### DIFF
--- a/pkg/ebpf/common/http/elasticsearch.go
+++ b/pkg/ebpf/common/http/elasticsearch.go
@@ -122,6 +122,9 @@ func extractOperationName(req *http.Request) string {
 		return ""
 	}
 	parts := strings.Split(path, "/")
+	if len(parts) == 0 {
+		return ""
+	}
 	name := parts[len(parts)-1]
 	return strings.TrimPrefix(name, "_")
 }
@@ -137,8 +140,12 @@ func extractDBCollectionName(req *http.Request) string {
 	}
 
 	parts := strings.Split(path, "/")
-	if len(parts) > 0 {
-		return parts[0]
+	if len(parts) == 0 {
+		return ""
 	}
-	return ""
+	first := parts[0]
+	if strings.HasPrefix(first, "_") {
+		return ""
+	}
+	return first
 }
--- a/pkg/ebpf/common/http/elasticsearch_test.go
+++ b/pkg/ebpf/common/http/elasticsearch_test.go
@@ -67,6 +67,26 @@ func TestParseElasticsearchRequest(t *testing.T) {
 			expected: elasticsearchOperation{},
 			wantErr:  true,
 		},
+		{
+			name:  "Valid Post request with wrong query JSON type",
+			input: newRequest(http.MethodPost, "/test_index/_search", `{"query": "not_object"}`),
+			expected: elasticsearchOperation{
+				DBQueryText:      "{\"query\":\"not_object\"}",
+				DBOperationName:  "search",
+				DBCollectionName: "test_index",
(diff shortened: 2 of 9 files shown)


pkg/ebpf/common/http/elasticsearch.go
```diff
@@ -0,0 +1,144 @@
...
+
+// extractOperationName is a generic function used to extract the operation name
+// that is the endpoint identifier provided in the request
+func extractOperationName(req *http.Request) string {
+	path := strings.Trim(req.URL.Path, "/")
+	if path == "" {
+		return ""
+	}
+	parts := strings.Split(path, "/")
+	name := parts[len(parts)-1]
```
grcevski, line 128: Let's be careful and check the len(parts) here.

pkg/ebpf/common/http/elasticsearch_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"strings"
+	"testing"
+)
+
+func TestParseElasticsearchRequest(t *testing.T) {
+	newRequest := func(method, target, body string) *http.Request {
+		return httptest.NewRequest(method, target, strings.NewReader(body))
+	}
+
+	tests := []struct {
```
grcevski, line 18: Can we add tests for various malformed bodies and incomplete queries? It can happen with partial buffers and customers sending wrong requests.

pkg/ebpf/common/http_transform.go
```diff
@@ -148,7 +154,7 @@ func HTTPInfoEventToSpan(parseCtx *EBPFParseContext, event *BPFHTTPInfo) (reques
 		requestBuffer = event.Buf[:]
 	}
 
-	if parseCtx != nil && !parseCtx.payloadExtraction.HTTP.GraphQL.Enabled {
+	if parseCtx != nil && !parseCtx.payloadExtraction.HTTP.GraphQL.Enabled && !parseCtx.payloadExtraction.HTTP.Elasticsearch.Enabled {
```
grcevski: Do you mind making a helper function like any payload extraction enabled etc.?

grcevski (approved):
Looks amazing!! I added some minor comments around testing certain functions and one bounds check. I'm approving and we are good to merge once these are addressed.

pkg/ebpf/common/http/elasticsearch_test.go
pinoOgni: Sure

pkg/ebpf/common/http_transform.go
pinoOgni: Done

pkg/ebpf/common/http/elasticsearch_test.go
pinoOgni: Hi @grcevski, I was implementing the required tests and it occurred to me: are we sure we want to consider a request with a malformed body/incomplete request as invalid? 

I'm currently checking if the JSON is good or not and for example this "{" is an invalid json, while this "{"query": "not_object"}" is a valid json but not to be a valid request it should be for example like this "`{"query":{"match_all":{}}}"`. 
From our point of view it's just a request to an elasticsearch server which will reply with `400 Bad request` and we can categorize it as elasticsearch request instead of just http.
What do you think?

grcevski: Oh that's a good point! let's keep it as is!

PR #812, opened 2025-10-27
http: add support for AWS S3 spans

mmat11:
This PR adds support for AWS s3 operations spans.

I noticed the integration test I added did flake a couple times but I can't reproduce locally -- not sure if it's related to the general flakes we've seen in the past few days
edit: looks like it's flaky, looking into it

### DIFF
--- a/bpf/generictracer/k_tracer.c
+++ b/bpf/generictracer/k_tracer.c
@@ -779,7 +779,7 @@ static __always_inline int return_recvmsg(void *ctx, struct sock *in_sock, u64 i
     pid_connection_info_t info = {};
 
     if (!args && !in_sock) {
-        goto done;
+        return 0;
     }
 
     void *sock_ptr = in_sock;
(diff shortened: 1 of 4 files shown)


bpf/generictracer/k_tracer.c
```diff
@@ -791,7 +791,9 @@ static __always_inline int return_recvmsg(void *ctx, struct sock *in_sock, u64 i
         }
     }
 
-    if (copied_len <= 0) {
+    // We want the full response (or most of it) to be able to parse HTTP headers/body.
+    // Drop 1 byte packets as they would end the request tracking prematurely.
+    if (copied_len <= 1) {
```
mmat11: Note for the reviewers: I had to do this because the AWS api would respond to a put object request with `100 continue`, sending ~22 packets containing 1 byte of data each :S

mariomac: I think it would be nice to get this mention to AWS API in the code comment so people know that 1-byte packets are actually happening! :)

PR #817, opened 2025-10-28
java: fix jvm attach cleanup function

mmat11:
- fix error message being printed regardless of errors
- fix cleanup function; swap setegid/seteuid

/cc @grcevski


grcevski (approved):
Thanks!

PR #822, opened 2025-10-28
Improve gibberish detection

grcevski:
Our heuristic detector for IDs within paths, which is used for converting paths to the low cardinality route has a blindspot when the path component is only partially gibberish. For example the following path doesn't properly identify the obvious IDs as IDs:

```
/bar/test/test/bar-attach-generic-product-apjkmyp/files/multi-test-version-jwbCm
```
We'd like to match `bar-attach-generic-product-apjkmyp` and `multi-test-version-jwbCm` as `*`, but because they are partially made of real words, they come out as OK.

This PR adds a bit more logic to iterate on the final word for gibberish detection by splitting by the typical path segments: -, _ and .

### DIFF
--- a/test/integration/red_test_python_aws_s3.go
+++ b/test/integration/red_test_python_aws_s3.go
@@ -16,6 +16,7 @@ import (
 	"github.com/mariomac/guara/pkg/test"
 	"github.com/stretchr/testify/require"
 
+	ti "go.opentelemetry.io/obi/pkg/test/integration"
 	"go.opentelemetry.io/obi/test/integration/components/jaeger"
 )
 
@@ -37,7 +38,7 @@ func testPythonAWSS3(t *testing.T) {
 
 	// Wait for /health to appear in jaeger
 	test.Eventually(t, testTimeout, func(t require.TestingT) {
-		doHTTPGet(t, "http://localhost:8381/health", 200)
+		ti.DoHTTPGet(t, "http://localhost:8381/health", 200)
 		resp, err := http.Get(jaegerQueryURL + "?service=python3.12&operation=GET%20%2Fhealth")
 		require.NoError(t, err)
 		if resp == nil {


test/integration/red_test_python_aws_s3.go
```diff
@@ -35,6 +35,21 @@ func testPythonAWSS3(t *testing.T) {
 	waitForTestComponentsNoMetrics(t, address+"/health")
 	waitForTestComponentsNoMetrics(t, localstackAddress)
 
+	// Wait for /health to appear in jaeger
```
grcevski, line 39: I made a small fix to the AWS S3 test since it failed on me with createBucket missing. I checked the logs and we caught the application too late, we saw the response of the `PUT  /obi-bucket` but not the actual request. I made us wait calling `/health` until we see at least one jaeger trace with `/health`. Then we go into verifying the buckets.

mmat11: Thanks! I thought the issue was on `PUT /bucket/item` 🙃 , let's hope this fixes it

pkg/internal/transform/route/clusterurl/cluster.go
```diff
@@ -52,6 +52,8 @@ func NewClusterURLClassifier(config *Config) (*ClusterURLClassifier, error) {
 		validCharTable[c] = true
 	}
 
+	validCharTable[' '] = true
```
skl: Not needed since `' '` is already in the char table above

skl (approved):
lgtm!

mariomac (approved):

pkg/internal/transform/route/clusterurl/cluster.go
grcevski: Good catch, I was trying to reuse the list of '_', '-', and '.' but I decided it was better to just do an || statement than another char table of 256. I forgot to revert the ' ' :).

PR #826, opened 2025-10-29
Remove bitfields from DNS code

rafaelroquetto:
Bitfields carry a lot of undefined or implementation defined behaviour. The original code assumed ordering and packing semantics that are not guaranteed.

Reference from the C17 standard:

> "The order of allocation of bit-fields within a unit (high-order to low-order or low-order to high-order) is implementation-defined."
(§6.7.2.1 paragraph 11)

and

> "If enough space remains, a bit-field that immediately follows another bit-field in a structure shall be packed into adjacent bits of the same unit. If insufficient space remains, whether a bit-field that does not fit entirely within the unit is put into the next unit or overlaps adjacent units is implementation-defined."
(§6.7.2.1 paragraph 12)

which may lead to obscure and hard to spot/debug bugs.

Follow up to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/793

### DIFF
--- a/bpf/common/dns.h
+++ b/bpf/common/dns.h
@@ -25,10 +25,31 @@
 enum dns_qr_type : u8 { k_dns_qr_query = 0, k_dns_qr_resp = 1 };
 
 // https://datatracker.ietf.org/doc/html/rfc1035#section-4.1.1
+//
+// 4.1.1. Header section format
+//
+// The header contains the following fields:
+//
+//                                     1  1  1  1  1  1
+//       0  1  2  3  4  5  6  7  8  9  0  1  2  3  4  5
+//     +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
+//     |                      ID                       |
+//     +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
+//     |QR|   Opcode  |AA|TC|RD|RA|   Z    |   RCODE   | <--- flags (1)
+//     +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
+//     |                    QDCOUNT                    |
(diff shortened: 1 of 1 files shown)


grcevski (commented):
The code is much harder to follow now, is this a real problem for us in this given scenario? If we look at the bpfcore imports, bitfields are quite common in there.

grcevski:
> @grcevski bitfield layout and ordering are implementation-defined in C. LLVM might emit different layouts or shifts across versions, architectures, or compiler flags, there's no ABI stability guarantee. Who knows?

I'm sorry @rafaelroquetto, this still doesn't answer my question. If we draw the line on unions as not being acceptable, we have a problem with our existing code base which we import from libbpf. 

I'm also not comfortable making any of the changes you have suggested since we have no tests at the moment and that will follow when we build metrics support. The PR returning green doesn't mean the changes work.

bpf/common/dns.h
```diff
@@ -25,31 +25,49 @@
...
-        u8 rcode : 4;  // response code
-    };
-    u16 flags;
-};
-
 struct dnshdr {
     u16 id;
 
-    union dnsflags flags;
+    u16 flags; // in network byte order
```
grcevski: Let's add the following docs on how the bits are laid out:

```
// DNS flags in network byte order
//  15  14 13 12 11  10  9   8   7   6   5   4   3   2   1   0
// +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
// |QR|   Opcode  |AA|TC|RD|RA|   Z    |   RCODE   |
// +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
```

grcevski (approved):
Should be rebased and merged only after tests are added. I'll post the PR with metrics and tests shortly.

PR #830, opened 2025-10-30
Enable trace export debug

grcevski:
After we started using the OTel collector SDK for traces export we ended breaking the support for enabling trace export logging. The SDKLogLevel option on the traces export was never used. 

This PR adds support for logging in the traces exporter and enables it for the AWS tests so we can see what's happening.

### DIFF
--- a/pkg/export/otel/traces.go
+++ b/pkg/export/otel/traces.go
@@ -295,6 +295,7 @@ func createZapLoggerDev(sdkLogLevel string) *zap.Logger {
 	var level zapcore.Level
 	if err := level.UnmarshalText([]byte(sdkLogLevel)); err != nil {
 		slog.Error("unsupported trace exporter logger level", "error", err, "level", sdkLogLevel)
+		return zap.NewNop()
 	}
 
 	config := zap.NewDevelopmentConfig()


mmat11 (approved):

pkg/export/otel/traces.go
```diff
@@ -286,11 +287,33 @@ func getTracesExporter(ctx context.Context, cfg otelcfg.TracesConfig, im imetric
...
 
-func getTraceSettings(dataTypeMetrics component.Type) exporter.Settings {
+func createZapLoggerDev(sdkLogLevel string) *zap.Logger {
+	if sdkLogLevel == "" {
+		return zap.NewNop()
+	}
+
+	var level zapcore.Level
+	if err := level.UnmarshalText([]byte(sdkLogLevel)); err != nil {
+		slog.Error("unsupported trace exporter logger level", "error", err, "level", sdkLogLevel)
```
rafaelroquetto, line 297: should we be returning `zap.NewNop()` here in case of error? Otherwise level gets used on 301 but I am assuming it's invalid then.

grcevski: Good point, let me add that

rafaelroquetto (approved):

PR #832, opened 2025-10-30
Verify already started protocol handlers

grcevski:
Right now our protocol control logic looks for pieces of the buffer to identify whether something is HTTP, HTTP2, MySQL or  Postgres. However, we can possibly have the situation where the first packet that came through identified a protocol, say HTTP, but then the new part of the packet somehow based on the contents looks like something else. 

This PR ensures we skip other protocol detectors once we've identified that something is HTTP.

### DIFF
--- a/bpf/common/connection_info.h
+++ b/bpf/common/connection_info.h
@@ -18,6 +18,7 @@ enum protocol_type : u8 {
     k_protocol_type_unknown = 0,
     k_protocol_type_mysql = 1,
     k_protocol_type_postgres = 2,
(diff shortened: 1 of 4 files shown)


mmat11:
there's already some infrastructure in place for storing protocol info, see `call_protocol_args_t->protocol_type` or `enum protocol_type`
if we extend that enum to http/http2, the checks for other tcp protocols are already in place, see: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/bpf/generictracer/protocol_mysql.h#L205-L208

Issue #753, opened 2025-10-09
DB Semantic convention attributes need to be updated


gouthamve-but-enterprise:
I am not seeing `server.address` in `db_client_operation_duration_seconds_count` metric. I think this is because we don't actually add the attribute: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/export/attributes/attr_defs.go#L277-L284

Can we review the semantic conventions and update the implementation to match?

PR #849, opened 2025-10-30
Update to match db semconv

grcevski:
This PR fixes the DB and messaging semantic conventions for metrics to include the `server.address` field. We are currently generating it as an IP address, so I used the new DNS code to implement an eBPF based RDNS for application metrics. I've only tested this in one scenario, the OATS SQL test, and we probably want to do more testing before we enable it by default.

I'm leaving couple of comments in the PR code to explain some of the changes.

Closes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/753

### DIFF
--- a/bpf/common/common.h
+++ b/bpf/common/common.h
@@ -253,7 +253,6 @@ typedef struct dns_req {
     u16 id;
     u8 _pad2[2];
     tp_info_t tp;
-    u64 ts;
     // we need this to filter traces from unsolicited processes that share the executable
     // with other instrumented processes
     pid_info pid;
--- a/bpf/common/dns.h
+++ b/bpf/common/dns.h
@@ -110,7 +110,6 @@ static __always_inline void populate_dns_record(dns_req_t *req,
     req->len = size;
     req->dns_q = qr;
     req->id = bpf_ntohs(id);
-    req->ts = bpf_ktime_get_ns();
     req->tp.ts = bpf_ktime_get_ns();
     req->pid = conn_pid->p_info;
 
--- a/pkg/ebpf/common/dns_request_transform.go
+++ b/pkg/ebpf/common/dns_request_transform.go
@@ -64,9 +64,9 @@ func readDNSEventIntoSpan(parseCtx *EBPFParseContext, record *ringbuf.Record) (r
 			Host:          hostname,
 			HostPort:      hostPort,
 			ContentLength: 0,
-			RequestStart:  int64(event.Ts),
-			Start:         int64(event.Ts),
-			End:           int64(event.Ts + 1),
+			RequestStart:  int64(event.Tp.Ts),
+			Start:         int64(event.Tp.Ts),
+			End:           int64(event.Tp.Ts + 1),
 			TraceID:       trace.TraceID(event.Tp.TraceId),
 			SpanID:        trace.SpanID(event.Tp.SpanId),
 			ParentSpanID:  trace.SpanID(event.Tp.ParentId),
@@ -122,7 +122,7 @@ func readDNSEventIntoSpan(parseCtx *EBPFParseContext, record *ringbuf.Record) (r
 	if msg.Response {
 		responseCode = uint16(msg.RCode)
 		span.Status = int(responseCode)
-		span.End = int64(event.Ts)
+		span.End = int64(event.Tp.Ts)
 	} else {
 		return *span, true, nil // ignore until we get a response or never hear back
 	}


bpf/common/common.h
```diff
@@ -246,9 +246,8 @@ typedef struct mongo_go_client_req {
 
 typedef struct dns_req {
     u8 flags; // Must be first we use it to tell what kind of packet we have on the ring buffer
-    u8 p_type;
     u8 dns_q;
-    u8 _pad1[1];
+    u8 _pad1[2];
```
grcevski, line 250: I removed an unused field.

bpf/common/common.h
```diff
@@ -259,4 +258,5 @@ typedef struct dns_req {
     // with other instrumented processes
     pid_info pid;
     unsigned char buf[k_dns_max_len];
+    u8 _pad3[4];
```
grcevski, line 260: I had to clamp_max a size for the BPF buffer, so I needed a power of 2. I've switched to 512 and I've added a padding.

bpf/common/dns.h
```diff
@@ -158,30 +189,51 @@ static __always_inline u8 handle_dns(struct __sk_buff *skb,
...
-                urand_bytes(req->tp.span_id, SPAN_ID_SIZE_BYTES);
-            } else {
-                init_new_trace(&req->tp);
-            }
-            read_skb_bytes(skb, dns_off, req->buf, sizeof(req->buf));
+            u32 len = skb->len - dns_off;
+            bpf_clamp_umax(len, 512);
+            populate_dns_record(req, &p_conn, orig_dport, len, qr, hdr.id, conn_pid);
+
+            read_skb_bytes(skb, dns_off, req->buf, len);
```
grcevski, line 195: This was a nasty bug. We were reading up to the max of req->buf, but read_skb_bytes failed to grab the last chunk of 16 bytes since we weren't specifically reading exactly as we should. This left us with garbage at the end.

bpf/common/dns.h
```diff
@@ -158,30 +189,51 @@ static __always_inline u8 handle_dns(struct __sk_buff *skb,
...
+
+    return 0;
+}
+
+static __always_inline u8 handle_dns_buf(const unsigned char *buf,
+                                         const int size,
+                                         pid_connection_info_t *p_conn,
+                                         u16 orig_dport) {
+
+    if (size < sizeof(struct dnshdr)) {
```
grcevski, line 211: I had to add an additional way to read the DNS traffic since if OBI is configured with host network, it may not be able to read some internal pod traffic, in this case it wasn't able to see the docker internal DNS traffic between the docker network and the service itself. I reproduced it in our test examples, with setting `network_mode: host` for OBI.

In this scenario, the socket_filter doesn't see the traffic, but the `udp_sendmsg` and `sock_recvmsg` kprobes fire. I added a userspace buffer helper here so we can capture those DNS requests too.

If OBI is running in the network of the target process, we'll see the event twice, once from the sock_filter another time by the kprobes. This isn't a problem because we deduplicate the DNS requests in user space and won't let them be emitted twice.

pkg/internal/rdns/store/memory.go
```diff
@@ -18,26 +20,37 @@ type InMemory struct {
...
 	// TODO: address scenarios where different hostnames point to a same IP
-	entries map[string][]string
+	entries *simplelru.LRU[string, []string]
 }
 
-func NewInMemory() *InMemory {
-	return &InMemory{
-		entries: map[string][]string{},
+func NewInMemory(cacheSize int) (*InMemory, error) {
+	cache, err := simplelru.NewLRU[string, []string](cacheSize, nil)
```
grcevski, line 27: I changed this implementation to ensure this cache is memory capped.

bpf/common/dns.h
```diff
@@ -97,6 +97,37 @@ static __always_inline u8 is_dns(connection_info_t *conn) {
...
+                                                const u16 id,
+                                                const conn_pid_t *conn_pid) {
+    __builtin_memcpy(&req->conn, &p_conn->conn, sizeof(connection_info_t));
+
+    req->flags = EVENT_DNS_REQUEST;
+    req->len = size;
+    req->dns_q = qr;
+    req->id = bpf_ntohs(id);
+    req->ts = bpf_ktime_get_ns();
+    req->tp.ts = bpf_ktime_get_ns();
```
mmat11, line 113: nit: use `req->ts` here?

pkg/internal/rdns/store/memory.go
```diff
@@ -18,26 +20,37 @@ type InMemory struct {
...
+	im.access.Lock()
+	defer im.access.Unlock()
+	im.entries.Add(ip, []string{name})
+}
+
 func (im *InMemory) GetHostnames(ip string) ([]string, error) {
 	im.access.RLock()
 	defer im.access.RUnlock()
-	return im.entries[ip], nil
+	r, _ := im.entries.Get(ip)
```
mmat11, line 54: does this return an error? should we directly `return im.entries.Get(ip)`?

mmat11 (approved):
lgtm!

bpf/common/dns.h
grcevski: Nice catch. I need the req->tp.ts so that we can perform the "is transaction too long" check for finding a parent request, but, since I have the timestamp there I don't need another. I'm removing the req->ts field.

pkg/internal/rdns/store/memory.go
grcevski: No unfortunately it returns a bool :(.

PR #853, opened 2025-10-31
Add support for _msearch, _bulk and _doc operations 

pinoOgni:
This PR continues the work done in #801 and it:
- adds support for `_msearch`, `_bulk` and `_doc` Elasticsearch operations.
- removes the `extractDBQueryText` function: we just return the body and we don't check if it's a valid JSON because from an observability point of view it is still an Elasticsearch request/response. This way, we gain performance and provide the user with the query of the request.
- updates tests
- updates operation name extraction


pinoOgni:
> The shard-4 failure looks related, it's failing in the ES test.

Let me investigate it!

NimrodAvni78 (approved):
nice!

pinoOgni:
> I can also take a look at the BPF logs to see if anything is weird, unix_stream_sendmsg should be triggering for unix socks, I wonder if another traffic somehow passes through there and it's impacting this code path.

Hi @grcevski, my comment on the `unix_stream_sendmsg` was wrong. I updated the comment today.

grcevski:
Increasing the timeout seems to have helped. The integration tests passed and the only failure seems a compile issue of unit tests. The k8s netolly test was a temporary issue and I've re-started it.

Issue #800, opened 2025-10-24
Client Spans are missing the now stable attibute peer.service


thoro:
I'm trying to ingest spans from beyla into instana, but it can't correlate the services correctly, because only server.address is sent in the span.

Since peer.service is stable, it should be set for all spans that are accessing a known k8s object (service, pod).

It probably should not be set for resolved ips.

PR #854, opened 2025-10-31
Add peer service to client spans

grcevski:
This PR adds the span attribute "peer.service" as per the spec for client calls https://opentelemetry.io/docs/specs/semconv/registry/attributes/peer/.

We only set the peer if there's a host name resolved, not using IP addresses.

Fixes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/800


mmat11 (approved):
lgtm!

PR #855, opened 2025-10-31
Delay route harvesting for Java apps

grcevski:
We currently harvest HTTP routes from Java apps by talking to them through the `jcmd` interface. The trouble is that Java apps start very slowly, there's the class loading, bytecode verification, JIT compilation, etc. In my experience I've seen enterprise Java apps that take 2 minutes to fully load.

OBI can be loaded as a daemonset quite some time before the Java app loads, so technically we'll see the Java application as soon as the process becomes live, so when we pull the symbols to be harvested from the Java application, they may not be the final symbols, since perhaps, number of HTTP handlers may not be loaded yet.

To mitigate this, I'm making the Java harvesting delayed by 60 seconds and it's configurable to a different value, using the advanced route harvesting options. I think 60 seconds will cover majority of Java applications and 100% of the micro-services, even if they are running on slow clusters.

### DIFF
--- a/pkg/appolly/discover/attacher.go
+++ b/pkg/appolly/discover/attacher.go
@@ -283,7 +283,10 @@ func (ta *traceAttacher) harvestRoutes(ie *ebpf.Instrumentable, reused bool) {
 		procAge := ta.processAgeFunc(ie.FileInfo.Pid)
 		if procAge < delayTime {
 			time.AfterFunc(delayTime-procAge, func() {
-				ta.harvestRoutesProcessor(ie, reused)
+				// sanity check that the program is still up and running and it's the same command
+				if exePath, ready := ExecutableReady(PID(ie.FileInfo.Pid)); ready && exePath == ie.FileInfo.CmdExePath {
+					ta.harvestRoutesProcessor(ie, reused)
+				}
 			})
 
 			return
--- a/pkg/appolly/discover/watcher_proc.go
+++ b/pkg/appolly/discover/watcher_proc.go
@@ -74,7 +74,7 @@ func ProcessWatcherFunc(cfg *obi.Config, ebpfContext *ebpfcommon.EBPFEventContex
 		pids:              map[PID]ProcessAttrs{},
 		pidPorts:          map[pidPort]ProcessAttrs{},
 		listProcesses:     fetchProcessPorts,
-		executableReady:   executableReady,
+		executableReady:   ExecutableReady,
 		loadBPFWatcher:    loadBPFWatcher,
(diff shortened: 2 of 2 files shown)


internal/test/oats/http/docker-compose-java-routes-custom.yml
```diff
@@ -28,6 +28,7 @@ services:
       OTEL_EBPF_OTLP_TRACES_BATCH_TIMEOUT: "1ms"
       OTEL_EBPF_LOG_LEVEL: "DEBUG"
       OTEL_EBPF_BPF_DEBUG: "true"
+      OTEL_EBPF_JAVA_ROUTE_HARVEST_DELAY: "10s"
```
grcevski, line 31: Setting to 10 seconds to still have the feature tested, but not cause our tests to slow too much.

pkg/appolly/discover/attacher.go
```diff
@@ -273,6 +278,21 @@ func (ta *traceAttacher) harvestRoutes(ie *ebpf.Instrumentable, reused bool) {
 	}
 }
 
+func (ta *traceAttacher) harvestRoutes(ie *ebpf.Instrumentable, reused bool) {
+	if delay, delayTime := ta.routeHarvester.HarvestRoutesDelay(ie.FileInfo); delay {
+		procAge := ta.processAgeFunc(ie.FileInfo.Pid)
+		if procAge < delayTime {
+			time.AfterFunc(delayTime-procAge, func() {
+				ta.harvestRoutesProcessor(ie, reused)
```
skl: What would happen if the target java process, during/after the timeout:

- the target process died
- the target pid was reused for a different process

Maybe a cancellable context could be used, or pid re-validation/age matching could happen - to ensure we're still acting on the same process?

skl (approved):
lgtm, one comment

pkg/appolly/discover/attacher.go
grcevski: I think that possibly was a problem even today, it's just less likely to happen since the time window of us detecting a process and it dying was very short. What would happen is that the call to `HarvestRoutes` would return with an error, it has cancellable and returns in case of a non-existing executable. I think it's sufficient, but if you think otherwise I can add a validation check that the process is still alive just before we call harvest.

skl: Ok, that sounds like it covers the case where the process died. I think it's important to check that proc age >= delay time after the timeout completes, to verify the pid wasn't reused (should be unlikely though).

grcevski: Done

Issue #251, opened 2025-07-03
CI: trigger all integration tests when anything ebpf related changes


mmat11:
Current behavior: full integration tests are triggered only on changes in the `bpf/` path

Desired behavior: full integration tests are triggered on changes of ebpf code, loader code, llvm toolchain

https://cloud-native.slack.com/archives/C08P9L4FPKJ/p1751552036683089

mmat11:
Closing as completed

PR #866, opened 2025-11-04
upgrade cilium/ebpf to v0.20.0

mmat11:
Changelog: https://github.com/cilium/ebpf/releases/tag/v0.20.0

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -27,7 +27,7 @@ jobs:
           go-version-file: "go.mod"
           cache: "false"
       - name: Run verification and unit tests
-        run: make docker-generate verify cov-exclude-generated check-go-mod
+        run: make docker-generate verify cov-exclude-generated check-go-mod notices-update check-ebpf-ver-synced check-clean-work-tree
       - name: Report coverage
         uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5.5.1
         env:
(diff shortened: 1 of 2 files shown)


bpf/bpfcore/placeholder.go
```diff
@@ -4,3 +4,6 @@
 //go:build obi_bpf
 
 package bpfcore
+
+// Store loader version here in order to run the full test suite on upgrades.
+// github.com/cilium/ebpf v0.20.0
```
MrAlias, line 9: This seems like it has a high likelihood to be missed during upgrades and quickly become out of sync. Is there are better place to track this?

mmat11: none that I know of in the `bpf` subdirectory, I am open to ideas

grcevski (approved):
LGTM!

bpf/bpfcore/placeholder.go
MrAlias: How is this used? Does the loader read this comment?

mmat11: This is just a trigger for bpf vm integration tests which run only if a file in the bpf subdirectory is changed (see https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/.github/workflows/workflow_integration_tests_vm.yml#L6)

For more context: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/251

I thought this was a simple way to do so; maybe I can think of some oneliner to check and make sure these 2 versions are in sync?

mmat11: I added a CI check

PR #869, opened 2025-11-06
ci: automatically create github release on tag

skl:
This PR creates a new GitHub actions workflow that triggers when a git tag following the pattern `v*.*.*` is pushed to this repo:

- Automatically publishes a GitHub release
- Uses GitHub's auto-generated release notes
- The release is named the same as the git tag (follows current convention in this repo)

NOTE: `publish_dockerhub_main.yml` and `publish_dockerhub_k8s_cache_main.yml` also currently trigger on this same tag pattern push, I thought best to solve the ordering/workflow-calling in a separate PR.

In future, it would be possible to modify this workflow to handle e.g. artifact signing, uploading, and attaching to releases.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,27 @@
+name: Release
+
+on:
+  push:
+    tags:
+      # N.B. These workflows also trigger on this tag pattern:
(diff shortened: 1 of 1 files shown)


grcevski (approved):
LGTM!

PR #870, opened 2025-11-06
ci: use alpine:3.20 to workaround insecure runc procfs

skl:
Checking VM workflows for runc with unsafe procfs.

Relates to:

- https://aws.amazon.com/security/security-bulletins/rss/aws-2025-024/
- https://advisories.gitlab.com/pkg/golang/github.com/opencontainers/runc/CVE-2025-52565/

Builds are failing with:

```
------
 > [testserver1 stage-1 2/7] RUN mkdir -p x509:
0.844 runc run failed: unable to start container process: error during container init: unable to setup user: unsafe procfs detected: openat2 fsmount:fscontext:proc/self/setgroups: no such file or directory
------
failed to solve: process "/bin/sh -c mkdir -p x509" did not complete successfully: exit code: 1
```

I tried the following workarounds, which were all unsuccessful:

- unconfined seccomp profile
- insecure buildx instance
- pinning an older runc version on `golang:1.25.1-alpine`

In the end, the only workaround I could get to run successfully was:

- Use an older alpine image `alpine:3.20`, and install go 1.25.1 there

### DIFF
--- a/internal/test/vm/Dockerfile
+++ b/internal/test/vm/Dockerfile
@@ -5,6 +5,7 @@ FROM alpine:3.20
 ARG target=run-integration-test-vm
 ARG test_pattern=TestMultiProcess
 ARG run_number=1
+ARG GO_VERSION=1.25.1
 
 # Pin Docker/runc to Alpine 3.20 versions (before November 2025 CVE patches)
 # The procfs security checks in newer runc (CVE-2025-52881, CVE-2025-52565,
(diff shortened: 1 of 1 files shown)


internal/test/vm/Dockerfile
```diff
@@ -1,21 +1,35 @@
...
     git \
     make \
     openrc \
     openssh \
-    shadow
+    shadow \
+    wget
+
+# Install Go 1.25.1 manually (not available in Alpine 3.20 repos)
+RUN wget -q https://go.dev/dl/go1.25.1.linux-amd64.tar.gz && \
```
mmat11: nit: can we put the go version in some env var so next time it will be upgraded it will be sufficient to change 1 line?

mmat11 (approved):
lgtm if it works

internal/test/vm/Dockerfile
skl: Done in c43b4086b

MrAlias:
@skl can you open an issue to track this? I'm okay with merging this, but don't want to get stuck on this version for ever. It would be good to have an issue to keep track work to find alternatives to this downgrade.

PR #886, opened 2025-11-12
testing http2go with http2.Protocols

sword-jin:
Since [go1.24] provided a simple way to configure HTTP/2, we can support it as well.
https://go.dev/doc/go1.24#nethttppkgnethttp

This PR only contains that testing http client with http2.Protocols

## Checklist


internal/tools/tools.go
```diff
@@ -11,8 +11,9 @@ import (
 	_ "github.com/google/go-licenses/v2"
 	_ "github.com/grafana/go-offsets-tracker/cmd/go-offsets-tracker"
 	_ "github.com/onsi/ginkgo/v2/ginkgo"
-	_ "go.opentelemetry.io/build-tools/multimod"
 	_ "gotest.tools/gotestsum"
 	_ "sigs.k8s.io/controller-runtime/tools/setup-envtest"
 	_ "sigs.k8s.io/kind"
+
+	_ "go.opentelemetry.io/build-tools/multimod"
```
sword-jin, line 18: why `fmt` format this file, I did't change it. 🤔

sword-jin:
> I'm wondering if this changed the package names which we rely on to instrument. 

Oh, I see `configs/offsets/http2/inspect.go` imports `"go.opentelemetry.io/obi/pkg/test/httplib"`, this anti-pattern which doesn't make sense. I checked the code base, not sure where is appropriate of putting `NewHTTP2Transport`, @grcevski could you please give some suggestions..

sword-jin:
I recovered [internal/test/integration/components/go_grpc_server_mux/main.go](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/886/commits/bf4c3be462f55604da2c474dfd0dc8d1665abf7c), because it makes `TestGRPCMux` very flaky and time-consuming, and I realize currently we only hook `x/net/http2` package for supporting http2 in [go_nethttp.c](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/bpf/gotracer/go_nethttp.c).

An interesting point is that nothing is broken in `TestHTTP2Go` which also tests http2 server and http2 client, I think this is worth to in-depth look

---
My project is only using `http.Protocols` to serve http2, and we are planing to migrate to [opentelemetry-ebpf-instrumentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation)

configs/offsets/http2/inspect.go
```diff
@@ -5,12 +5,11 @@ package main
 
 import (
 	"context"
-	"crypto/tls"
 	"fmt"
 	"net/http"
 	"os"
 
-	"golang.org/x/net/http2"
+	"go.opentelemetry.io/obi/pkg/test/httplib"
```
grcevski: I think we need to revert this file change. This project here is a dummy project we use to find the offsets of `x/net/http2`, which people may still use. The Go offsets should already be there because we have another offsets dummy project for the go standard library.

internal/test/integration/components/gohttp2/client/main.go
```diff
@@ -31,14 +29,20 @@ func main() {
 	}
 }
 
+func newHTTP2Transport() *http.Transport {
```
grcevski: I like these changes, but I was thinking that we can duplicate this test with another test directory instead of modifying the code. If we remove this test with `x/net/http2` we may break things in the future for people that still use the old approach. Would you please clone the test repo and keep the original one? Maybe call the new one `gohttp2std` or whatever else you like.

grcevski:
> I recovered [internal/test/integration/components/go_grpc_server_mux/main.go](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/886/commits/bf4c3be462f55604da2c474dfd0dc8d1665abf7c), because it makes `TestGRPCMux` very flaky and time-consuming, and I realize currently we only hook `x/net/http2` package for supporting http2 in [go_nethttp.c](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/bpf/gotracer/go_nethttp.c).
> 
> An interesting point is that nothing is broken in `TestHTTP2Go` which also tests http2 server and http2 client, I think this is worth to in-depth look
> 
> My project is only using `http.Protocols` to serve http2, and we are planing to migrate to [opentelemetry-ebpf-instrumentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation)

Good idea, I'm going to look into this today. Thanks for providing the testcase, I'm going to check if our http2 support works for your example and add what's missing if it doesn't.

internal/tools/tools.go
grcevski: that's odd, maybe something to do with your code editor auto-formatting?

grcevski:
> > I'm wondering if this changed the package names which we rely on to instrument.
> 
> Oh, I see `configs/offsets/http2/inspect.go` imports `"go.opentelemetry.io/obi/pkg/test/httplib"`, this anti-pattern which doesn't make sense. I checked the code base, not sure where is appropriate of putting `NewHTTP2Transport`, @grcevski could you please give some suggestions..

I think the symbols for the NewHTTP2Transport should already be there because of `config/offsets/std_inspect.go`, it contains a file with the go standard http library and this is now also supporting http2. I think we should be fine with your example. Like I mentioned, just copy the project into new files and it's sufficient for us to add support for the new http2 way.

internal/test/integration/components/gohttp2/client/main.go
sword-jin: @grcevski, and maybe we can just add a ENV to choose using `x/net/http2` or `http.Protocols`, then copy a new testcase to test `http.Protocols`, does that sound good?

internal/tools/tools.go
sword-jin: image

on macos.

configs/offsets/http2/inspect.go
sword-jin: reverted

internal/test/integration/components/gohttp2/client/main.go
grcevski: Sure that works as well if it's possible! Good idea.

grcevski:
looks like CI hit some intermittent issues, I'm working on fixing restrict_local_node in another PR, once it's merged please sync with main so we can get a clean CI run. Thanks!

PR #887, opened 2025-11-12
feat: add `network.type` as attribute for network flows

theSuess:
This adds support for the `network.type` attribute as specified in the OTEL [Semantic
Conventions](https://opentelemetry.io/docs/specs/semconv/registry/attributes/network/#network-type). It follows a similar approach to the `transport` attribute and can be extended to support more protocols easily

Adding this field allows for easier aggregation without resorting to checking for `:` in the address fields


pkg/export/attributes/attr_defs.go
```diff
@@ -95,6 +95,7 @@ func getDefinitions(
 			attr.ClientPort:     false,
 			attr.SrcZone:        false,
 			attr.DstZone:        false,
+			attr.NetworkType:    false,
```
grcevski, line 98: We have a few tests in the Kubernetes integration suite that verify the output, would it be possible for you to add this attribute and check if it's coming out as expected? If you need help with this let me know.

For example, modifying `testNetFlowBytesForExistingConnections` in `k8s_netolly_network_metrics.go` to test for this attribute too.

The `06-obi-netolly.yml` which drives that test will already include this attribute, since we add all network olly attributes in the YAML config:
```
      select:
        obi.network.flow.bytes:
          # assured cardinality explosion. Don't try in production!
          include: ["*"]
          exclude: ["src_port"]
```

So essentially, one liner in the test I mentioned above to check the value of the `network.type` attribute would work, e.g. adding:
```
		metric := results[0].Metric
		assertIsIP(t, metric["src_address"])
		assertIsIP(t, metric["dst_address"])
...
+		assert.Equal(t, "ipv4", metric["network_type"]) 
```

theSuess: I'll do that!

grcevski: Awesome!

grcevski (approved):

PR #889, opened 2025-11-12
Go route harvesting

grcevski:
This PR adds extraction of the HTTP routes from Go programs. I've added support for the built-in Go Mux that was introduced in 1.21, Gorilla Mux and Gin. This should cover a lot of the use-cases, the rest of the frameworks will use the regular heuristic matcher.

I also looked into reading the go binary symbol tables and extracting routes like we did for Java, but it proved to not be feasible. We essentially would need to disassemble all of the code and then look for instructions that read the symbol tables, then figure out if those are actually HTTP routes.

Instead my approach uses uprobes in the frameworks to read the route value when they match it internally and determine the handler.

### DIFF
--- a/configs/offsets/gin/inspect.go
+++ b/configs/offsets/gin/inspect.go
@@ -1,3 +1,6 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
 package main
 
 import (
--- a/configs/offsets/mux/inspect.go
+++ b/configs/offsets/mux/inspect.go
@@ -1,3 +1,6 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
 package main
 
 import (


bpf/gotracer/go_grpc.c
```diff
@@ -464,7 +464,7 @@ static __always_inline int grpc_connect_done(struct pt_regs *ctx, void *err) {
     bpf_dbg_printk("method ptr = %lx, method_len = %d", method_ptr, method_len);
 
     // Get method from the incoming call arguments
-    if (!read_go_str_n("method", method_ptr, (u64)method_len, &trace->path, sizeof(trace->path))) {
+    if (!read_go_str_n("method", method_ptr, (u64)method_len, trace->path, sizeof(trace->path))) {
```
rafaelroquetto, line 467: nice catch!

bpf/gotracer/go_nethttp.c
```diff
@@ -137,7 +139,7 @@ int obi_uprobe_ServeHTTP(struct pt_regs *ctx) {
         if (!read_go_str("method",
                          req,
                          go_offset_of(ot, (go_offset){.v = _method_ptr_pos}),
-                         &invocation.method,
+                         invocation.method,
```
rafaelroquetto, line 142: man, how come this was working? nice catch

rafaelroquetto (approved):

bpf/gotracer/go_nethttp.c
grcevski: madness :) I have no idea, address of an address.

PR #890, opened 2025-11-13
Standardise on C17 with GNU extensions

rafaelroquetto:
Our C code already relies on C17 with GNU extensions (aka GNU17) implicitly. The GNU extensions are widely used not only by our own codebase, but by the headers provided by libbpf and vmlinux.h itself.

This PR:

1. makes the C standard explicit
2. add configuration files to make the experience in VSCode smoother

The empty `compile_flags.txt` is required to trick `clangd` into resolving the current working directory to the correct value (i.e. `bpf/`).

### DIFF
--- a/devdocs/vscode.md
+++ b/devdocs/vscode.md
@@ -1,6 +1,7 @@
 # Working with C Code for eBPF in VS Code
 
 This repository already includes the configuration needed for working on the eBPF C sources:
+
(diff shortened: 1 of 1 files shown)


grcevski (approved):
LGTM! This is huge for me, I've had red squiggly lines since we started using enum : u8 {..

PR #891, opened 2025-11-14
Add Java TLS bpf support

grcevski:
This PR adds the eBPF hooks for allowing the future Java OBI agent to supply the missing TLS information for Java TLS applications. Essentially, this just provides the mechanism to receive the data and it uses the exiting eBPF infrastructure to produce the events.

There are no tests yet for this, I'll follow-up with further PRs to do the following:
1. Add the Java OBI agent source code to the repo.
2. Add ability for OBI to inject the agent and then I'll provide integration tests.

PR #896, opened 2025-11-17
OBI java agent

grcevski:
This PR adds the OBI Java agent for instrumenting TLS for Java applications. It's a follow-up to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/891, that is, the Step 1 of the follow-ups.

Follow-up work required:
- Build process, packaging and CI
- Enable the injection from OBI's code

The contributed README.md contains all of the details on how to get started and how this agent is structured and organised.

### DIFF
--- a/pkg/internal/java/README.md
+++ b/pkg/internal/java/README.md
@@ -101,6 +101,41 @@ A lightweight loader that:
    └──────────┘
 ```
 
+### 3. Communicating with the eBPF side
+
+The Java code prepares a packet which is supplied to the `ioctl` C library call. The
+packet will contain information about the connection as well as the decrypted payload.
+
+The `ioctl` call uses custom constant (`0x0b10b1`) for the `id` parameter. We look for
+this magic number on the eBPF side to ensure the `ioctl` call is for our purpose.
+
+The payload we send to the eBPF side has this format:
+
+```
+Memory Layout of Pointer p (after pos.write(new byte[] {42}))
+═══════════════════════════════════════════════════════════════
+
+Offset    Size    Value    Description
+───────────────────────────────────────────────────────────────
+  0        1B      0x01     OperationType.SEND.code
+                          ┌─────────────────────────────┐
+  1       36B      ...    |ConnectionInfo (36 bytes)    │ packetPrefixSize
+                          │ (socket connection data)    │ = 1 + 36 + 4
+                          └─────────────────────────────┘
+ 37        4B      0x01     Buffer length (int = 1)
+                          ┌─────────────────────────────┐
+ 41        1B      0x2A   |Data byte: 42                │ Actual payload
+                          └─────────────────────────────┘
+
+Total size: 1 + 36 + 4 + 1 = 42 bytes
+
+Test assertions:
+───────────────────────────────────────────────────────────────
+p.getByte(0)           → 1    (OperationType.SEND)
+p.getInt(1 + 36)       → 1    (Buffer length at offset 37)
+p.getByte(1 + 36 + 4)  → 42   (Data byte at offset 41)
+```
+
 ## 🔨 Building
 
 ### Prerequisites
--- a/pkg/internal/java/agent/src/test/java/io/opentelemetry/obi/java/ebpf/IOCTLPacketTest.java
+++ b/pkg/internal/java/agent/src/test/java/io/opentelemetry/obi/java/ebpf/IOCTLPacketTest.java
@@ -12,6 +12,29 @@ import com.sun.jna.Pointer;
 import io.opentelemetry.obi.java.instrumentations.data.Connection;
 import org.junit.jupiter.api.Test;
 
+// Memory Layout of Pointer p (after pos.write(new byte[] {42}))
+// ═══════════════════════════════════════════════════════════════
+
+// Offset    Size    Value    Description
+// ───────────────────────────────────────────────────────────────
+//   0        1B      0x01     OperationType.SEND.code
+//                           ┌─────────────────────────────┐
+//   1       36B      ...    |ConnectionInfo (36 bytes)    │ packetPrefixSize
+//                           │ (socket connection data)    │ = 1 + 36 + 4
+//                           └─────────────────────────────┘
+//  37        4B      0x01     Buffer length (int = 1)
+//                           ┌─────────────────────────────┐
+//  41        1B      0x2A   |Data byte: 42                │ Actual payload
+//                           └─────────────────────────────┘
+
+// Total size: 1 + 36 + 4 + 1 = 42 bytes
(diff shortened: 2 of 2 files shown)


pkg/internal/java/build.gradle
```diff
@@ -0,0 +1,26 @@
+plugins {
+    id 'java'
+}
+
+group = 'io.opentelemetry.obi'
+version = '0.1.0'
```
skl, line 6: What's the process to bump the version of the agent? I see `0.1.0` in a few places.

Follow-up question: can the user tell which version of the agent they're running for debugging purposes?

pkg/internal/java/loader/src/main/java/io/opentelemetry/obi/java/Loader.java
```diff
@@ -0,0 +1,72 @@
...
+import java.net.URLClassLoader;
+import java.nio.file.Files;
+
+public class Loader {
+  public static void agentCaller(String function, String agentArgs, Instrumentation inst) {
+    String agentResourcePath = "agent/agent.zip";
+
+    try {
+      Class.forName("io.opentelemetry.obi.java.ebpf.ProxyOutputStream");
+      System.err.println("agent already loaded, ignoring load request.");
```
skl, line 20: When a newer version of the java agent becomes available, how would that be loaded if the previous version is already attached?

pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/Agent.java
```diff
@@ -0,0 +1,239 @@
...
+import java.util.logging.Logger;
+import net.bytebuddy.agent.ByteBuddyAgent;
+import net.bytebuddy.agent.builder.AgentBuilder;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.dynamic.ClassFileLocator;
+import net.bytebuddy.dynamic.loading.ClassInjector;
+import net.bytebuddy.utility.JavaModule;
+
+public class Agent {
+  public static int IOCTL_CMD = 0x0b10b1;
```
skl, line 41: What is this magic number? 😄

pkg/internal/java/agent/src/test/java/io/opentelemetry/obi/java/ebpf/ProxyOutputStreamTest.java
```diff
@@ -0,0 +1,72 @@
...
+  void testWriteSingleByte() throws IOException {
+    ByteArrayOutputStream baos = new ByteArrayOutputStream();
+    ProxyOutputStream pos = new ProxyOutputStream(baos, new Socket());
+    pos.write(new byte[] {42});
+    pos.close();
+    assertArrayEquals(new byte[] {42}, baos.toByteArray());
+
+    assertEquals(1, p.getByte(0));
+    assertEquals(1, p.getInt(1 + 36));
+    assertEquals(42, p.getByte(1 + 36 + 4));
```
skl, line 44: I had to ask Cursor about these numbers, maybe worth a comment/adding to README?

```
Memory Layout of Pointer p (after pos.write(new byte[] {42}))
═══════════════════════════════════════════════════════════════

Offset    Size    Value    Description
───────────────────────────────────────────────────────────────
  0        1B      0x01     OperationType.SEND.code
                          ┌─────────────────────────────┐
  1       36B      ...    |ConnectionInfo (36 bytes)    │ packetPrefixSize
                          │ (socket connection data)    │ = 1 + 36 + 4
                          └─────────────────────────────┘
 37        4B      0x01     Buffer length (int = 1)
                          ┌─────────────────────────────┐
 41        1B      0x2A   |Data byte: 42                │ Actual payload
                          └─────────────────────────────┘

Total size: 1 + 36 + 4 + 1 = 42 bytes

Test assertions:
───────────────────────────────────────────────────────────────
p.getByte(0)           → 1    (OperationType.SEND)
p.getInt(1 + 36)       → 1    (Buffer length at offset 37)
p.getByte(1 + 36 + 4)  → 42   (Data byte at offset 41)
```

EDIT - I see there's a comment in pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/ebpf/IOCTLPacket.java

Maybe `36` could be some kind of const?

skl (approved):
Nice tests!

pkg/internal/java/build.gradle
jaydeluca: > What's the process to bump the version of the agent? I see 0.1.0 in a few places.

in the java repos, we have a [script](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/.github/scripts/update-version.sh) that is used on each release via a [github action](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/.github/workflows/prepare-release-branch.yml#L55-L56) to update a [version.gradle.kts](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/version.gradle.kts) file, so we could implement something similar. I assume that the agent version would follow/use the same version number os OBI itself? or would that be separate?

grcevski: My thoughts are that perhaps we shouldn't stress about the version of this code much, not until perhaps this is removed from OBI and becomes part of the official Java agent, if we ever go that direction. If we build this artefact as part of the OBI binaries, it should be the OBI version always. We need "some" version to make the gradle build work, but I guess it would be nice if we can automate this to match the OBI version. Since the Go standard for the main branch would be 'main' for version, I don't know if Gradle would let us do that?

pkg/internal/java/loader/src/main/java/io/opentelemetry/obi/java/Loader.java
grcevski: Very good question :)! I don't have support for that yet. I think it's possible but also not easy. We'd need to somehow unload the previous version and load the new one. I think we can add this later.

pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/Agent.java
grcevski: It matches the magic number in java_tls.c. We need to pick a constant for the `ioctl` call, so I just picked one saying `obiobi` in binary :).

pkg/internal/java/agent/src/test/java/io/opentelemetry/obi/java/ebpf/ProxyOutputStreamTest.java
grcevski: Good idea, I like your schematic, I'll add this to the README and the code.

PR #900, opened 2025-11-19
Update opentelemetrio.io dependencies to v0.140.0/v1.46.0

mariomac:
## Checklist


grcevski:
the kafka test needs to be fixed, they keep updating the package and the test doesn't pin it. I'd remove this line in red_test_kafka.go:

```
						attribute.String("messaging.client_id", "kafka-python-2.2.16"),

```
Now it's 2.2.16, but our test says 2.2.15. If we upgrade it, it will fail on the next package upgrade.

PR #901, opened 2025-11-19
Deprecate queue.Send in favor of queue.SendCtx

mariomac:
During OBI shutdown, if a node is trying to invoke `queue.Send` when the receiving node has already shut down, that goroutine will stay blocked until the whole OBI process is killed (which might cause that some resources are not properly freed).

This PR is the starting point of the progressive replacement by `queue.SendCtx`, which would abort the send operation as soon as the passed context is done.

This PR also removes the `forEach` syntactic sugar construct from the test-fake Collector, as we have now a confortable way to iterate the OTEL sequences via `iter.Seq2`.


grcevski (approved):
LGTM! Nice!

PR #904, opened 2025-11-20
Allow overriding SDK name

grcevski:
This PR adds another attribute that will allow vendors to override the SDK name of OBI, just like the prefix. This helps the Beyla team continue with the support of the Beyla product with vendored OBI code, without hacks, but it would also allow for others to put the name of their product if they fork OBI. This is similar to the existing attribute of VendorPrefix.

I also noticed that the Prometheus metrics were not using `opentelemetry-ebpf-instrumentation` for SDK name, so I made that consistent with the OTel metrics exporters.

### DIFF
--- a/pkg/export/attributes/names/attrs.go
+++ b/pkg/export/attributes/names/attrs.go
@@ -87,8 +87,10 @@ const (
 // VendorPrefix allows identifying some metrics (network, internal counters...)
 // as custom metrics, since at the moment they don't follow any semantic convention for them.
 // This value can be overridden when OBI is vendored as a library (e.g. from the OTEL collector)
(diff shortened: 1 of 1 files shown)


skl (approved):

pkg/export/attributes/names/attrs.go
```diff
@@ -88,6 +88,7 @@ const (
 // as custom metrics, since at the moment they don't follow any semantic convention for them.
```
MrAlias, line 88: Looks like the line above this needs to have a tab replace the comment prefix.

MrAlias (approved):
Looks good to me outside of fixing the lint failure.

PR #908, opened 2025-11-20
Add sdk language to span metrics

grcevski:
The program language is low cardinality and it's quite useful to be part of span metrics, avoiding expensive joins with target_info.


marctc (approved):
LGTM

PR #910, opened 2025-11-21
Update OTEL_EBPF_PROMETHEUS_FEATURES comment

pinoOgni:
This PR updates the `OTEL_EBPF_PROMETHEUS_FEATURES` comment because we can add other values ​​besides "application" and "network" (same as `OTEL_EBPF_METRICS_FEATURES`).

Also, I noticed how the `export/prom` package uses variables like `Feature*` and `Buckets` declared in `export/otel/otelfcfg`. Technically, it would be more correct to create a `common.go `file at the `export` pkg level and include all the common elements between `prom` and `otel`, or something similar. What do you think?

### DIFF
--- a/pkg/export/prom/prom.go
+++ b/pkg/export/prom/prom.go
@@ -114,7 +114,8 @@ type PrometheusConfig struct {
 
 	DisableBuildInfo bool `yaml:"disable_build_info" env:"OTEL_EBPF_PROMETHEUS_DISABLE_BUILD_INFO"`
 
-	// Features of metrics that can be exported. Accepted values are "application" and "network".
+	// Features of metrics that can be exported. Accepted values: application, network, application_process,
+	// application_span, application_service_graph, ...
 	Features []string `yaml:"features" env:"OTEL_EBPF_PROMETHEUS_FEATURES" envSeparator:","`
 	// Allows configuration of which instrumentations should be enabled, e.g. http, grpc, sql...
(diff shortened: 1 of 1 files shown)


grcevski:
> Also, I noticed how the export/prom package uses variables like Feature* and Buckets declared in export/otel/otelfcfg. Technically, it would be more correct to create a common.go file at the export pkg level and include all the common elements between prom and otel, or something similar. What do you think?

Yes sure, that sounds like a good refactor. We can keep this PR as-is and follow-up with a refactor, or if you wish, you can extend this PR with that change.

pinoOgni:
> > Also, I noticed how the export/prom package uses variables like Feature* and Buckets declared in export/otel/otelfcfg. Technically, it would be more correct to create a common.go file at the export pkg level and include all the common elements between prom and otel, or something similar. What do you think?
> 
> Yes sure, that sounds like a good refactor. We can keep this PR as-is and follow-up with a refactor, or if you wish, you can extend this PR with that change.

Thanks @grcevski! That's fine by me, you can go ahead and merge this one. I'll start the refactor in a new branch to keep things clear.

grcevski (approved):
LGTM!

grcevski:
I expect that we will always want to keep both OTel and Prometheus capabilities in sync. It shouldn't matter if people prefer pull or push, it should work the same way. Having said that, you never know, but I guess we'll cross that bridge when we get there, if we wanted to diverge at some point :).

PR #914, opened 2025-11-21
Add TCP Options Context Propagation Support

rafaelroquetto:
### Add TCP Options Context Propagation Support

This PR adds support for injecting distributed tracing context via TCP options (kind 25), providing a transport-layer fallback for context propagation when HTTP headers are unavailable (e.g., SSL/TLS traffic or non-HTTP protocols):

- New `tpinjector` now injects TCP options in addition to HTTP
- Flexible configuration with a bitmask-based `ContextPropagation` field supporting combinations like `headers`, `tcp`, `ip`, or `all` via `OTEL_EBPF_BPF_CONTEXT_PROPAGATION` (backwards compatible)
- Mutual exclusion on egress ensures only one injection method per connection, TCP or IP, in order of precedence
- "Last one wins" on ingress allows all enabled methods to parse independently, with later methods overwriting earlier ones to establish a natural priority hierarchy (IP < TCP < HTTP)
- Documentation added to `devdocs/context-propagation.md` explaining architecture, execution order, and coordination mechanisms

### Breaking Changes

Removed deprecated `ContextPropagationEnabled` boolean field - users must migrate to `context_propagation: "all"` or specific combinations like `"headers,tcp"`

### DIFF
--- a/pkg/config/ebpf_tracer_test.go
+++ b/pkg/config/ebpf_tracer_test.go
@@ -240,7 +240,6 @@ func TestContextPropagationMode_HasMethods(t *testing.T) {
 	}
 }
 
-
 func TestContextPropagationMode_TracerLoading(t *testing.T) {
 	// Test which tracers should be loaded for each configuration
 	// tpinjector handles: HTTP headers (sk_msg) and TCP options (BPF_SOCK_OPS)


bpf/common/connection_info.h
```diff
@@ -79,18 +82,30 @@ static __always_inline void dbg_print_http_connection_info(connection_info_t *in
                    info->d_port);
 }
 static __always_inline void dbg_print_http_connection_info_part(connection_info_part_t *info) {
+    if (!k_bpf_debug) {
```
rafaelroquetto, line 85: this is better than the macro - when it's defined to `0`, the compiler will trip out all the unreachable code anyway (so equivalent to the macro in that sense) from the final binary - but this ensures that the code is always sieved through the compiler for syntax and sanity checks.

bpf/tpinjector/tpinjector.c
```diff
@@ -22,59 +26,180 @@
...
+
+struct tp_option {
+    u8 kind;
+    u8 len;
+    unsigned char trace_id[TRACE_ID_SIZE_BYTES];
+    unsigned char span_id[SPAN_ID_SIZE_BYTES];
+};
+
+static __always_inline void
+encode_hex_skb(unsigned char *dst, const unsigned char *src, u32 src_len) {
```
mmat11: we already have this function in `common/trace_util.h`

bpf/tpinjector/tpinjector.c
```diff
@@ -22,59 +26,180 @@
...
+#pragma clang loop unroll(full)
+    for (u32 i = 0, j = 0; i < src_len; i++) {
+        unsigned char p = src[i];
+
+        dst[j++] = hex[(p >> 4) & 0xff];
+        dst[j++] = hex[p & 0x0f];
+    }
+}
+
+static __always_inline const char *tp_string_from_opt(const struct tp_option *opt) {
```
mmat11, line 70: we also have something similar in `common/tracing.h` (see: `make_tp_string()`), maybe we can use the tp_info instead of tp_option

bpf/tpinjector/tpinjector.c
```diff
@@ -110,69 +235,181 @@ static __always_inline connection_info_t sk_msg_extract_key_ip6(struct sk_msg_md
...
+    struct bpf_sock *sk = skops->sk;
+
+    if (!sk) {
+        return;
+    }
+
+    const tp_info_pid_t *tp_pid = bpf_sk_storage_get(&sk_tp_info_pid_map, sk, NULL, 0);
+
+    if (!tp_pid) {
+        bpf_dbg_printk("tp info not found");
```
mmat11: nit: can we prepend the function name before the log message?

pkg/config/ebpf_tracer.go
```diff
@@ -19,10 +17,19 @@ type RedisDBCacheConfig struct {
...
+	ContextPropagationDisabled  ContextPropagationMode = 0
+	ContextPropagationHeaders   ContextPropagationMode = 1 << 0 // HTTP headers
+	ContextPropagationTCP       ContextPropagationMode = 1 << 1 // TCP options
+	ContextPropagationIPOptions ContextPropagationMode = 1 << 2 // IP options
+
+	// Convenience aliases
+	ContextPropagationAll = ContextPropagationHeaders | ContextPropagationTCP | ContextPropagationIPOptions
+)
+
+// Deprecated aliases for backwards compatibility
```
mmat11: aren't we the only "users" of these aliases? can we also remove these since we deprecated the `ContextPropagationEnabled` opt?

pkg/config/ebpf_tracer.go
```diff
@@ -54,11 +61,9 @@ type EBPFTracer struct {
 	// Must be at least 0
 	HTTPRequestTimeout time.Duration `yaml:"http_request_timeout" env:"OTEL_EBPF_BPF_HTTP_REQUEST_TIMEOUT" validate:"gte=0"`
 
-	// Deprecated: equivalent to ContextPropagationAll
-	ContextPropagationEnabled bool `yaml:"enable_context_propagation" env:"OTEL_EBPF_BPF_ENABLE_CONTEXT_PROPAGATION" validate:"boolean"`
-
 	// Enables distributed context propagation.
-	ContextPropagation ContextPropagationMode `yaml:"context_propagation" env:"OTEL_EBPF_BPF_CONTEXT_PROPAGATION" validate:"oneof=0 1 2 3"`
+	// Can be a combination of: headers, tcp, ip (e.g., "headers,tcp" or "all")
+	ContextPropagation ContextPropagationMode `yaml:"context_propagation" env:"OTEL_EBPF_BPF_CONTEXT_PROPAGATION"`
```
mmat11, line 60: should we validate?

edit: I saw we probably do it in `UnmarshalText()`? in that case disregard this

devdocs/context-propagation.md
```diff
@@ -0,0 +1,304 @@
...
+
+#### Scenario A: Go HTTP or SSL/TLS (uprobes involved)
+
+1. **uprobes** (Go HTTP client or SSL detection)
+   - Populate `outgoing_trace_map` with initial trace context
+   - Set `valid=1` for non-SSL, `valid=0` for SSL
+
+2. **sk_msg (tpinjector)**
+   - Runs for packets in sockmap
+   - Can inject HTTP headers and/or schedule TCP options
```
mmat11, line 43: question: there is some Go code that uses `bpf_probe_write_user` and some part of tpinjector (`write_go_traceparent`) which seems to do the same. These two should be interchangeable as written in the comment for the Go handling:
```
        // For Go we support two types of HTTP context propagation for now.
        //   1. The one that this code does, which uses the locked down bpf_probe_write_user.
        //   2. By using a sock_msg program that will extend the packet.
        // If this code ran, we should ensure that the second part doesn't run, therefore
        // we remove the metadata setup in uprobe_persistConnRoundTrip(struct pt_regs *ctx), so
        // that approach 2. skips this packet.
```
Maybe it would make sense to add this also here?

bpf/tpinjector/tpinjector.c
rafaelroquetto: Nice catch, will reuse it.

bpf/tpinjector/tpinjector.c
rafaelroquetto: The idea of using tp_option is to ensure that whatever/however it was written it is correct - it's good for debugging and also to inspect it as a "source of truth" right before the options are injected or parsed. If we used tp_info instead, then the question of "does tp_info match this corresponding tp_option" would always remain.
I toyed with the idea of creating a lower-level helper that is used by both, but it looks less straightforward than the current minimal duplication, so I'd probably just leave this one as is.

bpf/tpinjector/tpinjector.c
rafaelroquetto: yes, we can!

pkg/config/ebpf_tracer.go
rafaelroquetto: Ah indeed, that's a left-over. I'll remove then. Thanks for spotting this.

pkg/config/ebpf_tracer.go
rafaelroquetto: yes, it will fail early on UnmarshalText

devdocs/context-propagation.md
rafaelroquetto: Good idea! Consider it done!

PR #924, opened 2025-11-25
Node route harvesting

grcevski:
This PR extends our route harvesting capabilities to Node.js as well as the existing ones for Java and Go. Since Node.js applications are deployed as scripts, the code does a directory scan for a few of the most common Node.js web framewrorks and discovers routes used.

Supported frameworks:
- Express (with route syntaxt too)
- Fastify (short and long syntax)
- Koa Router
- Hapi
- Restify
- NestJS
- HTTP Dispatcher
- NextJS

The majority of the code are tests. Main logic of the harvesting is in `js.go`, with some extra helper libraries that help us determine what's the path for the main node.js script.

### DIFF
--- a/pkg/internal/transform/route/harvest/js_test.go
+++ b/pkg/internal/transform/route/harvest/js_test.go
@@ -1034,7 +1034,7 @@ func TestExtractNodejsRoutes(t *testing.T) {
 	// Create test directory structure
 	tempDir := t.TempDir()
 	testAppDir := filepath.Join(tempDir, "app")
-	require.NoError(t, os.MkdirAll(testAppDir, 0755))
+	require.NoError(t, os.MkdirAll(testAppDir, 0o755))
 
 	// Create a simple test JavaScript file with routes
 	testFile := filepath.Join(testAppDir, "server.js")
@@ -1056,7 +1056,7 @@ app.get('/api/users/:id', (req, res) => {
 
 app.listen(3000);
 `
-	require.NoError(t, os.WriteFile(testFile, []byte(testContent), 0644))
+	require.NoError(t, os.WriteFile(testFile, []byte(testContent), 0o644))
 
 	tests := []struct {
 		name           string
@@ -1174,7 +1174,7 @@ app.listen(3000);
 				require.NoError(t, err)
 				require.NotNil(t, result)
 				assert.Equal(t, CompleteRoutes, result.Kind)
-				assert.Equal(t, tt.expectedCount, len(result.Routes))
+				assert.Len(t, result.Routes, tt.expectedCount)
 
 				// Check that expected routes are present
 				for _, expectedRoute := range tt.expectedRoutes {
@@ -1200,17 +1200,17 @@ func TestExtractNodejsRoutes_EmptyDirectory(t *testing.T) {
 	// Create empty directory
 	tempDir := t.TempDir()
 	emptyDir := filepath.Join(tempDir, "empty")
-	require.NoError(t, os.MkdirAll(emptyDir, 0755))
+	require.NoError(t, os.MkdirAll(emptyDir, 0o755))
 
-	rootDirForPID = func(pid int32) string {
+	rootDirForPID = func(_ int32) string {
 		return tempDir
 	}
 
-	cmdlineForPID = func(pid int32) (string, []string, error) {
+	cmdlineForPID = func(_ int32) (string, []string, error) {
 		return "node", []string{"node", "server.js"}, nil
 	}
 
-	cwdForPID = func(pid int32) (string, error) {
+	cwdForPID = func(_ int32) (string, error) {
 		return "/empty", nil
 	}
 
--- a/pkg/internal/transform/route/harvest/js.go
+++ b/pkg/internal/transform/route/harvest/js.go
@@ -495,19 +495,21 @@ func (e *RouteExtractor) FirstArg(args []string) string {
 }
 
 // testing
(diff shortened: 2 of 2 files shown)


pkg/internal/transform/route/harvest/harvester.go
```diff
@@ -121,7 +130,52 @@ func (h *RouteHarvester) HarvestRoutes(fileInfo *exec.FileInfo) (*RouteHarvester
...
+				firstArg := jsExtractor.FirstArg(args)
+
+				dir := findScriptDirectory(rootDir, firstArg, workdir)
+				if dir == "" {
+					h.log.Debug("failed to find script directory", "directory", dir, "root", rootDir, "firstArg", firstArg, "cwd", workdir)
+					resultChan <- result{err: fmt.Errorf("failed to find script directory for pid %d, script %s, cwd %s", fileInfo.Pid, firstArg, workdir)}
+					return
+				}
+				h.log.Debug("found node js application directory", "directory", dir, "root", rootDir, "firstArg", firstArg, "cwd", workdir)
+				err = jsExtractor.ScanDirectory(dir)
```
NimrodAvni78: just a thought, maybe we can have some common directories to ignore, mostly sutff like `node_modules` which routes under them probably wont be used and might cause some false positive

pkg/internal/transform/route/harvest/harvester.go
```diff
@@ -56,6 +61,9 @@ func NewRouteHarvester(cfg *services.RouteHarvestingConfig, disabled []string, t
 		if strings.ToLower(lang) == "java" {
 			dMap[svc.InstrumentableJava] = struct{}{}
 		}
+		if strings.ToLower(lang) == "nodejs" || strings.ToLower(lang) == "node.js" {
```
NimrodAvni78: might be out of the scope of this pr
but can we change to config `DisabledRouteHarvesters` to be an array a named type of the language, something [like this](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/926)
or just use the langauges type we already have

pkg/internal/transform/route/harvest/harvester.go
```diff
@@ -121,7 +130,52 @@ func (h *RouteHarvester) HarvestRoutes(fileInfo *exec.FileInfo) (*RouteHarvester
 			} else {
 				resultChan <- result{r: nil}
 			}
-		} else {
+		case svc.InstrumentableNodejs:
+			if _, ok := h.disabled[svc.InstrumentableNodejs]; !ok {
```
NimrodAvni78, line 133: maybe we can also move this to some function like `nodejsExtractRoutes`

pkg/internal/transform/route/harvest/js_test.go
```diff
@@ -0,0 +1,1018 @@
...
+			},
+		},
+		{
+			name:  "not a route() pattern",
+			line:  "  app.get('/users', handler)",
+			found: false,
+		},
+		{
+			name:  "route() with variable",
+			line:  "  app.route(apiPath)",
```
skl, line 271: Presumably, the regex approach cannot support variables and/or interpolation... You could potentially use something like https://github.com/dop251/goja - maybe only as a fallback where variables or interpolation are detected?

pkg/internal/transform/route/harvest/harvester.go
grcevski: I have quite a bit that I skip :), node_modules included, but I also skip the Linux root directories since in Dockerfiles this will cause insane amount of walking. Full list here:
https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/924/files#diff-8fe83386c3423f135b008340593ec8186fbfd16340ccb2b0c305f5266cc388d7R18

pkg/internal/transform/route/harvest/harvester.go
grcevski: Good idea! Let me make a list somehow, we have a InstrumentableLanguage type, but it's all ints, I can make a related struct that has the names.

pkg/internal/transform/route/harvest/harvester.go
grcevski: Good idea! I'll extract the logic.

pkg/internal/transform/route/harvest/js_test.go
grcevski: Do you think we should make a follow up issue for this? I think variables might be hard, especially with parsing conditionals etc.

skl: Yeah could definitely be it's own feature

grcevski: Added an issue to track this here: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/929

pkg/internal/transform/route/harvest/harvester.go
```diff
@@ -58,10 +56,10 @@ func (e *HarvestError) Error() string {
 func NewRouteHarvester(cfg *services.RouteHarvestingConfig, disabled []string, timeout time.Duration) *RouteHarvester {
```
NimrodAvni78, line 56: @grcevski so the config is still a string[] but we compare it to the string enum?
maybe we can create `MarshalText` and `UnmarshalText` on `InstrumentableType` to allow users to specify it and have config type safety?

grcevski: Good idea! I will follow-up with a PR to fix this, either standalone or when I add ruby.

PR #926, opened 2025-11-25
Add multiple named types for improved readability and maintainability

pinoOgni:
In this first PR, the following named types are created:

- Feature
- Instrumentation
- LogLevel

I saw that this pattern is already used elsewhere in OBI and it makes the code cleaner: now a developer can simply click on the type and see a list of all possible constants for that particular field.

The PR is split in multiple commits to make the review process smoother.

### DIFF
--- a/pkg/appolly/instrumenter_test.go
+++ b/pkg/appolly/instrumenter_test.go
@@ -77,7 +77,7 @@ func TestBasicPipeline(t *testing.T) {
 		MetricsEndpoint: tc.ServerEndpoint, Interval: 10 * time.Millisecond,
 		ReportersCacheLen: 16,
 		TTL:               5 * time.Minute,
-		Instrumentations: []string{
+		Instrumentations: []instrumentations.Instrumentation{
 			instrumentations.InstrumentationALL,
 		},
 	}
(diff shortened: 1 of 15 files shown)


pkg/export/otel/otelcfg/common.go
```diff
@@ -58,20 +58,24 @@ const (
 	envResourceAttrs   = "OTEL_RESOURCE_ATTRIBUTES"
 )
 
+type Feature string
```
mariomac, line 61: Since this feature applies exclusively to metrics and the package name just mentions "otelcfg", should we name it as `MetricFeature`?

mariomac (approved):
LGTM! I have a suggestion about naming but happy to leave it as now.

pkg/export/otel/otelcfg/common.go
pinoOgni: Hi @mariomac, I'm working on a common package for the `Feature` type  because both `PrometheusConfig` and `MetricsConfig` are using it. 
So perhaps `Feature` is more generic even if in both cases we are still talking about metrics. 

What do you think? I'm fine with both options.

mariomac: Fine! Let's see then after you move the Features to a common package.

Should we then merge this PR, or are you moving the package within this same PR?

pinoOgni: I'll move it in another PR. You can merge this one. Thanks! :)

PR #927, opened 2025-11-25
Metric features: replace string by otelcfg.MetricFeature type

mariomac:
Adds an extra bit of type safety.


mmat11:
@mariomac I believe this overlaps with https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/926

PR #930, opened 2025-11-25
Move Buckets and Feature definitions to pkg/export

pinoOgni:
This PR moves `Buckets` and `Feature` definitions from `otelcfg` pkg to `pkg/export` to allow shared usage by different packages, such as `prom` and `otel`.


mariomac (approved):
👏🏻

Issue #923, opened 2025-11-24
Support per-process declarative configuration


dashpole:
### Problem Statement

The OpenTelemetry Operator is [planning](https://github.com/open-telemetry/opentelemetry-operator/issues/4093) to rework their Instrumentation resource to support the new OpenTelemetry declarative config file format as part of their v1beta1 milestone. It would be really nice if I was able to use OBI as the implementation of the Instrumentation Kubernetes resources in the cluster the same way that they can be used with auto-instrumentation agents today. That would enable some pretty neat consumption experiences, like changing the sampling rate while an application is running, or changing which attributes are aggregated away with a View.

From a broader perspective, I want to be able to use OBI as a drop-in replacement for manual instrumentation.  Ideally, this means I get the same configuration capabilities (i.e. declarative configuration) at the same granularity (per-process).

Many of the pieces are already there: pod-label and k8s-namespace based service discovery, (global) exporter configuration, sampling config, aggregations, etc.

### Proposal

Overall, I propose starting with a subset of the declarative configuration, and adding support for more of it over time.

To start, migrate https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/94dc45ea3ac81cb46d964799af68503039995152/pkg/appolly/services/attr_glob.go#L82 and https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/94dc45ea3ac81cb46d964799af68503039995152/pkg/appolly/services/attr_glob.go#L80 to be structured the same way as the declarative configuration is structured.  Then, add fields over time as.

Potential issues and concerns (i'm sure there are many):

* Many exporters will be more resource-intensive than a single exporter, and lose some of the benefits of batching. This might be a bit of a footgun for users.
* Some parts of the declarative configuration might be impossible for us to support in a way that is exactly the same.  For example, OBI can't (can it?) open a port in the containers' namespace to serve a Prometheus exporter /metrics endpoint. Telemetry sent from the exporter also won't have the source IP of the pod, which means it won't interact with the k8sattributes processor the same way.
* Some parts of the declarative config won't apply: We (presumably) won't implement a "simple span processor" that blocks while the span is exported.

We need to think about how we will handle cases where the declarative configuration can't be implemented, or doesn't make sense. It would be nice if we could ignore unsupported config fields, and log a warning, rather than crashing.

### Alternative considered

[...]

PR #934, opened 2025-11-26
Replace []export.Feature by bitmap

mariomac:
In preparation of https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/923 , we will need to filter different types of metric exporters on a per-service basis. This requires something faster than `slices.Contains[string]`.

This PR replaces the `[]export.Feature` strings slice by a bitmap approach where looking for a given component is as efficient as checking a bit in an integer.

Nothing changes from a user perspective. We provide `UnmarshalYAML`and `UnmarshalText` methods to handle the stringlist-to-uint conversion.


rafaelroquetto (approved):
Awesome!

PR #941, opened 2025-11-27
Uniform debug print messages in ebpf code

pinoOgni:
This PR is a proposal to try to find a more uniform way to print ebpf debug information.

I only modified a few debug prints as a test and wrote the `bpf-print-format.md` file to document the proposal.

I'm open to suggestions.


bpf/tctracer/tctracer.c
```diff
@@ -58,18 +58,24 @@ static __always_inline void update_outgoing_request_span_id(pid_connection_info_
...
     go_addr_key_t *g_key = bpf_map_lookup_elem(&go_ongoing_http, e_key);
     if (g_key) {
-        bpf_dbg_printk("Found Go HTTP info, trying to find the span id");
+        bpf_dbg_printk("%s: found Go HTTP info, trying to find the span id", __FUNCTION__);
         http_func_invocation_t *invocation =
             bpf_map_lookup_elem(&go_ongoing_http_client_requests, g_key);
         if (invocation) {
-            bpf_dbg_printk(
-                "Found Go HTTP invocation, resetting the span id to %x%x", tcp->seq, tcp->ack);
+            bpf_dbg_printk("%s: found Go HTTP invocation, resetting the span id to seq=%x, ack=%x",
```
rafaelroquetto: I believe this is a mistake - the span id here is indeed meant to be the combination of `%x%x`, i.e. the concatenation of `seq` and `ack`.

devdocs/bpf-print-format.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+Log output: `extend_and_write_tp: failed to push data: -1`.
+
+#### Avoiding concatenation
+
+Addressing issues like logging multiple variables without a separator (e.g., avoiding `%x%x`). Example:
+
+```c
+bpf_dbg_printk("%s: found HTTP info, resetting the span id to seq=%x, ack=%x", __FUNCTION__, tcp->seq, tcp->ack);
```
rafaelroquetto: I get the point, but this is a bad example (see previous comment)

devdocs/bpf-print-format.md
```diff
@@ -0,0 +1,117 @@
...
+This section ensures every log message is contextualized.
+
+We use **`__FUNCTION__`** which is a compile-time string literal (not a runtime function call) to print the function name.
+
+#### Function name in eBPF probe
+
+At the beginning of an eBPF probe (not every probe, as it can be too verbose\!), print the function name in the following way:
+
+```c
+bpf_dbg_printk("=== %s ===", __FUNCTION__);
```
rafaelroquetto: we could wrap this in a macro, so that one simply writes:

```
bpf_dbg_log_func()
```

devdocs/bpf-print-format.md
```diff
@@ -0,0 +1,117 @@
...
+
+```c
+bpf_d_printk("%s: tailcall failed", __FUNCTION__);
+```
+
+Log output: `obi_packet_extender: tailcall failed`.
+
+#### Function name in generic functions
+
+To print the function name in a generic function, use the **`__FUNCTION__`** identifier in **every log statement** without the bounding **`===`**. The triple equals signs are reserved exclusively for entry points of eBPF probes. Example:
```
rafaelroquetto: I see the value here, but there's an issue as others have pointed out: some kernels only allow at most 3 arguments, and now we are "wasting" one with `__FUNCTION__` - this has been a problem before, and you will notice that because of that some bpf_dbg_printk statements had to be split into at least 2 as a workaround.

This why I've explicitly opted to spell out the function name as part of the format string rather than using `__FUNCTION__` in `tpinjector.c` and friends. I think it's a fair trade-off, given we are dealing with eBPF, but YMMV.

Now, if we want something like that anyway, we could probably have a macro like this:

```
#define bpf_func_printk(fmt, ...) \
    bpf_printk("%s: " fmt, __func__, ##__VA_ARGS__)
```

(or wrap around bpf_dbg_printk, I haven't tried with that).

If we do that, it should be a separate macro, whilst `bpf_dbg_printk` remains as is, giving the caller the flexibility to pick - think of it as a convenience macro.

devdocs/bpf-print-format.md
```diff
@@ -0,0 +1,117 @@
...
+bpf_dbg_printk("obi_packet_extender(%s): generating tp info", __FUNCTION__);
+```
+
+Log output: `obi_packet_extender(create_trace_info): generating tp info`.
+
+### Compatibility and workarounds
+
+It can happen, sometimes, that the number of arguments to print is 3, but when `__FUNCTION__` is included, the total reaches 4. In these cases, `bpf_dbg_printk` calls `bpf_printk`, which subsequently calls `___bpf_pick_printk`. This chain then selects and calls `__bpf_vprintk` (for 4+ arguments), which finally invokes `bpf_trace_vprintk`. Since `bpf_trace_vprintk` is only available starting in [kernel version 5.16](https://docs.ebpf.io/linux/helper-function/bpf_trace_vprintk/), two options are available to avoid errors when loading the eBPF program:
+
+1. Split the print into two separate calls: one with one argument and one with two arguments, where `__FUNCTION__` is added to both.
```
rafaelroquetto: So, in light of this comment, perhaps the way is to simply provide the convenience macro above - makes it flexible.

bpf/tctracer/tctracer.c
pinoOgni: ops, thanks!

devdocs/bpf-print-format.md
pinoOgni: yes, thanks

pinoOgni:
Hi @grcevski and @rafaelroquetto, about the wrapper/macro solution with `snprintf`. 
I modified `bpf_dbg_helper` like this:
```
...
            if (bpf_core_enum_value_exists(enum bpf_func_id___x, BPF_FUNC_snprintf___x)) {         \
                /* snprintf is available: Include the function name and the formatted message */     \ 
                /* we use __FUNCTION__ ": " fmt to prepend the function name to the log   */          \
                BPF_SNPRINTF((char *)__trace__->log,                                               \
                             sizeof(__trace__->log),                                               \
                             "%s: " fmt,                                                            \
                             __FUNCTION__,                                                           \
                             ##args);                                                              \
...
```
It works, and now we no longer have to write `__FUNCTION__` every time we use `bpf_dbg_printk`, but:
1. If we use this solution, we need to modify all `bpf_dbg_printk` where the function name is already written, for example `parse_ip_options`: I can do that.
2. When we have something like `bpf_dbg_printk("=== uprobe/proc http2WriteHeaders === ");` at the beginning of an ebpf probe, it will print `obi_uprobe_http2WriteHeaders: === uprobe/proc http2WriteHeaders ===`. What do you think?

grcevski:
This looks good, I would just like to reverse the order, first text then the function name. If the function name is cut off because of the string size, it's less damaging than if we cut off the text.

pinoOgni:
> This looks good, I would just like to reverse the order, first text then the function name. If the function name is cut off because of the string size, it's less damaging than if we cut off the text.

Nice suggestion!

grcevski (approved):
Amazing! LGTM!

pinoOgni:
> We had a failure on 5.15, I've restarted, but I wonder if we somewhere went above 3 arguments?

I thought the same thing, but the usage of `snprintf` is in `bpf_dbg_helper` while the argument limit is in `bpf_printk`, so `__FUNCTION__` doesn't affect it and we're safe.

PR #942, opened 2025-11-27
Implement trace-log correlation

mmat11:
## Checklist

### DIFF
--- a/pkg/appolly/discover/matcher.go
+++ b/pkg/appolly/discover/matcher.go
@@ -35,14 +35,15 @@ func criteriaMatcherProvider(
 ) swarm.InstanceFunc {
 	beylaNamespace, _ := namespaceFetcherFunc(int32(osPidFunc()))
 	m := &Matcher{
-		Log:              slog.With("component", "discover.CriteriaMatcher"),
-		Criteria:         FindingCriteria(cfg),
-		ExcludeCriteria:  ExcludingCriteria(cfg),
-		ProcessHistory:   map[PID]ProcessMatch{},
-		Input:            input.Subscribe(msg.SubscriberName("discover.CriteriaMatcher")),
-		Output:           output,
-		Namespace:        beylaNamespace,
-		HasHostPidAccess: hasHostPidAccess(),
+		Log:                 slog.With("component", "discover.CriteriaMatcher"),
+		Criteria:            FindingCriteria(cfg),
+		ExcludeCriteria:     ExcludingCriteria(cfg),
+		LogEnricherCriteria: LogEnricherFindingCriteria(cfg),
+		ProcessHistory:      map[PID]ProcessMatch{},
+		Input:               input.Subscribe(msg.SubscriberName("discover.CriteriaMatcher")),
+		Output:              output,
+		Namespace:           beylaNamespace,
+		HasHostPidAccess:    hasHostPidAccess(),
 	}
 	return swarm.DirectInstance(m.Run)
 }
@@ -50,9 +51,10 @@ func criteriaMatcherProvider(
 // Matcher is the component that matches the processes against the discovery criteria.
 // It filters the processes that match the discovery criteria and sends them to the output channel.
 type Matcher struct {
-	Log             *slog.Logger
-	Criteria        []services.Selector
-	ExcludeCriteria []services.Selector
+	Log                 *slog.Logger
+	Criteria            []services.Selector
+	ExcludeCriteria     []services.Selector
+	LogEnricherCriteria []services.Selector
 	// ProcessHistory keeps track of the processes that have been already matched and submitted for
 	// instrumentation.
 	// This avoids keep inspecting again and again client processes each time they open a new connection port
@@ -65,8 +67,13 @@ type Matcher struct {
 
 // ProcessMatch matches a found process with the first selection criteria it fulfilled.
 type ProcessMatch struct {
-	Criteria []services.Selector
-	Process  *services.ProcessInfo
+	Criteria            []services.Selector
+	LogEnricherCriteria []services.Selector
+	Process             *services.ProcessInfo
+}
+
+func (pm ProcessMatch) LogEnricherEnabled() bool {
+	return len(pm.LogEnricherCriteria) > 0
 }
 
 func (m *Matcher) Run(ctx context.Context) {
@@ -105,18 +112,36 @@ func (m *Matcher) alreadyMatched(pid PID) bool {
 
 func (m *Matcher) matchCriteria(obj ProcessAttrs, proc *services.ProcessInfo) *ProcessMatch {
 	criteria := make([]services.Selector, 0, len(m.Criteria))
-
 	for i := range m.Criteria {
 		if m.matchProcess(&obj, proc, m.Criteria[i]) && !m.isExcluded(&obj, proc) {
 			criteria = append(criteria, m.Criteria[i])
 		}
 	}
 
+	logEnricherCriteria := make([]services.Selector, 0, len(m.LogEnricherCriteria))
+	for i := range m.LogEnricherCriteria {
+		if m.matchProcess(&obj, proc, m.LogEnricherCriteria[i]) {
+			logEnricherCriteria = append(logEnricherCriteria, m.LogEnricherCriteria[i])
+		}
+	}
+
 	if len(criteria) > 0 {
 		m.Log.Debug("found process", "pid", proc.Pid, "comm", proc.ExePath, "metadata",
-			obj.metadata, "podLabels", obj.podLabels, "criteria", criteria)
+			obj.metadata, "podLabels", obj.podLabels, "criteria", criteria, "logEnricherCriteria", logEnricherCriteria)
 
-		return &ProcessMatch{Criteria: criteria, Process: proc}
+		return &ProcessMatch{Criteria: criteria, LogEnricherCriteria: logEnricherCriteria, Process: proc}
+	}
+
+	if len(logEnricherCriteria) > 0 {
+		m.Log.Info(
+			"avoiding log instrumentation for process, since it wasn't matched for instrumentation. Log enrichment criteria must be a subset of instrumentation matching criteria",
+			"pid", proc.Pid,
+			"comm", proc.ExePath,
+			"metadata", obj.metadata,
+			"podLabels", obj.podLabels,
+			"criteria", criteria,
+			"logEnricherCriteria", logEnricherCriteria,
+		)
 	}
 
 	return nil
@@ -295,6 +320,20 @@ func normalizeRegexCriteria(finderCriteria services.RegexDefinitionCriteria) []s
 		criteria = append(criteria, fc)
 	}
 	return criteria
+}
+
+func LogEnricherFindingCriteria(cfg *obi.Config) []services.Selector {
+	var selectors []services.Selector
+
+	if !cfg.EBPF.LogEnricher.Enabled() {
+		return selectors
+	}
+
+	for _, svcs := range cfg.EBPF.LogEnricher.Services {
+		selectors = append(selectors, NormalizeGlobCriteria(svcs.Service)...)
+	}
+
+	return selectors
 }
 
 func FindingCriteria(cfg *obi.Config) []services.Selector {
--- a/pkg/internal/ebpf/logenricher/logenricher.go
+++ b/pkg/internal/ebpf/logenricher/logenricher.go
@@ -83,6 +83,7 @@ func New(pf ebpfcommon.ServiceFilter, cfg *obi.Config) *Tracer {
 	)
 
 	tr.asyncWriter = asyncWriter
+
 	return tr
 }
 
@@ -288,7 +289,7 @@ func (p *Tracer) handleLogEvent(_ *ebpfcommon.EBPFParseContext, _ *config.EBPFTr
 	if err != nil {
 		// This should never happen -- if it does, we can't really recover
 		// and the targeted process will miss his logs.
-		return request.Span{}, true, err
+		return request.Span{}, true, nil
 	}
 
 	err = p.asyncWriter.Enqueue(p.ctx, LogEvent{
--- a/internal/test/integration/log_enricher_test.go
+++ b/internal/test/integration/log_enricher_test.go
@@ -74,7 +74,7 @@ func testLogEnricher(t *testing.T) {
 	require.NoError(t, err)
 	defer cl.Close()
 
-	test.Eventually(t, testTimeout, func(t require.TestingT) {
+	test.Eventually(t, 2*testTimeout, func(t require.TestingT) {
 		ti.DoHTTPGet(t, serverURL+jsonEndpoint, 200)
 
 		containerID := testContainerID(t, cl, containerImage)
--- a/pkg/appolly/app/svc/svc.go
+++ b/pkg/appolly/app/svc/svc.go
@@ -120,6 +120,8 @@ type Attrs struct {
 	// Service Graph Metrics for all the services but RED metrics only for a few
 	Features export.Features
 
+	LogEnricherEnabled bool
+
 	CustomInRouteMatcher  route.Matcher
 	CustomOutRouteMatcher route.Matcher
 	HarvestedRouteMatcher route.Matcher
--- a/pkg/appolly/discover/attacher.go
+++ b/pkg/appolly/discover/attacher.go
@@ -90,7 +90,6 @@ func (ta *traceAttacher) attacherLoop(_ context.Context) (swarm.RunFunc, error) 
 	}
 	ta.processInstances = maps.MultiCounter[uint64]{}
 	ta.EbpfEventContext.CommonPIDsFilter = ebpfcommon.NewPIDsFilter(&ta.Cfg.Discovery, slog.With("component", "ebpfCommon.CommonPIDsFilter"), ta.Metrics)
-	ta.EbpfEventContext.LogEnricherPIDsFilter = ebpfcommon.NewPIDsFilter(&ta.Cfg.Discovery, slog.With("component", "ebpfCommon.LogEnricherPIDsFilter"), ta.Metrics)
 	ta.routeHarvester = harvest.NewRouteHarvester(&ta.Cfg.Discovery.RouteHarvestConfig, ta.Cfg.Discovery.DisabledRouteHarvesters, ta.Cfg.Discovery.RouteHarvesterTimeout)
 	ta.processAgeFunc = ProcessAgeFunc()
 
@@ -157,12 +156,14 @@ func (ta *traceAttacher) getTracer(ie *ebpf.Instrumentable) bool {
 		ta.log.Debug(".done", "success", ok)
 		return ok
 	}
+
 	ta.log.Info("instrumenting process",
 		"cmd", ie.FileInfo.CmdExePath,
 		"pid", ie.FileInfo.Pid,
 		"ino", ie.FileInfo.Ino,
 		"type", ie.Type,
 		"service", ie.FileInfo.Service.UID.Name,
+		"logenricher", ie.FileInfo.Service.LogEnricherEnabled,
 	)
 
 	// builds a tracer for that executable
@@ -183,7 +184,7 @@ func (ta *traceAttacher) getTracer(ie *ebpf.Instrumentable) bool {
 				// instance of the executable has different DLLs loaded, e.g. libssl.so.
 				return ta.reuseTracer(ta.reusableTracer, ie)
 			} else {
-				programs = ta.withCommonTracersGroup(newGenericTracersGroup(ta.EbpfEventContext.CommonPIDsFilter, ta.EbpfEventContext.LogEnricherPIDsFilter, ta.Cfg, ta.Metrics))
+				programs = ta.withCommonTracersGroup(newGenericTracersGroup(ta.EbpfEventContext.CommonPIDsFilter, ta.Cfg, ta.Metrics))
 			}
 		} else {
 			if ta.reusableGoTracer != nil {
@@ -196,7 +197,7 @@ func (ta *traceAttacher) getTracer(ie *ebpf.Instrumentable) bool {
 		if ta.reusableTracer != nil {
 			return ta.reuseTracer(ta.reusableTracer, ie)
 		}
-		programs = ta.withCommonTracersGroup(newGenericTracersGroup(ta.EbpfEventContext.CommonPIDsFilter, ta.EbpfEventContext.LogEnricherPIDsFilter, ta.Cfg, ta.Metrics))
+		programs = ta.withCommonTracersGroup(newGenericTracersGroup(ta.EbpfEventContext.CommonPIDsFilter, ta.Cfg, ta.Metrics))
 	default:
 		ta.log.Warn("unexpected instrumentable type. This is basically a bug", "type", ie.Type)
 	}
--- a/pkg/appolly/discover/finder.go
+++ b/pkg/appolly/discover/finder.go
@@ -151,7 +151,7 @@ func newGoTracersGroup(pidFilter ebpfcommon.ServiceFilter, cfg *obi.Config, metr
 	return []ebpf.Tracer{gotracer.New(pidFilter, cfg, metrics)}
 }
 
-func newGenericTracersGroup(pidFilter, logEnricherFilter ebpfcommon.ServiceFilter, cfg *obi.Config, metrics imetrics.Reporter) []ebpf.Tracer {
+func newGenericTracersGroup(pidFilter ebpfcommon.ServiceFilter, cfg *obi.Config, metrics imetrics.Reporter) []ebpf.Tracer {
 	var tracers []ebpf.Tracer
 
 	// Add tracers based on configuration
@@ -164,7 +164,7 @@ func newGenericTracersGroup(pidFilter, logEnricherFilter ebpfcommon.ServiceFilte
 
 	// Enables log enricher which handles trace-log correlation
 	if cfg.EBPF.LogEnricher.Enabled() {
-		logEnricher := logenricher.New(logEnricherFilter, cfg)
+		logEnricher := logenricher.New(pidFilter, cfg)
 		if logEnricher != nil {
 			tracers = append(tracers, logEnricher)
 		}
--- a/pkg/appolly/discover/typer.go
+++ b/pkg/appolly/discover/typer.go
@@ -133,11 +133,12 @@ func (t *typer) makeServiceAttrs(processMatch *ProcessMatch) svc.Attrs {
 			Name:      name,
 			Namespace: namespace,
 		},
-		ProcPID:     processMatch.Process.Pid,
-		ExportModes: exportModes,
-		Sampler:     samplerFromConfig(samplerConfig),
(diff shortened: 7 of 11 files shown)


pkg/internal/ebpf/logenricher/logenricher.go
```diff
@@ -0,0 +1,316 @@
...
+
+func New(pf ebpfcommon.ServiceFilter, cfg *obi.Config) *Tracer {
+	return &Tracer{
+		log: slog.With("component", "logenricher"),
+		cfg: cfg,
+		pf:  pf,
+	}
+}
+
+func (p *Tracer) Load() (*ebpf.CollectionSpec, error) {
```
grcevski, line 90: Since we use bpf_probe_write_user, can we please check like we do in the gotracer if this is allowed:
```
func (p *Tracer) supportsContextPropagation() bool {
	return !ebpfcommon.IntegrityModeOverride && ebpfcommon.SupportsContextPropagationWithProbe(p.log)
}

```

If possible, when this is not allowed we should not load the program at all and tell people with WARN that log enrinchment will not work because they are on a locked kernel.

pkg/internal/ebpf/logenricher/logenricher.go
```diff
@@ -0,0 +1,316 @@
...
+	} else {
+		// TTY write
+		filePath = unix.ByteSliceToString(event.FilePath[:])
+		if filePath == "" {
+			// Fallback to process stdout in the case path resolver failed
+			filePath = procFdPath(1)
+		}
+	}
+
+	f, err := os.OpenFile(filePath, os.O_WRONLY, 0)
```
grcevski: Are we concerned this might be slow, opening and closing the file every time? Would it make sense to cache file descriptors with some sort of a timeout?

bpf/logenricher/logenricher.c
```diff
@@ -0,0 +1,191 @@
...
+    iovec_iter_ctx iov_ctx;
+    get_iovec_ctx(&iov_ctx, (struct iov_iter___dummy *)from);
+
+    if (iov_ctx.iter_type != ITER_UBUF) {
+        bpf_dbg_printk("logenricher: unsupported iter_type %d", iov_ctx.iter_type);
+        return 0;
+    }
+
+    size_t count = BPF_CORE_READ(from, count);
+    long offset = bpf_core_field_offset(struct iov_iter, count) - 8;
```
rafaelroquetto: could these also be const?

bpf/logenricher/logenricher.c
```diff
@@ -0,0 +1,191 @@
...
+
+    struct iovec iov = {};
+    bpf_probe_read(&iov, sizeof(iov), (char *)from + offset);
+
+    u64 pid_tgid = bpf_get_current_pid_tgid();
```
rafaelroquetto: idem

bpf/logenricher/logenricher.c
```diff
@@ -0,0 +1,191 @@
...
+    e->pid_tp = *tp_info;
+    e->fd = fd;
+    bpf_probe_read_user(e->log, e->len, iov.iov_base);
+
+    if (fd == 0) {
+        struct path path = BPF_CORE_READ(iocb, ki_filp, f_path);
+        resolve_path((char *)e->file_path, &path, task);
+    } else {
+        e->file_path[0] = '\0';
+    }
```
rafaelroquetto, line 101: would it be possible to add a comment explaining what this block does (or even splitting it into its own function if you think that is feasible)?

bpf/logenricher/logenricher.c
```diff
@@ -0,0 +1,191 @@
...
+
+    if (e->len > 0) {
+        // from this point on, the responsibility of writing to stdout is on us,
+        // so if something fails, we must always fallback to writing the original data
+        long err = bpf_ringbuf_output(
```
rafaelroquetto: I think this can also be `const`

bpf/logenricher/logenricher.c
```diff
@@ -0,0 +1,191 @@
...
+        // delete current buffer to avoid double logging
+        char *zero = bpf_map_lookup_elem(&zeros, &(u32){0});
+        if (!zero) {
+            bpf_dbg_printk("logenricher: failed to get zero buffer");
+            return 0;
+        }
+
+        // reapply bounds for the verifier
+        const u16 zeroed_count = e->len & k_log_event_max_log_mask;
+        if (zeroed_count) { // this is never zero, but the verifier is not smart enough
```
rafaelroquetto: I _think_ this is because it's theoretically not guaranteed that `e->len` hasn't been modified after your first check to `if (e->len > 0)` - it's a pointer to a map element - so theoretically userspace on another CPU could modify this element before this instruction point is reached.

You can try to store the len on the stack and give this a shot:

```
const size_t effective_len = count & k_log_event_max_log_mask;
e->len = effective_len;

/* ... use effective_len instead of e->len henceforth ... */

if (effective_len > 0) {
    /* ... */
}

```

bpf/logenricher/path_resolver.h
```diff
@@ -0,0 +1,144 @@
...
+    struct dentry **dentry_arr;
+
+    // Ensure we make buf an empty string early up here so if we exit with any
+    // sort of error, we won't leave garbage in it if it's uninitialized
+    buf[0] = '\0';
+
+    if (!(dentry_arr = bpf_map_lookup_elem(&path_resolver_scratch, &(u32){0}))) {
+        bpf_dbg_printk("resolve_path: could not get path resolver scratch area");
+        goto out_err;
+    }
```
rafaelroquetto, line 43:
```suggestion
    buf[0] = '\0';

    dentry_arr = bpf_map_lookup_elem(&path_resolver_scratch, &(u32){0});

    if (!dentry_arr) {
        bpf_dbg_printk("resolve_path: could not get path resolver scratch area");
        return false;
    }
```

rationale: clarity (and no need to set buf[0] twice)

bpf/logenricher/path_resolver.h
```diff
@@ -0,0 +1,144 @@
...
+#include <logger/bpf_dbg.h>
+
+#include <logenricher/types.h>
+
+#include <logenricher/maps/path_resolver_scratch.h>
+
+// Resolve a struct path to a string. Returns a boolean indicating whether the
+// path construction errored/was truncated or was successful.
+static __always_inline bool
+resolve_path(char *buf, struct path *path, const struct task_struct *task) {
```
rafaelroquetto, line 21: we should try to pass the buf_size here instead of implicitly relying on k_pts_file_path_len_max

bpf/logenricher/path_resolver.h
```diff
@@ -0,0 +1,144 @@
...
+
+    if (!(dentry_arr = bpf_map_lookup_elem(&path_resolver_scratch, &(u32){0}))) {
+        bpf_dbg_printk("resolve_path: could not get path resolver scratch area");
+        goto out_err;
+    }
+
+    // Loop 1, follow the dentry chain (up to a maximum of
+    // k_path_resolver_max_components) and store pointers to each dentry in
+    // dentry_arr
+    for (int i = 0; i < k_path_resolver_max_components; i++) {
```
rafaelroquetto, line 48: I wonder if we could simplify this into a single loop by simply walking backwards and write directly to buf. Something like this (pseudo-code):

```
static __always_inline size_t
resolve_path(char *buf, size_t buflen, struct path *path, const struct task_struct *task) {
    /* ... setup task_struct, etc ... */

    /* these are the main pointers, we iterate backwards into dentry, and then
     * once we hit the current mnt_root, we swap the pointers, rinse and
     * repeat
     */
    struct vfsmount *mnt;
    struct dentry *dentry;

    char *end = buf + buflen;

    *--end = '\0';

    /* so here we start from the leaf path part */
    mnt = path->mnt;
    dentry = path->dentry;

    for (u32 i = 0; i < k_path_resolver_max_components; ++i) {
        struct dentry *parent;

        /* while we are in the samen mnt, we keep prepending the path parts
         * for each dentry
         */

        while (dentry != mnt->mnt_root) {
            int len = dentry->d_name.len;

            /* account for '/' + name */
            if (end - buf < len + 1) {
                /* fail returning 0 bytes written */
                return 0u;
            }

            /* copy name backwards - this is pseudo / regular C, need to read
             * the len, qstr, etc... with bpf_probe_read_kernel() or
             * BPF_CORE_READ()
             */
            end -= len;

            /* read the path part  and preprent '/' */
            bpf_probe_read_kernel(end, len, dentry->d_name.name);

            *--end = '/';

            /* move to parent dentry */
            parent = dentry->d_parent;

            if (parent == dentry)
                break;

            dentry = parent;
        }

        /* if we getting here, we got to the current's dentry chain mnt_root -
         * so all we do is rinse and repeat and switch the dentry and mnt
         * pointers to the next level
         */
        if (mnt == mnt->mnt_parent)
            break;

        dentry = mnt->mnt_mountpoint;
        mnt = mnt->mnt_parent;
    }

    // If we didn't write anything but the terminator, path is "/"
    if (*end == '\0') {
        if (end == buf)
            return 0;
        *--end = '/';
    }

    /* return the number of bytes read, you can use 0 to check if anything was
     * written
     */
    return buflen - end + buf;
}
```

or something like that - then in _userspace_ you can simply iterate `bytes_written` event from the end of the buffer to extract the path. This makes it simpler and does ithout the need for the map to collect dentry pointers

bpf/logenricher/types.h
```diff
@@ -0,0 +1,97 @@
...
+    BPF_CORE_READ_INTO(&dev->minor, tty, driver, minor_start);
+    dev->minor += BPF_CORE_READ(tty, index);
+    dev->termios.c_lflag = BPF_CORE_READ(tty, termios.c_lflag);
+}
+
+static __always_inline bool tty_driver_is_pty(struct tty_struct *tty) {
+    if (bpf_core_enum_value_exists(enum tty_driver_type__new, TTY_DRIVER_TYPE_PTY)) {
+        int typ = bpf_core_enum_value(enum tty_driver_type__new, TTY_DRIVER_TYPE_PTY);
+        return BPF_CORE_READ(tty, driver, type) == typ;
+    } else {
```
rafaelroquetto: nit: we should remove this `else`, it's redundant (we should probably enable `-Wredundant-parens` in clang-tidy).

pkg/internal/ebpf/logenricher/logenricher.go
rafaelroquetto: Yeah, I believe we need a different approach here - opening and closing it for each event can cause a massive overhead.

mmat11: makes sense, I can add a fd cache

bpf/logenricher/logenricher.c
mmat11: I think you are right, I'll try next week

mmat11:
> Another pass. We should run some scalability tests here before being able to merge this (e.g. deploy it in a big cluster and gather performance data). I am worried about a few points:
> 
> pressure on the ringbuffer: the shared ringbuffer  will now be flooded by lines of logs  of every tracked process - that alone is concerning, and it gets worse if we take into account the same ring buffer is used for other kinds of events

I am using a dedicated ring buffer (see `log_events`), we can discuss whether we want its size configurable (or auto-adjusting depending on the number of tracked services)

> overhead of context-switching for each `_write` kprobe / CPU usage and ksoftirq handlers - you can gather performance data using `perf` for that

Will do that 👍🏻 

> The handler for the logs should probably grab the event, and put it into a processing queue to be picked up by another thread or thread pool - doing heavy work like deserialising json or opening and closing files (i.e. I/O) is very slow and will block the ring buffer. That handler should be doing the minimum work possible and return (e.g. copy the event, enqueue it and return).

I didn't want to add additional complexity without having measured some metrics or observed some bottlenecks, maybe we can do this after that step?

> We should also track the target process stdout (or log sink), open it once for writing, and keep the file descriptor open until either the process or OBI itself terminates, as @grcevski suggested.

👍🏻

bpf/logenricher/logenricher.c
mmat11: It really doesn't like it :D 

```
const size_t effective_len = count & k_log_event_max_log_mask;


; bpf_probe_write_user(iov.iov_base, zero, effective_len);
1923: (79) r1 = *(u64 *)(r1 +0)       ; R1_w=scalar() fp-104=mmmmmmmm
; bpf_probe_write_user(iov.iov_base, zero, effective_len);
1924: (bf) r2 = r0                    ; R0_w=map_value(map=zeros,ks=4,vs=16384) R2_w=map_value(map=zeros,ks=4,vs=16384)
1925: (79) r3 = *(u64 *)(r10 -184)    ; R3_w=scalar(smin=smin32=0,smax=umax=smax32=umax32=8191,var_off=(0x0; 0x1fff)) R10=fp0 fp-184=scalar(smin=smin32=0,smax=umax=smax32=umax32=8191,var_off=(0x0; 0x1fff))
1926: (85) call bpf_probe_write_user#36
R3 invalid zero-sized read: u64=[0,8191]
```

bpf/logenricher/path_resolver.h
mmat11: I'd like to avoid changing this code too much, as written on top of the file it's been taken from elastic/ebpf and ideally I'd use it as-is as I have the confidence it just works. This function has a lot of edge-cases handling and I don't wanna risk breaking it.
If the code is too much/too complex I can use the fd-based path construction the way I do for `pipe_write`, I preferred this as there are less moving parts (and less probes)

pkg/internal/ebpf/logenricher/logenricher.go
mmat11: I think it's supposed to work with `integrity` mode enabled, my tests were failing locally because in the container `integrity` is enabled instead of `none` and the tracer failed to start -- should I just return an error on `confidentiality`?

```
root@64be0b6a541e:/# cat /sys/kernel/security/lockdown
none [integrity] confidentiality
```

```
(venv) lima@lima-cx-new:~$ cat /sys/kernel/security/lockdown
[none] integrity confidentiality
```

bpf/logenricher/logenricher.c
rafaelroquetto: can you paste the rest of the verifier log that precedes it? The last line tells me we are not constraining r3 enough and I'd like to see what's going on.

mmat11: unfortunately there's too much in between initialization and usage so it's hard to follow -- I think it's fine using an additional variable for now

bpf/logenricher/path_resolver.h
rafaelroquetto: I don't know... the original code loops over the N components twice, and requires additional loading/storing from an array per part (plus a map lookup every time this is invoked) - whereas the approach I am proposing does without the array altogether and loops once (and no map lookups) in a single pass. This is something that we can test for.

The pseudo-code is lacking a few corner cases (truncation, root path `/` when no other parts exist, for instance) but these could be addressed.

Having said that, I understand this takes (human) time and see where you are coming from. I'd personally give it a shot (and _potentially_ fail miserably) but ultimately it's up to you - so feel free to leave it as is.

An aside point, since this code comes from elastic/ebpf and is dual-licensed (GPL/BSD2) you may want to confirm whether we need to keep the copyright and license notice or somenthing like that.

bpf/logenricher/logenricher.c
rafaelroquetto: It's ok, I am just curious to see the entire log output.

```
1924: (bf) r2 = r0                    ; R0_w=map_value(map=zeros,ks=4,vs=16384) R2_w=map_value(map=zeros,ks=4,vs=16384)
1925: (79) r3 = *(u64 *)(r10 -184)    ; R3_w=scalar(smin=smin32=0,smax=umax=smax32=umax32=8191,var_off=(0x0; 0x1fff)) R10=fp0 fp-184=scalar(smin=smin32=0,smax=umax=smax32=umax32=8191,var_off=(0x0; 0x1fff))
1926: (85) call bpf_probe_write_user#36
R3 invalid zero-sized read: u64=[0,8191]
```

If you look at smin and var_off, you can tell that we are not constraining the scalar away from `0` so the verifer rightfully catches that. But it should be possible to do that.

If you paste the log I can try to help, even if the verifier error is ugly.

bpf/logenricher/path_resolver.h
mmat11: For the first part, I'll give it a shot after load testing. For the second part, I think we should keep copyright, but I don't know how to do it. I added a comment at the top of the file with the source, should I copy something else? /cc @MrAlias

bpf/logenricher/types.h
```diff
@@ -0,0 +1,95 @@
...
+    u32 len;
+    u32 fd;
+    u32 _pad;
+    u8 file_path[k_pts_file_path_len_max];
+    u8 log[];
+} log_event_t;
+
+const log_event_t *log_event__unused __attribute__((unused));
+
+enum tty_driver_type__new {
```
rafaelroquetto:
```suggestion
enum tty_driver_type___new {
```

see below

bpf/logenricher/types.h
```diff
@@ -0,0 +1,95 @@
...
+    TTY_DRIVER_TYPE_SYSTEM,
+    TTY_DRIVER_TYPE_CONSOLE,
+    TTY_DRIVER_TYPE_SERIAL,
+    TTY_DRIVER_TYPE_PTY,
+    TTY_DRIVER_TYPE_SCC,
+    TTY_DRIVER_TYPE_SYSCONS,
+};
+
+enum tty_driver_subtype__new {
```
rafaelroquetto:
```suggestion
enum tty_driver_subtype___new {
```

the mechanism requires three underscores - I think we got lucky here.

Ref: https://kernel.googlesource.com/pub/scm/linux/kernel/git/torvalds/linux/%2B/4333a9b0b67bb4e8bcd91bdd80da80b0ec151162/tools/lib/bpf/libbpf.c#4019

bpf/logenricher/types.h
```diff
@@ -0,0 +1,95 @@
...
+
+static __always_inline void tty_dev__fill(struct tty_dev *dev, struct tty_struct *tty) {
+    BPF_CORE_READ_INTO(&dev->major, tty, driver, major);
+    BPF_CORE_READ_INTO(&dev->minor, tty, driver, minor_start);
+    dev->minor += BPF_CORE_READ(tty, index);
+    dev->termios.c_lflag = BPF_CORE_READ(tty, termios.c_lflag);
+}
+
+static __always_inline bool tty_driver_is_pty(struct tty_struct *tty) {
+    if (bpf_core_enum_value_exists(enum tty_driver_type__new, TTY_DRIVER_TYPE_PTY)) {
```
rafaelroquetto: This is actualy not enough, you need

```
    if (!bpf_core_field_exists(struct tty_driver, type)) {
        return false;
    }
```

otherwise it will fail in kernels that don't have that.

bpf/logenricher/types.h
```diff
@@ -0,0 +1,95 @@
...
+        int typ = bpf_core_enum_value(enum tty_driver_type__new, TTY_DRIVER_TYPE_PTY);
+        return BPF_CORE_READ(tty, driver, type) == typ;
+    }
+    return BPF_CORE_READ(tty, driver, type) == k_tty_driver_type_pty;
+}
+
+static __always_inline bool tty_driver_is_master(struct tty_struct *tty) {
+    if (bpf_core_enum_value_exists(enum tty_driver_subtype__new, PTY_TYPE_MASTER)) {
```
rafaelroquetto: same here for subtype (if that also can vary amongst kernel versions)

bpf/logenricher/logenricher.c
rafaelroquetto: So I have it a run on my kernel and it works fine - here's the diff, in case that helps:

```
diff --git a/bpf/logenricher/logenricher.c b/bpf/logenricher/logenricher.c
index 5c6d2b726..bdc7fd83a 100644
--- a/bpf/logenricher/logenricher.c
+++ b/bpf/logenricher/logenricher.c
@@ -82,8 +82,11 @@ __write(struct kiocb *iocb, struct iov_iter *from, const int fd, const struct ta
         bpf_dbg_printk("logenricher: failed to reserve event space");
         return 0;
     }
+
+    const u32 len = count & k_log_event_max_log_mask;
+
     e->tgid = pid_tgid >> 32;
-    e->len = count & k_log_event_max_log_mask;
+    e->len = len;
     e->pid_tp = *tp_info;
     e->fd = fd;
     bpf_probe_read_user(e->log, e->len, iov.iov_base);
@@ -101,11 +104,11 @@ __write(struct kiocb *iocb, struct iov_iter *from, const int fd, const struct ta
         e->file_path[0] = '\0';
     }
 
-    if (e->len > 0) {
+    if (len > 0) {
         // From this point on, the responsibility of writing to stdout is on us,
         // so if something fails, we must always fallback to writing the original data.
         const long err = bpf_ringbuf_output(
-            &log_events, e, (sizeof(log_event_t) + e->len) & k_log_event_max_size_mask, 0);
+            &log_events, e, (sizeof(log_event_t) + len) & k_log_event_max_size_mask, 0);
         if (err < 0) {
             bpf_dbg_printk("logenricher: failed to write log event to ringbuf");
             return 0;
@@ -118,11 +121,7 @@ __write(struct kiocb *iocb, struct iov_iter *from, const int fd, const struct ta
             return 0;
         }
 
-        // Reapply bounds for the verifier.
-        const u16 zeroed_count = e->len & k_log_event_max_log_mask;
-        if (zeroed_count) {
-            bpf_probe_write_user(iov.iov_base, zero, zeroed_count);
-        }
+        bpf_probe_write_user(iov.iov_base, zero, len);
     }
 
     return 0;
```

bpf/logenricher/path_resolver.h
rafaelroquetto: @mmat11 yeah that sounds good - come to think of it, you are right, let's do the testing first and see if that even matters.

pkg/internal/ebpf/logenricher/logenricher.go
grcevski: Yeah, it would be great if we can just tell the user, sorry this wont' work you have integrity on.

bpf/logenricher/types.h
mmat11: the `type` field is always present

bpf/logenricher/types.h
mmat11: same here

bpf/logenricher/types.h
mmat11: nice catch, don't know why it was working regardless

mmat11:
@rafaelroquetto @grcevski I did add a couple of optimizations, did some tests on a real cluster and applied all the other suggestions (I just left out the pathresolver.h changes). Let me know if I'm missing something

pkg/internal/ebpf/logenricher/logenricher.go
```diff
@@ -0,0 +1,382 @@
...
+
+	if err := p.initPIDsMap(p.bpfObjects.LogEnricherPids); err != nil {
+		p.log.Error("failed to init pids map, not starting", "error", err)
+		return
+	}
+
+	ebpfcommon.ForwardRingbuf(
+		&p.cfg.EBPF,
+		p.bpfObjects.LogEvents,
+		eventCtx.CommonPIDsFilter,
```
grcevski, line 270: Do we need to pass the log pids filter here to the forward ring buff? The common pids filter is for the other bpf programs. Maybe that makes sense actually, you want to filter all logs for non instrumented services. In a sense, tracking logs doesn't make sense if the pid is not also instrumented for tracing.

grcevski (commented):
This looks great @mmat11!

I had one question remaining and it's probably me not being able to see how this is done. I see we have another criteria definition for further filtering of the services we want to include for logging, how is this wired in?

I would've thought we needed maybe an extension to the pipeline to have Matcher filter out pids we don't want.

pkg/internal/ebpf/logenricher/logenricher.go
mmat11: IIRC I passed the common pids filter there because that filtering is happening at span level after data gets to userspace -- all I needed was to keep the ebpf map with pids aligned with the tracked processes (see AllowPID/BlockPID). Maybe I can add another check in ebpf which just calls `valid_pid()` on the pids map?

grcevski: Yeah I think it's fine, I think it will work, since we require that the logged pids are also traced. I think the main issue is that I still don't see how the logs filtering works, how the additional rules affect certain apps. If you can point me to the code that does the logs criteria matching, I must be missing it. Perhaps I need to download your branch.

mmat11: It works similarly to how generictracer's filter works, I add the pids key (ns + pid) to a map and check via ebpf the current pid is in that map (see `pid_tracked()` in `logenricher.c`)

the userspace tracer is responsible to reconcile this map (`log_enricher_pids`) with the discovered processes based on config

grcevski: Sorry I wasn't clear, so imagine we have instrumented two ports for tracing, OTEL_EBPF_OPEN_PORT="8080,8081". I see you have LogEnricherServicesConfig, so let's say I have a config:
```
log_enricher:
  open_port: 8080
```

I'm wondering where is that extra discovery config used to avoid enrichment of logs for 8081?

mmat11: I think I'm using the main `discovery` pids for filtering 🤦🏻‍♂️ , I'll fix it

grcevski: Heh no problem, I just thought I was missing something. I think it looks great to be honest, with adding the correct filtering we are good to go IMO.

mmat11: I had to change it at "matcher" level and pass the info all along the pipeline, the changes are in the last commit

pkg/appolly/discover/matcher.go
```diff
@@ -105,18 +112,24 @@ func (m *Matcher) alreadyMatched(pid PID) bool {
...
+	}
+
 	if len(criteria) > 0 {
 		m.Log.Debug("found process", "pid", proc.Pid, "comm", proc.ExePath, "metadata",
-			obj.metadata, "podLabels", obj.podLabels, "criteria", criteria)
+			obj.metadata, "podLabels", obj.podLabels, "criteria", criteria, "logEnricherCriteria", logEnricherCriteria)
 
-		return &ProcessMatch{Criteria: criteria, Process: proc}
+		return &ProcessMatch{Criteria: criteria, LogEnricherCriteria: logEnricherCriteria, Process: proc}
 	}
```
grcevski, line 133: I would just ask one last thing :), if you don't mind. Before returning `nil` here can you add an `if` statement to check if the logEnricherCriteria is not empty and with info log tell us about it.

Something like:
```
if len(logEnricherCriteria) > 0 {
  m.Log.Info("avoiding log instrumentation for process, since it wasn't matched for instrumentation. Log enrichment criteria must be a subset of instrumentation matching criteria", "pid", proc.Pid...
}
```

mmat11: yes, good suggestion!

mmat11: done

PR #943, opened 2025-11-27
Create placeholder for metrics common configuration

mariomac:
In preparation:
- [x] Create a base metrics global configuration. This structure will be reused for per-service configuration.
- [x] Move `otel_metric_export > features` to a common `metrics > features`, and deprecate it
- [x] Move `prometheus > features` to `metrics > features`, and deprecate it
- [x] Handle new configuration from integration tests
- [x] Update documentation --> Will submit a pr to `opentelemetry.io` docs as soon as the new property is accepted/agreed.

The reason to move the features specification to a common section is because in the next PRs, we will allow a declarative per-service metric features definition. Having different metric features per export endpoint adds a lot of complexity to the code.


internal/test/integration/configs/obi-config-go-otel-grpc.yml
```diff
@@ -2,11 +2,12 @@ routes:
   unmatched: path
 prometheus_export:
   port: 8999
-  features:
-    - application
-    - application_process
 otel_traces_export:
   endpoint: http://jaeger:4318
+meter_provider:
+  obi_features:
```
grcevski: why do we call them `obi_features`, instead of `features`?

mariomac: With the new changes, this is already: `metrics > features`

grcevski (approved):
LGTM!

PR #954, opened 2025-12-01
better error log on socket filter fail error

NimrodAvni78:
This is more likely to happen on older kernel versions or specific distributions like bottlerocket, especially with BPF_DEBUG enabled
give a clear error message with actionable step to solve the issue

log:
```
time=2025-11-29T07:58:59.683+02:00 level=ERROR msg="couldn't trace process. Stopping process tracer" component=discover.traceAttacher error="error attaching socket filter: ran out of memory, socket filter has a jitted size of 6792, consider increasing the value net.core.optmem_max kernel parameter to be larger then the program jitted size, this will not affect existing sockets but only future ones created"

```

### DIFF
--- a/pkg/ebpf/instrumenter.go
+++ b/pkg/ebpf/instrumenter.go
@@ -342,7 +342,7 @@ func (i *instrumenter) handleSockFilterErr(originalErr error, filter *ebpf.Progr
 	if err != nil {
 		return fmt.Errorf("getting jited size: %w", originalErr)
 	}
-	return fmt.Errorf("%s, socket filter has a jited size of %d, consider increasing the value net.core.optmem_max kernel parameter to be larger then the program jitted size, this will not affect existing sockets but only future ones created", originalErr.Error(), jitedSize)
+	return fmt.Errorf("%s, socket filter has a jited size of %d, consider increasing the value net.core.optmem_max kernel parameter to be larger then the program jited size, this will not affect existing sockets but only future ones created", originalErr.Error(), jitedSize)
 }
 
 func attachSocketFilter(filter *ebpf.Program) (int, error) {


pkg/ebpf/instrumenter.go
```diff
@@ -330,6 +330,21 @@ func (i *instrumenter) sockfilters(p Tracer) error {
 	return nil
 }
 
+func (i *instrumenter) handleSockFilterErr(originalErr error, filter *ebpf.Program) error {
+	if !errors.Is(originalErr, unix.ENOMEM) {
+		return originalErr
+	}
+	info, err := filter.Info()
+	if err != nil {
+		return originalErr
```
mmat11: question: should we wrap the original error with the new one here?

pkg/ebpf/instrumenter.go
```diff
@@ -330,6 +330,21 @@ func (i *instrumenter) sockfilters(p Tracer) error {
...
+	}
+	jittedSize, err := info.JitedSize()
```
mmat11:
```suggestion
	jitedSize, err := info.JitedSize()
```

pkg/ebpf/instrumenter.go
```diff
@@ -330,6 +330,21 @@ func (i *instrumenter) sockfilters(p Tracer) error {
...
+	if err != nil {
+		return originalErr
```
mmat11: same here, should we wrap?

pkg/ebpf/instrumenter.go
```diff
@@ -330,6 +330,21 @@ func (i *instrumenter) sockfilters(p Tracer) error {
...
+	}
+	return fmt.Errorf("%s, socket filter has a jitted size of %d, consider increasing the value net.core.optmem_max kernel parameter to be larger then the program jitted size, this will not affect existing sockets but only future ones created", originalErr.Error(), jittedSize)
```
mmat11:
```suggestion
	return fmt.Errorf("%s, socket filter has a jited size of %d, consider increasing the value net.core.optmem_max kernel parameter to be larger than the program jited size, this will not affect existing sockets but only future ones created", originalErr.Error(), jitedSize)
```

pkg/ebpf/instrumenter.go
rafaelroquetto: +1

pkg/ebpf/instrumenter.go
rafaelroquetto: I believe `jitted` is also correct - see https://en.wiktionary.org/wiki/jitted#English - but the rules are flexible since this is an acronym turned into verb (and not a real verb).

pkg/ebpf/instrumenter.go
```diff
@@ -330,6 +330,21 @@ func (i *instrumenter) sockfilters(p Tracer) error {
...
+	}
+	info, err := filter.Info()
+	if err != nil {
+		return fmt.Errorf("getting program info: %w", originalErr)
+	}
+	jitedSize, err := info.JitedSize()
+	if err != nil {
+		return fmt.Errorf("getting jited size: %w", originalErr)
+	}
+	return fmt.Errorf("%s, socket filter has a jited size of %d, consider increasing the value net.core.optmem_max kernel parameter to be larger then the program jitted size, this will not affect existing sockets but only future ones created", originalErr.Error(), jitedSize)
```
mmat11: last nit:

```suggestion
	return fmt.Errorf("%s, socket filter has a jited size of %d, consider increasing the value net.core.optmem_max kernel parameter to be larger then the program jited size, this will not affect existing sockets but only future ones created", originalErr.Error(), jitedSize)
```

pkg/ebpf/instrumenter.go
mmat11: didn't knew that! I think `jited` is preferable for consistency with the ebpf library methods

PR #966, opened 2025-12-05
Add support for python asyncio

marctc:
This PR adds support to track context creation of python asyncio framework in order to add this information of trace context propagation. 

For this case it only uses basic asyncio with [aiohttp](https://docs.aiohttp.org/en/stable/index.html).

Thanks @aabmass for the guidance.
Paired with @grcevski

### DIFF
--- a/bpf/common/trace_common.h
+++ b/bpf/common/trace_common.h
@@ -398,11 +398,6 @@ static __always_inline void server_or_client_trace(
     if (type == EVENT_HTTP_REQUEST) {
         trace_key_t t_key = {0};
         task_tid(&t_key.p_key);
-        bpf_dbg_printk("server_or_client: host_id=%llx, after task_tid: pid=%d tid=%d ns=%x",
-                       id,
-                       t_key.p_key.pid,
-                       t_key.p_key.tid,
-                       t_key.p_key.ns);
         t_key.extra_id = extra_runtime_id();
 
         connection_info_part_t conn_part = {};
@@ -423,11 +418,8 @@ static __always_inline void server_or_client_trace(
             return;
         }
 
-        bpf_dbg_printk("STORE server_traces[pid=%d,tid=%d,ns=%x,extra_id=%llx]",
-                       t_key.p_key.pid,
-                       t_key.p_key.tid,
-                       t_key.p_key.ns,
-                       t_key.extra_id);
+        bpf_dbg_printk(
+            "Saving thread server span for ns=%x, extra_id=%llx", t_key.p_key.ns, t_key.extra_id);
         bpf_map_update_elem(&server_traces, &t_key, tp_p, BPF_ANY);
     } else {
         // Setup a pid, so that we can find it in TC.
--- a/bpf/generictracer/k_tracer.c
+++ b/bpf/generictracer/k_tracer.c
@@ -644,7 +644,6 @@ int BPF_KPROBE(obi_kprobe_tcp_close, struct sock *sk, long timeout) {
         bpf_map_delete_elem(&accepted_connections, &info.conn);
     }
 
-    bpf_map_delete_elem(&pid_tid_to_conn, &id);
     bpf_map_delete_elem(&active_send_args, &id);
     bpf_map_delete_elem(&active_send_sock_args, &sock_p);
 


bpf/generictracer/python.c
```diff
@@ -0,0 +1,68 @@
...
+
+    u64 context_ptr = (u64)context;
+
+    bpf_dbg_printk("=== uprobe Python PyContext_Run id=%d context=%llx ===", id, context);
+    bpf_map_update_elem(&current_python_context_map, &t_key, &context_ptr, BPF_ANY);
+    return 0;
+}
+
+SEC("uprobe/libpython3.so:copy_current_ret")
+int BPF_URETPROBE(obi_uprobe_copy_current_ret, void *context) {
```
grcevski: We can't use those, we must always use UPROBE because all of these dynamic languages can grow the stack. Since uprobes bind to the return address of the function, if the stack grows it will crash the application. The current "uretprobes" are actually uprobes, where we search for the "ret" opcodes in the function and attach a regular "uprobe" to those. If you couldn't attach uprobes to the return of copy_current_ret, then it must do a jump an exit through another function. We need to disassemble source of the python binary and look at what happens.

bpf/maps/python_async_context.h
```diff
@@ -0,0 +1,36 @@
...
+
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, trace_key_t); // the trace key of the thread running this context
+    __type(value, u64);       // the python context pointer
+    __uint(max_entries, MAX_CONCURRENT_REQUESTS);
+    __uint(pinning, OBI_PIN_INTERNAL);
+} current_python_context_map SEC(".maps");
+
+struct {
```
grcevski: What's the reason we have to use two maps for tracking the parent context?

What I mean is, when you have the probe on copy current, why not just remember the parent trace information for the context, which is just a single lookup when we match the parent information?

bpf/generictracer/python.c
```diff
@@ -0,0 +1,76 @@
...
+    trace_key_t t_key = {0};
+    task_tid(&t_key.p_key);
+    t_key.extra_id = extra_runtime_id();
+
+    u64 context_ptr = (u64)context;
+
+    bpf_dbg_printk("=== uprobe Python PyContext_Run id=%d context=%llx ===", id, context);
+    bpf_map_update_elem(&current_python_context_map, &t_key, &context_ptr, BPF_ANY);
+
+    // If this thread has an active server request, store the trace info in context_to_server_map
```
grcevski: I'm not sure how does this work. My understanding is that the context_run probe is for the child thread, why are we looking up server traces for the current trace_key (which is thread_id based) on this thread? I'm assuming we find nothing always.

How does this enable sequential await calls?

Doesn't every new thread get a new context from copy_current, which is presumably done on the original server request thread?

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -421,6 +421,16 @@ func (p *Tracer) UProbes() map[string]map[string][]*ebpfcommon.ProbeDesc {
 				Start:    p.bpfObjects.ObiUvFsAccess,
 			}},
 		},
+		"libpython3.": {
```
mmat11: question: is there a minimum supported python version? should we document it somewhere?

bpf/maps/python_async_context.h
marctc: You are right, I created those to test the sequential calls, but it can be merged in a single one.

bpf/generictracer/python.c
marctc: Python with asyncio doesn't create new threads for each await call. We need to store this information somehow in order to connect spans when the execution of the awaits is sequential. In the case we have nested calls, asyncio will create a new context.

pkg/internal/ebpf/generictracer/generictracer.go
marctc: python 3.4 is where asyncio was introduced. I can add a comment here for now and document in our context propagation docs that we have support for asyncio from version 3.4

internal/test/integration/components/pythonasync/async_context_test.py
```diff
@@ -0,0 +1,82 @@
...
+        response = await http_client.get(f"http://localhost:8391/mock/task-{task_id}")
+        return response.json()
+    
+    async def outer_task():
+        task1 = asyncio.create_task(inner_task(1))
+        task2 = asyncio.create_task(inner_task(2))
+        results = await asyncio.gather(task1, task2)
+        return results
+    
+    results = await outer_task()
```
grcevski: This example is interesting, we call async on outer_task and then two inner tasks. Based on this I think we need to make the lookup a bit more generic, that is it has to recursively look up to say 3 levels for parents. I'm not sure that this await call on outer_task() launches a new thread, but it might. 

What we have right now is only one level lookup. We have an example of multi parent lookup in:
```
static __always_inline tp_info_pid_t *find_parent_process_trace(trace_key_t *t_key) 
```

internal/test/integration/traces_test.go
```diff
@@ -1525,3 +1525,150 @@ func testHTTPTracesNestedNodeJSLargeHTTPS(t *testing.T) {
...
 	require.Len(t, children, 2)
 }
+
+func testPythonAsyncContextPropagation(t *testing.T) {
+	var traceID string
+	var parentID string
+
+	waitForTestComponentsSub(t, "http://localhost:8391", "/mock/health")
+
+	traceID = createTraceID()
```
grcevski: What do we use traceID and parentID for? Your configuration doesn't include `parse_request_headers: true`, nor has enabled context propagation, which means we will not read the incoming `traceparent` in the headers. 

You can either enable the parsing in the config or the docker compose, or just not use precreated traceID and parentID. Your validation logic doesn't check that the traceID is the same as the one you made in createParentID().

bpf/generictracer/python.c
grcevski: The above one I understand, but why do we lookup the server traces below if this is the client thread:
```
    // This enables sequential await calls to inherit the parent trace
    tp_info_pid_t *server_tp = bpf_map_lookup_elem(&server_traces, &t_key);
    if (server_tp) {
        bpf_map_update_elem(&context_to_server_map, &context_ptr, server_tp, BPF_NOEXIST);
    }
```

I guess my question is, if we have two sequential calls from the same parent, who calls the copy_current for the second request, the parent or the first client call in the sequence?

marctc: The context is created once when the endpoint handler starts or when `asyncio.create_task()`. As far I understand these sequential calls would run asynchronously in same thread and sharing context.

internal/test/integration/traces_test.go
marctc: Good catch, probs copy paste from other test. Remvoing

bpf/common/trace_common.h
```diff
@@ -200,6 +201,18 @@ static __always_inline tp_info_pid_t *find_parent_process_trace(trace_key_t *t_k
     return NULL;
 }
 
+static __always_inline tp_info_pid_t *find_python_parent_trace(trace_key_t *t_key) {
```
rafaelroquetto:
```suggestion
static __always_inline tp_info_pid_t *find_python_parent_trace(const trace_key_t *t_key) {
```

bpf/common/trace_common.h
```diff
@@ -200,6 +201,18 @@ static __always_inline tp_info_pid_t *find_parent_process_trace(trace_key_t *t_k
...
+    u64 *context_ptr = bpf_map_lookup_elem(&current_python_context_map, t_key);
+    if (context_ptr) {
+        u64 context = *context_ptr;
+        tp_info_pid_t *parent_trace = bpf_map_lookup_elem(&context_to_server_map, &context);
+        if (parent_trace) {
+            return parent_trace;
+        }
+    }
+    return NULL;
+}
```
rafaelroquetto:
```suggestion
    const u64 *context = bpf_map_lookup_elem(&current_python_context_map, t_key);
    
    if (!context) {
        return NULL;
    }
    
    return bpf_map_lookup_elem(&context_to_server_map, context);
}
```

rationale:

 1. https://medium.com/softwarecraft-mastery/the-return-early-pattern-c554f47fe58a
 2. you can use the `context` pointer directly, no need to do `context = *context_ptr` just to then pass it as `&context` to the next call.

bpf/common/trace_common.h
```diff
@@ -215,6 +228,11 @@ static __always_inline tp_info_pid_t *find_parent_trace(const pid_connection_inf
                    t_key->p_key.ns,
                    t_key->extra_id);
 
+    tp_info_pid_t *python_parent = find_python_parent_trace(t_key);
+    if (python_parent) {
```
rafaelroquetto, line 307: nit: add empty line before the if to remain consistent with surrounding code

bpf/generictracer/python.c
```diff
@@ -0,0 +1,74 @@
...
+#include <common/tp_info.h>
+#include <pid/pid.h>
+#include <maps/python_async_context.h>
+#include <maps/server_traces.h>
+
+SEC("uprobe/libpython3.so:context_run")
+int BPF_UPROBE(obi_uprobe_context_run, void *context) {
+    (void)ctx;
+
+    u64 id = bpf_get_current_pid_tgid();
```
rafaelroquetto, line 26:
```suggestion
    const u64 id = bpf_get_current_pid_tgid();
```

bpf/generictracer/python.c
grcevski: But if they are on the same thread, wouldn't they just get the same context from the parent setup from copy_current?

Can I please see logs (trimmed please) for what happens on the parent thread for the copy current, showing what context pointer did the parent create and for the client async requests?

What I'm trying to understand is why the code in copy current (shown below) is insufficient:
```
    void *context = (void *)PT_REGS_RC(ctx);
    // find the current parent trace key for this thread that creates a new context
    trace_key_t t_key = {0};
    task_tid(&t_key.p_key);
    t_key.extra_id = extra_runtime_id();

    tp_info_pid_t *parent_tp = find_python_parent_trace(&t_key);
    if (!parent_tp) {
        return 0;
    }

    bpf_dbg_printk(
        "=== uretprobe Python PyContext_CopyCurrent id=%d context=%llx ===", id, context);

    u64 context_ptr = (u64)context;
    bpf_map_update_elem(&context_to_server_map, &context_ptr, parent_tp, BPF_ANY);
    return 0;
```

The code in copy current sets up the parent context (parent_tp) for the context pointer that is copied. That context pointer is going to be then seen on the context_run for the client. I don't understand the part why do we look up in server traces in context_run for the client thread, the t_key is for the clients and it shouldn't ever be found in server_traces. 

I'm also not sure why do we need to look up `tp_info_pid_t *parent_tp = find_python_parent_trace(&t_key);` twice, once where it should be in trace_common, but you are also using it here? This code sets up the parent information from a server request, we would never find parent_trace for the incoming server request, which is the first one doing copy_current.

bpf/generictracer/python.c
```diff
@@ -0,0 +1,74 @@
...
+    if (!valid_pid(id)) {
+        return 0;
+    }
+
+    // find the trace key for this thread
+    trace_key_t t_key = {0};
+    task_tid(&t_key.p_key);
+    t_key.extra_id = extra_runtime_id();
+
+    u64 context_ptr = (u64)context;
```
rafaelroquetto: you may not need this - bpf_map_update_elem takes a `const void*` for the value, so I guess you can just pass context directly to it.

bpf/generictracer/python.c
```diff
@@ -0,0 +1,74 @@
...
+    tp_info_pid_t *server_tp = bpf_map_lookup_elem(&server_traces, &t_key);
+    if (server_tp) {
+        bpf_map_update_elem(&context_to_server_map, &context_ptr, server_tp, BPF_NOEXIST);
+    }
+
+    return 0;
+}
+
+SEC("uprobe/libpython3.so:copy_current_ret")
+int obi_uprobe_copy_current_ret(struct pt_regs *ctx) {
```
rafaelroquetto: I noticed that, unlike in the other uprobe above, you are not using the BPF_UPROBE macro - any particular reason? Otherwise, we should probably normalise it and use it here as well.

bpf/generictracer/python.c
```diff
@@ -0,0 +1,74 @@
...
+
+    if (!valid_pid(id)) {
+        return 0;
+    }
+
+    void *context = (void *)PT_REGS_RC(ctx);
+    // find the current parent trace key for this thread that creates a new context
+    trace_key_t t_key = {0};
+    task_tid(&t_key.p_key);
+    t_key.extra_id = extra_runtime_id();
```
rafaelroquetto: I've seen this above in the previous uprobe as well - perhaps it's better if this gets turned into a convenience function, something like:

```
static __always_inline trace_key_t make_trace_key() {
   trace_key_t t_key = {};
   task_tid(&t_key.p_key);
   t_key.extra_id = extra_runtime_id();
   return t_key;
}
```

then you can _reuse_ it as:

```
   const trace_key_t t_key = make_trace_key();
```

rationale:
 1. reuse
 2. readability
 3. const correctness

bpf/generictracer/python.c
```diff
@@ -0,0 +1,74 @@
...
+    tp_info_pid_t *parent_tp = find_python_parent_trace(&t_key);
+    if (!parent_tp) {
+        return 0;
+    }
+
+    bpf_dbg_printk(
+        "=== uretprobe Python PyContext_CopyCurrent id=%d context=%llx ===", id, context);
+
+    u64 context_ptr = (u64)context;
+    bpf_map_update_elem(&context_to_server_map, &context_ptr, parent_tp, BPF_ANY);
```
rafaelroquetto:
```suggestion
    bpf_map_update_elem(&context_to_server_map, &context, parent_tp, BPF_ANY);
```

grcevski:
@marctc, I think what would really help is some simple diagram showing what happens at what point and how do we track the context, or an explanation step by step. The reason that I'm confused is that the original logs shown by the bpftrace program that Aaron wrote, things are not the same here in this implementation.

What I understood from those logs is that:
1. CopyCurrent runs on the server (parent) thread and sets up a copy of the context for the async task. This is the place where we would search in server_traces to pull the current trace information for the async tasks. 
2. run_context happens on all async tasks, server or client, but the ones we care about is the client tasks. The context found in run_context on the client task should match the context setup in 1. by copy current. So if we recorded information in the parent task for the context pointer, we should be able to find the trace parent tp_info_t for the client when we end up in trace_common looking up the context.

bpf/maps/python_async_context.h
```diff
@@ -0,0 +1,28 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
```
rafaelroquetto: we've been leaning towards one map per header file, allowing them to be individually included (and thus hinting to the header which maps a given translation unit is using).

But there's no hard rule, not sure how others feel about this.

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -417,6 +417,20 @@ func (p *Tracer) UProbes() map[string]map[string][]*ebpfcommon.ProbeDesc {
...
+		"libpython3.": { // python asyncio (from python 3.4)
+			"context_run": {{
+				Required: false,
+				Start:    p.bpfObjects.ObiUprobeContextRun,
+			}},
+			"PyContext_CopyCurrent": {{
+				Required: false,
+				End:      p.bpfObjects.ObiUprobeContextNewFromVarsRet,
+			}},
+			"context_new_from_vars": {{
```
grcevski: Why was this API needed? Does CopyCurrent call context_new_from_vars and the compiler optimises the return?

internal/test/integration/traces_test.go
```diff
@@ -1525,3 +1525,145 @@ func testHTTPTracesNestedNodeJSLargeHTTPS(t *testing.T) {
 	// We must see two children
 	require.Len(t, children, 2)
 }
+
+func testPythonAsyncSequential(t *testing.T) {
+	waitForTestComponentsSub(t, "http://localhost:8391", "/health")
+
+	ti.DoHTTPGet(t, "http://localhost:8391/sequential", 200)
```
grcevski: Can you modify these tests to run simultaneous requests in parallel, like `testHTTPTracesNestedNginxSQL` or  `testHTTPTracesNestedNginx` in red_test_ruby.go. This will ensure that the context matching truly works even when there are multiple requests in play.

Same for `testPythonAsyncToThread`.

internal/test/integration/components/pythonasync/async_context_test.py
```diff
@@ -0,0 +1,49 @@
...
+    def blocking_http_call(url: str):
+        response = requests.get(url, timeout=30)
+        return response.status_code
+
+    r1 = await asyncio.to_thread(blocking_http_call, f"{BACKEND_URL}/")
+    r2 = await asyncio.to_thread(blocking_http_call, f"{BACKEND_URL}/")
+    return {"calls": 2, "status_codes": [r1, r2]}
+
+
+if __name__ == "__main__":
```
fstab, line 83: Both endpoints (`/sequential` and `/to-thread`) run client requests sequentially: The `await` suspends the main thread until the client call is done, only then the next client call is scheduled.

It might be good to have a test that actually runs multiple requests in parallel.

pkg/internal/ebpf/generictracer/generictracer.go
marctc: Python in docker does tail call optimization:

```
000000000019fbf9 <PyContext_CopyCurrent>:
  19fbf9:	endbr64
  19fbfd:	push   %rbp
  19fbfe:	mov    %rsp,%rbp
  19fc01:	call   2f9ee0 <context_get>
  19fc06:	test   %rax,%rax
  19fc09:	je     19fc15 <PyContext_CopyCurrent+0x1c>
  19fc0b:	mov    0x18(%rax),%rdi
  19fc0f:	pop    %rbp
  19fc10:	jmp    19fb83 <context_new_from_vars>
```

where as my local python is not optimized:

```
00000000002aadc0 <PyContext_CopyCurrent>:
  2aadc0:	f3 0f 1e fa          	endbr64
  2aadc4:	41 54                	push   %r12
  2aadc6:	55                   	push   %rbp
  2aadc7:	53                   	push   %rbx
  2aadc8:	66 48 8d 3d 20 21 2b 	data16 lea 0x2b2120(%rip),%rdi        # 55cef0 <_DYNAMIC+0x8d0>
  2aadcf:	00 
  2aadd0:	66 66 48 e8 28 d8 e5 	data16 data16 rex.W call 108600 <__tls_get_addr@plt>
  2aadd7:	ff 
  2aadd8:	48 8b 28             	mov    (%rax),%rbp
  2aaddb:	48 89 c3             	mov    %rax,%rbx
  2aadde:	48 8b 85 d0 00 00 00 	mov    0xd0(%rbp),%rax
  2aade5:	48 85 c0             	test   %rax,%rax
  2aade8:	0f 84 e2 00 00 00    	je     2aaed0 <PyContext_CopyCurrent+0x110>
  2aadee:	4c 8b 60 18          	mov    0x18(%rax),%r12
  2aadf2:	48 8b 45 10          	mov    0x10(%rbp),%rax
  2aadf6:	8b 90 f0 2b 04 00    	mov    0x42bf0(%rax),%edx
  2aadfc:	85 d2                	test   %edx,%edx
  2aadfe:	0f 85 94 00 00 00    	jne    2aae98 <PyContext_CopyCurrent+0xd8>
  2aae04:	48 8b 3d cd 1c 2b 00 	mov    0x2b1ccd(%rip),%rdi        # 55cad8 <PyContext_Type@@Base-0x240c8>
  2aae0b:	e8 60 da e5 ff       	call   108870 <_PyObject_GC_New@plt>
  2aae10:	48 89 c3             	mov    %rax,%rbx
  2aae13:	48 85 c0             	test   %rax,%rax
  2aae16:	0f 84 d4 00 00 00    	je     2aaef0 <PyContext_CopyCurrent+0x130>
  2aae1c:	41 8b 04 24          	mov    (%r12),%eax
  2aae20:	48 c7 43 10 00 00 00 	movq   $0x0,0x10(%rbx)
  2aae27:	00 
  2aae28:	c7 43 28 00 00 00 00 	movl   $0x0,0x28(%rbx)
  2aae2f:	48 c7 43 20 00 00 00 	movq   $0x0,0x20(%rbx)
  2aae36:	00 
  2aae37:	83 c0 01             	add    $0x1,%eax
  2aae3a:	74 04                	je     2aae40 <PyContext_CopyCurrent+0x80>
  2aae3c:	41 89 04 24          	mov    %eax,(%r12)
  2aae40:	4c 89 63 18          	mov    %r12,0x18(%rbx)
  2aae44:	66 48 8d 3d a4 20 2b 	data16 lea 0x2b20a4(%rip),%rdi        # 55cef0 <_DYNAMIC+0x8d0>
  2aae4b:	00 
  2aae4c:	66 66 48 e8 ac d7 e5 	data16 data16 rex.W call 108600 <__tls_get_addr@plt>
  2aae53:	ff 
  2aae54:	48 8d 4b f0          	lea    -0x10(%rbx),%rcx
  2aae58:	48 8b 00             	mov    (%rax),%rax
  2aae5b:	48 8b 40 10          	mov    0x10(%rax),%rax
  2aae5f:	48 8b 90 d0 00 00 00 	mov    0xd0(%rax),%rdx
  2aae66:	48 8b 43 f8          	mov    -0x8(%rbx),%rax
  2aae6a:	48 8b 72 08          	mov    0x8(%rdx),%rsi
  2aae6e:	83 e0 03             	and    $0x3,%eax
  2aae71:	66 48 0f 6e c2       	movq   %rdx,%xmm0
  2aae76:	48 09 f0             	or     %rsi,%rax
  2aae79:	48 89 0e             	mov    %rcx,(%rsi)
  2aae7c:	66 48 0f 6e c8       	movq   %rax,%xmm1
  2aae81:	66 0f 6c c1          	punpcklqdq %xmm1,%xmm0
  2aae85:	0f 11 43 f0          	movups %xmm0,-0x10(%rbx)
  2aae89:	48 89 4a 08          	mov    %rcx,0x8(%rdx)
  2aae8d:	48 89 d8             	mov    %rbx,%rax
  2aae90:	5b                   	pop    %rbx
  2aae91:	5d                   	pop    %rbp
  2aae92:	41 5c                	pop    %r12
  2aae94:	c3                   	ret
  2aae95:	0f 1f 00             	nopl   (%rax)
  2aae98:	48 8b 98 e8 2b 04 00 	mov    0x42be8(%rax),%rbx
  2aae9f:	83 ea 01             	sub    $0x1,%edx
  2aaea2:	89 90 f0 2b 04 00    	mov    %edx,0x42bf0(%rax)
  2aaea8:	48 8b 53 20          	mov    0x20(%rbx),%rdx
  2aaeac:	48 89 df             	mov    %rbx,%rdi
  2aaeaf:	48 89 90 e8 2b 04 00 	mov    %rdx,0x42be8(%rax)
  2aaeb6:	48 c7 43 20 00 00 00 	movq   $0x0,0x20(%rbx)
  2aaebd:	00 
  2aaebe:	e8 1d fa e5 ff       	call   10a8e0 <_Py_NewReference@plt>
  2aaec3:	e9 54 ff ff ff       	jmp    2aae1c <PyContext_CopyCurrent+0x5c>
  2aaec8:	0f 1f 84 00 00 00 00 	nopl   0x0(%rax,%rax,1)
  2aaecf:	00 
  2aaed0:	e8 3b ec ff ff       	call   2a9b10 <context_new_empty>
  2aaed5:	48 85 c0             	test   %rax,%rax
  2aaed8:	74 16                	je     2aaef0 <PyContext_CopyCurrent+0x130>
  2aaeda:	48 89 85 d0 00 00 00 	mov    %rax,0xd0(%rbp)
  2aaee1:	48 8b 2b             	mov    (%rbx),%rbp
  2aaee4:	e9 05 ff ff ff       	jmp    2aadee <PyContext_CopyCurrent+0x2e>
  2aaee9:	0f 1f 80 00 00 00 00 	nopl   0x0(%rax)
  2aaef0:	31 db                	xor    %ebx,%ebx
  2aaef2:	eb 99                	jmp    2aae8d <PyContext_CopyCurrent+0xcd>
  2aaef4:	66 66 2e 0f 1f 84 00 	data16 cs nopw 0x0(%rax,%rax,1)
  2aaefb:	00 00 00 00 
  2aaeff:	90                   	nop
```

grcevski: Great, thanks for explaining that.

bpf/common/runtime.h
```diff
@@ -18,6 +19,11 @@ static __always_inline u64 extra_runtime_id_with_task_id(const u64 id) {
 
 static __always_inline u64 extra_runtime_id() {
     const u64 id = bpf_get_current_pid_tgid();
-
+    u64 *context = bpf_map_lookup_elem(&python_current_context, &id);
+    if (context) {
+        bpf_dbg_printk(
+            "extra_runtime_id: LOOKUP python_current_context[host_id=%llx] = %llx", id, *context);
+        return (u64)(*context);
```
rafaelroquetto, line 26:
```suggestion
        return *context;
```

bpf/common/trace_common.h
```diff
@@ -31,6 +31,8 @@
 #include <maps/server_traces.h>
 #include <maps/tp_info_mem.h>
 #include <maps/tp_char_buf_mem.h>
+#include <maps/python_thread_context.h>
+#include <maps/python_context_trace.h>
```
rafaelroquetto, line 35: nit: includes are ordered alphabetically (at least in a group)

bpf/common/runtime.h
```diff
@@ -18,6 +19,11 @@ static __always_inline u64 extra_runtime_id_with_task_id(const u64 id) {
 
 static __always_inline u64 extra_runtime_id() {
     const u64 id = bpf_get_current_pid_tgid();
-
+    u64 *context = bpf_map_lookup_elem(&python_current_context, &id);
```
rafaelroquetto, line 22: I reckon this is only relevant for Python, so I'd move this into `make_trace_key` in `python.c` and leave this alone.

Rationale:
 - improves readability by making it explicit these are orthogonal
 - improves performance by not wasting time on map lookups when extra_runtime_id is called for non-python processes
 - even though it should be impossible in practice, it would conceptually rule out a non python process from interacting with `python_current_context`

bpf/generictracer/python.c
```diff
@@ -0,0 +1,61 @@
...
+
+#include <pid/pid.h>
+
+static __always_inline trace_key_t make_trace_key(void *context) {
+    trace_key_t t_key = {};
+    task_tid(&t_key.p_key);
+    if (context) {
+        t_key.extra_id = (u64)context;
+    } else {
+        t_key.extra_id = extra_runtime_id();
```
rafaelroquetto, line 19: so with regards to my comments above, here you can lookup `python_current_context` explicitly, and then fall back to `extra_runtime_id()` if it is null

bpf/generictracer/python.c
```diff
@@ -0,0 +1,61 @@
...
+//go:build obi_bpf_ignore
+
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_helpers.h>
+
+#include <common/trace_common.h>
+
+#include <pid/pid.h>
+
+static __always_inline trace_key_t make_trace_key(void *context) {
```
rafaelroquetto, line 13:
```suggestion
static __always_inline trace_key_t make_trace_key(const void *context) {
```

bpf/generictracer/python.c
rafaelroquetto: Looks like it is not resolved

bpf/generictracer/python.c
```diff
@@ -0,0 +1,61 @@
...
+    if (!valid_pid(id)) {
+        return 0;
+    }
+
+    void *context = (void *)PT_REGS_PARM1(ctx);
+    const trace_key_t t_key = make_trace_key(context);
+
+    bpf_dbg_printk("context_run: tid=%d ctx=%llx", t_key.p_key.tid, context);
+    bpf_map_update_elem(&python_thread_context, &t_key, &context, BPF_ANY);
+    bpf_map_update_elem(&python_current_context, &id, &context, BPF_ANY);
```
rafaelroquetto, line 37: do we really need 2 maps here? `id` is basically tid_tgid (so unique) mapped to a context pointer.

But then t_key is basically the same information (the thread/task id encoded differently) mapped to a context (and more confusingly the `extra_id` is the context itself) - so basically, to look up the value in this map (the context) you need a key that contains the context -> so logically you don't need this map because you already know the context.

I hope that makes sense, perhaps I am misunderstanding it.

bpf/generictracer/python.c
```diff
@@ -0,0 +1,61 @@
...
+    return 0;
+}
+
+SEC("uprobe/libpython3.so:context_new_from_vars_ret")
+int obi_uprobe_context_new_from_vars_ret(struct pt_regs *ctx) {
+    u64 id = bpf_get_current_pid_tgid();
```
rafaelroquetto, line 43:
```suggestion
    const u64 id = bpf_get_current_pid_tgid();
```

bpf/generictracer/python.c
```diff
@@ -0,0 +1,61 @@
...
+        return 0;
+    }
+
+    void *context = (void *)PT_REGS_RC(ctx);
+
+    if (!context) {
+        return 0;
+    }
+
+    const trace_key_t t_key = make_trace_key(0);
```
rafaelroquetto, line 55: If I understand this correctly, this is where you associate this context with its parent?

If so, couldn't you do something like:
 - bpf_map_lookup_elem(&python_thread_context, id, ...) to lookup the context for this thread? I feel this is exactly what the current implementation ends up doing once it calls `extra_runtime_id`, just in a more convoluted fashion.

bpf/common/trace_common.h
```diff
@@ -220,6 +222,41 @@ static __always_inline tp_info_pid_t *find_parent_process_trace(trace_key_t *t_k
...
+        tp_info_pid_t *server_tp = bpf_map_lookup_elem(&server_traces, &key);
+
+        if (server_tp) {
+            bpf_dbg_printk(
+                "find_python_parent: FOUND tid=%d extra=%llx", key.p_key.tid, key.extra_id);
+            return server_tp;
+        }
+
+        // Not this thread's server request, try Python context chain
+        u64 *context_ptr = bpf_map_lookup_elem(&python_thread_context, &key);
```
rafaelroquetto, line 240: ok so here instead of `key`, I believe we could just pass `id` (pid_tgid)

rafaelroquetto (changes requested):
Please correct me if I am wrong, but my understanding is that you try to build a parent/child relationship between python threads.

If that's the case, I _think_ it may be possible to rely only on `bpf_get_pid_tgid()` aka `id`.

So in `context_new_from_vars_ret` (I am assuming this is where a new context is created in the parent thread) you can associated the new context with the parent, i.e.

```
map 1 <ctx_ptr, id (of the parent)>
```

then when the context runs (in `context_run`) this is where you map the context pointer to its own thread id, so:

```
map 2 <id, ctx_ptr>
```

then to find the parent trace:
1. get the current thread id via `bpf_get_pid_tgid()` (aka `id`)
2. build t_key and query server_traces
3. if nothing yields, then go one level deep: get the `ctx_ptr` using `id` (map 1), then use this `ctx_ptr` to get the `id` of the parent (map 2). Then goto 2.

Hopefully I got it right.

PR #968, opened 2025-12-07
Support for distributed tracing for Ruby on Rails

grcevski:
This design description was also added to the new `ruby.c` file.

PR to track the worker thread handoff for Puma used in Ruby on Rails and other Ruby based projects.
Puma architecture document: https://github.com/puma/puma/blob/main/docs/architecture.md

Basic concepts:
1. Server thread, which does all the accept(s) of the incoming requests, named `puma srv`.
2. Worker threads, that do all the request handling, named `puma srv tp NNN`
3. Reactor thread, which takes over the reading of the incoming request if there are no worker
   threads available.
4. Other worker threads, named `puma srv thread NNN`.

The `puma srv` thread adds the work when it's accepted, by sending a "client" object to
the "worker pool", called "pool" in the server.rb code: 
https://github.com/puma/puma/blob/731b97d2a5c7838c9d736462e259e629655b00a1/lib/puma/server.rb#L406

This is effectively an array push in ruby's world, handled by `rb_ary_push`.

The worker threads (`puma srv tp NNN`) wait around for work to arrive, and when there's work ready, they take
the head of the work queue setup by the `puma srv` thread, by calling "shift" on the array, i.e. `rb_ary_shift`.
This is the call to `todo.shft` here: 
https://github.com/puma/puma/blob/731b97d2a5c7838c9d736462e259e629655b00a1/lib/puma/thread_pool.rb#L178

We can see that the `<<` operator that `server.rb` calls, is essentially adding to the todo queue:
https://github.com/puma/puma/blob/731b97d2a5c7838c9d736462e259e629655b00a1/lib/puma/thread_pool.rb#L261

The approach we take to handle this is the follows:
1. Uprobe on `rb_obj_call_init_kw`, where we match the name of the server thread. Initially, a probe on
   `rb_ary_push` was considered, but this proved to be very expensive. Instead we look for calls to
   object initializations by this thread. This thread may be initializing random objects, but we don't care, 
   since the LRU map will flush the unwanted items. The "puma srv" thread is a light allocator.
2. At the time the "push" probe runs, we looks to see if we have recorded info in the accept4 kprobe. If we have,
   then we record metadata so that the worker thread can find it. What we record is the item pointer. 
   Since Ruby GC (so far) doesn't move objects, this should work 100% of the
   time. Even if they decide to move and compact the objects, the chances of this new object from moving
   is very small. If the references ever move, we'll fail to correlate the server and client call in
   that instance.
3. Uprobe on `rb_ary_shift`, where we match for the "puma srv tp" worker threads and we read the head
   of the todo array. We handle two cases, Ruby embedded arrays and Ruby heap arrays. Ruby frozen arrays
   don't matter, since we would be throwing an exception on shift if this array was frozen.
4. At the time of the "shift" probe, we take the item pointer and record it for the 
   current thread. 
5. find_parent_trace in trace_common.h, looks up the puma metadata for the current pid:tgid pair. It
   would find the metadata recorded by the `rb_ary_shft` uprobe, i.e. the "puma srv tp NNN" worker, which
   is the client thread. We take the returned "array:item" pair from the client and look up in the map
   setup by "puma srv" thread on `rb_array_push` for the connection information of the original server
   request that was accepted. We then use the connection information to hit `server_traces_aux` for the
   traceparent information, much like we do for "nginx".

In a sense, this design is very similar to what happens with nginx request tracking, except here we use
the array:item pair for the work, rather than the file descriptors.

TODO:
- [x] Integration tests with many parallel requests to overflow the default rails thread pool and make it stash work on the todo.

### DIFF
--- a/bpf/generictracer/ruby.c
+++ b/bpf/generictracer/ruby.c
@@ -143,7 +143,7 @@ int obi_rb_ary_shift(struct pt_regs *ctx) {
     // See if the array is not embedded anymore and we have to
     // find the heap pointer. If the 0x2000 flag is set, the array
     // is embedded.
-    const u8 embedded_arr = flags & 0x2000;
+    const u16 embedded_arr = flags & 0x2000;
 
     u64 item = 0;
     bpf_probe_read_user(&item, sizeof(u64), (void *)(item_ptr + k_rb_ary_embedded_ptr_pos));


internal/test/integration/components/nginx-rails-sql/docker-compose.yaml
```diff
@@ -0,0 +1,56 @@
+version: '3.8'
```
skl: I believe the version property can now be dropped from compose files

internal/test/integration/components/nginx-rails-sql/testapi/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+FROM ruby:3.0.2-slim
+
+RUN apt-get update && apt-get install -y default-libmysqlclient-dev
+
+WORKDIR /
+COPY --from=builder /build .
+COPY --from=builder /usr/local/bundle /usr/local/bundle
+
+# Run the node app
```
skl:
```suggestion
# Run the app
```

grcevski: Good catch of my copy pasta :).

internal/test/integration/components/nginx-rails-sql/docker-compose.yaml
grcevski: Sure!

bpf/generictracer/ruby.c
```diff
@@ -0,0 +1,190 @@
...
+#include <generictracer/maps/pid_tid_to_conn.h>
+#include <generictracer/maps/puma_tasks.h>
+
+#include <generictracer/types/puma_task_id.h>
+
+#include <logger/bpf_dbg.h>
+
+#include <pid/pid.h>
+
+enum { k_comm_len = 12 };
```
mmat11, line 22: shouldn't this be 16?

mmat11: or perhaps this is the minimum len to be able to compare? in that case disregard my comment

bpf/generictracer/ruby.c
```diff
@@ -0,0 +1,190 @@
...
+
+        bpf_map_update_elem(&puma_task_connections, &task_id, &conn_part, BPF_ANY);
+    }
+
+    return 0;
+}
+
+SEC("uprobe/ruby:rb_ary_shift")
+int obi_rb_ary_shift(struct pt_regs *ctx) {
+    (void)ctx;
```
mmat11:
```suggestion
```

this should be used below

bpf/generictracer/ruby.c
```diff
@@ -0,0 +1,190 @@
...
+    if (!valid_pid(id)) {
+        return 0;
+    }
+
+    u64 ary = (u64)PT_REGS_PARM1(ctx);
+    if (!ary) {
+        return 0;
+    }
+
+    u64 item_ptr = (u64)PT_REGS_PARM1(ctx);
```
mmat11: isn't this the same as `ary`?

bpf/generictracer/ruby.c
```diff
@@ -0,0 +1,190 @@
...
+
+    // See if the array is not embedded anymore and we have to
+    // find the heap pointer.
+    if ((flags & 0x2000) == 0) {
+        bpf_dbg_printk("rb_ary_shift ret ==> empty slice");
+        embedded_arr = 0;
+    }
+
+    u64 item = 0;
+    bpf_probe_read_user(&item, sizeof(u64), (void *)(item_ptr + 0x10));
```
mmat11: nit: can we define these offsets as enum/constants?

internal/test/integration/red_test_ruby.go
```diff
@@ -167,3 +167,51 @@ func testHTTPTracesNestedNginx(t *testing.T) {
 		}, test.Interval(100*time.Millisecond))
 	}
 }
+
+// Assumes we've run the metrics tests
+func testHTTPTracesNestedNginxSQL(t *testing.T) {
+	for i := 1; i <= 4; i++ {
+		go ti.DoHTTPGet(t, "https://localhost:8443/users/"+strconv.Itoa(i), 200)
```
mmat11, line 174: do these need to be ran async?

bpf/common/trace_common.h
```diff
@@ -3,6 +3,10 @@
 
 #pragma once
 
+#include "common/connection_info.h"
+#include "common/tp_info.h"
+#include "generictracer/maps/puma_tasks.h"
+#include "logger/bpf_dbg.h"
```
rafaelroquetto: This looks like vscode auto added the headers.

We need to replace the `"` with `<>` and group them accordingly with the rest of the headers down below.

bpf/common/trace_common.h
```diff
@@ -137,6 +143,24 @@ static __always_inline tp_info_pid_t *find_nginx_parent_trace(const pid_connecti
     return NULL;
 }
 
+static __always_inline tp_info_pid_t *find_puma_parent_trace(u64 id) {
+    puma_task_id_t *task_id = bpf_map_lookup_elem(&puma_worker_tasks, &id);
+    bpf_dbg_printk("puma lookup task_id %llx", task_id);
+    if (!task_id) {
+        return 0;
```
rafaelroquetto:
```suggestion
        return NULL;
```

(just a nit)

bpf/generictracer/ruby.c
```diff
@@ -0,0 +1,190 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build obi_bpf_ignore
+
+#include "common/connection_info.h"
```
rafaelroquetto: same here

bpf/generictracer/ruby.c
rafaelroquetto: +1

bpf/generictracer/ruby.c
```diff
@@ -0,0 +1,190 @@
...
+
+    u64 flags = 0;
+    bpf_probe_read_user(&flags, sizeof(u64), (void *)(item_ptr));
+
+    u8 embedded_arr = 1;
+
+    // See if the array is not embedded anymore and we have to
+    // find the heap pointer.
+    if ((flags & 0x2000) == 0) {
+        bpf_dbg_printk("rb_ary_shift ret ==> empty slice");
```
rafaelroquetto: is this debug message correct? I'd expect something like "array not embedded" or something like that - but I may be missing some context.

bpf/generictracer/ruby.c
```diff
@@ -0,0 +1,190 @@
...
+    if (!ary) {
+        return 0;
+    }
+
+    u64 item_ptr = (u64)PT_REGS_PARM1(ctx);
+
+    u64 flags = 0;
+    bpf_probe_read_user(&flags, sizeof(u64), (void *)(item_ptr));
+
+    u8 embedded_arr = 1;
```
rafaelroquetto: if you don't need the debug print below, you could shorten this (at your discretion) to

```
const u8 embedded_array = flags & 0x2000;
```

bpf/generictracer/ruby.c
grcevski: yeah, I only needed 12 characters to compare the names of the threads that do the work.

bpf/generictracer/ruby.c
grcevski: good catch, left over code when I initially wrote this on the return probe.

internal/test/integration/red_test_ruby.go
grcevski: yeah, if we wait one by one it's always using the same thread, I needed to ensure these run in parallel to expose the problem of multiple queuing requests.

bpf/generictracer/ruby.c
grcevski: right, I'll fix that.

bpf/generictracer/ruby.c
grcevski: yeah good suggestion, I'll print the flags and leave a comment saying 0x2000 is heap pointer.

bpf/generictracer/ruby.c
grcevski: I went with removing the print and printing the flags instead.

bpf/generictracer/ruby.c
grcevski: done

PR #984, opened 2025-12-10
Add SQL database hostname capture for MySQL

marctc:
Current implementation only captures IP of database connections, which ultimately are set as outgoing.
This PR adds extra logic to capture the hostname of  SQL operations to MySQL.

The implementation add hooks for https://github.com/go-sql-driver/mysql, retrieving config info of https://github.com/go-sql-driver/mysql/blob/master/dsn.go to finally fetch hostname

### DIFF
--- a/pkg/transform/name_resolver.go
+++ b/pkg/transform/name_resolver.go
@@ -223,12 +223,12 @@ func (nr *NameResolver) resolve(svc *svc.Attrs, ip string, fallback string) (str
 	if len(ip) > 0 {
 		var peer string
 		peer, ns = nr.dnsResolve(svc, ip)
+		name = ip
+		if fallback != "" {
+			name = fallback
+		}
 		if len(peer) > 0 && peer != ip {
 			name = peer
-		} else if fallback != "" {
-			name = fallback
-		} else {
-			name = ip
 		}
 	} else if fallback != "" {
 		name = fallback


bpf/gotracer/go_sql.c
```diff
@@ -26,10 +26,200 @@
...
+    // Config.Addr is a Go string represented as [data_ptr, length]
+    u64 addr_ptr = 0, addr_len = 0;
+    bpf_probe_read(&addr_ptr, sizeof(addr_ptr), (void *)((u64)cfg_ptr + addr_offset));
+    bpf_probe_read(&addr_len, sizeof(addr_len), (void *)((u64)cfg_ptr + addr_offset + 8));
+
+    if (addr_ptr == 0 || addr_len == 0) {
+        return 0;
+    }
+
+    if (addr_len > max_len - 1) {
```
grcevski: This code here should be replaced with a call to `read_go_str`. it does the exact thing, checking the length, putting a null terminator at the length etc.

bpf/gotracer/go_sql.c
```diff
@@ -26,10 +26,200 @@
...
+        addr_len = max_len - 1;
+    }
+
+    bpf_probe_read(hostname, addr_len, (void *)addr_ptr);
+    if (addr_len < max_len) {
+        hostname[addr_len] = '\0';
+    }
+
+    // Basic validation: check first few chars are printable
+#pragma unroll
```
grcevski: Why do we want to do this validation here? Do we expect the value to be corrupt? If you have to do validation, it's better done in userspace when we create the event.

bpf/gotracer/go_sql.c
```diff
@@ -26,10 +26,200 @@
...
+        return 0;
+    }
+
+    off_table_t *ot = get_offsets_table();
+    if (!ot) {
+        return 0;
+    }
+
+    // Read stdlib Conn.conn field offset
+    u64 pgx_conn_offset = go_offset_of(ot, (go_offset){.v = _pgstdlib_conn_pos});
```
grcevski: Where are these -1 offsets set to -1? Is it supposed to be done for this code to work?

bpf/gotracer/go_sql.c
```diff
@@ -26,10 +26,200 @@
...
+    // Config.Host is a Go string [data_ptr, length]
+    u64 host_ptr = 0, host_len = 0;
+    bpf_probe_read(&host_ptr, sizeof(host_ptr), (void *)((u64)config_ptr + host_offset));
+    bpf_probe_read(&host_len, sizeof(host_len), (void *)((u64)config_ptr + host_offset + 8));
+
+    if (host_ptr == 0 || host_len == 0) {
+        return 0;
+    }
+
+    if (host_len > max_len - 1) {
```
grcevski: Same here, use go_read_str and don't do validation in eBPF.

bpf/gotracer/go_sql.c
```diff
@@ -26,10 +26,200 @@
...
+        return 0;
+    }
+
+    off_table_t *ot = get_offsets_table();
+    if (!ot) {
+        return 0;
+    }
+
+    // Get offset of driverConn.ci field
+    u64 ci_offset = go_offset_of(ot, (go_offset){.v = _driverconn_ci_pos});
```
grcevski, line 39: There are so many exits from each of these if checks with offset -1, which I think are never set. It would be nice to have a debug print saying why we bailed if we would ever have to debug why this code doesn't work.

bpf/gotracer/go_sql.c
```diff
@@ -110,6 +309,42 @@ int obi_uprobe_queryReturn(struct pt_regs *ctx) {
 
         __builtin_memcpy(&trace->conn, &invocation->conn, sizeof(connection_info_t));
 
+        u64 driver_conn_ptr = invocation->driver_conn_ptr;
+        int hostname_extracted = 0;
```
grcevski: I think we can simply this by always initialising the hostname to '\0'. Maybe extract this code into a function  and return from error scenarions, but also put print statement so we know if we need to debug this.

configs/offsets/std_inspect.go
```diff
@@ -69,7 +70,23 @@ func regularGetRequest(ctx context.Context, url string) error {
 	return nil
 }
 
+func useDatabaseSQL() error {
```
grcevski, line 73: What about progams in config/offsets for the mysql and postgres drivers? I think we need those so that we can generate the offsets in tracker_input

internal/test/oats/sql/yaml/oats_sql_go_postgres.yaml
```diff
@@ -16,11 +16,13 @@ expected:
             db.collection.name: accounting.contacts
             db.system.name: other_sql
   metrics:
-    - promql: 'db_client_operation_duration_sum{db_operation_name="SELECT", db_system_name="other_sql"}'
+    - promql: 'db_client_operation_duration_sum{db_operation_name="SELECT", db_system_name="other_sql", server_address="sqlserver"}'
```
grcevski: This is great test, but I think we need to add one for mysql and the alternative postgres library you added.

configs/offsets/tracker_input.json
```diff
@@ -23,7 +23,8 @@
       "net/http.http2serverConn": ["conn"],
       "net/http.http2Framer": ["w"],
       "net/http.http2MetaHeadersFrame": ["Fields"],
-      "net/rpc.Request": ["ServiceMethod"]
+      "net/rpc.Request": ["ServiceMethod"],
```
grcevski, line 26: We need all of the postgres and mysql library offsets in the tracker_input as well. We can't expect that they'll be there in the binary.

pkg/internal/goexec/structmembers.go
```diff
@@ -414,6 +425,48 @@ var structMembers = map[string]structInfo{
...
+			"cfg": MySQLConnCfgPos,
+		},
+	},
+	"github.com/go-sql-driver/mysql.Config": {
+		lib: "github.com/go-sql-driver/mysql",
+		fields: map[string]GoOffset{
+			"Addr": MySQLConfigAddrPos,
+		},
+	},
+	"github.com/jackc/pgx/v5/pgconn.PgConn": {
```
grcevski: The v5 package seems quite new, would be possible to support v4 as well, how different are they?

internal/test/oats/sql/yaml/oats_sql_go_postgres.yaml
marctc: yeah, i'm working on it

bpf/common/common.h
```diff
@@ -35,6 +35,7 @@
 #define HOST_LEN 64 // can be a fully qualified DNS name
 #define TRACEPARENT_LEN 55
 #define SQL_MAX_LEN 500
+#define SQL_HOSTNAME_MAX_LEN 272
```
grcevski: Is there a reason we picked such a big buffer for the name of the host? I would imagine most hostnames will not be that long.

bpf/gotracer/go_sql.c
```diff
@@ -25,11 +25,99 @@
...
+// driverConn -> ci (interface) -> mysqlConn -> cfg (*Config) -> Addr (string)
+// Returns the Addr string which has format "host:port"
+static __always_inline int
+read_mysql_hostname_from_driverconn(u64 driver_conn_ptr, char *hostname, int max_len) {
+    if (driver_conn_ptr == 0) {
+        bpf_dbg_printk("can't read mysql hostname: driver_conn_ptr is null");
+        return 0;
+    }
+
+    off_table_t *ot = get_offsets_table();
```
grcevski, line 36: what's the reason we check for nullness of the offset table?

bpf/gotracer/go_sql.c
```diff
@@ -25,11 +25,99 @@
...
+        &driver_impl_ptr,
+        sizeof(driver_impl_ptr),
+        (void *)(driver_conn_ptr + go_offset_of(ot, (go_offset){.v = _driverconn_ci_pos}) + 8));
+
+    if (res || !driver_impl_ptr) {
+        bpf_dbg_printk("can't read sql driver implementation pointer");
+        return;
+    }
+
+    if (read_mysql_hostname_from_driverconn(
```
grcevski: I have general question of how this would work... This probe is called from the main (all sql drivers supported) code. As far as I can tell, it doesn't do any validation if the actual database type is MySQL. So if this was a postgres database, we'll simply read nonsense from the random pointers for the host.

Am I understanding this wrong? How do we check that the code is actually going to run only for mysql?

bpf/gotracer/go_sql.c
```diff
@@ -25,11 +25,99 @@
...
+        return;
+    }
+
+    off_table_t *ot = get_offsets_table();
+    if (!ot) {
+        bpf_dbg_printk("sql hostname extraction failed: offset table not available");
+        return;
+    }
+
+    void *driver_impl_ptr = 0;
```
grcevski: this code is duplicate from above, perhaps it can be extracted as a helper function

bpf/common/common.h
marctc: Previously we were capturing DSN, which are longer. So this is just leftover from the past.

bpf/gotracer/go_sql.c
marctc: Just to be extra defensive, but if not needed we can remove it.

bpf/gotracer/go_sql.c
marctc: No, you are absolutely right, and I'm having that problem right now testing postgres. The solution is validate the offset and return 0 if is not myql. Something like this:

```
    u64 ci_offset = go_offset_of(ot, (go_offset){.v = _driverconn_ci_pos});
    u64 cfg_offset = go_offset_of(ot, (go_offset){.v = _mysql_conn_cfg_pos});
    u64 addr_offset = go_offset_of(ot, (go_offset){.v = _mysql_config_addr_pos});

    if (!ci_offset || !cfg_offset || !addr_offset) {
        bpf_dbg_printk("mysql driver offsets not available");
        return 0;
    }
```
but if you have a better idea how to do it, please let me know because i haven't found any similar pattern with that.

bpf/common/common.h
grcevski: Cool, yeah, let's reduce the space we need for these events.

bpf/gotracer/go_sql.c
grcevski: it's too verbose and we have log errors when we can't register an offsets table. we should remove those lines.

bpf/gotracer/go_sql.c
marctc: or add more logic to detect the driver...

grcevski: That wouldn't work unfortunately, it would assume that I only use mysql as a database in my code. Imagine I have two different database I need to query, the offsets would be setup because of the mysql driver and we'll still run the code even for the other database.

There are few ways to do this, but it depends on what's available for your situation:
1. You can inspect the struct of the driver and confirm that it's mysql, e.g. there's some obvious field set that would be only set for this driver and it must match 100%.
2. Use the approach we have for the manual span probes. That would require that we extract the address of the interface type for the mysql driver and then you compare the interface address if it matches.

marctc: Implemented option number 2, let me know if how it looks!

bpf/common/common.h
```diff
@@ -81,14 +82,16 @@ typedef struct http_request_trace {
...
-    u8 _pad[1];
+    u8 _pad0[1];
     u16 status;
     pid_info pid;
     u64 start_monotime_ns;
     u64 end_monotime_ns;
     tp_info_t tp;
     connection_info_t conn;
     unsigned char sql[SQL_MAX_LEN];
+    unsigned char hostname[SQL_HOSTNAME_MAX_LEN];
```
grcevski, line 93: If you use MAX_LEN for the hostname of 104 we can use these extra bytes for the padding.

bpf/gotracer/go_sql.c
```diff
@@ -25,11 +25,120 @@
...
+        return NULL;
+    }
+
+    u64 mysql_type_addr = go_offset_of(ot, (go_offset){.v = _mysql_conn_type_off});
+    if (!mysql_type_addr) {
+        bpf_dbg_printk("can't read mysql.mysqlConn offset");
+        return NULL;
+    }
+
+    bpf_dbg_printk("validating mysql conn type %llx with %llx",
```
grcevski: This looks like you are missing a semicolon.

marctc: yep, i just pushed now

bpf/common/common.h
```diff
@@ -81,14 +82,15 @@ typedef struct http_request_trace {
 
 typedef struct sql_request_trace {
     u8 type; // Must be first
-    u8 _pad[1];
+    u8 _pad0[1];
```
rafaelroquetto: nit: I think this change is no longer required

bpf/gotracer/go_common.h
```diff
@@ -103,6 +103,7 @@ typedef struct sql_func_invocation {
     u64 start_monotime_ns;
     u64 sql_param;
     u64 query_len;
+    u64 driver_conn_ptr; // Pointer to driverConn (to get *DB)
```
rafaelroquetto: nit: redundant comment (driver_conn_ptr literally means pointer to driverConn)

bpf/gotracer/go_offsets.h
```diff
@@ -88,12 +88,17 @@ typedef enum {
...
+    _driverconn_ci_pos,
+    // mysql driver
+    _mysql_conn_cfg_pos,
+    _mysql_config_addr_pos,
+    _mysql_conn_type_off,
     // route resolution
     _mux_template_pos,
     _gin_fullpath_pos,
     _last_go_offset,
 } go_offset_const;
```
rafaelroquetto, line 101: nit: I'd keep the empty line - it's standard practice in C and improves readability

bpf/gotracer/go_sql.c
```diff
@@ -25,11 +25,119 @@
...
+    // Get driverConn.ci offset
+    u64 ci_offset = go_offset_of(ot, (go_offset){.v = _driverconn_ci_pos});
+    if (!ci_offset) {
+        bpf_dbg_printk("can't get driverConn.ci offset");
+        return NULL;
+    }
+
+    // driverConn.ci is a Go interface [type_ptr (8 bytes), data_ptr (8 bytes)]
+    // Read the type pointer (at ci_offset + 0) to validate driver type
+    void *ci_type_ptr = 0;
```
rafaelroquetto:
```suggestion
    void *ci_type_ptr = NULL;
```

rationale: NULL is more customary in C code that `0` for pointers - C23 has followed suit with C++ and introduced `nullptr` -  https://en.cppreference.com/w/c/language/nullptr.html - so using `NULL` instead of `0` has the added benefit that we can simply find/replace NULL with nullptr when we migrate.

bpf/gotracer/go_sql.c
```diff
@@ -25,11 +25,119 @@
...
+    int res =
```
rafaelroquetto: hint: try to use `const` for immutable variables - not only does it improve readability (it signals to the reader the variable doesn't change / is not being reused) but it actually allows the compiler to enforce immutability and prevent errors from side-effects.

bpf/gotracer/go_sql.c
```diff
@@ -25,11 +25,119 @@
...
+        bpf_probe_read(&ci_type_ptr, sizeof(ci_type_ptr), (void *)(driver_conn_ptr + ci_offset));
+
+    if (res || !ci_type_ptr) {
```
rafaelroquetto:
```suggestion
    if (res != 0) {
```

1. I'd leave the case in which `ci_type_ptr` is `NULL` to be caught by line 63 - this way the debug print on line 62 will print when the pointer is NULL and let us know (which is different than failing to read the pointer).
2. this one is mostly a nit - but `if (res)` is confusing for people who do not know the return semantics of `bpf_probe_read_*` because it implies "true" which is the opposite of an error - explicitly checking for ` != 0` conveys the meaning that this is one of those POSIX-like functions that return 0 on success, -1 on error.

bpf/gotracer/go_sql.c
```diff
@@ -25,11 +25,119 @@
...
+    if (!ci_offset) {
+        bpf_dbg_printk("can't get driverConn.ci offset");
+        return NULL;
+    }
+
+    // driverConn.ci is a Go interface [type_ptr (8 bytes), data_ptr (8 bytes)]
+    // Read the type pointer (at ci_offset + 0) to validate driver type
+    void *ci_type_ptr = 0;
+    int res =
+        bpf_probe_read(&ci_type_ptr, sizeof(ci_type_ptr), (void *)(driver_conn_ptr + ci_offset));
```
rafaelroquetto:
```suggestion
        bpf_probe_read_user(&ci_type_ptr, sizeof(ci_type_ptr), (void *)(driver_conn_ptr + ci_offset));
```

From the bpf helpers manpage:

```

       long bpf_probe_read(void *dst, u32 size, const void *unsafe_ptr)

              Description
                     For tracing programs, safely attempt to read size
                     bytes from kernel space address unsafe_ptr and store
                     the data in dst.

                     Generally, use bpf_probe_read_user() or
                     bpf_probe_read_kernel() instead.

              Return 0 on success, or a negative error in case of
                     failure.
```

The reason is because `bpf_probe_read` needs an extra step to resolve if the ptr in question points to user or kernel memory. It's useful  in common code that handles both kernel and userspace pointers, but here I believe it's all user memory (correct me if I am wrong) - so `bpf_probe_read_user` is more appropriate (and also tells the reader the nature of the pointer).

bpf/gotracer/go_sql.c
```diff
@@ -25,11 +25,119 @@
...
+        return NULL;
+    }
+
+    return mysql_conn_ptr;
+}
+
+// Extracts MySQL server hostname from a validated mysqlConn pointer.
+// Follows the pointer chain: mysqlConn -> cfg (*Config) -> Addr (string)
+static __always_inline int
+read_mysql_hostname_from_mysqlconn(void *mysql_conn_ptr, char *hostname, int max_len) {
```
rafaelroquetto:
```suggestion
read_mysql_hostname_from_mysqlconn(void *mysql_conn_ptr, char *hostname, u64 max_len) {
```

Sizes and lengths are usually represented by unsigned values - traditionally `size_t`, which is not available in eBPF so we use u8, u16, u32 or u64 depending on the context.

You will notice that `read_go_str` for instance takes a u64 - having an `int` here for instance allows the caller to pass `-1` which makes no sense and will wrap to a very big number once it gets converted to unsigned.

bpf/gotracer/go_sql.c
```diff
@@ -25,11 +25,119 @@
...
+    void *cfg_ptr = 0;
+    int res = bpf_probe_read(
+        &cfg_ptr,
+        sizeof(cfg_ptr),
+        (void *)((u64)mysql_conn_ptr + go_offset_of(ot, (go_offset){.v = _mysql_conn_cfg_pos})));
+
+    if (res || !cfg_ptr) {
+        bpf_dbg_printk("can't read mysql.mysqlConn.cfg");
+        return 0;
+    }
```
rafaelroquetto, line 100: same comments above apply

bpf/gotracer/go_sql.c
rafaelroquetto:
```suggestion
read_mysql_hostname_from_mysqlconn(const void *mysql_conn_ptr, char *hostname, int max_len) {
```

rationale: this pointer is immutable in this function (if you check the prototype of bpf_probe_read_* you will also see that it takes a `const void*` based on the same rationale).

bpf/gotracer/go_sql.c
```diff
@@ -25,11 +25,119 @@
...
+        bpf_dbg_printk("can't read MySQL connection data pointer");
+        return NULL;
+    }
+
+    return mysql_conn_ptr;
+}
+
+// Extracts MySQL server hostname from a validated mysqlConn pointer.
+// Follows the pointer chain: mysqlConn -> cfg (*Config) -> Addr (string)
+static __always_inline int
```
rafaelroquetto:
```suggestion
static __always_inline bool
```

and return `true` or `false` instead - we've been using this in new code for improved readability - but it's just a nit here.

bpf/gotracer/go_sql.c
```diff
@@ -25,11 +25,119 @@
...
+        bpf_dbg_printk("sql hostname extraction skipped: driver_conn_ptr is null");
+        return;
+    }
+
+    void *mysql_conn_ptr = get_mysql_conn_ptr(driver_conn_ptr);
+    if (mysql_conn_ptr) {
+        if (read_mysql_hostname_from_mysqlconn(
+                mysql_conn_ptr, (char *)trace->hostname, sizeof(trace->hostname))) {
+            bpf_dbg_printk("extracted MySQL hostname: %s", trace->hostname);
+            return;
```
rafaelroquetto: this return is redudant

bpf/gotracer/go_sql.c
```diff
@@ -25,11 +25,119 @@
...
+static __always_inline void extract_sql_hostname(sql_request_trace_t *trace, u64 driver_conn_ptr) {
+    trace->hostname[0] = '\0';
+
+    if (driver_conn_ptr == 0) {
+        bpf_dbg_printk("sql hostname extraction skipped: driver_conn_ptr is null");
+        return;
+    }
+
+    void *mysql_conn_ptr = get_mysql_conn_ptr(driver_conn_ptr);
+    if (mysql_conn_ptr) {
```
rafaelroquetto: you may want to avoid excessive nesting by doing an early return here.

Rationale: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f56-avoid-unnecessary-condition-nesting

pkg/transform/name_resolver.go
```diff
@@ -194,10 +194,10 @@ func (nr *NameResolver) resolveNames(span *request.Span) {
 	}
 	// don't set names if the peer and host names have been already decorated
 	// in a previous stage (e.g. Kubernetes decorator)
-	if pn != "" {
+	if pn != "" && span.PeerName == "" {
```
grcevski: Why was this change needed? What did we have in the name that we need to overwrite?

marctc:
> We can't discard the existing name just like that, the lookup function works for all spans, and it consults names set by OTEL_SERVICE_NAME, which should always take precedence. You need to find what was discovered in the name that is non empty.

I moved this logic to only client spans section to only override hostname if what's returned by resolve it's not an IP

pkg/transform/name_resolver.go
```diff
@@ -177,7 +177,12 @@ func (nr *NameResolver) resolveNames(span *request.Span) {
 	}
 
 	if span.IsClientSpan() {
-		hn, span.OtherNamespace = nr.resolve(&span.Service, span.Host)
+		hn = span.HostName
```
grcevski: This is exactly the same as before @marctc. The nr.resolve code will be returning an IP address (e.g. Host) only if we didn't find any host name. There fore we check that condision on line 186 below. Namely, if the returned resolved value is empty or it's the same as Host we try from the host scheme.

What you really want to do is if, the same condition is met, e.g. hn == "" || hn == span.Host, is to set hn to the prior host name if it's not empty.

An alternative is to pass a third argument to nr.resolve that is the previous value. So if dnsResolve didn't return a value, then we should return the existing value for the name if it was set.

grcevski: Also, make sure you add unit tests for this special case once you fix the code.

marctc: cool, will modify `resolve`

PR #987, opened 2025-12-10
Remove Java OTel SDK injection

grcevski:
The code when donated had support for experimental dynamic injection of the OTel Java SDK, since we didn't support Java TLS. This support didn't fully work, since some of the OTel SDK instrumentations don't work with dynamic injection. For example, gRPC never worked, because the interceptor is injected at build time, so OBI had to see the start of the Java process, which is not feasible.

I experimented recently with making various patches to the Java OTel SDK to support dynamic attach, and while I got things to work for gRPC and Netty, I was working on an old branch (2.14). After I attempted to migrate my code to the latest Java OTel SDK, I realised that many instrumentations that worked in my version don't work anymore, mainly because of refactoring in the Java OTel SDK to support java invoke dynamic (i.e. they call it indy). Many instrumentation have been modified to not pass data through locals, but through separate container classes which cannot be found by the application class loader. I attempted to fix this but I failed.

Since this is not a supported mode for the Java OTel SDK, and things have moved in the opposite direction, I'm removing our code for this here, but I'm keeping the injection classes, because we are going to use those to inject the OBI Java agent for TLS support.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,7 +85,6 @@ internal/test/integration/components/pythonserver/pyvenv.cfg
 internal/test/integration/components/gosql/qosql
 *_bpfel.go
 *_bpfel.o
-pkg/internal/otelsdk/grafana-opentelemetry-java.jar
 opentelemetry-ebpf-instrumentation.sln
 obi.sln
 internal/test/integration/components/javakafka/target/*


rafaelroquetto (approved):

marctc (approved):
I guess this PR has some breaking changes as we removed a field. That field wasn't documented tho.

skl (approved):

mariomac (approved):

Issue #982, opened 2025-12-10
ARM64: can't instrument NodeJS after "token __arm64_do_vfs_ioctl: not found: no such file or directory"


mariomac:
`TestSuite_NodeJS` can't run in different virtualized ARM64 Kernels. Beyla logs the following messages:

```
time=2025-12-10T14:06:03.101Z level=DEBUG msg="going to add kprobe to function" component=ebpf.Instrumenter probes=kprobes function=do_vfs_ioctl probes="{Required:true Start:Kprobe(obi_kprobe_do_vfs_ioctl)#83 End:<nil> StartOffset:0 ReturnOffsets:[]}"
time=2025-12-10T14:06:03.101Z level=WARN msg="couldn't load tracer" component=ebpf.ProcessTracer error="instrumenting function \"do_vfs_ioctl\": setting kprobe: creating perf_kprobe PMU (arch-specific fallback for \"do_vfs_ioctl\"): token __arm64_do_vfs_ioctl: not found: no such file or directory" required=true
time=2025-12-10T14:06:03.101Z level=ERROR msg="couldn't trace process. Stopping process tracer" component=discover.traceAttacher error="instrumenting function \"do_vfs_ioctl\": setting kprobe: creating perf_kprobe PMU (arch-specific fallback for \"do_vfs_ioctl\"): token __arm64_do_vfs_ioctl: not found: no such file or directory"
```

Tested environments:

* OrbStack Mac (Kernel Linux ubuntu 6.17.8-orbstack-00308-g8f9c941121b)
* Docker Desktop Mac (Engine: 29.1.2, Kernel: 6.12.54-linuxkit)

Complete logs:

```
time=2025-12-10T14:06:01.789Z level=INFO msg="OpenTelemetry eBPF Instrumentation" Version=unset Revision=unset "OpenTelemetry SDK Version"=1.38.0
time=2025-12-10T14:06:01.791Z level=DEBUG msg="trying to fetch host ID" component=ContextInfo func=fetchHostID fetcher=AWS
time=2025-12-10T14:06:02.296Z level=DEBUG msg="didn't get host ID" component=ContextInfo func=fetchHostID fetcher=AWS cause="timed out waiting for host ID connection"
time=2025-12-10T14:06:02.296Z level=DEBUG msg="trying to fetch host ID" component=ContextInfo func=fetchHostID fetcher=Azure
[...]

PR #990, opened 2025-12-10
Rework ioctl to tap into the syscall

grcevski:
Apparently do_vfs_ioctl doesn't exist on all kernels, so I'm modifying the code to probe into sys_ioctl. The syscall is a bit uglier in code, since we have to unwrap the registers from the syscall calling convention, but it will be there on all kernels.

Closes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/982

### DIFF
--- a/bpf/generictracer/java_tls.c
+++ b/bpf/generictracer/java_tls.c
@@ -36,18 +36,27 @@ static __always_inline u8 cmd_to_op(u8 cmd) {
     }
 }
 
-SEC("kprobe/do_vfs_ioctl")
-int BPF_KPROBE(obi_kprobe_do_vfs_ioctl, void *filp, unsigned int fd, unsigned int cmd, void *arg) {
-    (void)ctx;
-    (void)filp;
-
+SEC("kprobe/sys_ioctl")
+// unsigned int fd, unsigned int cmd, void *arg
+int BPF_KPROBE(obi_kprobe_sys_ioctl) {
     u64 id = bpf_get_current_pid_tgid();
 
     if (!valid_pid(id)) {
         return 0;
     }
 
-    bpf_dbg_printk("=== do_vfs_ioctl id=%d ===", id);
+    bpf_dbg_printk("=== sys_ioctl id=%d ===", id);
+
+    // unwrap the syscall arguments in __ctx
+    struct pt_regs *__ctx = (struct pt_regs *)PT_REGS_PARM1(ctx);
+
+    unsigned int fd = 0;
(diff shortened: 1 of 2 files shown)


bpf/generictracer/java_tls.c
```diff
@@ -36,18 +36,27 @@ static __always_inline u8 cmd_to_op(u8 cmd) {
...
 }
 
-SEC("kprobe/do_vfs_ioctl")
-int BPF_KPROBE(obi_kprobe_do_vfs_ioctl, void *filp, unsigned int fd, unsigned int cmd, void *arg) {
-    (void)ctx;
-    (void)filp;
-
+SEC("kprobe/sys_ioctl")
+// unsigned int fd, unsigned int cmd, void *arg
+int BPF_KPROBE(obi_kprobe_sys_ioctl) {
```
mmat11, line 41: maybe you can try using this macro https://github.com/libbpf/libbpf/blob/master/src/bpf_tracing.h#L899 which does the args unwrapping automatically; alternatively these helper macros could be used instead: https://docs.ebpf.io/ebpf-library/libbpf/ebpf/PT_REGS_PARM_SYSCALL/, https://docs.ebpf.io/ebpf-library/libbpf/ebpf/PT_REGS_SYSCALL_REGS/

note: to use any of this, the `bpf_tracing.h` header needs to be updated

grcevski: Oh interesting, I didn't know they added the macro... perhaps we can follow-up with a cleanup? I'm not sure what/if anything breaks if I update bpf_tracing? I can give it a shot.

grcevski: Argh, it requires a full upgrade of our libbpf. The new version depends on bpf_helpers and the PT_REGS_PARM_SYSCALL isn't defined. None of these are in our source.

grcevski: I'd rather do this upgrade separately, looks like a bigger change. I was thinking of bringing in that macro, but it depends on these:
```
#define ___bpf_syscall_args0()           ctx
#define ___bpf_syscall_args1(x)          ___bpf_syscall_args0(), (unsigned long long)PT_REGS_PARM1_SYSCALL(regs)
```
which we need to bring in, at which point it's a big upgrade.

mmat11: yeah if this works we can leave it as a follow up

mmat11 (approved):

PR #991, opened 2025-12-11
Refactor tctracer

rafaelroquetto:
This is a large PR that refactors the `tctracer`. The trigger for this work was originally https://github.com/grafana/beyla/issues/2293 - the major refactor happened when I decided to switch to direct packet access. In a nutshell, this PR:

1. fixes IPv4 packet checksum calculation
2. handles Ethernet VLAN extended headers
3. switches both IPv4 and IPv6 parsing to direct packet mode
4. uses direct packet mode to construct `connection_info_t`
5. adheres to IPv6 headers recommendations from [RFC-8200](https://www.rfc-editor.org/rfc/rfc8200) and [RFC-4727](https://www.rfc-editor.org/rfc/rfc4727.html) - we now use an appropriate IPv6 option code (`0x1e`) which carries [semantic meaning](https://www.rfc-editor.org/rfc/rfc4727.html#section-8)
6. only injects IPv6 options when no other extension headers are present - there's law and order ruling those headers (in terms of ordering and quantity), and this conservative approach prevents packet corruption.
7. simplifies `track_sock` to use verifier-friendly stack values rather than pointers into the packet data
8. improves debug logging by printing actual traceparents

### DIFF
--- a/bpf/tctracer/tctracer.c
+++ b/bpf/tctracer/tctracer.c
@@ -660,19 +660,9 @@ static __always_inline void inject_ip_options(struct __sk_buff *skb,
         return;
     }
 
-    // this shouldn't ever be reached, as the tp should have already been
-    // deleted by the kprobes when tp->written == 1, but it does not hurt to
-    // be robust
     if (tp->written) {
         bpf_dbg_printk("tp already written by L7, not injecting IP options");
         bpf_map_delete_elem(&outgoing_trace_map, &e_key);
-        return;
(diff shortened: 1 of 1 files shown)


grcevski:
I don't remember why we originally had the valid set to 1 and then to 2, rather than deleting the outgoing options. Maybe it was early prototype code, but this behaves slightly differently now. Although I tested the branch myself and it works as is :).

PR #997, opened 2025-12-11
fix: fallback to http host header for service graph

### LINKED ISSUE #995: Use HTTP Host header for service name resolution when K8s lookup fails
For HTTP client spans, when the destination Pod IP cannot be resolved via K8s metadata, fall back to parsing the HTTP Host header from `span.Statement`.

If the Host header contains a K8s Service FQDN (e.g., `my-service.my-namespace.svc.cluster.local`), extract the service name and namespace.

This provides more accurate service graph connections for HTTP traffic.

skl:
Fixes #995

## Checklist

### DIFF
--- a/pkg/transform/name_resolver.go
+++ b/pkg/transform/name_resolver.go
@@ -144,6 +144,19 @@ func isValidRDNS(ip string) bool {
 		ip != "::"
 }
 
+// parseK8sFQDN returns the service name and namespace from a Kubernetes FQDN.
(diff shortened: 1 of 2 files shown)


grcevski (approved):
This is a great fix, I think it's good to go as is out of draft.

I was initially worried that we need to handle partial strings of `svc.cluster.local` because we have a limit on the buffers from eBPF space, but we made the fix to read them fully or not use them at all. E.g. we look for the \r character to see if we see it at the end of the host header.

PR #1000, opened 2025-12-12
Inject IP options for every packet of the request

rafaelroquetto:
Restore the behaviour of the old implementation in which the traceparent was injected for every packet belonging to the same request.

Also switches the IPv4 option from StreamID to the experimental range, as StreamID needs to be constrained to 4 bytes.


rafaelroquetto:
@grcevski I looked into this, and the IP options are mucking up the docker network itself.... so the tests fail.

I will split this PR into a more straightforward one fixing the ipv4 option code and deal with the injection later.

PR #1004, opened 2025-12-12
Fix eBPF preverification error in Kernel 6.18

mariomac:
## Checklist

### DIFF
--- a/.github/workflows/pull_request_oats_test.yml
+++ b/.github/workflows/pull_request_oats_test.yml
@@ -76,7 +76,7 @@ jobs:
         if: always()
         with:
           name: Oats test logs - ${{ matrix.basename }}
-          path: test/oats/${{ matrix.basename }}/build/*
+          path: internal/test/oats/${{ matrix.basename }}/build/*
 
       - name: Process coverage data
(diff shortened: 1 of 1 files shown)


mmat11 (approved):
lgtm, out of curiosity, what was the error?

there's an additional `obi.yml`, not sure it's intentional

mariomac:
@mmat11  to me it looks as a flaw in the eBPF preverifier. The error was:
```
105: (85) call bpf_probe_read_kernel#113      ; R0=scalar() fp-8=mmmmmmmm fp-16=mmmmmmmm fp-24=mmmmmmmm
; tid->ns = (u32)ns.inum; @ pid_helpers.h:82
106: (61) r1 = *(u32 *)(r6 +24)
invalid read from stack R6 off=0 size=4
processed 105 insns (limit 1000000) max_states_per_insn 0 total_states 7 peak_states 7 mark_read 0
```
The preverifier didn't like accessing `ns.inum` outside `BPF_CORE_READ`, as if `ns` could be nil.

Thanks for noticing the `obi.yml`. It was added by mistake. Removing.

Issue #985, opened 2025-12-10
Next.js route harvesting via routes-manifest.json


matt-hensley:
As part of the Next.js build process, a file named `routes-manifest.json` in the `.next` directory is generated that contains a structured list of what appear to be all the route definitions. This includes pages, redirects, rewrites, and so on.

Finding and parsing this manifest would give better span names with appropriate cardinality. [Span name cardinality for Next.js is a known issue that hasn't had a fix land yet](https://github.com/vercel/next.js/pull/64852).

[I created a gist with the full `routes-manifest.json` from a demo app](https://gist.github.com/matt-hensley/8d5491185e3233039c6d28dc61faae30).

Quick example of some of the route data available:
```
{
  "version": 3,
  "caseSensitive": false,
  "basePath": "",
  "redirects": [
    {
      "source": "/:path+/",
      "destination": "/:path+",
      "internal": true,
      "priority": true,
      "statusCode": 308,
      "regex": "^(?:/((?:[^/]+?)(?:/(?:[^/]+?))*))/$"
    }
  ],
  "dynamicRoutes": [
    {
      "page": "/users/[userId]/posts/[postId]",
      "regex": "^/users/([^/]+?)/posts/([^/]+?)(?:/)?$",
      "routeKeys": {
        "nxtPuserId": "nxtPuserId",
        "nxtPpostId": "nxtPpostId"
      },
      "namedRegex": "^/users/(?<nxtPuserId>[^/]+?)/posts/(?<nxtPpostId>[^/]+?)(?:/)?$"
    }
  ],
  "staticRoutes": [
    {
      "page": "/favicon.ico",
      "regex": "^/favicon\\.ico(?:/)?$",
      "routeKeys": {},
      "namedRegex": "^/favicon\\.ico(?:/)?$"
    }
  ]
}
```

The Vercel serverless platform uses the manifest when serving Next.js apps, and while the format is not well documented, I'd expect it to be relatively stable since their platform relies on it.

PR #1007, opened 2025-12-14
feat: Next.js route harvesting from manifest

tongoss:
fix: #985


grcevski (approved):
Very cool! LGTM! If tests pass we should be good to merge this.

Issue #973, opened 2025-12-08
Need comprehensive documentation for OBI config file fields


itsCheithanya:
**Description:**
Currently, there doesn't appear to be complete documentation listing all available fields that the OBI config file supports when running `./obi -config instrument-config.yml`.

**Request:**
 providing comprehensive documentation that includes:
- A complete list of all supported config file fields
- A template/example config file with all available options
- Description of each field's purpose and accepted values
- Default values for optional fields
- Examples of common configuration scenarios

**Use Case:**
Users need a reference template to understand all configuration options available when instrumenting applications with OBI, rather than discovering fields through code inspection.

**Additional Context:**
This would help users:
- Quickly set up proper instrumentation configurations
- Understand all available customization options
- Avoid configuration errors

PR #1017, opened 2025-12-16
Add Config json schema generation

NimrodAvni78:
Fixes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/973
Resolve https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/53

Automatically generates json schema for config struct, with enum values, deprecated warnings and comments as descriptions

also validates in CI that json schema is the same

### DIFF
--- a/cmd/obi-schema/main.go
+++ b/cmd/obi-schema/main.go
@@ -301,7 +301,7 @@ func main() {
 		FieldNameTag:               "yaml",
 		Mapper:                     g.customMapper(),
 	}
-	if err := reflector.AddGoComments("go.opentelemetry.io/obi", "./"); err != nil {
+	if err := reflector.AddGoComments("go.opentelemetry.io/obi", "./", jsonschema.WithFullComment()); err != nil {
 		fmt.Fprintf(os.Stderr, "Warning: could not add Go comments: %v\n", err)
 	}
 
@@ -317,6 +317,9 @@ func main() {
 
 	// Add environment variable annotations
 	g.processEnvVars(schema)
+
+	// Normalize descriptions: collapse newlines into single spaces
+	normalizeDescriptions(schema)
 
 	// Sort properties for deterministic output
 	sortSchemaProperties(schema)
@@ -654,6 +657,16 @@ func processSchemaDeprecation(schema *jsonschema.Schema) {
 			return
 		}
 	}
+}
+
+// normalizeDescriptions collapses newlines in all schema descriptions into single spaces.
+func normalizeDescriptions(schema *jsonschema.Schema) {
+	visitNestedSchemas(schema, func(s *jsonschema.Schema) {
+		if s == nil || s.Description == "" {
+			return
+		}
+		s.Description = strings.ReplaceAll(s.Description, "\n", " ")
+	})
 }
 
 // jsonSchemaerType is the reflect.Type for the jsonSchemaer interface.
--- a/docs/config-schema.json
+++ b/docs/config-schema.json
@@ -52,12 +52,12 @@
         },
         "metric_span_names_limit": {
           "type": "integer",
-          "description": "MetricSpanNameAggregationLimit works PER SERVICE and only relates to span_metrics.\nWhen the span_name cardinality surpasses this limit, the span_name will be reported as AGGREGATED.\nIf the value <= 0, it is disabled.",
+          "description": "MetricSpanNameAggregationLimit works PER SERVICE and only relates to span_metrics. When the span_name cardinality surpasses this limit, the span_name will be reported as AGGREGATED. If the value <= 0, it is disabled.",
           "x-env-var": "OTEL_EBPF_METRIC_SPAN_NAMES_LIMIT"
         },
         "rename_unresolved_hosts": {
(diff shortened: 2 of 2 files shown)


pkg/config/ebpf_tracer.go
```diff
@@ -23,7 +25,13 @@ const (
...
 
 	// Convenience aliases
-	ContextPropagationAll = ContextPropagationHeaders | ContextPropagationTCP
+	ContextPropagationAll           = ContextPropagationHeaders | ContextPropagationTCP
+	ContextPropagationDisabledText  = "disabled"
+	ContextPropagationAllText       = "all"
+	ContextPropagationHeadersText   = "headers"
+	ContextPropagationHTTPText      = "http"
+	ContextPropagationTCPText       = "tcp"
+	ContextPropagationIPOptionsText = "ip"
```
mariomac: It might be a bit confusing that some "ContextPropagation..." variables are integers and others are strings.

I'd rename these variables to `StrContextPropagation`, for example.

Also, make them private if they are going to be only used here.

cmd/obi-schema/main.go
```diff
@@ -0,0 +1,363 @@
+// Copyright The OpenTelemetry Authors
```
marctc, line 1: I wonder if this belongs to this repo, as it looks pretty generic. Are there any project ins the open-telemetry ecosystem doing that? Maybe its worth to move this somewhere later if that can be reused?

NimrodAvni78: i couldn't find it in projects like otel-collector, maybe we can make it something that can be reused
im not familiar with how other otel projects manage schema.
maybe @MrAlias you have more insight?

itsCheithanya:
@NimrodAvni78 https://github.com/itsCheithanya/opentelemetry-ebpf-instrumentation/blob/f5edac2114cad288ee315019ddee3e15a62ed677/pkg/appolly/services/attr_glob.go check this if it can be be of any help

NimrodAvni78:
also noticed the field is actually inlined and not under another field called metadata, so i also fixed that

pkg/config/ebpf_tracer.go
MrAlias: +1 to making private.

docs/config-schema.json
MrAlias, line 1: What validation testing is done for this schema?

Is it all local? Should it be added to the CI?

cmd/obi-schema/main.go
MrAlias, line 1: This is missing tests demonstrating schema generation correctness.

docs/config-schema.json
NimrodAvni78: CI will fail if there is a diff between committed schema and newly generated schema, see `make generate-config-schema` and `make check-config-schema`

cmd/obi-schema/main.go
NimrodAvni78: will add some testing

docs/config-schema.json
NimrodAvni78: this means users need to run `make generate-config-schema`, maybe we can make this as part of `make fmt`

docs/config-schema.json
```diff
@@ -0,0 +1,1902 @@
...
+            "local"
+          ]
+        },
+        {
+          "type": "string",
+          "pattern": "^name:.+$"
+        }
+      ],
+      "title": "Agent Type Interface",
+      "description": "Specifies which interface should the agent pick the IP address from in order to report it in the AgentIP field on each flow. Accepted values are: external, local, or name:\u003cinterface name\u003e (e.g. name:eth0)."
```
antonjim-te: the generated JSON currently escapes < / > in descriptions (e.g., name:\u003cinterface name\u003e). 

Can we try to avoid HTML escaping? or were they expected?

docs/config-schema.json
```diff
@@ -0,0 +1,1902 @@
...
+          "items": {
+            "type": "string",
+            "enum": [
+              "headers",
+              "http",
+              "ip",
+              "tcp"
+            ]
+          },
+          "type": "array",
```
antonjim-te: it is not parsed as an array, it is parsed a comma-separated 
```
parts := strings.Split(str, ",")
```

is something wrong here?

pkg/config/ebpf_tracer.go
```diff
@@ -23,7 +25,13 @@ const (
 	ContextPropagationIPOptions ContextPropagationMode = 1 << 2 // IP options (dangerous)
 
 	// Convenience aliases
-	ContextPropagationAll = ContextPropagationHeaders | ContextPropagationTCP
+	ContextPropagationAll         = ContextPropagationHeaders | ContextPropagationTCP
```
antonjim-te, line 29: is it expected that `All` does not contain `IP` in the description it says "All contains all the methods"

NimrodAvni78: yes its intentional to be removed by default
see https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1001

MrAlias (approved):

PR #1018, opened 2025-12-16
Add SQL database hostname capture for Postgres

marctc:
Current implementation only captures IP of database connections, which ultimately are set as outgoing.
This PR adds extra logic to capture the hostname of  SQL operations to Postgres in Go.

The implementation add hooks for lib/pg for the function `network` (https://github.com/lib/pq/blob/master/conn.go#L436) which is called on each db operation.

### DIFF
--- a/bpf/gotracer/go_sql.c
+++ b/bpf/gotracer/go_sql.c
@@ -27,13 +27,7 @@
 #include <gotracer/go_common.h>
 #include <gotracer/go_str.h>
 
-// Store hostname by goroutine for PostgreSQL (lib/pq)
-struct {
-    __uint(type, BPF_MAP_TYPE_LRU_HASH);
-    __type(key, go_addr_key_t); // goroutine key
-    __type(value, char[SQL_HOSTNAME_MAX_LEN]);
-    __uint(max_entries, MAX_CONCURRENT_REQUESTS);
-} pq_hostnames SEC(".maps");
+#include <maps/go_sql.h>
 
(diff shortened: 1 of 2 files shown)


bpf/gotracer/go_sql.c
```diff
@@ -27,6 +27,14 @@
 #include <gotracer/go_common.h>
 #include <gotracer/go_str.h>
 
+// Store hostname by goroutine for PostgreSQL (lib/pq)
```
grcevski: Can we please move this code into a standalone file in the maps folder under the go tracer, like we've started doing for other maps?

bpf/gotracer/go_sql.c
```diff
@@ -114,7 +122,8 @@ read_mysql_hostname_from_mysqlconn(void *mysql_conn_ptr, char *hostname, u64 max
 
 // SQL hostname extraction with driver type routing.
 // Attempts to extract hostname by trying supported database drivers
-static __always_inline void extract_sql_hostname(sql_request_trace_t *trace, u64 driver_conn_ptr) {
+static __always_inline void
+extract_sql_hostname(sql_request_trace_t *trace, u64 driver_conn_ptr, void *goroutine_addr) {
     trace->hostname[0] = '\0';
 
     if (driver_conn_ptr == 0) {
```
grcevski: Could this check prevent us now from fetching the postgres data? it seems that the postgres support doesn't depend on the driver connection pointer. Perhaps best is to move the code for postgres forward ahead of this check.

PR #1019, opened 2025-12-16
Per-application metric features

mariomac:
In addition to the global metric features selector, it allows overriding the exported features per application. For example:
```yaml
metrics:
  features: ["application_span"]
discovery:
  instrument:
    - name: pytestserver
      open_ports: 7773
      metrics: { features: ["application", "application_span"] }
    - name: testserver
      open_ports: 8080
    - name: jtestserver
      open_ports: 8085
    - name: rtestserver
      open_ports: 8090
```

Above, by default, the services would only export Span Metrics, while the `pytestserver` service in port `7773` would also export application RED metrics, because we overrode the features in the `discovery > instrument` entry.


grcevski (approved):
LGTM! Nice work!

PR #1021, opened 2025-12-16
Inject OBI Java Agent for Java processes

grcevski:
This PR finishes the work to add the Java agent to Java processes so we can sniff out TLS traffic and get telemetry for it.

Our OBI Java agent had two components, one a loader program which has no dependencies and another the agent itself. I discovered an issue with classloading and the loader program, for all lazily loaded classes. If the classes weren't loaded at the time the agent is loaded, bytebuddy would have issues with Netty instrumentation. I've disabled the loader for now and we directly build the agent. This shouldn't be a problem, since we shade the classes and we wouldn't have duplicates. I added also an already loaded check in the Agent modules, since it was previously tested in the loader.

TODO:
- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/pkg/internal/java/java_inject_test.go
+++ b/pkg/internal/java/java_inject_test.go
@@ -38,7 +38,7 @@ func TestJavaInjector_CopyAgent(t *testing.T) {
 				require.NoError(t, os.WriteFile(tmpFile, []byte("test agent content"), 0o644))
 				return tmpFile
 			},
-			setupTempDir: func(t *testing.T, pid int32) string {
+			setupTempDir: func(t *testing.T, _ int32) string {
 				tmpDir := t.TempDir()
 				procRoot := filepath.Join(tmpDir, "proc", "root")
 				require.NoError(t, os.MkdirAll(filepath.Join(procRoot, "tmp"), 0o755))
@@ -56,7 +56,7 @@ func TestJavaInjector_CopyAgent(t *testing.T) {
 				require.NoError(t, os.WriteFile(tmpFile, []byte("test agent content"), 0o644))
 				return tmpFile
 			},
-			setupTempDir: func(t *testing.T, pid int32) string {
+			setupTempDir: func(t *testing.T, _ int32) string {
 				tmpDir := t.TempDir()
 				procRoot := filepath.Join(tmpDir, "proc", "root")
 				customTmpDir := filepath.Join(procRoot, "custom", "tmp")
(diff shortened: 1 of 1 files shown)


pkg/internal/java/java_inject.go
```diff
@@ -0,0 +1,351 @@
...
+		}
+
+		buf.WriteByte(b)
+		if b == '\n' {
+			if end, err := returnCodeLine(buf.String()); end {
+				return err
+			}
+
+			buf.Reset()
+		} else if b == 0 { // j9 terminates with 0
```
skl, line 279: Do you need to check if the `0` is at the end of the buffer? Otherwise it seems this conditional is true for any byte that is `0`?

pkg/internal/java/java_inject.go
```diff
@@ -0,0 +1,351 @@
...
+
+func (i *JavaInjector) verifyJVMVersion(attacher *jvm.JAttacher, pid int32) bool {
+	attacher.Init()
+
+	defer func() {
+		if err := attacher.Cleanup(); err != nil {
+			slog.Warn("error on JVM attach cleanup", "error", err)
+		}
+	}()
+	// OpenJ9 doesn't support VM.version command
```
skl, line 328: OpenJ9 was mentioned a couple of times in the code comments, is it also worth mentioning in the error message? Some users might be open to swapping out the runtime if they know the issue is specific to OpenJ9.

pkg/internal/java/java_inject.go
```diff
@@ -0,0 +1,351 @@
...
+
+	if out == nil {
+		return true
+	}
+
+	scanner := bufio.NewScanner(out)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.HasPrefix(line, "JDK ") {
+			return !strings.HasPrefix(line, "JDK 26")
```
skl, line 343: Why not JDK 26?

pkg/internal/java/java_inject.go
grcevski: The output from the JVMs is always textual (ASCII), so seeing 0 means it's over for J9, Hotspot never puts 0 in their stream and for them they properly send EOF. I think it's fine as is.

grcevski: This code was translated from jattach, this is how they check for end as well.

pkg/internal/java/java_inject.go
grcevski: In this case it's fine, since we call VM.version and the other unsupported APIs with ignoreOnJ9 = true. So if there's an error, it's a true error, on J9 we just say OK, we won't run that API. These calls are nice to have so that we know if something is supported, if it's not we'll attempt the attach and fail with an error.

pkg/internal/java/java_inject.go
grcevski: So the JVMs say that dynamic attach will be removed at some point, so far it wasn't in the JDK 25 which is an LTS, but they might in JDK 26 which is in rampdown phase 1 of development. According to what's landing in 26 (https://openjdk.org/projects/jdk/26/), they will not remove dynamic attach by default, but let's wait for a release candidate before we say it's OK.

PR #1023, opened 2025-12-17
Fix panic when parsing MongoDB client data

mariomac:
OBI sometimes crashes with the following panic:
```
1765971730951	2025-12-17T11:42:10.951Z	panic: runtime error: slice bounds out of range [:237] with capacity 236
1765971730951	2025-12-17T11:42:10.951Z		/src/vendor/go.opentelemetry.io/obi/pkg/ebpf/common/mongo_detect_transform.go:286 +0x3a5
```

This PR adds some extra guards to avoid reading beyond the buffer limits.

### DIFF
--- a/pkg/ebpf/common/mongo_detect_transform.go
+++ b/pkg/ebpf/common/mongo_detect_transform.go
@@ -233,6 +233,9 @@ func addSectionToMessage(isResponse bool, pendingRequest *MongoRequestValue, sec
 // |    16B      |     4B      |     ?     | optional 4B |
 // +------------+-------------+------------------+
 func parseOpMessage(buf []uint8, time int64, isResponse bool, pendingRequest *MongoRequestValue) (*MongoRequestValue, bool, error) {
+	if len(buf) < msgHeaderSize+int32Size {
+		return nil, false, errors.New("packet too short for MongoDB flag bits")
+	}
 	flagBits := int32(binary.LittleEndian.Uint32(buf[msgHeaderSize : msgHeaderSize+int32Size]))
 	err := validateFlagBits(flagBits)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/internal/transform/route/harvest/java_notlinux.go
```diff
@@ -13,5 +13,12 @@ type (
 	}
 )
 
-func NewJavaRoutesHarvester() *JavaRoutes                                  { return nil }
+func NewJavaRoutesHarvester() *JavaRoutes {
```
mariomac, line 16: Unrelated but this fixes compilation in Mac.

pkg/ebpf/common/mongo_detect_transform.go
```diff
@@ -259,11 +262,10 @@ func parseOpMessage(buf []uint8, time int64, isResponse bool, pendingRequest *Mo
 
 func parseSections(buf []uint8) ([]mongoSection, error) {
 	offSet := 0
-	sections := []mongoSection{}
+	var sections []mongoSection
 	for offSet < len(buf) {
-
-		if len(buf[offSet:]) < int32Size {
-			return nil, errors.New("not enough data for section header")
+		if len(buf[offSet:]) == 0 {
```
mmat11, line 267: isn't this less strict than `< int32Size`?

skl (approved):

grcevski (approved):
LGTM!

pkg/ebpf/common/mongo_detect_transform.go
mariomac: At this point we only read 1 byte, for the message type. Then se advance the buffer and check again for `len(buf) < int32size` before reading the actual integer.

An issue with the original check was that we checked for `len(buf) < int32Size` but we were actually reading 5 bytes (the message type and the message size)

But if you feel safer, I can replace this by: `< int32Size + 1` through I think it's unnecessary as long as we keep the `if len(buf) < msgHeaderSize+int32Size` in line 236 and the `len(buf) < offSet+int32Size` in line 288.

mmat11: ok then if we check later (or if we don't read past that), I didn't see the whole function 👍🏻

mmat11 (approved):
lgtm!

PR #1027, opened 2025-12-18
feat: geoip decorator support

theSuess:
This PR adds support for GeoIP lookups using [ipinfo.io](https://ipinfo.io) or [MaxMind GeoIP2-Lite](https://www.maxmind.com/en/geolite-free-ip-geolocation-data).

By using this decorator, users can get more insights into traffic patterns without exploding the metric cardinality. An example metric looks like this:
```
obi_network_flow_bytes_total{direction="response",dst_asn="",dst_cidr="192.168.42.0/24",dst_country="",obi_ip="192.168.42.203",src_asn="AS16509",src_cidr="0.0.0.0/0",src_country="US"} 132
```

This functionality is enabled by setting the path to either the IPInfo Lite or GeoIP2 Lite databases.

We can't bundle the MaxMind database due to licensing restrictions (as discussed in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33510#issuecomment-2245511892) but it should be possible for ipinfo as it's licensed under the `Creative Commons Attribution-ShareAlike 4.0` license.

The files used for testing are the [MaxMind test-data](https://github.com/maxmind/MaxMind-DB/tree/main/test-data) and the [IPInfo sample-database](https://github.com/ipinfo/sample-database/)

### DIFF
--- a/pkg/internal/netolly/flow/geoip.go
+++ b/pkg/internal/netolly/flow/geoip.go
@@ -9,7 +9,9 @@ import (
 	"fmt"
 	"log/slog"
 	"net"
+	"time"
 
+	"github.com/hashicorp/golang-lru/v2/expirable"
 	"github.com/oschwald/maxminddb-golang"
 
 	attr "go.opentelemetry.io/obi/pkg/export/attributes/names"
@@ -23,6 +25,8 @@ import (
 type GeoIP struct {
 	IPInfo      IPInfoConfig  `yaml:"ipinfo"`
 	MaxMindInfo MaxMindConfig `yaml:"maxmind"`
+	CacheLen    int           `yaml:"cache_len" env:"OTEL_EBPF_NETWORK_GEOIP_CACHE_LEN" validate:"gte=0"`
+	CacheTTL    time.Duration `yaml:"cache_expiry" env:"OTEL_EBPF_NETWORK_GEOIP_CACHE_TTL" validate:"gte=0"`
 }
 
 type IPInfoConfig struct {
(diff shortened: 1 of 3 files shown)


theSuess:
I didn't check the underlying lookup performance too closely, but I can write a benchmark to make sure this is fast enough and put a cache in between if not

grcevski (commented):
It looks like there are some lint issues.

pkg/internal/netolly/flow/geoip.go
```diff
@@ -0,0 +1,172 @@
...
+		if !cfg.Enabled() {
+			return swarm.Bypass(input, output)
+		}
+		lookupFn, err := getLookupFn(cfg)
+		if err != nil {
+			return nil, err
+		}
+
+		log := geoiplog()
+		in := input.Subscribe(msg.SubscriberName("flow.ReverseDNS"))
```
mariomac: Please change `flow.ReverseDNS` by something like `flow.GeoIP`

pkg/internal/netolly/flow/geoip.go
```diff
@@ -0,0 +1,172 @@
...
+			info, err := lookupFn(addr.IP())
+			if err != nil {
+				return info, err
+			}
+			cache.Add(*addr, info)
+			return info, nil
+		}
+		return func(_ context.Context) {
+			defer output.Close()
+			log.Debug("starting reverse DNS node")
```
mariomac: Should be maybe something like `starting GeoIP node`?

pkg/internal/netolly/flow/geoip.go
```diff
@@ -0,0 +1,172 @@
...
+			for flows := range in {
+				for _, flow := range flows {
+					srcInfo, err := cachedLookup(flow.Id.SrcIP())
+					if err != nil {
+						log.Warn("failed to perform geoip lookup for source", "err", err)
```
mariomac: Could this flood the user output? In this case I'd either:
* set this log as `Debug` level
* or print the first message as warning, then the rest of messages as debug.

pkg/internal/netolly/flow/geoip.go
```diff
@@ -0,0 +1,172 @@
...
+					}
+					dstInfo, err := cachedLookup(flow.Id.DstIP())
+					if err != nil {
+						log.Warn("failed to perform geoip lookup for destination", "err", err)
```
mariomac: Same as for the previous `log.Warn`

mariomac (approved):
LGTM! Thanks for addressing the changes

grcevski (approved):
LGTM!

PR #1034, opened 2025-12-19
Add large buffer support for Kafka

pinoOgni:
This PR continues the work done  #512 

Now OBI parses kafka data in kernel space and it sends a large buffer event to the userspace. In this way we can parse big `MetadataResponse`. This is really useful when the packet is too big, for example when there are multiple brokers and unfortunately the needed topic is after that.

A dedicated integration test has also been added to test only Kafka large buffer with 3 Kafka brokers, allowing for a list of brokers and increasing the size of the `MetadataResponse`. 
Since we need the `MetadataResponse` to create the spans, in the testServer the `Consumer` has been configured to trigger the `Metadata` refresh so that OBI can capture it.

Moreover, a dedicated unit test has been added.

Finally, to increase the size of the packet, an ad hoc topic name was used, it is a poem on OBI (written by Gemini not by me).

### DIFF
--- a/bpf/generictracer/protocol_kafka.h
+++ b/bpf/generictracer/protocol_kafka.h
@@ -44,9 +44,9 @@ struct kafka_response_hdr {
     s32 correlation_id; // The correlation ID of this response
 };
 
-struct kafka_state_data {
+typedef struct kafka_state_data {
     s32 message_size;
-};
+} kafka_state_data_t;
 
 typedef struct kafka_state_key {
     connection_info_t conn;
@@ -57,18 +57,18 @@ typedef struct kafka_state_key {
 struct {
     __uint(type, BPF_MAP_TYPE_LRU_HASH);
     __type(key, kafka_state_key_t);
-    __type(value, struct kafka_state_data);
+    __type(value, kafka_state_data_t);
     __uint(max_entries, MAX_CONCURRENT_REQUESTS);
 } kafka_state SEC(".maps");
 
-struct kafka_correlation_data {
+typedef struct kafka_correlation_data {
     s32 correlation_id;
-};
+} kafka_correlation_data_t;
 
 struct {
     __uint(type, BPF_MAP_TYPE_LRU_HASH);
     __type(key, connection_info_t);
-    __type(value, struct kafka_correlation_data);
+    __type(value, kafka_correlation_data_t);
     __uint(max_entries, MAX_CONCURRENT_REQUESTS);
 } kafka_ongoing_requests SEC(".maps");
 
@@ -78,7 +78,7 @@ enum {
     k_kafka_hdr_request_api_version = 2,
     k_kafka_hdr_correlation_id = 4,
 
-    k_kafka_min_response_message_size_value = 4, //  correlation_id (4)
+    k_kafka_min_response_message_size_value = 4, // correlation_id (4)
 
     // https://kafka.apache.org/protocol#protocol_api_keys
     k_kafka_api_key_metadata = 3,
@@ -99,10 +99,16 @@ static __always_inline int kafka_read_message_size(const unsigned char *data, si
     bpf_probe_read(&message_size, k_kafka_hdr_message_size, (const void *)data);
     message_size = bpf_ntohl(message_size);
 
+    // we can be in the case where we already have the first part
+    // of the header saved in the map and we are reading the second
+    // part so we think that is message_size but it is actually
+    // key+version in case of request or the correlation id in case of
+    // response
+
     if (message_size < k_kafka_min_response_message_size_value ||
         message_size > k_kafka_message_size_max) {
-        bpf_dbg_printk("invalid message_size: %d", message_size);
-        return -1;
+        bpf_dbg_printk("possible invalid message_size: %d", message_size);
+        return 0;
     }
     return message_size;
 }
@@ -127,7 +133,7 @@ static __always_inline int kafka_store_state_data(const connection_info_t *conn_
     if (message_size == -1) {
         return 0;
     }
-    struct kafka_state_data new_state_data = {};
+    kafka_state_data_t new_state_data = {};
     new_state_data.message_size = message_size;
     kafka_state_key_t state_key = {.conn = *conn_info, .direction = direction};
     bpf_map_update_elem(&kafka_state, &state_key, &new_state_data, BPF_ANY);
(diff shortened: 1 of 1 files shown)


internal/test/integration/red_test_kafka.go
```diff
@@ -174,6 +174,54 @@ func testJavaKafka(t *testing.T) {
...
+			Route:   "http://localhost:8381",
+			Subpath: "message",
+			Comm:    "javakafka-lb",
+			Spans: []TestCaseSpan{
+				{
+					Name: "publish theotelebpfagentisperfectlyimpatientitskipsthecodethesdkandfindsthekernelssecretkeyitwatcheshttpandgrpctogiveyoumetricsforfreeapowerfulkernellevelspree",
+					Attributes: []attribute.KeyValue{
+						attribute.String("span.kind", "producer"),
+						attribute.String("messaging.operation.type", "publish"),
+						attribute.String("messaging.destination.name", "theotelebpfagentisperfectlyimpatientitskipsthecodethesdkandfindsthekernelssecretkeyitwatcheshttpandgrpctogiveyoumetricsforfreeapowerfulkernellevelspree"),
```
skl, line 195: I'm here for this

pinoOgni: For the poem? :)

> The OTel eBPF agent, Is perfectly impatient. It skips the code, the SDK, and finds the kernel's secret key. It watches HTTP and gRPC, to give you metrics for free. A powerful, kernel-level spree!

internal/test/integration/red_test_kafka.go
```diff
@@ -148,7 +148,7 @@ func testJavaKafka(t *testing.T) {
 				},
 				{
 					// TODO: in here we can't recognize the topic name since the metadata response is cut to the first 4 bytes
-					// in java, to get this to work we need to use eBPF large buffers for kafka, will do so in a future PR
+					// in java, to get this to work we need to use eBPF large buffers for kafka -> testJavaKafkaLargeBuffer
```
skl: is this still a TODO? sounds like it's implemented now

bpf/generictracer/protocol_tcp.h
```diff
@@ -187,7 +193,8 @@ static __always_inline void handle_unknown_tcp_connection(pid_connection_info_t
                                                           u16 orig_dport,
                                                           enum protocol_type protocol_type) {
     tcp_req_t *existing = bpf_map_lookup_elem(&ongoing_tcp_req, pid_conn);
-
+    u32 netns = task_netns();
+    bool is_server = is_listening(pid_conn->conn.d_port, netns);
```
skl, line 199: This check could misclassify a client as a server in the case that some unrelated service happens to be bound to this process's port (such as a metrics server). Not sure how likely that is, but it could be worth having a more robust classification here?

bpf/generictracer/protocol_kafka.h
```diff
@@ -0,0 +1,443 @@
...
+
+struct kafka_state_key {
+    connection_info_t conn;
+    u8 direction;
+    u8 _pad[3];
+};
+
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, connection_info_t);
```
skl: Should this be `kafka_state_key`?

```suggestion
    __type(key, kafka_state_key);
```

bpf/generictracer/protocol_kafka.h
```diff
@@ -0,0 +1,443 @@
...
+                                                            u8 direction) {
+    u8 offset = 0;
+
+    struct kafka_state_key state_key = {.conn = *conn_info, .direction = direction};
+    struct kafka_state_data *state_data = bpf_map_lookup_elem(&kafka_state, &state_key);
+    if (state_data != NULL && (state_data->message_size == data_len)) {
+        state_data->message_size = bpf_htonl(state_data->message_size);
+        bpf_probe_read(buf, k_kafka_hdr_message_size, (const void *)state_data);
+        offset += k_kafka_hdr_message_size;
+        bpf_map_delete_elem(&kafka_state, conn_info);
```
skl: Should this be `&state_key`?

```suggestion
        bpf_map_delete_elem(&kafka_state, &state_key);
```

bpf/generictracer/protocol_kafka.h
```diff
@@ -0,0 +1,443 @@
...
+    // Sanity checks
+    k_kafka_message_size_max = 1 << 13, // 8K
+};
+
+// Structure to define the min/max supported versions for an API key
+typedef struct {
+    s16 api_key;
+    s16 min_version;
+    s16 max_version;
+} kafka_api_version_info_t;
```
skl: Is this used?

pinoOgni: Nope! It's old code. Thanks

bpf/generictracer/protocol_kafka.h
pinoOgni: Yes! Thanks

bpf/generictracer/protocol_kafka.h
pinoOgni: Yes again, and thanks again!

internal/test/integration/red_test_kafka.go
pinoOgni: Yes, I can remove the **TODO**. I just wanted to leave a comment explaining why the asterisk is there.

bpf/generictracer/protocol_tcp.h
pinoOgni: ~off topic: In the first implementation, the algorithm tried to parse the event as a request and then as a response because there was a bug and `is_server` had the wrong values.~

The `is_listening` function is used in the [infer_packet_type](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/3781840a42ffed54dbeb070aa22ec72799ed61b0/bpf/generictracer/protocol_common.h#L44), shouldn't we have the same problem there too?

mariomac: I think is worth commenting this in the code so if at some point in the future this issue emerges, someone trying to look for the cause can have more context here.

pinoOgni: Sure!

pinoOgni:
Hi @grcevski, thanks so much for the advice! 

I followed your example and created a unit test for Kafka. I tested the two main functions `is_kafka` and `kafka_send_large_buffer`, as well as all the internally called functions. I added various tests, but I'm open to other uses cases.

P.s. Thanks to the unit test, I found a bug in the ebpf code that the integration tests hadn't detected!

grcevski (approved):
LGTM! Thanks for doing this.

mariomac (approved):

PR #1044, opened 2025-12-22
Update makefile goarch detection

pinoOgni:
This PR adds automatic detection of the system architecture for `GOARCH` var if is not manually provided via the command line.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,20 +7,7 @@ CACHE_CMD ?= k8s-cache
 CACHE_MAIN_GO_FILE ?= cmd/$(CACHE_CMD)/main.go
 
 GOOS ?= linux
-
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -7,7 +7,20 @@ CACHE_CMD ?= k8s-cache
 CACHE_MAIN_GO_FILE ?= cmd/$(CACHE_CMD)/main.go
 
 GOOS ?= linux
-GOARCH ?= amd64
```
mmat11, line 10: would this work?

`GOARCH ?= $(shell go env GOARCH || echo amd64)`

pinoOgni: even better!

grcevski (approved):
LGTM! Sweet!

PR #1050, opened 2026-01-02
rb: flush stale data on interval

mmat11:
This PR adds a goroutine that flushes the ringbuffers when:

- stale data is detected (available bytes > 0)
- last read was longer than the predefined interval (3s)

This was practically never needed till now because in normal systems there is always data flowing which triggers user space notifications, but there's now an use case for it (logs ringbuffer with "silent" targeted processes -- the 4kb watermark is never hit if the targeted app just prints once and never logs again)

## Checklist

### DIFF
--- a/pkg/ebpf/common/ringbuf.go
+++ b/pkg/ebpf/common/ringbuf.go
@@ -20,12 +20,17 @@ import (
 	"go.opentelemetry.io/obi/pkg/pipe/msg"
 )
 
+// Max interval before reading stale available bytes from the ring buffer
+const flushInterval = 3 * time.Second
+
 // ringBufReader interface extracts the used methods from ringbuf.Reader for proper
 // dependency injection during tests
 type ringBufReader interface {
 	io.Closer
 	Read() (ringbuf.Record, error)
 	ReadInto(*ringbuf.Record) error
+	AvailableBytes() int
+	Flush() error
 }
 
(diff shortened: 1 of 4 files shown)


pkg/ebpf/common/ringbuf.go
```diff
@@ -20,12 +20,17 @@ import (
 	"go.opentelemetry.io/obi/pkg/pipe/msg"
 )
 
+// Max interval before reading stale available bytes from the ring buffer
+const flushInterval = 3 * time.Second
```
NimrodAvni78, line 24: maybe this should be part of config?

mmat11: while it wouldn't hurt to have this configurable, I struggle to find use cases/examples for which configuring it higher or lower would make sense:

- if the ringbuffer experiences high traffic, this interval is high enough to be negligible from the performance perspective
- if the ringbuffer experiences very low traffic, 1 or 2 seconds less shouldn't change much at the end of the day

I avoided to put this in the config to avoid adding bloat for now, we can always add it later if we find we need it -- if we make this configurable we should also make it configurable per-ringbuffer.

let me know if this makes sense

grcevski: That explanation works for me, if @NimrodAvni78 agrees, I think we are good to merge.

NimrodAvni78: yeah also agree 👌

PR #1052, opened 2026-01-06
Java thread pool support (context propagation)

grcevski:
Since we now have the OBI java agent which can be dynamically loaded, we can extend the Java support to have proper context propagation for Java thread pools, which are the most common way async work is done in Java. This PR adds support for: Runnable, ForkJoin and Callable. There are other types of thread pools in the wild, and this doesn't add support for Reactive Java (e.g.RxJava), but it gets us out of the box coverage for many cases.

I had to add another native call, to be able to get the Java thread system thread ID. Java has a native API to get the system thread ID in Java 19 and newer, but we want to support older Java versions too (e.g. Java 8 and 11 together still have more usage than Java 17).

TODO:
- [x] Integration tests
- [x] Verify if it's possible to replace the newly added Caffeine LRU with Java Concurrent HashMap (faster, could be done in a follow-up PR). I think it should be, since we can clean-up at the time the callable, runnable or ForkJoinTask executes.

### DIFF
--- a/internal/test/integration/components/java_tls/wrapper/wrapper.go
+++ b/internal/test/integration/components/java_tls/wrapper/wrapper.go
@@ -55,7 +55,6 @@ func main() {
 
 	http.HandleFunc("/request", func(w http.ResponseWriter, r *http.Request) {
 		err := regularGetRequest(r.Context(), "https://testserver:8443/greeting")
-
 		if err != nil {
 			fmt.Printf("Error %v\n", err)
 		}
--- a/internal/test/integration/java_dist_test.go
+++ b/internal/test/integration/java_dist_test.go
@@ -21,7 +21,6 @@ import (
 )
 
 func testJavaNestedTraces(t *testing.T, slug string) {
-
 	// give enough time for the Java injector to finish and to
 	// harvest the routes
 	t.Log("checking proper server to client nesting for [/api/" + slug + "]")


pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/instrumentations/data/SSLStorage.java
```diff
@@ -129,4 +132,26 @@ public static Field getBootDebugOn() {
 
     return bootDebugOn;
   }
+
+  public static void trackTask(long threadId, Object task) {
+    if (task == null) {
+      return;
+    }
+    tasks.put(task.hashCode(), threadId);
```
skl: Apparently `task.hashCode()` is non-final so might be overridden with non-unique return value... you might consider `System.identityHashCode(task)` instead.

skl (commented):
💯

pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/instrumentations/data/SSLStorage.java
grcevski: Good idea!

bpf/common/trace_common.h
```diff
@@ -226,6 +227,36 @@ static __always_inline tp_info_pid_t *find_parent_process_trace(trace_key_t *t_k
     return NULL;
 }
 
+static __always_inline tp_info_pid_t *find_parent_java_trace(trace_key_t *t_key) {
+    // Up to 3 levels of thread nesting allowed
+    enum { k_max_depth = 3 };
```
mariomac, line 232: Stupid question (to learn). Why not using `const k_max_depth = 3`?

grcevski: I think it's all the same, I just followed what was done on the thread to thread lookup. We can probably refactor how we use constants in the future and make it standard across the project. At the moment we use a combination of defines, enums and constants. At the end of the day clang will generate the same code.

PR #1053, opened 2026-01-06
OBI as otel collector receiver

### LINKED ISSUE #1094: Allow OBI to run as OTEL collector receiver

NimrodAvni78:
This adds support of running OBI as an otel receiver
this helps consolidate with other components like the otel-profiler, plus gives us OpAPM support to allow managing obi with remote config

resolves #1094 
## Checklist

### DIFF
--- a/collector/factory_linux.go
+++ b/collector/factory_linux.go
@@ -20,11 +20,8 @@ import (
 	"go.opentelemetry.io/obi/pkg/obi"
 )
 
-// loggerOnce ensures we only set the default logger once
-var loggerOnce sync.Once
-
-func setDefaultLogger(rs receiver.Settings) {
-	loggerOnce.Do(func() {
+var setDefaultLogger = func(rs receiver.Settings) func() {
+	return sync.OnceFunc(func() {
 		slog.SetDefault(slog.New(zapslog.NewHandler(rs.Logger.Core())))
 	})
 }
--- a/internal/test/integration/components/gomysql-otelsql/gomysql_otelsql.go
+++ b/internal/test/integration/components/gomysql-otelsql/gomysql_otelsql.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/XSAM/otelsql"
 	_ "github.com/go-sql-driver/mysql"
+
 	semconv "go.opentelemetry.io/otel/semconv/v1.21.0"
 )
 
--- a/pkg/instrumenter/instrumenter.go
+++ b/pkg/instrumenter/instrumenter.go
@@ -137,7 +137,7 @@ func buildServiceNameTemplate(config *obi.Config) (*template.Template, error) {
 	return templ, nil
 }
 
-// BuildContextInfo populates some globally shared components and properties
+// BuildCommonContextInfo populates some globally shared components and properties
 // from the user-provided configuration
 func BuildCommonContextInfo(
 	ctx context.Context, config *obi.Config,


florianl (commented):
Appreciate the approach and effort 🙏

grcevski (commented):
This looks great @NimrodAvni78 ! What do we need to do to finish this PR? Would it be possible to add some sort of an integration test? I don't mind if that's in a separate PR.

collector/internal/controller.go
```diff
@@ -0,0 +1,119 @@
...
+		return nil
+	}
+
+	c.shared.started = true
+	ctx, c.shared.cancel = context.WithCancel(ctx)
+
+	// Run OBI in a goroutine
+	go func() {
+		defer close(c.shared.runDone)
+		c.shared.runErr = instrumenter.Run(ctx, c.shared.config)
```
florianl: If something fails in the start up phase in `instrumenter.Run()` in `buildCommonContextInfo()`, then users only will learn from it, once `(*Controller) Shutdown()` is called.
Maybe, as a follow up, `instrumenter.Run()` should be split, so errors from preparations like `buildCommonContextInfo()` could be reported earlier.

collector/internal/controller.go
```diff
@@ -0,0 +1,119 @@
...
+	"go.opentelemetry.io/obi/pkg/obi"
+)
+
+// sharedController manages a singleton OBI instance that can be shared between
+// traces and metrics receivers.
+type sharedController struct {
+	mu      sync.Mutex
+	config  *obi.Config
+	cancel  context.CancelFunc
+	started bool
```
florianl: I think, either `started` or `refCnt` should be required but not both. Is there a scenario, where `sharedController` is started multiple times?

collector/factory_linux.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+// loggerOnce ensures we only set the default logger once
+var loggerOnce sync.Once
+
+func setDefaultLogger(rs receiver.Settings) {
+	loggerOnce.Do(func() {
+		slog.SetDefault(slog.New(zapslog.NewHandler(rs.Logger.Core())))
+	})
+}
```
florianl: This could be simplified to this:
```suggestion
var setDefaultLogger = func(rs receiver.Settings) func() {
	return sync.OnceFunc(func() {
		slog.SetDefault(slog.New(zapslog.NewHandler(rs.Logger.Core())))
	})
}()
```

collector/internal/controller.go
NimrodAvni78: sounds like a good idea, i'll do that

collector/internal/controller.go
NimrodAvni78: yeah you are right i think only one is required
in the common case where obi is initiated as both a metrics receiver and a tracers receiver, we want them to use the same underlying obi process

actually this might interfere if someone will want to different obi receivers configured differently
i might need to have this by component ID

pkg/export/otel/otelcfg/consumer_exporter.go
```diff
@@ -0,0 +1,371 @@
+// Copyright The OpenTelemetry Authors
```
marctc, line 1: Interesting stuff, back then I didn't contemplate to do this for tracing. Is there somewhere in OTEL landscape that we could re-use it? Did you measure performance implications of collecting metrics with SDK and later converting to all of this? In any case, i think we should a good amount of tests.

NimrodAvni78: didn't get the chance to look at performance impact of this conversion, i was planning to have this merged and building a distribution which i can test under load

regarding tests i agree, both for this conversion and integration tests, i can add those now or in a following PR

grcevski: I think follow-up PR is fine, since this isn't used at the moment, we can build on top of it.

PR #1059, opened 2026-01-08
Add connection_type to service graph metrics

marctc:
This field was present but values were never set propperly.
Adding this back to be consistent of service graph metrics generated by tempo.

More information https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/servicegraphconnector

### DIFF
--- a/pkg/appolly/app/request/span.go
+++ b/pkg/appolly/app/request/span.go
@@ -648,7 +648,7 @@ func (s *Span) ServiceGraphKind() string {
 }
 
 // ServiceGraphConnectionType returns the connection_type for service graph metrics.
-// See: https://grafana.com/docs/tempo/latest/metrics-from-traces/service_graphs/
+// See: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/servicegraphconnector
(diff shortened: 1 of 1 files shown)


pkg/appolly/app/request/span.go
```diff
@@ -647,6 +647,25 @@ func (s *Span) ServiceGraphKind() string {
...
+	switch s.Type {
+	case EventTypeSQLClient, EventTypeRedisClient, EventTypeMongoClient:
+		return "database"
+	case EventTypeKafkaClient:
+		return "messaging_system"
+	case EventTypeHTTPClient:
+		if s.SubType == HTTPSubtypeAWSSQS {
+			return "messaging_system"
+		}
+		if s.SubType == HTTPSubtypeElasticsearch {
```
mmat11, line 662: probably `HTTPSubtypeAWSS3` belongs here too

marctc: isn't S3 object storage? i think semantic conventios for s3 are in object store category: https://opentelemetry.io/docs/specs/semconv/object-stores/

mmat11: you are right, I thought it belonged to "database"

PR #1061, opened 2026-01-08
bpf: proposal: remove compile-time defines in favor of runtime constants

mmat11:
Both approaches have pros/cons: while using runtime configuration via .rodata constants introduces some additional overhead (eg. unused maps being included in the final object file), it reduces the number of generated ebpf object files, shortens build times, reduces complexity and improves test coverage by making it easier to validate a single object file instead of all possible combinations

### DIFF
--- a/bpf/common/globals.h
+++ b/bpf/common/globals.h
@@ -7,3 +7,4 @@
 
 volatile const bool g_bpf_debug = false;
 volatile const bool g_bpf_traceparent_enabled = false;
+volatile const bool g_bpf_header_propagation = false;
--- a/bpf/gotracer/go_grpc.c
+++ b/bpf/gotracer/go_grpc.c
@@ -17,6 +17,7 @@
 
 #include <bpfcore/utils.h>
 
+#include <common/globals.h>
 #include <common/ringbuf.h>
 
 #include <gotracer/go_byte_arr.h>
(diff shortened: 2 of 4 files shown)


grcevski (commented):
Nice initiative!

I thought that we must have a define for NO_HEADER_PROPAGATION, because the write helper must never be seen by the verifier. Perhaps it's not an issue and this would work, in which case I think the compiler strips the code as if we were using a define.

Can you confirm if possible if this works when the kernel is set in integrity mode?

For the debug ring buffer, I think we can solve the memory increase by making the size of the ringbuf configurable by userspace. For example, if we are not in debug we can set the map size to 0 and otherwise the value it has today.

mmat11:
I did test on kernel 6.8 with lockdown=confidentiality and lockdown=integrity: while the former doesn't even allow reading memory, the latter seems to work ok:

```
time=2026-01-09T17:01:47.423+01:00 level=DEBUG msg="checking kernel lockdown mode, [none] allows us to propagate trace context" component=ebpf.ProcessTracer
time=2026-01-09T17:01:47.423+01:00 level=INFO msg="Kernel in lockdown mode or missing CAP_SYS_ADMIN." component=go.Tracer
time=2026-01-09T17:01:47.458+01:00 level=DEBUG msg="Linux kernel version" component=go.Tracer major=6 minor=8
```

```
lima@lima-cx-new:~$ sudo cat /sys/kernel/security/lockdown
none [integrity] confidentiality

lima@lima-cx-new:~$ sudo bpftool prog dump xlated name obi_uprobe_writeSubset
int obi_uprobe_writeSubset(struct pt_regs * ctx):
; int obi_uprobe_writeSubset(struct pt_regs *ctx) {
   0: (bf) r7 = r1
; if (!g_bpf_header_propagation) {
   1: (18) r1 = map[id:164][0]+2
   3: (71) r1 = *(u8 *)(r1 +0)
; }
   4: (b7) r0 = 0
   5: (95) exit
```

the verifier seems to be able to strip the branch as those global vars are constant. I would like to test this on an older 5.x kernel to make sure this isn't a new optimization -- also if you happen to have some test env with lockdown enabled on which to run a quick test would help me feel safer :D

mmat11:
Can confirm it works on 5.15

Issue #1030, opened 2025-12-18
findOvnMp0IP fails on OpenShift 4.20


pellared:
Environment ([OpenShift Local](https://developers.redhat.com/products/openshift-local)):

```
$ crc version
CRC version: 2.57.0+ae41f6
OpenShift version: 4.20.5
MicroShift version: 4.20.0
```

OBI fails in this line

https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/17b0cbe1b3b5ce4bae7c52ddb22b9dde1254a7ee/pkg/kube/kubecache/meta/cni/ovn_kubernetes.go#L54

with

```
cannot read annotation k8s.ovn.org/node-subnets: json: cannot unmarshal array into Go value of type string
```

As far as I see that subnets structure can have different forms. See:

https://github.com/ovn-kubernetes/ovn-kubernetes/blob/5d56a53df520a085e629cdc71be092afed9c3f0f/go-controller/pkg/util/subnet_annotations.go#L15-L39

I also found https://techdocs.broadcom.com/us/en/vmware-security-load-balancing/avi-load-balancer/avi-kubernetes-operator/1-13/avi-kubernetes-operator-guide-1-13/install-avi-kubernetes-operator/cni-support/ovn-kubernetes-cni-plugin-support-on-openshift.html

Side note:
Maybe it would be good to log the input that failed to parse?
Then it should be easier to fix the bug when even just the error message would be provided.

pellared:
I can work on it, but no sooner than in January.

PR #1062, opened 2026-01-08
refactor(cni): indent error flow in findOvnMp0IP

pellared:
Follow https://go.dev/wiki/CodeReviewComments#indent-error-flow

- Reduce nesting
- Return early
- Error checking first

PR #1063, opened 2026-01-09
fix(cni): change log level to error for failed OVN mp0 IP indexing

pellared:
Change log level from INFO to ERROR for failed OVN mp0 IP indexing.

Add a comment from where the fix for https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1030 comes from.


pkg/kube/kubecache/meta/cni/ovn_kubernetes.go
```diff
@@ -40,8 +40,8 @@ func AddOvnIPs(ips []string, node *v1.Node) []string {
 	// (no IP / error returned when not using ovn-k)
 	ip, err := findOvnMp0IP(node.Annotations)
 	if err != nil {
-		// Log the error as Info, do not block other ips indexing
-		slog.Info("failed to index OVN mp0 IP", "error", err)
+		// Log the error and do not block other ips indexing
+		slog.Error("failed to index OVN mp0 IP", "error", err)
```
pellared, line 44: I was also considering

```suggestion
		slog.Warn("failed to index OVN mp0 IP", "error", err)
```

but I think that if this fails then it means that the code is "outdated" or there is a bug

mariomac (approved):

pkg/kube/kubecache/meta/cni/ovn_kubernetes.go
grcevski: I think this is a good change.

MrAlias:
@pellared after merging https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1062 there are now conflicts. Sorry for the trouble. Can you resolve the merge conflicts?

PR #1066, opened 2026-01-09
test: initial coverage of attacher.go

skl:
I noticed `pkg/appolly/discover/attacher.go` had no unit test coverage, so added a few test cases.

### DIFF
--- a/pkg/appolly/discover/attacher_test.go
+++ b/pkg/appolly/discover/attacher_test.go
@@ -5,7 +5,6 @@ package discover
 
 import (
 	"context"
-	"os"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/appolly/discover/attacher_test.go
```diff
@@ -0,0 +1,381 @@
...
+
+				// Verify all fields are initialized
+				assert.NotNil(t, ta.log)
+				assert.NotNil(t, ta.existingTracers)
+				assert.NotNil(t, ta.nodeInjector)
+				assert.Equal(t, os.Getpid(), ta.obiPID)
+				assert.NotNil(t, ta.processInstances)
+				assert.NotNil(t, ta.processAgeFunc)
+				assert.NotNil(t, ta.EbpfEventContext.CommonPIDsFilter)
+				assert.NotNil(t, ta.routeHarvester)
```
mariomac: I'm not usually in favor of unit tests that just check that error is not nil or that private fields are not nil. Despite they increase the coverage data, they might give the false impression of that component being properly tested at a behavioral level.

skl: @mariomac thanks for the review! I removed a bunch of checks and test cases like this in 25349b58a, so it still provides basic coverage without being too specific.

I didn't really think to check the online coverage in codecov, I was just looking locally in the repo, I'll check online next time 👍

PR #1055, opened 2026-01-06
feat(mqtt): PUBLISH parser

skl:
This PR implements a parser for MQTT publish packets. Tested with real-world PUBLISH packet data using the test setup in #937.

Follow-up to:

- #986

This package is not yet in use. Actual integration will follow in subsequent PRs, along with more packet type parsers.

Relevant MQTT specifications:

- https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718037
- https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901100

PR #1070, opened 2026-01-09
feat(mqtt): SUBSCRIBE parser

skl:
This PR implements a parser for MQTT SUBSCRIBE packets. Tested with real-world SUBSCRIBE packet data using the test setup in #937, added a `/mqtt/subscribe` endpoint to the test.

Follow-up to:

- #1055

This package is not yet in use. Actual integration will follow in subsequent PRs.

Relevant MQTT specifications for SUBSCRIBE packets:

- https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718063
- https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901161

N.B. MQTT protocol version is only specified in CONNECT packets, but as OBI is stateless it may be that SUBSCRIBE is the first MQTT packet we see, so I added some logic to try to detect the protocol version in this packet parser.

### DIFF
--- a/internal/test/integration/components/pythonmqtt/pythonmqtt.py
+++ b/internal/test/integration/components/pythonmqtt/pythonmqtt.py
@@ -6,7 +6,8 @@ import time
 import logging
 from datetime import datetime
 from threading import Lock
-from fastapi import FastAPI, HTTPException
+from fastapi import FastAPI, HTTPException, Query
+from typing import Optional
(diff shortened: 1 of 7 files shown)


grcevski (approved):
LGTM!

PR #1073, opened 2026-01-09
Java agent cleanup

grcevski:
This PR makes small improvements to the OBI java agent:

- I added a check to prevent the agent to be loaded (manually by using -javaagent:...) on non Linux systems. Since the ioctl API is Linux only, this will cause unexpected behaviour.
- I learned that you can pass data from a ByteBuddy method enter to exit with the `@Advice.Enter` annotation, so I changed to using that approach instead of the thread locals I was using.
- The thread pool instrumentations I added (the ForkJoin specifically) are used in the concurrent LRU implementation we are using (Caffeine), so I made sure we are not instrumenting those thread pools.

## Checklist

- [X] If this enhances / fixes / changes a core feature, I have updated the [features documentation](../devdocs/features.md)


pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/instrumentations/SSLEngineInst.java
```diff
@@ -276,6 +268,7 @@ public static void wrap(
         @Advice.Argument(1) final ByteBuffer dst,
         @Advice.Return SSLEngineResult result) {
       if (src == null || dst == null) {
+        SSLStorage.unencrypted.remove();
```
grcevski, line 271: I caught two missing cleanup on return cases.

skl (approved):
lgtm

PR #1086, opened 2026-01-13
Experiment: replacing informers by watch API

mariomac:
This is just a vibe-coded PoC. Informers have a big impact in the memory of each K8s cache instance. This PoC replaces them by the Watch API.

This is just a test to see if it passes all the tests, and a starting point for human review and later optimization/improvements.

### DIFF
--- a/pkg/kube/kubecache/meta/informers.go
+++ b/pkg/kube/kubecache/meta/informers.go
@@ -8,8 +8,6 @@ import (
 	"log/slog"
 	"slices"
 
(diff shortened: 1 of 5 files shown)


mariomac:
It worked, but after testing it internally I haven't found any memory or CPU improvement.

Digging more into the code, there is little difference between this implementation and the informers. It's like it just replicated the informers implementation, so we still have a lot of intermediate/unnecessary JSON parsing/transformation/store steps that we could avoid.

PR #1088, opened 2026-01-14
feat: service name lookup

theSuess:
This PR adds a way to perform service name lookups. By enabling this attribute, the server port is looked up against the list of well known privileged ports (e.g 53 -> dns, 80 -> http, etc.). While it technically doesn't add any new data, this results in more legible queries.

I considered multiple implementation strategies but decided on generating the lookup code based on `/etc/services`. This has the following advantages:
* no runtime dependencies
  * allows running obi in environments without `/etc/services` present
* no encoding required
  * by generating code, we don't need to build the lookup table at runtime or deserialize parse anything

I've only included privileged ports (<= 1024) to cover the most used protocols without exploding the binary size. With this PR, it's a 20Ki increase on my machine.

If the runtime dependency on `/etc/services` is acceptable, I'm happy to rework this PR to use that file instead.


mariomac (commented):
Amazing @theSuess !! 

One question: can we assume that what's inside `/etc/services` is somewhat standard so the generated port->service mapping file is portable?

PR #1089, opened 2026-01-14
Add pgx support

marctc:
Currently OBI supports https://github.com/lib/pq for postgres which works on top of `database/sql`. However, `pq` seems to be unmaintained and the most popular driver postgres is https://github.com/jackc/pgx (with 13k stars on github). While this drivers can also be use with `database/sql` the recommended way is to use directly their own API.

This PR adds inital support for pgx for operations such as `QueryRow` and `Query`. Implementation works very similar as `queryDC` but connection info pointer is the struct of the function and not a parameter.

In a follow-up PR I will add support for capturing database name and `Exec`.

### DIFF
--- a/internal/test/oats/sql/yaml/oats_sql_go_pgx.yaml
+++ b/internal/test/oats/sql/yaml/oats_sql_go_pgx.yaml
@@ -10,7 +10,7 @@ interval: 500ms
 expected:
   traces:
     # Successful query
-    - traceql: '{ .db.operation.name = "SELECT" && .db.system.name = "other_sql"}'
+    - traceql: '{ name = "SELECT accounting.contacts" }'
       spans:
         - name: 'SELECT accounting.contacts'
           attributes:
@@ -23,6 +23,8 @@ expected:
         - name: 'SELECT nonexistent_table'
           attributes:
             db.operation.name: SELECT
+            db.collection.name: nonexistent_table
+            db.system.name: other_sql
   metrics:
     - promql: 'db_client_operation_duration_seconds_sum{db_operation_name="SELECT", db_system_name="other_sql"}'
       value: "> 0"


bpf/gotracer/go_sql.c
```diff
@@ -272,6 +286,57 @@ int obi_uprobe_queryReturn(struct pt_regs *ctx) {
     return 0;
 }
 
+SEC("uprobe/pgx_Query_return")
+int obi_uprobe_pgx_Query_return(struct pt_regs *ctx) {
+    bpf_dbg_printk("=== uprobe/pgx_Query_return === ");
```
marctc, line 291: maybe we want to refactor this to have a common function between probes?

grcevski: yeah, how different is this code now from the return probe on sql.Query? If it's the same or mostly the same, we can refactor the code to be reused. But if there's difference in a few places, it's OK to be left as-is.

grcevski (approved):
LGTM! this looks like a great start. You mentioned the regular SQL package can be used, how hard would it be to add support for that? I guess in terms of extracting the host name.

Also, to confirm this PR doesn't handle the host name extraction yet. Is that correct?

bpf/gotracer/go_sql.c
marctc: There are mostly  the same but mysql returns   `(*Rows, error)`  and pgx `(Rows, error)`

grcevski: I think then we should be able to combine them, we don't care about the rows part in our code, correct?

marctc: So i guess `trace->status = (resp_ptr == NULL);` has to be changed here

marctc: or refactor both to check err instead

grcevski: yeah, so we can extract the code in a static __always_inline and call it with the status. How do we tell the status with pgx? This Rows structure, how many fields does it have? Maybe we should instead look at the error field for being non-null.

I didn't realise that the return struct is not a pointer, then the way you set the status will not work. Maybe for future reference paste the full signature of the function. 

To be able to determine the status correctly, you need to figure out if `err` is in a register or somewhere on the stack, which would depend on how big Rows is as a struct. Given this, it's probably worthwhile adding a test with broken SQL.

grcevski: Checking err when the Rows is a pointer is very simple, it's just the next register. But when they return a full struct, each field of the struct will take a register, so which register is err in will be different if you return a *Rows or Rows. If Rows has many fields, it will eventually overflow the registers used for the linkage and err will be stored on the stack.

marctc: Yeah, it only works for the happy path. I noticed the discrepancy but I forgot to address it, my bad. Will add tests for the failure mode

marctc: `Rows` it's an interface here, so we can just check `GO_PARAM3`, right?

grcevski: Ah cool, then yeah I think so!

grcevski: So it would be GO_PARAM2 in the normal case for `err` and GO_PARAM3 in this API. If you extract the function and call it, you can pass the err value from GO_PARAM2 or GO_PARAM3 in it.

marctc: Done, let me add the refactor in a separete PR and I will also add a test for mysql

PR #1101, opened 2026-01-15
Refactor query return handler for Go SQL probes

marctc:
This PR refactors the duplicated probe function for Go SQL added in #1089 
Now Mysql and PGX use same return method for query operations, having as parameter
the `err` return to set `status` of the SQL operation.

I also added an extra OATS test for mysql for the failure mode.


mariomac (approved):
Looks good to me!

PR #1104, opened 2026-01-15
[draft] Fix memory leak

mariomac:
Fixes an issue when OBI instrumented many services, OBI was consuming 1GB of memory due to unnecessary invocations to "NewExecutable" function.

image

⚠️ Do not merge!

I've been trying some manual tests and where we should see 3 instrumented instances, we only see 1:

image

TODO:

- [ ] add a test that verifies that multiple instances of the same executable are instrumented


pkg/appolly/discover/attacher.go
```diff
@@ -316,8 +317,11 @@ func (ta *traceAttacher) reuseTracer(tracer *ebpf.ProcessTracer, ie *ebpf.Instru
 		return false
 	}
 
-	if err := tracer.NewExecutable(exe, ie); err != nil {
-		ta.log.Debug("Failed to attach uprobes for new executable", "pid", ie.FileInfo.Pid, "error", err)
+	// if this executable is already instrumented, avoid instrumenting it again!
```
grcevski, line 328: I think you might need to call NewExecutableInstance?

mariomac:
Independently of the correctness of the first provided solution, it seems that it does not have effect in the `heap` profile for the `inuse_space` metric:

image

PR #1106, opened 2026-01-15
Capture hostname for Go  pgx queries

marctc:
This PR adds support to capture the hostname of query operations using postgres pgx driver.
It works on similar way as MySQL works which is deferrencing conection struct untila arrive to
`Hostname` field of `pgconn.Config`.


bpf/gotracer/go_sql.c
```diff
@@ -162,6 +200,14 @@ extract_sql_hostname(sql_request_trace_t *trace, u64 driver_conn_ptr, void *goro
         return;
     }
 
+    if (conn_type == SQL_CONN_TYPE_PGX) {
```
marctc, line 203: I had to this check because i couldn't do the trick of mysql of checking types

grcevski: This works even better IMO, since we know what we are dealing with exactly from the new uprobe.

PR #1109, opened 2026-01-16
fix: use stable /etc/services reference from OpenBSD

theSuess:
Turns out, the `/etc/services` file varies widely between distributions - even in the port range <= 1024.

This PR changes the source from the local `/etc/services` to the services database shipped by OpenBSD. It includes the most popular services, doesn't include obsolete and old services, and is permissively licensed. Ideally I'd have liked to use the database from the [debian netbase](https://salsa.debian.org/md/netbase/-/blob/master/etc/services?ref_type=heads) package because it contains some newer services, but since it is licensed under GPLv2, choosing the OpenBSD variant seems like the safer choice.

A possible future enhancement would be to load the services database dynamically if desired by the user.

Thanks to @marctc for pointing out the generation issues!


skl:
Should this be used instead or as well as #1110?

PR #1110, opened 2026-01-16
Don't regenerate protocols.go each time docker-generate is invoked

mariomac:


skl (approved):
lgtm but when should it be regenerated? does it need a scheduled action maybe?

cmd/generate-port-lookup/main.go
```diff
@@ -15,11 +16,11 @@ import (
 const fileTemplate = `// Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-//go:generate go run ../../../../../cmd/generate-port-lookup
-//go:generate gofmt -w protocol.go
```
pellared, line 24: Can we add a [package comment with the usage](https://go.dev/doc/comment#cmd)?

For example this could be:

```go
/*
generate-port-lookup reads /etc/services
and generates a Go file (protocol.go by default)
with a lookup table mapping well-known ports
to service names for use by the transport package.

Usage:

	generate-port-lookup [flags]

The flags are:

	-dst filepath
		Destination file path for generated code (default "protocol.go")
*/
package main
```

More: https://go.dev/doc/effective_go#commentary

cmd/generate-port-lookup/main.go
```diff
@@ -68,7 +72,7 @@ func main() {
 		return
 	}
 
-	out, err := os.Create("protocol.go")
+	out, err := os.Create(*dst)
```
pellared: nit: it is more idiomatic to name the variable for the file `f`

```suggestion
	f, err := os.Create(*dst)
```

However, it would be more Unix-style to render the file to standard output instead.
Then the usage in `Makefile` would be:

```
go run cmd/generate-port-lookup/main.go > pkg/internal/netolly/flow/transport/protocol.go
```

cmd/generate-port-lookup/main.go
```diff
@@ -15,11 +16,11 @@ import (
 const fileTemplate = `// Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-//go:generate go run ../../../../../cmd/generate-port-lookup
-//go:generate gofmt -w protocol.go
-
 package transport
 
+// This is an auto-generated file. To regenerate it, run
+// make regenerate-port-lookup
```
pellared: See https://go.dev/blog/generate and https://go.dev/blog/generate on how to make it Go idiomatic

This could be:

```go
// Code generated by generate-port-lookup; DO NOT EDIT.

// Copyright The OpenTelemetry Authors
// SPDX-License-Identifier: Apache-2.0

package transport
```

pellared (commented):
:+1:

PR #1113, opened 2026-01-16
build: upgrade to go 1.25.5

skl:
- Bump build to go 1.25.5 to address [CVE-2025-61729](https://avd.aquasec.com/nvd/2025/cve-2025-61729/)
- I had an issue with the `ghcr.io/open-telemetry/obi-generator:0.2.4` - split into a separate PR on an origin branch so that 0.2.5 could be published #1114


pkg/internal/netolly/flow/transport/protocol.go
```diff
@@ -7,32 +7,33 @@
 package transport
```
marctc, line 7: you can exclude this file as #1109 has fix to avoid generate this file all the time

skl: Restored in 4d64b3517

PR #1115, opened 2026-01-16
Fix features link with absolute path

pinoOgni:
This PR fixes the features doc link by using an absolute path. It also updates the *Checklist* heading to satisfy the linter.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,3 +1,5 @@
 ## Checklist
 
-- [ ] If this enhances / fixes / changes a core feature, I have updated the [features documentation](../devdocs/features.md)
+- [ ] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md)
+
+<!-- markdownlint-disable-file MD041 -->


.github/pull_request_template.md
```diff
@@ -1,3 +1,3 @@
-## Checklist
+# Checklist
```
pellared: I think this is an overkill (see preview), I suggest adding an ignore for the linter at the end of the file

```
<!-- markdownlint-disable-file MD041 -->
```

pinoOgni: Hi @pellared I changed it because there was an error from the linter:
```
Run docker://avtodev/markdown-lint:v1@sha256:6aeedc2f49138ce7a1cd0adffc1b1c0321b841dc2102408967d9301c031949ee
  with:
    args: .github/pull_request_template.md
...
.github/pull_request_template.md:1 MD041/first-line-heading/first-line-h1 First line in file should be a top level heading [Context: "## Checklist"]
```
I can even do a revert, no problem.

pellared: I suggest reverting this line and adding

```
<!-- markdownlint-disable-file MD041 -->
```

at the end of the file

example: https://github.com/goyek/goyek/blob/main/.github/pull_request_template.md?plain=1

PR #1154, opened 2026-01-21
add url.scheme on http server spans

### LINKED ISSUE #1103: url.scheme is missing on http server spans
we dont have the full URL in the http server, but we can know that if its SSL then it is https and if not it is http

itsCheithanya:
## Checklist

fixes #1103


mariomac (approved):
Thank you for your contribution!

grcevski:
@itsCheithanya there seem to be some minor formatting issues with the code, we should be able to merge this PR once those changes are addressed.

PR #1162, opened 2026-01-21
Do not inject TP header when it already exists

### LINKED ISSUE #976: `Traceparent` header may get duplicated with `context_propagation: headers`
Hello! I'm running into the following:

* an instance of nginx (OBI-instrumented, through Beyla) receives a request with a `Traceparent` header already set
* by the time the request _leaves_ nginx, it has two different `Traceparent` headers

Headers obtained by `tcpdump`ing in the container running nginx. Only including relevant sections.

Case 1: external tracing instrumentation injects `traceparent`, OBI injects `Traceparent`.

```
Traceparent: 00-51c4b295117d40a6a3f531e715ce23d7-a24f141cea4f7d1d-01
X-Forwarded-Proto: http
traceparent: 00-51c4b295117d40a6a3f531e715ce23d7-a8e0bed6092f2234-01
```

I initially thought it may be due to case difference, so I reconfigured the upstream instrumentation to use `Traceparent`, matching case. Unfortunately didn't work, leading to:

Case 2: both inject `Traceparent`, duplicating the header:

Incoming into nginx:

```
Traceparent: 00-f6ed85f8071a40eca61f756f991ceb30-b7410955bd7a4a59-01
```

Outgoing _from_ nginx:

```
Traceparent: 00-f6ed85f8071a40eca61f756f991ceb30-a41beb8a36e8dbe3-01
Traceparent: 00-f6ed85f8071a40eca61f756f991ceb30-b7410955bd7a4a59-01
```

I was able to confirm this as the source of some down-stream issues with consuming this data by adding a custom de-duplicating context propagator. Otherwise the resulting traces aren't very usable - the trees are "split" in the middle.

Based on some exploration of the codebase, it seems like `bpf/tpinjector/tpinjector.c` always prepends a `Traceparent` header, without attempting to check if there already exists a header. Is this by design, please? 🙏 I believe the W3C spec states there should only be one Traceparent header present.

rafaelroquetto:
Instead of unconditionally injecting the `Traceparent` HTTP header, only do so when it does not exist (e.g. it could have been already added by external actors) - in which case, we parse it and use it instead.

We also attempt to gracefully handle the case in which a proxy (e.g. `nginx`) forwards the traceparent header unchanged, i.e. instead of generating a new span ID, `nginx` will forwarded the received `Traceparent:` header from the incoming call it received, effectively creating 2 requests with the same traceparent header. The current code detects that case and overrides the latter span ID to a newly generated one.

Fixes #976 

## Checklist

### DIFF
--- a/bpf/tpinjector/tpinjector.c
+++ b/bpf/tpinjector/tpinjector.c
@@ -72,7 +72,7 @@ struct {
 
 typedef struct tailcall_ctx {
     pid_connection_info_t p_conn;
-    tp_info_pid_t parent_tp;
+    tp_info_t parent_tp;
     egress_key_t e_key;
     u8 niter;
     bool has_parent_tp;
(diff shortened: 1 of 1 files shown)


bpf/tpinjector/tpinjector.c
```diff
@@ -665,6 +665,59 @@ static __always_inline void handle_existing_tp_pid(struct sk_msg_md *msg,
...
+    const unsigned char eoh[] = {'\r', '\n', '\r', '\n'};
+    const u32 eoh_niter = TP_SIZE / sizeof(eoh);
+
+    for (u32 i = 0; i < niter; ++i) {
+        if (ptr + TP_SIZE >= e) {
+            break;
+        }
+
+        if (is_traceparent(ptr)) {
+            ptr += TP_TID_PREFIX_SIZE;
```
grcevski: Honestly, I think we don't need to worry about parsing it right, as far as we are concerned, I think we should correct any wrong Traceparents. As long as it's the same size. Different size would complicate things, we'd need to know that we should punch a partial hole, I think that can be a follow-up if we want to handle that.

PR #1164, opened 2026-01-22
test: dockertest instead of docker-compose-go-otel.yml

pellared:
Towards https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1199

First PR  towards https://cloud-native.slack.com/archives/C08P9L4FPKJ/p1768473135948999 for `docker-compose-go-otel.yml`

Use https://pkg.go.dev/github.com/ory/dockertest/v3 instead of `docker compose` (via `go.opentelemetry.io/obi/internal/test/integration/components/docker`).

The test execution time is similar. Yet, it does NOT need the docker binary installed. It gives as also more fine-grained control as it uses [Docker Engine API](https://docs.docker.com/reference/api/engine/).

Most importantly, it allows some code reuse. E.g. we have some helpers to start the obi services instead of having almost the same YAML in all `docker-compose.yml` files.

Another notable change is that now have  `docker build` output in `go test` output instead of having this output in a separate file.

I also think that in future instead of using OTel Collector, Jaeger, Prometheus containers we should use test fakes written in Go. But, one thing at a time :wink:

### DIFF
--- a/internal/test/integration/http_go_otel_test.go
+++ b/internal/test/integration/http_go_otel_test.go
@@ -33,7 +33,7 @@ func setupGoOTelTestServer(t *testing.T, network *dockertest.Network, env []stri
 	t.Helper()
 
 	buildGoOTelTestServerOnce.Do(func() {
-		t.Log("Building Go OpenTelemetry test server image")
+		t.Log("Building Go OpenTelemetry test server image...")
 		buildGoOTelTestServerErr = dockerPool.Client.BuildImage(docker.BuildImageOptions{
 			Name:         "hatest-testserver",
 			ContextDir:   pathRoot,


internal/test/integration/http_go_otel_test.go
```diff
@@ -129,25 +132,53 @@ func testInstrumentationMissing(t *testing.T, route, svcNs string) {
...
+		Repository:   "hatest-testserver",
+		Tag:          "latest",
+		Name:         fmt.Sprintf("testserver-otel-test-%d", time.Now().UnixNano()),
+		Networks:     []*dockertest.Network{network},
+		ExposedPorts: []string{"8080/tcp"},
+		PortBindings: map[docker.Port][]docker.PortBinding{
+			"8080/tcp": {{HostIP: "localhost", HostPort: "8080"}},
+		},
+	})
+	require.NoError(t, err, "could not start test server container")
```
pellared: I created https://github.com/ory/dockertest/issues/629 so maybe this could be simplified in future.

internal/test/integration/http_go_otel_test.go
```diff
@@ -170,65 +201,122 @@ func otelWaitForTestComponents(t *testing.T, url, subpath string) {
...
+		},
+		ExposedPorts: []string{"8080/tcp"},
+		PortBindings: map[docker.Port][]docker.PortBinding{
+			"8080/tcp": {{HostIP: "localhost", HostPort: "8080"}},
+		},
+	})
+	require.NoError(t, err, "could not start test server container")
+	t.Cleanup(func() {
+		require.NoError(t, pool.Purge(testserver), "could not remove test server container")
+	})
```
pellared: This could be extracted into a function and struct, similar to `instrumentWithOBI` and `obiConfig`. However, I don’t think that would bring much benefit. In fact, it might make the test more obscure. Keeping the container setup here reduces the number of abstractions to reason about. Previously, understanding the test required jumping between instrumented application code, `docker-compose.yml`, and the test itself. Now, it’s limited to just the instrumented application code and the test.

internal/test/integration/dockerutil_test.go
```diff
@@ -0,0 +1,204 @@
...
+			"8999/tcp": {{HostIP: "localhost", HostPort: "8999"}},
+		},
+	}, func(hc *docker.HostConfig) {
+		hc.PidMode = "container:" + resource.Container.ID
+	})
+	require.NoError(t, err, "could not start OBI container")
+	t.Cleanup(func() {
+		require.NoError(t, pool.Purge(obi), "could not remove OBI container")
+	})
+}
```
pellared: I think in future this could be "donated" to `pkg/test/integration`.

internal/test/integration/http_go_otel_test.go
```diff
@@ -129,25 +132,53 @@ func testInstrumentationMissing(t *testing.T, route, svcNs string) {
...
-	compose, err := docker.ComposeSuite("docker-compose-go-otel.yml", path.Join(pathOutput, "test-suite-go-otel.log"))
-	require.NoError(t, err)
-
-	// we are going to setup discovery directly in the configuration file
-	compose.Env = append(compose.Env, `OTEL_EBPF_EXECUTABLE_PATH=`, `OTEL_EBPF_OPEN_PORT=8080`, `APP_OTEL_ENDPOINT=http://localhost:1111`)
-	lockdown := KernelLockdownMode()
+	pool, network := setupDockertest(t)
+	setupContainerPrometheus(t, pool, network)
+	setupContainerJaeger(t, pool, network)
+	setupContainerCollector(t, pool, network)
```
skl: These could maybe be parallelised

internal/test/integration/http_go_otel_test.go
```diff
@@ -129,25 +132,53 @@ func testInstrumentationMissing(t *testing.T, route, svcNs string) {
...
 	}
-
-	require.NoError(t, compose.Up())
+	if !KernelLockdownMode() {
+		obiConfig.SecurityConfigSuffix = "_none"
+	}
+	instrumentWithOBI(t, pool, network, testserver, obiConfig)
 
 	t.Run("Go RED metrics: http service instrumented with OTel", func(t *testing.T) {
 		waitForTestComponents(t, "http://localhost:8080")
```
skl, line 210: Docker compose had `depends_on` for startup ordering, I suppose you could reuse  `waitForTestComponents` to wait for dependent containers to start before starting the OBI container, or some other approach?

internal/test/integration/http_go_otel_test.go
pellared: What do you mean by parallelized here?
That all tests run in parallel? Then we will need to expose different ports.
That containers are shared across tests? Then I think the results of one test can affect others (e.g., we may expect no metrics in prometheus but they may be there because of other test).

internal/test/integration/http_go_otel_test.go
pellared: It does not seem this is necessary.

internal/test/integration/http_go_otel_test.go
skl: I mean that all the containers within a single test could be started in parallel (as I presume compose would do), or are these functions non-blocking already?

internal/test/integration/http_go_otel_test.go
skl: Great if it's not needed 👍 I mentioned it because sometimes in CI we see quite variable performance, and containers can take longer than expected to start up.

internal/test/integration/http_go_otel_test.go
pellared: AFAIK there are blocking. Maybe this is where the delay comes from. Nice idea

> I just took a guess and thought it might be due to the containers being started sequentially, but that is just a guess.

I think you are right👍

MrAlias (approved):

PR #1166, opened 2026-01-22
Slightly improve HTTP2grpc memory generation

mariomac:
Replaced `bytes.NewBuffer` by a zero-copy alternative, as`byteFramer` function reported to generate a lot of memory in some super-high-load scenarios:
image

After the changes, there is a slight improvement in terms of pod memory:

image

The Garbage collector also shows some slight improvement in terms of cpu (-3%) under low-load scenarios


mmat11 (approved):
🎉

PR #1171, opened 2026-01-22
Fix java agent for JDK8 part 2

grcevski:
With the reported slack issue related to Apollo Config I was able to uncover and fix couple of issues with the Java agent:

1. To interface with JNI I used the JNA library, which does nice packaging and minimizes boilerplate, however, with whatever SpringBoot 2.0 in Apollo Config 1.6 does with class loaders, we just couldn't properly load the native code. I switched the implementation to use pure JNI and it resolved that issue. The benefit of direct JNI is that it's easier to reason with and reduces the dependencies of the OBI java agent, for extra work around the build process.
2. Some of the bytebuffers used by the project had side-effects when calling flip(). According to the base class spec, all it should do is set the position and the limit, but it caused issues with the internal state of the buffers we were reading. Instead of doing that I changed all code to use ByteBuffer.duplicate to get a copy of the internals of the original bytebuffer, without copying the content. These duplicates can be messed with as we want and it's easier to understand the code.
3. The application exposed one more issue with our sync SSL instrumentation. We require that getInputStream() and getOutputStream() are called to install our proxy streams that capture the traffic. However, this only works for webservers that keep on establishing new streams. For things like databases, where the stream is only established once, we were not capturing any traffic. I added instrumentation for the most common class SSLSocketImpl streams, the rest will use the old approach.
4. I cleaned up some verbosity in the code.

I'll follow-up with more tests around TLS with databases and Kafka in another PR.

### DIFF
--- a/pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/instrumentations/JavaExecutorInst.java
+++ b/pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/instrumentations/JavaExecutorInst.java
@@ -102,7 +102,7 @@ public class JavaExecutorInst {
 
   @SuppressWarnings("unused")
   public static final class SetExecuteRunnableStateAdvice {
-    @Advice.OnMethodEnter // (suppress = Throwable.class)
+    @Advice.OnMethodEnter(suppress = Throwable.class)
     public static void enterJobSubmit(
         @Advice.Argument(value = 0, readOnly = false) Runnable task, @Advice.Origin String method) {
       long threadId = Agent.NativeLib.gettid();


pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/instrumentations/SocketChannelInst.java
```diff
@@ -71,20 +100,24 @@ public static AgentBuilder.Transformer transformer() {
...
+    @Advice.OnMethodExit(suppress = Throwable.class)
     public static void write(
         @Advice.Argument(0) final ByteBuffer src,
         @Advice.Enter int savedPos,
-        @Advice.FieldValue("localAddress") SocketAddress localSocket,
-        @Advice.FieldValue("remoteAddress") SocketAddress remoteSocket) {
+        @Advice.This Object thiz) {
+
+      // Get address fields - try multiple names for JDK8/11 compatibility
+      Object localSocket = getFieldValue(thiz, "localAddress", "localAddr", "local");
```
skl: There are other occurrences of the `@Advice.FieldValue("localAddress")` pattern on a few other methods in this file - do they also need to be updated with the same compat fix?

grcevski: Ah this is some test code I need to remove, weird, I thought I cleaned it up. I'll remove it. I check now the conditions for localAddress that must exist in the object.

PR #1184, opened 2026-01-26
bpf: add shared traces_ctx_v1 map with initial server traces info

mmat11:
This PR:

- fixes a bug in logenricher which adds/deletes the same pid twice and causes unwanted warn logs
- adds a base shared obi context info map which can be used by external processes (eg. the profiler)
- use the shared context map in logenricher -- this will make fetching context by pid easier to test in the future and consistent with external processes

For this to work properly, object files need to be available in the external processes' imported code as the map spec is generated via bpf2go and replaced at runtime.
The context map should be loaded by the first process who starts: if the profiler starts first, it should create the map and OBI will load the pinned map in bpffs - if OBI starts first, it should create the map and the profiler will just load it.
Versioning is not handled at the moment, the assumption is that new changes in the map structure will happen in a forward-compatible way, so new fields should be added at the end of the struct and deprecated ones shouldn't be deleted (I doubt this will happen as we only have trace and span IDs for now).

I pushed a PoC here which uses the otel profiler as an external reader and tested it by printing the span ID while running a python server in the background: https://github.com/coralogix/opentelemetry-ebpf-profiler/pull/1/files


bpf/shared/obi_ctx.h
```diff
@@ -0,0 +1,38 @@
...
+
+#pragma once
+
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_helpers.h>
+#include <bpfcore/bpf_builtins.h>
+
+#include <common/tp_info.h>
+
+typedef struct obi_ctx_info {
```
rafaelroquetto, line 12: here's a suggestion: you could publish two maps, namely `obi_ctx_meta` as a single entry _array_ providing metadat information including versioning (to make it truly forward compatible) and `obi_ctx_trace_info` which is what you currently call `obi_ctx`.

Any interested part can open `obi_ctx_meta` and from there infer what other maps may exist alongside their layouts.

internal/test/integration/docker-compose-log-enricher.yml
```diff
@@ -18,6 +18,7 @@ services:
       - /sys/kernel/security:/sys/kernel/security:ro
       - ../../../testoutput:/coverage
       - ../../../testoutput/run-python:/var/run/beyla
+      - /sys/fs/bpf:/sys/fs/bpf
```
rafaelroquetto, line 21: in older versions of Beyla we had code to detect the existence of `bpffs` (and tried to mount it). We removed the `bpffs` requirement because it had issues in a few settings involving AppArmour.

As a matter of fact, I recommend having a look at this PR: https://github.com/grafana/beyla/pull/1251 for some insights, in particular:

> the AppArmour policy restricts the use of `mount` and the access to `/sys/fs/` directories. Beyla uses the BPF Linux file system to store pinned BPF maps, for communication among the different BPF programs. For this reason, Beyla either needs to `mount` a BPF file system, or write to `/sys/fs/bpf`, which are both restricted.

[(see here)](https://github.com/grafana/beyla/pull/1251/changes#diff-6336b3c645a0790a22759fb9e1f7d64a414fb51875cfbea823104ec19b4cc4bb)

We may need to document this thoroughly again.

pkg/ebpf/tracer_linux.go
```diff
@@ -72,26 +73,38 @@ func resolveMaps(eventContext *common.EBPFEventContext, spec *ebpf.CollectionSpe
...
-				return nil, fmt.Errorf("failed to load shared map: %w", err)
+				eventContext.EBPFMaps[k] = internalMap
+				runtime.SetFinalizer(internalMap, (*ebpf.Map).Close)
 			}
 
-			eventContext.EBPFMaps[k] = internalMap
-			runtime.SetFinalizer(internalMap, (*ebpf.Map).Close)
+			collOpts.MapReplacements[k] = internalMap
+		case ebpf.PinByName:
+			if k == shared.ObiCtxMapName {
```
rafaelroquetto: why is this necessary? Could we not simply rely on the `LoadAndAssign()` default behaviour for pinned maps in which it will create the pinned map if it does not yet exist?

bpf/shared/obi_ctx.h
mmat11: the issue with this approach is that a consumer - given the fact it needs a reference to the ebpf map when loading its programs - will need to load the map beforehand if it doesn't exist so it's not always obi instantiating it for the first time. If we do this we will need to support different versions of the same map in OBI (eg. obi_ctx_v1, obi_ctx_v2, etc.) and load/update all of them, which is something I'd like to avoid for now

correct me if I got your comment wrong or if I'm overthinking it

internal/test/integration/docker-compose-log-enricher.yml
mmat11: > it had issues in a few settings involving AppArmour

damn :/ 

> We may need to document this thoroughly again.

good suggestion, I'll go through it

pkg/ebpf/tracer_linux.go
mmat11: you need to specify the pin path in case of pinned maps so I wrapped it in a function for simplicity, which is also an "api" for external consumers

bpf/shared/obi_ctx.h
```diff
@@ -0,0 +1,38 @@
...
+    unsigned char trace_id[16];
+    unsigned char span_id[8];
+} obi_ctx_info_t;
+
+struct {
```
florianl, line 19: Maybe we should start a OTLP specification about this map and its pinning location, so it can be shared and communicated more easily.

pkg/ebpf/tracer_linux.go
rafaelroquetto: I see what you mean, yet it may be worth checking that Beyla PR again - you will see it did things a bit differently by allowing the bpffs path was configurable (and not hardcoded to /sys/...), and managed - the following comment sheds some light on why:

```
       # We need to set beyla container as unconfined so it is able to write
        # the BPF file system.
        # Otherwise, we can have error like:
        # "failed to create server failed to create or mount folder for pinning bpf maps: can't mount BPF filesystem: permission denied""
        # 
        # Default rules can be found here: https://github.com/moby/moby/blob/master/profiles/apparmor/template.go
        # Both 'mount' and access to /sys/fs/bpf are denied by default.
        #
        # Instead of 'unconfined', you can define a more refined policy which allows Beyla to use 'mount' 
```

If we are going to be supporting pinned maps again, I think we should provide a mechanism that is not necessarily tied to the ctx map. In fact, it's just another map and there should be nothing special to it, so it should be loaded like any other map. If you look at that Beyla PR, you will see that all it did was:
```
					err = spec.LoadAndAssign(p.BpfObjects(), &ebpf.CollectionOptions{
						Programs: ebpf.ProgramOptions{LogSize: 640 * 1024},
						Maps: ebpf.MapOptions{
							PinPath: pt.PinPath,
						}})
```
for _any_ map.

In OBI's case, it's just a matter of:

```
func (pt *ProcessTracer) loadAndAssign(eventContext *common.EBPFEventContext, p Tracer) error {
	spec, err := pt.loadSpec(p)
	if err != nil {
		return err
	}

	collOpts, err := resolveMaps(eventContext, spec)
	if err != nil {
		return err
	}

	collOpts.Programs = ebpf.ProgramOptions{LogSizeStart: 640 * 1024}
	collOpts.Maps.PinPath = pt.PinPath // <----- THIS

	return spec.LoadAndAssign(p.BpfObjects(), collOpts)
}
```

It's then up to the user to decide what the pinpath is, and point everyone interested in pinned maps to it (including obi ctx) - we can of course provide a sane default so that the user can reason about it.

bpf/shared/obi_ctx.h
rafaelroquetto: No, you are right and it makes sense - and actually, I believe something like `obi_ctx_v1` makes total sense - would that fit?

internal/test/integration/docker-compose-log-enricher.yml
mmat11: wouldn't adding the `container.apparmor.security.beta.kubernetes.io/` annotation work for our use case? I don't understand why the path name should be configurable

bpf/shared/obi_ctx.h
mmat11: updated the code with latest OTEP changes which include versioning in the name

internal/test/integration/docker-compose-log-enricher.yml
mmat11: made it configurable

pkg/ebpf/tracer_linux.go
mmat11: done 👍🏻

bpf/logenricher/types.h
```diff
@@ -28,11 +28,11 @@ enum {
 };
 
 typedef struct log_event {
-    tp_info_pid_t pid_tp;
+    obi_ctx_info_t ctx;
+    u8 _pad[8];
```
rafaelroquetto: what happens to the padding if you move `ctx` after `fd`?

bpf/shared/obi_ctx.h
```diff
@@ -0,0 +1,40 @@
...
+#pragma once
+
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_helpers.h>
+#include <bpfcore/bpf_builtins.h>
+
+#include <common/tp_info.h>
+
+typedef struct obi_ctx_info {
+    unsigned char trace_id[16];
```
rafaelroquetto: are you purposedly using number literals here instead of `TRACE_ID_SIZE_BYTES` for instance, to prevent any further dependencies in terms of headers?

bpf/shared/obi_ctx.h
```diff
@@ -0,0 +1,40 @@
...
+} otel_traces_ctx_v1 SEC(".maps");
+
+static __always_inline obi_ctx_info_t *obi_ctx__get(const u64 pid_tgid) {
+    return bpf_map_lookup_elem(&otel_traces_ctx_v1, &pid_tgid);
+}
+
+static __always_inline long obi_ctx__set(const u64 pid_tgid, const tp_info_t *info) {
+    obi_ctx_info_t obi_info = {};
+    bpf_memcpy(obi_info.trace_id, info->trace_id, 16);
+    bpf_memcpy(obi_info.span_id, info->span_id, 8);
```
rafaelroquetto:
```suggestion
    bpf_memcpy(obi_info.trace_id, info->trace_id, sizeof(obi_info.trace_id));
    bpf_memcpy(obi_info.span_id, info->span_id, sizeof(obi_info.span_id));
```

pkg/internal/ebpf/logenricher/logenricher.go
```diff
@@ -227,27 +228,30 @@ func (p *Tracer) AllowPID(pid, ns uint32, _ *svc.Attrs) {
 	}
 
 	for _, nsPid := range nsPids {
-		if err := p.addPID(p.pidKey(ns, nsPid)); err != nil {
+		if pid == nsPid {
```
rafaelroquetto, line 231: are these related?

pkg/shared/obi_ctx.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("loading %s spec: %w", OtelTracesCtxV1MapName, err)
+	}
+
+	var maps BpfMaps
+	err = spec.LoadAndAssign(&maps, &ebpf.CollectionOptions{
+		Maps: ebpf.MapOptions{
+			PinPath: bpfFsPath,
+			LoadPinOptions: ebpf.LoadPinOptions{
+				ReadOnly: true, // BPF_F_RDONLY
```
rafaelroquetto: genuine question: I thought this made the map truly read-only to ebpf helpers - is that not the case? can one do `bpf_map_update_elem` on a read-only map?

pkg/shared/obi_ctx.go
```diff
@@ -0,0 +1,54 @@
...
+	})
+	if err != nil {
+		return nil, fmt.Errorf("creating new %s map: %w", OtelTracesCtxV1MapName, err)
+	}
+
+	return maps.OtelTracesCtxV1, nil
+}
+
+func TeardownCtxMap(bpfFsPath string) error {
+	m, err := LoadOrCreateCtxMap(bpfFsPath)
```
rafaelroquetto: I think here you can use https://pkg.go.dev/github.com/cilium/ebpf#LoadPinnedMap which is more lightweight.

pkg/ebpf/tracer_linux.go
```diff
@@ -72,49 +74,66 @@ func resolveMaps(eventContext *common.EBPFEventContext, spec *ebpf.CollectionSpe
...
-				return nil, fmt.Errorf("failed to load shared map: %w", err)
+				eventContext.EBPFMaps[k] = internalMap
+				runtime.SetFinalizer(internalMap, (*ebpf.Map).Close)
 			}
 
-			eventContext.EBPFMaps[k] = internalMap
-			runtime.SetFinalizer(internalMap, (*ebpf.Map).Close)
+			collOpts.MapReplacements[k] = internalMap
+		case ebpf.PinByName:
+			if k == shared.OtelTracesCtxV1MapName {
```
rafaelroquetto: Why not simply pass `bpffFsPath` directly to LoadAndAssign here: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/b97b052a428d266a9c03545ce06787ad1ccdccbe/pkg/ebpf/tracer_linux.go#L196

I.e. it's just another line

```
collOpts.Maps = ebpf.MapOptions{PinPath: bpfFsPath}
```

For the readonly flag, I think you can just declare it directly on the map definition with:

```
__uint(map_flags, BPF_F_RDONLY);
```

although the documentation states that "Setting this flag will make it so the map can only be read via the [syscall](https://docs.ebpf.io/linux/syscall/) interface, but not written to." - I am not sure if it applies to bpf helpers or just the syscall iface.

---

Doing the above enables us to fully support pinned maps again, should push comes to shove, instead of artificially constraining it to the ctx map. Any tracer can simply include the map header and be happy. `LoadOrCreateCtxMap` is a mere convenience function/abstraction (it does nothing much differently), it shouldn't be mandatory IMHO.

bpf/shared/obi_ctx.h
mmat11: no, I was just lazy, added the defines

pkg/internal/ebpf/logenricher/logenricher.go
mmat11: I changed the map used in logsenricher, this is just a bugfix (I was adding/removing the same pid twice)

pkg/shared/obi_ctx.go
mmat11: the read-only part should only apply to userspace (see: https://docs.ebpf.io/linux/syscall/BPF_MAP_CREATE/#bpf_f_rdonly)

pkg/shared/obi_ctx.go
mmat11: latest commit still calls that function because it needs to (load the spec and create the path if it doesn't exist)

pkg/ebpf/tracer_linux.go
mmat11: I see what you mean, I'll try to make it work

mmat11: looks much cleaner, and seems to work, thanks!

pkg/ebpf/tracer_linux.go
```diff
@@ -72,26 +74,28 @@ func resolveMaps(eventContext *common.EBPFEventContext, spec *ebpf.CollectionSpe
 	for k, v := range spec.Maps {
 		alignMaxEntriesIfRingBuf(v)
 
-		if v.Pinning != PinInternal {
-			continue
-		}
+		switch v.Pinning {
```
rafaelroquetto: Do these changes do anything different than the original code at this stage? If not, you can just revert it (the less diffs the better for bisecting).

mmat11: not really, I just thought it was clearer since there are now 3 pinning modes, I can revert if you prefer

mmat11: reverted

Issue #1138, opened 2026-01-20
Improve service metadata when not running in Kubernetes


MrAlias:
Currently, OBI is very Kubernetes-centric when it comes to determining service metadata.

- [x] Investigate the other platforms OBI is commonly run on
- [x] Determine strategies to populate metadata for services (compliant with semantic conventions) for each of these platforms.

NimrodAvni78:
GKE Autopilot
https://github.com/open-telemetry/community/issues/3083

MrAlias:
Done! 🎉

PR #1190, opened 2026-01-27
Docker metadata support for app metrics

mariomac:
Parent issue: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1138

Provides a new source of metadata for application-level metrics: docker and docker-compose metadata, so you now can select applications by container name, and application metrics will be decorated with container name and ID.

Also the `instance` and `service_name` values will be now composed according to container metadata.

After this PR, new PRs will follow up:
* A refactor to avoid keep duplicating nodes in several parts of the pipeline, each time we want to add a new metadata source (e.g. cloud metadata or VM metadata)
* Container metadata also in network metrics.
* Documentation

### DIFF
--- a/pkg/internal/docker/docker_api_client.go
+++ b/pkg/internal/docker/docker_api_client.go
@@ -6,6 +6,7 @@ package docker // import "go.opentelemetry.io/obi/pkg/internal/docker"
 import (
 	"context"
 	"log/slog"
+	"maps"
 	"strings"
 	"sync"
 
@@ -15,6 +16,8 @@ import (
 	attr "go.opentelemetry.io/obi/pkg/export/attributes/names"
 	"go.opentelemetry.io/obi/pkg/internal/helpers/container"
 )
+
+const composeServiceLabelKey = "com.docker.compose.service"
 
 func cmlog() *slog.Logger {
 	return slog.With("component", "docker.ContainerStore")
@@ -78,12 +81,14 @@ func (s *ContainerStore) initialize(ctx context.Context) {
 	if info, err := docker.Info(ctx); err != nil {
 		s.log.Debug("failed to get docker info", "error", err)
 		return
-	s.log.Info("Docker info",
-		"driver", info.Driver,
-		"version", info.ServerVersion,
-		"cgroupDriver", info.CgroupDriver,
-		"cgroupVersion", info.CgroupVersion)
-	s.docker = docker
+	} else {
+		s.log.Info("Docker info",
+			"driver", info.Driver,
+			"version", info.ServerVersion,
+			"cgroupDriver", info.CgroupDriver,
+			"cgroupVersion", info.CgroupVersion)
+		s.docker = docker
+	}
 }
 
 func (s *ContainerStore) ContainerInfo(ctx context.Context, pid PID) (ContainerMeta, bool) {
@@ -109,7 +114,7 @@ func (s *ContainerStore) ContainerInfo(ctx context.Context, pid PID) (ContainerM
 
 	composeSvcName := ""
 	if inspectInfo.Config != nil && len(inspectInfo.Config.Labels) > 0 {
-		composeSvcName = inspectInfo.Config.Labels["com.docker.compose.service"]
+		composeSvcName = inspectInfo.Config.Labels[composeServiceLabelKey]
 	}
 
 	return ContainerMeta{
@@ -149,12 +154,14 @@ func (ci *ContainerMeta) DecorateService(s *svc.Attrs) {
 }
 
 func ContainerMetadata[T ~string](dst map[T]string, ci *ContainerMeta, stringer func(attr.Name) T) map[T]string {
-    // Copy map to avoid concurrent read/write on shared Metadata
-    out := make(map[T]string, len(dst)+2)
-    if dst != nil {
-        maps.Copy(out, dst)
-    }
-    out[stringer(attr.ContainerName)] = ci.Name
-    out[stringer(attr.ContainerID)] = ci.ID
-    return out
+	// Copy map to avoid concurrent read/write on shared Metadata
+	var out map[T]string
+	if dst == nil {
+		out = map[T]string{}
+	} else {
+		out = maps.Clone(dst)
+	}
+	out[stringer(attr.ContainerName)] = ci.Name
+	out[stringer(attr.ContainerID)] = ci.ID
+	return out
 }


Makefile
```diff
@@ -479,7 +479,7 @@ notices-update: docker-generate go-notices-update $(TARGET_BPF)
 
 .PHONY: go-notices-update
 go-notices-update: $(GOLICENSES)
-	@$(GOLICENSES) save ./... --save_path=$(NOTICES_DIR) --force
+	@GOOS=$(GOOS) GOARCH=amd64 $(GOLICENSES) save ./... --save_path=$(NOTICES_DIR) --force
```
rafaelroquetto, line 528: did you mean to have this explicit GOARCH?

pkg/appolly/discover/container_updater.go
```diff
@@ -18,6 +18,7 @@ import (
 // ContainerDBUpdaterProvider is a stage in the Process Finder pipeline that will be
 // enabled only if Kubernetes decoration is enabled.
 // It just updates part of the kubernetes database when a new process is discovered.
+// TODO: rename to avoid confusions with Docker-only containers
```
MrAlias, line 21: Do we have an issue to track this work?

pkg/internal/docker/docker_api_client.go
```diff
@@ -0,0 +1,160 @@
...
+	docker, err := client.NewClientWithOpts(
+		client.WithAPIVersionNegotiation(),
+		client.FromEnv,
+	)
+	if err != nil {
+		s.log.Debug("trying to instantiate docker client", "error", err)
+		return
+	}
+	if info, err := docker.Info(ctx); err != nil {
+		s.log.Debug("trying to get get docker info", "error", err)
```
MrAlias:
```suggestion
		s.log.Debug("failed to get docker info", "error", err)
```

pkg/internal/docker/docker_api_client.go
```diff
@@ -0,0 +1,160 @@
...
+		return
+	} else {
+		s.log.Info("Docker info",
+			"driver", info.Driver,
+			"version", info.ServerVersion,
+			"cgroupDriver", info.CgroupDriver,
+			"cgroupVersion", info.CgroupVersion)
+	}
```
MrAlias: The above clause returns, no need for the indentation.

```suggestion
	s.log.Info("Docker info",
		"driver", info.Driver,
		"version", info.ServerVersion,
		"cgroupDriver", info.CgroupDriver,
		"cgroupVersion", info.CgroupVersion)
```

pkg/internal/docker/docker_api_client.go
```diff
@@ -0,0 +1,160 @@
...
+		Name:           strings.Trim(inspectInfo.Name, "/"),
+		ID:             containerID,
+		ComposeService: composeSvcName,
+	}, true
+}
+
+func identity(n attr.Name) attr.Name { return n }
+
+func (ci *ContainerMeta) DecorateService(s *svc.Attrs) {
+	s.Metadata = ContainerMetadata(s.Metadata, ci, identity)
```
MrAlias: Why not inline this?

```suggestion
func (ci *ContainerMeta) DecorateService(s *svc.Attrs) {
	s.Metadata = ContainerMetadata(s.Metadata, ci, func(n attr.Name) attr.Name {
		return n
	})
```

pkg/internal/docker/docker_api_client.go
```diff
@@ -0,0 +1,160 @@
...
+		s.log.Info("Docker info",
+			"driver", info.Driver,
+			"version", info.ServerVersion,
+			"cgroupDriver", info.CgroupDriver,
+			"cgroupVersion", info.CgroupVersion)
+	}
+	s.docker = docker
+}
+
+func (s *ContainerStore) ContainerInfo(ctx context.Context, pid PID) (ContainerMeta, bool) {
```
MrAlias, line 94: Can you document in a method comment what the returned `bool` signifies?

It seems like it is a substitute for an error. Consider just returning the error and having the caller unify any logging it want to do within its own error handling.

pkg/transform/transform_docker.go
```diff
@@ -0,0 +1,124 @@
...
+		// only enable this node if Docker is available, but also
+		// if we aren't running on Kubernetes
+		if ctxInfo.K8sInformer.IsKubeEnabled() ||
+			!ctxInfo.DockerMetadata.IsEnabled(ctx) {
+			return swarm.Bypass(input, output)
+		}
+
+		in := input.Subscribe()
+		containers := ctxInfo.DockerMetadata
+		containerByPID := map[int32]docker.ContainerMeta{}
```
MrAlias, line 98: nit: this is an unneeded type conversion below.

```suggestion
		containerByPID := map[docker.PID]docker.ContainerMeta{}
```

pkg/internal/docker/docker_api_client.go
```diff
@@ -0,0 +1,160 @@
...
+
+	const abbreviationLength = 12
+	containerID := inspectInfo.ID
+	if len(containerID) > abbreviationLength {
+		containerID = containerID[:abbreviationLength]
+	}
+
+	composeSvcName := ""
+	if inspectInfo.Config != nil && len(inspectInfo.Config.Labels) > 0 {
+		composeSvcName = inspectInfo.Config.Labels["com.docker.compose.service"]
```
MrAlias: This is a magic string constant. Consider defining it as an explicit package level `const` to communicate its purpose.

For example, above:

```
const composeServiceLabelKey = "com.docker.compose.service"
```

then:

```suggestion
		composeSvcName = inspectInfo.Config.Labels[composeServiceLabelKey]
```

MrAlias (approved):
Looks good overall. Minor nits.

Makefile
mariomac: Yeah, otherwise if you run `go-notices-update` on an Darwin/ARM machine, the generated NOTICES folder will differ from the folder generated by the GitHub action and the PR fails.

pkg/export/prom/prom.go
```diff
@@ -1099,6 +1114,9 @@ func labelNamesTargetInfo(kubeEnabled bool, extraMetadataLabelNames []attr.Name)
 	if kubeEnabled {
 		names = appendK8sLabelNames(names)
 	}
+	if dockerEnabled {
```
ajimenez1503, line 1117: I have a question. What is the reason of not having docker decorator and Kube decorator at the same time?

I have seen that both add `service.name`, `service.namespace`, and `service.instance.id`; and that could lead into conflict. 
But they also add other attributes which are independent, like `container.name` and `container.id` or `k8s.*`

Is there a way that we can have both adding their unique attribute and a common `serviceDecoreator` that get the data from Kube or container API.

I am adding the comment in that line because here the docker attributes are being adding only if `dockerEnabled`, but on the decorator we are populating those attribute only if `dockerEnabled` and not `kubeEnabled`. So it seems inconsistent.

Thanks. I may be missing context too...

pkg/transform/transform_docker.go
```diff
@@ -0,0 +1,124 @@
...
+	})
+}
+
+func (dd *dockerEnricher) containerInfo(ctx context.Context, pid docker.PID) (docker.ContainerMeta, bool) {
+	if ci, ok := dd.containerByPID[pid]; ok {
+		return ci, true
+	}
+	ci, ok := dd.docker.ContainerInfo(ctx, pid)
+	if ok {
+		dd.containerByPID[pid] = ci
```
ajimenez1503, line 73: we are adding container info by PID but never deleting when PID is deleted. Can we have here a memory issue? or incorrect docker info for a PID reused?

Can we somehow listen to `case exec.ProcessEventTerminated:` and delete on those case? 

Maybe, it may not be need it :)

pkg/internal/docker/docker_api_client.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+}
+
+func ContainerMetadata[T ~string](dst map[T]string, ci *ContainerMeta, stringer func(attr.Name) T) map[T]string {
+	if dst == nil {
+		dst = map[T]string{}
+	}
+	dst[stringer(attr.ContainerName)] = ci.Name
+	dst[stringer(attr.ContainerID)] = ci.ID
+	return dst
```
ajimenez1503: Same as K8s copy shared Metadata https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/e1f623a32e5c64af3f195e2022c502cc5fdaa382/pkg/transform/k8s.go#L398 , we can do the same here
```suggestion
    // Copy map to avoid concurrent read/write on shared Metadata
    out := make(map[T]string, len(dst)+2)
    if dst != nil {
        maps.Copy(out, dst)
    }
    out[stringer(attr.ContainerName)] = ci.Name
    out[stringer(attr.ContainerID)] = ci.ID
    return out
```

pkg/appolly/discover/container_updater.go
mariomac: I put it into a more general issue: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1263

pkg/export/prom/prom.go
mariomac: Hi @ajimenez1503 actually the container name and container ID are also provided by the Kubernetes decorator, but they got the container name and ID from the Kube API

pkg/transform/transform_docker.go
mariomac: The processEventTerminated case is actually few lines below, wrapped into the `DockerProcessEventDecoratorProvider`, which accesses a shared instance of the container Store and is the component that has visibility about process deletion events.

pkg/internal/docker/docker_api_client.go
mariomac: Nice!

pkg/internal/docker/docker_api_client.go
mariomac: Done!

pkg/internal/docker/docker_api_client.go
mariomac: Sorry I missed this comment and I merged. Will address it into a new PR

pkg/transform/transform_docker.go
mariomac: Sorry I missed this comment and I merged. Will address it into a new PR

PR #1195, opened 2026-01-27
Improve service graph metrics

grcevski:
Service graph metrics typically have a client and server label for the namespace and the otel name, however when running on Kubernetes it's useful to differentiate the Kubernetes namespace name. Customers may bundle the service namespaces much differently than the Kubernetes namespaces, and this additional connectivity can help with debugging issues.

I've also added the cluster name with client and server, although we are limited to in-cluster and cluster to external at the moment.


mariomac (approved):
🚀

PR #1214, opened 2026-01-30
Remove unused headers in ebpf code 

pinoOgni:
This PR removes unused includes and in some cases removes "umbrella" headers with the specific files actually required. 

I was hoping to automate everything but using `clang-tidy` there are a lot of false positives because maybe we include a header that is not used directly but indirectly so I just cleaned up the code a bit.

Moreover, I would like to know what do you think about [IWYU](https://github.com/include-what-you-use/include-what-you-use), I know it can increase the number of lines at the top of a file but it should clarify dependencies and it should reduce preprocessor overhead.

P.s. I ran some tests to check the compilation time, and it didn't really reduce much because most of the compilation time is spent on large headers like `utils.h`, `vmlinux.h`, and so on. To check this, I used [ClangBuildAnalyzer](https://github.com/aras-p/ClangBuildAnalyzer). 
Maybe it's worth looking into where we're wasting time and improving the situation?

### DIFF
--- a/bpf/generictracer/protocol_mysql.h
+++ b/bpf/generictracer/protocol_mysql.h
@@ -184,10 +184,6 @@ static __always_inline u32 data_offset(struct mysql_hdr *hdr) {
                             : k_mysql_hdr_size;
 }
 
-static __always_inline u32 mysql_command_offset(struct mysql_hdr *hdr) {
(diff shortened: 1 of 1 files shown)


rafaelroquetto (approved):
Fantastico!

Issue #1185, opened 2026-01-26
Migrate tests from guara `Eventually` to testify `EventuallyWithT`


MrAlias:
Migrate test code from `github.com/mariomac/guara/pkg/test.Eventually` to `github.com/stretchr/testify`'s plain `github.com/stretchr/testify/require.EventuallyWithT`. This aligns the test suite with widely adopted tooling, improves integration with existing `github.com/stretchr/testify/assert` and `github.com/stretchr/testify/require`, simplifies the contributor experience, and reduces a niche dependency.

Constraints for this migration:

- Preserve existing polling intervals (tick) exactly where they are already specified.
- Add interval values where 'busy-loop' (no explicit interval) is currently provided

## Motivation

- Consistency: `github.com/stretchr/testify` is already a project dependency and widely used across the tests. Consolidating on its `Eventually` avoids mixing two near-identical constructs.
- Developer experience: Familiar APIs, clearer failure messages via standard `msgAndArgs`, and less cognitive overhead for contributors.
- Maintenance: One fewer external testing utility to track; unified assertion style across the suite.

### Shortfalls of guara `test.Eventually` argument type

`github.com/mariomac/guara/pkg/test.Eventually` passes a `github.com/stretchr/testify/require.TestingT` into the inner function rather than a `*testing.T`. This difference leads to several pitfalls in practice:

- Limited interface: `github.com/stretchr/testify/require.TestingT` exposes only assertion-related methods (`Errorf`, `FailNow`). It does not support `testing.T.Log`, `testing.T.Logf`, `testing.T.Helper`, `testing.T.Cleanup`, `testing.T.Run`, `testing.T.Skip`, `testing.T.Parallel`, or deadlines. Using the inner parameter as if it were `*testing.T` either fails to compile or drops expected behavior (for example, missing logs).
- Goroutine semantics: Implementations commonly run the inner function in a separate goroutine to emulate `FailNow`. Using the outer `*testing.T` from within that goroutine is unsafe and can trigger errors such as "FailNow called from non-test goroutine" or produce interleaved or missing logs.
- Failure surfacing: Assertion failures inside the inner function are collected and surfaced only at timeout, which can obscure where the failure occurred and hide intermediate logs expected from `*testing.T`.
- Developer confusion: The inner parameter name suggests a test handle, but it is not a full `*testing.T`. Misuse leads to broken logging, missing helpers or cleanup, and brittle tests.

[...]

PR #1215, opened 2026-01-30
Remove dependency on github.com/mariomac/guara

MrAlias:
Follow up to #1185 

- Decouple testing from the github.com/mariomac/guara
- Add a depguard to not use github.com/mariomac/guara going forward

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -97,7 +97,7 @@ linters:
               desc: "Use go.uber.org/multierr instead of github.com/hashicorp/go-multierror"
             - pkg: "math/rand$"
               desc: "Use the newer 'math/rand/v2' instead of math/rand"
-            - pkg: "github.com/mariomac/guara/pkg/test"
+            - pkg: "github.com/mariomac/guara"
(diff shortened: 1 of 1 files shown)


florianl (approved):

marctc (approved):

pkg/kube/kubecache/envtest/integration_test.go
```diff
@@ -71,11 +71,14 @@ func TestMain(m *testing.M) {
 		slog.Error("creating K8s manager client", "error", err)
 		os.Exit(1)
 	}
-	freePort, err = test.FreeTCPPort()
+	// Create a temporary listener just to get a free port
+	tempListener, err := net.Listen("tcp", ":")
```
mmat11, line 75: can we use testutil.FreeTCPPort() also here?

MrAlias: Unfortunately, no. There is no `*testing.T` available, only a `*testing.M`.

Issue #1025, opened 2025-12-17
Significant CPU impact on instrumented Redis


paweljw:
Hi! I noticed a significant impact on the CPU load from a Redis service instrumented with OBI. I whittled it down to a minimal reproduction: https://github.com/paweljw/redis-impact-repro I generally run `run_load.sh` first, then `run_obi.sh`.

This minimal configuration file seems to be enough to trigger it:

```
discovery:
  instrument:
  - open_ports: 6379
trace_printer: text
```

Here are the results I'm getting with the synthetic reproduction on a spare, otherwise empty 4-core server (not anywhere close to 100% loaded):

```
load-generator-1  | [STATS] avg_latency=10.73ms | redis_cpu=27.9% | ops/sec=9813
load-generator-1  | [STATS] avg_latency=9.49ms | redis_cpu=29.0% | ops/sec=10792
load-generator-1  | [STATS] avg_latency=8.49ms | redis_cpu=29.0% | ops/sec=12281
load-generator-1  | -------------- OBI STARTED -----------------
load-generator-1  | [STATS] avg_latency=9.55ms | redis_cpu=28.2% | ops/sec=10899
load-generator-1  | [STATS] avg_latency=10.75ms | redis_cpu=47.1% | ops/sec=9654
load-generator-1  | [STATS] avg_latency=10.32ms | redis_cpu=47.2% | ops/sec=10035
load-generator-1  | [STATS] avg_latency=10.78ms | redis_cpu=46.7% | ops/sec=9555
load-generator-1  | [STATS] avg_latency=11.64ms | redis_cpu=45.7% | ops/sec=8773
load-generator-1  | [STATS] avg_latency=10.92ms | redis_cpu=46.4% | ops/sec=9350
load-generator-1  | [STATS] avg_latency=10.41ms | redis_cpu=46.7% | ops/sec=9827
load-generator-1  | [STATS] avg_latency=11.45ms | redis_cpu=45.5% | ops/sec=8985
load-generator-1  | [STATS] avg_latency=11.78ms | redis_cpu=40.7% | ops/sec=8709
load-generator-1  | -------------- OBI STOPPED -----------------
load-generator-1  | [STATS] avg_latency=8.54ms | redis_cpu=28.2% | ops/sec=12126
load-generator-1  | [STATS] avg_latency=9.78ms | redis_cpu=28.2% | ops/sec=10658
[...]

PR #1216, opened 2026-01-30
Improve performance of the eBPF ring buffer and Go tracer

grcevski:
This work was prompted because of a related Beyla customer reported issue https://github.com/grafana/beyla/issues/2425.

Essentially, we can construct example service programs that stress the eBPF probe usage and the instrumentation overhead. Even though typically our eBPF instrumentation overhead is small for many services, if the service itself is spending very little CPU and it's responding in microseconds, then the eBPF overhead becomes visible.

The issue above was reported for Go, however we have similar issue reported for the generic tracer, https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1025. Redis has very low response latency, so with running high volume requests on such a service, the impact of the eBPF instrumentation will be visible.

To be fair, I don't think there's another type of instrumentation that will have lesser impact, certainly not adding instrumentation in the code or by running a proxy service, but it's good to know the limits and overhead of what we do.

I implemented two changes:

1. Based on profiling the application, one of the major overheads is the default trigger of the ringbuffer to userspace. I increased the ringbuffer size and I set a defaulf of wakeup_len to make the ringbuffer accumulate more aggressively before shipping the data to userspace. This improvement impacts the gotracer and the generic tracer.
2. The next big performance impact is the number of probes that run. I used bpftop to find which probes run the most and I found that the probe for extracting the Go HTTP traceparent header was the top hitter. I implemented an approach using bpf_loop just like we do in the generic tracer to reduce the per header probe to 1 per request probe. This only helps the Go tracer.

Here are the performance results (benchmark is provided with the PR, where I described the methodology I used, with CPU pinning and running on quiet hardware):

### Baseline without OBI (CPU core utilisation 30%):
```
    HTTP
    http_req_duration.......................................................: avg=49.53µs min=32.49µs med=39.64µs max=4.26ms p(90)=64.99µs p(95)=83.76µs 
      { expected_response:true }............................................: avg=49.53µs min=32.49µs med=39.64µs max=4.26ms p(90)=64.99µs p(95)=83.76µs 
```

### Baseline with OBI (CPU core utilisation 70%)
```
    HTTP
    http_req_duration.......................................................: avg=80.95µs  min=54.78µs med=72.31µs max=2.93ms p(90)=100.42µs p(95)=115.49µs
      { expected_response:true }............................................: avg=80.95µs  min=54.78µs med=72.31µs max=2.93ms p(90)=100.42µs p(95)=115.49µs

```

As you can see the numbers don't look great.

### After optimising the ring buffer (CPU core utilisation 60%)
```
    HTTP
    http_req_duration.......................................................: avg=71.23µs min=53.7µs  med=63.57µs max=4.16ms p(90)=88.71µs  p(95)=100.01µs
      { expected_response:true }............................................: avg=71.23µs min=53.7µs  med=63.57µs max=4.16ms p(90)=88.71µs  p(95)=100.01µs
```

### After optimising the Go readContinuedSlice probe (CPU core utilisation 54%)
```
    HTTP
    http_req_duration.......................................................: avg=66.66µs min=50.16µs med=59.5µs  max=3.73ms p(90)=82.19µs p(95)=93.26µs 
      { expected_response:true }............................................: avg=66.66µs min=50.16µs med=59.5µs  max=3.73ms p(90)=82.19µs p(95)=93.26µs
```

There's still a lot of overhead left, but this change makes a decent dent in the overhead.

When I went to add the benchmark test code, I realised we had some very old test programs in the `test/integration/cmd` directory that have since been migrated to `test/integration/components` and I removed this unused test code.

### DIFF
--- a/internal/test/integration/test_python_elasticsearchclient.go
+++ b/internal/test/integration/test_python_elasticsearchclient.go
@@ -63,9 +63,7 @@ func assertElasticsearchOperation(t *testing.T, dbSystemName, op, queryText, ind
 	}
 	fullJaegerURL := fmt.Sprintf("%s?%s", jaegerQueryURL, params.Encode())
 
-	tt := t
-
-	tt.Log(fullJaegerURL)
+	t.Log(fullJaegerURL)
 
 	require.EventuallyWithT(t, func(ct *assert.CollectT) {
 		resp, err := http.Get(fullJaegerURL)
@@ -84,34 +82,35 @@ func assertElasticsearchOperation(t *testing.T, dbSystemName, op, queryText, ind
 		for i := range lastTrace.Spans {
 			span := &lastTrace.Spans[i]
 
-			tt.Log(span)
+			t.Log(span)
 			tag, found := jaeger.FindIn(span.Tags, "db.operation.name")
 
 			if !found || tag.Value == nil {
 				continue
 			}
 
-		assert.Contains(ct, span.OperationName, operationName)
+			assert.Contains(ct, span.OperationName, operationName)
 
-		tag, found := jaeger.FindIn(span.Tags, "db.query.text")
-		assert.True(ct, found)
-		assert.Equal(ct, queryText, tag.Value.(string))
+			tag, found = jaeger.FindIn(span.Tags, "db.query.text")
+			assert.True(ct, found)
+			assert.Equal(ct, queryText, tag.Value.(string))
 
-		tag, found = jaeger.FindIn(span.Tags, "db.collection.name")
-		assert.True(ct, found)
-		assert.Equal(ct, index, tag.Value)
+			tag, found = jaeger.FindIn(span.Tags, "db.collection.name")
+			assert.True(ct, found)
+			assert.Equal(ct, index, tag.Value)
 
-		tag, found = jaeger.FindIn(span.Tags, "db.namespace")
-		assert.True(ct, found)
-		assert.Empty(ct, tag.Value)
+			tag, found = jaeger.FindIn(span.Tags, "db.namespace")
+			assert.True(ct, found)
+			assert.Empty(ct, tag.Value)
 
-		tag, found = jaeger.FindIn(span.Tags, "db.system.name")
-		assert.True(ct, found)
-		assert.Equal(ct, dbSystemName, tag.Value)
+			tag, found = jaeger.FindIn(span.Tags, "db.system.name")
+			assert.True(ct, found)
+			assert.Equal(ct, dbSystemName, tag.Value)
 
-		tag, found = jaeger.FindIn(span.Tags, "elasticsearch.node.name")
-		assert.True(ct, found)
-		assert.Empty(ct, tag.Value)
+			tag, found = jaeger.FindIn(span.Tags, "elasticsearch.node.name")
+			assert.True(ct, found)
+			assert.Empty(ct, tag.Value)
+		}
 	}, testTimeout, 100*time.Millisecond)
 }
 


bpf/gotracer/go_nethttp.c
```diff
@@ -15,14 +15,17 @@
 
 //go:build obi_bpf_ignore
 
+#include "gotracer/go_offsets.h"
```
rafaelroquetto: vscode being vscode

bpf/gotracer/go_nethttp.c
```diff
@@ -411,6 +420,80 @@ static __always_inline unsigned char *match_header(
...
 
+SEC("uprobe/readMimeHeader")
+int obi_uprobe_readMimeHeader(struct pt_regs *ctx) {
+    bpf_dbg_printk("=== uprobe/proc ReadMimeHeader === ");
+
+    void *goroutine_addr = GOROUTINE_PTR(ctx);
+    bpf_dbg_printk("goroutine_addr %lx", goroutine_addr);
+    go_addr_key_t g_key = {};
+    go_addr_key_from_id(&g_key, goroutine_addr);
+    connection_info_t *existing = bpf_map_lookup_elem(&ongoing_server_connections, &g_key);
```
rafaelroquetto:
```suggestion
    const connection_info_t *existing = bpf_map_lookup_elem(&ongoing_server_connections, &g_key);
```

just a nit - having said that, do you really mean to just check if the value exists and then bail (I noticed `existing` is not used after line 432 so just double checking it's intended)

bpf/gotracer/go_nethttp.c
```diff
@@ -411,6 +420,80 @@ static __always_inline unsigned char *match_header(
...
+
+    bpf_clamp_umax(len, TRACE_BUF_SIZE);
+
+    bpf_probe_read_user(buf, len, arr);
+
+    bpf_dbg_printk("buf %s", buf);
+
+    unsigned char *tp_ptr = bpf_strstr_tp_loop(buf, len);
+
+    bpf_dbg_printk("tp %llx", tp_ptr);
```
rafaelroquetto: another nit, but just to be in line with the changes by Giuseppe
```suggestion
    bpf_dbg_printk("tp=%llx", tp_ptr);
```

bpf/gotracer/go_sdk.c
```diff
@@ -18,6 +18,7 @@
 
 //go:build obi_bpf_ignore
 
+#include "common/map_sizing.h"
```
rafaelroquetto: vsstudio shenanigans

mmat11 (approved):
lgtm!

grcevski:
https://www.githubstatus.com

pkg/obi/config.go
```diff
@@ -95,6 +95,7 @@ var DefaultConfig = Config{
 		BatchLength:        100,
 		BatchTimeout:       time.Second,
 		HTTPRequestTimeout: 0,
+		WakeupLen:          500,
```
mmat11, line 98: just one last comment/question: given the fact that this is around 250KB should we start with batching something less in order to not send spikes of work to userspace? I was thinking something like `100` maybe

grcevski: ok, sure, let me test a bit and see what happens with 100. I think it will be more less the same for performance and I'll watch for OBI memory usage.

grcevski: I measured and there's very little difference between 100 and 500 in terms of performance, 2K on average more memory for OBI userspace with 500. I think  it's a wash to be honest.

grcevski: Let's try 500 and then if issues arise we can drop it to 100. The ringbuf is 1MB now, so we have 3/4 free with 250K for the default limit.

PR #1218, opened 2026-02-01
Fix code coverage in parallel integration test

### LINKED ISSUE #1100: Fix code coverage reports
Image

Real code coverage was usually near 80% but suddenly went down to 50%. I suspect this might be related with the new way of running integration tests in parallel.

ajimenez1503:
## Checklist

Fix #1100

## Summary

Fix code coverage reporting in parallelized integration tests by adding unique identifiers to Codecov uploads.

## Problem

After parallelizing integration tests into 10 partitions, code coverage dropped from ~80% to ~55%. Each partition was uploading coverage independently, but Codecov wasn't merging them properly.

## Solution

Added unique identifiers to each Codecov upload:

```
name: integration-test-${{ matrix.id }}
```

cc @open-telemetry/ebpf-instrumentation-approvers

### DIFF
--- a/.github/workflows/workflow_integration_tests_vm.yml
+++ b/.github/workflows/workflow_integration_tests_vm.yml
@@ -167,21 +167,16 @@ jobs:
         uses: codecov/codecov-action@671740ac38dd9b0130fbe1cec585b89eea48d3de # v5.5.2
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-          ARCH: ${{ matrix.kernel.arch }}
-          KERNEL_VERSION: ${{ matrix.kernel.kernel-version }}
         with:
           files: ./testoutput/itest-covdata.txt
-          flags: integration-test-vm-${ARCH}-${KERNEL_VERSION}
-          name: vm-integration-test-coverage-${KERNEL_VERSION}-${ARCH}-${{ matrix.test.id }}-${{ github.run_number }}
+          flags: integration-test-vm-${{ matrix.kernel.arch }}-${{ matrix.kernel.kernel-version }}
+          name: vm-integration-test-coverage-${{ matrix.kernel.kernel-version }}-${{ matrix.kernel.arch }}-${{ matrix.test.id }}-${{ github.run_number }}
 
       - name: Upload integration test logs
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
-        env:
-          KERNEL_VERSION: ${{ matrix.kernel.kernel-version }}
-          ARCH: ${{ matrix.kernel.arch }}
         if: always()
         with:
-          name: vm-integration-test-logs-${KERNEL_VERSION}-${ARCH}-${{ matrix.test.id }}-${{ github.run_number }}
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request_integration_tests.yml
```diff
@@ -159,6 +159,7 @@ jobs:
         with:
           files: ./testoutput/itest-covdata.txt
           flags: integration-test
+          name: integration-test-${{ matrix.id }}
```
florianl: Should we use something like `${{ github.run_id }}-${{ github.run_number }}` instead?

I didn't find documentation around `matrix.id` in https://docs.github.com/en/actions/reference/workflows-and-actions/contexts.

skl: `matrix` is defined above at `jobs.test-matrix.outputs.matrix` (line 39). For the logs we use `${{ github.run_number }}-${{ matrix.id }}`, the matrix.id is equal to the shard number so helps to more easily associate a file with a shard.

.github/workflows/pull_request_integration_tests.yml
```diff
@@ -159,20 +159,21 @@ jobs:
         with:
           files: ./testoutput/itest-covdata.txt
           flags: integration-test
+          name: go-integration-test-coverage-${{ github.run_number }}-${{ matrix.id }}
```
mmat11: should these be `matrix.id`-`github.run_number` for consistency with the other jobs?

.github/workflows/workflow_integration_tests_vm.yml
```diff
@@ -172,12 +172,13 @@ jobs:
...
           files: ./testoutput/itest-covdata.txt
           flags: integration-test-vm-${ARCH}-${KERNEL_VERSION}
+          name: vm-integration-test-coverage-${ARCH}-${{ github.run_number }}-${{ matrix.kernel.kernel-version }}-${{ matrix.test.id }}
 
       - name: Upload integration test logs
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         if: always()
         with:
-          name: vm-test-logs-${{ github.run_number }}-${{ matrix.kernel.kernel-version }}-${{ matrix.test.id }}
+          name: vm-integration-test-logs-${{ matrix.kernel.arch }}-${{ github.run_number }}-${{ matrix.kernel.kernel-version }}-${{ matrix.test.id }}
```
mmat11: what about `kernel version`-`arch`-`matrix.id`-`github.run_number`

mmat11:
> Hi team, The test `TestAvoidedServicesMetrics` is being `flaky test`. It is failing due to a data race in the `dockertest` library. Should I run the docker setup in sequence to avoid those race conditions?
> 
> https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/5dcbb8fd4701ea89dbdc104ff7451b4c50634fda/internal/test/integration/internal_metrics_test.go#L42-L60

See: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1232 // https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1233

PR #1230, opened 2026-02-04
Refactor and streamline NVIDIA GPU metrics support

grcevski:
## Refactoring and cleanup of the NVIDIA Cuda GPU support

A lot of the code was never production ready and it was inherited from a hackathon. The major change I want to do with this PR is to:
1. Remove the code that symbolicates the Cuda kernel names. While this is useful in profiling scenarios, it's a bad idea for the metrics we are generating, since typically there are many kernels and it will create a cardinality explosion. At the same time, given the large number of kernels it's almost impossible to visualise this data in metrics. We should maybe consider porting the Cuda kernel name extraction to the OTel profiler for the creation of a count profile.
2. Rename all symbols from generic GPU to actually say that this is Cuda specific code, since the same concepts and metrics series would make no sense for another GPU vendor. I've extracted the cuda specifc code in separate bpf files.
3. Recently, advanced GenAI frameworks (like VLLM) tend to optimise the kernel launch code into groups of "graphs", invoked by cudaGraphLaunch. The graphs then launch cuda kernels. I added metrics to track the graph launches as well.

I also added autodetection for NVIDIA drivers present on the system to enable/disable CUDA instrumentation support.

## Checklist

### DIFF
--- a/bpf/gpuevent/gpu_ringbuf.h
+++ b/bpf/gpuevent/gpu_ringbuf.h
@@ -7,4 +7,5 @@
 
 struct {
     __uint(type, BPF_MAP_TYPE_RINGBUF);
+    __uint(max_entries, 1 << 16);
 } gpu_events SEC(".maps");


bpf/common/trace_common.h
```diff
@@ -477,4 +477,17 @@ static __always_inline u8 find_parent_trace_for_client_request(const pid_connect
...
         p_conn, orig_dport, &t_key, pid_tgid, tp);
+}
+
+static __always_inline void init_new_trace(tp_info_t *tp) {
+    bpf_d_printk("Generating new traceparent id [%s]", __FUNCTION__);
+    new_trace_id(tp);
+    urand_bytes(tp->span_id, SPAN_ID_SIZE_BYTES);
+    __builtin_memset(tp->parent_id, 0, sizeof(tp->span_id));
+
+    if (g_bpf_debug) {
```
rafaelroquetto, line 488: Search for `print_tp` - it appears in one or two places, perhaps we should consolidate those.

It does not have to be in this PR, happy to do it as a follow-up - just mentioning it in case you've got the appetite.

bpf/gpuevent/cuda.c
```diff
@@ -0,0 +1,153 @@
...
+#include <logger/bpf_dbg.h>
+
+#include <pid/pid.h>
+
+const cuda_kernel_launch_t *unused_gpu __attribute__((unused));
+const cuda_malloc_t *unused_gpu1 __attribute__((unused));
+const cuda_memcpy_t *unused_gpu2 __attribute__((unused));
+const cuda_graph_launch_t *unused_gpu3 __attribute__((unused));
+
+#define EVENT_GPU_KERNEL_LAUNCH 1
```
rafaelroquetto: **nit** _(feel free to ignore)_: I understand this is Meta code? In any case, `enum { k_foo = 1}` is what we've been pushing for in new OBI code. The rationale summary:

- type safety - that's the main one: the compiler can catch invalid assignments and arguments, unlike #define which is untyped text substitution
- avoids global macro namespace pollution
- better compiler diagnostics (e.g. missing switch cases, invalid values)
- debuggers display enum symbolic names instead of raw integers
- tooling (static analysis, IDE refactors) works reliably with enums
- exception: use #define only when preprocessor behavior is required (e.g., macros, #if, conditional compilation).

Most the above are general guidelines and may not impact the context of this code or eBPF.

bpf/gpuevent/cuda.c
```diff
@@ -0,0 +1,153 @@
...
+#define EVENT_GPU_MALLOC 2
+#define EVENT_GPU_MEMCPY 3
+#define EVENT_CUDA_GRAPH_LAUNCH 4
+
+// The caller uses registers to pass the first 6 arguments to the callee.  Given
+// the arguments in left-to-right order, the order of registers used is: %rdi,
+// %rsi, %rdx, %rcx, %r8, and %r9. Any remaining arguments are passed on the
+// stack in reverse order so that they can be popped off the stack in order.
+#define SP_OFFSET(offset) (void *)PT_REGS_SP(ctx) + (offset * 8)
```
rafaelroquetto: This only works for x64, not arm - just making sure that's intended.

bpf/gpuevent/cuda.h
```diff
@@ -11,46 +11,41 @@
...
     int grid_z;
     int block_x;
     int block_y;
     int block_z;
-    uint64_t stream;
-    uint64_t args[MAX_GPUKERN_ARGS];
-    size_t ustack_sz;
-    stack_trace_t ustack;
-} gpu_kernel_launch_t;
+    u64 stream;
```
rafaelroquetto: nit: I'd move this under `kern_func_off` - it makes it easier to reason about the alignment and padding (if another `int` is added, the padding would be in the middle, not the end of the struct).

bpf/gpuevent/cuda.c
grcevski: Actually I'll remove it, we don't use the stream

bpf/gpuevent/cuda.c
grcevski: I can change that to enum.

bpf/common/trace_common.h
```diff
@@ -477,4 +477,17 @@ static __always_inline u8 find_parent_trace_for_client_request(const pid_connect
 
     return find_parent_trace_for_client_request_with_t_key(
         p_conn, orig_dport, &t_key, pid_tgid, tp);
+}
+
+static __always_inline void init_new_trace(tp_info_t *tp) {
+    bpf_d_printk("Generating new traceparent id [%s]", __FUNCTION__);
+    new_trace_id(tp);
+    urand_bytes(tp->span_id, SPAN_ID_SIZE_BYTES);
+    __builtin_memset(tp->parent_id, 0, sizeof(tp->span_id));
```
mmat11, line 486: nit: we should import `#include <bpfcore/bpf_builtins.h>` and use `__bpf_memzero(tp->parent_id, sizeof(tp->span_id));`, ideally all these builtin calls should use the optimized versions one day

bpf/gpuevent/gpu_ringbuf.h
```diff
@@ -0,0 +1,10 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <bpfcore/utils.h>
+
+struct {
+    __uint(type, BPF_MAP_TYPE_RINGBUF);
```
mmat11, line 9: doesn't this need to be sized? or will it be resized in userspace?

mmat11 (approved):
lgtm, just a couple comments

bpf/common/trace_common.h
grcevski: I just moved that function, we can refactor that later IMO.

bpf/gpuevent/gpu_ringbuf.h
grcevski: yeah good point, let me put a size, I think it just gets some default size.

bpf/common/trace_common.h
grcevski: yeah that's a moved function since I initially wanted to reuse it, but later I found out that all of these LLM solutions use custom protocols for talking from the main application to the one that works with cuda, so we can't correlate anyway.

PR #1236, opened 2026-02-04
Make bpffs errors into warnings

rafaelroquetto:
Instead of showing the lack of `bpffs` as an error, show them as more informative warnings in order not to spook customers (as OBI continues to run).

Bonus: normalise variable names casing based on the go convention (and assuming `bpffs` to be a single term, as is the default in other tools).

## Checklist

### DIFF
--- a/pkg/ebpf/tracer_linux.go
+++ b/pkg/ebpf/tracer_linux.go
@@ -196,9 +196,9 @@ func (pt *ProcessTracer) makeOtelBPFFSPath() (string, error) {
 }
 
 func (pt *ProcessTracer) setupBPFFS(spec *ebpf.CollectionSpec) string {
-	var err error
+	otelBPFFSPath, err := pt.makeOtelBPFFSPath()
 
-	if otelBPFFSPath, err := pt.makeOtelBPFFSPath(); err == nil {
+	if err == nil {
 		return otelBPFFSPath
 	}
 


pkg/ebpf/tracer_linux.go
```diff
@@ -195,6 +195,26 @@ func (pt *ProcessTracer) makeOtelBpfFsPath() (string, error) {
 	return otelPath, nil
 }
 
+func (pt *ProcessTracer) setupBPFFS(spec *ebpf.CollectionSpec) string {
+	var err error
```
mmat11: is this needed?

pkg/ebpf/tracer_linux.go
```diff
@@ -195,6 +195,26 @@ func (pt *ProcessTracer) makeOtelBpfFsPath() (string, error) {
...
+
+	if otelBPFFSPath, err := pt.makeOtelBPFFSPath(); err == nil {
+		return otelBPFFSPath
+	}
+
+	slog.Warn("creating OTEL namespace in bpffs failed (is bpffs mounted?)", "bpffs_path", pt.bpffsPath, "err", err)
+	slog.Warn("OBI will still work, but features depnding on pinned maps (e.g., log enricher, profile correlation) will be disabled")
```
mmat11: typo
```suggestion
	slog.Warn("OBI will still work, but features depending on pinned maps (e.g., log enricher, profile correlation) will be disabled")
```

pkg/ebpf/tracer_linux.go
rafaelroquetto: yes, otherwise `err` is not accessible outside the scope of the `if` statement (it's used in the `Warn` calls

mmat11: ah didn't notice it's used after, 👍🏻

pkg/ebpf/tracer_linux.go
```diff
@@ -195,6 +195,26 @@ func (pt *ProcessTracer) makeOtelBpfFsPath() (string, error) {
 	return otelPath, nil
 }
 
+func (pt *ProcessTracer) setupBPFFS(spec *ebpf.CollectionSpec) string {
+	var err error
+
+	if otelBPFFSPath, err := pt.makeOtelBPFFSPath(); err == nil {
```
MrAlias: This is shadowing `err`. If you want to assign to the `err` from the parent context the function call cannot be returned to the declaration of the clause context.

```suggestion
	if otelBPFFSPath, err = pt.makeOtelBPFFSPath(); err == nil {
```

MrAlias: note: `otelBPFFSPath` needs to be declared separate as well based on that suggestion.

You could also do something like:

```suggestion
	if otelBPFFSPath, e := pt.makeOtelBPFFSPath(); e == nil {
		err = e
```

rafaelroquetto: Nice catch, many thanks for this. Fixed!

PR #1250, opened 2026-02-06
Fix support of different values for SDKLogLevel field

### LINKED ISSUE #1036: SDKLogLevel field
Hi folks!

I noticed that the `SDKLogLevel` field in `MetricsConfig` and in `TracesConfig` have the same env variable name `OTEL_EBPF_SDK_LOG_LEVEL`  () but the supported values are different.
1. [`SDKLogLevel`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/3781840a42ffed54dbeb070aa22ec72799ed61b0/pkg/export/otel/otelcfg/config_metrics.go#L45) in `MetricsConfig` is used as `slog.Level` which accepts: debug, info, warn and error
2. [`SDKLogLevel`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/3781840a42ffed54dbeb070aa22ec72799ed61b0/pkg/export/otel/otelcfg/config_traces.go#L55) in `TracesConfig` is used as `zapcore.Level` which accepts: debug, info, warn/warning, error, dpanic, panic and fatal

I wanted to change the name of the environment variable, something like `OTEL_EBPF_METRICS_SDK_LOG_LEVEL` and `OTEL_EBPF_TRACES_SDK_LOG_LEVEL`, but this is a breaking change, so I'd like to hear your opinion first.

ajimenez1503:
## Proposal: align `OTEL_EBPF_SDK_LOG_LEVEL` with `OTEL_EBPF_LOG_LEVEL`

- **Supported values**: `debug`, `info`, `warn`, `error` (case‑insensitive).
- **Backward compatibility**: if `dpanic`, `panic`, or `fatal` are provided, **downgrade to `error`** and emit a warning.
- **Traces logger**: after `zapcore.Level.UnmarshalText`, map any level above `error` to `error` with a warning.
- **Metrics/SDK logger**: no change needed; `slog.Level` already only supports `debug/info/warn/error`.
- **Docs/comments**: update `otel_sdk_log_level` comments to explicitly list supported values and note the downgrade behavior.

This keeps the SDK log level consistent with our global log level without breaking existing configs.

Resolve #1036

## Checklist

### DIFF
--- a/pkg/export/otel/traces_test.go
+++ b/pkg/export/otel/traces_test.go
@@ -14,6 +14,7 @@ import (
 	expirable2 "github.com/hashicorp/golang-lru/v2/expirable"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/zap/zapcore"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/consumer"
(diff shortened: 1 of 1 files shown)


grcevski (approved):
LGTM! Thanks!

PR #1251, opened 2026-02-06
Improve gRPC client performace

grcevski:
While running on large systems with multiple memory controllers it appears that the cost of reading user memory (in uprobes) becomes very high. In gRPC client workloads, the NewStream call happens on every gRPC request which tend to be quick. We were reading the connection information in this probe every time, even though typically grpc requests reuse the same long lived client. 

With this PR I'm making the following improvements:
- Caching of the connection information for grpc clients so we don't have to read it every time.
- I've removed the probe on goroutine exit, it was there to cleanup data for the map, which is already LRU and it costs us probe invocation.
- The netFD.Read probe wasn't returning upon finding the specific information.

This change cuts the time in NewStream by 3/4. On my tests this isn't too bad, around 5% of CPU when I run a grpc benchmark, but on a 48 core machine the time was 25%.

### DIFF
--- a/bpf/gotracer/go_net.c
+++ b/bpf/gotracer/go_net.c
@@ -49,20 +49,6 @@ int obi_uprobe_netFdRead(struct pt_regs *ctx) {
         return 0;
     }
 
-    // lookup active HTTP connection
-    connection_info_t *conn = bpf_map_lookup_elem(&ongoing_server_connections, &g_key);
-    if (conn) {
-        if (conn->d_port == 0 && conn->s_port == 0) {
-            bpf_dbg_printk(
-                "Found existing server connection, parsing FD information for socket tuples, %llx",
-                goroutine_addr);
-
-            void *fd_ptr = GO_PARAM1(ctx);
-            get_conn_info_from_fd(fd_ptr, conn); // ok to not check the result, we leave it as 0
-        }
-        //dbg_print_http_connection_info(conn);
-        return 0;
-    }
     // lookup active sql connection
     sql_func_invocation_t *sql_conn = bpf_map_lookup_elem(&ongoing_sql_queries, &g_key);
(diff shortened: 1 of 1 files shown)


bpf/gotracer/go_net.c
```diff
@@ -0,0 +1,88 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build obi_bpf_ignore
+
+#include <bpfcore/vmlinux.h>
```
grcevski, line 18: This code was in go_nethttp.c, it should be standalone so I moved it out in its own file.

bpf/gotracer/go_grpc.c
```diff
@@ -590,34 +597,45 @@ int obi_uprobe_transport_http2Client_NewStream(struct pt_regs *ctx) {
...
+        // reading the connection can be expensive for high volume of
+        // new grpc client connections. We cache it, since most grpc client
+        // connections are long lived.
+        if (!cached_conn) {
+            void *s_ptr = 0;
+            buf[0] = 0;
+            bpf_probe_read(&s_ptr, sizeof(s_ptr), (void *)(t_ptr + grpc_t_conn_pos));
+            bpf_probe_read(buf, sizeof(buf), s_ptr);
+
+            //bpf_dbg_printk("scheme %s", buf);
```
rafaelroquetto, line 611: intended left-over?

grcevski: Yeah it was always there in case we find a problem and we need to see what it is.

PR #1265, opened 2026-02-09
Add support for parsing couchbase Flexiable framing

NimrodAvni78:
## Checklist

Couchbase packets have an alternative header structure for [flexible frames](https://github.com/couchbase/kv_engine/blob/master/docs/BinaryProtocol.md?#response-header-with-flexible-framing-extras)

We were incorrectly parsing these packets and creating a situation where the size of the key looks to big to be included in the packet, hence why we didn't classify the header as valid

### DIFF
--- a/pkg/ebpf/common/tcp_detect_transform_test.go
+++ b/pkg/ebpf/common/tcp_detect_transform_test.go
@@ -323,7 +323,7 @@ func TestReadTCPRequestIntoSpan_CouchbaseKeyNotFound(t *testing.T) {
 	binaryRecord := bytes.Buffer{}
 	require.NoError(t, binary.Write(&binaryRecord, binary.LittleEndian, tri))
 
-	fltr := TestPidsFilter{services: map[uint32]svc.Attrs{}}
+	fltr := TestPidsFilter{services: map[app.PID]svc.Attrs{}}
 
 	span, ignore, err := ReadTCPRequestIntoSpan(ctx, &cfg, &ringbuf.Record{RawSample: binaryRecord.Bytes()}, &fltr)
 	require.NoError(t, err)
@@ -385,7 +385,7 @@ func TestReadTCPRequestIntoSpan_CouchbaseFlexibleFraming(t *testing.T) {
 	binaryRecord := bytes.Buffer{}
 	require.NoError(t, binary.Write(&binaryRecord, binary.LittleEndian, tri))
 
-	fltr := TestPidsFilter{services: map[uint32]svc.Attrs{}}
+	fltr := TestPidsFilter{services: map[app.PID]svc.Attrs{}}
 
 	span, ignore, err := ReadTCPRequestIntoSpan(ctx, &cfg, &ringbuf.Record{RawSample: binaryRecord.Bytes()}, &fltr)
 	require.NoError(t, err)


pkg/internal/ebpf/couchbasekv/header.go
```diff
@@ -108,6 +139,19 @@ func ParsePacket(pkt []byte) (*Packet, error) {
...
 	remaining := len(pkt) - offset
 
+	// Parse framing extras first (only for flexible framing packets)
+	if header.FramingExtrasLen > 0 {
+		framingExtrasLen := int(header.FramingExtrasLen)
+		if remaining < framingExtrasLen {
+			framingExtrasLen = remaining
+		}
+		if framingExtrasLen > 0 {
+			packet.FramingExtras = pkt[offset : offset+framingExtrasLen]
```
mmat11: do we need some safety checks here?

pkg/internal/ebpf/couchbasekv/header.go
```diff
@@ -224,6 +269,11 @@ func (p *Packet) HasFullExtras() bool {
 	return len(p.Extras) == int(p.Header.ExtrasLen)
 }
 
+// HasFullFramingExtras returns true if the complete framing extras were parsed.
+func (p *Packet) HasFullFramingExtras() bool {
```
mmat11: I think this is unused

pkg/internal/ebpf/couchbasekv/header.go
grcevski: +1 maybe also good idea to add some checks on remaining not going < 0?

pkg/internal/ebpf/couchbasekv/header.go
```diff
@@ -12,19 +12,25 @@ import (
...
-	DataType  DataType
-	VBucketID uint16 // For requests: VBucket ID; For responses: Status code
-	Status    Status // Alias for VBucketID when parsing responses
-	BodyLen   uint32 // Total body length = extras + key + value
-	Opaque    uint32 // Echoed back in response, like correlation ID
-	CAS       uint64 // Compare-and-swap value
+	Magic            Magic
+	Opcode           Opcode
+	FramingExtrasLen uint8  // Only used for flexible framing (magic 0x08/0x18)
+	KeyLen           uint16 // For flexible framing, this is only 8 bits but stored as uint16
```
rafaelroquetto:
```suggestion
	KeyLen           uint16 // For flexible framing, this is only 8 bits but stored as uint16
	FramingExtrasLen uint8  // Only used for flexible framing (magic 0x08/0x18)
```

rationale: https://github.com/couchbase/libcouchbase/blob/a129df617431cac279413fcdb2ec808c3f9c890e/include/memcached/protocol_binary.h#L428

pkg/internal/ebpf/couchbasekv/header.go
```diff
@@ -12,19 +12,25 @@ import (
...
+	ExtrasLen        uint8
+	DataType         DataType
+	VBucketID        uint16 // For requests: VBucket ID; For responses: Status code
+	Status           Status // Alias for VBucketID when parsing responses
```
rafaelroquetto: wouldn't it make sense to not reuse this, instead declaring a new [struct for the response](https://github.com/couchbase/libcouchbase/blob/a129df617431cac279413fcdb2ec808c3f9c890e/include/memcached/protocol_binary.h#L447)?

The way this is laid out now, it prevents us from reinterpreting binary data into this type (also `For requests: VBucket ID; For responses: Status code` is detrimental for code readability - the comment is buried in the definition)

pkg/internal/ebpf/couchbasekv/header.go
NimrodAvni78: will do!

pkg/internal/ebpf/couchbasekv/header.go
NimrodAvni78: yeah you are right, just seemed like code duplication only for 1 field
but that said it does remove ambiguity so i agree its a good idea

pkg/internal/ebpf/couchbasekv/header.go
NimrodAvni78: Correct me if i am wrong but i think it already checks that remaining can't be under 0
it first sets what we try to read (framingExtrasLen / KeyLen, ...) to be no bigger then `remaining`, then we check if it is larger then 0 then we read, and continue to the next, then if `remaining` is smaller then 0 we will stop reading the next variables no?

pkg/internal/ebpf/couchbasekv/header.go
```diff
@@ -60,171 +88,148 @@ func ParseHeader(pkt []byte) (*Header, error) {
...
+	}
+	return end
 }
 
-// HasFullValue returns true if the complete value was parsed.
-func (p *Packet) HasFullValue() bool {
-	return len(p.Value) == p.Header.ValueLen()
+// validateHeader checks for basic validity of the parsed header.
+func validateHeader(h Header) error {
+	minBodyLen := int(h.FramingExtrasLen()) + int(h.ExtrasLen()) + int(h.KeyLen())
```
grcevski, line 228: Is it theoretically possible for us to create a header with less bytes than required? I'm wondering if we should check the total header len before we call the FramingExtrasLen etc, because they pull data out of the header and if it's say of size 2 (or 0) we'll get a panic.

NimrodAvni78: this is done after a validation that the header it self is of valid size, i.e if len(h) > HeaderLen (==24)
its done in `ParseHeader` before Header is not created before this validation (and that Magic byte is a valid one)

grcevski: Excellent!

PR #1266, opened 2026-02-09
Unify PID type

mariomac:
We had:
* `svc.PID`
* `docker.PID`
* `uint32`
* `int32`

This PR replaces all by `app.PID` and keeps raw int types for some accessory libraries (gopsutil and Cilium-autogenerated code)


grcevski (approved):
Nice, thanks for unifying this! LGTM!

Issue #1257, opened 2026-02-07
Feature: Per-service/namespace control for Java and Node.js agent injection


gk-clode:
## Problem Statement

Currently, the Java agent (`javaagent`) and Node.js agent (`nodejs`) injectors can only be enabled/disabled globally via:

- `OTEL_EBPF_JAVAAGENT_ENABLED=false`
- `OTEL_EBPF_NODEJS_ENABLED=false`

This is problematic when:
1. **Some applications have library conflicts** - e.g., Neo4j bundles caffeine cache which conflicts with OBI's shaded `io.opentelemetry.obi.com.github.benmanes.caffeine.*` classes, causing `NoClassDefFoundError`
2. **Users want eBPF traces but not agent injection** for specific services
3. **Different namespaces have different requirements** - production apps may need full instrumentation while third-party databases should be excluded

## Current Workaround

Users must either:
1. Disable agents globally (losing functionality for ALL Java/Node.js apps)
2. Use `exclude_instrument` to exclude the namespace entirely (losing ALL eBPF traces for that service)

Neither option is ideal.

## Proposed Solution

Add per-service agent injection control to the `instrument` discovery configuration, similar to existing per-service options like `exports`, `sampler`, and `routes`.

### Option 1: Unified control

```yaml
discovery:
  instrument:
    - k8s_namespace: juno
      k8s_statefulset_name: neo4j
      agent_injection: false  # Disable both Java + Node.js agents
      
    - k8s_namespace: production
      agent_injection: true   # Enable (default)
```

### Option 2: Granular control

```yaml
discovery:
  instrument:
    - k8s_namespace: juno
      javaagent_enabled: false
      nodejs_enabled: true
      
    - k8s_pod_labels:
        app: legacy-app
      javaagent_enabled: false
```

### Option 3: Exclusion list (similar to `exclude_instrument`)

```yaml
javaagent:
  enabled: true
  exclude:
    - k8s_namespace: juno
    - k8s_statefulset_name: neo4j

nodejs:
  enabled: true
  exclude:
    - k8s_namespace: kube-system
```

## Implementation Notes

Based on code analysis:

1. **`GlobAttributes` struct** (`pkg/appolly/services/attr_glob.go`) already supports per-service config for `exports`, `sampler`, `routes`, `metrics`. Adding `javaagent_enabled` and `nodejs_enabled` fields would follow the same pattern.

[...]

PR #1267, opened 2026-02-09
Fix java caffeine class not found

grcevski:
We were using an LRU map to ensure we have capped hashmap for the datastructures the java agent uses, if because of exceptions we don't always clean-up the data we have. However this was an overkill for a few reasons:
1. We don't actually need LRU semantics, all data is short lived and we just need a capped map.
2. Caffeine is great, but also there are warnings when we load it on JDK 25.
3. We hit an issue with lazy loading in classes that are only used during actual eviction, which apparently isn't hit by our tests.

So I had two options:
1. Fix the lazy loading of the caffeine classes by iterating the agent jar and forcing load of all classes.
2. Replace caffeine with another implementation that only uses pure JDK classes.

I opted for approach 2. I created a simple fifo capped concurrent hashmap, which simply removes the oldest element on the limit being reached. This also reduces dependencies and the jar file size.

Relates to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1257

### DIFF
--- a/pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/Agent.java
+++ b/pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/Agent.java
@@ -227,8 +227,7 @@ public class Agent {
 
     for (Class<?> clazz : instrumentation.getAllLoadedClasses()) {
       TypeDescription desc = new TypeDescription.ForLoadedType(clazz);
-      if (desc.getName().startsWith("io.opentelemetry.obi.")
-          || desc.getName().startsWith("com.github.benmanes.")) {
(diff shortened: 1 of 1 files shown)


pkg/internal/java/agent/build.gradle.kts
```diff
@@ -50,6 +49,8 @@ dependencies {
     testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:5.13.3")
 }
 
+tasks.register("prepareKotlinBuildScriptModel"){}
```
grcevski, line 52: Newer gradle versions complained about this not existing.

PR #1273, opened 2026-02-10
Introduce StatsO11y (statsolly) with tcp rtt sample metric

pinoOgni:
This PR introduces the appnetworktracer, which allows you to attach custom ebpf probes to calculate network metrics related to an instrumented service.

If the event is of interest to us, it is sent to userspace using a different ringbuf than the generic one.
The event is then converted to Span and the already established pipeline (decorator and others) is used (with some modifications).

The first metric implemented is the TCP RTT using tcp_close. This is an example metric, I plan to implement more and write documentation that allows anyone to add their own.


bpf/appnetworktracer/k_tcp.c
```diff
@@ -0,0 +1,82 @@
...
+#include <common/connection_info.h>
+#include <common/sockaddr.h>
+
+#include <logger/bpf_dbg.h>
+
+#include <maps/app_network_events.h>
+
+#include <pid/pid.h>
+
+#include "types.h"
```
rafaelroquetto: nit: even when the header is "local", we use the full "qualified" path, i.e.


```suggestion
#include <appnetworktracer/types.h>
```

bpf/appnetworktracer/k_tcp.c
```diff
@@ -0,0 +1,82 @@
...
+
+enum {
+    usec_per_sec = 1000000ULL,
+    max_srtt_allowed = 60 * usec_per_sec,
+};
```
rafaelroquetto, line 22:
```suggestion
enum {
    k_usec_per_sec = 1000000ULL,
    k_max_srtt_allowed = 60 * k_usec_per_sec,
};
```

just because we have been consistently using the prefix `k_` for constants (although we do have `UPPERCASE` macros - eventually we should normalise all of them to using enums and using `k_` or all going `UPPERCASE`)

bpf/appnetworktracer/k_tcp.c
```diff
@@ -0,0 +1,82 @@
...
+    pid_info pid_info;
+    u32 srtt;
+    connection_info_t conn;
+} app_net_tcp_rtt_t;
+
+SEC("kprobe/tcp_close")
+int BPF_KPROBE(obi_kprobe_tcp_close_rtt, struct sock *sk) {
+    (void)ctx;
+
+    u64 id = bpf_get_current_pid_tgid();
```
rafaelroquetto:
```suggestion
    const u64 id = bpf_get_current_pid_tgid();
```

bpf/appnetworktracer/types.h
```diff
@@ -0,0 +1,10 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build obi_bpf_ignore
+
+// These need to line up with some Go identifiers defined in the appnetworktracer.go file
+#pragma once
+enum {
+    event_app_net_tcp_rtt = 1, // EventTypeAppNetTcpRtt
```
rafaelroquetto:
```suggestion
    k_event_app_net_tcp_rtt = 1, // EventTypeAppNetTcpRtt
```

bpf/statsolly/maps/stats_events.h
```diff
@@ -0,0 +1,17 @@
+// Copyright The OpenTelemetry Authors
```
rafaelroquetto, line 1: since this map is private to `appnetworktracer`, you can move it to `appnetworktracer/maps` instead to be consistent with other tracers.

bpf/appnetworktracer/maps/app_network_events.h
```diff
@@ -0,0 +1,17 @@
...
+
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_helpers.h>
+
+struct {
+    __uint(type, BPF_MAP_TYPE_RINGBUF);
+    __uint(max_entries, 1 << 16);
+} app_network_events SEC(".maps");
+
+static __always_inline long app_network_events_flags() {
```
rafaelroquetto: genuine (non-rhetorical) question: why do we need a specialised version of `get_flags()` only for this, instead of using `get_flags()`?

rafaelroquetto (commented):
I know this is a draft, so I know this is subject to change, but thought I'd offer early feedback.

Out of curiosity, why did you opt to use a separate ring buffer other than using the shared one? A separate ring buffer is the more intuitive approach for me, but I wonder if the shared ring buffer exists for another good reason.

bpf/appnetworktracer/maps/app_network_events.h
pinoOgni: The `get_flags()` function is working on the `events` ebpf ringbuf map while `app_network_events_flags` is working on the `app_network_events` map.

bpf/statsolly/maps/stats_events.h
pinoOgni: Right, thanks!

bpf/appnetworktracer/types.h
pinoOgni: Right, "k_" for kprobe!

bpf/appnetworktracer/maps/app_network_events.h
rafaelroquetto: Oh right! Makes sense - perhaps we shouldn't hardcode 4096 and make this configurable/query the available data in the ringbuffer. Or do something like this: https://github.com/rafaelroquetto/opentelemetry-ebpf-instrumentation/blob/1ac990697a904212ead446f62814e69b55a90fcc/bpf/netolly/flows_sock.c#L75

(this will take into account a flush period and the ring buffer capacity).

mmat11: I previously suggested to hardcode it in order to avoid adding complexity and not query it every time -- the ringbuffers are now automatically flushed after a deadline so I thought it would be enough to not send a notification for every single event and 4KB should be big enough to contain a lot of these network events in a batch

rafaelroquetto: @mmat11 cool, makes sense to keep it simple, thanks for clarifying!

devdocs/app-network-tracer.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+And the same metric in a Kubernetes environment:
+
+```
+obi_net_tcp_rtt_seconds_bucket{dst_address="192.168.187.167",dst_name="go-client-3-deployment-795484488d-nhfm9",dst_port="",dst_zone="us-east-2c",instance="default.go-server-3-deployment-674fb9748f-hpzrf.go-server",job="default/go-server-3-deployment",k8s_cluster_name="",k8s_dst_name="go-client-3-deployment-795484488d-nhfm9",k8s_dst_namespace="default",k8s_dst_node_ip="192.168.177.102",k8s_dst_node_name="i-08e6fe8a9b5968e8a",k8s_dst_owner_name="go-client-3-deployment",k8s_dst_owner_type="Deployment",k8s_dst_type="Pod",k8s_src_name="go-server-3-deployment-674fb9748f-hpzrf",k8s_src_namespace="default",k8s_src_node_ip="192.168.177.102",k8s_src_node_name="i-08e6fe8a9b5968e8a",k8s_src_owner_name="go-server-3-deployment",k8s_src_owner_type="Deployment",k8s_src_type="Pod",service_namespace="default",src_address="192.168.187.168",src_name="go-server-3-deployment-674fb9748f-hpzrf",src_port="",src_zone="us-east-2c",le="0.001"} 7
```
pinoOgni: Need to update this example with `dst_port` and `src_port`.

devdocs/app-network-tracer.md
```diff
@@ -0,0 +1,40 @@
...
+- [Example](#example)
+- [Add a new application network metric](#add-a-new-application-network-metric)
+- [Notes on attributes](#notes-on-attributes)
+
+## Example
+
+In a non Kubernetes environment:
+
+```
+obi_net_tcp_rtt_seconds_bucket{dst_address="127.0.0.1",dst_name="",dst_port="",dst_zone="",instance="lima-ubuntu-ebpf:651294",job="main",service_namespace="",src_address="127.0.0.1",src_name="",src_port="",src_zone="",le="0.5"} 8
```
pinoOgni: Need to update this example with `dst_port` and `src_port`.

bpf/appnetworktracer/maps/app_network_events.h
```diff
@@ -0,0 +1,17 @@
...
+    long avail_data = bpf_ringbuf_query(&app_network_events, BPF_RB_AVAIL_DATA);
```
rafaelroquetto:
```suggestion
    const u64 avail_data = bpf_ringbuf_query(&app_network_events, BPF_RB_AVAIL_DATA);
```

otherwise we are mixed unsigned and signed types and `long` is not guaranteed to be 64-bit (although it is for our platforms I suppose)

pinoOgni: Hi @rafaelroquetto fine by me.
We should update `get_flags()` and `log_events_flags()` as well. Wdyt?

rafaelroquetto: Good idea!

bpf/appnetworktracer/k_tcp.h
```diff
@@ -0,0 +1,70 @@
...
+    app_net_tcp_rtt_t *se = bpf_ringbuf_reserve(&app_network_events, sizeof(*se), 0);
+    if (!se) {
+        return;
+    }
+    se->flags = k_event_app_net_tcp_rtt;
+    task_pid(&se->pid_info);
+    se->srtt = srtt / 1000; // convert to millisecond
+    se->conn = *conn;
+
+    bpf_d_printk("pid %u, src port %d, dst port %d, srtt %d",
```
grcevski: Oh this is going to get you on kernel 5.15. That's why maybe the tests are failing, we can only print 3 arguments at a time.

pkg/appolly/discover/attacher.go
```diff
@@ -201,9 +204,13 @@ func (ta *traceAttacher) getTracer(ie *ebpf.Instrumentable) bool {
 			return ta.reuseTracer(ta.reusableTracer, ie)
 		}
 		programs = ta.withCommonTracersGroup(newGenericTracersGroup(ta.EbpfEventContext.CommonPIDsFilter, ta.Cfg, ta.Metrics))
+		isGenericTracerActive = true
 	default:
 		ta.log.Warn("unexpected instrumentable type. This is basically a bug", "type", ie.Type)
 	}
+	if ta.Cfg.AppNetworkMetrics.Enabled {
```
grcevski: Ah this might be problematic with the suggestion I made to combine the tcp_close probes. We might load the generic tracer later, the first application might be Go and then the network tracer will trigger and then the generic tracer will load and we'll duplicate the event.

I didn't think of how we could enable this when I suggested the other change.

I guess you want to handle the case when it could be just Go apps + network.

I think we either need to have this reverted back to your original approach, or add a new probe in Go. I think we should go back to two probes on tcp_close if network application monitoring is enabled.

However, I came up with another problem when thinking about this and I think we need to come up with another solution.

Go applications don't use the PID filter, since we attach to the binary, not to the kernel. So technically, the network monitoring will skip those applications now if you have generic filter enabled. The generic PID filter avoids Go applications, so we don't process those events in vain.

I think the best approach would be to do what Mattia did with the log enricher, have a custom PID filter for just this network application tracer, separate all probes and make it filter both Go and non-Go apps.

pkg/internal/ebpf/appnetworktracer/appnetworktracer.go
```diff
@@ -0,0 +1,298 @@
...
+
+func pidSegmentBit(k uint64) (uint32, uint32) {
+	h := uint32(k % primeHash)
+	segment := h / 64
+	bit := h & 63
+
+	return segment, bit
+}
+
+func (p *Tracer) buildPidFilter() []uint64 {
```
grcevski: If we are reusing this kind of PID filter, might be possible to common the code into a single implementatoin.

grcevski (commented):
I took another pass at reviewing this and I think we might have some outstanding issues with how the tracers react with the PID filter.

I was wondering though after I wrote all of this, would it be simpler if we made this metrics part of the network tracer? Do we see potentially people not caring about TCP resets for all applications and they'd like to filter based on applications?

If we made this part of the network tracer, things will get much simpler to implement and if people really wanted to filter some events they can use the OBI filtering syntax or use the collector to filter. I honestly believe this might belong in the network tracer code.

bpf/appnetworktracer/k_tcp.h
pinoOgni: Oh nice catch, I was using it for tests! Thanks

pkg/internal/ebpf/appnetworktracer/appnetworktracer.go
pinoOgni: Yep, I left a todo, I thought is was too much for a single PR.

pkg/appolly/discover/attacher.go
pinoOgni: > Ah this might be problematic with the suggestion I made to combine the tcp_close probes. We might load the generic tracer later, the first application might be Go and then the network tracer will trigger and then the generic tracer will load and we'll duplicate the event.

Yep, right. I didn't think about that! We may do a double check to have only one event but I don't think is the best idea.

> 
> I didn't think of how we could enable this when I suggested the other change.
> 
> I guess you want to handle the case when it could be just Go apps + network.
> 
> I think we either need to have this reverted back to your original approach, or add a new probe in Go. I think we should go back to two probes on tcp_close if network application monitoring is enabled.

Sure.

> 
> However, I came up with another problem when thinking about this and I think we need to come up with another solution.
> 
> Go applications don't use the PID filter, since we attach to the binary, not to the kernel. So technically, the network monitoring will skip those applications now if you have generic filter enabled. The generic PID filter avoids Go applications, so we don't process those events in vain.

I ran some tests with Go and didn't encounter this problem. Thanks, Nikola! I'll check it.

> 
> I think the best approach would be to do what Mattia did with the log enricher, have a custom PID filter for just this network application tracer, separate all probes and make it filter both Go and non-Go apps.

So the final idea is to revert the networktracer to the way I did it, but change the pid management. I previously used the generictracer's, now I need to have the networktracer manage PIDs. Finally, even if both the generictracer and the networktracer are loaded together, there will be two separate probes (in this case, tcp_close for rtt).

PR #1279, opened 2026-02-10
Fix go duplicate traceparent in injected headers

grcevski:
This fixes two pretty nasty edge cases in our traceparent injection:
~~1. When the new code was looking for the traceparent it might inject it even if it didn't find `\r\n\r\n`.~~ This needs another fix, otherwise we miss many parents. We need to improve the de-duplication with split packets.
2. We weren't resetting the buffer for split packets, TCP continuations, so if the reading of the kernel buffer failed, we'd have stale buffer and the detection of is_http would succeed on non header data, causing double injection of the traceparent in a middle of the body.

I exposed this by accident with self instrumentation of OBI. I used the old style `services` definition for criteria, which wasn't excluding obi and ebpf-instrument. So we'd damage the OBI export of traces and traces wouldn't be sent to the collector.

This has been there for a while, however with the recent performance improvements to make the ring buffer send large packets, we make larger payloads for the OTLP export, which caused more split packets and exposed this bug.

There's one follow-up fix we need to make. We have detection for non-go that we've extended this packet before, such that it doesn't trigger on payloads that have HTTP prefixes inside, so split packets can still cause issues but it would be very rare. I'll follow-up with that once I find a good solution for it.

### DIFF
--- a/bpf/tpinjector/tpinjector.c
+++ b/bpf/tpinjector/tpinjector.c
@@ -868,10 +868,7 @@ int obi_packet_extender_find_existing_tp(struct sk_msg_md *msg) {
     const u32 data_size = (e - ptr) & 0x3ff; // 1KB chunks per iteration
 
     for (u32 i = 0; i < data_size; ++i) {
-        if (ptr + TP_SIZE >= e) {
-            return SK_PASS;
-        }
-        if (is_eoh(ptr)) {
+        if ((ptr + TP_SIZE >= e) || is_eoh(ptr)) {
             bpf_tail_call_static(msg, &extender_jump_table, k_tail_create_tp);
             break;
         }


skl (approved):

bpf/tpinjector/tpinjector.c
```diff
@@ -475,6 +475,7 @@ static __always_inline u8 protocol_detector(struct sk_msg_md *msg,
         return 0;
     }
 
+    msg_ptr[0] = 0;
```
mmat11, line 478: question: isn't this getting overwritten in the next line?

grcevski: Yes and that's the expected case, however if that read fails, the buffer is from the previous packet, so we think it's a start of the http request and slap another traceparent in a middle of a payload

PR #1281, opened 2026-02-11
feat: add podname to env for filtering when running OBI as a sidecar

### LINKED ISSUE #1240: Filtering on pod via environment variables when running OBI as a sidecar
when running OBI in sidecar mode, the [example](https://opentelemetry.io/docs/zero-code/obi/setup/kubernetes/#deploy-obi-as-a-sidecar-container) shows matching the process to instrument either via port (OTEL_EBPF_OPEN_PORT) or executable path (OTEL_EBPF_AUTO_TARGET_EXE). that can work until you duplicate this configuration many times across many pods

ideally you would want the OBI sidecar to instrument only processes on the current pod, OBI supports this via config map in `discovery.instrument.[0].k8s_pod_name`, but not via environment variable like port and path

my suggestion is to add something similar to OTEL_EBPF_OPEN_PORT and OTEL_EBPF_AUTO_TARGET_EXE for pod name, for example `OTEL_EBPF_K8S_POD_NAME`. which will allow to define repeatable configuration for sidecar deployments

for example:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
spec:
  containers:
    - name: obi
      image: otel/ebpf-instrument
      env:
        - name: OTEL_EBPF_K8S_POD_NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
```

itsCheithanya:
## Checklist
fixes #1240

### DIFF
--- a/pkg/appolly/discover/matcher.go
+++ b/pkg/appolly/discover/matcher.go
@@ -359,13 +359,14 @@ func FindingCriteria(cfg *obi.Config) []services.Selector {
 
 	if len(cfg.Discovery.Instrument) > 0 {
 		finderCriteria := cfg.Discovery.Instrument
-		if cfg.AutoTargetExe.IsSet() || cfg.Port.Len() > 0 {
+		if cfg.AutoTargetExe.IsSet() || cfg.Port.Len() > 0 || cfg.PodName.IsSet() {
 			finderCriteria = slices.Clone(cfg.Discovery.Instrument)
 			finderCriteria = append(finderCriteria, services.GlobAttributes{
 				Name:      cfg.ServiceName,
 				Namespace: cfg.ServiceNamespace,
 				Path:      cfg.AutoTargetExe,
 				OpenPorts: cfg.Port,
+				Metadata:  map[string]*services.GlobAttr{services.AttrPodName: &cfg.PodName},
 			})
 		}
 		return NormalizeGlobCriteria(finderCriteria)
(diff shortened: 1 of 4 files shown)


pkg/appolly/discover/matcher.go
mmat11, line 354: should we also add it here for consistency?

NimrodAvni78:
> Good proposal! I'd be happy to merge it once the tests pass. But I'd also add an integration test, or at least modify one of the current sidecar tests to make use of this.
> 
> I also have a question, why using this option and not directly `OTEL_EBPF_AUTO_TARGET_EXE=*`? As long as it runs as a sidecar, it will only instrument the processes of that pod, which is equivalent to the proposed action, and less error-prone since you can't assign a pod outside of your sidecar.

If OBI it runs as a sidecar does it only have access to pod processes?
if so maybe i was mistaken and doing `OTEL_EBPF_AUTO_TARGET_EXE=*` would solve the issue, maybe we should document this in the example since i think most of the time its to go to scenario

itsCheithanya:
Yeah I agree with @mariomac ,Just thinking if can it be useful when OBI runs as a DaemonSet which needs to filter on pod basis?🤔

mariomac:
Hi @NimrodAvni78 , by default, it will only see Pod's processes because it shares the Pod's PID namespace.

... unless you configure that pod to access to the whole host namespace (same as we currently do for OBI pods).

NimrodAvni78:
Ok, then I think it's fine to close it, maybe we should update the docs to add an example with this since this is probably the most common use case

grcevski:
> [itsCheithanya/opentelemetry.io@54f6d23](https://github.com/itsCheithanya/opentelemetry.io/commit/54f6d23315a0fc10ffe1b541824cb2a7b11f9e98)

This looks good to me! Why don't you make the PR for the opentelemetry docs :).

PR #1300, opened 2026-02-13
AWS cloud metadata decoration

mariomac:
## Checklist

- [x] AWS metadata
- [x] Remove old "host-id" fetching code
- [x] Integration tests AWS

The code also includes some methods for retrieving GCP and Azure metadata, but they will be properly tested in other PRs, to keep the focus of this PR.

### DIFF
--- a/pkg/appolly/meta/meta_node_kube.go
+++ b/pkg/appolly/meta/meta_node_kube.go
@@ -6,17 +6,23 @@ package meta // import "go.opentelemetry.io/obi/pkg/appolly/meta"
 import (
 	"context"
 	"fmt"
+	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"go.opentelemetry.io/obi/pkg/kube"
 )
 
+const kubeTimeout = 30 * time.Second
+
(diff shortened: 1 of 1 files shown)


pkg/appolly/meta/meta_node_cloud.go
```diff
@@ -0,0 +1,51 @@
...
+		"detector", fmt.Sprintf("%T", detector)[1:])
+
+	return func(ctx context.Context) (NodeMeta, error) {
+		// we expect very short response time in a cloud environment
+		ctx, cancel := context.WithTimeout(ctx, connectionTimeout)
+		defer cancel()
+		resource, err := detector.Detect(ctx)
+		// none of the errors from the ec2 detect are retriable, so we just log them.
+		if err != nil {
+			log.Debug("failed to detect AWS EC2 metadata", "error", err)
```
grcevski: I think this is common for all cloud vendors, perhaps it shouldn't say AWS?

mariomac: Ohh good catch. I refactored this part to make it generic and I forgot to change some EC2 references. Thanks!

Issue #1302, opened 2026-02-13
Add language instrumentation selector


grcevski:
We currently have number of application instrumentation selection criteria, for example:
- executable name (path)
- open ports
- kubernetes metadata selectors, e.g names, labels, annotations

Thinking forward with integrating OBI with the OpenTelemetry Injector, especially in non K8S environments, it would be useful to provide blanket instrumentation capability to instrument services written in a specific programming language. For example, the OpenTelemetry Injector currently supports Java, .NET, Nodejs and Python. It would be nice to create a combined package (with the new packaging SIG) that will bundle OBI with the injector, where we can instruct OBI to instrument Go and Rust based services, thus providing a complete automatic instrumentation for end users.

There are few things that need to be implemented for this to be done well:
1. Add the language based selector capability in the selection criteria for the matcher pipeline.
2. Add better language detection. We currently have good language detection for various services, but we can improve it for few languages and add some new ones.

Relates to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1140

PR #1304, opened 2026-02-13
Add language type detection pipeline before matcher

grcevski:
This PR adds language decoration pipeline component to run before the process matcher component, so that we can implement the language selection criteria.

I added a new config option to exclude common Linux system paths that shouldn't be investigated for language, mainly because I was worried about the overhead of parsing all executables to find what language they are, even with our `fastelf` library.

With this PR I did couple of things:
- I fixed a bug in the language detection for Go programs. It's unlikely, but we can't assume a Go program will not be called `node`. So I made the cheaper module path based detector run only for shared libraries, and then the more complete detector runs after we've looked at the binary symbols.
- I added C++ language detection, which runs after all others because most runtimes are implemented in C++ (e.g. node and hotspot jvm)
- I added the detector in the pipeline which is unused for now.

I measured the impact on startup with what I've done and I can't see visible regression (the process is single threaded) so logging wall-clock time should be sufficient to tell. I tested on my host system, so a lot of system services and I ran bunch of them in kubernetes and docker compose to make sure there are a lot of processes.

Without the language decorator timings, starting the pipeline to first instrumented process:
```
time=2026-02-13T16:28:09.498-05:00 level=INFO msg="Starting main node" component=obi.Instrumenter
time=2026-02-13T16:28:11.008-05:00 level=INFO msg="instrumenting process" component=discover.traceAttacher cmd=/usr/bin/ruby3.2 pid=1430026 ino=34086778 type=ruby service="" logenricher=false
```

With language decorator timings, starting the pipeline to first instrumented process:
```
time=2026-02-13T16:26:20.584-05:00 level=INFO msg="Starting main node" component=obi.Instrumenter
time=2026-02-13T16:26:22.303-05:00 level=INFO msg="instrumenting process" component=discover.traceAttacher cmd=/usr/bin/ruby3.2 pid=1430026 ino=34086778 type=ruby service="" logenricher=false
```

Both are under 2 seconds, with the decorator maybe 200ms slower. I think this is acceptable for being able to instrument based on the detected language.

Relates to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1302

### DIFF
--- a/internal/test/oats/http/yaml/oats_java_tls_sync_async.yaml
+++ b/internal/test/oats/http/yaml/oats_java_tls_sync_async.yaml
@@ -13,7 +13,7 @@ expected:
           attributes:
             server.port: "8443"
             url.path: "/greeting"
-    - traceql: '{ .url.full = "https://example.com/" }'
+    - traceql: '{ .url.full = "https://google.com/" }'
       spans:
         - name: "GET /"
           attributes:


pkg/appolly/discover/language_decorator.go
```diff
@@ -0,0 +1,93 @@
...
+type languageDecorator struct {
+	in           <-chan []Event[ProcessAttrs]
+	out          *msg.Queue[[]Event[ProcessAttrs]]
+	typeCache    *lru.Cache[uint64, svc.InstrumentableType]
+	log          *slog.Logger
+	ignoredPaths []string
+}
+
+func (ld *languageDecorator) isIgnoredPath(exePath string) bool {
+	for _, prefix := range ld.ignoredPaths {
```
MrAlias, line 51: Do we expect this to be a relatively small number of paths? Or should we look into a binary search after sorting this list on creation, or use a trie?

pkg/appolly/app/svc/svc.go
```diff
@@ -18,7 +18,8 @@ import (
 type InstrumentableType int
 
 const (
-	InstrumentableGolang InstrumentableType = iota + 1
+	InstrumentableUnknown InstrumentableType = iota + 1
```
MrAlias: Any reason to not make this `iota` so the zero value of `InstrumentableType` is `InstrumentableUnknown`?

grcevski: Good idea.

pkg/appolly/discover/language_decorator.go
grcevski: I expect a short list, I added 5 elements. I think it's fine unless some customer adds something very large here.

pkg/appolly/app/svc/svc.go
grcevski: Done

MrAlias (approved):

PR #1318, opened 2026-02-17
bpf: proposal: resize eBPF maps at configuration time

### LINKED ISSUE #1287: Resize eBPF maps at configuration time
UPDATE

While I was looking through the code, I found this [TODO](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d12e1612cf8060b28e16862585c524c4cbe41b1d/bpf/common/map_sizing.h#L6) there, all alone, and I thought it needed attention. :)

As discussed in the SIG, resize eBPF maps at configuration time can be useful for various reasons (such as memory efficiency because static large maps may be smaller in some cases and so on).

## How much freedom does the user need? (Update after internal discussion)

The User can potentially modify all maps. We can expose some functional dimensions in the configuration, something like:
* capacity: `max_concurrent_requests`, `max_custom_spans` and so on. Divided by maps.
* multiplier (as suggested by @florianl) to scale the dimensions by a factor (e.g., `0.5x` for some cases and `2x` for others). [Profiler reference](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/main/tracer/tracer.go#L539)

## Linked maps

Some maps could be mathematically linked. For example, if we increase map A, map B must grow proportionally. In this case we can give the possibility to the user to configure only the "Parent" dimension, and we calculate  automatically the "Child" map sizes to ensure internal consistency.

## Work done and next steps

I already did something. Right now, during loading, change the size of all maps with max_entries of the following values: MAX_CONCURRENT_REQUESTS, MAX_CONCURRENT_SHARED_REQUESTS, and MAX_CONCURRENT_CUSTOM_SPANS. 
Maps are divided by tracer. Each tracer must define the size they want for each map. The use can do:
```
ebpf:
  map_sizes:
    max_concurrent_requests: 9999
    ...
```
**Note**: there are maps shared between various tracers (even if they're perhaps not used, for example, `listening_ports` between `tpinjector` and `generictracer). This could be solved by cleaning up the ebpf code and making each tracer only use its own things or common data structures but not those of another.

If we agree on the approach, I can open a draft PR implementing these changes so we can discuss the implementation details there.

WIP solution:
```
ebpf:
    map_sizes:
         - listening_ports:
             max_entries: <size or multiplier>
         - kafka_state:
            max_entries: <size or multiplier>
```

FYI @rafaelroquetto @grcevski

pinoOgni:
Hi everyone, after various discussions and some needed refactors, this is my proposal to allow changing the size of ebpf maps during configuration (resolves #1287).

Example configuration:
```
log_config: yaml

open_port: "9092"
log_level: debug
prometheus_export:
  port: 9876

metrics:
  features:
    - network
    - stats
    - application

ebpf:
  maps_config:
    global_scale_factor: 1
```

In brief, we have a single `global_scale_factor` that allows us to change the size of all maps, with a few caveats:
1. The `.rodata`, `.bss`, and `.data` maps are not modified.
2. Maps with `max_entries < 64` are not modified, because the `global_scale_factor` can reach -3, so by setting the lower limit to 64, it can reach up to 8. We can decide also to modify it.
3. Various types of maps that I consider "special", such as `BPF_MAP_TYPE_PROG_ARRAY`, are not modified (I'm open to suggestions if the list needs to be changed).
4. If a map is pinned, the size is not updated, otherwise we would get an error. This is especially true for `traces_ctx_v1`; the size of this map will only be set the first time OBI is launched on the node.
5. All the resizable maps from appolly, statsolly, netolly and rdns can be modified.

Unfortunately, this is both the simplest solution and the one with the least customization. If we decide, we can also
- add greater granularity as in the old version where each map had an editable size (with its pros and cons) 
- or create editable map categories and not

Note: the `MaxMapEntries` is `1<<24`, so the `direct_flows` map which already has `max_entries = 1 << 24` will not grow in size.

### DIFF
--- a/pkg/internal/ebpf/convenience/convenience.go
+++ b/pkg/internal/ebpf/convenience/convenience.go
@@ -141,7 +141,7 @@ func SetupMapSizes(spec *ebpf.CollectionSpec, globalScaleFactor int, bpffsPinPat
 			continue
 		}
 
-		if bpffsPinPath != "" {
+		if bpffsPinPath != "" && mSpec.Pinning == ebpf.PinByName {
 			mapPath := filepath.Join(bpffsPinPath, name)
 			if _, err := os.Stat(mapPath); err == nil {
 				continue
--- a/docs/config-schema.json
+++ b/docs/config-schema.json
@@ -1142,7 +1142,7 @@
       "properties": {
         "global_scale_factor": {
           "type": "integer",
-          "description": "GlobalScaleFactor scales ALL maps that we expose by this factor If GlobalScaleFactor > 0 --> NewSize = DefaultSize << GlobalScaleFactor // double it If GlobalScaleFactor < 0 --> NewSize = DefaultSize >> GlobalScaleFactor // halve it"
+          "description": "GlobalScaleFactor scales map sizes in powers of two:   > 0: grows size (2x per step)   < 0: shrinks size (1/2 per step)   = 0: no change"
         }
       },
       "type": "object"
--- a/pkg/internal/ebpf/convenience/convenience_test.go
+++ b/pkg/internal/ebpf/convenience/convenience_test.go
@@ -144,21 +144,26 @@ func TestSetupMapSizes_SkipsBelowMinResizableMapSize(t *testing.T) {
 
 func TestSetupMapSizes_SkipsPinnedMaps(t *testing.T) {
 	pinDir := t.TempDir()
-	// Create a file to simulate a pinned map
-	pinnedPath := filepath.Join(pinDir, "pinned_map")
-	if err := os.WriteFile(pinnedPath, []byte{}, 0o600); err != nil {
-		t.Fatal(err)
+	// Create files to simulate pinned maps on bpffs
+	for _, name := range []string{"pinned_map", "not_pinned_map"} {
+		if err := os.WriteFile(filepath.Join(pinDir, name), []byte{}, 0o600); err != nil {
+			t.Fatal(err)
+		}
 	}
 
 	spec := makeSpec(map[string]*ebpf.MapSpec{
-		"pinned_map":   {Type: ebpf.Hash, MaxEntries: 1024},
-		"unpinned_map": {Type: ebpf.Hash, MaxEntries: 1024},
+		"pinned_map":     {Type: ebpf.Hash, MaxEntries: 1024, Pinning: ebpf.PinByName},
+		"not_pinned_map": {Type: ebpf.Hash, MaxEntries: 1024, Pinning: ebpf.PinNone},
+		"unpinned_map":   {Type: ebpf.Hash, MaxEntries: 1024},
 	})
 
 	SetupMapSizes(spec, 2, pinDir)
 
 	if spec.Maps["pinned_map"].MaxEntries != 1024 {
-		t.Errorf("pinned map should be skipped: got %d, want 1024", spec.Maps["pinned_map"].MaxEntries)
+		t.Errorf("PinByName map with file on disk should be skipped: got %d, want 1024", spec.Maps["pinned_map"].MaxEntries)
+	}
+	if spec.Maps["not_pinned_map"].MaxEntries == 1024 {
+		t.Error("PinNone map should be resized even if a file with same name exists on disk")
 	}
 	if spec.Maps["unpinned_map"].MaxEntries == 1024 {
 		t.Error("unpinned map should have been resized")


pinoOgni:
> Hey @pinoOgni, thank you for putting the thought and effort on this. I am starting to wonder if exposing the maps to the outer world (by the means of configuration), is such a good idea. For instance, quoting your example:
> 
> ```
>       outgoing_trace_map:
>         max_entries: 3456
> ```
> 
> Now the world knows that there's an `outgoing_trace_map`, and that makes be a bit nervous for two reasons, **the first being the most concerning one** IMHO:
> 
> 1. not only are we now exposing implementation detail: we are somehow committing to backwards compatibility. That feels very limiting from a development standpoint. For example, what if:
>    
>    * we want to rename the map (e.g. I'm working on renaming `incoming_trace_maps` to `incoming_tcp_opts`)
>    * get rid of the map altogether
>    * introduce new maps/tracers - now we need to expose those as well, and somehow let the users know (the users need to be cognizant about tracers or OBI internals somehow, see next point)
>    * overall, this suffers from the same problems affecting code with poor or no abstractions

Agree. There might be a half-solution which is to not have the maps hardcoded as variables but to use a string map but we will still have other problems.

> 2. The cognitive burden on the user becomes arguably unreasonable
>    
>    * the user needs to understand the role of each map to be able to properly tune them
>    * I can tell first hand that even I don't know what every map inside OBI does - I wouldn't be surprised if a user messed it up, misunderstanding it, and that creates more support pressure from our end

Totally agree.

> 3. We cannot rely on the premise of dividing maps by tracers: there _will be_ cases in which maps will be shared, forked, and maybe tracers won't be present/instantiated.

Agree.

> 
> So after giving this some thought, I think we need to swallow the cognitive burden ourselves and instead try to present to the user a more "rational" configuration experience: in particular, this the idea behind:
> 
> ```
> ebpf:
>   maps_config:
>     global_scale_factor: 2
> ```
> 
> is much more appetising as it abstracts all of the above on something more palatable to everyone (users and devs alike - it does not take our freedom and does not unreasonably impose on the user).

Fine by me. We should still document that not all maps are actually modified (for example, there are some with `max_entries` equal to 1 and I think they should remain that way)
> 
> But I wonder if we can go the extra mile and offer even more meaningful settings, such as:
> 
> ```
> ebpf:
>   max_concurrent_requests: 30000 # maximum number of requests OBI can track at once
> ```

This was my first idea, and then the desire to add things took over.

> or a different dimension affecting the same thing as an alternative:
> 
> ```
> ebpf:
>    max_alloc: 30MB # perhaps a bad example, but still easier to reason
> ```
> 
> I am happy to engage and try to map what those would be, if that helps.

Fine by me, I appreciate! :)

> 
> And I do agree we should probably revisit all of our ebpf-related settings - we've got a lot already going on - like an airplane cockpit, and a lot of them are obscure.

Totally agree.

bpf/common/map_sizing.h
```diff
@@ -3,8 +3,8 @@
 
 #pragma once
 
-// TODO: make this user-configurable and modify the value from the userspace when
-// loading the maps with the Cilium library
+// Default values that can be overrided at runtime and are needed to compile OBI
```
mmat11:
```suggestion
// Default values that can be overridden at runtime and are needed to compile OBI
```

pkg/config/ebpf_tracer.go
```diff
@@ -34,6 +34,13 @@ const (
 	StrContextPropagationTCP      = "tcp"
 )
 
+type MapsConfig struct {
+	// GlobalScaleFactor scales ALL maps that we expose by this factor
+	// If GlobalScaleFactor > 0 --> NewSize = DefaultSize << GlobalScaleFactor // double it
+	// If GlobalScaleFactor < 0 --> NewSize = DefaultSize >> GlobalScaleFactor // halve it
```
```suggestion
// GlobalScaleFactor scales map sizes in powers of two:
//   > 0: grows size (2x per step)
//   < 0: shrinks size (½ per step)
//   = 0: no change
```

pkg/internal/ebpf/convenience/convenience.go
```diff
@@ -101,6 +102,75 @@ func LoadSpec(spec *ebpf.CollectionSpec, objects any, constants map[string]any,
...
+		if mSpec.MaxEntries < MinResizableMapSize {
+			continue
+		}
+
+		if bpffsPinPath != "" {
+			mapPath := filepath.Join(bpffsPinPath, name)
+			if _, err := os.Stat(mapPath); err == nil {
+				continue
+			}
+		}
```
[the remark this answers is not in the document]
rafaelroquetto: @pinoOgni that's a very good point here

pkg/config/ebpf_tracer.go
pinoOgni: Yes, I like more your description!

pkg/internal/ebpf/convenience/convenience.go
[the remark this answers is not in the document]
pinoOgni: wow, yes!

PR #1321, opened 2026-02-17
feat: Allow selecting instrumentation by PIDs

damemi:
This makes it possible to instrument by passing a PID, or list of PIDs, directly through the OBI config.

As a next step, it would be nice to be able to dynamically add/remove PIDs during runtime

## Checklist

### DIFF
--- a/pkg/appolly/discover/matcher.go
+++ b/pkg/appolly/discover/matcher.go
@@ -212,13 +212,7 @@ func (m *Matcher) isExcluded(obj *ProcessAttrs, proc *services.ProcessInfo) bool
 func (m *Matcher) matchProcess(obj *ProcessAttrs, p *services.ProcessInfo, a services.Selector) bool {
 	log := m.Log.With("pid", p.Pid, "exe", p.ExePath)
 	if pids, ok := a.GetPIDs(); ok && len(pids) > 0 {
-		if !pidInList(p.Pid, pids) {
-			return false
-		}
-		// PID matches; if this is PID-only criteria, skip path/port checks
-		if !a.GetPath().IsSet() && !a.GetPathRegexp().IsSet() && a.GetOpenPorts().Len() == 0 {
-			return m.matchByAttributes(obj, a)
-		}
+		return pidInList(p.Pid, pids)
 	}
 	if !a.GetPath().IsSet() && !a.GetLanguages().IsSet() && a.GetOpenPorts().Len() == 0 && len(obj.metadata) == 0 {
 		pids, hasPIDs := a.GetPIDs()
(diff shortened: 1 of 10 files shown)


pkg/appolly/services/attr_pid.go
```diff
@@ -0,0 +1,43 @@
...
+	return func(_ func(string, StringMatcher) bool) {}
+}
+
+func (p *PidSelector) RangePodLabels() iter.Seq2[string, StringMatcher] {
+	return func(_ func(string, StringMatcher) bool) {}
+}
+
+func (p *PidSelector) RangePodAnnotations() iter.Seq2[string, StringMatcher] {
+	return func(_ func(string, StringMatcher) bool) {}
+}
```
mariomac: I think introducing a whole new `Selector` implementation type is not needed.

Analogously to what we do with the Open Ports attribute, you can just keep the `GetPID()` method in the `Selector` interface, plus implement it in the `RegexSelector` and `GlobAttributes` implementations, so any user can add the PID selection criteria to either the `discovery > instrument` or the (deprecated) `discovery > services`.

damemi: sgtm, updated to just keep GetPID in the Selector interface and added to RegexSelector and GlobAttributes

pkg/appolly/discover/matcher.go
```diff
@@ -211,9 +211,21 @@ func (m *Matcher) isExcluded(obj *ProcessAttrs, proc *services.ProcessInfo) bool
 
 func (m *Matcher) matchProcess(obj *ProcessAttrs, p *services.ProcessInfo, a services.Selector) bool {
 	log := m.Log.With("pid", p.Pid, "exe", p.ExePath)
+	if pids, ok := a.GetPIDs(); ok && len(pids) > 0 {
+		if !pidInList(p.Pid, pids) {
+			return false
+		}
+		// PID matches; if this is PID-only criteria, skip path/port checks
+		if !a.GetPath().IsSet() && !a.GetPathRegexp().IsSet() && a.GetOpenPorts().Len() == 0 {
```
grcevski: I'm wondering, why not just return if the PID matches and ignore the rest? We can put a validation rule on the selectors that will ensure that if you choose a PID, nothing else should be set.

I guess one reason to think this would need further selectors is if OBI  is namespaced, but in that case it can only see a subset of the PIDs. I think we should make it absolute, you've set a PID on this selector you match that one directly.

pkg/obi/config.go
```diff
@@ -48,6 +49,63 @@ const (
 	LogLevelError LogLevel = "ERROR"
 )
 
+// TargetPIDs is a list of process IDs to instrument. Supports YAML arrays (e.g. [1234, 5678]),
+// a single YAML number, or env/comma-separated (e.g. OTEL_EBPF_TARGET_PID=1234,5678).
+type TargetPIDs []uint32
```
grcevski: What do you think about reusing PortEnum and renaming it into something more appropriate? It supports something similar, e.g `80,443,8000-8999`. I don't mind the new format, but it creates an inconsistency in how the end users would define this.

pkg/appolly/discover/matcher.go
damemi: true, PID is a pretty explicit/atomic selector so it is weird to then scope up to other fields.

we might want to reevaluate selector criteria at some point, I could see reasons for using pid selector alongside others but for now I think just `return pidInList()` pretty much covers it here

pkg/obi/config.go
damemi: Sounds good to me, renamed PortEnum to IntEnum (wdyt?)

Issue #1291, opened 2026-02-12
Filtering should support <,> for numeric values.


pilhuhn:
Filtering traces for e.g `sql.client.duration`  would most of the time not work with the current `match` or `not_match` operators, as times or number of bytes rarely match an exact number, but it is more common to filter for "outliers" in the sense of very small or very large values.

mariomac:
Good Idea @pilhuhn !

What would you think that, for consistency with the open_port discovery selectors, using the same matchers (renaming [PortEnum/PortRange](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/appolly/services/criteria.go#L173-L181) to something like NumEnum, NumRange).

However I'm not 100% convinced because on one side it would provide consistent specification across the different number matchers, but on the other side looks a bit weird something like: `sql.client.duration: { match: 0-1000 }`

PR #1323, opened 2026-02-18
Add numeric operators for filters.

pilhuhn:
Addresses #1291

This introduces new numeric filter options  
* equals
* not_equals
* greater_than
* greater_equals
* less_than
* less_equals

To be used e.g. like the following:

```yaml
filter:
  application:
    server.port:
      equals: 8787
```

### DIFF
--- a/pkg/filter/attribute.go
+++ b/pkg/filter/attribute.go
@@ -103,18 +103,43 @@ func buildMatcher[T any](getters attributes.NamedGetters[T, string], attribute a
 	if err := def.Validate(); err != nil {
 		return m, err
 	}
-	if def.Match != "" {
+	switch {
+	case def.Match != "":
 		var err error
 		if m.Glob, err = glob.Compile(def.Match); err != nil {
 			return m, fmt.Errorf("invalid glob in match property: %w", err)
 		}
-	} else {
+	case def.NotMatch != "":
 		var err error
 		if m.Glob, err = glob.Compile(def.NotMatch); err != nil {
 			return m, fmt.Errorf("invalid glob in not_match property: %w", err)
 		}
 		m.Negate = true
+	default:
+		// use match-all as dummy for numeric-only comparisons
+		m.Glob = glob.MustCompile("*")
+
+		// Set up numeric comparisons
+		if def.Equals != nil {
+			m.Equals = def.Equals
+		}
+		if def.NotEquals != nil {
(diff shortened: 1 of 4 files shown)


pkg/filter/matcher.go
```diff
@@ -33,11 +78,21 @@ type MatchDefinition struct {
 	Match string `yaml:"match"`
 	// NotMatch stores the glob that a given attribute MUST NOT match to let the record pass
 	NotMatch string `yaml:"not_match"`
+	// Numerical comparison for e.g. http.code or timings
+	GreaterThan   float64 `yaml:"greater_than"`
```
grcevski: What do you think about using *float64 instead of float64? In that case we can tell if the option is not set and allow us to filter on 0. We often have cases where maybe by error we cannot capture the http.code and we write 0, so I can see folks wanting to set a filter on that.

pkg/filter/attribute.go
```diff
@@ -103,18 +103,43 @@ func buildMatcher[T any](getters attributes.NamedGetters[T, string], attribute a
...
-	} else {
+	case def.NotMatch != "":
 		var err error
 		if m.Glob, err = glob.Compile(def.NotMatch); err != nil {
 			return m, fmt.Errorf("invalid glob in not_match property: %w", err)
 		}
 		m.Negate = true
+	default:
+		// use match-all as dummy for numeric-only comparisons
+		m.Glob = glob.MustCompile("*")
```
grcevski, line 120: I think we need to add some statement that checks if none of these are set and report an error.

pkg/filter/matcher.go
grcevski: What do you think about making these *float64? Like this 
```
GreaterThan   *float64 `yaml:"greater_than,omitempty"`
```
That way we can test for the 0 port, 0 status code etc. Sometimes bugs in the eBPF code prevent us from finding the correct connection information so folks can filter on 0 server port etc.

grcevski: Actually I also wonder if we can make them ints? Technically, eBPF doesn't support floating point operations, it's not in the instruction set, so we can never get float value from there.

pilhuhn: I was originally thinking about floats for timings. On a closer look after your comment, I see that timings are either a float as string ("1.2ms") or int in µs, so we should be good here.

PR #1326, opened 2026-02-18
Add programming language selector

grcevski:
This PR implements the second part of the issue, to take the input from the language detector and use it to be able to select programming languages to instrument. This can be especially useful for deployment in non Kubernetes environments alongside the OpenTelemetry Injector or with the OpenTelemetry operator in daemonset mode.

Here's an example on how we can use this:
```
services:
  instrument:
    - language: {go,rust}
``` 

With this configuration option OBI will instrument all services found on the host that are implemented in `golang` or `rust`.

I'll follow-up with a docs PR for opentelemetry.io when the change is merged. 

Closes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1302

### DIFF
--- a/internal/test/integration/multiprocess_test.go
+++ b/internal/test/integration/multiprocess_test.go
@@ -324,7 +324,7 @@ func testPartialLanguageHTTPProbes(t *testing.T) {
 	}, testTimeout, 100*time.Millisecond)
 }
 
-func TestMultiProcessLanguage(t *testing.T) {
+func TestLanguageSelectors(t *testing.T) {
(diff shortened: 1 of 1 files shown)


grcevski:
> > Amazing! What about adding also an `OTEL_EBPF_AUTO_TARGET_LANGUAGE` environment variable?
> 
> Good idea! I'll follow-up with a PR on that. It can be handy.

Added here https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1331

PR #1327, opened 2026-02-18
ci: use arm runner for arm image build

skl:
The image builds for obi and obi-cache were taking [~45m](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/22142448947) and [~18m](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/22142448985) respectively, mostly due to the ARM images being built inside a QEMU ARM VM on x86 runners.

This PR instead uses native ARM runners in parallel with the standard x86.

Affected workflows:

- `pull_request_docker_build_test.yml` (PR build only, ARM previously skipped)
- `publish_dockerhub_main.yml` (main obi build and push)
- `publish_dockerhub_k8s_cache_main.yml` (main cache build and push)

Outputs:

- Same tags, registries, multi-arch manifest, labels, and signing as before.
- Adds arm64 on a native runner and removes the QEMU VM for image building.
- Only the build process itself has been optimised.

The expectation is that these workflows should be significantly faster (est. <10m).


MrAlias (approved):
🥇

PR #1334, opened 2026-02-19
Azure and GCP metadata decoration: integration tests

mariomac:
It also improves the AWS EC2 tests by removing the need of mocking the `AWS_EC2_METADATA_SERVICE_ENDPOINT` variable. Now the fake IMDS is reachable at the standard `169.254.169.254` IP.

It also uses the low-level docker client to improve some aspects of `docker_util_test`:
* Enables container log reports after integration tests
* We need to assign the `obi` network alias to OBI so the Prometheus scraping work. The `dockerutil` high-level API required that the network was created shortly after the OBI container was created. This caused that sometimes it retrieved for metadata before the network was ready, causing the tests to fail.

### DIFF
--- a/internal/test/integration/cloud_aws_test.go
+++ b/internal/test/integration/cloud_aws_test.go
@@ -25,7 +25,7 @@ func TestCloudResourceMetadata_AWS(t *testing.T) {
 	setupContainerJaeger(t, network)
 	setupContainerCollector(t, network, "otelcol-config.yml")
 	defer network.Close()
-	testserver := setupGoOTelTestServer(t, network, nil)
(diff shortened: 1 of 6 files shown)


mariomac:
@grcevski concretely the Azure test, which works locally but in the GitHub runner, which works in Azure, it contacts to the real Azure metadata endpoint instead of the mocked endpoint. Unlike the other OTEL resource detectors, the Azure detector doesn't seem to allow overriding the Azure endpoint.

I need to have a look at the logs and see if I can configure better the Docker network. I have had to implement logging report in the new Go-based `dockertest` code.

PR #1341, opened 2026-02-20
refactor: apply `const` to bpf variables 

pinoOgni:
This is the first phase of what I hope will be a larger refactoring. 
This PR updates variable declarations to use the `const` across the following bpf directories:
- bpf/gotracer
- bpf/common
- bpf/generictracer

Add `const` is a best practice that offers various benefits like readability, intent and safety. It should also add compiler optimization and help the verifier.

I find the code much more readable this way, I hope it works for you.

### DIFF
--- a/bpf/common/dns.h
+++ b/bpf/common/dns.h
@@ -148,7 +148,7 @@ static __always_inline u8 handle_dns(struct __sk_buff *skb,
 
         // The data offset field in the header is specified in 32-bit words. We
         // have to multiply this value by 4 to get the TCP header length in bytes.
-        __u8 tcp_header_len = tcph.doff * 4;
+        const u8 tcp_header_len = tcph.doff * 4;
 
         // DNS is after the TCP header and the 2 bytes of the length of the DNS packet
         const u16 size_bytes_len = 2;
--- a/bpf/common/trace_common.h
+++ b/bpf/common/trace_common.h
@@ -52,8 +52,7 @@ static __always_inline tp_info_pid_t *tp_buf() {
 static __always_inline void trace_key_from_pid_tid(trace_key_t *t_key) {
     task_tid(&t_key->p_key);
 
-    const u64 extra_id = extra_runtime_id();
-    t_key->extra_id = extra_id;
+    t_key->extra_id = extra_runtime_id();
 }
 
 static __always_inline void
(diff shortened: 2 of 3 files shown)


bpf/common/dns.h
mmat11:
```suggestion
        const u8 tcp_header_len = tcph.doff * 4;
```

this should also be const I believe; did you run some tool or script to find all of them?

bpf/generictracer/k_tracer.c
```diff
@@ -275,7 +274,7 @@ static __always_inline void setup_cp_support_conn_info(pid_connection_info_t *p_
     }
 
     task_tid(&ct.t_key.p_key);
-    u64 extra_id = extra_runtime_id();
+    const u64 extra_id = extra_runtime_id();
```
mmat11: I think we can inline all these `extra_id` instances, not sure we should do it in this PR tho

bpf/common/dns.h
pinoOgni: I tried `clang-tidy` but it was not working (maybe I missed something) so I just used this regex in vsocde `^((?!const).)*\b(u8|u16|u32|bool|u64|long)\b\s+\w+\s*=` to search.

bpf/generictracer/k_tracer.c
pinoOgni: why not, I'll do it

bpf/common/dns.h
```diff
@@ -148,7 +148,7 @@ static __always_inline u8 handle_dns(struct __sk_buff *skb,
 
         // The data offset field in the header is specified in 32-bit words. We
         // have to multiply this value by 4 to get the TCP header length in bytes.
-        __u8 tcp_header_len = tcph.doff * 4;
+        const __u8 tcp_header_len = tcph.doff * 4;
```
mmat11:
```suggestion
        const u8 tcp_header_len = tcph.doff * 4;
```

bpf/common/trace_common.h
```diff
@@ -52,15 +52,15 @@ static __always_inline tp_info_pid_t *tp_buf() {
 static __always_inline void trace_key_from_pid_tid(trace_key_t *t_key) {
     task_tid(&t_key->p_key);
 
-    u64 extra_id = extra_runtime_id();
-    t_key->extra_id = extra_id;
+    t_key->extra_id = extra_runtime_id();
+    ;
```
mmat11:
```suggestion
```

pinoOgni: ops

PR #1342, opened 2026-02-22
Add SendTimeout field in kubecache config

pinoOgni:
This PR allows to set the `SendTimeout` value of the `k8s-cache` component during configuration. 
It also adds some config validations.


mmat11 (approved):
can you rebase and run `make generate-config-schema`?

PR #1350, opened 2026-02-23
ci: shard unit tests

skl:
Unit test workflow was taking ~10 minutes, let's see if sharding can help:

- Split the one-huge-job "PR checks" job into several jobs and steps
- Run lint in parallel with the tests
- Re-use the docker-generate caching from the lint-darwin workflow
- Re-use the go caching from various workflows
- Use a simple sharding strategy based on `go list`

Should be able to see the run time on this PR:

- https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/22314090260?pr=1350

FYI: There's a [GitHub Actions incident](https://www.githubstatus.com) currently, so it's making it difficult to measure timings as runners are taking a while to pick up jobs.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -20,12 +20,59 @@ permissions:
   contents: read
 
 jobs:
-  test:
-    permissions:
-      # Required for codecov
-      checks: write
-      pull-requests: write
-    name: test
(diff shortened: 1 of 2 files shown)


grcevski (approved):
LGTM!

PR #1351, opened 2026-02-23
Config v2.0

MrAlias:
- Design plan methodology for v2.0 of OBI config
- JSON Schema for v2.0
- Migration plan
- Examples and tooling


devdocs/config/version-2.0/examples/default-configuration.yaml
```diff
@@ -0,0 +1,562 @@
...
+                - /lib/systemd/*
+                - /usr/lib/systemd/*
+                - /usr/libexec/*
+                - /sbin/*
+                - /usr/sbin/*
+
+    instrumentation:
+      # HTTP protocol instrumentation and HTTP-specific enrichment/parsing.
+      http:
+        enabled:
```
NimrodAvni78: don't we already define this under `tracer_provider->instrumentations` and `meter_provider->instrumentations`?
same for other protocols

devdocs/config/version-2.0/examples/default-configuration.yaml
```diff
@@ -0,0 +1,562 @@
...
+          elasticsearch:
+            enabled: false
+          aws:
+            enabled: false
+          sqlpp:
+            enabled: false
+            endpoint_patterns:
+              - /query/service
+        # HTTP log enrichment controls.
+        log_enrichment:
```
NimrodAvni78: should probably not be under http or instrumentations, but a different feature entierly

devdocs/config/version-2.0/examples/default-configuration.yaml
```diff
@@ -0,0 +1,562 @@
...
+          # Services for which log enrichment is enabled.
+          services: []
```
NimrodAvni78: should probably be similar to the `selection` segmant above

devdocs/config/version-2.0/examples/default-configuration.yaml
```diff
@@ -0,0 +1,562 @@
...
+          metrics: true
+        filters:
+          traces: {}
+          metrics: {}
+        debug:
+          enabled: false
+          bytecode_instrumentation: false
+        attach_timeout: 10s
+      # Node.js instrumentation toggle.
+      nodejs:
```
NimrodAvni78: maybe we should seperate go and nodejs to another section

nodejs on/off only effects if to inject the small program via the SIGUSR1 signal, so not real sense to have it as enabled per signal type and filters

go maybe makes sense to have some programs use the go instrumentations some the generic ones

devdocs/config/version-2.0/examples/default-configuration.yaml
```diff
@@ -0,0 +1,562 @@
...
+          max_duration: 5m0s
+        bpf_filesystem:
+          path: /sys/fs/bpf/
+      # Logging behavior and startup config dump.
+      logging:
+        level: INFO
+        startup_dump: ""
+        debug_trace_output: disabled
+      # Optional profiling endpoint.
+      profiling:
```
NimrodAvni78, line 590: not sure if super related to the new config initiative, but
for profiling and internal metrics, we can think on how to use the existing instrastructure in the collector (pprof extension and internal telemetry) while allowing to also run OBI as independent binary

NimrodAvni78 (commented):
nice!
i know its in draft but wanted to give some input as well

devdocs/config/version-2.0/examples/default-configuration.yaml
MrAlias: Ah, that's a mistake. The `tracer_provider` and `meter_provider` should not have `instrumentation`. They should be vanilla OTel declarative configuration properties. I'll fix that.

MrAlias: Updated. The missing refactor has been moved into the `obi` section.

devdocs/config/version-2.0/examples/default-configuration.yaml
MrAlias: I have moved into its own section under the new top-level `correlation` section.

devdocs/config/version-2.0/examples/default-configuration.yaml
MrAlias: I have refactored to follow the principle "One concern, one place" and have the log_trace_annotation inherit the selection of services and then have the ability to filter down from there, similar to the reset of the config.

devdocs/config/version-2.0/examples/default-configuration.yaml
MrAlias: I've moved Node.js, Go, and Java to a new top-level "runtimes" section that defines how language runtimes are instrumented. PTAL.

devdocs/config/version-2.0/examples/default-configuration.yaml
MrAlias: What do you think about removing these from the OBI config and moving into the command options/args?

Meaning the `obi` command would take these as flags and setup appropriate profiling/internal_metrics and then, since they are not in the config, the collector receiver would just assume the user will configure these profiling/interna metric settings in the appropriate part of the collector configuration.

MrAlias:
> One requirement that we are frequently seeing at Grafana is configuration per service.

@fstab yes, that should be possible. As I have designed the configuration, something like this would address the user concern:

```yaml
extensions:
  obi:
    version: "2.0"
    selection:
      policy:
        # Exclude all services not matched
        default_action: exclude
      rules:
        - action: include
          match:
            process:
              exe_path_glob:
                - "/path/to/my_service"
```

devdocs/config/version-2.0/examples/default-configuration.yaml
```diff
@@ -0,0 +1,554 @@
...
+        # Include by default unless matched by an exclude rule.
+        default_action: include
+        # First matching rule decides outcome.
+        match_order: first_match_wins
+        # Process re-scan interval (0s uses implementation default/event-driven behavior).
+        poll_interval: 0s
+        # Ignore very new processes until they are old enough to evaluate.
+        min_process_age: 5s
+      # Ordered include/exclude rules.
+      rules:
```
grcevski, line 99: I think we need to allow for defining overrides from the defaults per selection criteria. Initially when we built the config, all selection was global, but overtime we've had the need to override for given `rules` of the selection criteria.

One example is overriding what telemetry is exported for a `rule`, e.g. maybe globally we want just metrics, but for this subset we want traces. We also allow folks to override the sampler config for a given `rule`, maybe some namespace is not very important, so it's OK to sample with 1%.

Right now we have these:
```
	// Configures what to export. Allowed values are 'metrics', 'traces',
	// or an empty array (disabled). An unspecified value (nil) will use the
	// default configuration value
	ExportModes ExportModes `yaml:"exports"`

	SamplerConfig *SamplerConfig `yaml:"sampler"`

	Routes *CustomRoutesConfig `yaml:"routes"`

	// Metrics configuration that is custom for this service match
	Metrics perapp.SvcMetricsConfig `yaml:"metrics" env:"-"`
```

grcevski (commented):
This is fantastic! I really like the structure and the focus on what the end user wants to achieve.

I just have a comment on providing a way to override defaults for a `selection` -> `rule`. Given we run in a "daemonset" mode and instrument many services, more and more so, we've had the request to make certain settings be configurable per selection criteria.

It would be really nice if we could provide a way to override certain global settings.

devdocs/config/version-2.0/examples/default-configuration.yaml
MrAlias: You’re right: v2 has scoped filtering controls, but it does not yet have explicit per-selection-rule overrides equivalent to legacy exports/sampler/routes/metrics. I'm looking into ways to address this.

MrAlias: I've updated with details on how we can support custom sampling and still use the OTel declarative configuration semantics for the tracing pipeline as the unique definition. Given the sampler there is extensible, we can provide our own sampler with matching and tuning configuration.

I'm still looking at the refinement/override of instrumentation per selection.

devdocs/config/version-2.0/.verify/default-config-current.yaml
```diff
@@ -0,0 +1,371 @@
...
+trace_printer: disabled
+executable_path: ""
+autotargetexe: ""
+open_port: ""
+autotargetlanguage: ""
+target_pids: ""
+service_name: ""
+service_namespace: ""
+metrics:
+    features: 8
```
mariomac, line 267: The global "features" section was deprecated in favor of each entry of "discovery > instrument".

devdocs/config/version-2.0/.verify/default-config-current.yaml
```diff
@@ -0,0 +1,371 @@
...
+discovery:
+    services: []
+    exclude_services: []
+    default_exclude_services:
```
mariomac, line 271: services, exclude_services and default_exclude_services (Regexp-based) were kept for backwards compatibility, but they were deprecated in favor of "instrument, exclude_instrument and default_exclude_instrument (Glob-based).

Maybe we can take the opportunity to remove them.

devdocs/config/version-2.0/.verify/default-config-current.yaml
mariomac: Also the auto-generation seems to have failed because the `export.Features` type is internally an integer, but we actually override the `UnmarshalYAML` method to accept a list of strings.

The real default value should be: `["application"]`

mariomac (commented):
Amazing! I'd like to drop few minor comments.

devdocs/config/version-2.0/examples/default-configuration.yaml
```diff
@@ -0,0 +1,585 @@
...
+              size: 256
+              ttl: 1h0m0s
+        # Debug/diagnostics output controls.
+        diagnostics:
+          print_flows: false
+
+    enrich:
+      # Runtime configuration for metadata enrichers.
+      enrichers:
+        kubernetes:
```
dmitryax, line 481: If this is run as a collector receiver it'll be duplicated cache/load on the k8s api with https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/k8sattributesprocessor which does the same enrichment

devdocs/config/version-2.0/examples/default-configuration.yaml
```diff
@@ -0,0 +1,585 @@
...
+          size: 1024
+          ttl: 5m0s
+        # Rule-based service identity mapping.
+        # Rule order is the precedence model: earlier rules win when they set the same target field.
+        rules:
+          - id: k8s-default
+            from: kubernetes
+            description: Default Kubernetes label mapping into canonical service identity.
+            map:
+              service.name:
```
dmitryax, line 519: k8sattributes does this as well following OTel semantic convensions for sourcing servce.name as prescribed [here](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#how-servicename-should-be-calculated)

devdocs/config/version-2.0/examples/default-configuration.yaml
```diff
@@ -0,0 +1,585 @@
...
+          #     service.version:
+          #       - cloud.service.version
+        # Fallback names when peer/host resolution fails.
+        unresolved_hosts:
+          names:
+            default: unresolved
+            outgoing: outgoing
+            incoming: incoming
+      # Attribute enrichment/selection controls.
+      attributes:
```
dmitryax, line 532: Should be done by collector processors

devdocs/config/version-2.0/examples/default-configuration.yaml
MrAlias: This is now addressed. The updated design adds a `refine` block on include rules with an explicit, closed vocabulary of overridable fields:

- `exports`: override which signals (`traces`, `metrics`) are emitted for this workload
- `http.routes`: override HTTP route patterns and fallback policy for this workload  
- `http.filters`: replace HTTP trace/metric filters for this workload

For sampling overrides, the design keeps sampling in `tracer_provider.sampler` (canonical OTel location) via a new `obi_rule_based` custom sampler plugin that supports per-workload matching. This avoids introducing a parallel pipeline model inside `extensions.obi`.

The decision to use an explicit vocabulary rather than generic deep-merge was deliberate — deep-merge has ambiguous semantics for arrays and absent fields, and the actual override use cases are a small, well-defined set. See the updated `config-v2.md` for the full rationale.

devdocs/config/version-2.0/examples/default-configuration.yaml
MrAlias: Agreed — and the updated design addresses this structurally. `enrich` is now a standalone-mode-only section. When OBI runs as a Collector receiver, `enrich`, `correlation`, and `daemon` are not valid and the parser will reject them. The receiver embeds only `capture`, which contains no enrichment logic.

For Kubernetes environments using OBI as a receiver, the recommendation is to use `k8sattributesprocessor` in the Collector pipeline and omit `enrich` entirely from the receiver config (or set `enrich.enrichers.kubernetes.mode: disabled` if the section is present). The rationale is documented in the updated `config-v2.md` under the `enrich` section.

devdocs/config/version-2.0/examples/default-configuration.yaml
MrAlias: Same answer as the k8s enrichment thread: `enrich` (which contains `service_name`) is now standalone-mode only. In receiver deployments, service name derivation follows OTel semantic conventions via `k8sattributesprocessor` — no duplication with OBI's built-in service name logic.

devdocs/config/version-2.0/examples/default-configuration.yaml
MrAlias: Correct — and the `enrich.attributes` section is now standalone-mode only for exactly this reason. In receiver mode, attribute enrichment belongs in the Collector pipeline via processors.

devdocs/config/version-2.0/examples/default-configuration.yaml
MrAlias: The updated design keeps `profiling`, `internal_metrics`, and `logging` in config under the new `daemon` section, which is explicitly standalone-mode only. When OBI runs as a Collector receiver, these are absent — the Collector manages process lifecycle, its own pprof extension, and internal telemetry. So the receiver embedding only includes `capture`; the Collector operator configures profiling/internal metrics in the appropriate Collector config sections.

devdocs/config/version-2.0/.verify/default-config-current.yaml
MrAlias: Good catch on both points. The v2 mapping for `metrics.features` fans out to per-protocol `capture.instrumentation.<protocol>.enabled.metrics` — so the deprecated global features field maps cleanly into the per-protocol structure. The `verify.go` parity check will need to be updated to use the `["application"]` default value you noted (not the raw integer), and to map against `discovery > instrument` rather than the deprecated top-level `features` field. I'll update the parity checker to reflect this.

devdocs/config/version-2.0/.verify/default-config-current.yaml
MrAlias: Agreed — the v2 design uses only glob-based matching (`exe_path_glob`, namespace/pod globs, etc.) and doesn't carry forward the regexp-based `services`/`exclude_services`/`default_exclude_services` aliases. The v1→v2 migration CLI will map the glob-based variants; the regexp-based deprecated aliases are intentionally dropped. The migration CLI will flag these with an explicit warning if it encounters them in a v1 config being migrated.

devdocs/config/version-2.0/migration.md
```diff
@@ -0,0 +1,119 @@
...
+- Deterministic parsing and validation for v2 inputs.
+- Consistent behavior across standalone host and collector receiver host.
+- Actionable diagnostics for operators before rollout.
+
+## v2 Configuration Parsing
+
+A new configuration package will be added. It's purpose will be to provide:
+
+- Parsing functionality of the `extension.obi` portion of the `v2` configuration
+- Export types representing the OBI conifguration
```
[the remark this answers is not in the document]
MrAlias: Fixed in the doc rewrite.

devdocs/config/version-2.0/migration.md
```diff
@@ -0,0 +1,119 @@
...
+
+Using this new package, both the OBI command and the collector receiver will parse user provided configuration.
+It will be up to these callers to deterine:
```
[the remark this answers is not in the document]
MrAlias: Fixed in the doc rewrite.

devdocs/config/version-2.0/migration.md
```diff
@@ -0,0 +1,119 @@
...
+
+- how to fallback to v1 support when the parser informs it that is then input format
+- how to setup the SDK which is outside the scope of the v2 configuration package
+
+### Integration with `otelconf`
+
+It is assumed that users that need SDK will use the `go.opentelemetry.io/contrib/otelconf` package to parse top-level objects of the declartive config accordingly.
```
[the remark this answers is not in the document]
MrAlias: Fixed in the doc rewrite.

devdocs/config/version-2.0/migration.md
```diff
@@ -0,0 +1,119 @@
...
+It does not merge or translate `instrumentation/development` into OBI-owned settings.
+
+### Backward compatibility behavior
+
+Based on the structure of the configuration, the version of that configuration can be determined from:
+
+- Root `version` identifies OTel declarative document contract.
+- `extensions.obi.version` identifies OBI extension contract.
+
+From this, the v2 configuration pacakge will behave as follows:
```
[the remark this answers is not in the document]
MrAlias: Fixed in the doc rewrite.

devdocs/config/version-2.0/config-v2.md
```diff
@@ -0,0 +1,404 @@
...
+- **Journey-first, user-mental-model first**
+  - Configuration should match what users are trying to do, not internal implementation layering.
+  - Structure should optimize for readability and safe default operation.
+
+- **One concern, one place**
+  - Every concern has one canonical home.
+  - Avoid parallel knobs for the same behavior across sections.
+  - OBI-specific concerns remain under `extensions.obi`, independent of generic instrumentation sections.
+
+- **Compatible with OpenTelemetry declarative configuartion**
```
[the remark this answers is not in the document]
MrAlias: Fixed in the doc rewrite.

devdocs/config/version-2.0/config-v2.md
```diff
@@ -0,0 +1,404 @@
...
+
+The current configuration model has evolved organically with a focus on implementation needs and incremental user feedback.
+This has led to some structural inconsistencies, redundant controls, and a mix of user-facing and internal configuration in the same sections.
+To address this, a user-centric redesign of the configuration schema that optimizes for common user journeys, clear ownership of concerns, and a clean separation between user-facing configuration and internal implementation details is being proposed here.
+
+Goals:
+
+- Define a clear, consistent configuration schema that maps directly to user intent and common use cases.
+- Provide an extension to the OpenTelemetry declarative configuration model that configures OBI-specific behavior.
+- Grantee a smooth migration path from the current v1 configuration shape to the new v2 shape, with clear validation and tooling support.
```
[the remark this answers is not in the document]
MrAlias: Fixed in the doc rewrite.

devdocs/config/version-2.0/migration.md
```diff
@@ -0,0 +1,119 @@
...
+- Lock version-detection and compatibility behavior.
+
+### Phase 2 — Dual-read period
+
+- Attempt v2 parser first; on explicit not-v2 result, invoke legacy parser path.
+
+### Phase 3 — v2-first default
+
+- Default docs/examples/CI to v2.
+- Deprecate the v1 configuration. Warn users, and tell them how migrate with tooling.
```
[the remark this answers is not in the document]
MrAlias: Fixed in the doc rewrite.

MrAlias:
## Design update — `capture`/`daemon` split + per-workload `refine`

The design has been significantly revised based on reviewer feedback. Summary of the main changes:

### Deployment-aware structure (`capture` vs. `daemon`/`enrich`/`correlation`)

The flat `extensions.obi` top-level structure has been reorganized around deployment scope:

- **`capture`** — receiver-embeddable, valid in all deployment modes. Contains everything OBI needs to select workloads and capture telemetry: `policy`, `rules`, `instrumentation`, `runtimes`, `network`, `limits`, `engine`, `safety`, `channels`, `telemetry`.
- **`enrich`**, **`correlation`**, **`daemon`** — standalone-mode only. Not valid in Collector receiver deployments.

This directly addresses @dmitryax's concerns: when OBI runs as a Collector receiver, `enrich` (including k8s metadata, service name, attribute enrichment) is absent. The Collector pipeline handles enrichment via `k8sattributesprocessor` and other processors — no duplication. `log_trace_annotation` (`correlation`) is also standalone-only; a future standalone Collector component is the planned path there.

### Per-workload `refine` block

Addresses the feedback from @grcevski and @fstab on per-service configuration. Include rules now support an optional `refine` block:

```yaml
capture:
  rules:
    - action: include
      match:
        kubernetes:
          namespace_glob: ["staging-*"]
      refine:
        exports:
          traces: false
          metrics: true
    - action: include
      match:
        kubernetes:
          namespace_glob: ["orders"]
      refine:
        http:
          routes:
            unmatched: wildcard
            patterns:
              - /orders/{id}
```

`refine` uses an explicit, closed vocabulary (`exports`, `http.routes`, `http.filters`) rather than generic deep-merge, to avoid ambiguous array/map merge semantics. New fields can be added deliberately as use cases emerge.

Per-workload **sampling** stays in `tracer_provider.sampler` (canonical OTel location) via the planned `obi_rule_based` custom sampler plugin — keeping sampling out of the OBI extension namespace.

### `policy`/`rules` flattened into `capture`

The earlier `capture.selection.policy` / `capture.selection.rules` nesting has been removed. `policy` and `rules` are now direct children of `capture`, saving one indentation level on the most-written field.

### `operations` split and renamed

The old `operations` section mixed capture-valid and daemon-only fields. It's now split:
- eBPF engine internals → `capture.engine`
- Process management (logging, profiling, shutdown, internal metrics) → `daemon`

---

Updated files: `config-v2.md` (design rationale for every structural decision), `migration.md` (deployment-mode validation, non-deterministic migration definition, phased rollout rationale), `obi-extension.schema.json`, `examples/default-configuration.yaml`, `verify.go`.

Marking this ready for re-review. Happy to discuss any of the above.

MrAlias:
@fstab — yes, both of your examples are now supported. The updated design adds a `refine` block on include rules with an explicit vocabulary of per-workload overrides:

- `exports`: override which signals (`traces`, `metrics`) are emitted for a matched workload — so you can disable traces for a specific service while keeping them globally.
- `http.routes`: override HTTP route patterns and fallback policy for a specific service.
- `http.filters`: replace HTTP trace/metric filters for a specific service.

Example covering both your cases:

```yaml
capture:
  rules:
    # Disable traces for shopping cart, keep metrics
    - action: include
      name: shopping-cart
      match:
        kubernetes:
          namespace_glob: ["shopping-cart"]
      refine:
        exports:
          traces: false
          metrics: true

    # Custom HTTP route matchers for orders service
    - action: include
      name: orders-service
      match:
        kubernetes:
          namespace_glob: ["orders"]
      refine:
        http:
          routes:
            unmatched: wildcard
            patterns:
              - /orders/{id}
              - /orders/{id}/items
```

For per-workload sampling overrides, the design keeps sampling in `tracer_provider.sampler` (the canonical OTel location) via a planned `obi_rule_based` custom sampler plugin that supports workload-matching rules. See the updated `config-v2.md` for the full design rationale.

devdocs/config/version-2.0/examples/default-configuration.yaml
MrAlias: Small correction to my previous reply: runtimes ended up as `capture.runtimes`, not a standalone top-level section. The `capture` block groups everything valid in both standalone and Collector receiver deployments — runtimes included. Full paths are `capture.runtimes.go`, `capture.runtimes.nodejs`, `capture.runtimes.java`.

devdocs/config/version-2.0/examples/default-configuration.yaml
```diff
@@ -0,0 +1,613 @@
...
+            discovery:
+              # Global timeout for route template discovery runs.
+              timeout: 10s
+              # Languages to skip for route template discovery.
+              disabled_languages: []
+              java:
+                # Delay before Java route template discovery to allow runtime readiness.
+                delay: 1m0s
+          # HTTP payload-level extraction features.
+          payload_extraction:
```
NimrodAvni78, line 232: I think for all of these (and other protocols we infer on top of http) the current configuration is really complicated, needed a new field and env var per protocol, unlike the normal `instrumentations` field where its just a list of protocols, the only difference is that this detection is done over http, and probably needs http large buffers enabled to work reliably
i think we should just unify this config instrumentations config, maybe warn on startup / document that some of these wont work reliably without large buffers.
and allow for some unifying semantics (for example instead of specifying `anthropic, openai, gemini` we can allow for users to write `genai`, maybe something similar for other stuff like couchbase (KV + n1ql), aws (sqs and s3), elasticsearch (elasticsearch + opensearch), sql (mysql + postgresql + mssql) and so on.
this could maybe be fixed in OBI now or we can wait for config v2

devdocs/config/version-2.0/examples/default-configuration.yaml
NimrodAvni78: @dmitryax @MrAlias the difference with our k8s enrichment and why we can't just let the collector handle it, even if we are running as part of the collector is that we also use k8s metadata to get the service.peer.{name|namespace}. we resolve the outgoing ip (in case of client span, analogous in server case) and get service name from it if it is inside the same k8s cluster
in most deployments the k8sattributes processor only uses metadata from the same node to enrich spans (to not overload k8s api server), we use k8s-cache to have only select consumers selecting metadata from *all k8s nodes* and sharing them will all obi instances
so if we dont use k8s enrichment in obi we wont have the incoming / outgoing info
i think that might be a concern

grcevski: There's also another issue, if I'm not mistaken the enrichment in the k8s collector processor requires that it can find a pod IP. So I think we need to at least encode the k8s.pod.ip in resource attributes always to be 100% sure this works in all cases. So we must have access to k8s metadata at leas to some extent.

grcevski (approved):
LGTM! Great work!

mariomac (approved):
Good to me! Great improvement

devdocs/config/version-2.0/examples/default-configuration.yaml
```diff
@@ -0,0 +1,613 @@
+# Authored v2 configuration example (journey-first, one-concern-one-place).
+# Values are mapped from current defaults in pkg/obi/DefaultConfig.
+
+# OTel declarative document version.
+file_format: '1.0-rc.1'
```
pellared: What is the reason we use https://github.com/open-telemetry/opentelemetry-configuration/releases/tag/v1.0.0-rc.1 and not https://github.com/open-telemetry/opentelemetry-configuration/releases/tag/v1.0.0 or even https://github.com/open-telemetry/opentelemetry-configuration/releases/tag/v1.0.0-rc.3 ?

devdocs/config/version-2.0/config-v2.md
```diff
@@ -0,0 +1,642 @@
...
+At a high level, the target configuration shape is a standard [OpenTelemetry declarative configuration](https://github.com/open-telemetry/opentelemetry-configuration) document with a root `file_format` field and top-level sections for `resource`, `propagator`, `tracer_provider`, and `meter_provider`.
+All OBI-specific configuration lives under `extensions.obi`.
+
+The `extensions.obi` block is divided by deployment scope:
+
+- `capture`: valid in **all** deployment modes. Contains everything OBI needs to select workloads and capture telemetry. When running OBI as a Collector receiver, this block is embedded directly in the receiver configuration — no manual extraction required.
+- `enrich`, `correlation`, `daemon`: **standalone-mode only**. These sections are not valid in Collector receiver deployments. The Collector pipeline handles enrichment (via processors) and process lifecycle (logging, profiling, shutdown) in receiver mode.
+
+```yaml
+file_format: '1.0-rc.1'
```
pellared: Same as in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1351/files#r3199965696

devdocs/config/version-2.0/examples/default-configuration.yaml
MrAlias: Addressed in 1e04a8079. The payload-extraction shape is now reduced to a single enablement surface: `payload_extraction.enabled` is a list of extractor names, and nested blocks are reserved for tuning only (`sqlpp.endpoint_patterns`). That keeps the config aligned with the list-based instrumentation model without introducing parallel knobs. I also documented the large-buffer caveat as part of the design rationale for HTTP-derived parsing.

devdocs/config/version-2.0/examples/default-configuration.yaml
MrAlias: Addressed in b93a5926f. The example now uses `file_format: "1.0"`, and the docs clarify that this field tracks the declarative schema version (`major.minor`), not an upstream release tag such as `v1.0.0` or `v1.0.0-rc.3`.

devdocs/config/version-2.0/config-v2.md
MrAlias: Same fix as the example thread: b93a5926f updates the design doc to use `file_format: "1.0"` and explains that the field follows the declarative schema version, not a release tag.

NimrodAvni78 (approved):
🔥

MrAlias:
I'm merging this given we just made the v0.9.0 release and there seems to be general acceptance of the new config form. Merging does not immediately switch us over to the new config, but it does land this code to start that process (see https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2080 for next steps).

Thank you all. :pray:

Issue #1134, opened 2026-01-20
🏆 Epic: Additional Protocol Support


MrAlias:
Increase OBI's value by supporting widely-used communication protocols (beyond HTTP/SQL), addressing user demand for unified, zero-code observability.

## Acceptance Criteria

- [x] OBI can instrument and collect telemetry for MQTT
- [ ] OBI can instrument and collect telemetry for AMQP
- [ ] OBI can instrument and collect telemetry for NATS
- [ ] OBI can instrument and collect telemetry for Redis pubsub
- [ ] OBI can instrument (context propagation) and collect telemetry for gRPC
- [ ] OBI supports MongoDB version 5 and below
- [ ] OBI supports compression for MongoDB
- [ ] OBI can instrument and collect telemetry for Cloud service SDKs
- [x] OBI supports Couchbase
- [x] OBI support genAI: openAI, gemini, anthropic

PR #1353, opened 2026-02-24
Improve large buffers and demonstrate with OpenAI protocol support

grcevski:
This PR improves the large buffer support by capturing responses in large buffers too and demonstrates this with implementing the first GenAI protocol - OpenAI.

There are couple of things that I had to do to make this happen:

1. We now delay the HTTPS requests just like the HTTP requests. I need to see if we can pass cleanly our SSL test suite. I believe we had resolved all issues with finding the end of TLS requests, but we'll see.
2. We count correctly the request and response sizes.
3. I enabled us to capture larger than 32K buffers by splitting them and shipping more than one.
4. OpenAI reponds with gzip bodies, so I had to add generic parsing for HTTP requests for compressed packets (gzip, brotli, deflate and zstd).

And finally, since GenAI support in OTel SDKs in general is in infancy, we can really spearhead the OTel support and get across language GenAI observabilty with OBI. We can extend what I did with OpenAI to Anthropic and AWS and others and we'll have a pretty complete solutions and more and more GenAI workloads are being developed and used.

This PR adds only traces support. GenAI spec Metrics will follow.

Big chunk of this PR is just tests. I had to create a mock OpenAI server and wrapper client programs to ensure we capture the payloads correctly.

Relates to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1134

### DIFF
--- a/pkg/ebpf/common/http/openai.go
+++ b/pkg/ebpf/common/http/openai.go
@@ -42,12 +42,12 @@ func OpenAISpan(baseSpan *request.Span, req *http.Request, resp *http.Response) 
 
 	var parsedRequest request.OpenAIInput
 	if err := json.Unmarshal(reqB, &parsedRequest); err != nil {
-		return *baseSpan, false
+		slog.Debug("failed to parse OpenAI request", "error", err)
 	}
 
 	var parsedResponse request.OpenAI
 	if err := json.Unmarshal(respB, &parsedResponse); err != nil {
-		return *baseSpan, false
+		slog.Debug("failed to parse OpenAI response", "error", err)
 	}
 
 	parsedResponse.Request = parsedRequest


pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -543,12 +543,12 @@ func (p *Tracer) lookForTimeouts(ctx context.Context, parseCtx *ebpfcommon.EBPFP
 					// but it hasn't been posted yet, likely missed by the logic that looks at finishing requests
 					// where we track the full response. If we haven't updated the EndMonotimeNs in more than some
 					// short interval, we are likely not going to finish this request from eBPF, so let's do it here.
-					if v.EndMonotimeNs != 0 && v.Submitted == 0 && t.After(kernelTime(v.EndMonotimeNs).Add(2*time.Second)) {
+					if v.EndMonotimeNs != 0 && v.Submitted == 0 && t.After(kernelTime(v.EndMonotimeNs).Add(10*time.Second)) {
```
mmat11, line 546: would it work if we parse the content-length and emit the event once the response body matches it so we can remove this userspace deadline?

grcevski: Yes that's a good idea, I think we can follow-up with that work. It won't help with OpenAI, since they use chunked responses, so ContentLength is -1. But it will work in many other cases.

bpf/generictracer/protocol_http.h
```diff
@@ -483,22 +477,46 @@ static __always_inline int http_send_large_buffer(http_info_t *req,
...
 
-    req->has_large_buffers = true;
+    bpf_dbg_printk("sending large buffer, total size=%d, packet_type=%d, direction %d",
+                   bytes_len,
+                   packet_type,
+                   direction);
 
-    bpf_dbg_printk("sending large buffer, size=%d", bytes_len);
+    int b = 0;
+    for (; b <= (max / k_large_buf_payload_max_size); b++) {
```
rafaelroquetto: I think here you may end up missing trailing bytes or oversending when `max % k_large_buf_payload_max_size == 0`

You might want to do:


```suggestion
const u32 niter = (max / k_large_buf_payload_max_size) + ((max % k_large_buf_payload_max_size) > 0);

    for (; b < niter; b++) {
```

bpf/generictracer/protocol_http.h
```diff
@@ -483,22 +477,46 @@ static __always_inline int http_send_large_buffer(http_info_t *req,
...
 
-    large_buf->len = bytes_len;
-    if (large_buf->len >= http_buffer_size) {
-        large_buf->len = http_buffer_size;
-        bpf_dbg_printk("WARN: buffer is full, truncating data");
-    }
+    req->has_large_buffers = true;
 
-    bpf_probe_read(large_buf->buf, large_buf->len & k_large_buf_payload_max_size_mask, u_buf);
+    u32 max = bytes_len;
```
rafaelroquetto: nit:

```suggestion
    u32 available_bytes = bytes_len;
```

bpf/generictracer/protocol_http.h
```diff
@@ -483,22 +477,46 @@ static __always_inline int http_send_large_buffer(http_info_t *req,
...
+    bpf_dbg_printk("sending large buffer, total size=%d, packet_type=%d, direction %d",
+                   bytes_len,
+                   packet_type,
+                   direction);
 
-    bpf_dbg_printk("sending large buffer, size=%d", bytes_len);
+    int b = 0;
+    for (; b <= (max / k_large_buf_payload_max_size); b++) {
+        u32 offset = b * k_large_buf_payload_max_size;
+        if (offset >= k_large_buf_abs_max_size) {
```
rafaelroquetto: I found this a bit confusing - so `k_large_buf_abs_max_size` is the maximum total value we want to read, not the maximum size of a single large buffer, correct?

If so, could we rename it to `k_large_buf_max_read_size` or even better `k_large_buffer_read_limit`?

bpf/generictracer/protocol_http.h
```diff
@@ -483,22 +477,46 @@ static __always_inline int http_send_large_buffer(http_info_t *req,
...
-    req->has_large_buffers = true;
+    bpf_dbg_printk("sending large buffer, total size=%d, packet_type=%d, direction %d",
+                   bytes_len,
+                   packet_type,
+                   direction);
 
-    bpf_dbg_printk("sending large buffer, size=%d", bytes_len);
+    int b = 0;
+    for (; b <= (max / k_large_buf_payload_max_size); b++) {
+        u32 offset = b * k_large_buf_payload_max_size;
```
```suggestion
        const u32 offset = b * k_large_buf_payload_max_size;
```

bpf/generictracer/protocol_http.h
```diff
@@ -483,22 +477,46 @@ static __always_inline int http_send_large_buffer(http_info_t *req,
...
+        if (offset >= k_large_buf_abs_max_size) {
+            break;
+        }
+        u32 read_size = chunk;
+        bpf_clamp_umax(read_size, k_large_buf_payload_max_size);
```
rafaelroquetto: you may not need `chunk` altogether, instead:


```suggestion
        const u32 read_size = available_bytes % k_large_buf_payload_max_size;
```

(`available_bytes` is `max`, renamed as per nit comment above)

bpf/generictracer/protocol_http.h
```diff
@@ -483,22 +477,46 @@ static __always_inline int http_send_large_buffer(http_info_t *req,
...
+
+        large_buf->len = read_size;
+
+        u32 total_size = sizeof(tcp_large_buffer_t);
+        total_size += large_buf->len > sizeof(void *) ? large_buf->len : sizeof(void *);
+
+        bpf_clamp_umax(total_size, k_large_buf_max_size);
+        bpf_ringbuf_output(&events, large_buf, total_size, get_flags());
+
+        if (read_size <= k_large_buf_payload_max_size) {
```
rafaelroquetto: I could be misunderstanding this, but if you apply my suggestions to the for loop, you don't need this

bpf/generictracer/protocol_http.h
```diff
@@ -483,22 +477,46 @@ static __always_inline int http_send_large_buffer(http_info_t *req,
...
+            break;
+        }
+        chunk -= k_large_buf_payload_max_size;
```
```suggestion
        available_bytes -= read_size;
```

you may not have read an entire chunk of k_large_buf_payload_max_size - so if you decrement all of that, you may be skipping bytes

bpf/generictracer/protocol_http.h
```diff
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include "bpfcore/utils.h"
```
rafaelroquetto: vscode vscoding

bpf/generictracer/protocol_http.h
grcevski: Yeah, you know what, I'll actually write a unit test for this and see if I can prove it out.

pkg/ebpf/common/http/openai.go
```diff
@@ -0,0 +1,55 @@
...
+func OpenAISpan(baseSpan *request.Span, req *http.Request, resp *http.Response) (request.Span, bool) {
+	reqB, err := io.ReadAll(req.Body)
+	if err != nil {
+		return *baseSpan, false
+	}
+	req.Body = io.NopCloser(bytes.NewBuffer(reqB))
+
+	// Check any of the well known response headers that OpenAI would use
+	isOpenAI := false
+	for _, header := range []string{"Openai-Version", "Openai-Organization", "Openai-Project", "Openai-Processing-Ms"} {
```
rafaelroquetto, line 19: here I think you can use `resp.Get("...")`

then you don't need to check `ok` and most importantly, it canonicalises the string - see comment on https://cs.opensource.google/go/go/+/refs/tags/go1.26.0:src/net/http/header.go;l=43

pkg/ebpf/common/http/openai.go
```diff
@@ -0,0 +1,55 @@
...
+	"encoding/json"
+	"io"
+	"log/slog"
+	"net/http"
+
+	"go.opentelemetry.io/obi/pkg/appolly/app/request"
+)
+
+func OpenAISpan(baseSpan *request.Span, req *http.Request, resp *http.Response) (request.Span, bool) {
+	reqB, err := io.ReadAll(req.Body)
```
rafaelroquetto: instead of reading the entire body upfront (which can be large and allocate) you could check the headers first, and only do this if it is openAI

bpf/generictracer/protocol_http.h
grcevski: I found a bug, but this was correct. I was testing the read_size for loop exit, not the chunk.

bpf/common/large_buffers.h
```diff
@@ -16,12 +16,15 @@ enum {
...
-    k_large_buf_max_size = 1 << 14, // 16K
+    k_large_buf_max_size = 1 << 15, // 32K
     k_large_buf_max_size_mask = k_large_buf_max_size - 1,
 
     // Maximum size for a large buffer payload.
-    k_large_buf_payload_max_size = 1 << 13, // 8K
+    k_large_buf_payload_max_size = 1 << 14, // 16K
     k_large_buf_payload_max_size_mask = k_large_buf_payload_max_size - 1,
+
+    // Absolute maximum
```
mmat11: clarity nit: can we change the comment to indicate that if the large buffer is bigger than `k_large_buf_max_size`, multiple chunks are sent until `k_large_buf_abs_max_size`?

bpf/generictracer/k_tracer.c
```diff
@@ -1162,10 +1164,17 @@ int obi_handle_buf_with_args(void *ctx) {
     } else { // large request tracking and generic TCP
         http_info_t *info = bpf_map_lookup_elem(&ongoing_http, &args->pid_conn);
 
+        bpf_d_printk("http? info %llx, submitted %d, still reading %d",
```
mmat11:
```suggestion
        bpf_d_printk("http info %llx, submitted %d, still reading %d",
```

bpf/generictracer/protocol_http.h
grcevski: Sure

bpf/generictracer/protocol_http.h
grcevski: Sure

bpf/generictracer/protocol_http.h
grcevski: I had simpler code, but the verifier was complaining

bpf/generictracer/protocol_http.h
grcevski: I need to tell apart if we still have more to read or we read exactly the chunk size. Since I clamp max, we can't tell if it was exact or clamped without a second variable.

bpf/generictracer/protocol_http.h
grcevski: yeah this was the actual bug

pkg/ebpf/common/http/openai.go
grcevski: Good idea.

pkg/appolly/app/request/span.go
```diff
@@ -226,6 +227,94 @@ type AWSSQS struct {
...
+
+	return u.CompletionTokens
+}
+
+type OpenAIError struct {
+	Message string `json:"message"`
+	Type    string `json:"type"`
+}
+
+type OpenAI struct {
```
mmat11, line 259: I wonder how much can we generalize this -- what about having an "LLM" struct or similar name, with all the generic fields and then two (or more) additional fields:

- flavor -> openai, ...
- *openai -> openai specific data

This way we don't have to handle specific vendors in a different way when exposing metrics or handling the span in general

pkg/ebpf/common/http/openai.go
```diff
@@ -0,0 +1,55 @@
...
+
+	respB, err := getResponseBody(resp)
+	if err != nil && len(respB) == 0 {
+		return *baseSpan, false
+	}
+
+	slog.Debug("OpenAI", "request", string(reqB), "response", string(respB))
+
+	var parsedRequest request.OpenAIInput
+	_ = json.Unmarshal(reqB, &parsedRequest)
```
mmat11: I'd handle the errors here and 2 lines below

bpf/generictracer/protocol_http.h
grcevski: I think I have the loop how you thought it should be and I've modified the unit test to reflect it. Thanks for pointing it out, it's much simpler. I need to check if the verifier likes it and I'll push.

pkg/ebpf/common/http/responses.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// http.ReadResponse does NOT auto-decompress Content-Encoding
+	// (only http.Transport does, and only for gzip). Decompress manually.
+	body := respB
+	if enc := resp.Header.Get("Content-Encoding"); enc != "" && len(respB) > 0 {
+		if dec, derr := decompressBody(enc, respB); derr == nil {
+			body = dec
+		} else {
+			return nil, fmt.Errorf("decompress error (enc=%s, truncated body?): %w", enc, derr)
+		}
```
mmat11: clarity nit:

```suggestion
		dec, err := decompressBody(enc, respB)
		if err != nil {
			return nil, fmt.Errorf("decompress error (enc=%s, truncated body?): %w", enc, err)
		}
		body = dec
```

pkg/ebpf/common/http/responses.go
```diff
@@ -0,0 +1,68 @@
...
+			return nil, err
+		}
+		defer zr.Close()
+		return io.ReadAll(zr)
+	case "deflate":
+		fr := flate.NewReader(bytes.NewReader(b))
+		defer fr.Close()
+		return io.ReadAll(fr)
+	case "br":
+		return io.ReadAll(brotli.NewReader(bytes.NewReader(b)))
```
mmat11, line 64: should the brotli reader be closed like all the others?

bpf/common/large_buffers.h
grcevski: Good idea.

grcevski: Actually, I think I should probably make the maximum configurable.

pkg/appolly/app/request/span.go
grcevski: Yeah absolutely. I want to see what it would take to do Anthropic and then I'll combine them, perhaps have separate JSON reader structs and put them into a generic struct. I'd do this after we have a second provider.

pkg/ebpf/common/http/openai.go
grcevski: Good idea.

pkg/ebpf/common/http/responses.go
grcevski: ~~Nice catch, I think I need to.~~

grcevski: Oh I remember, there's no close method on this one.

bpf/common/large_buffers.h
grcevski: I was missing `    bpf_clamp_umax(available_bytes, http_buffer_size); `.

pkg/config/ebpf_tracer.go
```diff
@@ -152,7 +152,7 @@ func (e *EBPFTracer) CudaInstrumentationEnabled() bool {
 // Max: 8192 bytes.
```
mmat11: not super important, but this needs to be updated

bpf/generictracer/protocol_http.h
```diff
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include <bpfcore/utils.h>
```
rafaelroquetto: one last nit - this may be passing because `protocol_http.h` is being indirectly included, but theoretically this needs to go under `vmlinux.h` as it is what defines types such as `u16` and what not.

I'd just move this after line 9

bpf/generictracer/protocol_http.h
```diff
@@ -483,22 +477,49 @@ static __always_inline int http_send_large_buffer(http_info_t *req,
...
 
-    bpf_probe_read(large_buf->buf, large_buf->len & k_large_buf_payload_max_size_mask, u_buf);
+    bpf_dbg_printk("sending large buffer, total size=%d, packet_type=%d, direction %d",
+                   bytes_len,
+                   packet_type,
+                   direction);
 
-    u32 total_size = sizeof(tcp_large_buffer_t);
-    total_size += large_buf->len > sizeof(void *) ? large_buf->len : sizeof(void *);
+    const uint32_t niter = (available_bytes / k_large_buf_payload_max_size) +
```
mmat11, line 495: out of curiosity, did you try to just increase the numbers and send a single event without chunking?

pkg/config/ebpf_tracer.go
grcevski: Done!

bpf/generictracer/protocol_http.h
grcevski: Yes, eBPF would not let me allocate more than 32K for the in memory map array.

mmat11: ahhh yes makes sense

bpf/generictracer/protocol_http.h
```diff
@@ -483,22 +477,49 @@ static __always_inline int http_send_large_buffer(http_info_t *req,
...
-    if (large_buf->len >= http_buffer_size) {
-        large_buf->len = http_buffer_size;
-        bpf_dbg_printk("WARN: buffer is full, truncating data");
+    req->has_large_buffers = true;
+
+    u32 available_bytes = bytes_len;
+    // limit by the userspace requested size
+    if (available_bytes > http_buffer_size) {
+        available_bytes = http_buffer_size;
     }
```
rafaelroquetto, line 486: I might be misunderstanding so please bear with me.

`http_buffer_size` is always meant to be less than `k_large_buf_payload_max_size` (i.e. `k_large_buf_payload_max_size` is a ceiling).

So capping `available_bytes` to `http_buffer_size` means that you will always end up sending a single large buffer (`niter == 1`) and I am assuming the intent here is to slice `available_bytes` into `N` large buffers, so I think this block should be removed - then see below.

bpf/generictracer/protocol_http.h
```diff
@@ -483,22 +477,49 @@ static __always_inline int http_send_large_buffer(http_info_t *req,
...
-    bpf_probe_read(large_buf->buf, large_buf->len & k_large_buf_payload_max_size_mask, u_buf);
+    bpf_dbg_printk("sending large buffer, total size=%d, packet_type=%d, direction %d",
+                   bytes_len,
+                   packet_type,
+                   direction);
 
-    u32 total_size = sizeof(tcp_large_buffer_t);
-    total_size += large_buf->len > sizeof(void *) ? large_buf->len : sizeof(void *);
+    const uint32_t niter = (available_bytes / k_large_buf_payload_max_size) +
+                           ((available_bytes % k_large_buf_payload_max_size) > 0);
```
rafaelroquetto, line 496: this should actually be


```suggestion
    const uint32_t niter = (available_bytes / http_buffer_size) +
                           ((available_bytes % http_buffer_size) > 0);
```

this will yield the effective number of iterations to consume all of the available bytes

bpf/generictracer/protocol_http.h
```diff
@@ -483,22 +477,49 @@ static __always_inline int http_send_large_buffer(http_info_t *req,
...
 
-    req->has_large_buffers = true;
+    int b = 0;
+    for (; b < niter; b++) {
+        const u32 offset = b * k_large_buf_payload_max_size;
```
rafaelroquetto, line 500: and then this becomes
```suggestion
        const u32 offset = b * http_buffer_size;
```

otherwise your stride is potentially larger than `http_buffer_size` and you skip bytes. I think this only worked so far because we are consistently using `k_large_buf_payload_max_size` to read, meaning we are not respecting `http_buffer_size` and always sending the maximum number of bytes

bpf/generictracer/protocol_http.h
```diff
@@ -483,22 +477,49 @@ static __always_inline int http_send_large_buffer(http_info_t *req,
...
+        if (offset >= k_large_buffer_read_limit) {
```
rafaelroquetto, line 501:
```suggestion
        if (offset + read_size >= k_large_buffer_read_limit) {
```

if we can read at most `k_large_buffer_read_limit` in total, we need to account for the bytes already read (i.e. `offset` bytes) + the bytes we are about to read, otherwise we can overflow.

bpf/generictracer/protocol_http.h
```diff
@@ -483,22 +477,49 @@ static __always_inline int http_send_large_buffer(http_info_t *req,
...
+            break;
+        }
+        u32 read_size = available_bytes;
```
rafaelroquetto, line 504:
```suggestion
        u32 read_size = available_bytes % http_buffer_size;
```

we need to cap to `http_buffer_size` here as well, to read at most `http_buffer_size` bytes per buffer

bpf/generictracer/protocol_http.h
```diff
@@ -483,22 +477,49 @@ static __always_inline int http_send_large_buffer(http_info_t *req,
...
+        bpf_clamp_umax(read_size, k_large_buf_payload_max_size);
```
rafaelroquetto, line 505: and this is correct, it should keep the verifier happy about `read_size` which at this point is guaranteed to be less than `k_large_buf_payload_max_size`

bpf/generictracer/protocol_http.h
grcevski: not necessarily, it's set by userspace and while there's a cap on the config setting, I don't want to leave it up to userspace to decide.

User space can set 2K, or 200K. If it's 2K we should only send 2K. If it's 200K, it should send 64K.

bpf/generictracer/protocol_http.h
grcevski: No, it should be capped by what the buffer is possible with eBPF, not the userspace set value.

rafaelroquetto (approved):
Discussed offline, follow up PRs to come.

PR #1354, opened 2026-02-24
Allow to limit process selection by command line argument

pilhuhn:
This allows to limit process selection by command line argument. E.g.:
```
discovery:
  instrument:
    - exe_path: "*python*"
      cmd_args: "server*.py"
```

Especially in standalone case there may be a lot of Python processes running (on my RHEL one there are currently 3 system services), that should not be observed.

### DIFF
--- a/docs/config-schema.json
+++ b/docs/config-schema.json
@@ -551,6 +551,10 @@
     },
     "GlobAttributes": {
       "properties": {
+        "cmd_args": {
+          "$ref": "#/$defs/GlobAttr",
+          "description": "CmdArgs allows to limit by matching command line arguments"
+        },
         "container_name": {
           "$ref": "#/$defs/GlobAttr"
         },
@@ -1439,6 +1443,10 @@
     },
     "RegexSelector": {
       "properties": {
+        "cmd_args": {
+          "$ref": "#/$defs/RegexpAttr",
+          "description": "CmdArgs allows matching by command line arguments"
+        },
         "container_name": {
           "$ref": "#/$defs/RegexpAttr"
         },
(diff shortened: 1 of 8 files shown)


docs/config-schema.json
```diff
@@ -551,6 +551,10 @@
     },
     "GlobAttributes": {
       "properties": {
+        "cmd_args": {
+          "$ref": "#/$defs/GlobAttr",
+          "description": "CmdArgs allows to further limit by command line arguments"
```
MrAlias:
```suggestion
          "description": "cmd_args allows to further limit by command line arguments"
```

Similar below.

docs/config-schema.json
```diff
@@ -551,6 +551,10 @@
     },
     "GlobAttributes": {
       "properties": {
+        "cmd_args": {
+          "$ref": "#/$defs/GlobAttr",
```
MrAlias, line 555: Saying "further" here implies it doesn't work without an exec defined, but it should, right?

pilhuhn: It should (only) work with an exec defined.  
Are you suggesting, that it should not be in addition, but a "standalone" filter?

docs/config-schema.json
pilhuhn: I don't think this suggestion is valid, as my code follows the pattern of e.g. OpenPorts, See https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/b8a6436179eb7e14170f4ce2e164d9a87349f557/pkg/appolly/services/attr_glob.go#L86-88
Also the config-schema.json is generated, so the change should not happen here anyway.

docs/config-schema.json
grcevski: I think we probably don't want to make it used only when exec is defined. I'd leave it as standalone selector option, the users can always add exe_path if needed.

pilhuhn: Ok, I'll change it

mmat11 (approved):

PR #1364, opened 2026-02-25
ci: tentative fix for the release job

mmat11:
AI explanation:

```
The Problem
The top-level permissions: {} sets all permissions to none for the entire workflow. In GitHub Actions, job-level permissions can only restrict, not expand, beyond the workflow-level ceiling. So even though individual jobs declare permissions like contents: read or checks: write, those declarations are blocked by the {} ceiling above them.

This is the line causing the issue:
# https://docs.zizmor.sh/audits/#excessive-permissions
permissions: {}

The Fix
Remove the top-level permissions: {}. Since every job in this workflow already has its own explicit permissions: block, the per-job declarations alone are sufficient to enforce least-privilege — which is exactly what the zizmor audit recommends. The top-level block is redundant here and actively harmful.
REMOVE this block entirely:
# permissions: {}

Each job already defines only what it needs, for example:
validate-tag:
  permissions:
    contents: read   # ✅ this is now respected

release:
  permissions:
    contents: write  # ✅ this is now respected

Why This Is Still Secure
The zizmor excessive-permissions audit is satisfied as long as each job declares its own minimal permissions — the top-level block isn't required for that. Its only unique effect here is preventing the per-job grants from working, which is the opposite of the intent. Removing it while keeping all job-level permissions: blocks intact gives you the same security posture without the breakage.
```

Fixes: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/22405505860

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,9 +16,6 @@ on:
         required: true
         type: string
 
-# https://docs.zizmor.sh/audits/#excessive-permissions
-permissions: {}
-
 jobs:
   # Validate tag format before proceeding with release.
   validate-tag:


.github/workflows/release.yml
```diff
@@ -16,9 +16,6 @@ on:
         required: true
         type: string
 
-# https://docs.zizmor.sh/audits/#excessive-permissions
-permissions: {}
```
skl, line 20: It's usually recommended to do this as standard practice, as per the linked doc:

> In practice, this means that workflows should almost always set permissions: {} at the workflow level to disable all permissions by default, and then set specific job-level permissions as needed.

Generally the permissions needed should be set within each job as required.

skl: Although I do see the permissions at the job level, so maybe this general advice does not apply when nested workflows are used?

skl: Ah no, the jobs [at that time](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d12e1612cf8060b28e16862585c524c4cbe41b1d/.github/workflows/release.yml#L92-L97) were missing job-level permissions which [they now have](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/4793b9bd5e3d377db4a7976f9654e4594c85aad7/.github/workflows/release.yml#L96-L99).

So this has already been fixed and this change is not required.

mmat11: the original error message suggests it somehow cannot use the permissions specified in the nested job, not sure if the msg is misleading or zizmor is wrong (?)

skl: If you look at the most [recent run in main](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/21923175763) of the release action, rather than 0.5.0, it does not have this permissions issue.

The problem is that the 0.5.0 tag points to an outdated workflow, because the workflow did not have the right permissions at the time the tag was created.

As this problem is already resolved in main, there's nothing to do here.

MrAlias: Yes, this was addressed already in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1284

MrAlias: [This run](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/22405505860) was run against the tagged commit of `v0.5.0` which doesn't include that fix ^ FWIW.

mmat11: I see, closing then

PR #1365, opened 2026-02-25
Fix release image tag

MrAlias:
- Update the docker metadata step to find manually set tags.
- Add support to additionally tag the released image as latest.
- Detect if the pushed tag that triggered the release is the latest and tag the docker images accordingly.

## Related Documentation

- [docker/metadata-action inputs](https://github.com/docker/metadata-action?tab=readme-ov-file#inputs)


NimrodAvni78 (approved):
LGTM

PR #1370, opened 2026-02-26
gotracer: set trace context in shared map for http and other protocols

mmat11:
- set trace context in shared map for http and other protocols
- refactor go probes with gotracer/maps/ and gotracer/types/
- remove unused hpack.h

### DIFF
--- a/bpf/gotracer/go_common.h
+++ b/bpf/gotracer/go_common.h
@@ -35,9 +35,7 @@ char __license[] SEC("license") = "Dual MIT/GPL";
 
 enum { W3C_KEY_LENGTH = 11, W3C_VAL_LENGTH = 55 };
 
-#define TP_ENCODED_LEN 8
-
-static unsigned char tp_encoded[TP_ENCODED_LEN] = {
+static unsigned char tp_encoded[] = {
     0x4d, 0x83, 0x21, 0x6b, 0x1d, 0x85, 0xa9, 0x3f}; // hpack encoded "traceparent"
 
 // Temporary information about a function invocation. It stores the invocation time of a function
--- a/bpf/gotracer/types/kafka.h
+++ b/bpf/gotracer/types/kafka.h
@@ -8,9 +8,11 @@
 #include <common/common.h>
 #include <common/tp_info.h>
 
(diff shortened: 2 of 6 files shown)


bpf/gotracer/types/kafka.h
```diff
@@ -0,0 +1,24 @@
...
+#pragma once
+
+#include <bpfcore/vmlinux.h>
+
+#include <common/common.h>
+#include <common/tp_info.h>
+
+#define KAFKA_API_FETCH 0
+#define KAFKA_API_PRODUCE 1
+#define KAFKA_API_KEY_POS 5
```
rafaelroquetto: nit: can we use enum here?

bpf/gotracer/go_common.h
```diff
@@ -27,12 +27,19 @@
...
 #include <logger/bpf_dbg.h>
 
 char __license[] SEC("license") = "Dual MIT/GPL";
 
 enum { W3C_KEY_LENGTH = 11, W3C_VAL_LENGTH = 55 };
 
+#define TP_ENCODED_LEN 8
+
+static unsigned char tp_encoded[TP_ENCODED_LEN] = {
+    0x4d, 0x83, 0x21, 0x6b, 0x1d, 0x85, 0xa9, 0x3f}; // hpack encoded "traceparent"
```
rafaelroquetto:
```suggestion
static unsigned char tp_encoded[] = {
    0x4d, 0x83, 0x21, 0x6b, 0x1d, 0x85, 0xa9, 0x3f}; // hpack encoded "traceparent"
```

if you ever need an actual size constant, you can then do:
```
const u32 TP_ENCODED_LEN = sizeof(tp_encoded);
```

the verifier should treat this as a compile time constant

rafaelroquetto (commented):
Man, these maps are complicated... but it LGTM, apart from the usage of macros - perhaps this is a good opportunity for us to migrate away from them in the code we are touching anyway, what do you think?

(I'm mostly referring to the `#define <CONSTANT>` statements that could be turned into `enum { k_constant = X };`

grcevski:
Amazing, thanks for doing the run @mmat11 !

PR #1373, opened 2026-02-26
Allow loading multiple specs per tracer

rafaelroquetto:
This effectively allows a tracer to be composed of multiple independent translation units - in other words, instead of doing:

```
// mytracer.c

#include "file1.c"
#include "file2.c"
```

A tracer can now `go:generate` individual `.c` files independently, each yielding its own `.o` file, treated as independent translation units. Map sharing continue to work as usual due to `OBI_PIN_INTERNAL`.

### But why?

Glad you asked! Having different translation units allow us to have multiple/different definitions for a single function. For example, sock msg programs operate on a `struct sk_msg` context, whilst sock skb programs operate on `struct __skb_buf` contexts. These have similar semantics, e.g.:
 - you can pull data
 - they both provide direct packet access (`data` and `data_end` pointers)
 - likewise, they both have the concept of length/size

Being able to have multiple implementations allow writing common code. For instance:

```
//common_defs.h
// defines a common """interface"""
static __always_inline u32 ctx_len(void *ctx);
static __always_inline void ctx_pull_data(void *ctx, u32 len);
static __always_inline void *ctx_data(void *ctx);
static __always_inline void *ctx_data_end(void *ctx);

// sockmsg_defs.h
// sk_msg implementations

static __always_inline u32 ctx_len(void *ctx) {
    return ((struct sk_msg_md *)ctx)->size;
}

static __always_inline void ctx_pull_data(void *ctx, u32 len) {
    bpf_msg_pull_data(ctx, 0, len, 0);
}

static __always_inline void *ctx_data(void *ctx) {
    return ((struct sk_msg_md *)ctx)->data;
}

static __always_inline void *ctx_data_end(void *ctx) {
    return ((struct sk_msg_md *)ctx)->data_end;
}

// skb_defs.h
// skb implementations
static __always_inline u32 ctx_len(void *ctx) {
    return ((struct __sk_buff *)ctx)->len;
}

static __always_inline void ctx_pull_data(void *ctx, u32 len) {
    bpf_skb_pull_data(ctx, len);
}

static __always_inline void *ctx_data(void *ctx) {
    return ctx_skb_data((struct __sk_buff *)ctx);
}

static __always_inline void *ctx_data_end(void *ctx) {
    return ctx_skb_data_end((struct __sk_buff *)ctx);
}

// common logic code (context agnostic)
static __always_inline void init_http_request(void *ctx, struct socket_data *sk_data) {
    const u32 len = ctx_len(ctx);

    /* ... */
    const u32 nbytes = len > sizeof(info->buf) ? sizeof(info->buf) : len;

    ctx_pull_data(ctx, nbytes);

    const unsigned char *ptr = ctx_data(ctx);
    const unsigned char *e = ctx_data_end(ctx);

    /* ... */
}
```

The use case is exactly the one illustrated above: allowing programs featuring direct packet access to be generalised (WIP/different PR).

This PR changes `tpinjector/sock_iter.c` to be a separate translation unit as an example.

## Checklist

### DIFF
--- a/pkg/ebpf/tracer_linux.go
+++ b/pkg/ebpf/tracer_linux.go
@@ -211,7 +211,7 @@ func (pt *ProcessTracer) makeOtelBPFFSPath() (string, error) {
 	return otelPath, nil
 }
 
-func (pt *ProcessTracer) setupOtelBPFFFSPath(bundles []*common.SpecBundle) string {
+func (pt *ProcessTracer) setupOtelBPFFSPath(bundles []*common.SpecBundle) string {
 	// Set up BPF FS path once for all specs
 	otelBPFFSPath, err := pt.makeOtelBPFFSPath()
 
@@ -245,7 +245,7 @@ func (pt *ProcessTracer) loadAndAssign(eventContext *common.EBPFEventContext, p 
 		return fmt.Errorf("loading eBPF program specs: %w", err)
 	}
 
-	otelBPFFSPath := pt.setupOtelBPFFFSPath(bundles)
+	otelBPFFSPath := pt.setupOtelBPFFSPath(bundles)
 
 	for i, bundle := range bundles {
 		if err := loadSpec(eventContext, bundle, otelBPFFSPath, i); err != nil {
--- a/pkg/internal/ebpf/tpinjector/tpinjector_test.go
+++ b/pkg/internal/ebpf/tpinjector/tpinjector_test.go
@@ -94,7 +94,7 @@ func TestTracer_Constants(t *testing.T) {
 
 			bundles, err := New(cfg).LoadSpecs()
 			require.NoError(t, err)
-			require.Len(t, bundles, expectedSpecCount, "tpinjector should have exactly one spec")
+			require.Len(t, bundles, expectedSpecCount, "tpinjector bundle count must match")
 
 			// Spec 0 (tpinjector) carries the main constants.
 			c := bundles[0].Constants


pkg/ebpf/tracer_linux.go
```diff
@@ -99,6 +99,34 @@ func resolveMaps(eventContext *common.EBPFEventContext, spec *ebpf.CollectionSpe
 	return &collOpts, nil
 }
 
+func loadSpec(eventContext *common.EBPFEventContext, bundle *common.SpecBundle, otelBPFFSPath string, idx int) error {
+	if err := ebpfconvenience.RewriteConstants(bundle.Spec, bundle.Constants); err != nil {
+		return fmt.Errorf("rewriting BPF constants for spec %d: %w", idx, err)
```
mmat11, line 104: is there a better identifier to use instead of the index? as a developer I think it would be hard to pinpoint where the issue is if we read "error at index 2"

pkg/internal/ebpf/gotracer/gotracer.go
```diff
@@ -84,42 +84,46 @@ func (p *Tracer) supportsContextPropagation() bool {
...
-		"attr_type_int64slice":      uint64(attribute.INT64SLICE),
-		"attr_type_float64slice":    uint64(attribute.FLOAT64SLICE),
-		"attr_type_stringslice":     uint64(attribute.STRINGSLICE),
-		"g_bpf_traceparent_enabled": true,
-		"g_bpf_loop_enabled":        p.supportsBPFLoop,
-	}
+	return []*ebpfcommon.SpecBundle{{
+		Spec:    spec,
+		Objects: &p.bpfObjects,
+		Constants: map[string]any{
```
mmat11: nit: can we keep the `func constants() map[string]any {}` function in every tracer for clarity?

pkg/ebpf/tracer_linux.go
rafaelroquetto: we could add a "name" to the spec bundle - my humble opinion: it'd keep it as is, as I don't think the trade-off of naming every spec in every tracer is worth it - but I don't feel strong about it, so if you think that's best say the words and I will make the change :)

pkg/internal/ebpf/gotracer/gotracer.go
rafaelroquetto: yes, good idea! will do!

pkg/ebpf/tracer_linux.go
mmat11: I realized that indexes are in the order of 0/1/2/3 so it should be fine, let's keep it simple

pkg/ebpf/tracer_linux.go
```diff
@@ -195,43 +211,50 @@ func (pt *ProcessTracer) makeOtelBPFFSPath() (string, error) {
 	return otelPath, nil
 }
 
-func (pt *ProcessTracer) setupBPFFS(spec *ebpf.CollectionSpec) string {
+func (pt *ProcessTracer) setupOtelBPFFFSPath(bundles []*common.SpecBundle) string {
```
grcevski: this looks like it has an extra `F`.

pkg/internal/ebpf/tpinjector/tpinjector_test.go
```diff
@@ -109,13 +89,35 @@ func TestTracer_Constants_FilterPids(t *testing.T) {
 					MaxTransactionTime: 10 * time.Second,
 				},
 			}
+			err := cfg.EBPF.ContextPropagation.UnmarshalText([]byte(tt.contextPropagation))
+			require.NoError(t, err)
+
+			bundles, err := New(cfg).LoadSpecs()
+			require.NoError(t, err)
+			require.Len(t, bundles, expectedSpecCount, "tpinjector should have exactly one spec")
```
grcevski: I think the test comment may be outdated, should it say "tpinjector bundle count must match"?

pkg/ebpf/tracer_linux.go
```diff
@@ -99,6 +99,34 @@ func resolveMaps(eventContext *common.EBPFEventContext, spec *ebpf.CollectionSpe
 	return &collOpts, nil
 }
 
+func loadSpec(eventContext *common.EBPFEventContext, bundle *common.SpecBundle, otelBPFFSPath string, idx int) error {
```
grcevski, line 102: Can bundle ever be nil here? It would cause a panic, I'm wondering if we can return an error message instead...

pkg/ebpf/tracer_linux.go
rafaelroquetto: Ah FFS! I will fix it :)

pkg/internal/ebpf/tpinjector/tpinjector_test.go
rafaelroquetto: good catch, will fix

pkg/ebpf/tracer_linux.go
rafaelroquetto: I don't think so. Every LoadSpecs implementation either returns `nil, err` on failure (in which case the caller returns early before the loop runs), no implementation ever appends a nil pointer element (that would be a major WTF), so this should be safe.

PR #1374, opened 2026-02-26
fix: set telemetry.sdk.name to underlying sdk and add telemetry.distro.name

### LINKED ISSUE #1375: Correct telemetry.sdk.* and telemetry.distro.* Resource Attributes
Our current telemetry attributes don't follow the [OpenTelemetry Semantic Conventions for Telemetry](https://opentelemetry.io/docs/specs/semconv/registry/attributes/telemetry/). We are currently misusing `telemetry.sdk.name` to identify the agent rather than the underlying library.

**The Fix**

Update the OBI Resource attributes to align with the spec:

- `telemetry.sdk.name`: Change to `"opentelemetry-go"`.
- `telemetry.distro.name`: Set to `"opentelemetry-ebpf-instrumentation"`.

**Why?**

Per the spec:

- `telemetry.sdk.name` is the name of the telemetry SDK (which is opentelemetry-go in our case).
- `telemetry.distro.name` is the name of the auto instrumentation agent or distribution".

Related Slack Conversation: https://cloud-native.slack.com/archives/C08P9L4FPKJ/p1772129318780629

svencowart:
- change VendorSDKName to "opentelemetry-go" to correctly identify the underlying SDK
- introduce TelemetryDistroName ("opentelemetry-ebpf-instrumentation") for the distro attribute
- emit telemetry.distro.name in otel metrics, internal metrics, svc graph, and prom reporters
- update all tests to expect the new attribute split

Fixes #1375

### DIFF
--- a/pkg/appolly/instrumenter_test.go
+++ b/pkg/appolly/instrumenter_test.go
@@ -108,6 +108,8 @@ func TestBasicPipeline(t *testing.T) {
 		delete(event.ResourceAttributes, string(semconv.ServiceInstanceIDKey))
 		assert.NotEmpty(t, event.ResourceAttributes, string(semconv.TelemetrySDKVersionKey))
 		delete(event.ResourceAttributes, string(semconv.TelemetrySDKVersionKey))
+		assert.NotEmpty(t, event.ResourceAttributes, string(semconv.TelemetryDistroVersionKey))
+		delete(event.ResourceAttributes, string(semconv.TelemetryDistroVersionKey))
 		if event.Name == "http.server.request.duration" {
 			break
 		}
@@ -136,7 +138,7 @@ func TestBasicPipeline(t *testing.T) {
 			string(semconv.ServiceNameKey):          "foo-svc",
 			string(semconv.ServiceNamespaceKey):     "ns",
 			string(semconv.TelemetrySDKLanguageKey): "go",
-			string(semconv.TelemetrySDKNameKey):     "opentelemetry-go",
+			string(semconv.TelemetrySDKNameKey):     "opentelemetry",
 			string(semconv.TelemetryDistroNameKey):  "opentelemetry-ebpf-instrumentation",
 			string(semconv.OSTypeKey):               "linux",
 		},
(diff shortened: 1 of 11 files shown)


grcevski:
I left a comment on the issue, I think we need to agree if this is the right way to tag the SDK. I see various vendors use these fields like OBI does. telemetry.sdk.name and telemetry.sdk.version explain the product that generated the telemetry and what was the version. I think that should stay OBI and the OBI version. We can add distro name.

grcevski:
Based on the discussion in the issue, I think you are right, we need to change this, but I'd make it 100% conform with the spec:
```
telemetry.distro.name: "opentelemetry-ebpf-instrumentation"
telemetry.sdk.name: "opentelemetry"
```

pkg/appolly/instrumenter_test.go
```diff
@@ -328,6 +331,8 @@ func TestRouteConsolidation(t *testing.T) {
 		delete(event.ResourceAttributes, string(semconv.ServiceInstanceIDKey))
 		assert.NotEmpty(t, event.ResourceAttributes, string(semconv.TelemetrySDKVersionKey))
 		delete(event.ResourceAttributes, string(semconv.TelemetrySDKVersionKey))
+		assert.NotEmpty(t, event.ResourceAttributes, string(semconv.TelemetryDistroVersionKey))
+		delete(event.ResourceAttributes, string(semconv.TelemetryDistroVersionKey))
```
MrAlias, line 335: Similar needs to be done on line 610.

svencowart: I believe I've addressed what you are looking for with the change here: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1374/changes#diff-bc139b8bef44df8910a8f62a0c799899a6754fe838353f2a667498a54edef4eaR892-R893

I'll wait for you to resolve this comment. Thank you.

MrAlias: It looks like the test are still failing: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/22683993252/job/65768935154?pr=1374

svencowart: I believe it's fixed now.

Sorry, I'm having problems running `make test` on my local mac machine. However, I was able to run the specific tests that were failing in the CI pipeline, so I have verification on those.

mmat11 (approved):

PR #1376, opened 2026-02-26
Embed the Java agent in the OBI binary

MrAlias:
This PR embeds the Java agent JAR into the OBI binary so Java instrumentation works in single-binary distributions without shipping a separate JAR. At runtime, OBI extracts the embedded JAR into a user cache location and reuses a checksum-addressed artifact across runs. The extraction path uses temp-file staging plus atomic rename for crash safety and concurrent startup safety.

### Runtime behavior

- Switched Java injector startup to resolve the embedded agent from cache instead of external path lookup.
- Removed Java agent path override surfaces (CLI/env/config runtime path wiring).

### Build and packaging

- Simplified Java build flow to copy directly into the embed directory (no redundant sync target).
- Updated Docker build flow to ensure Java artifact is embedded before Go compile.
- Kept release artifacts aligned with the single-binary model.

### Notices and compliance

- Added Java dependency notice generation into notices workflows.
- Made java-notices-update CI-safe by setting writable Gradle home paths in containerized runs.
- Moved Gradle image pinning to the renovate-managed dependency image pattern.
- Normalized the generated "report was generated at …" footer to keep notice outputs deterministic and avoid check-clean-work-tree failures.

## Validation

- Go unit tests pass.
- Dry-run verified updated make targets and image resolution.
- Re-ran Java notices generation and confirmed stable output hash across runs after normalization.
- Verified generated notice footer remains deterministic between runs.

### DIFF
--- a/pkg/internal/java/java_inject.go
+++ b/pkg/internal/java/java_inject.go
@@ -235,22 +235,27 @@ func ensureEmbeddedAgentInCache() (string, error) {
 	}()
 
 	written, err := tmpFile.Write(embeddedJavaAgentBytes)
+	func() {
+		defer func() {
+			e := tmpFile.Close()
+			if e != nil {
+				err = errors.Join(err, e)
+			}
+		}()
+		if err != nil {
+			return
+		}
+		if written != len(embeddedJavaAgentBytes) {
+			err = errors.New("short write")
+			return
+		}
+		if err = tmpFile.Chmod(0o644); err != nil {
+			err = fmt.Errorf("unable to set permissions on temporary java agent file: %w", err)
+			return
+		}
+	}()
 	if err != nil {
-		_ = tmpFile.Close()
 		return "", fmt.Errorf("unable to write embedded java agent: %w", err)
-	}
-	if written != len(embeddedJavaAgentBytes) {
-		_ = tmpFile.Close()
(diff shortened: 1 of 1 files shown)


pkg/internal/java/java_inject.go
```diff
@@ -178,32 +188,78 @@ func (i *JavaInjector) NewExecutable(ie *ebpf.Instrumentable) error {
...
+	defer func() {
+		if cleanup {
+			_ = os.Remove(tmpPath)
+		}
+	}()
+
+	written, err := tmpFile.Write(embeddedJavaAgentBytes)
 	if err != nil {
-		return "", err
+		_ = tmpFile.Close()
```
rafaelroquetto: Couldn't this also be on a `defer` statement?

rafaelroquetto (approved):
Minor comment, LGTM! Thank you so much for this, good stuff!

pkg/internal/java/java_inject.go
MrAlias: I'll need to refactor a bit. Just `defer`ing here would mean that we will rename the file then close the file handle to the old file. But I can restructure to manage the temp file in its own scope.

MrAlias: Updated.

PR #1379, opened 2026-02-27
Add integration test for sock_iter pre-existing socket tracking

rafaelroquetto:
Validates that the `obi_sk_iter_tcp` correctly registers TCP sockets that exist before OBI starts, so that obi_packet_extender (sk_msg) can inject Traceparent headers into requests sent over those pre-existing connections.

The test uses an echo endpoint to prove wire-level header injection: the server reflects back any Traceparent it receives, which cannot be faked by OBI's internal black box context propagation correlation.

## Checklist

### DIFF
--- a/internal/test/integration/keepalive_test.go
+++ b/internal/test/integration/keepalive_test.go
@@ -11,7 +11,6 @@ import (
 	"testing"
 	"time"
 
-	json "github.com/goccy/go-json"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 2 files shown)


internal/test/integration/keepalive_test.go
```diff
@@ -0,0 +1,40 @@
...
+	require.EventuallyWithT(t, func(ct *assert.CollectT) {
+		resp, err := http.Get("http://localhost:9091/status")
+		require.NoError(ct, err)
+		defer resp.Body.Close()
+		var status struct {
+			TraceparentSeen bool `json:"traceparent_seen"`
+		}
+		require.NoError(ct, json.NewDecoder(resp.Body).Decode(&status))
+		require.True(ct, status.TraceparentSeen)
+	}, 2*time.Minute, time.Second)
```
mmat11:
```suggestion
	}, testTimeout, time.Second)
```

nit: can we use testTimeout? (does it take more than a minute after the healthcheck pass?)

rafaelroquetto: yes, thanks for spotting that - I also got rid of the JSON parsing, it was overengineered

PR #1388, opened 2026-03-02
Dynamic add/remove of PID selection

damemi:
Follow up to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1321

This adds the ability to add and remove PID selections dynamically for callers that import the Instrumenter. It's used by providing an option to `instrumenter.Run()` called `WithDynamicPIDSelector()`. The caller can initialize a `discover.NewDynamicPIDSelector()` which they can hold reference to and add/update pids dynamically.

This is meant as an incremental step toward the ultimate goal of modularizing the Instrumenter and Tracers. In the future, it would be nice to support returning an instance of the Instrumenter directly. But in the interest of a smaller PR, I wanted to wire this through as much existing code as possible.

The intended usage is documented in a new test, but it is essentially:

```go
package main

import (
	"context"
	"go.opentelemetry.io/obi/pkg/appolly/discover"
	"go.opentelemetry.io/obi/pkg/instrumenter"
	"go.opentelemetry.io/obi/pkg/obi"
)

func main() {
	ctx := context.Background()
	cfg := &obi.Config{ /* ... */ }

	// Create the selector and pass it into Run. No callback and no reference to the instrumenter.
	sel := discover.NewDynamicPIDSelector()
	_ = instrumenter.Run(ctx, cfg, instrumenter.WithDynamicPIDSelector(sel))

	// Add/remove PIDs at runtime by calling the selector directly.
	sel.AddPIDs(42, 100)
	pids, _ := sel.GetPIDs()
	sel.RemovePIDs(42)
}
```

t;dr

* New `DynamicPIDSelector` type providing `AddPIDs()`, `RemovePIDs()`, and `GetPIDs()` functions
* `DynamicPIDSelector` implemented as a `Criteria` for Matcher when process created events are received
* `RemovedNotify()` channel handles synthetic delete events from `DynamicPIDSelector.RemovePIDs()` which triggers the existing `m.filterDeleted()` flow

## Checklist

### DIFF
--- a/pkg/appolly/discover/watcher_proc.go
+++ b/pkg/appolly/discover/watcher_proc.go
@@ -127,6 +127,8 @@ type pollAccounter struct {
 	ebpfContext       *ebpfcommon.EBPFEventContext
 	// when non-nil, PIDs received here are removed from pids/pidPorts so they are re-emitted as new on next poll
 	addedPIDsNotify <-chan []app.PID
+	// pidsMu protects pids and pidPorts so the addedPIDsNotify goroutine can call forgetPIDs while snapshot runs
+	pidsMu sync.Mutex
 }
 
 func (pa *pollAccounter) run(ctx context.Context) {
@@ -150,6 +152,10 @@ func (pa *pollAccounter) run(ctx context.Context) {
 
 	go pa.watchForProcessEvents(ctx, log, bpfWatchEvents)
 
+	if pa.addedPIDsNotify != nil {
+		go pa.runAddedPIDsNotify(ctx, log)
+	}
+
 	for {
 		procs, err := pa.listProcesses(pa.portFetchRequired())
 		if err != nil {
@@ -160,26 +166,29 @@ func (pa *pollAccounter) run(ctx context.Context) {
 				pa.output.Send(events)
 			}
 		}
-		switch {
-		case pa.addedPIDsNotify != nil:
-			select {
-			case <-ctx.Done():
-				log.Debug("context canceled. Exiting")
+		select {
+		case <-ctx.Done():
+			log.Debug("context canceled. Exiting")
+			return
+		case <-time.After(pa.interval):
+			// poll again
+		}
+	}
+}
+
+// runAddedPIDsNotify runs in a goroutine; it receives PIDs added to the dynamic selector
+// and calls forgetPIDs so they are re-emitted as new on the next poll.
+func (pa *pollAccounter) runAddedPIDsNotify(ctx context.Context, log *slog.Logger) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case pids, ok := <-pa.addedPIDsNotify:
+			if !ok {
 				return
-			case pids := <-pa.addedPIDsNotify:
-				pa.forgetPIDs(pids)
-				log.Debug("forgot PIDs so they can be re-emitted as new", "pids", pids)
-			case <-time.After(pa.interval):
-				// poll again
 			}
-		default:
-			select {
-			case <-ctx.Done():
-				log.Debug("context canceled. Exiting")
-				return
-			case <-time.After(pa.interval):
-				// poll again
-			}
+			pa.forgetPIDs(pids)
+			log.Debug("forgot PIDs so they can be re-emitted as new", "pids", pids)
 		}
 	}
 }
@@ -187,6 +196,8 @@ func (pa *pollAccounter) run(ctx context.Context) {
 // forgetPIDs removes the given PIDs from the watcher's tracked state so they will be
 // reported as new on the next poll (e.g. when added to the dynamic PID selector).
 func (pa *pollAccounter) forgetPIDs(pids []app.PID) {
+	pa.pidsMu.Lock()
+	defer pa.pidsMu.Unlock()
 	for _, pid := range pids {
 		delete(pa.pids, pid)
 	}
@@ -264,6 +275,8 @@ func (pa *pollAccounter) processTooNew(proc ProcessAttrs) bool {
 // snapshot compares the current processes with the status of the previous poll
 // and forwards a list of process creation/deletion events
 func (pa *pollAccounter) snapshot(fetchedProcs map[app.PID]ProcessAttrs) []Event[ProcessAttrs] {
+	pa.pidsMu.Lock()
+	defer pa.pidsMu.Unlock()
 	log := wplog()
 	var events []Event[ProcessAttrs]
 	currentPidPorts := make(map[pidPort]ProcessAttrs, len(fetchedProcs))


grcevski (commented):
This is pretty cool! I was wondering what you think about making the whole selection criteria reloadable? Was that difficult to achieve? Because now we have the pids in the selection criteria, I was thinking maybe it's doable to just pass new criteria to the matcher...

How does the dynamic PID communicate to the attacher that it should remove the process? Right now the only mechanism is from the watcher to send the process terminated, I think we need to send an event like that on remove...

pkg/appolly/discover/matcher.go
```diff
@@ -81,6 +90,10 @@ func (pm ProcessMatch) LogEnricherEnabled() bool {
 func (m *Matcher) Run(ctx context.Context) {
 	defer m.Output.Close()
 	m.Log.Debug("starting criteria matcher node")
+	if m.RemovedPIDsNotify != nil {
+		m.runWithNotify(ctx)
+		return
+	}
```
mariomac: For simplification, I think you can just replace this addition + the `swarms.ForEachInput` function below just by the contents of `m.runWithNotify`. If `m.RemovedPID` is nil, then the `case removedPIDs := <-m.RemovedPIDsNotify:` will never trigger but that's fine.

pkg/appolly/services/criteria.go
```diff
@@ -171,6 +171,10 @@ type Selector interface {
 	GetLanguages() StringMatcher
 	// GetPIDs returns the list of target PIDs and true when this selector has PID criteria (analogous to OpenPorts).
 	GetPIDs() ([]app.PID, bool)
+	// AddPIDs adds PIDs to the selector's list; no-op for selectors that do not support dynamic PIDs.
+	AddPIDs(pids ...uint32)
+	// RemovePIDs removes PIDs from the selector's list; no-op for selectors that do not support dynamic PIDs.
+	RemovePIDs(pids ...uint32)
```
mariomac: Selector should be a read-only only interface for accepting user-provided configuration and now we are turning it mutable, and overloading it with extra methods that have to be implemented with no-ops in most implementations. Also we have to consider that OBI can manage multiple `GlobAttributes` implementations (one for each user-defined criteria set).

What I would suggest:
1. Remove AddPIDs/RemovePIDs from the `Selector` interface. So the finding criteria types are just user-configuration types without extra logic.
2. Create a `DynamicPIDSelector` (or whatever type) concrete type that is unique to OBI. Probably needs to be preloaded by the aggregation of all the user-provided `target_pids` configurations.
3. The instrumenter or appo11y types will hold a reference to the `DynamicPIDSelector` type. It can call AddPIDs/RemovePIDs directly without the interface.
4. Only Matcher should know about dynamic PID selection. Matcher already owns ProcessHistory, so it is the right place to: (1) treat runtime PIDs as an override or supplement, (2) emit synthetic deletes from `Removed`.

This also would avoid most of the `WithPIDSelectorChangeNotifier`, `WithPIDSelector`, `Set..` wiring logic. You can pas one dependency into discovery, not two. Replace WithPIDSelector + WithPIDSelectorChangeNotifier with one WithPIDController(...), or keep it entirely internal if vendored users do not need it.

pkg/appolly/discover/watcher_proc.go
```diff
@@ -67,8 +67,8 @@ func wplog() *slog.Logger {
 }
 
 // ProcessWatcherFunc polls every PollInterval for new processes and forwards either new or deleted process PIDs
-// as well as PIDs from processes that setup a new connection
-func ProcessWatcherFunc(cfg *obi.Config, ebpfContext *ebpfcommon.EBPFEventContext, output *msg.Queue[[]Event[ProcessAttrs]]) swarm.RunFunc {
+// as well as PIDs from processes that setup a new connection.
+func ProcessWatcherFunc(cfg *obi.Config, ebpfContext *ebpfcommon.EBPFEventContext, output *msg.Queue[[]Event[ProcessAttrs]], pidSelector services.Selector) swarm.RunFunc {
```
mariomac: I think that once we implement my suggestion for `criteria.go`, we can revert all the changes in ProcessWatcher because the selection logic is really handled in the Matcher node.

pkg/instrumenter/opts.go
```diff
@@ -12,6 +13,29 @@ import (
...
+	AddTargetPIDs(pids ...int)
+	RemoveTargetPIDs(pids ...int)
+	TargetPIDs() []app.PID
+}
+
+// WithTargetPIDsUpdater calls the given function with a TargetPIDsUpdater after the App O11y
+// instrumenter is created. Callers receive the App O11y instrumenter (as this interface); store it
+// and call AddTargetPIDs/RemoveTargetPIDs at runtime to change which PIDs are instrumented, or
+// TargetPIDs to inspect the currently tracked set.
+func WithTargetPIDsUpdater(onReady func(TargetPIDsUpdater)) Option {
```
mariomac: Yeah! I just realized it and I removed my initial comment just before your reply.

pkg/appolly/services/criteria.go
damemi: agreed, I like this suggestion. I was trying to wire these changes through without changing too much of the existing pathways, but that made the overall approach look weird with the problems you're pointing out. This can definitely be more like what you're saying

pkg/instrumenter/opts.go
damemi: ah ok lol, I'll remove mine then too :)

pkg/internal/appolly/appolly.go
```diff
@@ -96,16 +98,26 @@ func New(ctx context.Context, ctxInfo *global.ContextInfo, config *obi.Config) (
...
-		ctxInfo:           ctxInfo,
-		tracersWg:         &sync.WaitGroup{},
-		tracesInput:       tracesInput,
-		processEventInput: processEventsInput,
-		bp:                bp,
-		peGraphBuilder:    swi,
-		ebpfEventContext:  ebpfcommon.NewEBPFEventContext(),
-	}, nil
+	var sel *discover.DynamicPIDSelector
+	if v := ctxInfo.AppO11y.DynamicPIDSelector; v != nil {
```
grcevski, line 102: I'm reviewing this a second time now, do we need to seed the DynamicPidSelector from the target_pids we read from the config? If we configure a target_pids based on the previous PR and add a dynamic selector, do they work together?

pkg/appolly/discover/dynamic_pid_selector.go
```diff
@@ -0,0 +1,155 @@
...
+	out := make([]app.PID, len(d.pids))
+	for i, p := range d.pids {
+		out[i] = app.PID(p)
+	}
+	return out, true
+}
+
+// AddPIDs adds PIDs to the set (deduplicated).
+func (d *DynamicPIDSelector) AddPIDs(pids ...uint32) {
+	if len(pids) == 0 {
```
grcevski, line 81: Do we need a notify path here too? What if we dropped OBI in an existing environment and wanted to dynamically add a process that has already been running a while. I think watcher_proc will remember it as something it has seen and unless something tells the pipeline to conside this new PID we may not re-evaluate?

pkg/appolly/discover/dynamic_pid_selector.go
```diff
@@ -0,0 +1,155 @@
...
+			continue
+		}
+		removedPIDs = append(removedPIDs, app.PID(p))
+	}
+	d.pids = newPids
+	d.mu.Unlock()
+	d.notifyRemoved(removedPIDs)
+}
+
+func (d *DynamicPIDSelector) notifyRemoved(removedPIDs []app.PID) {
```
grcevski, line 125: Is there a timing issue here if RemovePIDs is called twice and matcher hasn't drained the first PID? Could this cause the second pid to never be uninstrumented?

pkg/instrumenter/instrumenter_test.go
```diff
@@ -39,3 +45,38 @@ func TestServiceNameTemplate(t *testing.T) {
...
+			},
+		},
+	}
+	require.True(t, cfg.Enabled(obi.FeatureAppO11y), "test config must enable App O11y")
+
+	opts := []Option{WithDynamicPIDSelector(sel)}
+	done := make(chan error, 1)
+	go func() { done <- Run(ctx, cfg, opts...) }()
+
+	time.Sleep(2 * time.Second)
```
grcevski, line 72: This might cause flakyness in CI, do you need to know somehow that FindAndInstrument has started?

pkg/appolly/discover/matcher.go
```diff
@@ -107,16 +153,41 @@ func (m *Matcher) filter(events []Event[ProcessAttrs]) []Event[ProcessMatch] {
...
 }
 
 func (m *Matcher) matchCriteria(obj ProcessAttrs, proc *services.ProcessInfo) *ProcessMatch {
-	criteria := make([]services.Selector, 0, len(m.Criteria))
-	for i := range m.Criteria {
-		if m.matchProcess(&obj, proc, m.Criteria[i]) && !m.isExcluded(&obj, proc) {
-			criteria = append(criteria, m.Criteria[i])
+	effectiveCriteria := m.Criteria
+	if m.DynamicPIDs != nil {
+		effectiveCriteria = append([]services.Selector{m.DynamicPIDs.AsSelector()}, m.Criteria...)
```
grcevski: This would allocate on every call to matchCriteria, can we build it once in criteriaMatcherProvider?

pkg/internal/appolly/appolly.go
damemi: I was originally doing some seeding from the static config to DynamicSelector, but I realized for a dynamic use case that's probably just unnecessary, since the dynamic caller can just seed *using* the DynamicSelector.Add() if they want to. So in this change, I'm actually not even handling any static config if DynamicSelector is present.

To that extent, I wonder if the end-user-facing static PIDs config I added is even necessary at all. I thought it would be crucial for dynamic PID selection, but now I see that the path to wire that up doesn't really need to touch the static config.

grcevski: I actually like your target pid selection, I've been using it :) Sometimes I have more than one python process and it's easy with your option to pick the one I want :).

If it's out of scope, then don't worry about it. We can log a debug message if there's one defined in the config and we are using the dynamic selector to tell the end user that this is not supported. I thought you maybe missed seeding the pids. I personally think it's a good idea to seed them from the config if there are any, but I don't mind it like this.

pkg/appolly/discover/matcher.go
damemi: good catch this is redundant

pkg/instrumenter/instrumenter_test.go
damemi: nope this was leftover from my original approach when we were waiting for the callback

pkg/appolly/discover/dynamic_pid_selector.go
damemi: good point, what I ended up doing here was give dynamicselector an addqueue+removequeue, so it can store pending calls and process them in batches. do you have any other suggestions for an approach there?

pkg/appolly/discover/dynamic_pid_selector.go
damemi: added something similar to the remove channel, but it's an add notify to the watcher_proc that tells it to "forget this pid" so the next snapshot picks it up and matches it to the dynamic selector

pkg/appolly/discover/dynamic_pid_selector.go
grcevski: I think that should work and we can fix it if it causes you deadlocks.

PR #1431, opened 2026-03-03
Replace largeBuffer []byte with chunked LargeBuffer

rafaelroquetto:
The previous `largeBuffer` type was a plain `struct{ buf []byte }`. When a TCP payload spans multiple ring-buffer events, each `largeBufferActionAppend` call did `append(lb.buf, chunk...)`, which copies the entire accumulated buffer every time it grows past capacity.

Replace it with LargeBuffer, a chunked store that keeps each ring-buffer chunk as an independent allocation. Appending is always a single copy into a freshly allocated slice plus one pointer append into the chunk index. Previously written data is never moved. Reading is handled by `LargeBufferReader` (cursor-based, `io.Reader`-compatible) for sequential parsers, and by `UnsafeViewAt/UnsafeView` (zero-copy within a chunk, scratch-backed across boundaries) for parsers that need all bytes at once.

Parsers that only need the full payload are changed to take `*LargeBuffer` and call `UnsafeView()` directly, eliminating the reader allocation at each call site. Parsers doing genuine sequential reads (`handlePostgres`, `Kafka`, `http.ReadRequest/ReadResponse`) keep `*LargeBufferReader`. The `kafkaparser` sub-package is migrated from raw `[]byte` slices to a `byteReader` interface so it can consume a `LargeBufferReader` without importing ebpfcommon.

The new `large_buffer_test.go` covers construction, all cursor operations, random-access methods, scalar helpers, zero-alloc hot paths, and UnsafeView stale-slice semantics.

---
This is the foundation PR. Once merged, individual parsers can be migrated in follow-up PRs to take advantage of chunked reads, for example, avoiding the full-buffer materialisation in Postgres, or streaming Kafka frames directly from a LargeBufferReader without copying into an intermediate `[]byte` first. For now, I tried to keep the code somewhat consistent.

I recommend reviewing `pkg/ebpf/common/large_buffer.go` as that contextualises the rest of the PR.

## Checklist

### DIFF
--- a/pkg/ebpf/common/http_transform.go
+++ b/pkg/ebpf/common/http_transform.go
@@ -243,10 +243,6 @@ func HTTPInfoEventToSpan(parseCtx *EBPFParseContext, event *BPFHTTPInfo) (reques
 // HTTP response buffers might have been sent incomplete, before the full body.
 // Try to parse the original buffer first, if an EOF is encountered, append an empty
 // body to the buffer and try again.
-//
-// HTTP response buffers might have been sent incomplete, before the full body.
-// Try to parse the original buffer first, if an EOF is encountered, append an empty
-// body to the buffer and try again.
 func httpSafeParseResponse(responseBuffer *LargeBuffer, req *http.Request) (*http.Response, error) {
 	rd := bufio.NewReader(responseBuffer.NewReader())
 	resp, err := http.ReadResponse(rd, req)
--- a/pkg/ebpf/common/kafka_detect_transform.go
+++ b/pkg/ebpf/common/kafka_detect_transform.go
@@ -67,9 +67,6 @@ func ProcessPossibleKafkaEvent(event *TCPRequestInfo, pkt *LargeBufferReader, rp
 }
 
 func ProcessKafkaEvent(pkt *LargeBufferReader, rpkt *LargeBufferReader, kafkaTopicUUIDToName *simplelru.LRU[kafkaparser.UUID, string]) (*KafkaInfo, bool, error) {
-	if pkt == nil {
-		return nil, true, errors.New("nil request buffer")
-	}
 	hdr, err := kafkaparser.ParseKafkaRequestHeader(pkt)
 	if err != nil {
 		return nil, true, err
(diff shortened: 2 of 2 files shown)


pkg/ebpf/common/large_buffer.go
```diff
@@ -0,0 +1,561 @@
...
+
+	// Slow path: crosses chunk boundary — copy into reusable scratch.
+	if cap(lb.scratch) < n {
+		lb.scratch = make([]byte, n)
+	}
+
+	lb.scratch = lb.scratch[:n]
+
+	for filled := 0; filled < n; {
+		copied := copy(lb.scratch[filled:], lb.chunks[ci][off:])
```
mmat11, line 187: is the access at `lb.chunks[ci]` implicitly safe because of the previous check on `lb.total`?

pkg/ebpf/common/http_transform.go
```diff
@@ -242,46 +243,52 @@ func HTTPInfoEventToSpan(parseCtx *EBPFParseContext, event *BPFHTTPInfo) (reques
 // HTTP response buffers might have been sent incomplete, before the full body.
 // Try to parse the original buffer first, if an EOF is encountered, append an empty
 // body to the buffer and try again.
-func httpSafeParseResponse(responseBuffer []byte, req *http.Request) (*http.Response, error) {
-	rd := bufio.NewReader(bytes.NewReader(responseBuffer))
+//
+// HTTP response buffers might have been sent incomplete, before the full body.
```
mmat11, line 243: this comment is repeated from line 243

pkg/ebpf/common/kafka_detect_transform.go
```diff
@@ -59,25 +66,28 @@ func ProcessPossibleKafkaEvent(event *TCPRequestInfo, pkt []byte, rpkt []byte, k
 	return k, ok, err
 }
 
-func ProcessKafkaEvent(pkt []byte, rpkt []byte, kafkaTopicUUIDToName *simplelru.LRU[kafkaparser.UUID, string]) (*KafkaInfo, bool, error) {
-	hdr, offset, err := kafkaparser.ParseKafkaRequestHeader(pkt)
+func ProcessKafkaEvent(pkt *LargeBufferReader, rpkt *LargeBufferReader, kafkaTopicUUIDToName *simplelru.LRU[kafkaparser.UUID, string]) (*KafkaInfo, bool, error) {
+	if pkt == nil {
```
mmat11: is this possible? should we also check rpkt?

pkg/ebpf/common/http_transform.go
rafaelroquetto: frigging Claude :P - nice catch, will fix it

pkg/ebpf/common/kafka_detect_transform.go
rafaelroquetto: you are right, it is not possible - I will remove it.

pkg/ebpf/common/large_buffer.go
rafaelroquetto: Yes, the check relies on ` absOff+n > lb.total` and on the fact that AppendChunk always does `lb.total += len(data)` in lockstep with the append, so `lb.total` == sum of the lens of the chunks always holds.

PR #1441, opened 2026-03-03
ci: allow conditional secrets for docker login

skl:
Turns out `if: secrets` is not valid GitHub Actions workflow syntax on forks, so to make this docker login step optional we need to first use another step to actually check the secret.

Then use output from above step as the condition. This should be more compatible.

To take advantage of this during the release process, the secret needs to be passed in from the parent workflow ~~using `secrets: inherit`~~ and registered as an optional secret in the `workflow_call` definition.

### DIFF
--- a/internal/test/oats/kafka/docker-compose-obi-java-kafka.yml
+++ b/internal/test/oats/kafka/docker-compose-obi-java-kafka.yml
@@ -8,7 +8,7 @@ services:
       - "2181:2181"
 
   kafka:
-    image: confluentinc/cp-kafka:8.2.0@sha256:acbbf674f2ed40e5d0a8ca51beb0f00692c866fc22b5ce06f8cadbdc54cd4436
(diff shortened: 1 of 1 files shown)


mmat11 (approved):
🤞🏻

.github/workflows/pull_request_integration_tests.yml
```diff
@@ -113,8 +117,19 @@ jobs:
           ref: ${{ inputs.ref || github.sha }}
           persist-credentials: false
 
+      - name: Check for Docker Hub token
+        id: check-docker-token
+        run: |
+          if [ -n "${DOCKER_TOKEN}" ]; then
```
MrAlias, line 123: Should this be:


```suggestion
          if [ -n "${DOCKER_TOKEN_EBPF_INSTRUMENTATION}" ]; then
```

Or is this intentionally different?

MrAlias: Ah, I now see below. Disregard.

PR #1443, opened 2026-03-03
Revert merges with missing CI

MrAlias:
- https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1413
- https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1410
- https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1414


internal/test/vm/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.23@sha256:25109184c71bdad752c8312a8623239686a9a2071e8825f20acb8f2198c3f659
+FROM alpine:3.20@sha256:a4f4213abb84c497377b8544c81b3564f313746700372ec4fe84653e4fb03805
```
skl, line 1: this line is the breaking change for the VM workflow, the other alpine updates should be ok I think

MrAlias: I can take a look at adding back the other updates after merging all these reverts. Better to have the full CI system evaluate those changes as well when it is working completely again.

PR #1448, opened 2026-03-04
Allow adding http headers to spans

NimrodAvni78:
## Checklist

Allow for extracting HTTP headers into spans by set of rules, inspired structure by [config-v2 structure](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1351/changes#diff-9c5c3212f4f63b10031cdd20d2456ad6121498022013980bf3a50b04399ee54dR53)
for now it is called http enrichment in the config and code, since i plan to add support for http body rules as well
if you have a better name please suggest 

example:

```yaml
ebpf:
  payload_extraction:
    http:
      enrichment:
        enabled: false
        policy:
          # exclude by default unless matched by an include or obfuscate rule.
          default_action: exclude
          # First matching rule decides outcome.
          match_order: first_match_wins
          # Obfuscate String to use on action: obfuscate
          obfuscation_string: "***"
        # Ordered include/exclude/obfuscate rules.
        rules:
          - action: obfuscate
            name: obfuscate-auth
            description: Obfuscate all autherization and token headers
            scope: request
            type: headers
            match:
              case_sensetive: false
              patterns:
                - "*auth*"
                - "*credential*"
                - "*cookie*"
                - "*token*"
                - "*api-key*"
                - "*apikey*"
          - action: include
            name: include-aws-headers
            description: include all aws headers
            scope: both
            type: headers
            match:
              case_sensetive: false
              patterns:
                - "*aws*"
```

### DIFF
--- a/internal/test/integration/components/jaeger/jaeger.go
+++ b/internal/test/integration/components/jaeger/jaeger.go
@@ -6,6 +6,7 @@
 package jaeger // import "go.opentelemetry.io/obi/internal/test/integration/components/jaeger"
 
 import (
+	"encoding/json"
 	"fmt"
 	"regexp"
 	"strings"
@@ -144,12 +145,29 @@ func FindIn(tags []Tag, key string) (Tag, bool) {
 	return Tag{}, false
 }
 
+// tagParseStringSlice attempts to parse a JSON-encoded string array.
+// Jaeger 1.x serializes OTLP StringSlice attributes as a single string
+// containing the JSON array, e.g. `["value-one","value-two"]`.
+func tagParseStringSlice(s string) ([]string, bool) {
+	if len(s) < 2 || s[0] != '[' {
+		return nil, false
+	}
+	var parsed []string
+	if err := json.Unmarshal([]byte(s), &parsed); err != nil {
+		return nil, false
+	}
+	return parsed, true
+}
+
 // TagFirstStringValue extracts the first string value from a tag.
-// Handles both plain string values and array values (as returned by Jaeger
-// for OTLP StringSlice attributes).
+// Handles plain strings, JSON-encoded string arrays (Jaeger 1.x),
+// and native array values.
 func TagFirstStringValue(tag Tag) (string, bool) {
 	switch v := tag.Value.(type) {
 	case string:
+		if vals, ok := tagParseStringSlice(v); ok && len(vals) > 0 {
+			return vals[0], true
+		}
 		return v, true
 	case []any:
 		if len(v) > 0 {
@@ -162,9 +180,14 @@ func TagFirstStringValue(tag Tag) (string, bool) {
 }
 
 // TagStringValues extracts all string values from a tag.
+// Handles plain strings, JSON-encoded string arrays (Jaeger 1.x),
+// and native array values.
 func TagStringValues(tag Tag) []string {
 	switch v := tag.Value.(type) {
 	case string:
+		if vals, ok := tagParseStringSlice(v); ok {
+			return vals
+		}
 		return []string{v}
 	case []any:
 		result := make([]string, 0, len(v))


pkg/config/payload_extraction.go
```diff
@@ -3,12 +3,20 @@
...
+	"gopkg.in/yaml.v3"
+)
+
 type PayloadExtraction struct {
 	HTTP HTTPConfig `yaml:"http"`
 }
 
 func (p PayloadExtraction) Enabled() bool {
-	return p.HTTP.GraphQL.Enabled || p.HTTP.Elasticsearch.Enabled || p.HTTP.AWS.Enabled || p.HTTP.SQLPP.Enabled || p.HTTP.OpenAI.Enabled
+	return p.HTTP.GraphQL.Enabled || p.HTTP.Elasticsearch.Enabled || p.HTTP.AWS.Enabled || p.HTTP.SQLPP.Enabled || p.HTTP.OpenAI.Enabled || p.HTTP.Enrichment.Enabled
```
rafaelroquetto: very minor nit: eventually we should start wrapping this line

pkg/config/payload_extraction.go
```diff
@@ -51,3 +61,148 @@ type OpenAIConfig struct {
...
+	// Match defines the matching criteria for this rule
+	Match HTTPParsingMatch `yaml:"match"`
+}
+
+// HTTPParsingRuleType specifies the target of a parsing rule.
+type HTTPParsingRuleType string
+
+const (
+	HTTPParsingRuleTypeHeaders HTTPParsingRuleType = "headers"
+)
```
rafaelroquetto: is it possible to use an integer type here instead of carrying a string forward? Analogous to what we do with [ContextPropagationMode](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/5f7d95562d02c22c40e77c121b4006498098d0d8/pkg/config/ebpf_tracer.go#L16) but simpler, as this is not a flag

pkg/config/payload_extraction.go
```diff
@@ -51,3 +61,148 @@ type OpenAIConfig struct {
...
+
+func (t *HTTPParsingRuleType) UnmarshalText(text []byte) error {
+	str := HTTPParsingRuleType(strings.TrimSpace(strings.ToLower(string(text))))
```
rafaelroquetto: there's nothing wrong with ToLower here and I personally do not have a strong opinion, but I just wanted to point out we don't do it consistently throughout the code (like ContextPropagationMode above will error out if you type HeAders or HEADERS instead of "headers").

pkg/config/payload_extraction.go
```diff
@@ -51,3 +61,148 @@ type OpenAIConfig struct {
...
+		}
+		m.Regex = append(m.Regex, re)
+	}
+	return nil
+}
+
+// HTTPParsingAction represents the action for a generic parsing rule or default policy.
+type HTTPParsingAction string
+
+const (
```
rafaelroquetto, line 171: Same here, _if possible_ I'd turn this into a uint8 and leave the strings for the marshalling/unmarshalling process

pkg/config/payload_extraction.go
```diff
@@ -51,3 +61,148 @@ type OpenAIConfig struct {
...
+	case HTTPParsingActionInclude, HTTPParsingActionExclude, HTTPParsingActionObfuscate:
+		*a = str
+		return nil
+	default:
+		return fmt.Errorf("invalid parsing action: %q (valid: include, exclude, obfuscate)", string(text))
+	}
+}
+
+// HTTPParsingAction represents the action for a http parsing rule or default policy.
+type HTTPParsingScope string
```
rafaelroquetto: ditto

pkg/ebpf/common/http/http_enrichment.go
```diff
@@ -0,0 +1,97 @@
...
+	for name, values := range req.Header {
+		action := resolveHeaderAction(name, cfg.Rules, cfg.Policy, config.HTTPParsingScopeRequest)
+		applyHeaderAction(action, name, values, reqHeaders, cfg.Policy.ObfuscationString)
+	}
+
+	// Process response headers
+	for name, values := range resp.Header {
+		action := resolveHeaderAction(name, cfg.Rules, cfg.Policy, config.HTTPParsingScopeResponse)
+		applyHeaderAction(action, name, values, respHeaders, cfg.Policy.ObfuscationString)
+	}
```
rafaelroquetto: nit: I'd split this further to improve readability (and reduce duplication as a bonus)

something like

```suggestion
    reqHeaders := processHeaderAction(req.Header, cfg, config.HTTPParsingScopeRequest)
	
	respHeaders := processHeaderAction(resp.Header, cfg, config.HTTPParsingScopeResponse)

```

you don't even need comments then

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -298,6 +299,22 @@ var (
 	spanMetricsSkip     = attribute.Bool(string(attr.SkipSpanMetrics), true)
 )
 
+// httpHeaderAttributes converts extracted HTTP headers to OTel span attributes
+// following the semantic convention: http.request.header.<key> and http.response.header.<key>
+// where <key> is the lowercase, hyphen-separated header name.
+func httpHeaderAttributes(span *request.Span) []attribute.KeyValue {
+	var attrs []attribute.KeyValue
```
rafaelroquetto: we should always allocate with the capacity specified when we know it upfront to prevent needless reallocs

```suggestion
	attrs := make([]attribute.KeyValue, 0, len(span.RequestHeaders) + len(span.ResponseHeaders))
```

pkg/config/payload_extraction.go
NimrodAvni78: is this something that should be done by the linter or by us?
i can do it

rafaelroquetto: maybe the linter - I will investigate if there's a way to automate this - @MrAlias thoughts?

pkg/ebpf/common/http/http_enrichment.go
```diff
@@ -0,0 +1,97 @@
...
+
+// EnrichHTTPSpan applies generic HTTP parsing rules to extract headers into the span.
+// Regex patterns in rules are already compiled during YAML deserialization.
+// Unlike other parsers, this enriches the span with headers rather than replacing it.
+func EnrichHTTPSpan(
+	baseSpan *request.Span,
+	req *http.Request,
+	resp *http.Response,
+	cfg config.EnrichmentConfig,
+) bool {
```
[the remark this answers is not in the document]
NimrodAvni78: no need since this check is done in the call site

NimrodAvni78:
@rafaelroquetto decided to go with glob and so a significant performance improvement, I think its general enough for what we want users to match, especially with case-insensitivity support

pkg/ebpf/common/http/http_enrichment.go
```diff
@@ -0,0 +1,111 @@
...
+	cfg config.EnrichmentConfig,
+	scope config.HTTPParsingScope,
+) map[string]string {
+	result := make(map[string]string)
+	for name, values := range headers {
+		action := resolveHeaderAction(name, cfg.Rules, cfg.Policy, scope)
+		applyHeaderAction(action, name, values, result, cfg.Policy.ObfuscationString)
+	}
+	return result
+}
```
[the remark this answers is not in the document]
rafaelroquetto: yes please

pkg/config/payload_extraction.go
```diff
@@ -51,3 +68,218 @@ type OpenAIConfig struct {
 	// Enable OpenAI payload extraction and parsing
 	Enabled bool `yaml:"enabled" env:"OTEL_EBPF_HTTP_OPENAI_ENABLED" validate:"boolean"`
 }
+
+// EnrichmentConfig configures generic HTTP header and payload extraction.
+type EnrichmentConfig struct {
+	// Enable generic HTTP header and payload extraction
+	Enabled bool `yaml:"enabled" env:"OTEL_EBPF_HTTP_GENERIC_PARSING_ENABLED" validate:"boolean"`
```
grcevski: I wonder if it would be better to name the env variable and the externally visible config to explicitly mention the word "headers"?

NimrodAvni78: I want this part to also include the http body extraction rules
my plans was to have it in the same rule list with a different `HTTPParsingRuleType`
anyway i wanted to change from `GENERIC_PARSING` to `ENRICHMENT`
WDYT?

grcevski: Ah I see, sounds good then, ENRICHMENT is better :)

pkg/ebpf/common/http_transform.go
```diff
@@ -172,6 +172,10 @@ func httpRequestResponseToSpan(parseCtx *EBPFParseContext, event *BPFHTTPInfo, r
 		}
 	}
 
+	if parseCtx != nil && parseCtx.payloadExtraction.HTTP.Enrichment.Enabled {
```
grcevski, line 175: What do you think about running the enrichment before we run any of the payload extraction? I can see that we'd want to extract certain headers regardless of the payload extraction outcome. Perhaps we can create the enriched span first and then let the payload extractors handle the remaining on top of the initially created span.

pkg/ebpf/common/http/http_enrichment.go
```diff
@@ -0,0 +1,118 @@
...
+	return policy.DefaultAction
+}
+
+// scopeApplies returns true if the rule scope covers the given header source.
+func scopeApplies(ruleScope config.HTTPParsingScope, headerSource config.HTTPParsingScope) bool {
+	return ruleScope == config.HTTPParsingScopeAll || ruleScope == headerSource
+}
+
+// applyHeaderAction adds the header to the map based on the resolved action.
+func applyHeaderAction(
```
grcevski, line 102: If I understand correctly this will do last one wins for duplicate headers. I believe the HTTP header spec allows for duplicates and they should be concatenated. For example:

Accept: text/json
Accept: http/proto-buf

is the same as `Accept: text/json, http/proto-buf`. 

https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

pkg/ebpf/common/http_transform.go
NimrodAvni78: I wonder how that will work.
Because the purpose of those classifiers is to turn the http span into another span (Aws/ graphql / OpenAI ) and adding http headers to those spans doesn't make much sense since it's not part of the semantic convention

grcevski: I guess it's true, if it's not part of the convention it might look weird. I imagined someone for example building a service that say uses S3 or OpenAI. And they have their own customers and for example, all incoming requests have `X-Customer-Id` header. I think those folks would likely want to get that associated, if they are choosing it, to these spans too. But I think I was overthinking it, these spans that we detect at the moment will mostly be client spans made by the client libraries, and those client libraries will not be adding these headers.

I'm good with this to go as is.

NimrodAvni78: maybe in the future we can add the http spans as child spans for the openAI / aws instrumentation and apply the http enrichments there
for example, the field `suppressInternalInstrumentation` in the elastic instrumentation for nodejs
https://github.com/aspecto-io/opentelemetry-ext-js/blob/master/packages/instrumentation-elasticsearch/README.md

PR #1454, opened 2026-03-04
bpf: nodejs: set trace context in shared map via async callbacks

mmat11:
Similarly to how done for the Go runtime, also add trace context for NodeJS runtime to the shared map

## Checklist

### DIFF
--- a/bpf/generictracer/nodejs.c
+++ b/bpf/generictracer/nodejs.c
@@ -17,6 +17,60 @@
 
 #include <shared/obi_ctx.h>
 
+enum {
+    k_delim_offset = 13,
+    k_fd1_offset = 14,
+    k_fd2_offset = 18,
+    k_ctx_fd_offset = 18,
+    k_max_fd_digits = 4
+};
+
+static __always_inline int handle_async_switch(char *buf, const u64 pid_tgid) {
+    u32 fd = 0;
+    for (u8 i = 0; i < k_max_fd_digits; ++i) {
+        fd *= 10;
+        fd += buf[k_ctx_fd_offset + i] - '0';
+    }
+
+    bpf_dbg_printk("nodejs_async_switch: %s, pid_tgid = %llx, fd = %u", buf, pid_tgid, fd);
+
+    const fd_key fkey = {.pid_tgid = pid_tgid, .fd = (s32)fd};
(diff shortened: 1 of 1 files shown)


bpf/generictracer/nodejs.c
```diff
@@ -8,27 +8,42 @@
...
+    // the obi nodejs agent (fdextractor.js) passes signals to the ebpf layer
+    // by invoking uv_fs_access() with a fake path. Two formats are used:
+    //
+    // 1. fd pair correlation (outgoing -> incoming):
+    //    /dev/null/obi/<fd1><fd2>  — each fd is a left-zero-padded 4-digit number
+    //
+    // 2. async context switch (before-hook fires before each JS callback):
+    //    /dev/null/obi-ctx/<fd>    — 4-digit incoming fd for the current async context
+    //
+    // Both paths share the prefix "/dev/null/obi" (13 chars). The character at
```
rafaelroquetto, line 89: clever!

bpf/generictracer/nodejs.c
```diff
@@ -38,6 +53,38 @@ int BPF_KPROBE(obi_uv_fs_access, void *loop, void *req, const char *path) {
         return 0;
     }
 
+    const u64 pid_tgid = bpf_get_current_pid_tgid();
+
+    // Async context switch: /dev/null/obi-ctx/XXXX
+    // Fires from the async_hooks 'before' callback in fdextractor.js to refresh
+    // traces_ctx_v1 before each JS callback.
+    if (buf[13] == '-') {
```
rafaelroquetto:
```suggestion
    if (buf[k_delim_offset] == '-') {
```

also, can we split this into two static inline subroutines, so that it reads:

```
const char delim = buf[k_delim_offset];

switch (delim) {
    case '-':
      handle_obi_ctx_correlation(...); // or whatever name
      break;
    case '/':
      handle_fd_correlation(...); // or whatever name
      break;
     default:
       bpf_dbg_printk("invalid delim: '%c'", delim);
       break;
}

return 0;
```

rafaelroquetto (approved):
just a nit, feel free to disregard!

LGTM

bpf/generictracer/nodejs.c
mmat11: good idea 👍🏻

PR #1461, opened 2026-03-04
Remove the gradle-wrapper.jar

MrAlias:
Stop tracking the binary Gradle wrapper JAR in git history. Update the Gradle wrapper JAR/workflow from the Java integration and standardizes builds on the host/container `gradle` CLI.

## Changes

- Removed wrapper artifacts/scripts from `pkg/internal/java` (`gradle-wrapper.jar`, `gradlew`, `gradlew.bat`, wrapper properties).
- Updated Java build/test/format paths to use `gradle` directly in:
  - `.github/workflows/java-agent.yml`
  - `Makefile`
  - `Dockerfile`
  - `javaagent.Dockerfile`
  - `internal/test/integration/components/obi/Dockerfile-with-javaagent`
- Updated Java docs to reflect current prerequisites and build commands:
  - `pkg/internal/java/README.md`
  - `pkg/internal/java/agent/src/jmh/.../BENCHMARK_README.md`
- Refreshed embedded Java agent artifact: `pkg/internal/java/embedded/obi-java-agent.jar`.

## Fixes

Added fix for Java 8 CI correctness by enforcing Java 8 API targeting at compile time (`options.release = 8` in the Java Gradle config). This prevents newer-JDK method signatures from leaking into compiled classes (e.g., `ByteBuffer.flip()/position()` return types), which caused `NoSuchMethodError` when tests were executed on JDK 8. The Java 8 matrix path now runs successfully with Gradle on the pinned JDK and tests on the matrix JDK.

## Validation

- Java notices are up to date (`make java-notices-update` produced no diffs).
- Updated Docker build paths were validated by building changed Docker targets successfully.

### DIFF
--- a/pkg/internal/java/build.gradle.kts
+++ b/pkg/internal/java/build.gradle.kts
@@ -2,6 +2,8 @@ plugins {
     java
 }
 
+import org.gradle.api.tasks.compile.JavaCompile
+
 group = "io.opentelemetry.obi"
 version = "0.1.0"
 
@@ -15,6 +17,10 @@ subprojects {
 
     repositories {
         mavenCentral()
+    }
+
+    tasks.withType<JavaCompile>().configureEach {
+        options.release.set(8)
     }
 }
 


grcevski (approved):
LGTM! Yeah we should remove them, the build will make them anyway.

PR #1485, opened 2026-03-06
Use Server Spans when instrumenting SQL databases

marctc:
Currently when we instrument a MySQL server and  OBI generates a client span for each incoming SQL call.
It should be a server span.

Fixes https://github.com/grafana/beyla/issues/1695

### DIFF
--- a/internal/test/oats/sql/docker-compose-obi-mysqlserver.yml
+++ b/internal/test/oats/sql/docker-compose-obi-mysqlserver.yml
@@ -1,0 +1,55 @@
+version: "3.9"
+services:
+  mysqlserver:
+    image: mysql:9.6@sha256:e5dc14f6e01c3e577e669337d2855c6d1561b30d8ef2c592e63e4e8a9a52650a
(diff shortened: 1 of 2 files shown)


grcevski:
I don't think we should be dropping the spans just because there's no semantic convention. We can create a server span or internal span. A user may want to instrument MySQL, that's not out of the realm of possibility.

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -475,13 +475,15 @@ func TraceAttributesSelector(span *request.Span, optionalAttrs map[attr.Name]str
 			request.PeerService(request.PeerServiceFromSpan(span)),
 			request.ServerPort(span.HostPort),
 		}
-	case request.EventTypeSQLClient:
+	case request.EventTypeSQLClient, request.EventTypeSQLServer:
```
grcevski, line 478: Can we add a unit test for this change please?

PR #1500, opened 2026-03-09
Parameterize ringBufForwarder and update SharedRingbuf/ForwardRingbuf to generic

pinoOgni:
In preparation of #1273, we will need to use `ringBufForwarder` and `SharedRingbuf/ForwardRingbuf` not only for `request.Span`, so this PR makes them generic with a `[T any]` type parameter. It also changes all event handler signatures in various packages.

### DIFF
--- a/pkg/ebpf/common/common.go
+++ b/pkg/ebpf/common/common.go
@@ -6,6 +6,7 @@ package ebpfcommon // import "go.opentelemetry.io/obi/pkg/ebpf/common"
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


pkg/ebpf/common/ringbuf.go
```diff
@@ -39,75 +38,82 @@ var readerFactory = func(rb *ebpf.Map) (ringBufReader, error) {
...
+// If the shared ring buffer forwarder already exists, subsequent calls return a no-op
+// that simply waits for context cancellation.
+func SharedRingbuf[T any](
 	eventContext *EBPFEventContext,
-	parseContext *EBPFParseContext,
 	cfg *config.EBPFTracer,
-	filter ServiceFilter,
 	ringbuffer *ebpf.Map,
+	parse func(*ringbuf.Record) (T, bool, error),
+	filter func([]T) []T, // nil = no batch filter
```
rafaelroquetto: could we `type` these functions to improve readability? e.g. `type RingBufParseFunc func(....)...`

pinoOgni: great idea!

pkg/ebpf/common/common.go
```diff
@@ -187,7 +187,7 @@ type EBPFParseContext struct {
 
 type EBPFEventContext struct {
 	CommonPIDsFilter ServiceFilter
-	SharedRingBuffer *ringBufForwarder
+	SharedRingBuffer any
```
mmat11: would it be a mess to make the struct generic? or at least define an interface for the forwarder?

pinoOgni: > would it be a mess to make the struct generic? 

If you mean to do something like `type EBPFEventContext[T any] struct {..}` IMO yes, it will be a mess. 

> or at least define an interface for the forwarder?

If you mean to do something like:
```
type EBPFEventContext struct {
  ...
  SharedRingBuffer sharedForwarder
  ...
}
```
So we remove `any` and the type assertion, I think it's a good idea. I'll try it asap.

PR #1514, opened 2026-03-10
Verify SIGUSR1 handler is not registered before trying to inject nodejs agent

NimrodAvni78:
## Summary

Added a pre-check before sending SIGUSR1 to Node.js processes for agent injection, preventing crashes when the application has a custom SIGUSR1 handler (like `process.on('SIGUSR1', ...)`)                          
This is done by reading libuv's internal uv__signal_tree RB-tree directly from `/proc/<pid>/mem` to detect if a SIGUSR1 handler is registered, without sending any signal to the process

In addition added a check if the default debug port (9229) is already open on the target process, if so no need to send SIGUSR1 to open it, this happens if node already runs with `node --inspect` and has a custom SIGUSR1 handler, or something like `node --inspect --disable-sigusr1`. we can still inject our script without sending SIGUSR1

## Details
  The built-in inspector handler is registered via sigaction() ([inspector_agent.cc](https://github.com/nodejs/node/blob/main/src/inspector_agent.cc#L143), [node.cc](https://github.com/nodejs/node/blob/main/src/node.cc#L476)), while user JS handlers go through SignalWrap::Start ([signal_wrap.cc](https://github.com/nodejs/node/blob/main/src/signal_wrap.cc#L121)) → [libuv's signal.c](https://github.com/libuv/libuv/blob/v1.x/src/unix/signal.c#L427), which adds an entry to the uv__signal_tree internal RB-tree. By reading this tree from process memory
  and checking for signum == 10, we can distinguish user handlers from the built-in one — without any side effects.

  The signum offset (0x68) within uv_signal_s is determined by the [signal.h](https://github.com/libuv/libuv/blob/v1.x/include/uv.h#L1723) which includes UV_HANDLE_FIELDS + uv_signal_cb + int signum. This layout has been stable across all
   libuv 1.x versions (Node.js 4.x through 22.x+). All official Node.js Docker images (including alpine, slim, and distroless) ship unstripped binaries with the uv__signal_tree symbol available.

## Testing
  - Unit tests covering three scenarios: no handler, SIGUSR1 handler, other signal only (SIGINT)
  - Verified with bpftrace that uv_signal_start(signum=10) only fires for process.on('SIGUSR1', ...)
  - Verified uv__signal_tree symbol is present in node:22, node:22-slim, node:22-alpine, and gcr.io/distroless/nodejs22-debian12 images

### References
  - https://www.mehulkar.com/blog/2023/01/investigating-signal-handling-in-node.js/
  - https://docs.libuv.org/en/v1.x/signal.html

### DIFF
--- a/pkg/internal/nodejs/nodejs.go
+++ b/pkg/internal/nodejs/nodejs.go
@@ -6,9 +6,11 @@ package nodejs // import "go.opentelemetry.io/obi/pkg/internal/nodejs"
 import (
 	"debug/elf"
 	_ "embed"
-	"errors"
+	"fmt"
 	"log/slog"
+	"net"
 	"syscall"
+	"time"
 
 	"go.opentelemetry.io/obi/pkg/appolly/app/svc"
 	"go.opentelemetry.io/obi/pkg/ebpf"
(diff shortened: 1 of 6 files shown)


pkg/internal/nodejs/nodejs.go
```diff
@@ -43,13 +44,26 @@ func (i *NodeInjector) NewExecutable(ie *ebpf.Instrumentable) {
...
+func (i *NodeInjector) attachAgent(pid int, elfFile *elf.File) error {
+	// If the inspector port is already open (e.g. --inspect flag), skip SIGUSR1
+	// and inject directly.
+	if i.isInspectorOpen(pid) {
+		i.log.Debug("Node.js inspector already open, injecting agent", "pid", pid)
+		return i.inject(pid)
+	}
+
+	if elfFile != nil && hasUserSIGUSR1Handler(pid, elfFile) {
+		i.log.Debug("Node.js process has a custom SIGUSR1 handler, skipping agent injection",
```
grcevski: Can we please make this an INFO?

pkg/internal/nodejs/signal_check.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+	}
+	return 0, fmt.Errorf("symbol %q not found", name)
+}
+
+// findExeBaseAddr reads /proc/<pid>/maps to find the base virtual address
+// where the main executable is mapped. This is needed for PIE binaries where
+// ELF symbol addresses are relative to the load base.
+func findExeBaseAddr(pid int) (uint64, error) {
+	mapsPath := fmt.Sprintf("/proc/%d/maps", pid)
```
grcevski: You might be able to reuse some of the code in procmaps.go here to avoid code duplication.

pkg/internal/nodejs/nodejs.go
rafaelroquetto: In addition to that, we should probably tailor the message to say something along the lines of "... NodeJS trace correlation will not work"

pkg/internal/nodejs/nodejs.go
```diff
@@ -59,5 +73,40 @@ func (i *NodeInjector) attachAgent(pid int) error {
 	return i.inject(pid)
 }
 
+// isInspectorOpen checks if the Node.js inspector port (9229) is already
+// accepting connections in the target process's network namespace.
+// It validates that the listener is actually a Node.js inspector by
+// requesting /json/version and checking for a valid response.
```
rafaelroquetto: nit (feel free to ignore as this is highly subjective): I'd actually do without this comment - `isInspectorOpen()` kinda tells the whole story succinctly. 


**non-nit**: I am not sure if this is the best approach - we basically connect to the inspector, close the connection, just to try to connect to it again. 

How about actually doing the work in `injectFile()` ? You could attempt to connect first (as it already does), and if it fails, either send SIGUSR1 and retry or bail if SIGUSR1 is not available.

pkg/internal/nodejs/signal_check.go
```diff
@@ -0,0 +1,147 @@
...
+	sigusr1 = 10
+	// Offset of the signum field within the uv_signal_s struct (libuv 1.x, x86-64).
+	// This offset is stable across all libuv 1.x versions (used by Node.js 4.x through 22.x+).
+	// Layout: UV_HANDLE_FIELDS (0x60) + uv_signal_cb (0x08) = 0x68.
+	uvSignalSigNumOffset = 0x68
+	// Offsets of the RB-tree left/right child pointers within uv_signal_s.
+	// These are part of UV_SIGNAL_PRIVATE_FIELDS.tree_entry, which follows signum.
+	uvSignalTreeLeftOffset  = 0x70
+	uvSignalTreeRightOffset = 0x78
+	// Maximum number of tree nodes to visit to prevent runaway reads.
```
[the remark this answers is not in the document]
rafaelroquetto: Is this something we should address?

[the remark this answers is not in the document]
NimrodAvni78: already addressed, we are now using `elfFile.ByteOrder` on each read in the latest  version

pkg/internal/nodejs/nodejs.go
NimrodAvni78: fixed this to reuse the same connection and have it all under `injectFile` function

pkg/internal/nodejs/signal_check.go
NimrodAvni78: used `procs.FindLibMaps` and `procs.FindExeSymbols` after some small refactoring to find the other types of symbols beside `STT_FUNC`

pkg/internal/nodejs/nodejs.go
NimrodAvni78: applied and moved it to warn

grcevski (approved):
LGTM!

PR #1518, opened 2026-03-10
fix dependency policy script

mmat11:
Fixes `lint-dependency-policy.sh` on some shells:

```
./scripts/lint-dependency-policy.sh: line 513: ARGS[@]: unbound variable
make: *** [lint-dependency-policy] Error 1
```

### DIFF
--- a/scripts/lint-dependency-policy.sh
+++ b/scripts/lint-dependency-policy.sh
@@ -16,7 +16,6 @@ set -euo pipefail
 
 PROGNAME="$(basename "$0")"
 readonly PROGNAME
-declare -ar ARGS=("$@")
 ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 readonly ROOT_DIR
 readonly DEFAULT_BASE_REF="main"
@@ -510,4 +509,4 @@ main() {
   echo "Dependency policy lint passed for ${#files[@]} Dockerfile(s)."
 }
 
-main "${ARGS[@]}"
+main "$@"


scripts/lint-dependency-policy.sh
```diff
@@ -510,4 +510,4 @@ main() {
   echo "Dependency policy lint passed for ${#files[@]} Dockerfile(s)."
 }
 
-main "${ARGS[@]}"
+main ${ARGS[@]+"${ARGS[@]}"}
```
MrAlias: I see, the error you are seeing is likely coming from `set -u` when no arguments are passed.

`${ARGS[@]+"${ARGS[@]}"}` is a bit hard to read. What about:


```suggestion
main "${ARGS[@]-}"
```

Or, just remove the error by removing line 19 (`declare -ar ARGS=("$@")`), and passing the args directly:


```suggestion
main "$@"
```

mmat11: anything works for me :D

scripts/lint-dependency-policy.sh
```diff
@@ -510,4 +510,4 @@ main() {
   echo "Dependency policy lint passed for ${#files[@]} Dockerfile(s)."
 }
 
-main "${ARGS[@]}"
+main "$@"
```
MrAlias, line 512: The declaration of ARGS above should be removed.

mmat11:
this changes don't seem to fix it -> https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/22933672090/job/66560431503?pr=1526
closing, will fix in that PR

PR #1531, opened 2026-03-11
Generalize netolly pipeline nodes and decorators

pinoOgni:
This PR continues the work started in #1500  in preparation for #1273. 

In short, it updates the netolly pipeline and makes the majority of its nodes generic. 

Key changes include:
- Created `CommonAttrs`, which contains a list of common attributes not exclusive to `netolly` (src ip, dst ip and so on).
- Created `NetAttrs` for attributes specifically related to `netolly`.
- Updated the following decorators to be generic: `CIDR`, `GeoIP`, `RDNS`, and `Kubernetes metadata`.
- Split the `FlowDecorator` (i.e. `Decorator`) into two parts: one generic and one netolly-specific.
-  Moved all generic decorators outside of the netolly folder.

Moreover, for `CIDR` and `GeoIP` decorators, it also changes the environment variable names, accepting old names for backward compatibility.

### DIFF
--- a/pkg/export/otel/metrics_net.go
+++ b/pkg/export/otel/metrics_net.go
@@ -185,7 +185,7 @@ func (me *netMetricsExporter) Do(ctx context.Context) {
 				flowBytes, attrs := me.flowBytes.ForRecord(v)
 				flowBytes.Add(ctx, int64(v.Metrics.Bytes), metric2.WithAttributeSet(attrs))
 			}
-			if me.interZoneBytes != nil && v.CommonAttrs.SrcZone != v.CommonAttrs.DstName {
+			if me.interZoneBytes != nil && v.CommonAttrs.SrcZone != v.CommonAttrs.DstZone {
 				izBytes, attrs := me.interZoneBytes.ForRecord(v)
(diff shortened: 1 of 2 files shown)


pkg/export/otel/metrics_net.go
```diff
@@ -185,7 +185,7 @@ func (me *netMetricsExporter) Do(ctx context.Context) {
 				flowBytes, attrs := me.flowBytes.ForRecord(v)
 				flowBytes.Add(ctx, int64(v.Metrics.Bytes), metric2.WithAttributeSet(attrs))
 			}
-			if me.interZoneBytes != nil && v.Attrs.SrcZone != v.Attrs.DstZone {
+			if me.interZoneBytes != nil && v.CommonAttrs.SrcZone != v.CommonAttrs.DstName {
```
grcevski: I think this should be != v.CommonAttrs.Dst**Zone** :)

I wonder if we can add unit tests for this code, since we didn't catch it.

pinoOgni: Oh, nice catch! 

vscode autocomplete betrayed me there :) 

I'll fix it and add some unit tests.

grcevski (approved):
LGTM!

PR #1533, opened 2026-03-11
feat: Support MSSQL packets

### LINKED ISSUE #1285: Add support for MSSQL
HI :)
I am using OBI on my k8s cluster and my application is using Microsoft SQL Server, but the issue is that OBI did not recognize that.
The issue is that MSSQL packet looking  different than other SQL Databases because they are using UTF-16LE while PostgresQL using UTF-8.
Is there a possibility to add support for MSSQL too?

Orenico10:
This enables instrumentation of requests and queries to Microsoft SQL Server (mssql) from OBI services. Implemented similarly to how MySQL and PostgreSQL instrumentation are implemented.
Resolve #1285 

## Checklist


pkg/ebpf/common/sql_detect_mssql.go
```diff
@@ -0,0 +1,273 @@
...
+			curr := data[i+1:]
+			if len(curr) < 3 {
+				continue
+			}
+			// Ordinal (2)
+			curr = curr[2:]
+			// NameLen (1)
+			nameLen := int(curr[0])
+			curr = curr[1:]
+			if len(curr) < nameLen*2+7 {
```
NimrodAvni78: would love a comment / extract it to const to understand why

pkg/ebpf/common/sql_detect_mssql.go
```diff
@@ -0,0 +1,273 @@
...
+
+func parseHandleFromPrepareResponse(b []byte) uint32 {
+	if len(b) < kMSSQLHeaderLen {
+		return 0
+	}
+	data := b[kMSSQLHeaderLen:]
+
+	// Scan for 0xAC (RETURNVALUE)
+	for i := 0; i < len(data); i++ {
+		if data[i] == 0xAC {
```
NimrodAvni78: also extracting those to const

NimrodAvni78 (commented):
This is great, thanks!

would love to see some integration tests for this, im pretty sure there are some open source images for mssql we can use

i think after pushing a new commit tests will also start running

also adding copilot for review

bpf/generictracer/protocol_mssql.h
```diff
@@ -0,0 +1,144 @@
...
+    u8 status;
+    u16 length; // big-endian
+    u16 spid;   // big-endian
+    u8 packet_id;
+    u8 window;
+};
+
+enum {
+    // TDS header
+    k_mssql_hdr_size = 8,
```
rafaelroquetto, line 33: instead of declaring this, should we simply use `sizeof(mssql_hdr)`? This way we prevent them from drifting

pkg/ebpf/common/sql_detect_mssql.go
```diff
@@ -0,0 +1,302 @@
...
+func mssqlPreparedStatements(b *largebuf.LargeBuffer) (string, string, string) {
+	if b.Len() <= kMSSQLHeaderLen {
+		return "", "", ""
+	}
+
+	pktType, _ := b.U8At(0)
+	if pktType == kMSSQLBatch {
+		// SQL Batch
+		payload, _ := b.UnsafeViewAt(kMSSQLHeaderLen, b.Len()-kMSSQLHeaderLen)
+		stmt := ucs2ToUTF8(payload)
```
rafaelroquetto: so, here we allocate a string, just to convert it back to a byte slice. Any chance we can change `ucs2ToUTF8` to operate on a byte slice directly and decrease allocs?

pkg/ebpf/common/sql_detect_mssql.go
```diff
@@ -0,0 +1,302 @@
...
+		return span, errFallback
+	}
+
+	sqlCommand := sqlprune.SQLParseCommandID(request.DBMSSQL, reqRaw)
+	sqlError := sqlprune.SQLParseError(request.DBMSSQL, respRaw)
+
+	return TCPToSQLToSpan(event, op, table, stmt, request.DBMSSQL, sqlCommand, sqlError), nil
+}
+
+func parseMSSQLRPC(b []byte) (uint16, []byte) {
```
rafaelroquetto: perhaps it's best to pass a large buffer here directly, and use the buffer methods, or largeBuffer.NewReader() - you can then consume the data (and skip the data) on demand instead of having to materialise the entire byte slice (and as a bonus, I suspect the reader API might simplify the code a bit since you are skipping bytes and calculating offsets)

pkg/ebpf/common/sql_detect_mssql.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	// Skip options
+	if len(data) < 2 {
+		return procID, nil
+	}
+	data = data[2:]
+	return procID, data
+}
+
+func parseHandleFromExecute(data []byte) uint32 {
```
rafaelroquetto: same as above

pkg/ebpf/common/sql_detect_mssql.go
```diff
@@ -0,0 +1,302 @@
...
+		length := data[0]
+		data = data[1:]
+		if length == 4 {
+			return binary.LittleEndian.Uint32(data[:4])
+		}
+	}
+	return 0
+}
+
+func parseHandleFromPrepareResponse(b []byte) uint32 {
```
rafaelroquetto: ditto

rafaelroquetto (commented):
Good stuff! Thank you so much for spearheading this!

If possible, avoid materialising a byte slice and instead use the large buffer API directly - or, I see that some of the functions here are a perfect fit for the large buffer Reader (cursor) API - using those prevents us from having to allocate a big scratch chunk and instead will consume the bytes mostly in-place from the original backing store.

Also, this will conflict with: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1513 - since that PR is already approved, could you wait until that's in and rebase on top of it?

bpf/generictracer/protocol_mssql.h
Orenico10: Since this is the size defined in the [protocol that MSSQL uses (TDS)](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/7af53667-1b72-4703-8258-7984e838f746#:~:text=It%20is%20always%208%20bytes%20in%20length), I think it should be a constant. This is not different from how `protocol_mysql.h` is currently implemented:

```c
enum {
    // MySQL header sizes
    k_mysql_hdr_size = 5,
    k_mysql_hdr_command_id_size = 1,
    k_mysql_hdr_without_command_size = 4,

    // Command IDs
    k_mysql_com_query = 0x3,
    k_mysql_com_stmt_prepare = 0x16,
    k_mysql_com_stmt_execute = 0x17,

    // Sanity checks
    k_mysql_payload_length_max = 1 << 13, // 8K
};
```

pkg/ebpf/common/sql_detect_mssql.go
Orenico10: You are right. My original method uses `utf16.Decode`, which requires allocating a string (`utf16.Decode` returns `[]rune`, which can be cast to `string` - thus requires allocating a string to be later cast to `[]byte`). I took a look at how it is working behind the scenes and modified it to keep the data in `[]byte` to decrease the amount of allocs.

bpf/generictracer/protocol_mssql.h
```diff
@@ -0,0 +1,144 @@
...
+
+    *protocol_type = k_protocol_type_mssql;
+    bpf_map_update_elem(&protocol_cache, conn_info, protocol_type, BPF_ANY);
+
+    bpf_dbg_printk("is_mssql: mssql detected!");
+    return 1;
+}
+
+// Emit a large buffer event for MSSQL protocol.
+// The return value is used to control the flow for this specific protocol.
```
mmat11: the second and third line of this comment are not valid here

pkg/ebpf/common/sql_detect_mssql.go
```diff
@@ -0,0 +1,317 @@
...
+	// Length is big-endian in TDS. It's the length of the packet including the header.
+	length, err := b.U16BEAt(2)
+	if err != nil {
+		return false
+	}
+	// The length must be at least the header length.
+	// We also add an upper bound to avoid misinterpreting other protocols as MSSQL.
+	// The default TDS packet size is 4096, but can be negotiated up to 32767.
+	// We'll use a generous upper bound.
+	if length < uint16(kMSSQLHeaderLen) || length > 32768 {
```
mmat11: should we use the large buffer length here instead of 32768?

mmat11 (approved):
left a couple nits, lgtm!

pkg/config/ebpf_tracer.go
```diff
@@ -101,6 +101,8 @@ type EBPFTracer struct {
 	// Postgres prepared statements cache size.
 	PostgresPreparedStatementsCacheSize int `yaml:"postgres_prepared_statements_cache_size" env:"OTEL_EBPF_BPF_POSTGRES_PREPARED_STATEMENTS_CACHE_SIZE" validate:"gt=0"`
 
+	MSSQLPreparedStatementsCacheSize int `yaml:"mssql_prepared_statements_cache_size" env:"OTEL_EBPF_BPF_MSSQL_PREPARED_STATEMENTS_CACHE_SIZE" validate:"gt=0"`
```
pinoOgni, line 113: Since the configuration has changed, please regenerate the schema by running `make generate-config-schema`. This will update `docs/config-schema.json` file.

pkg/ebpf/common/sql_detect_mssql.go
Orenico10: I will move this to a constant so it is clearer why it is there. The goal is to verify whether this is MSSQL (TDS). If I check it against the buffer size, I won't be able to properly verify that this is indeed TDS.

pkg/ebpf/common/sql_detect_mssql.go
rafaelroquetto: awesome, thanks for this!

bpf/bpfcore/bpf_helpers.h
```diff
@@ -13,7 +13,7 @@
  */
 #include "bpf_helper_defs.h"
 
-#define __uint(name, val) int(*name)[val]
+#define __uint(name, val) int (*name)[val]
```
mmat11: I think this can be reverted

bpf/bpfcore/bpf_tracing.h
```diff
@@ -518,8 +518,8 @@ struct pt_regs;
                 ___bpf_ctx_cast(args));                                                            \
         _Pragma("GCC diagnostic pop")                                                              \
     }                                                                                              \
-    static                                                                                         \
-        __attribute__((always_inline)) typeof(name(0)) ____##name(unsigned long long *ctx, ##args)
+    static __attribute__((always_inline)) typeof(name(0)) ____##name(unsigned long long *ctx,      \
```
mmat11: same here

bpf/generictracer/protocol_mssql.h
```diff
@@ -0,0 +1,185 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include "logger/bpf_dbg.h"
```
mmat11:
```suggestion
#include <logger/bpf_dbg.>
```

also fix ordering

bpf/generictracer/protocol_mssql.h
```diff
@@ -0,0 +1,185 @@
...
+                                                   enum large_buf_action action) {
+    bool is_request = (direction == req->direction);
+    u32 expected_len = 0;
+
+    if (!is_request) {
+        // Response Phase
+        // We use _pad2 to store the expected length of the response
+        u32 *expected_len_ptr = (u32 *)req->_pad2;
+        if (req->resp_len > 0) {
+            action = k_large_buf_action_append;
```
mmat11: isn't action type being handled by the caller here?

bpf/generictracer/protocol_mssql.h
```diff
@@ -0,0 +1,185 @@
...
+                                                   u8 packet_type,
+                                                   u8 direction,
+                                                   enum large_buf_action action) {
+    bool is_request = (direction == req->direction);
+    u32 expected_len = 0;
+
+    if (!is_request) {
+        // Response Phase
+        // We use _pad2 to store the expected length of the response
+        u32 *expected_len_ptr = (u32 *)req->_pad2;
```
mmat11: mmm this is not clear to me, generally I don't think we should use padding fields to store data

bpf/generictracer/protocol_mssql.h
```diff
@@ -0,0 +1,185 @@
...
+        if (req->resp_len > 0) {
+            action = k_large_buf_action_append;
+        } else if (bytes_len >= k_mssql_hdr_size) {
+            struct mssql_hdr hdr = mssql_parse_hdr(u_buf);
+            *expected_len_ptr = hdr.length;
+            bpf_dbg_printk("mssql response: expected_len=%d", hdr.length);
+
+            // Copy the first part of the response for userspace parsing
+            bpf_probe_read(req->rbuf, K_TCP_RES_LEN, u_buf);
```
mmat11: I believe this shouldn't be done in a largebuffer function

bpf/generictracer/protocol_tcp.h
```diff
@@ -270,7 +274,9 @@ static __always_inline void handle_unknown_tcp_connection(pid_connection_info_t
                                existing->resp_len);
 
                 __builtin_memcpy(trace, existing, sizeof(tcp_req_t));
-                bpf_probe_read(trace->rbuf, bytes_len, u_buf);
+                if (existing->protocol_type != k_protocol_type_mssql) {
```
mmat11: why is this needed?

bpf/tests/bpfcore/bpf_helpers.h
```diff
@@ -15,7 +15,7 @@
 #define NULL ((void *)0)
 #endif
 
-#define __uint(name, val) int(*name)[val]
+#define __uint(name, val) int (*name)[val]
```
mmat11: this can also be reverted

bpf/bpfcore/bpf_helpers.h
Orenico10: This was caused by running `make clang-format`, should I still revert it?

mmat11: I'd revert it as it's not failing in main, maybe you ran a different version or rules from the one being used in CI?

Orenico10: I merged main to my branch twice, I don't think my version of the Makefile is so different from the CI's one

bpf/generictracer/protocol_tcp.h
Orenico10: When a TDS response spans multiple packets, only the first packet contains the header and error token. subsequent packets carry only payload, so without this guard the generic path would overwrite rbuf with the last payload-only chunk, making parseMSSQLError unable to find the header it expects at offset 0.

bpf/generictracer/protocol_mssql.h
Orenico10: This is really similar to 2 comments I got from Copilot so I will try to explain it here. This copies the first response packet into req->rbuf, which is what OBI's Go side reads. Only the first packet is needed since that's where the TDS header and error token are, without this, parseMSSQLError would have nothing to work with.

bpf/bpfcore/bpf_helpers.h
Orenico10: I tried reverting this but the pre-commit hook undid my changes (applied the formatting back) you can see it in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1533/commits/412161138f52bfe68631a4ce3d9ffd0d9b536050

rafaelroquetto: It's your clang-format version that is different - for now, I reckon just disable the pre-commit hook.

bpf/generictracer/protocol_mssql.h
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    *protocol_type = k_protocol_type_mssql;
+    bpf_map_update_elem(&protocol_cache, conn_info, protocol_type, BPF_ANY);
+
+    return 1;
+}
+
+// Emit a large buffer event for MSSQL protocol.
+static __always_inline int mssql_send_large_buffer(tcp_req_t *req,
```
rafaelroquetto: things have evolved a bit in regards to the large buffers, so we now have a helper you could/should use: `large_buf_emit_chunks`

bpf/generictracer/protocol_tcp.h
```diff
@@ -144,6 +145,9 @@ static __always_inline int tcp_send_large_buffer(tcp_req_t *req,
         return kafka_send_large_buffer(req, pid_conn, u_buf, bytes_len, direction, action);
     case k_protocol_type_http:
     case k_protocol_type_mqtt:
+        break;
+    case k_protocol_type_mssql:
+        return mssql_send_large_buffer(req, u_buf, bytes_len, packet_type, direction, action);
```
rafaelroquetto:
```suggestion
case k_protocol_type_mssql:
       return mssql_send_large_buffer(req, u_buf, bytes_len, packet_type, direction, action);
case k_protocol_type_http:
case k_protocol_type_mqtt:
```

bpf/bpfcore/bpf_helpers.h
Orenico10: You're right. I was using llvm@21 while the CI uses version clang-format-19. I ran it again, and it undid the changes (d283e7e6f4b62481e8a10c269d40bc75114ee9ba).

bpf/generictracer/protocol_tcp.h
rafaelroquetto: This doesn't sound right, but maybe I am missing the point here, so please bear with me: when large buffers are active, the "generic path" is not used. In ReadTCPRequestIntoSpan(), we have:

```
requestBuffer, responseBuffer := getBuffers(parseCtx, event)
```
which will return the large buffers when they are enabled (or the generic small buffers otherwise). The code shouldn't be affected by this distinction, except that perhaps without large buffers it may not have enough info to derive the protocol.

bpf/generictracer/protocol_mssql.h
Orenico10: I'm pretty sure I'm already using that :)

bpf/generictracer/protocol_mssql.h
mmat11: unresolving: action is passed as a parameter to mssql_send_large_buffer(), I believe it should be used as-is

bpf/generictracer/protocol_mssql.h
mmat11: what I was trying to say is that if large buffers are disabled, this won't get filled. We should fill non-largebuffers related fields outside of this function. To validate this, I think you could try to run the integration test after removing the mssql large buffer size in config

bpf/generictracer/protocol_mssql.h
Orenico10: you're right. there were lots of changes to `protocol_tcp.h` that i didn't notice that this works when i modified the code to work with the large buffers.

Edit: nvm, this causes it to fail the tests, I will try to debug it to see why this happens.

Orenico10: OK, I tried debugging it some more, apparently without having the action override it just doesn't have enough data for it to infer the error.

mssql large_buf: packet_type=2 action=0 resp_len=0 bytes_sent=0   <- 1st chunk: init - OK
  mssql response: partial, acc=8 exp=43                              <- partial, returns -1
  mssql large_buf: packet_type=2 action=0 resp_len=8 bytes_sent=8   <- 2nd chunk: init - not OK (it should be append)

the fix for this is either keeping the override or changing protocol_tcp.h to send the append action if `existing->lb_res_bytes` is larger than 0 (where we need to append)

mmat11: I think we should find a way to fix it in the caller (I believe protocol_tcp.h, the generic TCP code which calls this function)

Orenico10: Alr, added it

bpf/generictracer/protocol_tcp.h
```diff
@@ -266,13 +269,11 @@ static __always_inline void handle_unknown_tcp_connection(pid_connection_info_t
...
         existing->is_server = is_server;
-        if (tcp_send_large_buffer(existing,
-                                  pid_conn,
-                                  u_buf,
-                                  bytes_len,
-                                  direction,
-                                  protocol_type,
-                                  k_large_buf_action_init) < 0) {
+        const enum large_buf_action response_action =
+            (existing->lb_res_bytes > 0) ? k_large_buf_action_append : k_large_buf_action_init;
```
mmat11, line 273: just one last question: did you try adding a tcp_send_large_buffer() call to the `else` branch? I wonder why the second chunk of the packet came in this branch since the direction didn't change.. 🤔

bpf/generictracer/protocol_mssql.h
Orenico10: I think it should work now with my latest changes

bpf/generictracer/protocol_tcp.h
Orenico10: I did, it did not pick it up for some reason. I'm not entirely sure why this happens tbh

Orenico10:
I merged origin/main to my branch to deal with the conflicts. it should be good to go :)

bpf/generictracer/protocol_mssql.h
```diff
@@ -0,0 +1,208 @@
...
+                                                   u8 packet_type,
+                                                   u8 direction,
+                                                   enum large_buf_action action) {
+    if (mssql_max_captured_bytes > k_large_buf_max_mssql_captured_bytes) {
+        bpf_dbg_printk("BUG: mssql_max_captured_bytes exceeds maximum allowed value.");
+    }
+
+    if (packet_type == PACKET_TYPE_RESPONSE && req->resp_len == 0 &&
+        bytes_len >= k_mssql_hdr_size) {
+        bpf_probe_read(req->rbuf, k_mssql_hdr_size, u_buf);
```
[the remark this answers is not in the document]
Orenico10: We have been over this already, with bpf_probe_read_user, it just doesn't work. By the time the kprobe fires, the kernel has already copied that data into its own internal buffers; it's no longer a live userspace address. Trying to read it as one fails.

bpf/generictracer/protocol_mssql.h
```diff
@@ -0,0 +1,208 @@
...
+    k_mssql_msg_rpc = 0x03,
+    k_mssql_msg_response = 0x04,
+    k_mssql_msg_login7 = 0x10,
+    k_mssql_msg_prelogin = 0x12,
+};
+
+static __always_inline struct mssql_hdr mssql_parse_hdr(const unsigned char *data) {
+    struct mssql_hdr hdr = {};
+
+    bpf_probe_read(&hdr, sizeof(hdr), (const void *)data);
```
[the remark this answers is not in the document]
Orenico10: Again, mssql_parse_hdr is called from three places:
- Line 178: with u_buf, user memory
- Line 198: with req->rbuf, kernel memory (a BPF map field)
- Line 80: with data, which could be either, depending on the caller

Since the function is called with both kernel and user memory, bpf_probe_read (the generic version) is the correct choice. Switching to bpf_probe_read_user would break the call at line 198, where req->rbuf is kernel memory.

pkg/ebpf/common/sql_detect_mssql.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+
+	// Check the length:
+	// 1. MUST be at least 8 bytes (the size of the header itself).
+	// 2. MUST be less than or equal to the maximum allowable negotiated packet
+	//    size (32,767 bytes).
+	// 3. MUST fit within the captured buffer (ensures the first packet is complete).
+	// Note: While the *negotiated* packet size must be between 512 and 32,767,
+	// individual packets can be much smaller (e.g., a simple SELECT batch).
+	if length < uint16(kMSSQLHeaderLen) || length > kMSSQLMaxPacketSize || int(length) > b.Len() {
```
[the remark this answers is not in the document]
Orenico10: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1533#discussion_r3105021556
I just want to say that it told me the complete opposite on the previous review.

mmat11 (approved):
lgtm!

bpf/generictracer/protocol_mssql.h
```diff
@@ -0,0 +1,219 @@
...
+                req->lb_res_bytes += consumed_bytes;
+            }
+
+            if (consumed_bytes > 0) {
+                req->has_large_buffers = true;
+            }
+        }
+    }
+
+    if (packet_type == PACKET_TYPE_RESPONSE) {
```
rafaelroquetto: I am not sure if this is the best place for this - this seems to be detecting the end of a message, even when large buffers are not enabled for this protocol (i.e. `mssql_max_captured_byted == 0`). In this scenario, no buffer will ever be sent but yet this function is performing unrelated work for every packet.

 Maybe it would be cleaner to have someething like `mssql_respnse_complete()` instead, to be called from `handle_unknown_tcp_connection()` alongside the large buffer call, as it is somewhat unrelated.

rafaelroquetto (changes requested):
This looks great, there's just one major last change we should do - decouple the eom detection from mssql_send_large_buffer, as that's completely orthogonal and does not belong there (unless I am missing something).

Otherwise, some minor feedback:

 - missing default for `MSSQLPreparedStatementsCacheSize` - since the field has validate:"gt=0", OBI will fail validation on startup unless the user explicitly sets it
 - mssqlPreparedStatements function is misnamed: it handles SQL_BATCH plain queries, not prepared statements. RPC/prepared-statement handling is in handleMSSQL directly. Name it mssqlExtractBatchSQL or similar.
- improvements to multi-packet test: testPythonSQLMultiPacketResponse only asserts a SELECT span appeared, it doesn't verify the SQL text was correctly reassembled. The whole point of the /largeresult endpoint and bulk_actor table is to force multi-packet reassembly, but the assertion doesn't confirm it worked.
- BPF/Go detection divergence: BPF is_mssql accepts login7/prelogin packets; Go isMSSQL does not. Not a bug (login packets don't carry SQL), but undocumented and will confuse anyone trying to understand why the two functions differ.

(these bullet points were generated with the help of Claude).

The PR looks good overall, I am marking this with "request changes" just to prevent it from accidentally being merged since it's already gotten an approval.

Orenico10:
> * missing default for `MSSQLPreparedStatementsCacheSize` - since the field has validate:"gt=0", OBI will fail validation on startup unless the user explicitly sets it

Is the default config not under pkg/obi/config.go?
https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1533/changes#diff-c63675318e07cb375297629188e240c497c099f4580120308e86201f67e2934cR145
If so, then it already contains a default value :)

Orenico10:
> * improvements to multi-packet test: testPythonSQLMultiPacketResponse only asserts a SELECT span appeared, it doesn't verify the SQL text was correctly reassembled. The whole point of the /largeresult endpoint and bulk_actor table is to force multi-packet reassembly, but the assertion doesn't confirm it worked.

I'm not sure I understand what I should be testing here. OBI does not capture response content for successful queries, so the span is indistinguishable from a single packet response.

bpf/generictracer/protocol_mssql.h
Orenico10: So you want me to add something very specific to MSSQL EOM under `handle_unknown_tcp_connection()`? I feel like introducing this special handling specifically for MSSQL seems out of place.

edit: Please confirm whether the change I made matches what you meant.

rafaelroquetto (approved):

grcevski:
Thank you so much for your contribution @Orenico10 !!! :tada:

PR #1543, opened 2026-03-12
Add support for python asyncio uvloop

marctc:
This PR adds support for context propagation for python asyncio uvloop.

The implementation is built around four probe points:

- `_asyncio_Task___init__`
  * Runs when a new asyncio task is created.
  * Records the child task and its parent task.
  * Captures the request connection that should belong to that task.
  * This is the point where we build the task lineage that will later be used during parent lookup.
  
- `PyContext_CopyCurrent()`
  * Runs when Python copies the active context.
  * Binds the copied context to the task that currently owns that execution flow.
  * This is important both for normal asyncio task creation and for `asyncio.to_thread()`

- `task_step`
  * Tracks which asyncio task is currently active on the event-loop thread.

- `context_run`
  * Tracks which Python context is active on the thread.
  * This is the bridge used when there is no direct task identity on that thread.

When a Python client request needs a trace parent, the lookup works in two phases:

First, resolve the current logical task:
* try the current task recorded for the thread,
* if there is no current task, use the current active context for the thread,

Second, walk task lineage until finding the request owner:
* start from the resolved task,
* follow the stored parent chain,
* for each task, check whether it owns request connection state,
* once connection state is found, use it to resolve the server trace from `server_traces_aux`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,6 +26,7 @@ require (
 	github.com/hashicorp/golang-lru/v2 v2.0.7
 	github.com/invopop/jsonschema v0.13.0
 	github.com/klauspost/compress v1.18.4
+	github.com/lib/pq v1.10.9
 	github.com/ory/dockertest/v3 v3.12.0
 	github.com/oschwald/maxminddb-golang v1.13.1
 	github.com/prometheus/client_golang v1.23.2


bpf/common/python_task.h
```diff
@@ -0,0 +1,21 @@
...
+#include <maps/python_task_state.h>
+
+static __always_inline u64 resolve_python_context_task(const python_context_task_t *context_task) {
+    if (!context_task || !context_task->task) {
+        return 0;
+    }
+
+    const python_task_state_t *task_state =
+        (const python_task_state_t *)bpf_map_lookup_elem(&python_task_state, &context_task->task);
+    if (context_task->version && (!task_state || task_state->version != context_task->version)) {
```
grcevski: I think you can optimise this so that context_task->version is checked early in your initial if, no need to perform the map lookup if your next if fails if context_task->version is NULL.

bpf/common/trace_parent.h
```diff
@@ -129,6 +131,78 @@ static __always_inline tp_info_pid_t *find_parent_process_trace(trace_key_t *t_k
     return NULL;
 }
 
+static __always_inline u64 resolve_python_current_task(const trace_key_t *t_key, u64 pid_tgid) {
+    u64 task_id = 0;
+    const python_thread_state_t *thread_state =
+        (const python_thread_state_t *)bpf_map_lookup_elem(&python_thread_state, &pid_tgid);
+
+    if (thread_state && thread_state->current_task) {
+        task_id = thread_state->current_task;
```
grcevski: We should return the task_id right here. If I understand the code correctly, if we find it on the thread_state later the code will not reassign it if it's already set.

bpf/common/trace_parent.h
```diff
@@ -129,6 +131,78 @@ static __always_inline tp_info_pid_t *find_parent_process_trace(trace_key_t *t_k
     return NULL;
 }
 
+static __always_inline u64 resolve_python_current_task(const trace_key_t *t_key, u64 pid_tgid) {
+    u64 task_id = 0;
+    const python_thread_state_t *thread_state =
+        (const python_thread_state_t *)bpf_map_lookup_elem(&python_thread_state, &pid_tgid);
+
+    if (thread_state && thread_state->current_task) {
```
grcevski: I think you can simplify the ifs if you just do:
if (!thread_state) {
    return 0;
}

bpf/common/trace_parent.h
```diff
@@ -129,6 +131,78 @@ static __always_inline tp_info_pid_t *find_parent_process_trace(trace_key_t *t_k
...
+                               task_state->conn.port);
+                return server_tp;
+            }
+        }
+
+        u64 parent_task = 0;
+        if (task_state) {
+            parent_task = task_state->parent;
+        }
+        if (!parent_task || parent_task == task_id) {
```
grcevski: Under which condition is parent_task == task_id?

bpf/common/trace_parent.h
```diff
@@ -129,6 +131,78 @@ static __always_inline tp_info_pid_t *find_parent_process_trace(trace_key_t *t_k
...
+        bpf_dbg_printk("find_python_parent: no current task pid=%d tid=%d",
+                       t_key->p_key.pid,
+                       t_key->p_key.tid);
+        return NULL;
+    }
+
+    for (u8 i = 0; i < k_max_depth; ++i) {
+        const python_task_state_t *task_state =
+            (const python_task_state_t *)bpf_map_lookup_elem(&python_task_state, &task_id);
+        if (task_state && task_state->conn.port) {
```
grcevski: I think the code makes more sense if you simply break if (!task_state). We check it here and then the logic below is complex again with another if (task_state)

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+#include <common/connection_info.h>
+
+#include <generictracer/maps/pid_tid_to_conn.h>
+
+#include <pid/pid.h>
+
+static __always_inline u64 python_next_task_version(u64 task) {
+    const python_task_state_t *task_state =
+        (const python_task_state_t *)bpf_map_lookup_elem(&python_task_state, &task);
+    if (!task_state || !task_state->version) {
```
grcevski: I don't understand the logic of this code. It seems like we want to:
1. Return 1 if !task_state
2. Set it to 1 on task_state->version and return that
3. Bump it by one and return that.

I think this can be written much simpler.

if (!task_state) {
    return 1; //?? should it be 0 maybe
}

task_state->version = task_state->version + 1;
return task_state->version;

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+    const python_task_state_t *task_state =
+        (const python_task_state_t *)bpf_map_lookup_elem(&python_task_state, &task);
+    if (task_state) {
+        mapping.version = task_state->version;
+    }
+
+    bpf_map_update_elem(&python_context_task, &context, &mapping, BPF_ANY);
+}
+
+static __always_inline python_thread_state_t load_python_thread_state(u64 id) {
```
grcevski: This pattern is pretty inefficient. We copy the contents of thread_state every time. We should return a pointer and only initialise in the caller if the value from the map is 0. Once you pull a value from a map, you can just update it, no need to do bpf_map_update_elem again.

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+    ssl_pid_connection_info_t *info = bpf_map_lookup_elem(&pid_tid_to_conn, &id);
+    if (info) {
+        connection_info_part_t conn_part = {};
+        u32 host_pid = pid_from_pid_tgid(id);
+        populate_ephemeral_info(
+            &conn_part, &info->p_conn.conn, info->orig_dport, host_pid, FD_SERVER);
+        if (parent_task) {
+            const python_task_state_t *parent_state =
+                (const python_task_state_t *)bpf_map_lookup_elem(&python_task_state, &parent_task);
+            if (parent_state && parent_state->conn.port &&
```
grcevski: Ugh, I'm not sure how this works. This seems scary. How can the parent port be same as the the child port? Isn't a parent something like a server request, meaning the emphemeral port will be the port of the client connecting to the server, while the child is an outgoing connection which has an ephemeral port to yet a third service?

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+                parent_state->conn.port != conn_part.port) {
+                // Under concurrent gather workloads, pid_tid_to_conn can point
+                // at a different in-flight request on the same thread. If the
+                // parent already has a different request connection, prefer the
+                // parent so the child stays in the same request lineage.
+                conn_part = parent_state->conn;
```
grcevski: The comment here makes me think this isn't a reliable solution. I don't think we can rely on the connection.

pkg/ebpf/tracer_linux.go
```diff
@@ -386,6 +386,7 @@ func (pt *ProcessTracer) NewExecutable(exe *link.Executable, ie *Instrumentable)
 
 	for _, p := range pt.Programs {
 		p.RegisterOffsets(ie.FileInfo, ie.Offsets)
+		p.ProcessBinary(ie.FileInfo)
```
grcevski: Why did we need this?

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -210,7 +214,60 @@ func (p *Tracer) constants() map[string]any {
...
+	rest := path[start:]
+	end := strings.Index(rest, "/lib-dynload/_asyncio")
+	if end < 0 {
+		return ""
+	}
+
+	return rest[:end+len("/lib-dynload/_asyncio")]
+}
+
+func pythonAsyncioTaskStepStart(module string, legacyProbe, defaultProbe *ebpf.Program) *ebpf.Program {
```
grcevski: I think there's a better way of doing this. The way this is implemented we'll have individual probe for every python binary, which will accumulate on our daemonset environments.

We already have a pattern that works. The Go tracer has per inode number offsets map, which is a way to communicate for the eBPF side per binary file context information, including library version changes. We should generalise that approach and make it work for the generic tracer.

Technically all you need is per inode map that says which is the python version.

I'm also not sure if this is a reliable way to tell the python version. What if the user renamed the binary to `python`?

I'd check how the Otel eBPF Profiler finds the python version. If they do it like this, then perhaps it's good, but they may have more advanced ways of doing this.

I think we should scope the legacy out of this PR to make things simpler and then follow-up with PR to add the legacy support.

Another, perhaps a much simpler way is if we can peek in the eBPF side at the REG1 and see if we can tell by what the memory looks like if it's state or not. There might be some pattern, e.g. the first word contains flags and the next looks like int etc.

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -412,7 +484,34 @@ func (p *Tracer) UProbes() map[string]map[string][]*ebpfcommon.ProbeDesc {
 				Start:    p.bpfObjects.ObiRbObjCallInitKw,
 			}},
 		},
+		"libpython3.": {
+			"context_run": {{
+				Required: false,
+				Start:    p.bpfObjects.ObiUprobeContextRun,
+			}},
+			"context_run.lto_priv.0": {{ // In Python 3.14, context_run has different symbols due to Link Time Optimization
```
grcevski, line 421: Hm, this kind of symbol is complier dependent. I think it will depend on the toolchain used to build python and can easily break. Is there no other symbol we can pick up that context_run calls?

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+
+    bpf_dbg_printk("task_step: tid=%d task=%llx", id, task);
+    python_thread_state_t thread_state = load_python_thread_state(id);
+    thread_state.current_task = task;
+    bpf_map_update_elem(&python_thread_state, &id, &thread_state, BPF_ANY);
+    return 0;
+}
+
+SEC("uprobe/_asyncio.so:task_step_legacy")
+int obi_uprobe_task_step_legacy(struct pt_regs *ctx) {
```
mmat11, line 68: can we use a global variable to set the index of the task parameter instead of having 2 probes?

bpf/maps/python_context_task.h
```diff
@@ -0,0 +1,21 @@
...
+typedef struct python_context_task {
+    u64 task;    // task that owned this PyContext* when the mapping was written
+    u64 version; // task version captured at bind time to reject TaskObj* reuse
+} python_context_task_t;
+
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, u64); // PyContext*
+    __type(value, python_context_task_t);
+    __uint(max_entries, MAX_CONCURRENT_SHARED_REQUESTS);
```
mmat11, line 21: should this also be pinned internally?

bpf/maps/python_thread_state.h
```diff
@@ -0,0 +1,22 @@
...
+    u64 current_task;
+    u64 current_context;
+    u64 inflight_task;
+} python_thread_state_t;
+
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, u64); // pid:tgid
+    __type(value, python_thread_state_t);
+    __uint(max_entries, MAX_CONCURRENT_SHARED_REQUESTS);
```
mmat11, line 22: should this also be pinned internally?

devdocs/python-asyncio-context-propagation.md
```diff
@@ -0,0 +1,222 @@
+# Python asyncio and uvloop Context Propagation
```
mariomac, line 1: Do we know the range of compatibility with Python versions?

internal/test/vm/Dockerfile
```diff
@@ -90,6 +90,9 @@ echo "Docker daemon is ready"
 # job timeout. Combined with 'panic' boot param and QEMU -no-reboot,
 # this causes QEMU to exit after the first stall.
 sysctl -w kernel.panic_on_rcu_stall=1 2>/dev/null || true
+# Increase per-socket option memory so larger socket filters still attach
+# on older kernels used by the VM integration jobs.
+sysctl -w net.core.optmem_max=262144 2>/dev/null || true
```
mariomac: Is this something we should document for our users, or it's just a patch for testing?

internal/test/integration/traces_test.go
```diff
@@ -1514,3 +1525,123 @@ func testHTTPTracesNestedNodeJSLargeHTTPS(t *testing.T) {
 	// We must see two children
 	require.Len(t, children, 2)
 }
+
+func testPythonAsyncEndpoint(t *testing.T, endpoint string, expectedClientCalls int) {
+	waitForTestComponentsSub(t, "http://localhost:8391", "/health")
+
+	const requests = 200
+	for i := 1; i <= requests; i++ {
```
mmat11, line 1541: why do we need to do 200 requests with different slugs? there's an "Eventually" in the loop below so this might take a lot of time

internal/test/vm/Dockerfile
mmat11: I don't think we should increase the value here, we had some production cases where the limit was hit even with a full green test suite; I hit this in my branch as well, I think we should add a tail call to the socket filter

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -413,7 +484,34 @@ func (p *Tracer) UProbes() map[string]map[string][]*ebpfcommon.ProbeDesc {
...
+				Required: false,
+				Start:    p.bpfObjects.ObiUprobeContextRun,
+			}},
+			"PyContext_CopyCurrent": {{
+				Required: false,
+				End:      p.bpfObjects.ObiUprobeCopyContext,
+			}},
+			"context_new_from_vars": {{ // In Docker, PyContext_CopyCurrent has Tail Recursion Optimization, so we need this function instead
```
mmat11, line 429: can we add a check on the ELF symbols to verify either one of this probes will be loaded and return an error if no symbols or both symbols are available? (if there are two, we can also log a warn and load one of the two)

bpf/common/trace_parent.h
```diff
@@ -129,6 +131,78 @@ static __always_inline tp_info_pid_t *find_parent_process_trace(trace_key_t *t_k
...
+                task_id = resolved_task;
+                bpf_dbg_printk("find_python_parent: context fallback tid=%d ctx=%llx task=%llx",
+                               t_key->p_key.tid,
+                               thread_state->current_context,
+                               task_id);
+            }
+        }
+    }
+
+    bpf_dbg_printk("find_py_parent: resolved tid=%d task=%llx", t_key->p_key.tid, task_id);
```
pinoOgni: nit: here and in some other `bpf_dbg_printk` calls, the function name (e.g., "find_python_parent" or "find_py_parent", "task_step_ret" and so on) is included in the string. These can be removed, as the function name is now added automatically.

internal/test/integration/traces_test.go
marctc: To ensure the validity of the implementation I had to run this with many requests, as race conditions start to happen with around 1k requests. I can for sure lower it here for the CI.

internal/test/vm/Dockerfile
marctc: I added this because integration tests for old kernels failed because `net.core.optmem_max` was too low. However @mmat11 suggests to do in another way.

devdocs/python-asyncio-context-propagation.md
marctc: I had it documented in `devdocs/features.md` but somehow I lost it. I will change and also here

bpf/generictracer/python.c
marctc: It won't work in the case we instrument several pythons which different version each other.

bpf/maps/python_context_task.h
marctc: We could but I don't think its necessary

bpf/common/python_task.h
```diff
@@ -0,0 +1,21 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <maps/python_context_task.h>
```
rafaelroquetto, line 8: you are missing `<vmlinux>` here - we shouldn't rely on indirect includes

bpf/common/trace_parent.h
```diff
@@ -129,6 +131,78 @@ static __always_inline tp_info_pid_t *find_parent_process_trace(trace_key_t *t_k
...
+    return task_id;
+}
+
+static __always_inline tp_info_pid_t *find_python_parent_trace(const trace_key_t *t_key,
+                                                               u64 pid_tgid) {
+    enum { k_max_depth = 4 };
```
rafaelroquetto, line 170: how was this value calculated? I wonder if it is a reasonable value: in my experiments with NodeJS, which does something very similar to this, the depth was much deeper than 4 layers. This may of course not be the case with Python, but I am just wondering

bpf/common/trace_parent.h
```diff
@@ -129,6 +131,78 @@ static __always_inline tp_info_pid_t *find_parent_process_trace(trace_key_t *t_k
...
+                               task_id,
+                               task_state->conn.port);
+                return server_tp;
+            }
+        }
+
+        u64 parent_task = 0;
+        if (task_state) {
+            parent_task = task_state->parent;
+        }
```
rafaelroquetto: I think these lines can be simplified to:

```
if (!task_state || task_state->parent == task_id) {
   break;
}

task_id = task_state->parent;
```

bpf/generictracer/generictracer.c
```diff
@@ -9,5 +9,6 @@
 #include "nodejs.c"
 #include "java_tls.c"
 #include "ruby.c"
+#include "python.c"
```
rafaelroquetto, line 11: nit: alphabetic order

bpf/generictracer/python.c
rafaelroquetto: this can be made even simpler

```
return task_state ? ++task_state->version : 0;
```

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+    u64 id = bpf_get_current_pid_tgid();
```
rafaelroquetto:
```suggestion
    const u64 id = bpf_get_current_pid_tgid();
```

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+
+    if (!valid_pid(id)) {
+        return 0;
+    }
+
+    return update_current_task(id, (u64)PT_REGS_PARM1(ctx));
```
rafaelroquetto:
```suggestion
    const u64 task = (u64) PT_REGS_PARAM1(ctx);
    return update_current_task(id, task);
```

rationale: tells the reader what param 1 holds (task) - otherwise one has to look into `update_current_task` signature.

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+}
+
+SEC("uprobe/_asyncio.so:task_step")
+int obi_uprobe_task_step(struct pt_regs *ctx) {
+    u64 id = bpf_get_current_pid_tgid();
```
rafaelroquetto:
```suggestion
    const u64 id = bpf_get_current_pid_tgid();
```

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+
+    if (!valid_pid(id)) {
+        return 0;
+    }
+
+    return update_current_task(id, (u64)PT_REGS_PARM2(ctx));
```
rafaelroquetto: same as above

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+}
+
+SEC("uprobe/_asyncio.so:task_step_ret")
+int obi_uprobe_task_step_ret(struct pt_regs *ctx) {
+    (void)ctx;
+    u64 id = bpf_get_current_pid_tgid();
```
rafaelroquetto:
```suggestion
    const u64 id = bpf_get_current_pid_tgid();
```

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+
+    if (!valid_pid(id)) {
+        return 0;
+    }
+
+    bpf_dbg_printk("task_step_ret: clearing tid=%d", id);
+    python_thread_state_t thread_state = load_python_thread_state(id);
+    thread_state.current_task = 0;
```
rafaelroquetto: I've come across `0` a few times and I suspect it may have a meaning, but I am unsure - are you using `0` as a sentinel value or does `0` really mean the very first task?

If it's a sentinel value, we should probably define a constant, such as `k_current_task_none` or `k_current_task_uninitialised`

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+        return 0;
+    }
+
+    bpf_map_update_elem(&python_thread_state, &id, &thread_state, BPF_ANY);
+    return 0;
+}
+
+SEC("uprobe/libpython3.:context_run")
+int obi_uprobe_context_run(struct pt_regs *ctx) {
+    u64 id = bpf_get_current_pid_tgid();
```
rafaelroquetto:
```suggestion
    const u64 id = bpf_get_current_pid_tgid();
```

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+    if (!valid_pid(id)) {
+        return 0;
+    }
+
+    u64 context = (u64)PT_REGS_PARM1(ctx);
```
rafaelroquetto:
```suggestion
    const u64 context = (u64)PT_REGS_PARM1(ctx);
```

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+    python_thread_state_t thread_state = load_python_thread_state(id);
+    thread_state.current_context = context;
+    bpf_map_update_elem(&python_thread_state, &id, &thread_state, BPF_ANY);
+
+    return 0;
+}
+
+// PyContext_CopyCurrent is called in two key places:
+//   1. Inside _asyncio_Task___init___impl when context=Py_None (task creation)
+//   2. In asyncio.to_thread via contextvars.copy_context() (thread dispatch)
```
rafaelroquetto, line 142: I'd remove this comment - it's meaningless in the current context - or I'd rephrase it to mention what the purpose of PyContext_CopyCurrent is rather than where it's called ,although I personally find the function name (PyContext_CopyCurrent) pretty self-explanatory: it copies the current context (if that's not the case then a proper comment does come in handy)

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+SEC("uprobe/libpython3.:PyContext_CopyCurrent")
+int obi_uprobe_copy_context(struct pt_regs *ctx) {
+    u64 id = bpf_get_current_pid_tgid();
```
rafaelroquetto:
```suggestion
    const u64 id = bpf_get_current_pid_tgid();
```

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+
+    if (!valid_pid(id)) {
+        return 0;
+    }
+
+    u64 context = (u64)PT_REGS_RC(ctx);
```
rafaelroquetto:
```suggestion
    const u64 context = (u64)PT_REGS_RC(ctx);
```

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+    if (!context) {
+        return 0;
+    }
+
+    // Task initialization copies the new context before the child task shows up
+    // in task_step, so the inflight child is the only safe owner here.
+    const python_thread_state_t *thread_state =
+        (const python_thread_state_t *)bpf_map_lookup_elem(&python_thread_state, &id);
+    if (thread_state && thread_state->inflight_task) {
```
rafaelroquetto: you are checking thread_state twice here  and on line 166

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+        bpf_dbg_printk(
+            "copy_context: mapped context=%llx task=%llx", context, thread_state->current_task);
+        return 0;
+    }
+    return 0;
+}
+
+SEC("uprobe/_asyncio.so:_asyncio_Task___init__")
+int obi_uprobe_task_init(struct pt_regs *ctx) {
+    u64 id = bpf_get_current_pid_tgid();
```
rafaelroquetto:
```suggestion
    const u64 id = bpf_get_current_pid_tgid();
```

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+
+    if (!valid_pid(id)) {
+        return 0;
+    }
+
+    u64 child_task = (u64)PT_REGS_PARM1(ctx);
```
rafaelroquetto:
```suggestion
    const u64 child_task = (u64)PT_REGS_PARM1(ctx);
```

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+    };
+
+    bpf_dbg_printk("task_init: parent_lookup tid=%d parent=%llx", id, parent_task);
+    bpf_dbg_printk("task_init: tid=%d child=%llx parent=%llx", id, child_task, parent_task);
+
+    // Read the current request connection from this thread now. After the child
+    // task starts switching contexts, this thread-local value may no longer
+    // belong to the same request.
+    ssl_pid_connection_info_t *info = bpf_map_lookup_elem(&pid_tid_to_conn, &id);
+    if (info) {
```
rafaelroquetto: can you replace this with an early return?

```
if (!info) {
  // cleanup and return
}
```

rationale: reduces indentation/improves clarity

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+        .version = python_next_task_version(child_task),
+    };
+
+    bpf_dbg_printk("task_init: parent_lookup tid=%d parent=%llx", id, parent_task);
+    bpf_dbg_printk("task_init: tid=%d child=%llx parent=%llx", id, child_task, parent_task);
+
+    // Read the current request connection from this thread now. After the child
+    // task starts switching contexts, this thread-local value may no longer
+    // belong to the same request.
+    ssl_pid_connection_info_t *info = bpf_map_lookup_elem(&pid_tid_to_conn, &id);
```
rafaelroquetto: can this be const?

bpf/generictracer/python.c
```diff
@@ -0,0 +1,256 @@
...
+    if (info) {
+        connection_info_part_t conn_part = {};
+        u32 host_pid = pid_from_pid_tgid(id);
```
rafaelroquetto:
```suggestion
        const u32 host_pid = pid_from_pid_tgid(id);
```

bpf/generictracer/python.c
rafaelroquetto: Piggy-backing on Nikola's comments, why are these heuristics required? And what happens when they fail? Can they not be true?

bpf/maps/python_context_task.h
rafaelroquetto: We should probably pin in anyway - when code gets refactored, and other tracers end up using maps, it's easy to miss and forget this, and then weird bugs come up. Perhaps only SCRATCH_MEM maps should not be pinned, and I am not even sure about that.

internal/test/integration/traces_test.go
marctc: reduced to 20

bpf/generictracer/python.c
marctc: Discussed, clarified and fix in main PRs comments section.

bpf/generictracer/python.c
marctc: Initially i implemented this using as default behaviour of fetching conn info for each child task creation. Once I started to increase the concurrency, race conditions started to happen so this fix the issue (use parent's connection). Discussing with Nikola we agreed that this should be the default behaviour.

bpf/common/trace_parent.h
marctc: This value was not calculated per-se as there's no Python runtime limit for that. I started with a higher number and i lowered to 4 as tests passed. A depth >4 would be rather unusual for python programs.

bpf/generictracer/python.c
marctc: I'd rather leave it as reader can understand the modes where this code is executed. But if you really against it, I'm open to any other suggestions!

internal/test/integration/traces_test.go
```diff
@@ -1524,3 +1525,123 @@ func testHTTPTracesNestedNodeJSLargeHTTPS(t *testing.T) {
...
+		go ti.DoHTTPGet(t, "http://localhost:8391"+endpoint+strconv.Itoa(i), 200)
```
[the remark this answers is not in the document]
marctc: we need to test concurrency

pkg/internal/ebpf/generictracer/generictracer.go
marctc: I checked and `context_run()` does not call the public `PyContext_Enter()` / `PyContext_Exit()` wrappers. It calls the internal `_PyContext_Enter()` / `_PyContext_Exit()` helpers directly (but those are also LTO).

What about change how symbols are selected and have something like `re:^context_run(?:\\..+)?$` for this case? If we think its a good idea i could to in a follow up PR to reduce the scope of this

pkg/internal/ebpf/generictracer/generictracer.go
marctc: Makes sense, but the PR is getting big. Do you think we can do in a follow up PR or should be part of this one?

mmat11: follow up is also fine 👍🏻

internal/test/vm/Dockerfile
marctc: fixed here https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1543/commits/a63cb6e8f01a4ad5025372f19d263062418520c9

PR #1550, opened 2026-03-12
collector: fix cross-platform compilation

MrAlias:
`go build ./collector/...` failed on any non-Linux or non-amd64/arm64 target (Windows, macOS, `linux/ppc64le`, `linux/s390x`, etc.) with errors like:

```
# go.opentelemetry.io/obi/pkg/internal/ebpf/tcmanager
pkg/internal/ebpf/tcmanager/tcmanager.go:90:17: undefined: link.AttachTCX
# go.opentelemetry.io/obi/pkg/internal/rdns/ebpf/xdp
pkg/internal/rdns/ebpf/xdp/dnstrtacer.go:85:21: undefined: link.AttachXDP
# go.opentelemetry.io/obi/pkg/ebpf/common
pkg/ebpf/common/common.go:42:23: undefined: BpfHttpRequestTraceT
...
```

This blocked inclusion of the `obi` receiver in any collector distribution (e.g. `otelcol-contrib`) because the OCB-generated `components.go` imports `go.opentelemetry.io/obi/collector` unconditionally, causing cross-compile jobs for Windows/macOS/non-x86 Linux to fail.

Two root causes:

1. **`collector/internal/controller.go` had no build tag.**

   Running `go build ./collector/...` recursively compiled the `collector/internal` sub-package, pulling in `pkg/obi`, `pkg/instrumenter`, and the generated eBPF types — none of which compile outside `linux/amd64` or `linux/arm64`.

2. **The `//go:build !linux` tag on `factory_others.go` was too broad.**

   OBI's generated eBPF artifacts only exist for `amd64` and `arm64`; `linux/ppc64le` and other Linux architectures also lack them and need the stub path.

### `collector/internal/controller.go`

Added `//go:build linux && (amd64 || arm64)` so the eBPF-dependent controller is only compiled on the two supported architectures.

### `collector/internal/controller_others.go` *(new)*

Empty stub package file with `//go:build !(linux && (amd64 || arm64))` to satisfy the Go toolchain's requirement that a package is non-empty on every build target.

### `collector/factory_linux.go`

Narrowed build tag from `//go:build linux` to `//go:build linux && (amd64 || arm64)`.

### `collector/factory_others.go`

- Narrowed build tag from `//go:build !linux` to `//go:build !(linux && (amd64 || arm64))`.
- Moved `defaultConfig()` here (was incorrectly in `factory.go` alongside a platform-specific `pkg/obi` import).
- Removed the now-duplicate `errUnsupportedPlatform` var (moved to `factory.go`).

### `collector/factory.go`

- Removed the `defaultConfig()` function and its `pkg/obi` / `consumertest` imports (those are now in the platform-specific files).
- Added `errUnsupportedPlatform` so it is always compiled and accessible to both the non-Linux stub and `factory_test.go`.

### `collector/factory_test.go`

Added `TestNewFactoryUnsupportedPlatform`: skips on `linux/amd64` and `linux/arm64`, and on all other platforms asserts that both `BuildTracesReceiver` and `BuildMetricsReceiver` return `errUnsupportedPlatform`. This follows the `journaldreceiver` pattern: the factory is present and registered but returns a clear, testable error on startup.

### `.github/workflows/pull_request.yml`

Added `collector-cross-compile` job with a 9-platform matrix that runs on every PR, verifying `go build ./collector/...` succeeds for:

| Platform | Behaviour |
|---|---|
| `linux/amd64` | Full eBPF (supported) |
| `linux/arm64` | Full eBPF (supported) |
| `linux/ppc64le` | Stub — `errNotLinux` at runtime |
| `linux/s390x` | Stub — `errNotLinux` at runtime |
| `linux/386` | Stub — `errNotLinux` at runtime |
| `windows/amd64` | Stub — `errNotLinux` at runtime |
| `windows/arm64` | Stub — `errNotLinux` at runtime |
| `darwin/amd64` | Stub — `errNotLinux` at runtime |
| `darwin/arm64` | Stub — `errNotLinux` at runtime |

## Behaviour on unsupported platforms

`NewFactory()` returns a real `receiver.Factory` on all platforms — the `obi` type is registered in the component registry and appears in config validation. Receivers fail to start with a clear error:

```
OBI receiver is only supported on Linux amd64/arm64
```

This is identical to the approach used by `journaldreceiver` in `opentelemetry-collector-contrib` and is the pattern recommended in [opentelemetry-collector-contrib#46192](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46192).

## Testing

```
# Linux/amd64 (supported — full tests pass)
go test ./collector/...

# Windows cross-compile (test binary compiles, TestNewFactoryUnsupportedPlatform would run)
GOOS=windows GOARCH=amd64 go test -c -o /dev/null ./collector/...

# All nine platforms build cleanly
GOOS=linux   GOARCH=amd64   go build ./collector/...  # EXIT:0
GOOS=linux   GOARCH=arm64   go build ./collector/...  # EXIT:0
GOOS=linux   GOARCH=ppc64le go build ./collector/...  # EXIT:0
GOOS=windows GOARCH=amd64   go build ./collector/...  # EXIT:0
GOOS=darwin  GOARCH=arm64   go build ./collector/...  # EXIT:0
```

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -174,7 +174,7 @@ jobs:
           # Supported platforms (full OBI eBPF support)
           - { goos: linux,   goarch: amd64  }
           - { goos: linux,   goarch: arm64  }
-          # Unsupported platforms (factory stubs — must compile, return errNotLinux at runtime)
+          # Unsupported platforms (factory stubs — must compile, return errUnsupportedPlatform at runtime)
           - { goos: linux,   goarch: ppc64le }
           - { goos: linux,   goarch: s390x   }
           - { goos: linux,   goarch: "386"   }


.github/workflows/pull_request.yml
```diff
@@ -161,3 +161,54 @@ jobs:
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # Supported platforms (full OBI eBPF support)
+          - { goos: linux,   goarch: amd64  }
+          - { goos: linux,   goarch: arm64  }
+          # Unsupported platforms (factory stubs — must compile, return errNotLinux at runtime)
```
mmat11:
```suggestion
          # Unsupported platforms (factory stubs — must compile, return errUnsupportedPlatform at runtime)
```

PR #1553, opened 2026-03-13
nodejs detector: fallback to scan code for SIGUSR1 when traditional signal check fails

NimrodAvni78:
## Description
in continuation to [this comment](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1514#issuecomment-4045044941)

This is a backup logic to check for SIGUSR1 handling in the case that we can't extract it by lookin at the process memory (in case of stripped binary with dynamic libuv for example).
I've extracted some logic from the js route harvester into common js code scanning logic, and reused it to find references to SIGUSR1 constant in the

## Test cases

|                                                                | Agent injected? | Explination                                                                                                            
|:---------------------------------------------------------------|:---------------:|------------------------------------------------------------------------------------------------------------------------
| Has SIGUSR1 handler + Offical node image                       |       No        | After succesfully reading process memory and finding SIGUSR1 handler in libuv, classified as not safe to inject                                          
| Has SIGUSR1 handler +  Ubuntu with node any dynamic libuv      |       No        | After failing to read libuv offset, found SIGUSR1 constant in js code, classified as not safe to inject                                           
| Does not SIGUSR1 handler +  Offical node image                 |       Yes       | After succesfully reading process memory and not finding SIGUSR1 handler in libuv, classifed as safe to inject         
| Does not SIGUSR1 handler +  Ubuntu with node any dynamic libuv |       Yes       | After failing to read libuv offset, search for SIGUSR1 constant in js code and not found, classified as safe to inject

### Caveats
False positive detections may occur in cases like:
```js
const myVar = "SIGUSR1" // not used
setTimeout(() => {}, 600000);
```
where this application won't get the agent injected

false negatives are rarer but still may occur with something like:

```js
process.on("SIGUSR" + "1", () => process.exit());
```

but if someone runs this in production i think their application deserves to be restarted (only half kidding, they can still turn off js injection entirely with `OTEL_EBPF_NODEJS_ENABLED=false`)
## Checklist

### DIFF
--- a/pkg/internal/nodejs/source_scan.go
+++ b/pkg/internal/nodejs/source_scan.go
@@ -13,26 +13,40 @@ import (
 	"go.opentelemetry.io/obi/pkg/internal/transform/route/harvest"
 )
 
-// sigusr1Quoted contains the patterns we search for in source files to detect
-// SIGUSR1 handler registration. We look for the string SIGUSR1 wrapped in
-// quotes (single, double, or backtick) to avoid matching comments or logs.
 var sigusr1Quoted = []string{`"SIGUSR1"`, `'SIGUSR1'`, "`SIGUSR1`"}
 
 // sourceHasSIGUSR1Reference scans the Node.js application's source files for
 // references to "SIGUSR1", 'SIGUSR1', or `SIGUSR1`. This is a fallback
 // detection method used when the symbol-based detection fails (e.g. stripped
-// binaries with dynamic libuv). It reuses the same directory discovery and
-// file walking logic as the Node.js route harvester.
-//
-// Returns true if any source file contains a quoted SIGUSR1 reference.
-// Returns false on any error or if no reference is found.
+// binaries with dynamic libuv).
 func sourceHasSIGUSR1Reference(pid int) bool {
-	dir, err := harvest.FindAppDir(app.PID(pid))
+	dir, err := harvest.FindNodeJSAppDir(app.PID(pid))
 	if err != nil {
 		return false
 	}
 
 	return dirHasSIGUSR1Reference(dir)
+}
+
+func lineContainsSIGUSR1(line string) bool {
+	for _, pattern := range sigusr1Quoted {
+		if strings.Contains(line, pattern) {
+			return true
+		}
+	}
+	return false
+}
+
+func scanFileForSIGUSR1(path string) bool {
+	found := false
+	_ = harvest.ScanJSFileLines(path, func(line string) bool {
+		if lineContainsSIGUSR1(line) {
+			found = true
+			return true
+		}
+		return false
+	})
+	return found
 }
 
 // dirHasSIGUSR1Reference scans JS/TS source files in the given directory for
@@ -41,16 +55,8 @@ func dirHasSIGUSR1Reference(dir string) bool {
 	found := false
 
 	_ = harvest.WalkJSFiles(dir, func(path string) error {
-		_ = harvest.ScanJSFileLines(path, func(line string) bool {
-			for _, pattern := range sigusr1Quoted {
-				if strings.Contains(line, pattern) {
-					found = true
-					return true
-				}
-			}
-			return false
-		})
-		if found {
+		if scanFileForSIGUSR1(path) {
+			found = true
 			return filepath.SkipAll
 		}
 		return nil
--- a/pkg/internal/transform/route/harvest/js.go
+++ b/pkg/internal/transform/route/harvest/js.go
@@ -351,11 +351,26 @@ func ScanJSFileLines(path string, fn func(line string) bool) error {
 	}
 	defer file.Close()
 
+	inBlockComment := false
 	scanner := bufio.NewScanner(file)
 	for scanner.Scan() {
 		line := strings.TrimSpace(scanner.Text())
 
-		if line == "" || strings.HasPrefix(line, "//") || strings.HasPrefix(line, "/*") {
+		if inBlockComment {
+			if strings.Contains(line, "*/") {
+				inBlockComment = false
+			}
+			continue
+		}
+
+		if line == "" || strings.HasPrefix(line, "//") {
+			continue
+		}
+
+		if strings.HasPrefix(line, "/*") {
+			if !strings.Contains(line, "*/") {
+				inBlockComment = true
+			}
 			continue
 		}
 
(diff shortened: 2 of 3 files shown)


pkg/internal/nodejs/nodejs.go
```diff
@@ -75,10 +75,23 @@ func (i *NodeInjector) injectFile(pid int, elfFile *elf.File) error {
...
+		switch hasUserSIGUSR1Handler(pid, elfFile) {
+		case signalCheckFound:
+			i.log.Warn("Node.js process has a custom SIGUSR1 handler, skipping agent injection. "+
+				"Node.js trace correlation will not work", "pid", pid)
+			return nil
+		case signalCheckFailed:
+			// Symbol-based detection failed (e.g. stripped binary with dynamic libuv).
+			// Fall back to scanning the application's source files for quoted SIGUSR1 references.
+			if sourceHasSIGUSR1Reference(pid) {
+				i.log.Warn("Node.js source files reference SIGUSR1, skipping agent injection. "+
```
mmat11, line 88: can this be a false positive? and if yes, should we have some config flag to force attaching it?

NimrodAvni78: yes it can, if for some reason there is SIGUSR1 const and it is not a signal handler
i can add a config flag, something like `OTEL_EBPF_NODEJS_HURISTIC_DETECT=true/false` ?

pkg/internal/nodejs/source_scan.go
```diff
@@ -0,0 +1,60 @@
...
+	"path/filepath"
+	"strings"
+
+	"go.opentelemetry.io/obi/pkg/appolly/app"
+	"go.opentelemetry.io/obi/pkg/internal/transform/route/harvest"
+)
+
+// sigusr1Quoted contains the patterns we search for in source files to detect
+// SIGUSR1 handler registration. We look for the string SIGUSR1 wrapped in
+// quotes (single, double, or backtick) to avoid matching comments or logs.
```
rafaelroquetto: nit: I'd delete these comments

pkg/internal/nodejs/source_scan.go
```diff
@@ -0,0 +1,60 @@
...
+var sigusr1Quoted = []string{`"SIGUSR1"`, `'SIGUSR1'`, "`SIGUSR1`"}
+
+// sourceHasSIGUSR1Reference scans the Node.js application's source files for
+// references to "SIGUSR1", 'SIGUSR1', or `SIGUSR1`. This is a fallback
+// detection method used when the symbol-based detection fails (e.g. stripped
+// binaries with dynamic libuv). It reuses the same directory discovery and
+// file walking logic as the Node.js route harvester.
+//
+// Returns true if any source file contains a quoted SIGUSR1 reference.
+// Returns false on any error or if no reference is found.
```
rafaelroquetto:
```suggestion
// sourceHasSIGUSR1Reference scans the Node.js application's source files for
// references to "SIGUSR1", 'SIGUSR1', or `SIGUSR1`. This is a fallback
// detection method used when the symbol-based detection fails (e.g. stripped
// binaries with dynamic libuv).
```

pkg/internal/nodejs/nodejs.go
mmat11: mmm that would enable/disable it for all JS apps on that node. I'm thinking we might want to keep it enabled by default, but disable it only for certain apps, not sure if we should just have a bool or a whole selector, WDYT?

pkg/internal/nodejs/source_scan.go
```diff
@@ -0,0 +1,60 @@
...
+
+	return dirHasSIGUSR1Reference(dir)
+}
+
+// dirHasSIGUSR1Reference scans JS/TS source files in the given directory for
+// quoted SIGUSR1 references.
+func dirHasSIGUSR1Reference(dir string) bool {
+	found := false
+
+	_ = harvest.WalkJSFiles(dir, func(path string) error {
```
rafaelroquetto, line 57: nit: can we define these lambdas out of line (e.g. above this function) to reduce code indentation and improve clarity? I personally find this harder to review with all these different `return` statements returning different types (boolean, error, all chained)

pkg/internal/transform/route/harvest/js.go
```diff
@@ -341,6 +341,32 @@ func (e *RouteExtractor) extractNextJSRoutesFromManifest(dir string) error {
...
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+
+		if line == "" || strings.HasPrefix(line, "//") || strings.HasPrefix(line, "/*") {
+			continue
+		}
```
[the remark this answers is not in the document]
rafaelroquetto: Fair point here - can we address that?

pkg/internal/transform/route/harvest/js.go
```diff
@@ -574,25 +576,68 @@ var (
 	cwdForPID     = ebpfcommon.CWDForPID
 )
 
-func ExtractNodejsRoutes(pid app.PID) (*RouteHarvesterResult, error) {
+// FindAppDir locates the root directory of a Node.js application by reading
+// its command line and working directory from /proc. This is used by both
+// the route harvester and the SIGUSR1 source scanner.
+func FindAppDir(pid app.PID) (string, error) {
```
rafaelroquetto:
```suggestion
func FindNodeJSAppDir(pid app.PID) (string, error) {
```

pkg/internal/transform/route/harvest/js.go
```diff
@@ -574,25 +576,68 @@ var (
...
+	}
+	return dir, nil
+}
+
+// WalkJSFiles walks a directory tree, skipping known non-application directories
+// (node_modules, .git, system dirs, etc.), and calls fn for each JS/TS source
+// file found (.js, .ts, .mjs, .cjs). The callback can return filepath.SkipAll
+// to stop the walk early.
+func WalkJSFiles(root string, fn func(path string) error) error {
+	return filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
```
rafaelroquetto: optional nit: define lambda out of line

pkg/internal/nodejs/nodejs.go
mmat11: after some thought, I think there's no perfect way to detect it with 100% precision so we can handle these cases in the future

PR #1554, opened 2026-03-13
Fix hostname capturing for lib/pq versions > 1.11

### LINKED ISSUE #1535: Blocked upgrade: github.com/lib/pq v1.11.2
Upgrading the Go SQL test component from github.com/lib/pq v1.10.9 to v1.11.2 breaks the PostgreSQL OATS case on branch renovate/go.

The failing case is internal/test/oats/sql/yaml/oats_sql_go_postgres.yaml.

Identified in #1496 

## Failing symptom

The failing assertion is the Prometheus selector that expects the SQL hostname label:

- run-oats_sql_go_postgres
- expected metric selector: db_client_operation_duration_seconds_sum{db_operation_name="SELECT", db_system_name="other_sql", server_address="sqlserver"}

The metric family is present, but no series is emitted with server_address="sqlserver".

## Root cause

OBI currently hooks the lib/pq hostname helper through the symbol github.com/lib/pq.network.

The lib/pq upgrade changes that hook target:

- github.com/lib/pq v1.10.9 exposes github.com/lib/pq.network
- github.com/lib/pq v1.11.2 exposes github.com/lib/pq.Config.network

Because gotracer.go still instruments github.com/lib/pq.network, hostname extraction for lib/pq stops working on v1.11.2 and the resulting SQL spans/metrics lose server.address.

## Evidence

Binary-level check from the same gosql test app:

- with github.com/lib/pq v1.10.9: go tool nm shows github.com/lib/pq.network
- with github.com/lib/pq v1.11.2: go tool nm shows github.com/lib/pq.Config.network

Focused SQL OATS rerun with the downgraded gosql component showed that OBI once again detected and instrumented github.com/lib/pq.network in the test binary.

Relevant autoinstrumenter log lines from the focused rerun:

- library detected: github.com/lib/pq version=v1.10.9
- found relevant function for instrumentation: github.com/lib/pq.network
- going to instrument function: github.com/lib/pq.network

## Local verification notes

The local focused SQL OATS rerun on this machine still timed out on the same server_address selector even after the lib/pq downgrade, so the downgrade did not produce a fully green local OATS result here.

However, the downgrade did restore the old hook target in the rebuilt gosql binary, which confirms the package upgrade is a real blocker for the current probe implementation.

Verification commands used:

- cd internal/test/oats/sql && TESTCASE_TIMEOUT=5m TESTCASE_BASE_PATH=./yaml .tools/ginkgo -v -r --focus='run-oats_sql_go_postgres'
- go tool nm on a gosql binary built with v1.10.9 and with v1.11.2

## Proposed unblock

Teach OBI to handle the newer lib/pq symbol layout.

Possible directions:

- instrument github.com/lib/pq.Config.network in addition to github.com/lib/pq.network
- add version-aware symbol selection for lib/pq hooks
- add an integration test that asserts server.address survives across both symbol layouts

Until that probe update lands, the gosql test component should stay pinned to github.com/lib/pq v1.10.9.

marctc:
New version of lib/pq changes the way the hostname is fetched. Now it's part of the config struct
and the approach is closer than how MySQL works. I keept the probe that hooks in `network()` function
to keep backwards compatible.

Fixes #1535

### DIFF
--- a/bpf/gotracer/go_sql.c
+++ b/bpf/gotracer/go_sql.c
@@ -212,6 +212,12 @@ read_pq_hostname_from_pqconn(void *pq_conn_ptr, char *hostname, u64 max_len) {
     return hostname[0] != '\0';
 }
 
+static __always_inline bool supports_pq_conn_cfg_hostname() {
+    off_table_t *ot = get_offsets_table();
+
+    return go_offset_of(ot, (go_offset){.v = _pq_one_eleven_zero}) != 0;
+}
+
 // SQL hostname extraction with driver type routing.
 // Uses conn_type to determine which driver-specific extraction to use or
 // attempts to extract hostname by trying supported database drivers
@@ -236,10 +242,12 @@ static __always_inline void extract_sql_hostname(sql_request_trace_t *trace,
 
     void *pq_conn_ptr = get_pq_conn_ptr(driver_conn_ptr);
     if (pq_conn_ptr) {
-        if (read_pq_hostname_from_pqconn(
-                pq_conn_ptr, (char *)trace->hostname, sizeof(trace->hostname))) {
-            bpf_dbg_printk("extracted lib/pq hostname from conn.cfg: %s", trace->hostname);
-            return;
+        if (supports_pq_conn_cfg_hostname()) {
+            if (read_pq_hostname_from_pqconn(
+                    pq_conn_ptr, (char *)trace->hostname, sizeof(trace->hostname))) {
+                bpf_dbg_printk("extracted lib/pq hostname from conn.cfg: %s", trace->hostname);
+                return;
+            }
         }
 
         // lib/pq < v1.11 does not expose the selected host on conn, so keep the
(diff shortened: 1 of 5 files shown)


bpf/gotracer/go_sql.c
```diff
@@ -207,14 +234,36 @@ static __always_inline void extract_sql_hostname(sql_request_trace_t *trace,
         return;
     }
 
-    void *mysql_conn_ptr = get_mysql_conn_ptr(driver_conn_ptr);
-    if (!mysql_conn_ptr) {
+    void *pq_conn_ptr = get_pq_conn_ptr(driver_conn_ptr);
+    if (pq_conn_ptr) {
+        if (read_pq_hostname_from_pqconn(
```
grcevski: Don't we need to add another field with offset to say which version of the postgres library we have? Like the versioning checks we do for grpc?

I'm afraid that this may read garbage on older versions and we'll successfully return.

grcevski: Like these:
```
		// grpc versioning
		goexec.GrpcOneSixZero,
		goexec.GrpcOneSixNine,
		goexec.GrpcOneSevenSeven,
		// HTTP2 versioning
		goexec.HTTP2ZeroFortyFive,

```

marctc: Ah good idea, let's do that

Issue #1153, opened 2026-01-21
Extend network metrics


MrAlias:
Support exporting [similar metrics from the `opentelemetry-network` project](https://github.com/open-telemetry/opentelemetry-network/blob/main/docs/metrics/metrics.yaml).

svencowart:
@NimrodAvni78 @MrAlias I'm curious how much of the old `opentelemetry-network` project is now reproduced within OBI? If we have feature parity, I'd like to move towards deprecating the `opentelemetry-network` project. In it's current state, it has no active maintainers and therefore is stale. I see the following benefits to the open-telemetry ecosystem by deprecating `opentelemetry-network`:

1. Solves the confusing caused by having two projects that solve the same problem. We can provide a clear answer to everyone that they should use OBI and not `opentelemetry-network` for eBPF-based network observability.
2. Opens the project "domain", to be repurposed by the [repurposed network SIG](https://github.com/open-telemetry/community/pull/3560).

I know the decision to deprecate `opentelemetry-network` won’t be made by this SIG, but at least understanding if OBI can already replace it's functionality would enable use to start making that move.

PR #1565, opened 2026-03-16
refactor: move IP discovery logic to `pkg/netip` for reuse 

pinoOgni:
In preparation for #1153.
This PR simply moves IP discovery logic to `pkg/netip` for reuse.


pkg/internal/netip/ip.go
```diff
@@ -39,39 +37,39 @@ var (
 	}
 )
 
-// fetchAgentIP guesses the non-loopback IP address of the Agent host, according to the
+// FetchAgentIP guesses the non-loopback IP address of the Agent host, according to the
 // user-provided configuration:
 //   - If OBIIP is provided, this value is used whatever is the real IP of the Agent.
```
mmat11: I think comment should be updated to reflect the new parameter names

pinoOgni: oh, nice catch!

PR #1567, opened 2026-03-16
Added obi_bpf_network_ignored_packet_total metric

mariomac:
We observed that, in big hosts, the default OBI configuration can miss up to 80% of the network flows traffic due to collisions in the eBPF aggregated_flows hashmap. Increasing the size in `OTEL_EBPF_NETWORK_CACHE_MAX_FLOWS` allows more accurate flow data at the cost of more CPU and memory.

In case you are fine getting incomplete network flows but want to get a rough estimation of how much traffic you are missing, this PR provides two metrics that allow calculating a factor of "ignored packets":
```
obi_bpf_network_packets_total
obi_bpf_network_ignored_packets_total
```

### DIFF
--- a/bpf/netolly/flows_common.h
+++ b/bpf/netolly/flows_common.h
@@ -60,7 +60,7 @@ typedef struct packet_count_t {
 } packet_count;
 
 // Accounts the proportion of packets not reaching the userspace due to an error
-// in the allocation/update of direct_flows or aggregatted_flows
+// in the allocation/update of direct_flows or aggregated_flows
 struct {
     __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
     __type(key, u32);


bpf/netolly/flows_common.h
```diff
@@ -54,6 +54,20 @@ struct {
     __type(value, flow_metrics);
 } aggregated_flows SEC(".maps");
 
+typedef struct packet_count_t {
+    u64 total;
+    u64 ignored;
```
mmat11, line 59: nit: should we name this "dropped"? as far as I could see these are errored updates because of ringbuf or map update failures, right?

bpf/netolly/flows.c
```diff
@@ -169,6 +169,14 @@ static inline int flow_monitor(struct __sk_buff *skb) {
 
     const u64 current_time = bpf_ktime_get_ns();
 
+    const u32 key = 0;
+    packet_count safe = {0, 0};
+    packet_count *packet_stats = (packet_count *)bpf_map_lookup_elem(&flow_packet_stats, &key);
+    if (packet_stats == NULL) {
```
mmat11: is this required by the verifier? array maps should be preallocated and initialized to zero

pkg/internal/netolly/ebpf/sock_tracer.go
mmat11: should we use `errors.Join` instead of this?

pkg/internal/netolly/ebpf/tracer.go
```diff
@@ -178,7 +178,10 @@ func (m *FlowFetcher) Close() error {
...
 
 	if m.objects != nil {
-		errs = append(errs, m.closeObjects()...)
+		if err := m.objects.Close(); err != nil {
+			errs = append(errs, err)
+		}
+		m.objects = nil
 	}
 
 	var errStrings []string
```
mmat11: same here, errors.Join?

pkg/internal/netolly/flow/packet_stats.go
```diff
@@ -0,0 +1,68 @@
...
+func StartInternalMetrics(bpfPacketStats *ebpf.Map) (PacketStats, error) {
+	possibleCPUs, err := ebpf.PossibleCPU()
+	if err != nil {
+		_ = bpfPacketStats.Close()
+		return PacketStats{}, err
+	}
+
+	// initialize all the counters to 0
+	if err := bpfPacketStats.Put(uint32(0), make([]ebpf2.NetPacketCount, possibleCPUs)); err != nil {
+		return PacketStats{}, err
```
mmat11: should we close `bpfPacketStats` in case of error?

mmat11: also, I don't think this is needed, the map should be initialized to zero values

bpf/netolly/flows_common.h
mariomac: I named it "dropped" initially but later I prefered to rename it as "ignored" to avoid misleading users and thinking that the packages are actually dropped from the network stack.

bpf/netolly/flows.c
mariomac: Ah! I was not able to find this part in the eBPF spec so I took the defensive path.
This makes code cleaner. Thanks!

bpf/netolly/flows_common.h
mmat11: sounds good 👍🏻

bpf/netolly/flows_common.h
```diff
@@ -54,6 +54,20 @@ struct {
...
+} packet_count;
+
+// Accounts the proportion of packets not reaching the userspace due to an error
+// in the allocation/update of direct_flows or aggregatted_flows
```
mmat11:
```suggestion
// in the allocation/update of direct_flows or aggregated_flows
```

PR #1571, opened 2026-03-16
Fix missing requests on pipe splice

grcevski:
Proxies can implement buffer forwarding using pipes, system call `do_splice()` and `splice_to_socket()` which will skip tcp_recvmsg and may use kernel buffers for tcp_sendmsg. This causes us to miss tracking certain requests, but also see partial requests that are force closed (e.g. response code 499).

It can easily be reproduced by running external requests on a docker compose file and making the external and internal docker ports the same, using a port tracking option. Docker-proxy does this.

We used to have support for this kind of issue through buffer capture with the socket filter (nothing escapes the socket filter), but our initial implementation was very expensive. We removed it (a customer reported issue) and at that time I couldn't remember what it was for, I never added a test when I implemented the first version.

This time I've taken a different approach, we do heavy filtering on the sock filter on which buffers to capture. I've left comments in the code on how this works.

```
    // We check here for problematic buffer captures
    // There are two situatios:
    //   1. The sendmsg couldn't capture the buffer, we need to do it for them.
    //      This is the lookup by connection
    //   2. We have problematic receive port recorded. These are receive buffers we
    //      couldn't read, but since the socket filter runs before the receive probe
    //      we rely on a prior port connection recorded by tcp_close which saw an
    //      incomplete request.
```

TODO:
- [x] Add integration tests with a mock proxy. I'm working on something in C that will do this low level buffer forwarding. I've tested and now this correctly works with docker-proxy, all the way to the external curl call.
- [x] Benchmark the base case against the new branch on requests that don't need this new path.

### DIFF
--- a/bpf/generictracer/k_tracer.c
+++ b/bpf/generictracer/k_tracer.c
@@ -3,7 +3,6 @@
 
 //go:build obi_bpf_ignore
 
-#include "common/http_types.h"
 #include <bpfcore/vmlinux.h>
 #include <bpfcore/bpf_helpers.h>
 #include <bpfcore/bpf_tracing.h>
@@ -11,6 +10,7 @@
 #include <common/backup_buffer.h>
 #include <common/common.h>
 #include <common/connection_info.h>
+#include <common/http_types.h>
 #include <common/iov_iter.h>
 #include <common/msg_buffer.h>
 #include <common/protocol_defs.h>
@@ -1125,6 +1125,93 @@ int obi_socket_flt_buf(struct __sk_buff *skb) {
         return 0;
     }
 
+    // Save the original destination port before sorting. For incoming connections this is the
+    // local server port.
+    const u16 orig_dport = t_ctx->conn.d_port;
+
+    //d_print_http_connection_info(&conn);
+
+    sort_connection_info(&t_ctx->conn);
+
+    // Check if this is a connection we should be looking at.
+    // For outgoing connections, connect() populates connection_tracker before any data flows.
+    // For incoming connections, the socket filter fires before accept(), so connection_tracker
+    // won't have the entry yet. We fall back to checking filter_ports (listening port, no netns
+    // because we can't find it here in the socket filter).
+    //
+    // By definition we'll miss the first accept connection buffers.
+    tracked_connection_t *t_conn = bpf_map_lookup_elem(&connection_tracker, &t_ctx->conn);
+    if (!t_conn) {
+        bool *fp = bpf_map_lookup_elem(&filter_ports, &orig_dport);
+        if (!fp) {
+            // We finally check if we've missed the accept, but we have asked for backup buffer
+            // in tcp_sendmsg
+            backup_buffer_t *back_buf = bpf_map_lookup_elem(&sock_filter_buffers, &t_ctx->conn);
+            if (!back_buf) {
+                return 0;
+            }
+        }
+    }
+
+    // we don't want to read the whole buffer for every packed that passes our checks, we read only a bit and check if it's truly HTTP request/response.
+    unsigned char buf[MIN_HTTP_SIZE] = {0};
+    bpf_skb_load_bytes(skb, t_ctx->tcp.hdr_len, (void *)buf, sizeof(buf));
+    // technically the read should be reversed, but eBPF verifier complains on read with variable length
+    u32 len = skb->len - t_ctx->tcp.hdr_len;
+    if (len > MIN_HTTP_SIZE) {
+        len = MIN_HTTP_SIZE;
+    }
+
+    u8 packet_type = 0;
+    if (is_http(
+            buf,
+            len,
+            &packet_type)) { // we must check tcp_close second, a packet can be a close and a response
+        // this can be very verbose
+        //bpf_d_printk("http buf=[%s] [%s]", buf, __FUNCTION__);
+        //d_print_http_connection_info(&conn);
+        if (packet_type == PACKET_TYPE_REQUEST) {
+            const u64 cookie = bpf_get_socket_cookie(skb);
+            //bpf_dbg_printk("cookie=%llx, len=%d, buf=[%s]", cookie, len, buf);
+            //dbg_print_http_connection_info(&conn);
+
+            // The code below is looking to see if we have recorded black-box trace info on
+            // another interface. We do this for client calls, where essentially the original
+            // request may go out on one interface, but then get re-routed to another, which is
+            // common with some k8s environments.
+            partial_connection_info_t partial = {
+                .d_port = t_ctx->conn.d_port,
+                .s_port = t_ctx->conn.s_port,
+                .tcp_seq = t_ctx->tcp.seq,
+            };
+            __builtin_memcpy(partial.s_addr, t_ctx->conn.s_addr, sizeof(partial.s_addr));
+
+            tp_info_pid_t *trace_info = trace_info_for_connection(&t_ctx->conn, TRACE_TYPE_CLIENT);
+            if (trace_info) {
+                if (cookie) { // we have an actual socket associated
+                    bpf_map_update_elem(&tcp_connection_map, &partial, &t_ctx->conn, BPF_ANY);
+                }
+            } else if (!cookie) { // no actual socket for this skb, relayed to another interface
+                connection_info_t *prev_conn = bpf_map_lookup_elem(&tcp_connection_map, &partial);
+
+                if (prev_conn) {
+                    tp_info_pid_t *trace_info =
+                        trace_info_for_connection(prev_conn, TRACE_TYPE_CLIENT);
+                    if (trace_info) {
+                        if (current_immediate_epoch(trace_info->tp.ts) ==
+                            current_immediate_epoch(bpf_ktime_get_ns())) {
+                            //bpf_dbg_printk("Found trace info on another interface, setting it up for this connection");
+                            tp_info_pid_t other_info = {0};
+                            __builtin_memcpy(&other_info, trace_info, sizeof(tp_info_pid_t));
+                            set_trace_info_for_connection(
+                                &t_ctx->conn, TRACE_TYPE_CLIENT, &other_info);
+                        }
+                    }
+                }
+            }
+        }
+    }
+
     // We check here for problematic buffer captures
     // There are two situations:
     //   1. The sendmsg couldn't capture the buffer, we need to do it for them.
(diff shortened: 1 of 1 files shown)


bpf/common/backup_buffer.h
```diff
@@ -0,0 +1,13 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <bpfcore/vmlinux.h>
+
+enum { k_backup_buffer_len = 256 };
+
+typedef struct backup_buffer {
+    u8 buf[k_backup_buffer_len];
```
rafaelroquetto: nit:

```suggestion
    unsigned char buf[k_backup_buffer_len];
```

in the past (last year) I did a refactor in which I standardised every buffer / byte view to `unsigned char` for various reasons (strict aliasing being one of them)

bpf/generictracer/maps/backup_buffer_mem.h
```diff
@@ -0,0 +1,21 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_helpers.h>
+
+#include <common/backup_buffer.h>
+
+struct {
```
rafaelroquetto: we should use `SCRATCH_MEM_TYPED(backup_buffer, backup_buffer_t)` instead of having this header.

bpf/generictracer/maps/sock_filter_buffers.h
```diff
@@ -0,0 +1,18 @@
...
+
+#include <common/backup_buffer.h>
+#include <common/connection_info.h>
+#include <common/map_sizing.h>
+
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __uint(max_entries, MAX_CONCURRENT_REQUESTS);
+    __type(key, connection_info_t);
+    __type(value, backup_buffer_t);
```
rafaelroquetto, line 18:
```suggestion
    __type(value, backup_buffer_t);
    __uint(pinning, OBI_PIN_INTERNAL);
```

bpf/generictracer/k_send_receive.h
```diff
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include "logger/bpf_dbg.h"
```
rafaelroquetto: vscode vscoding

bpf/generictracer/k_send_receive.h
```diff
@@ -49,14 +50,39 @@ static __always_inline void ensure_sent_event(u64 id, u64 *sock_p, u8 direction)
     }
 }
 
-static __always_inline void force_sent_event(u64 id, u64 *sock_p) {
+// Unreadable buffers on responses may cause unexpected force closed connections, e.g. 499. We track
+// those in unreadable buffer ports, but they are setup on demand in tcp_cleanup_rbuf which is too
+// late for the first request. This code ensures we don't cause 499 on the first one, but a silently
+// missed event.
+static __always_inline u8 should_ignore_unreadable(pid_connection_info_t *p_conn,
```
rafaelroquetto:
```suggestion
static __always_inline bool should_ignore_unreadable(pid_connection_info_t *p_conn,
```

bpf/generictracer/k_send_receive.h
```diff
@@ -49,14 +50,39 @@ static __always_inline void ensure_sent_event(u64 id, u64 *sock_p, u8 direction)
...
+                                                   const bool *unreadable) {
```
rafaelroquetto:
```suggestion
                                                   bool unreadable) {
```

I do understanding the semantics are "if the pointer is there, then it is unreadable", i.e. the mere pointer existence conveys the information, rather than the boolean value, but that confused the hell out of me until I read the entire implementation, as it's not obvious at first sight. A plain bool with a helper (see comments below) is probably best here.

bpf/generictracer/k_tracer.c
```diff
@@ -3,10 +3,12 @@
 
 //go:build obi_bpf_ignore
 
+#include "common/send_args.h"
```
rafaelroquetto: vscode

bpf/generictracer/k_tracer.c
```diff
@@ -18,14 +20,17 @@
...
 #include <common/tcp_info.h>
+#include <common/tracked_connection.h>
 
 #include <generictracer/dns.h>
 #include <generictracer/k_send_receive.h>
 #include <generictracer/k_tracer_defs.h>
 #include <generictracer/k_unix_sock.h>
 #include <generictracer/maps/active_accept_args.h>
 #include <generictracer/maps/active_connect_args.h>
+#include <generictracer/maps/backup_buffer_mem.h>
```
rafaelroquetto: use SCRATCH_MEM

bpf/generictracer/k_tracer.c
```diff
@@ -904,10 +967,38 @@ static __always_inline int return_recvmsg(void *ctx, struct sock *in_sock, u64 i
...
+
+            backup_buffer_t *backup = bpf_map_lookup_elem(&sock_filter_buffers, &info.conn);
+            if (backup) {
+                buf = backup->buf;
+                bpf_dbg_printk("found backup buf=%llx", buf);
+                // delete right away to avoid duplicate responses. If a sendmsg needs this buffer
+                // they've already set it up, since sendmsg buffers are captured between the probe/retprobe.
+                bpf_map_delete_elem(&sock_filter_buffers, &info.conn);
+            } else {
+                bool unreadable = true;
```
rafaelroquetto:
```suggestion
                const bool unreadable = true;
```

bpf/generictracer/k_send_receive.h
```diff
@@ -49,14 +50,39 @@ static __always_inline void ensure_sent_event(u64 id, u64 *sock_p, u8 direction)
...
+    }
+    http_info_t *info = bpf_map_lookup_elem(&ongoing_http, p_conn);
+    if (!info) {
+        return false;
+    }
+
+    return !http_info_complete(info);
+}
+
+static __always_inline void force_sent_event(u64 id, u64 *sock_p, bool *unreadable) {
```
rafaelroquetto:
```suggestion
static __always_inline void force_sent_event(u64 id, u64 *sock_p, bool unreadable) {
```

bpf/generictracer/k_tracer.c
```diff
@@ -641,15 +693,26 @@ int BPF_KPROBE(obi_kprobe_tcp_close, struct sock *sk, long timeout) {
...
+    bool *unreadable = 0;
+    if (success) {
+        if (info.conn.d_port) {
+            unreadable = bpf_map_lookup_elem(&unreadable_buffer_ports, &info.conn.d_port);
+        }
+
+        if (!unreadable && info.conn.s_port) {
+            unreadable = bpf_map_lookup_elem(&unreadable_buffer_ports, &info.conn.s_port);
+        }
+    }
```
rafaelroquetto: Here's where the helper comes into play:
```
static __always_inline bool is_port_unreadable(u16 port) {
    if (port == 0) {
        return false;
    }

    const bool *unreadable = bpf_map_lookup_elem(&unreadable_buffer_ports, &port);

   return unreadable && *unreadable;
}

static __always_inline bool is_conn_unreadable(const connection_info_t *conn) {
  return is_port_unreadable(conn->d_port) || is_port_unreadable(conn->s_port);
}

// usage
const bool unreadable = is_conn_unreadable(&info.conn);
```

bpf/generictracer/k_tracer.c
```diff
@@ -904,10 +967,38 @@ static __always_inline int return_recvmsg(void *ctx, struct sock *in_sock, u64 i
 
     if (parse_sock_info((struct sock *)sock_ptr, &info.conn)) {
         const u16 orig_dport = info.conn.d_port;
+        const u16 orig_sport = info.conn.s_port;
         d_print_http_connection_info(&info.conn);
         sort_connection_info(&info.conn);
         info.pid = pid_from_pid_tgid(id);
 
+        if (buf == 0) {
```
rafaelroquetto:
```suggestion
        if (buf == NULL) {
```

or


```suggestion
        if (!buf) {
```

bpf/generictracer/k_tracer.c
```diff
@@ -904,10 +967,38 @@ static __always_inline int return_recvmsg(void *ctx, struct sock *in_sock, u64 i
...
+                bpf_map_delete_elem(&sock_filter_buffers, &info.conn);
+            } else {
+                bool unreadable = true;
+                // We have anunreadable connection, we mark both ports as unreadable.
+                // Tecnically we have information in tracked connection if this was connect or accept,
+                // however if OBI tracks both processes, we'll see connect and accept on the same pair
+                // and last one wins.
+                bpf_dbg_printk("setting unreadable buffer ports=%d,%d", orig_dport, orig_sport);
+                bpf_map_update_elem(&unreadable_buffer_ports, &orig_dport, &unreadable, BPF_ANY);
+                bpf_map_update_elem(&unreadable_buffer_ports, &orig_sport, &unreadable, BPF_ANY);
```
rafaelroquetto: likewise here I'd put this into a helper such as:

```
static __always_inline void mark_port_unreadable(u16 port) {
    if (port > 0) {
        bpf_map_update_elem(&unreadable_buffer_ports, &port, &(bool){true}, BPF_ANY);
    }
}

```


then


```suggestion
                bpf_dbg_printk("setting unreadable buffer ports=%u,%u", orig_dport, orig_sport);
                mark_port_unreadable(orid_dport);
                mark_port_unreadable(orig_sport);
```

bpf/generictracer/k_tracer.c
```diff
@@ -1052,6 +1164,48 @@ int obi_socket__http_filter(struct __sk_buff *skb) {
         }
     }
 
+    // We check here for problematic buffer captures
+    // There are two situatios:
```
rafaelroquetto:
```suggestion
    // There are two situations:
```

bpf/generictracer/k_tracer.c
```diff
@@ -1052,6 +1164,48 @@ int obi_socket__http_filter(struct __sk_buff *skb) {
...
+        if (back_buf->tcp_seq == tcp.seq) {
+            return 0;
+        }
+        back_buf->tcp_seq = tcp.seq;
+        capture_buf = back_buf->buf;
+    } else { // Scenario 2.
+        bool *unreadable = bpf_map_lookup_elem(&unreadable_buffer_ports, &conn.d_port);
+        if (!unreadable) {
+            unreadable = bpf_map_lookup_elem(&unreadable_buffer_ports, &conn.s_port);
+        }
```
rafaelroquetto: use the helper

bpf/generictracer/k_tracer.c
```diff
@@ -1052,6 +1164,48 @@ int obi_socket__http_filter(struct __sk_buff *skb) {
...
+
+        if (unreadable) {
+            backup_buffer_t *bbuf = backup_buf_memory();
+            if (bbuf) {
+                bbuf->tcp_seq = tcp.seq;
+                bpf_map_update_elem(&sock_filter_buffers, &conn, &bbuf, BPF_ANY);
+                back_buf = bpf_map_lookup_elem(&sock_filter_buffers, &conn);
+
+                capture_buf = back_buf->buf;
+            }
```
rafaelroquetto: this would also look better / decluttered if you moved it to its own function so you can call it as:

```
if (unreadable) {
    capture_buf = new_capture_buf(); // does the map insert and lookup and stuff
}
```

bpf/generictracer/protocol_http.h
```diff
@@ -3,19 +3,22 @@
 
 #pragma once
 
+#include "common/event_defs.h"
```
rafaelroquetto: vscode

bpf/maps/filter_ports.h
```diff
@@ -0,0 +1,18 @@
...
+
+#include <common/map_sizing.h>
+
+// Port-only (no netns) version of listening_ports, for use in socket filter
+// programs where the network namespace is not reliably accessible from __sk_buff.
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, u16);
+    __type(value, bool);
+    __uint(max_entries, MAX_CONCURRENT_REQUESTS);
```
rafaelroquetto, line 18: might wanna PIN_INTERNAL here as well (we never know when other traces like tpinjector or gotracer may need this)

bpf/maps/unreadable_buffer_ports.h
```diff
@@ -0,0 +1,18 @@
...
+#include <bpfcore/bpf_helpers.h>
+
+#include <common/map_sizing.h>
+
+// Port-only (no netns) version of listening_ports, for use in socket filter
+// programs where the network namespace is not reliably accessible from __sk_buff.
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, u16);
+    __type(value, bool);
```
rafaelroquetto, line 17: same

bpf/generictracer/k_tracer.c
grcevski: which helper?

bpf/generictracer/k_tracer.c
```diff
@@ -3,10 +3,12 @@
 
 //go:build obi_bpf_ignore
 
+#include "common/http_types.h"
```
rafaelroquetto: vscode again

PR #1572, opened 2026-03-17
Introduce StatsO11y 

### LINKED ISSUE #1153: Extend network metrics

pinoOgni:
This PR introduces **StatsO11y**, a new component designed to add custom ebpf probes for calculating node-wide statistical metrics (e.g., TCP RTT).

In brief:
- events of interest are captured and sent to user space via the `stat_events` ring buffer.
- events are converted to `ebpf.Stat` and routed through a dedicated processing pipeline.
- implemented TCP RTT using `tcp_close` hook.
- included documentation on how to implement and integrate additional metrics.

To maintain focus on the core StatsO11y implementation, the following improvements are planned for immediate follow-up (post-merge TODOs):
- explore merging common fields with `NetworkConfig` and aligning the Agent service status between StatsO11y and NetO11y.
- define default buckets for statistical metrics once a broader set of metrics is available.
- update NetO11y to include source/destination ports in `CommonAttrs` to match the StatsO11y implementation.

Example of tcp rtt metric:
```
obi_stat_tcp_rtt_seconds_bucket{dst_address="10.100.x.x",dst_name="quote",dst_port="8080",dst_zone="",k8s_cluster_name="",k8s_dst_name="quote",k8s_dst_namespace="default",k8s_dst_node_ip="",k8s_dst_node_name="",k8s_dst_owner_name="quote",k8s_dst_owner_type="Service",k8s_dst_type="Service",k8s_src_name="shipping-76f697f685-2wqwc",k8s_src_namespace="default",k8s_src_node_ip="192.168.x.x",k8s_src_node_name="i-0xxxxxxxxxxxxx",k8s_src_owner_name="shipping",k8s_src_owner_type="Deployment",k8s_src_type="Pod",obi_ip="192.168.x.x",src_address="192.168.x.x",src_name="shipping-76f697f685-2wqwc",src_port="39658",src_zone="us-west-2",le="0.01"} 1
```

resolves #1153 (even if the name of the issue is about network)
old closed PR #1273

### DIFF
--- a/bpf/statsolly/k_tcp.c
+++ b/bpf/statsolly/k_tcp.c
@@ -63,7 +63,7 @@ int BPF_KPROBE(obi_kprobe_tcp_close_srtt, struct sock *sk) {
     se->srtt_us = srtt_us;
     se->conn = conn;
 
-    bpf_d_printk("s_port=%d, d_port=%d, srtt_us=%d", se->conn.s_port, se->conn.d_port, se->srtt_us);
+    bpf_d_printk("s_port=%d, d_port=%d, srtt_us=%u", se->conn.s_port, se->conn.d_port, se->srtt_us);
     bpf_ringbuf_submit(se, stats_events_flags());
 
     return 0;
--- a/pkg/internal/statsolly/ebpf/stats_tracer.go
+++ b/pkg/internal/statsolly/ebpf/stats_tracer.go
@@ -53,16 +53,6 @@ func NewStatsFetcher() (*StatsFetcher, error) {
 		MaxEntries: uint32(os.Getpagesize()),
 	}
 
-	// msg_buffer_mem map is unsupported due to pinning
-	const kMsgBufferSizeMax = 4096
-
-	spec.Maps["msg_buffer_mem"] = &ebpf.MapSpec{
-		Type:       ebpf.PerCPUArray,
-		KeySize:    4,
-		ValueSize:  uint32(kMsgBufferSizeMax),
-		MaxEntries: 1,
-	}
-
 	if err := spec.LoadAndAssign(&objects, nil); err != nil {
 		return nil, fmt.Errorf("loading and assigning BPF objects: %w", err)
 	}
--- a/pkg/statsolly/agent/agent.go
+++ b/pkg/statsolly/agent/agent.go
@@ -80,7 +80,6 @@ type ebpFetcher interface {
 	StatsEventsMap() *ciliumebpf.Map
 }
 
-// StatsAgent instantiates a new agent, given a configuration.
 func StatsAgent(ctxInfo *global.ContextInfo, cfg *obi.Config) (*Stats, error) {
 	alog := alog()
 	alog.Info("initializing Stats agent")


bpf/statsolly/maps/stats_events.h
```diff
@@ -0,0 +1,17 @@
...
+#include <bpfcore/bpf_helpers.h>
+
+struct {
+    __uint(type, BPF_MAP_TYPE_RINGBUF);
+    __uint(max_entries, 1 << 16);
+} stats_events SEC(".maps");
+
+static __always_inline long stats_events_flags() {
+    const u64 avail_data = bpf_ringbuf_query(&stats_events, BPF_RB_AVAIL_DATA);
+    return avail_data >= 4096 ? BPF_RB_FORCE_WAKEUP : BPF_RB_NO_WAKEUP;
```
mariomac, line 16: Optional suggestion (can be addressed in the future, in another PR). Could we make this `4096` configurable? Even if 4096 is a reasonable size, in some big nodes it could still be increased, especially when we start to add more and more event types.

bpf/statsolly/maps/stats_events.h
```diff
@@ -0,0 +1,17 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_helpers.h>
+
+struct {
+    __uint(type, BPF_MAP_TYPE_RINGBUF);
+    __uint(max_entries, 1 << 16);
```
rafaelroquetto, line 11: does it make sense to pin_internal?

bpf/statsolly/k_tcp.c
```diff
@@ -0,0 +1,72 @@
...
+    u8 flags; // Must be first, we use it to tell what kind of event we have on the ring buffer
+    u8 _pad[3];
+    u32 srtt_us;
+    connection_info_t conn;
+} tcp_rtt_t;
+
+// Force tcp_rtt_t
+const tcp_rtt_t *unused_tcp_rtt __attribute__((unused));
+
+SEC("kprobe/tcp_close")
```
rafaelroquetto, line 33: just a random thought, since this is new code - should we attempt to use tracing programs instead of kprobes, as they have less overhead? Might matter of something in the tcp_close codepath, and easy to replace the kprobe with

See https://github.com/rafaelroquetto/opentelemetry-ebpf-instrumentation/blob/22ff5f85a6ee9f06e33e0d096185b72d9cb9fe8e/bpf/tpinjector/ingress.c#L331 for an example

bpf/statsolly/k_tcp.c
```diff
@@ -0,0 +1,72 @@
...
+    tcp_rtt_t *se = bpf_ringbuf_reserve(&stats_events, sizeof(*se), 0);
+    if (!se) {
+        return 0;
+    }
+
+    se->flags = k_event_stat_tcp_rtt;
+    se->srtt_us = srtt_us;
+    se->conn = conn;
+
+    bpf_d_printk("s_port=%d, d_port=%d, srtt_us=%d", se->conn.s_port, se->conn.d_port, se->srtt_us);
```
rafaelroquetto: nit: should be `%u` instead of `%d`

pkg/internal/statsolly/ebpf/stats_tracer.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	objects := StatsObjects{}
+	spec, err := LoadStats()
+	if err != nil {
+		return nil, fmt.Errorf("loading BPF data: %w", err)
+	}
+
+	// Debug events map is unsupported due to pinning
+	spec.Maps["debug_events"] = &ebpf.MapSpec{
```
rafaelroquetto, line 49: I wonder if there's a better way here (genuinely, not talking rhetorically) for these maps. This tells me the ebpf code is using the logger/debug events but yet here it's marked as unsupported. So theoretically either it should not be using it altogether.

Correct me if I am wrong, but what you are trying to solve here is the case in which only statso11y is active, and thus the ring buffer used by the ebpf logger is not available, is that right? Why does `k_tcp.c` need to include `bpf_dbg.h`? I don't see any calls to `bpf_dbg_printk` or the likes.

This touches on a deeper point: perhaps for this case we can do without logging, but we should think about a way to share maps across orthogonal tracers seamlessly. tpinjector is one such tracer, completely orthogonal, but it works because it's still a `ebpf.Tracer` and gets loaded together with the others.

Here we have an entirely orthogonal pipeline, which is good. Usually we could solve this by pinning the map, but we don't want to do that. So perhaps down the road we need to make `eventContext.EBPFMaps` a singleton that can transcend "pipeline" boundaries, or something like that.

pkg/internal/statsolly/ebpf/stats_tracer.go
```diff
@@ -0,0 +1,101 @@
...
+		Name:       "dummy_map",
+		Type:       ebpf.RingBuf,
+		Pinning:    ebpf.PinNone,
+		MaxEntries: uint32(os.Getpagesize()),
+	}
+
+	// msg_buffer_mem map is unsupported due to pinning
```
rafaelroquetto: ditto - it would be really good if we could properly address this before merging this PR (but I am not sure about the effort required). This feels like a bad short-cut but perhaps this is what it is for now.

pkg/internal/statsolly/ebpf/stats_tracer_nonlinux.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"structs"
+
+	ciliumebpf "github.com/cilium/ebpf"
+)
+
+type StatsFetcher struct{}
+
+type StatsTCPRtt struct {
```
rafaelroquetto, line 16: is this really required? could the code that use this type be made linux only as well?

pkg/statsolly/agent/agent.go
```diff
@@ -0,0 +1,194 @@
...
+
+	status Status
+}
+
+type ebpFetcher interface {
+	io.Closer
+	StatsEventsMap() *ciliumebpf.Map
+}
+
+// StatsAgent instantiates a new agent, given a configuration.
```
rafaelroquetto: imho comments like this should be purged, they are stating the obvious

bpf/statsolly/k_tcp.c
mmat11: @rafaelroquetto tracing programs are not viable on arm64 and kernel 5.x, we should have both with fallback to kprobe after verifying that the kernel has bpf trampolines -- I think we need that for `fexit/inet_csk_accept` too

rafaelroquetto: oh damn, good to know! it is what it is then!

pkg/statsolly/agent/agent.go
pinoOgni: Fine by me, I just copied it from `FlowsAgent` :)

pkg/internal/statsolly/ebpf/stats_tracer_nonlinux.go
pinoOgni: I added it because (If I remember correctly) the lint on mac os was complaining about it. I'll check again.

pkg/internal/statsolly/ebpf/stats_tracer.go
pinoOgni: Ops, it's a typo. I just removed it and it's working. I think it was solved after my two refactors and your work done in #1373

bpf/statsolly/maps/stats_events.h
pinoOgni: Yes, I made a note of it and then forgot. Thanks!
Update: I had a note about it but I'm thinking that right now is not needed to pin it. Later with multiple probes it will be needed.

pkg/internal/statsolly/ebpf/stats_tracer.go
pinoOgni: > I wonder if there's a better way here (genuinely, not talking rhetorically) for these maps. This tells me the ebpf code is using the logger/debug events but yet here it's marked as unsupported. So theoretically either it should not be using it altogether.
> 
> Correct me if I am wrong, but what you are trying to solve here is the case in which only statso11y is active, and thus the ring buffer used by the ebpf logger is not available, is that right? 

Nope but it can be a use case.

> Why does `k_tcp.c` need to include `bpf_dbg.h`? I don't see any calls to `bpf_dbg_printk` or the likes.

`k_tcp` doesn't use `bpf_dbg_printk` because I changed it to `bpf_d_printk` in one of my recent refactors. Basically evenf if i'm not using `bpf_dbg_printk` the `debug_events` map is present in the Spec because, for example `k_tcp.c` has `#include <common/connection_info.h>`. I added this shortcut because the verifier was complaining, and I noticed the same solution in netolly.

> 
> This touches on a deeper point: perhaps for this case we can do without logging, but we should think about a way to share maps across orthogonal tracers seamlessly. tpinjector is one such tracer, completely orthogonal, but it works because it's still a `ebpf.Tracer` and gets loaded together with the others.
> 
> Here we have an entirely orthogonal pipeline, which is good. Usually we could solve this by pinning the map, but we don't want to do that. So perhaps down the road we need to make `eventContext.EBPFMaps` a singleton that can transcend "pipeline" boundaries, or something like that.

Fine by me :)

bpf/statsolly/maps/stats_events.h
pinoOgni: Rafael had the same idea in the previous PR. Here's the reply: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1273#discussion_r2790431527

WDYT?

bpf/statsolly/maps/stats_events.h
rafaelroquetto: So, I was thinking about these maps in general, and do you think we should perhaps pin all of those internally by default, just to reduce the cognitive load?

bpf/statsolly/maps/stats_events.h
mariomac: Ok! Let's keep it simple ATM.

bpf/statsolly/maps/stats_events.h
pinoOgni: Do you mean to route `statsolly` (also `netolly` and `dnstracer`) through the same `ProcessTracer/loadSpec` path so the maps will get pinning, bpffs fallback, and lifecycle management? I mean remove the current workaround and do something consistent. 
Fine by me, IMO I can do this in another PR.

rafaelroquetto: that's right - and yeah, that may involve some refactoring, so you are right, another PR is probably best here than this one.

pkg/internal/statsolly/ebpf/stats_tracer.go
rafaelroquetto: cool - if it makes sense (you be the judge) you can include something on the next PR as well

pkg/internal/statsolly/ebpf/stats_tracer_nonlinux.go
pinoOgni: I checked and IMO is ok to leave as is. It's the same as `netolly` and others. Unfortunately, the linter mac os is complaining, and I have no other ideas at the moment.

internal/test/integration/test_utils.go
```diff
@@ -359,3 +365,20 @@ func newHTTP2Transport() *http.Transport {
...
+		address = u.Host
+	}
+
+	require.EventuallyWithT(t, func(ct *assert.CollectT) {
+		conn, err := net.DialTimeout("tcp", address, 2*time.Second)
+
+		if assert.NoError(ct, err, "TCP server at %s is not reachable", address) {
+			conn.Close()
+		}
+	}, time.Duration(minutes)*time.Minute, 1*time.Second)
```
mmat11, line 383:
```suggestion
	}, minutes*time.Minute, time.Second)
```

isn't int*time.Minute already a time.Duration?

pinoOgni: Go doesn't allow `int * time.Duration` directly, from the doc: https://pkg.go.dev/time#pkg-constants
```
second := time.Second
fmt.Print(int64(second/time.Millisecond)) // prints 1000
```

mmat11: I see 👍🏻

PR #1584, opened 2026-03-18
Move DynamicPIDSelector to its own mutually exclusive swarm node

damemi:
https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1388 added DynamicPIDSelector to be able to restrict instrumentation to specific processes. 

The problem I was seeing is that even when selecting just 1 pid with the Dynamic selector, I was still seeing OBI traces from other processes in the cluster. It seems like there are several default/empty fallback matchers that needed to be excluded when using DynamicPIDSelector.

Trying to add more checks for that was complicated, and @mariomac suggested moving DynamicSelector to its own swarm node, with the static matcher being mutually exclusive based on if dynamicselector != nil.

This PR does that, along with an updated change to pipe/msg/queue to ensure thread safety, that was caught by go race tests for both consumers subscribing to the same input queue

## Checklist

### DIFF
--- a/pkg/appolly/discover/dynamic_pid_selector_test.go
+++ b/pkg/appolly/discover/dynamic_pid_selector_test.go
@@ -29,7 +29,7 @@ func pidMultisetEqual(a, b []app.PID) bool {
 
 // readPIDNotifyBatchesUntil reads from ch until the concatenation of batches matches want
 // as a multiset (order of batches and within batches does not matter).
-func readPIDNotifyBatchesUntil(t *testing.T, ch <-chan []app.PID, want []app.PID) []app.PID {
+func readPIDNotifyBatchesUntil(t *testing.T, ch <-chan []app.PID, want []app.PID) {
 	ctx, cancel := context.WithTimeout(t.Context(), 2*time.Second)
 	defer cancel()
 	var got []app.PID
@@ -44,7 +44,6 @@ func readPIDNotifyBatchesUntil(t *testing.T, ch <-chan []app.PID, want []app.PID
 			t.Fatalf("timeout reading notify batches: got %v want %v", got, want)
 		}
 	}
-	return got
 }
 
 func TestDynamicPIDSelector_AddPIDs_RemovePIDs_GetPIDs(t *testing.T) {


pkg/appolly/discover/finder.go
```diff
@@ -79,6 +79,14 @@ func (pf *ProcessFinder) Start(ctx context.Context, opts ...ProcessFinderStartOp
 	tracerEvents := msgh.QueueFromConfig[Event[*ebpf.Instrumentable]](pf.cfg, "tracerEvents")
 
 	configCriteria := FindingCriteria(pf.cfg, startConfig.dynamicPIDSelector != nil)
+	// When using the dynamic PID selector, use only it: pass nil so the matcher gets
+	// just [dynamicSelector.AsSelector()]. Otherwise the matcher would also have config
+	// criteria (e.g. the default RegexSelector with empty path/port), and a process
+	// matches if it matches any criterion—so every process with pod metadata would
+	// match the empty criterion and get instrumented alongside the dynamic PIDs.
+	if startConfig.dynamicPIDSelector != nil {
```
mariomac: Given that we are passing this condition as second argument to `FindingCriteria`, I think we should then remove the second argument in FindingCriteria (and assume it's `false` in the body o the function).

mariomac (commented):
Good finding! I wonder if we can simplify the implementation. Instead of adding the dynamic PIDs logic in the criteria matcher, and fill it with if-elses to consider separating the DyamicPIDSelector to another node:

1. Change the signature of `criteriaMatcherProvider` to return `swarm.Instancer`
2. If DynamicPidSelector != nil, criteriaMatcherProvider will just return `swarm.EmptyRunFunc`, that would prevent any message from flowing through there.
3. Move the PIDs filtering logic to a simple node with similar input/output channels. It would return the PIDs matching function, or `swarm.EmptyRunFunc` if the dynamic PID selector is null.

The architecture would be something like:
```mermaid
flowchart LR
    DockerEnricher --> |if pidsMatcher == nil| CriteriaMatcher --> ExecTyper
    DockerEnricher --> |if pidsMatcher != nil| PidsMatcher --> ExecTyper
```

pkg/appolly/discover/finder.go
mmat11: what about reverting the order?

eg.
```
if dynamicSelector == nil {
    configCriteria = FindingCriteria(pf.cfg, false)
}
```

damemi:
@mariomac thanks, I agree. DynamicPIDSelector is showing more and more that it fits as its own thing and not quite into the existing flows. Even more so, I am picturing this being expanded to more than just PID selection as an overall dynamic matcher. On the surface, that seems like it would fit into the existing flows but in practice I think it stands alone.

I'll try out the refactor you suggested and ping your for another round. Thanks for the example too!

pkg/appolly/discover/finder.go
damemi: this whole check actually isn't needed anymore with DynamicSelector as its own swarm mode

PR #1585, opened 2026-03-18
Optimize LookupAndDeleteMap in network flow fetcher

mariomac:
`LookupAndDeleteMap` is generating a lot of memory:

Screenshot 2026-03-18 at 16 28 41

LookupAndDeleteMap stores partial copies of the network flows in slices, which are aggregated in the invoker function.

To unnecessarily save pressure in the GC, the slices are directly aggregated in `lookupAndDeletemap`. This simplifies the code and saves many slice allocations and re-allocations.

## Checklist


rafaelroquetto (approved):
Nice! Thanks for this - once this is in we can maybe look into adding the batch API stuff, happy to help there.

PR #1589, opened 2026-03-19
Fix bpf_xdp_load_bytes unkown function on kernel < 5.18

pinoOgni:
The `bpf_xdp_load_bytes` helper function is used to copy DNS packet data into a ring buffer and it was added in  kernel version [5.18](https://docs.ebpf.io/linux/helper-function/bpf_xdp_load_bytes/). In kernel versions < 5.18 we have this error:

```
program dns_response_tracker: load program: invalid argument: invalid func unknown#189 (221 line(s) omitted)
```

So, this PR removes `bpf_xdp_load_bytes` and adds a bounded loop (`RB_RECORD_LEN`) with per-iteration packet bounds checks. 

AFAIK this helper is the safe way to copy packet data when we have multi buffer packets but we are attaching the program without `BPF_F_XDP_HAS_FRAGS` so we don't care. Note from [ebpf docs](https://docs.ebpf.io/linux/program-type/BPF_PROG_TYPE_XDP/) 
> To indicate that a program is "XDP Fragment aware" the program should be loaded with the [BPF_F_XDP_HAS_FRAGS](https://docs.ebpf.io/linux/syscall/BPF_PROG_LOAD/#bpf_f_xdp_has_frags) flag.

### DIFF
--- a/bpf/rdns/rdns_xdp.c
+++ b/bpf/rdns/rdns_xdp.c
@@ -159,7 +159,7 @@ static __always_inline void submit_dns_packet(struct xdp_md *ctx, const unsigned
         if (i >= data_len) {
             break;
         }
-        if ((const void *)(data + i + 1) > (const void *)end) {
+        if (data + i + 1 > end) {
             break;
         }
         buf[i] = data[i];


bpf/rdns/rdns_xdp.c
```diff
@@ -155,7 +155,15 @@ static __always_inline void submit_dns_packet(struct xdp_md *ctx, const unsigned
         return;
     }
 
-    bpf_xdp_load_bytes(ctx, data_offset, buf, data_len);
+    for (__u32 i = 0; i < RB_RECORD_LEN; i++) {
+        if (i >= data_len) {
+            break;
+        }
+        if ((const void *)(data + i + 1) > (const void *)end) {
```
rafaelroquetto: I reckon these casts are not required and could be removed since everything is `unsigned char*`

bpf/rdns/rdns_xdp.c
```diff
@@ -155,7 +155,15 @@ static __always_inline void submit_dns_packet(struct xdp_md *ctx, const unsigned
         return;
     }
 
-    bpf_xdp_load_bytes(ctx, data_offset, buf, data_len);
+    for (__u32 i = 0; i < RB_RECORD_LEN; i++) {
```
mmat11, line 158:
```suggestion
    for (u32 i = 0; i < RB_RECORD_LEN; i++) {
```

mmat11: nvm, this file is all underscored

PR #1598, opened 2026-03-19
Add bpf-verifier test 

pinoOgni:
## Summary

This PR was born from the need to test OBI, or at least the loading of the ebpf part on various kernels. This arose from some issues, already solved:
- #1590 
- #1589 
- #1581 

This PR adds
- a local test that loads all the ebpf objects, which can be run locally with the make `run-bpf-verifier-vm` command
- a CI test that runs it on the following kernels: `5.15 `and `6.10` for x86 and `6.14.0 `(if I'm not mistaken) for arm, which uses `ubuntu-24.04-arm`.

## Validation

### DIFF
--- a/pkg/internal/ebpf/verifier/bpf_verifier_test.go
+++ b/pkg/internal/ebpf/verifier/bpf_verifier_test.go
@@ -6,6 +6,7 @@
 package bpf_verifier_test
 
 import (
+	"errors"
 	"os"
 	"testing"
 
@@ -58,7 +59,13 @@ func loadAndVerify(t *testing.T, name string, loadFn func() (*ebpf.CollectionSpe
 				LogSizeStart: 10 * 1024 * 1024,
 			},
 		})
-		require.NoError(t, err, "BPF verifier rejected program(s)")
+		if err != nil {
+			var ve *ebpf.VerifierError
+			if errors.As(err, &ve) {
+				t.Fatalf("BPF verifier rejected program(s):\n%+v", ve)
+			}
+			require.NoError(t, err, "failed to load BPF collection")
+		}
 		coll.Close()
 	})
 }
@@ -78,12 +85,14 @@ func TestBPFVerifier(t *testing.T) {
 
 	// netolly: TC-based flow monitor
 	loadAndVerify(t, "netolly/Net", netollybpf.LoadNet)
+
 	// netolly: socket-filter-based flow monitor
 	loadAndVerify(t, "netolly/NetSk", netollybpf.LoadNetSk)
 
-	// generic tracer (uprobe-based HTTP/gRPC/...)
+	// generictracer (iter programs like ObiIterTcp are included in the main Bpf spec)
 	loadAndVerify(t, "generictracer/Bpf", generictracerbpf.LoadBpf)
-	// Go-specific tracer
+
+	// gotracer
 	loadAndVerify(t, "gotracer/Bpf", gotracerbpf.LoadBpf)
 
 	// tracepoint injector
@@ -96,7 +105,7 @@ func TestBPFVerifier(t *testing.T) {
 	// GPU event tracer
 	loadAndVerify(t, "gpuevent/Bpf", gpueventbpf.LoadBpf)
 
-	// BPF ring-buffer logger
+	// logger
 	loadAndVerify(t, "logger/Bpf", loggerbpf.LoadBpf)
 
 	// log enricher


.github/workflows/pull_request.yml
```diff
@@ -212,3 +212,76 @@ jobs:
...
+          sudo apt-get install -y --no-install-recommends qemu-utils qemu-system-x86
+
+      - name: Run BPF verifier tests in VM (kernel ${{ matrix.kernel }})
+        if: matrix.use_vm
+        env:
+          KERNEL_VERSION: ${{ matrix.kernel }}
+        run: |
+          sudo make -C internal/test/vm \
+            KERNEL_VER=${KERNEL_VERSION} \
+            ARCH=x86_64 \
```
rafaelroquetto: is this meant to be hardcoded?

pinoOgni: right now yes, it only runs when `matrix.use_vm `is `true`, which only applies to the two x86_64 matrix entries, but let me update it now so we are ok for the future. I can use `ARCH=${{ matrix.arch }}`

pkg/internal/ebpf/verifier/bpf_verifier_test.go
```diff
@@ -0,0 +1,107 @@
...
+	if err := rlimit.RemoveMemlock(); err != nil {
+		t.Skipf("cannot remove memlock limit (insufficient privileges?): %v", err)
+	}
+
+	// netolly: TC-based flow monitor
+	loadAndVerify(t, "netolly/Net", netollybpf.LoadNet)
+	// netolly: socket-filter-based flow monitor
+	loadAndVerify(t, "netolly/NetSk", netollybpf.LoadNetSk)
+
+	// generic tracer (uprobe-based HTTP/gRPC/...)
```
mmat11: nit: I think the comment is wrong here

pkg/internal/ebpf/verifier/bpf_verifier_test.go
```diff
@@ -0,0 +1,107 @@
...
+	loadAndVerify(t, "generictracer/Bpf", generictracerbpf.LoadBpf)
```
mmat11, line 93: I think generictracer's iter is missing

pkg/internal/ebpf/verifier/bpf_verifier_test.go
```diff
@@ -0,0 +1,107 @@
...
+	gpueventbpf "go.opentelemetry.io/obi/pkg/internal/ebpf/gpuevent"
+	logenricherbpf "go.opentelemetry.io/obi/pkg/internal/ebpf/logenricher"
+	loggerbpf "go.opentelemetry.io/obi/pkg/internal/ebpf/logger"
+	tpinjectorbpf "go.opentelemetry.io/obi/pkg/internal/ebpf/tpinjector"
+	watcherbpf "go.opentelemetry.io/obi/pkg/internal/ebpf/watcher"
+	netollybpf "go.opentelemetry.io/obi/pkg/internal/netolly/ebpf"
+	rdnsxdpbpf "go.opentelemetry.io/obi/pkg/internal/rdns/ebpf/xdp"
+)
+
+const privilegedEnv = "PRIVILEGED_TESTS"
```
mmat11, line 28: do we need this?

pkg/internal/ebpf/verifier/bpf_verifier_test.go
```diff
@@ -0,0 +1,107 @@
...
+			}
+		}
+
+		coll, err := ebpf.NewCollectionWithOptions(spec, ebpf.CollectionOptions{
+			Programs: ebpf.ProgramOptions{
+				// Increase log buffer so verifier rejections are not truncated.
+				LogSizeStart: 10 * 1024 * 1024,
+			},
+		})
+		require.NoError(t, err, "BPF verifier rejected program(s)")
```
mmat11: can we handle the error manually and print the verifier error in case it fails?

pkg/internal/ebpf/verifier/bpf_verifier_test.go
pinoOgni: I found it in other tests and I thought it was a convention to use. There is just a check that verifies only that it is set.

pkg/internal/ebpf/verifier/bpf_verifier_test.go
pinoOgni: sure

pkg/internal/ebpf/verifier/bpf_verifier_test.go
pinoOgni: The `ObiIterTcp` is part of the main `bpfObjects` but I was missing the `gotracerbpf.LoadBpfTP`.
Update: Ok it was old generated code.

PR #1615, opened 2026-03-20
Update `CheckOSSupport`

pinoOgni:
## Summary

This PR adds a kernel version check for vanilla kernels (5.8+) and RHEL (4.18+) and it also adds a BTF check. 

This is useful but not strictly necessary for correctness, it adds clear (early) error messages instead of a confusing BPF/BTF failure and it correctly distinguishes RHEL from vanilla kernels.

## Validation

### DIFF
--- a/pkg/obi/os.go
+++ b/pkg/obi/os.go
@@ -6,6 +6,7 @@
 package obi // import "go.opentelemetry.io/obi/pkg/obi"
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"strings"
@@ -31,11 +32,12 @@ func parseOSReleaseIsRHEL(data []byte) bool {
 	// matches ID="rhel" or ID_LIKE containing "rhel" (e.g. Rocky, AlmaLinux, CentOS set ID_LIKE="rhel ...")
 	for _, line := range strings.Split(content, "\n") {
 		var val string
-		if strings.HasPrefix(line, "id_like=") {
+		switch {
+		case strings.HasPrefix(line, "id_like="):
 			val = line[len("id_like="):]
-		} else if strings.HasPrefix(line, "id=") {
+		case strings.HasPrefix(line, "id="):
 			val = line[len("id="):]
-		} else {
+		default:
 			continue
 		}
 		val = strings.Trim(val, "\"'")
@@ -93,17 +95,17 @@ var hasBTF = func() bool {
 // tests call this directly.
 func checkOSSupport() error {
 	major, minor := kernelVersion()
-	maj, min := minKernMaj, minKernMin
+	maj, mnr := minKernMaj, minKernMin
 	if isRHELBased() {
-		maj, min = minRHELKernMaj, minRHELKernMin
+		maj, mnr = minRHELKernMaj, minRHELKernMin
 	}
-	if major < maj || (major == maj && minor < min) {
+	if major < maj || (major == maj && minor < mnr) {
 		return fmt.Errorf("kernel version %d.%d not supported. Minimum required version is %d.%d",
-			major, minor, maj, min)
+			major, minor, maj, mnr)
 	}
 
 	if !hasBTF() {
-		return fmt.Errorf("kernel does not support BTF (CONFIG_DEBUG_INFO_BTF): no vmlinux BTF found")
+		return errors.New("kernel does not support BTF (CONFIG_DEBUG_INFO_BTF): no vmlinux BTF found")
 	}
 
 	return nil


pkg/obi/os.go
```diff
@@ -7,27 +7,59 @@ package obi // import "go.opentelemetry.io/obi/pkg/obi"
...
+// Minimum required Kernel version: 5.8 (or 4.18 for RHEL-based distros)
+const (
+	minKernMaj, minKernMin         = 5, 8
+	minRHELKernMaj, minRHELKernMin = 4, 18
+)
 
 var kernelVersion = ebpfcommon.KernelVersion
 
+func isRHELBased() bool {
+	data, err := os.ReadFile("/etc/os-release")
```
mariomac, line 54: could we move this to something like:

```go
var readOSRelease = func() ([]byte, error) {
   return os.ReadFile("/etc/os-release")
}
```

So we can mock it and add some extra test cases in the `os_test.go` unit tests.

pinoOgni: Oh, yes!

pkg/obi/os.go
```diff
@@ -7,27 +7,59 @@ package obi // import "go.opentelemetry.io/obi/pkg/obi"
...
-const minKernMaj, minKernMin = 4, 18
+// Minimum required Kernel version: 5.8 (or 4.18 for RHEL-based distros)
+const (
+	minKernMaj, minKernMin         = 5, 8
+	minRHELKernMaj, minRHELKernMin = 4, 18
+)
 
 var kernelVersion = ebpfcommon.KernelVersion
 
+func isRHELBased() bool {
```
mmat11: I think this should be wrapped in a sync.Once or a sync.OnceFunc

pkg/obi/os.go
```diff
@@ -7,27 +7,59 @@ package obi // import "go.opentelemetry.io/obi/pkg/obi"
...
+	if isRHELBased() {
+		maj, min = minRHELKernMaj, minRHELKernMin
+	}
+	if major < maj || (major == maj && minor < min) {
 		return fmt.Errorf("kernel version %d.%d not supported. Minimum required version is %d.%d",
-			major, minor, minKernMaj, minKernMin)
+			major, minor, maj, min)
 	}
+
+	if _, err := btf.LoadKernelSpec(); err != nil {
```
mmat11: this call might be expensive, can we just search the vmlinux btf and os.Stat() it? you can find all paths here https://github.com/libbpf/libbpf/blob/master/src/btf.c#L5494

pkg/obi/os.go
pinoOgni: let me check

pkg/obi/os.go
pinoOgni: Thanks!

pkg/obi/os.go
```diff
@@ -7,27 +7,59 @@ package obi // import "go.opentelemetry.io/obi/pkg/obi"
...
 
+func isRHELBased() bool {
+	data, err := os.ReadFile("/etc/os-release")
+	if err != nil {
+		return false
+	}
+	content := strings.ToLower(string(data))
+	// matches ID="rhel" or ID_LIKE containing "rhel" (e.g. Rocky, AlmaLinux, CentOS set ID_LIKE="rhel ...")
+	for _, line := range strings.Split(content, "\n") {
+		if strings.HasPrefix(line, "id=") || strings.HasPrefix(line, "id_like=") {
```
rafaelroquetto: so, `os-release` accepts quoted values, i.e `id="rhel"` so we should probably handle this here for robustness.

If you want to go the extra mile, but I am not sure if this is worth it, you can use a combination of `split.NewBytesIterator()`, `bytes.EqualFold()` `bytes.Contains()` to do without most of memory allocations, but perhaps that's premature optimisation territory (it depends on how that code would look like)

pkg/obi/os.go
```diff
@@ -7,27 +7,59 @@ package obi // import "go.opentelemetry.io/obi/pkg/obi"
...
+	return false
+}
+
 // CheckOSSupport returns an error if the running operating system does not support
-// the minimum required Beyla features.
+// the minimum required OBI features.
 func CheckOSSupport() error {
 	major, minor := kernelVersion()
-	if major < minKernMaj || (major == minKernMaj && minor < minKernMin) {
+	maj, min := minKernMaj, minKernMin
```
rafaelroquetto: could we further wrap this so that it reads:

```
majVerMin, minVerMin := minimumSupporterdVersion()
```

that should also shut the linter up

pinoOgni:
Updates after review:
- `checkOSSupport()` contains the actual logic and is wrapped in `sync.OnceValue`
-  `isRHELBased()` and `hasBTF()` (checks sysfs and fallback paths) are inside `checkOSSupport()`

rafaelroquetto (approved):

PR #1628, opened 2026-03-23
Uniform ebpf map loading using `ebpfconvenience` package

pinoOgni:
## Summary

This PR removes the duplicated `debug_events` dummy-map workaround from statsolly, netolly, and rdns by extracting the  logic from `tracer_linux.go` into a common `ebpfconvenience` pkg. Now, all eBPF loaders go through the same path, getting proper PinInternal map resolution, ringbuf alignment, and bpffs pin path support.

Finally it shares `stats_events` map across multiple eBPF programs using `OBI_PIN_INTERAL`.

## Validation

### DIFF
--- a/pkg/internal/netolly/ebpf/sock_tracer.go
+++ b/pkg/internal/netolly/ebpf/sock_tracer.go
@@ -97,7 +97,7 @@ func NewSockFlowFetcher(
 		"g_bpf_debug":      true,
 	}, sharedMaps, &mu, ""); err != nil {
 		printVerifierErrorInfo(err)
-		return nil, fmt.Errorf("loading and assigning BPF objects: %w", err)
+		return nil, err
 	}
 
 	fd, err := unix.Socket(unix.AF_PACKET, unix.SOCK_RAW, int(htons(unix.ETH_P_ALL)))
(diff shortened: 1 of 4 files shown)


pkg/ebpf/tracer_linux.go
```diff
@@ -45,74 +42,15 @@ type instrumenter struct {
...
-
-	collOpts.Programs = ebpf.ProgramOptions{LogSizeStart: 640 * 1024}
-	collOpts.Maps = ebpf.MapOptions{PinPath: otelBPFFSPath}
-
-	if err := bundle.Spec.LoadAndAssign(bundle.Objects, collOpts); err != nil {
+	if err := ebpfconvenience.LoadSpec(
+		bundle.Spec,
+		bundle.Objects,
+		bundle.Constants,
+		eventContext.EBPFMaps,
```
mmat11, line 50: if every subsystem forwards their map of shared maps, the "internal pinning" is within the same subsystem, right? eg. an internal pinned map in gotracer will be different than the same map in netolly

pinoOgni: Yep, each subsystem (statsolly, netolly, appolly) owns its own `sharedMaps` instance, so `OBI_PIN_INTERNAL` means that is shared within this subsystem. 

For example for `debug_events`, using this configuration:
```
open_port: "9092"
...
metrics:
  features:
    - network
    - application
    - stats
 ```
we have:
```
sudo bpftool map show | grep debug_events
9754: ringbuf  name debug_events  flags 0x0
9755: ringbuf  name debug_events  flags 0x0
9766: ringbuf  name debug_events  flags 0x0
```

PR #1633, opened 2026-03-24
Move src and dst port from `NetFlowId` to `CommonAttrs`

pinoOgni:
## Summary

This PR continues the work done in #1572. It updates netolly to include source and destination ports in `CommonAttrs` to match the statsolly implementation. Finally it changes the field type from `int` to `uint16` to match the ebpf generated types and avoid unnecessary conversions.

## Validation

### DIFF
--- a/pkg/export/otel/expirer_test.go
+++ b/pkg/export/otel/expirer_test.go
@@ -71,12 +71,12 @@ func TestNetMetricsExpiration(t *testing.T) {
 	// WHEN it receives metrics
 	metrics.Send([]*ebpf.Record{
 		{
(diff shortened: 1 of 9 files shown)


mariomac:
I just realized something I feel a bit "dangerous" from a programmers' side.

Now an `ebpf.Record` type has several duplicate attributes: `id.SrcPort` and `commonAttrs.SrcPort` (same for addresses and DST port). It could be pretty easy that any contributor in the future uses/modifies the `id.SrcPort` where it seems it should use `commonAttrs.SrcPort`. Is there any way to prevent the risks associated to duplicities of the same data?

PR #1655, opened 2026-03-26
Add payload extraction support for Anthropic Claude

grcevski:
## Summary

Adding payload extraction for Anthropic (Claude). This works more less the same as the OpenAI support, except it handles the Anthropic format and special features. I've tested with regular message calls, streams, multi tool calls, errors.

Next on my list on this topic is adding metrics support now that we have both OpenAI and Anthropic. We can then expand to perhaps with adding AWS Bedrock.

## Validation

### DIFF
--- a/devdocs/features.md
+++ b/devdocs/features.md
@@ -17,8 +17,7 @@
 | AWS S3        |    All    |         All | CreateBucket, DeleteBucket, PutObject, DeleteObject, ListBuckets, ListObjects, GetObject |  Yes   |                 No |                                                                                                                             N/A
 | AWS SQS       |    All    |         All | All                                                                                      |  Yes   |                 No |                                                                                                                             N/A
 | SQL++         |    All    |         All | All                                                                                      |  Yes   |                 No |                                                                                                                             N/A
-| OpenAI        |    All    |         All | All                                                                                      |  Yes   |                 No |                                                                                                                             N/A
-| Anthropic     |    All    |         All | All                                                                                      |  Yes   |                 No |                                                                                                                             N/A
+| GenAI         |    All    |         All | All                                                                                      |  Yes   |                 No |                                                                                            Supported vendors: OpenAI, Anthropic
 
 ## Go Instrumentation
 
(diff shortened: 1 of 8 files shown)


internal/test/oats/ai/configs/instrumenter-config-traces-genai.yml
pinoOgni, line 1: nit: should we rename `instrumenter-config-traces-openai.yml` to something more generic, like `instrumenter-config-traces-genai.yml` or `instrumenter-config-traces-llm.yml`?

devdocs/features.md
```diff
@@ -17,6 +17,8 @@
 | AWS S3        |    All    |         All | CreateBucket, DeleteBucket, PutObject, DeleteObject, ListBuckets, ListObjects, GetObject |  Yes   |                 No |                                                                                                                             N/A
 | AWS SQS       |    All    |         All | All                                                                                      |  Yes   |                 No |                                                                                                                             N/A
 | SQL++         |    All    |         All | All                                                                                      |  Yes   |                 No |                                                                                                                             N/A
+| OpenAI        |    All    |         All | All                                                                                      |  Yes   |                 No |                                                                                                                             N/A
```
NimrodAvni78: Do you think we should generelize it to GenAI (matching the semantic conventions), and under description we could add a list of vendors we support?

pkg/config/payload_extraction.go
```diff
@@ -37,6 +38,8 @@ type HTTPConfig struct {
 	SQLPP SQLPPConfig `yaml:"sqlpp"`
 	// OpenAI payload extraction
 	OpenAI OpenAIConfig `yaml:"openai"`
```
NimrodAvni78: maybe we should unify those here as well to be GenAI?

grcevski: yeah good idea, keep this as genai to match the spec and then put the config details for each supported vendor inside.

devdocs/features.md
grcevski: Done

internal/test/oats/ai/configs/instrumenter-config-traces-genai.yml
grcevski: Done

PR #1662, opened 2026-03-26
Move gputracer in common tracer

pinoOgni:
## Summary

This PR moves the GPU tracer from the generic tracers group to the common one. This way the GPU tracer loads once regardless of whether Go or generic tracers are active, just like `logenricher` and `tpinjector`. 

**Note:** Not sure if this is the best idea. Loading the GPU tracer in the common group means it would also activate when only Go programs are being traced, which is wasteful. AFAIK, CUDA is used by C/C++/Python programs and not go.
The fact is that, GPU tracer needs to be independent of the generic tracer, but also shouldn't load for Go-only scenarios. Maybe we should add its own tracer group (like a "gpu tracers group") that loads independently?

## Validation


pkg/appolly/discover/finder.go
mmat11: the whole function body can be rewritten as:
```
return []ebpf.Tracer{generictracer.New(pidFilter, cfg, metrics)}
```

PR #1672, opened 2026-03-28
GenAI metrics

grcevski:
## Summary

This PR adds metrics support for the two GenAI client metrics according to the OTel Semantic conventions.

I'm adding token usage and client timing support metrics. https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics

## Validation


pkg/export/attributes/names/attrs.go
```diff
@@ -234,3 +234,14 @@ const (
 const (
 	DNSQuestionName = Name(semconv.DNSQuestionNameKey)
 )
+
+// GenAI events
+
+const (
+	GenAIOperationName   = Name(semconv.GenAIOperationNameKey)
+	GenAIProviderName    = Name(semconv.GenAIProviderNameKey)
+	GenAITokenTypeInput  = Name(semconv.GenAITokenTypeKey)
```
NimrodAvni78, line 247: I think this should be `GenAITokenType` and its values are input / output?
https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/

grcevski: Yes that's how it is, I know it's confusing with the split of input and output attributes, but it was the most elegant way to do it. If you see the actual attribute name for both input and output is `semconv.GenAITokenTypeKey`. 

Essentially, with GenAI events, it's the first time we have a single event produce two separate metrics: input->tokens, output->tokens. All of our current metrics have one set of attributes and a value for that. But these events have two. I had several options:

1. Keep one set of attributes, but then write hacky code to manipulate them at the time we write them in 
the series here: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1672/changes#diff-0edcf6672a788ef8fb6d31c28bc6551d54009745630033924b2e6e6060cabf44R920. I would have to turn the set into array, find the type and replace. It would be very messy.
2. Create fake input and output attributes that map to the same semconv and then use our existing infrastructure `genAIInputTokenUsage, attrs := r.genAIInputTokenUsage.ForRecord(span)` and `genAIOutputTokenUsage, attrs := r.genAIOutputTokenUsage.ForRecord(span)`. Then all I had to do is return `input` or `output` in the span_getters.go code, but for the same field `semconv.GenAITokenType`.
3. Split the event when generated from the payload extraction, but then we'd have to de-duplicate them for traces which is a mess.

I opted for option 2. which was the cleanest to implement.


If you see the tests, this is how it looks in the final metrics:

```
 - promql: gen_ai_client_token_usage_count{gen_ai_operation_name="response",gen_ai_provider_name="openai",gen_ai_token_type="input",gen_ai_request_model="gpt-5-mini",gen_ai_response_model="gpt-5-mini-2025-08-07"}
    value: '> 0'
  - promql: gen_ai_client_token_usage_count{gen_ai_operation_name="response",gen_ai_provider_name="openai",gen_ai_token_type="output",gen_ai_request_model="gpt-5-mini",gen_ai_response_model="gpt-5-mini-2025-08-07"}
    value: '> 0'
```

NimrodAvni78: ohhh ok got it, i would maybe leave some comment on this?
or at least make both attributes not part of the semantic convention, if they are just fake attributes and not used at the end. having 1 semantic convention attribute and 1 not makes me think think there is something off
WDYT?

grcevski: good idea, I'll add a comment here as well. I added comments where the metrics are created but not here.

PR #1654, opened 2026-03-25
Split`bpf_dbg_printk` to have less than 3 arguments

pinoOgni:
## Summary

This PR solves the following error on kernel `5.15`, when `g_bpf_debug=true`:
```
2026-03-24T12:42:54.5889915Z         load program: invalid argument:
2026-03-24T12:42:54.5911977Z         	; asm("%[res] = *(u32 *)(%[base] + %[offset])"
2026-03-24T12:42:54.5930805Z         	0: (61) r4 = *(u32 *)(r1 +0)
2026-03-24T12:42:54.5952407Z         	; asm("%[res] = *(u32 *)(%[base] + %[offset])"
2026-03-24T12:42:54.5970596Z         	1: (61) r2 = *(u32 *)(r1 +4)
2026-03-24T12:42:54.5998464Z         	; return (data + sizeof(struct iphdr) > ctx_xdp_data_end(ctx)) ? NULL : data;
2026-03-24T12:42:54.6009150Z         	2: (bf) r3 = r4
2026-03-24T12:42:54.6020387Z         	3: (07) r3 += 34
2026-03-24T12:42:54.6031873Z         	; if (!iph) {
2026-03-24T12:42:54.6048345Z         	4: (2d) if r3 > r2 goto pc+1550
2026-03-24T12:42:54.6084919Z         	 R1=ctx(id=0,off=0,imm=0) R2_w=pkt_end(id=0,off=0,imm=0) R3_w=pkt(id=0,off=34,r=34,imm=0) R4_w=pkt(id=0,off=0,r=34,imm=0) R10=fp0
2026-03-24T12:42:54.6095731Z         	5: (bf) r9 = r4
2026-03-24T12:42:54.6106714Z         	6: (07) r9 += 14
2026-03-24T12:42:54.6122821Z         	; if (iph->protocol != IPPROTO_UDP) {
...
// very long output
...
2026-03-24T12:42:56.0378749Z         	535: (7b) *(u64 *)(r10 -152) = r3
2026-03-24T12:42:56.0398563Z         	; bpf_dbg_printk("id=%x, qr=%u, opcode=%u, aa=%u, tc=%u, rd=%u, ra=%u\n",
2026-03-24T12:42:56.0408867Z         	536: (7b) *(u64 *)(r10 -16) = r3
2026-03-24T12:42:56.0419175Z         	537: (7b) *(u64 *)(r10 -104) = r1
[...]

PR #1680, opened 2026-03-30
Update bpf verifier test with constants

pinoOgni:
## Summary

This PR replaces `TestBPFVerifier` with `TestBPFVerifierWithConstants`, which tests the bpf verifier across all combinations of constant values (debug flags, feature toggles, sampling rates, etc.). 

This is useful to check the verifier accepts all bpf programs under every combination of runtime constants, since different constant values cause the verifier to evaluate different code paths. 

For example, it found a bug on kernel 5.15 solved here #1654.

## Validation

### DIFF
--- a/pkg/internal/ebpf/verifier/bpf_verifier_test.go
+++ b/pkg/internal/ebpf/verifier/bpf_verifier_test.go
@@ -166,16 +166,11 @@ func TestBPFVerifierWithConstants(t *testing.T) {
 	})
 
 	// gotracer
-	// g_bpf_loop_enabled=true requires bpf_loop (kernel >= 5.17).
-	bpfLoopValues := []any{false}
-	if ebpfcommon.SupportsEBPFLoops(slog.Default(), false) {
-		bpfLoopValues = []any{true, false}
-	}
 	forEachCombination(t, "gotracer/Bpf", gotracerbpf.LoadBpf, []constOption{
 		{"g_bpf_debug", []any{true, false}},
 		{"g_bpf_traceparent_enabled", []any{true, false}},
 		{"g_bpf_header_propagation", []any{true, false}},
-		{"g_bpf_loop_enabled", bpfLoopValues},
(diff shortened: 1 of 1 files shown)


pkg/internal/ebpf/verifier/bpf_verifier_test.go
```diff
@@ -83,34 +145,79 @@ func TestBPFVerifier(t *testing.T) {
...
+	// g_bpf_loop_enabled=true requires bpf_loop (kernel >= 5.17).
+	bpfLoopValues := []any{false}
+	if ebpfcommon.SupportsEBPFLoops(slog.Default(), false) {
+		bpfLoopValues = []any{true, false}
+	}
+	forEachCombination(t, "gotracer/Bpf", gotracerbpf.LoadBpf, []constOption{
+		{"g_bpf_debug", []any{true, false}},
+		{"g_bpf_traceparent_enabled", []any{true, false}},
+		{"g_bpf_header_propagation", []any{true, false}},
+		{"g_bpf_loop_enabled", bpfLoopValues},
```
mmat11: I think we can inline it here
 
```suggestion
		{"g_bpf_loop_enabled", ebpfcommon.SupportsEBPFLoops(slog.Default(), false)},
```

pinoOgni: We may do:
```
{"g_bpf_loop_enabled", []any{ebpfcommon.SupportsEBPFLoops(slog.Default(), false)}},
```
but that only tests the single supported value, losing coverage of the false path when loops are available. Right now we are testing both true and false on capable kernels.

mmat11: we practically never load it with `false` if it's supported

pinoOgni: and you're right, I'll do it.

PR #1685, opened 2026-03-30
Add example use of OBI for nginx

MrAlias:
Highlight how OBI supports instrumentation of the nginx web server.

### DIFF
--- a/examples/nginx/generate-traffic.sh
+++ b/examples/nginx/generate-traffic.sh
@@ -78,12 +78,12 @@ route_specs() {
 EOF
 }
 
-request_route() {
-  local path="$1"
-  local expected_code="$2"
-  local http_code
+fetch_http_code() {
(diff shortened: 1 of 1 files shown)


examples/nginx/README.md
```diff
@@ -0,0 +1,220 @@
...
+helm repo update
+helm upgrade --install obi open-telemetry/opentelemetry-ebpf-instrumentation \
+  --namespace obi-nginx-example \
+  -f examples/nginx/k8s/04-obi-values.yaml
+```
+
+Port-forward the UIs:
+
+```bash
+kubectl -n obi-nginx-example port-forward svc/edge-nginx 8080:8080
```
grcevski, line 141: I think it would be nice to tell the users here that they need two separate terminal windows. The first command will block, so if someone is just copy pasting the grafana port won't ever start.

PR #1689, opened 2026-03-31
Add support for memcached protocol

marctc:
## Summary

Add support for [memcached protocol](https://github.com/memcached/memcached/blob/master/doc/protocol.txt). Implementation is quite straightforward because it is a text protocol, but `noreply` needs explicit handling since some valid requests intentionally produce no response and therefore cannot follow the normal request/response pairing path.

Fixes https://github.com/grafana/beyla/issues/2187

## Validation


bpf/generictracer/k_tracer.c
```diff
@@ -719,8 +719,23 @@ int BPF_KPROBE(obi_kprobe_tcp_close, struct sock *sk, long timeout) {
...
+            tcp_req_t *trace = bpf_ringbuf_reserve(&events, sizeof(tcp_req_t), 0);
+            if (trace) {
+                __builtin_memcpy(trace, existing_tcp, sizeof(tcp_req_t));
+                trace->end_monotime_ns = bpf_ktime_get_ns();
+                trace->resp_len = 0;
+
+                bpf_dbg_printk("Sending pending TCP trace on close: len=%d", trace->len);
+                bpf_ringbuf_submit(trace, get_flags());
+            }
+        }
```
rafaelroquetto: could you move this to a helper called along the lines of `finish_ongoing_tcp_req()`?

bpf/generictracer/k_tracer.c
```diff
@@ -719,8 +719,23 @@ int BPF_KPROBE(obi_kprobe_tcp_close, struct sock *sk, long timeout) {
...
+
         cleanup_tcp_trace_info_if_needed(&info);
+        bpf_map_delete_elem(&ongoing_tcp_req, &info);
```
rafaelroquetto: this can be moved inside the if block on line 724

bpf/generictracer/k_tracer.c
```diff
@@ -719,8 +719,23 @@ int BPF_KPROBE(obi_kprobe_tcp_close, struct sock *sk, long timeout) {
...
         info.pid = pid_from_pid_tgid(id);
         terminate_http_request_if_needed(&info);
-        bpf_map_delete_elem(&ongoing_tcp_req, &info);
+
+        tcp_req_t *existing_tcp = bpf_map_lookup_elem(&ongoing_tcp_req, &info);
+        if (existing_tcp && existing_tcp->end_monotime_ns == 0 &&
+            existing_tcp->protocol_type == k_protocol_type_unknown) {
+            tcp_req_t *trace = bpf_ringbuf_reserve(&events, sizeof(tcp_req_t), 0);
+            if (trace) {
+                __builtin_memcpy(trace, existing_tcp, sizeof(tcp_req_t));
```
rafaelroquetto: I think this works:
```suggestion
                *trace = *existing_tcp;
```

but double check it.

you can actually just use `bpf_ringbuf_output` and pass `existing_tcp` directly after adjusting the end monotime and resp len.

pkg/ebpf/common/memcached_detect_transform.go
```diff
@@ -0,0 +1,551 @@
...
+
+func isMemcached(buf *largebuf.LargeBuffer) bool {
+	if buf.Len() < minMemcachedFrameLen {
+		return false
+	}
+
+	return isMemcachedOp(buf.UnsafeView())
+}
+
+func isMemcachedOp(buf []byte) bool {
```
rafaelroquetto: instead of materialising the entire buffer with `UnsafeView()`, it's best to pass `buf` here, and refactor the code to use the large buffer API directly. For example:

```
func largebufFirstLine(buf *largebuf.LargeBuffer) ([]byte, bool) {
    crPos := buf.IndexByteAt(0, '\r')

    if crPos <= 0 || crPos+1 >= buf.Len() {
        return nil, false
    }

    lf, err := buf.U8At(crPos + 1)

    if err != nil || lf != '\n' {
        return nil, false
    }

    line, err := buf.UnsafeViewAt(0, crPos) // materialise only here

    return line, err == nil
}
```

bpf/generictracer/k_tracer.c
marctc: But this was already before. Wouldn't that prevent cleanups?

bpf/generictracer/protocol_tcp.h
```diff
@@ -106,6 +106,29 @@ static __always_inline void cleanup_tcp_trace_info_if_needed(pid_connection_info
...
+        existing_tcp->protocol_type == k_protocol_type_unknown) {
+        tcp_req_t *trace = bpf_ringbuf_reserve(&events, sizeof(tcp_req_t), 0);
+        if (trace) {
+            *trace = *existing_tcp;
+            trace->end_monotime_ns = bpf_ktime_get_ns();
+            trace->resp_len = 0;
+
+            bpf_dbg_printk("Sending pending TCP trace on close: len=%d", trace->len);
+            bpf_ringbuf_submit(trace, get_flags());
+        }
```
rafaelroquetto:
```suggestion
        existing_tcp->end_monotime_ns = bpf_ktime_get_ns();
        existing_tcp->resp_len = 0;

        bpf_dbg_printk("Sending pending TCP trace on close: len=%d", existing_tcp->len);
        
        bpf_ringbuf_output(&events, existing_tcp, sizeof(*existing_tcp), get_flags());
```

rationale: simplifies the code, and let the kernel copy the data directly into the ring buffer instead of explicitly doing it here.

bpf/generictracer/k_tracer.c
rafaelroquetto: since now you look it up with `bpf_map_lookup_elem`, you now know when it exists and when it doesn't, so it makes sense to only clean it up if it really exists to avoid calling the delete helper unconditionally. You fixed that in the new version, so all good :)

pkg/ebpf/common/common.go
```diff
@@ -267,6 +269,18 @@ func NewEBPFParseContext(cfg *config.EBPFTracer, spansChan *msg.Queue[[]request.
...
 
+	if spansChan != nil {
+		emitSpans = func(spans []request.Span) {
+			if len(spans) == 0 {
+				return
+			}
+			if filter != nil {
+				spans = filter.Filter(spans)
+			}
+			spansChan.Send(spans)
```
grcevski: this is supposedly using a deprecated API...

pkg/ebpf/common/tcp_detect_transform.go
```diff
@@ -161,6 +177,18 @@ func ReadTCPRequestIntoSpan(parseCtx *EBPFParseContext, cfg *config.EBPFTracer,
 			}
 			return TCPToRedisToSpan(event, op, text, status, db, redisErr), false, nil
 		}
+	case isMemcached(requestBuffer) && isMemcached(responseBuffer):
+		span, err := ProcessPossibleMemcachedEvent(parseCtx, event, requestBuffer, responseBuffer)
+		if errors.Is(err, errFallback) {
```
grcevski: Should we add the errFallback condition to the case condition? Can we mistakenly think this is memcached on other types of requests and then fail to properly parse the other protocols.

pkg/ebpf/common/common.go
```diff
@@ -267,6 +269,18 @@ func NewEBPFParseContext(cfg *config.EBPFTracer, spansChan *msg.Queue[[]request.
 		dnsEvents = expirable.NewLRU(1024, dnsEventExpireHandler(spansChan, filter), cfg.DNSRequestTimeout)
 	}
 
+	if spansChan != nil {
```
grcevski: since we are adding this generic function, can we please reuse it in dnsEventExpireHandler. Instead of passing the spansChan and the filter we can pass the emit spans function.

pkg/ebpf/common/tcp_detect_transform.go
marctc: I removed the `errFallback` condition and I made `isMemcached` function more robust

PR #1694, opened 2026-03-31
Remove check-ebpf-ver-synced make target

MrAlias:
This target was used to ensure the github.com/cilium/ebpf package version was the same in the generator as that used locally. This was made unneeded in #1290.

Blocking #1624

### DIFF
--- a/bpf/bpfcore/placeholder.go
+++ b/bpf/bpfcore/placeholder.go
@@ -4,6 +4,3 @@
 //go:build obi_bpf
 
 package bpfcore // import "go.opentelemetry.io/obi/bpf/bpfcore"
-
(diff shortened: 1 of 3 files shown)


bpf/bpfcore/placeholder.go
```diff
@@ -4,6 +4,3 @@
 //go:build obi_bpf
 
 package bpfcore // import "go.opentelemetry.io/obi/bpf/bpfcore"
-
-// Store loader version here in order to run the full test suite on upgrades.
-// github.com/cilium/ebpf v0.20.0
```
mmat11, line 9: this is used to trigger a full CI run when the library is bumped

PR #1710, opened 2026-04-02
Adding payload extraction for Gemini (Google AI Studio)

ishanjainn:
## Summary

Adding payload extraction for Gemini (Google AI Studio). This works more less the same as the OpenAI support, except it handles the Gemini format and special features. I've tested with regular message calls, streams, multi tool calls, errors.

## Validation


pkg/ebpf/common/http/gemini.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+		if strings.Contains(host, "generativelanguage.googleapis.com") ||
+			strings.Contains(host, "aiplatform.googleapis.com") {
+			return true
+		}
+	}
+
+	if req != nil {
+		for _, header := range []string{
+			"X-Goog-Api-Key",
```
grcevski: I wonder if these might be set to true for other kinds of Google API calls? I'm guessing the response header might be sufficient?

What I used in the past to look for interesting headers is print all headers here and look for something that says it's openai or anthropic. Maybe you've done that already, just suggesting :).

pkg/export/otel/metrics.go
```diff
@@ -914,7 +914,7 @@ func (r *Metrics) record(span *request.Span, mr *MetricsReporter) {
 			if mr.is.DBEnabled() && (span.SubType == request.HTTPSubtypeSQLPP || span.SubType == request.HTTPSubtypeElasticsearch) {
 				dbClientDuration, attrs := r.dbClientDuration.ForRecord(span)
 				dbClientDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			} else if mr.is.GenAIEnabled() && (span.SubType == request.HTTPSubtypeAnthropic || span.SubType == request.HTTPSubtypeOpenAI) {
+			} else if mr.is.GenAIEnabled() && request.IsGenAISubtype(span.SubType) {
```
grcevski, line 921: Nice!

grcevski (approved):
Looks great, I had some minor comments, but otherwise LGTM! I see some formatting issues, so we'll need to get this through the CI :).

pkg/ebpf/common/http/gemini.go
```diff
@@ -0,0 +1,106 @@
...
+			host = req.URL.Host
+		}
+		if strings.Contains(host, "generativelanguage.googleapis.com") ||
+			strings.Contains(host, "aiplatform.googleapis.com") {
+			return true
+		}
+	}
+
+	if req != nil {
+		for _, header := range []string{
```
rafaelroquetto: here the slice literal escapes to the heap - since you are only testing for 2 strings, just use an "or", e.g.

```
if req.Header.Get("X-Goog-Api-Key") != "" || req.Header.Get("X-Goog-Api-Client") != "" {
    return true
}
```

ishanjainn: Removed the block itself as wasnt needed as @grcevski mentioned

pkg/appolly/app/request/span.go
```diff
@@ -781,6 +875,9 @@ func HTTPSpanStatusCode(span *Span) string {
 				if span.GenAI.Anthropic != nil && span.GenAI.Anthropic.Output.Error != nil && span.GenAI.Anthropic.Output.Error.Type != "" {
 					return StatusCodeError
 				}
+			if span.GenAI.Gemini != nil && span.GenAI.Gemini.Output.Error != nil && span.GenAI.Gemini.Output.Error.Status != "" {
+				return StatusCodeError
+			}
```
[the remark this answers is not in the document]
ishanjainn: Fixed

pkg/appolly/app/request/span.go
```diff
@@ -238,8 +245,9 @@ type AWSSQS struct {
 }
 
 type GenAI struct {
-	OpenAI    *VendorOpenAI
-	Anthropic *VendorAnthropic
+	OpenAI       *VendorOpenAI
+	Anthropic    *VendorAnthropic
+	Gemini    *VendorGemini
```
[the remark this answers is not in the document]
ishanjainn: fixed

pkg/ebpf/common/http/gemini.go
ishanjainn: Sweet, Yes fixed

pkg/ebpf/common/http/gemini.go
```diff
@@ -0,0 +1,95 @@
...
+	"io"
+	"log/slog"
+	"net/http"
+	"strings"
+
+	"go.opentelemetry.io/obi/pkg/appolly/app/request"
+)
+
+func isGemini(req *http.Request, respHeader http.Header) bool {
+	if req != nil {
```
NimrodAvni78: not sure we ever do checks on the host for other protocols, Im assuming due to any number of reasons like proxies or private environments
if we have response headers that we can reliably rely one (like it seems here with X-Gemini-Service-Tier) i think that should be the priority

pkg/appolly/app/request/span.go
```diff
@@ -369,6 +377,92 @@ type AnthropicError struct {
...
+
+type VendorGemini struct {
+	Input  GeminiRequest
+	Output GeminiResponse
+	Model  string
+}
+
+type GeminiRequest struct {
+	Contents         json.RawMessage `json:"contents"`
+	SystemInstruct   *GeminiContent  `json:"systemInstruction,omitempty"`
```
NimrodAvni78: nit: why not `SystemInstruction`?

NimrodAvni78 (approved):
This is great! thank you
couple minor stuff
otherwise LGTM

mmat11 (approved):

pkg/ebpf/common/http/gemini.go
ishanjainn: Simplified to only check the X-Gemini-Service-Tier, Thanks for checking!

pkg/appolly/app/request/span.go
ishanjainn: Yup, Fixed

rafaelroquetto (approved):

grcevski:
Thanks for your contribution @ishanjainn !

PR #1712, opened 2026-04-02
Add support for NATS protocol

### LINKED ISSUE #1143: Instrument and collect telemetry for NATS

marctc:
## Summary

Add support to capture span from clients using [NATS protocol](https://docs.nats.io/reference/reference-protocols/nats-protocol).

Fixes #1143 

## Validation


pkg/ebpf/common/nats_detect_transform.go
```diff
@@ -0,0 +1,381 @@
...
+
+	return rest, nil
+}
+
+func validateNATSSID(field []byte) error {
+	if len(field) == 0 {
+		return errors.New("invalid NATS sid")
+	}
+	for _, b := range field {
+		if (b < '0' || b > '9') && (b < 'a' || b > 'z') && (b < 'A' || b > 'Z') {
```
grcevski: sql_detect_transform has the same code, maybe we can combine this into a common helper?

pkg/ebpf/common/nats_detect_transform.go
```diff
@@ -0,0 +1,381 @@
...
+		var meta struct {
+			Name string `json:"name"`
+		}
+		if err := json.Unmarshal(rest, &meta); err != nil {
+			return natsFrame{}, errors.New("invalid NATS JSON payload")
+		}
+
+		return natsFrame{clientID: meta.Name, valid: true}, nil
+	case "SUB":
+		if len(fields) != 3 && len(fields) != 4 {
```
grcevski: Nit: I found these statements confusing with != 3 && !=4. It might be better to write them as !(len==3 || len==4)

pkg/ebpf/common/nats_detect_transform_test.go
```diff
@@ -0,0 +1,371 @@
...
+	header := []byte("NATS/1.0\r\nX-Test: python\r\n\r\n")
+	payload := []byte("python-nats-1")
+	hdrLen := len(header)
+	totalLen := hdrLen + len(payload)
+	assert.True(t, isNATS(largebuf.NewLargeBufferFrom(
+		[]byte(fmt.Sprintf("PONG\r\nHMSG updates.orders subA %d %d\r\n%s%s\r\n", hdrLen, totalLen, header, payload)),
+	)))
+}
+
+func TestParseNATSFrame(t *testing.T) {
```
grcevski, line 17: Can we add more tests with all the variations of !=4 && !=5?

grcevski (commented):
I just had a few minor comments, but this is looking pretty good!

pkg/ebpf/common/nats_detect_transform.go
marctc: this is how it was before, but linter is not happy:

```
Error: pkg/ebpf/common/nats_detect_transform.go:139:6: QF1001: could apply De Morgan's law (staticcheck)
		if !(len(fields) == 3 || len(fields) == 4) {

```

grcevski: ah ok, cool!

grcevski (approved):
LGTM!

pkg/ebpf/common/nats_detect_transform.go
```diff
@@ -0,0 +1,386 @@
...
+	if err != nil {
+		return natsFrame{}, err
+	}
+
+	fields := bytes.Fields(line)
+	if len(fields) == 0 {
+		return natsFrame{}, errors.New("empty NATS command")
+	}
+
+	command := string(bytes.ToUpper(fields[0]))
```
rafaelroquetto: protocol sql has `equalFoldASCII` - we should use that here instead, as it operates directly on byte slices (as a rule of thumb, don't allocate strings when byte slices cut it)

The pattern is to define constants once and use them like this:

```
var (
    natsCmdPUB  = []byte("PUB")
    natsCmdHPUB = []byte("HPUB")
    // ...
)
/* ... */
equalFoldASCII(fields[0], natsCmdPUB)
```

pkg/ebpf/common/nats_detect_transform.go
```diff
@@ -0,0 +1,386 @@
...
+	rest := bytes.TrimSpace(line[len(command):])
+	if !json.Valid(rest) {
+		return nil, errors.New("invalid NATS JSON payload")
+	}
+
+	return rest, nil
+}
+
+func validateNATSSID(field []byte) error {
+	if !isASCII(string(field)) {
```
rafaelroquetto: also here, we allocate a string each time - so instead of using `isACII()`, we should introduce `isASCIIBytes()` that takes `[]byte` instead and use it throughout

pkg/ebpf/common/nats_detect_transform.go
```diff
@@ -0,0 +1,386 @@
...
+		return natsFrame{valid: true}, nil
+	case "UNSUB":
+		if len(fields) != 2 && len(fields) != 3 {
+			return natsFrame{}, errors.New("invalid UNSUB control line")
+		}
+		if err := validateNATSSID(fields[1]); err != nil {
+			return natsFrame{}, err
+		}
+		if len(fields) == 3 {
+			if _, err := strconv.Atoi(string(fields[2])); err != nil {
```
rafaelroquetto: this `strconv.Atoi(string(fields[XXX))` allocates each time, we should use a small helper:

```
func parseIntField(b []byte) (int, error) {
    return strconv.Atoi(unsafe.String(unsafe.SliceData(b), len(b)))
}
```

pkg/ebpf/common/nats_detect_transform.go
```diff
@@ -0,0 +1,386 @@
...
+				return natsFrame{}, errors.New("invalid UNSUB max_msgs")
+			}
+		}
+		return natsFrame{valid: true}, nil
+	case "PUB":
+		subject, size, err := parseNATSPayloadFields(fields)
```
rafaelroquetto: so here, you dispatch to `parseNATSPayloadFields` (good!), but in every other switch you are doing the parsing inline.

I feel this entire function is a bit out of hand/too big and should be split.

imagine something like:

```
case "PUB":
   handlePUB(...)
case "MSG":
   handleMSG(...)
```
etc...

as this is hurting clarity/readability as it is and makes it hard to review

pkg/ebpf/common/tcp_detect_transform.go
```diff
@@ -186,6 +188,23 @@ func ReadTCPRequestIntoSpan(parseCtx *EBPFParseContext, cfg *config.EBPFTracer,
 			return request.Span{}, true, fmt.Errorf("failed to handle Memcached event: %w", err)
 		}
 		return span, false, nil
+	case isNATS(requestBuffer) || isNATS(responseBuffer):
+		info, extraInfo, ignore, err := ProcessPossibleNATSEvent(event, requestBuffer, responseBuffer)
```
rafaelroquetto: `isNATS` walks all frames of the buffer.` ProcessPossibleNATSEvent` then calls `ProcessNATSEvent` which walks the same buffer again from scratch. Each buffer can be parsed up to twice, when both sides are checked that's 4 passes total.

have a look at detectSQLPayload: it is the only scan, there's no separate heuristic that does its own full parse first.

pkg/ebpf/common/nats_detect_transform.go
```diff
@@ -0,0 +1,386 @@
...
+	hostname := ""
+	hostPort := 0
+
+	if trace.ConnInfo.S_port != 0 || trace.ConnInfo.D_port != 0 {
+		peer, hostname = (*BPFConnInfo)(unsafe.Pointer(&trace.ConnInfo)).reqHostInfo()
+		hostPort = int(trace.ConnInfo.D_port)
+	}
+
+	reqType := request.EventTypeNATSClient
+	if trace.Direction == 0 {
```
rafaelroquetto: AGENTS.md says "Do not use magic numbers." - not sure why copilot did not latch onto that.

Claude also notes: "Worth also checking whether Direction is the right discriminator vs IsServer. SQL, Postgres, and MySQL all use IsServer. Using Direction is consistent with MQTT, but IsServer is set by the eBPF layer based on the socket's listening state, which is more reliable."

pkg/ebpf/common/nats_detect_transform.go
```diff
@@ -0,0 +1,386 @@
...
+	return false
+}
+
+func TCPToNATSToSpan(trace *TCPRequestInfo, data *NATSInfo) request.Span {
+	peer := ""
+	hostname := ""
+	hostPort := 0
+
+	if trace.ConnInfo.S_port != 0 || trace.ConnInfo.D_port != 0 {
+		peer, hostname = (*BPFConnInfo)(unsafe.Pointer(&trace.ConnInfo)).reqHostInfo()
```
rafaelroquetto: I think this can work without unsafe.Pointer

peer, hostname = (*BPFConnInfo)(&trace.ConnInfo).reqHostInfo()

rafaelroquetto (changes requested):
Overall, I think we should do another pass to split the code for clarity, and also ensure no unecessary string allocations are taking place (that's one of the reasons why I refactored the sql code when doing the large buffers API, so I would prefer if we did not drift from the approach).

It might be worth it doing a local review telling your agent to look into AGENTS.md - for some reason I think copilot (the github one) has been sloppy and missing a lot of obvious cases.

pkg/appolly/app/request/span.go
```diff
@@ -824,6 +830,14 @@ func spanAttributes(s *Span) SpanAttributes {
 			}
 		}
 		return attrs
+	case EventTypeNATSServer, EventTypeNATSClient:
+		return SpanAttributes{
+			"serverAddr": SpanHost(s),
+			"serverPort": strconv.Itoa(s.HostPort),
+			"operation":  s.Method,
+			"clientId":   s.Statement,
+			"topic":      s.Path,
```
rafaelroquetto: "topic" is not NATS terminology — the protocol calls this a "subject". This key surfaces in the JSON span representation (via `MarshalJSON`), so a debug log would emit `"topic": "updates.orders"` where `"subject": "updates.orders"` would be accurate. Suggest renaming to `"subject"` for consistency with the protocol spec (cf. MQTT which correctly uses `"topic"` for its equivalent concept).

---
*Generated by Claude (claude-sonnet-4-6) via Claude Code*

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -762,6 +765,20 @@ func TraceAttributesSelector(span *request.Span, optionalAttrs map[attr.Name]str
...
+	case request.EventTypeNATSServer, request.EventTypeNATSClient:
+		operation := request.MessagingOperationType(span.Method)
+		attrs = []attribute.KeyValue{
+			request.ServerAddr(request.HostAsServer(span)),
+			request.ServerPort(span.HostPort),
+			messagingSystemNATS,
+			semconv.MessagingDestinationName(span.Path),
+			semconv.MessagingClientID(span.Statement),
+			operation,
+		}
```
rafaelroquetto, line 778: span.ContentLength carries the payload size captured by the parser (for HPUB/HMSG this is total size including headers, not body-only), but it is not emitted here. Was this intentional, or should it be exported as `messaging.message.body.size` / `messaging.message.envelope.size` per OTel messaging semconv? If payload size export is planned, note that the HPUB/HMSG parser stores headers+body as a single total — body-only size would need to be tracked separately before either attribute can be emitted correctly.

---
*Generated by Claude (claude-sonnet-4-6) via Claude Code*

pkg/appolly/app/request/span.go
rafaelroquetto: "topic" is not NATS terminology — the protocol calls this a "subject". This key surfaces in the JSON span representation (via `MarshalJSON`), so a debug log would emit `"topic": "updates.orders"` where `"subject": "updates.orders"` would be accurate. Suggest renaming to `"subject"` for consistency with the protocol spec (cf. MQTT which correctly uses `"topic"` for its equivalent concept).

---
*Generated by Claude (claude-sonnet-4-6) via Claude Code*

internal/test/oats/nats/yaml/oats_python_nats.yaml
```diff
@@ -0,0 +1,36 @@
...
+interval: 500ms
+expected:
+  traces:
+  - traceql: '{ .messaging.system = "nats" && .messaging.operation.type = "publish" && .messaging.destination.name = "updates.orders" }'
+    equals: publish updates.orders
+    attributes:
+      messaging.destination.name: updates.orders
+      messaging.operation.type: publish
+      messaging.system: nats
+      server.port: '4222'
```
rafaelroquetto, line 16: The Python client connects with `name="python-nats-test"`, which OBI captures from the CONNECT frame and emits as `messaging.client_id`. This attribute isn't asserted here — if CONNECT frame extraction were silently broken, this test would still pass. Consider adding:
```yaml
messaging.client_id: python-nats-test
```
to both span attribute blocks. The unit tests cover CONNECT parsing, but an end-to-end assertion here would close the loop from wire capture to exported attribute.

---
*Generated by Claude (claude-sonnet-4-6) via Claude Code*

rafaelroquetto (approved):
Hey Marc, this look really great! Goooood stuff! It looks great to the naked eye - I had Claude do a pass on it nonetheless and it found some small things, up to you to decide whether they are wroth fixing.

mmat11 (approved):
lgtm!

PR #1734, opened 2026-04-02
Export NetNS functionality

rafaelroquetto:
## Summary

Export `WithNetNS` so it can be used outside the nodejs code. This is part of an incoming effort to:

 - fix the socket iterator to register all connections (it needs to be run on the target process netnamespace)
 - the upcoming socket tracer (WIP) which has a socket iterator as well
 - it also makes sense overall to have this public as there's nothing nodejs specific to it.

## Validation


pkg/internal/netns/netns.go
```diff
@@ -40,7 +40,6 @@ func withNetNS(hostPid int, fn func() error) error {
 
 	defer func() {
 		if err := unix.Setns(int(selfNS.Fd()), unix.CLONE_NEWNET); err != nil {
-			// FIXME log instead
```
NimrodAvni78, line 43: should we fix this?

rafaelroquetto: I double checked it and if this happens we are toast - it's pretty much unrecoverable at this stage, so we should probably panic and respawn after all...

PR #1748, opened 2026-04-04
Fix panic call to proper error handling in prom/expirer.go

yaten2302:
## Summary

This PR removes `panic` call in `pkg/export/prom/expirer.go`, and add error handling instead of panic.

## Validation

### DIFF
--- a/pkg/export/expire/expiry_map.go
+++ b/pkg/export/expire/expiry_map.go
@@ -67,6 +67,37 @@ func (ex *ExpiryMap[T]) GetOrCreate(lbls []string, instancer func() T) T {
 	return instance
 }
 
+// GetOrCreateWithError is like GetOrCreate but supports an instancer that may fail.
+func (ex *ExpiryMap[T]) GetOrCreateWithError(lbls []string, instancer func() (T, error)) (T, error) {
+	now := ex.clock()
+
+	h := labelsKey(lbls)
+	ex.mt.RLock()
+	e, ok := ex.entries[h]
+	ex.mt.RUnlock()
+	if ok {
+		ex.mt.Lock()
+		e.lastAccess = now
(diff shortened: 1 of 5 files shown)


pkg/export/expire/expiry_map.go
```diff
@@ -67,6 +67,37 @@ func (ex *ExpiryMap[T]) GetOrCreate(lbls []string, instancer func() T) T {
 	return instance
 }
 
+// GetOrCreateWithError is like GetOrCreate but supports an instancer that may fail.
+func (ex *ExpiryMap[T]) GetOrCreateWithError(lbls []string, instancer func() (T, error)) (T, error) {
```
NimrodAvni78, line 71: i would just replace the `GetOrCreate` implementation with this one, no need to keep both implementation

yaten2302: I tried doing that, but the issue is that then we would need to change all the components using this `GetOrCreate` func and there are several other components attached directly or indirectly to this func.

Should I still try to replace this 👀 ? WDYT?

NimrodAvni78: How many components are there more then you changed?
if not many i would opt to change all callsites, and worst case panic with the response and keep current behavior, best case if you can fix those too :)

grcevski:
Hi @yaten2302, what's the motivation for this change? Did you experience this panic while running OBI?

pkg/export/expire/expiry_map.go
yaten2302: Don't remember the exact number :) but yeah, I'll try to fix those as well!

yaten2302:
Got this @NimrodAvni78 👍
I'll surely pick up something from the good first issues and the docs changes required.
Closing this PR!

PR #1749, opened 2026-04-05
Replace fatal startup paths with error returns

ethical-buddy:
## Summary
Replace several abrupt startup and cleanup exits with normal error returns.

This change:

  - returns discovery criteria normalization errors instead of panicking
  - returns OTEL reporter-pool construction errors instead of panicking
  - returns Node.js netns restore failures instead of panicking during deferred cleanup
  - updates the offset example programs touched here to avoid panic/log.Fatal in setup and runtime stop
    paths
  - updates the affected tests for the new discovery criteria error-return contract

## Validation

Validation performed:

  - gofmt -w on touched files
  - git diff --check
  - attempted go build ./...
  - attempted go test ./...

Note:
  Full go build ./... and go test ./... currently fail in this checkout due to missing generated eBPF Go
  bindings/types, for example in pkg/ebpf/common, pkg/internal/netolly/ebpf, pkg/internal/statsolly/ebpf,
  and pkg/internal/rdns/ebpf/xdp. Those failures appear unrelated to this patch.

### DIFF
--- a/devdocs/features.md
+++ b/devdocs/features.md
@@ -64,6 +64,12 @@ Large payloads are streamed to userspace across multiple ring-buffer events and 
 
 Equivalent YAML keys live under `ebpf.buffer_sizes.{http,mysql,kafka,postgres}`.
 
+## Startup And Shutdown Behavior
+
+OBI prefers explicit error returns during startup and controlled shutdown over abrupt process termination.
+Initialization failures should be propagated to the caller so cleanup, exporter flushing, and other deferred shutdown
+logic can still run when applicable.
+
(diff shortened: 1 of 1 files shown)


ethical-buddy:
Hi @grcevski  
  I didn’t hit those invariant panics while running OBI. My motivation was broader than it should
  have been: I was trying to reduce abrupt exits in long-lived paths so cleanup and shutdown could
  still happen. After your feedback, I agree I over-applied that rule and converted some panics that
  are really there to catch contributor mistakes, not recoverable runtime errors. I’ve revised the new PR accordingly.

PR #1750, opened 2026-04-05
Add HTTP full body extraction 

### LINKED ISSUE #396: HTTP header and body extraction

NimrodAvni78:
## Summary

Resolves https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/396

Add HTTP body extraction capabilities to OBI, in the same http_enrichment pipeline as header extraction
rules now are either valid to headers or body
only works on json payloads detected by content-type
allows for obfuscation based on jsonpaths
allows filtering on route globs and methods

If you think the API it selfs needs change regarding what we support, if we need to separate the config from the headers config or anything else would love some feedback!

note:
request response payload semantic conventions do not exist yet, but we are following the [suggested spec](https://github.com/open-telemetry/semantic-conventions/pull/3521) here

## Validation


docs/config-schema.json
```diff
@@ -954,6 +988,11 @@
       "type": "object",
       "description": "InternalMetricsConfig options for the different metrics exporters"
     },
+    "JSONPathExpr": {
+      "properties": {},
+      "type": "object",
+      "description": "JSONPathExpr holds a JSONPath expression string and its compiled form."
```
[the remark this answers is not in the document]
NimrodAvni78: need to implement JSONSchema for this struct

pkg/config/payload_extraction.go
```diff
@@ -89,31 +90,78 @@ type AnthropicConfig struct {
...
+				return err
+			}
+		case HTTPParsingRuleTypeBody:
+			if err := validateBodyRule(i, rule); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
NimrodAvni78: its is called in `main.go`

pkg/config/payload_extraction.go
```diff
@@ -267,35 +399,46 @@ func (HTTPParsingScope) JSONSchema() *jsonschema.Schema {
...
+	*m = upper
+	return nil
 }
 
-func (HTTPParsingMatchOrder) JSONSchema() *jsonschema.Schema {
+func (HTTPMethod) JSONSchema() *jsonschema.Schema {
 	return &jsonschema.Schema{
 		Type: "string",
-		Enum: []any{"first_match_wins"},
+		Enum: []any{"GET", "POST", "PUT", "DELETE", "PATCH", "HEAD", "OPTIONS"},
```
grcevski: Is it possible to return HTTPMethodGET...?

pkg/ebpf/common/http/http_enrichment.go
```diff
@@ -115,3 +250,43 @@ func applyHeaderAction(
...
+		if routePatterns[i].MatchString(path) {
+			return true
+		}
+	}
+	return false
+}
+
+// methodMatches returns true if the request method matches any of the specified methods.
+// If no methods are specified, all methods match.
+func methodMatches(methods []config.HTTPMethod, method string) bool {
```
grcevski, line 299: Could we turn this into a map lookup, to save some processing time?

pkg/config/payload_extraction.go
```diff
@@ -89,31 +90,78 @@ type AnthropicConfig struct {
...
+	// Policy controls the default behavior
 	Policy HTTPParsingPolicy `yaml:"policy"`
 	// Rules is an ordered list of include/exclude/obfuscate rules.
-	// Rules are evaluated according to Policy.MatchOrder.
 	Rules []HTTPParsingRule `yaml:"rules"`
 }
 
-// HTTPParsingPolicy defines the default action and match strategy for http enrichment rules.
+// Validate checks the enrichment config for cross-field consistency errors.
+// Required fields (action, type, scope) are enforced by validate:"required" tags.
```
grcevski, line 100: Do you mind adding unit tests for these validate functions?

pkg/config/payload_extraction.go
```diff
@@ -89,31 +90,78 @@ type AnthropicConfig struct {
...
+func (c EnrichmentConfig) Validate() error {
```
grcevski, line 101: Maybe I missed it, but I think this Validate() function is never called. I think we are missing a call from config.go.

pkg/config/payload_extraction.go
```diff
@@ -89,31 +90,78 @@ type AnthropicConfig struct {
...
+	}
+	return nil
+}
+
+// HTTPParsingPolicy defines the default action for http enrichment rules.
 type HTTPParsingPolicy struct {
-	// DefaultAction specifies what to do when no rule matches: "include" or "exclude"
-	DefaultAction HTTPParsingAction `yaml:"default_action" env:"OTEL_EBPF_HTTP_ENRICHMENT_DEFAULT_ACTION"`
-	// MatchOrder controls how rules are evaluated: "first_match_wins"
-	MatchOrder HTTPParsingMatchOrder `yaml:"match_order" env:"OTEL_EBPF_HTTP_ENRICHMENT_MATCH_ORDER"`
```
grcevski, line 104: I remember we had a test with this removed option set, can you double check and remove it if it's there, to avoid confusion. I think the config for the http-enrichment integration tests.

pkg/config/payload_extraction.go
```diff
@@ -147,29 +200,91 @@ func (t HTTPParsingRuleType) MarshalText() ([]byte, error) {
...
-	// Patterns is a list of glob patterns to match the rule against
+	// Patterns is a list of glob patterns to match header names against (headers only)
 	Patterns []services.GlobAttr `yaml:"patterns"`
-	// CaseSensitive controls whether matching is case-sensitive.
+	// CaseSensitive controls whether header matching is case-sensitive (headers only)
 	CaseSensitive bool `yaml:"case_sensitive"`
+	// ObfuscationJSONPaths is a list of JSONPath expressions for fields to obfuscate (body only)
+	ObfuscationJSONPaths []JSONPathExpr `yaml:"obfuscation_json_paths"`
+	// RoutePatterns is a list of glob patterns to match the request path against (shared)
+	RoutePatterns []services.GlobAttr `yaml:"route_patterns"`
```
grcevski: These are called route patterns, but they actually match the span.Path, not the reduced span.Route. I'd suggest we rename the option to url_path_patterns or similar, since the route matcher hasn't run yet when we do the extraction.

pkg/ebpf/common/http/http_enrichment.go
```diff
@@ -89,7 +128,104 @@ func resolveHeaderAction(
...
+
+	if effectiveAction == config.HTTPParsingActionExclude {
+		return ""
+	}
+
+	// Read and re-wrap body
+	bodyBytes, err := io.ReadAll(body)
+	_ = body.Close()
+	if err != nil || len(bodyBytes) == 0 {
+		return ""
```
grcevski, line 215: should we call resetBody("") here?

grcevski (commented):
This is great work! I left a few comments!

pkg/ebpf/common/http/http_enrichment.go
```diff
@@ -89,7 +128,104 @@ func resolveHeaderAction(
...
+	default:
+		effectiveAction = config.HTTPParsingActionExclude
+	}
+
+	if effectiveAction == config.HTTPParsingActionExclude {
+		return ""
+	}
+
+	// Read and re-wrap body
+	bodyBytes, err := io.ReadAll(body)
```
grcevski: I think we need to call the uncompress helper, since the response might be gzip encoded, for example: func getResponseBody(resp *http.Response) ([]byte, error) { in responses.go, the Body will be the zipped stream

pkg/config/payload_extraction.go
NimrodAvni78: you are right i thought the Validate call was recursive to all structs for some reason, will fix this

pkg/ebpf/common/http/http_enrichment.go
NimrodAvni78: changed body handling to close the buffer

pkg/config/payload_extraction.go
NimrodAvni78: done

pkg/config/payload_extraction.go
NimrodAvni78: fixed integration tests config

grcevski (approved):
LGTM! Awesome stuff!

mmat11 (approved):
lgtm!

Issue #1504, opened 2026-03-09
Go HTTP Should have the same features as TCP based HTTP


NimrodAvni78:
For Go Http spans, that arrive from uprobe instrumentation, is sent as a different event type then the normal TCP based http instrumentation. which means it is missing the support for some features in the TCP based http pipeline. most notably large buffer support and post classification of http (for things like graphql, aws, openAI and so on)

NimrodAvni78:
solved by https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1766

Issue #1031, opened 2025-12-18
add support for twmb/franz-go go kafka client


rlankfo:
A few kafka golang clients are already supported. Requesting support for https://github.com/twmb/franz-go as well. Thanks!

PR #1766, opened 2026-04-07
Implement generic Go protocol support

grcevski:
## Summary

This PR provides generic Go instrumentation support for libraries that aren't covered by the targeted Go instrumentations.

Basic design:

- I added a new map that tracks the goroutines and connections that are handled by the targeted Go instrumentations.
- I added new probes to the return of netFD.read and on netFD.write, which are the entry points in the generic tracer.
- I refactored the code a bit so I can reuse it from Go, which now has this extra argument - the light weight thread (e.g. Goroutine).

The current first iteration of this code doesn't support context propagation within the Go applications. I'll follow-up with a PR on this. I also decided to never try sending HTTP2/gRPC through the generic code, since I don't think there are popular implementations outside of the base libraries. We can enable this in the future if we encounter any.

I added an integration test with Go fiber and Franz-Go (Kafka client for Go we don't support). I also tried the Go MQTT test and it worked on first try:
```
obi-1                | 2026-04-07 21:00:48.479048 (397.234µs[397.234µs]) MQTTClient(subType=0) 0 publish test/topic(test/topic) [172.22.0.6 as testserver.integration-test:58070]->[172.22.0.3 as 172.22.0.3:1883] contentLen:0B responseLen:0B svc=[integration-test/testserver go] traceparent=[00-a45afa449123027bd2f5e57e41525a7b-2316337560a60c16[0000000000000000]-00]
```

I also tested each OATS test individually and looked at the kernel trace to ensure we didn't end up in the generic tracer when we didn't intend to.

TODO in this PR:
- [x] Measure the performance impact on the Go benchmark
I measured the performance overhead of adding the new probes and it ended up not impacting the main Go path at all. There was a small slowdown because of inefficiency in the netFD.Write probe, but I changed the code to fix this.

In follow-up work:
- Make context propagation work, should be easy, I just need to add the goroutine parent lookup in the same spot where we check the rest of the runtimes.
- Add probes for Go TLS in the same way as netFD.Read/Write. 

Relates to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1504
Relates to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1031

## Validation


bpf/common/http_types.h
```diff
@@ -31,19 +32,30 @@
 // Preface PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n https://datatracker.ietf.org/doc/html/rfc7540#section-3.5
 #define MIN_HTTP2_SIZE 24
 
+typedef struct protocol_selector {
```
rafaelroquetto, line 35: we shouldn't use bitfields - see previous discussion on https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/826

This should be easy to be refactored into a flag, e.g.

```
typedef u8 protocol_selector_t;

enum : protocol_selector_t {
  k_protocol_http = 0x1,
  k_protocol_http2 = 0x2,
  k_protocol_tcp = 0x4,
  k_protocol_selector_all = k_protocol_http | k_protocol_http2 | k_protocol_tcp
};

```

you get the idea...

If you _really_ want type-safety:

```
typedef struct protocol_selector {
  u8 flags;
} protocol_selector_t;
```
you can then pass `protocol_selector_t` around and it will not implicitly convert to a `u8`.

If you want syntatic sugar/helpers for the above:

```
static __always_inline bool protocol_selector_has_flag(protocol_selector_t s, u8 flag) {
  return s.flags & flag;
}

static __always_line bool protocol_selector_http(protocol_selector_t s) {
    return protocol_selector_has_flags, k_protocol_selector_http);
}
```

bpf/gotracer/maps/handled_by_go.h
```diff
@@ -0,0 +1,46 @@
...
+#include <common/pin_internal.h>
+#include <common/tp_info.h>
+
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, egress_key_t); // key: the connection info
+    __type(value, bool);       // value: true
+    __uint(max_entries, MAX_CONCURRENT_REQUESTS);
+    __uint(pinning, OBI_PIN_INTERNAL);
+} handled_by_go_conn SEC(".maps");
```
MrAlias, line 21: `handled_by_go_conn` is now keyed by `egress_key_t`, which only preserves `s_port` and `d_port`. After one Go-specific request stores `(client_port, server_port)`, `already_handled_request` will suppress generic tracing for any other connection that reuses the same port pair, even when the peer IP differs, right? On busy servers or clients talking to multiple backends on the same destination port, wont this silently drop unrelated spans?

bpf/gotracer/go_net.c
```diff
@@ -15,18 +15,72 @@
...
+    }
+
+    return false;
+}
+
+static __always_inline void
+cleanup_duplicate_generic_events(const pid_connection_info_t *pid_conn) {
+    bpf_map_delete_elem(&ongoing_http, pid_conn);
+    bpf_map_delete_elem(&ongoing_tcp_req, pid_conn);
+}
```
MrAlias: `cleanup_duplicate_generic_events` only deletes `ongoing_http` and `ongoing_tcp_req`. When Go gRPC / HTTP/2 traffic has already been recognized by the generic path, `ongoing_http2_connections` remains populated, right? If so, later packets on that socket will still go through the generic HTTP/2 tail calls and emit duplicate spans even after a Go-specific probe has claimed the connection.

internal/test/integration/components/gogeneric/main.go
```diff
@@ -0,0 +1,133 @@
...
+	log.Println("Starting server on " + *addr)
+	if err := app.Listen(*addr); err != nil {
+		log.Fatalf("Failed to start server: %v", err)
+	}
+}
+
+func newKafkaClient(brokers string) *kgo.Client {
+	for {
+		client, err := kgo.NewClient(
+			kgo.SeedBrokers(brokers),
```
MrAlias: I think this wants a split version of `brokers` IIRC. I.e. `["host1:9092", "host2:9092"]`, not `"host1:9092,host2:9092"`. Right?

MrAlias (commented):
What a tour de force 👏, well done.

Overall, this looks close. The main issue is the new Go/generic deduplication logic looks to suppress unrelated connections because it keys handled connections too loosely, and it does not fully clear generic HTTP/2 state when Go-specific probes take ownership.

bpf/gotracer/maps/handled_by_go.h
grcevski: Yes you are right, I'll put the full connection info.

bpf/gotracer/go_net.c
grcevski: This is fine, because we never send http2/grpc traffic from the Go code to the generic tracer logic. I think the http2 implementations all use the Go standard library and I explicitly disable those handlers, example here `http2 = 0`:

https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1766/changes#diff-ac050a7a19465fd28a670f847e5d4b68d40ea01a1c46cf55dd6621e7237b83ccR199

internal/test/integration/components/gogeneric/main.go
grcevski: ah yeah, I'll check this, I think maybe it works since I only used one broker in the test.

bpf/gotracer/go_net.c
MrAlias: `http2 = 0` in the Go path makes sense to me, and I agree that it prevents the Go-side buffer handling from invoking the generic HTTP/2 parser.

My concern was about the generic kernel-side path rather than the Go-side parsing itself. In `protocol_handler.c`, once a connection is recognized as HTTP/2/gRPC and added to `ongoing_http2_connections`, later packets for that same `pid_connection_info_t` appear to remain eligible for the generic HTTP/2 tail call while that entry exists. `cleanup_duplicate_generic_events()` currently clears `ongoing_http` and `ongoing_tcp_req`, but not `ongoing_http2_connections`, so I was worried about the case where generic tracking is established first and Go-specific tracing takes ownership afterward.

That said, my confidence here is not high. You likely know the code here better than me. I may be missing another place where HTTP/2 state is cleared, or a control-flow guarantee that prevents the generic kernel path from continuing once the Go tracer owns the connection. If that guarantee exists, then this comment can be ignored.

grcevski: Yeah, so I added a config option to our generic tracer main handler function `obi_handle_buf_with_args` https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1766/changes#diff-4d8f08da7aa4b48695574369f2e7249af510d6eb877da0178dd5a9e75eca6ec7R23.

This config now affects if we'll ever process certain type of event by the generic tracer, and those are `http, tcp and http2`. The existing code for all other languages calls the default with all protocols enabled, but when Go passes buffers to it, it explicitly says don't handle `http2`.

The generic kprobes will never fire for Go programs because we don't add the Go programs to the pid filter, since we instrument the binaries directly, so essentially we should never have that case. Having said that I can add the delete, it won't hurt in the grand scheme of things. Plus some of the Go grpc tests are failing, so there might be something here.

bpf/generictracer/k_tracer_defs.h
```diff
@@ -51,8 +54,13 @@ static __always_inline call_protocol_args_t *make_protocol_args(const pid_connec
     args->direction = direction;
     args->orig_dport = orig_dport;
     args->u_buf = (u64)u_buf;
+    args->lw_thread = lw_thread;
+    args->protocols = protocols;
     args->protocol_type = protocol_type_for_conn_info(info);
 
+    __builtin_memcpy(&args->pid_conn, info, sizeof(pid_connection_info_t));
```
rafaelroquetto:
```suggestion
    args->pid_conn = *info;
```

bpf/generictracer/protocol_http.h
```diff
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include "common/lw_thread.h"
```
rafaelroquetto: vscode shenanigans

bpf/generictracer/protocol_http.h
```diff
@@ -434,6 +439,7 @@ static __always_inline void process_http_request(
     info->status = 0;
     info->submitted = 0;
     info->len = len;
+    info->no_pid_filter = (lw_thread != lw_thread_none); // generic events generated from Go
```
rafaelroquetto: I wonder if there's a better name for `no_pid_filter`. At this point of the review it is still not clear to me what kind of value it stores (apart from maybe some go thread id?)

bpf/common/lw_thread.h
```diff
@@ -0,0 +1,10 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <bpfcore/vmlinux.h>
+
+typedef u64 lw_thread_t;
+
+enum : lw_thread_t { lw_thread_none = 0 };
```
rafaelroquetto:
```suggestion
enum : lw_thread_t { k_lw_thread_none = 0 };
```

bpf/generictracer/protocol_http.h
```diff
@@ -554,7 +562,8 @@ static __always_inline int __obi_continue2_protocol_http(struct pt_regs *ctx,
     // we copy some small part of the buffer to the info trace event, so that we can process an event even with
     // incomplete trace info in user space.
     bpf_probe_read(info->buf, FULL_BUF_SIZE, (void *)args->u_buf);
-    process_http_request(info, args->bytes_len, meta, args->direction, args->orig_dport);
+    process_http_request(
```
rafaelroquetto, line 565: not on the scope of this PR, but I wonder if we should just start passing `args` to these methods given the amount of parameters coming from `args` itself.

bpf/generictracer/protocol_tcp.h
```diff
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include "common/lw_thread.h"
```
rafaelroquetto: ditto

bpf/gotracer/maps/handled_by_go.h
```diff
@@ -0,0 +1,46 @@
...
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, go_addr_key_t); // key: goroutine
+    __type(value, bool);        // value: true
+    __uint(max_entries, MAX_CONCURRENT_REQUESTS);
+    __uint(pinning, OBI_PIN_INTERNAL);
+} handled_by_go SEC(".maps");
+
+static __always_inline void store_go_handled_connection_info(const connection_info_t *conn) {
+    if (conn) {
+        egress_key_t e_key = make_egress_key(conn);
```
rafaelroquetto:
```suggestion
        const egress_key_t e_key = make_egress_key(conn);
```

bpf/gotracer/maps/handled_by_go.h
```diff
@@ -0,0 +1,46 @@
...
+        bpf_map_update_elem(&handled_by_go_conn, &e_key, &(bool){true}, BPF_ANY);
+    }
+}
+
+static __always_inline void store_go_handled_goroutine(const go_addr_key_t *goaddr) {
+    bpf_map_update_elem(&handled_by_go, goaddr, &(bool){true}, BPF_ANY);
```
rafaelroquetto: should we be checking for nullptr here as well?

bpf/gotracer/go_net.c
```diff
@@ -15,18 +15,72 @@
...
+    pid_connection_info_t p_conn;
+} net_args_t;
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __type(key, go_addr_key_t); // goroutine
+    __type(value, net_args_t);
+    __uint(max_entries, MAX_CONCURRENT_REQUESTS);
+    __uint(pinning, OBI_PIN_INTERNAL);
+} ongoing_fd_reads SEC(".maps");
```
rafaelroquetto: I think this should be on its own header

bpf/gotracer/go_net.c
```diff
@@ -15,18 +15,72 @@
...
+                                                    const go_addr_key_t *goaddr) {
+    egress_key_t e_key = make_egress_key(conn);
+    go_addr_key_t *g = bpf_map_lookup_elem(&handled_by_go_conn, &e_key);
+    if (g) {
+        return true;
+    }
+    bool *found = bpf_map_lookup_elem(&handled_by_go, goaddr);
+    if (found) {
+        return true;
+    }
```
rafaelroquetto:
```suggestion
    const egress_key_t e_key = make_egress_key(conn);
    go_addr_key_t *g = bpf_map_lookup_elem(&handled_by_go_conn, &e_key);
    
    if (g) {
        return true;
    }
    
    const bool *found = bpf_map_lookup_elem(&handled_by_go, goaddr);
    
    if (found) {
        return true;
    }
```

rafaelroquetto (commented):
First pass - will revisit

does .codex need to be committed?

bpf/common/http_types.h
grcevski: I think that discussion was mostly around not serialising the bit fields and I get that, but this struct is just to pass data over to the protocol detector and while the flags approach works, it's a lot more code and it will make things even harder to read. Without the helpers I think parsing the flags would be even more error prone.

rafaelroquetto: as discussed offline, in this case bitfields are harmless and likely a simpler solution than flags (as there is no serialisation involved or anything that assumes the struct layout)

bpf/generictracer/protocol_http.h
grcevski: Yeah I've been struggling with the naming of this field. But essentially, the problem is the following:
1. Generic tracer uses the PID filter, so when event makes it to userspace it also checks the userspace filter map, apart from us checking the is_valid_pid. We did this as a safety case for Go programs accidentally being picked up in the pid filter, since it's using a bloom filter and it can have false positives.
2. So now we are sending TCP events and HTTP events from the Go program for real, selectively, those that we cannot handle.
3. I need to signal to the userspace to let this event go, so that it will get passed up the pipeline.

bpf/generictracer/protocol_http.h
grcevski: I agree.

bpf/gotracer/maps/handled_by_go.h
grcevski: this will get changed now, since I'm switching to full connection.

bpf/gotracer/maps/handled_by_go.h
grcevski: good idea.

bpf/gotracer/go_net.c
grcevski: this changes too now with connection.

bpf/gotracer/go_net.c
grcevski: I added the http2 delete to be safe.

grcevski:
> First pass - will revisit
> 
> does .codex need to be committed?

No, thanks for noticing this. it keeps making that empty file all the time in my project.

bpf/gotracer/go_net.c
grcevski: Done

internal/test/integration/components/gogeneric/main.go
grcevski: Fixed.

bpf/common/http_types.h
```diff
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include "common/lw_thread.h"
```
rafaelroquetto: vscode

bpf/generictracer/generictracer.c
```diff
@@ -4,6 +4,7 @@
 //go:build obi_bpf_ignore
 #include "iter_tcp.c"
 #include "java_tls.c"
+#include "protocol_handler.c"
```
rafaelroquetto: pedantic nit: alphabetic order to comply with surrounding code

bpf/generictracer/protocol_handler.c
```diff
@@ -0,0 +1,162 @@
...
+
+    bpf_dbg_printk("=== kprobe buf=[%s], pid=%d, len=%d ===",
+                   args->small_buf,
+                   args->pid_conn.pid,
+                   args->bytes_len);
+
+    if (args->protocols.http && is_http(args->small_buf, MIN_HTTP_SIZE, &args->packet_type)) {
+        bpf_tail_call(ctx, &jump_table, k_tail_protocol_http);
+    } else if (is_http2_or_grpc(args->small_buf, MIN_HTTP2_SIZE) &&
+               (args->protocol_type != k_protocol_type_http)) {
```
rafaelroquetto: is it possible to flip this to:

```suggestion
    } else if (args->protocol_type != k_protocol_type_http && is_http2_or_grpc(args->small_buf, MIN_HTTP2_SIZE))) {
```

so that we don't waste time on is_http2_or_grpc unnecessarily?

bpf/generictracer/protocol_http.h
```diff
@@ -458,11 +464,13 @@ static __always_inline void process_http_response(http_info_t *info, const unsig
...
                                                  pid_connection_info_t *pid_conn,
                                                  http_info_t *info,
-                                                 int orig_len) {
+                                                 int orig_len,
+                                                 lw_thread_t lw_thread) {
     process_http_response(info, small_buf);
     cleanup_http_request_data(pid_conn, info);
 
-    if (high_request_volume) {
+    // Generic Go events cannot be delayed for now since we don't probe on net_close
```
rafaelroquetto, line 472: so does that mean lw_thread != lw_thread_none -> go event?

bpf/gotracer/maps/handled_by_go.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include <common/connection_info.h>
+#include <common/go_addr_key.h>
+#include <common/map_sizing.h>
+#include <common/pin_internal.h>
+#include <common/tp_info.h>
+
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, connection_info_t); // key: the connection info
```
rafaelroquetto:
```suggestion
    __type(key, connection_info_t);
```

the comment is very redundant

bpf/gotracer/maps/handled_by_go.h
```diff
@@ -0,0 +1,47 @@
...
+    __type(value, bool);            // value: true
```
rafaelroquetto:
```suggestion
    __type(value, bool);            // placeholder, we check for value presence
```

bpf/gotracer/maps/handled_by_go.h
```diff
@@ -0,0 +1,47 @@
...
+    __uint(max_entries, MAX_CONCURRENT_REQUESTS);
+    __uint(pinning, OBI_PIN_INTERNAL);
+} handled_by_go_conn SEC(".maps");
+
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, go_addr_key_t); // key: goroutine
+    __type(value, bool);        // value: true
```
rafaelroquetto: ditto

bpf/gotracer/go_net.c
```diff
@@ -20,13 +20,69 @@
...
+static __always_inline bool already_handled_request_sorted(const connection_info_t *conn,
+                                                           const go_addr_key_t *goaddr) {
+    if (conn) {
+        bool *found = bpf_map_lookup_elem(&handled_by_go_conn, conn);
+        if (found) {
+            return true;
+        }
+    }
+    if (goaddr) {
+        bool found = bpf_map_lookup_elem(&handled_by_go, goaddr);
```
rafaelroquetto:
```suggestion
        const bool *found = bpf_map_lookup_elem(&handled_by_go, goaddr);
```

missing pointer, it's probably implicitly casting to bool

bpf/gotracer/go_net.c
```diff
@@ -20,13 +20,69 @@
...
+#include <gotracer/types/net_args.h>
+
+#include <pid/pid_helpers.h>
+
+#include <shared/obi_ctx.h>
+
+static __always_inline bool already_handled_request_sorted(const connection_info_t *conn,
+                                                           const go_addr_key_t *goaddr) {
+    if (conn) {
+        bool *found = bpf_map_lookup_elem(&handled_by_go_conn, conn);
```
rafaelroquetto:
```suggestion
        const bool *found = bpf_map_lookup_elem(&handled_by_go_conn, conn);
```

bpf/gotracer/go_nethttp.c
```diff
@@ -15,6 +15,7 @@
 
 //go:build obi_bpf_ignore
 
+#include "common/connection_info.h"
```
rafaelroquetto: vscode shenanigans

bpf/gotracer/gotracer.c
```diff
@@ -26,3 +26,6 @@
 #include "go_sarama.c"
 #include "go_sdk.c"
 #include "go_mongo.c"
+#include "generictracer/protocol_handler.c"
```
rafaelroquetto, line 30: this is not optimal, as we try to keep tracers self-contained - ideally, we'd like to move protocol_handler.c to a common location, but I see it includes a bunch of generictracer headers which are now transiently included into gotracer - so I guess that entails a larger refactoring which may be outside the scope of this PR (you be the judge, perhaps it's not as hard as thought but I'm assuming it is).


Assuming we won't refactor it as part of this PR: can we add a `//FIXME - move common code to common location` ?

I can help with it later.

rafaelroquetto (commented):
I gotta be honest, I don't know enough about the go stuff to be able to assert it, but otherwise it LGTM - just minor nits

bpf/generictracer/protocol_http.h
grcevski: Yes, but I didn't want to call it Go, since we may add a new path in the future, e.g. some other runtime.

bpf/generictracer/protocol_handler.c
grcevski: sure, good idea.

rafaelroquetto (approved):

bpf/gotracer/go_net.c
```diff
@@ -81,11 +150,130 @@ int obi_uprobe_netFdRead(struct pt_regs *ctx) {
...
+    bpf_dbg_printk("buf=%llx, len=%d === ", buf, len);
+    if (buf && len > 0) {
+        dbg_print_http_connection_info(&net_ptr->p_conn.conn);
+
+        u16 orig_dport = net_ptr->p_conn.conn.d_port;
+        sort_connection_info(&net_ptr->p_conn.conn);
+
+        dbg_print_http_connection_info(&net_ptr->p_conn.conn);
+
+        bpf_map_delete_elem(&ongoing_fd_reads, &g_key);
```
MrAlias, line 212: Should this be delayed until after `handle_light_weight_thread_buf` is called with `&net_ptr->p_conn`? Isn't that pointer only valid while the map element remains valid?

bpf/gotracer/go_net.c
```diff
@@ -81,11 +150,130 @@ int obi_uprobe_netFdRead(struct pt_regs *ctx) {
...
     }
 
+    const u64 id = bpf_get_current_pid_tgid();
+
+    void *fd_ptr = GO_PARAM1(ctx);
+    void *byte_addr = GO_PARAM2(ctx);
+    net_args_t net_args = {
+        .byte_ptr = (u64)byte_addr,
+    };
+    get_conn_info_from_fd(fd_ptr, &net_args.p_conn.conn, false);
```
MrAlias, line 168: Should below be guarded from executing when this fails?

bpf/gotracer/maps/handled_by_go.h
```diff
@@ -0,0 +1,47 @@
...
+    __type(value, bool); // placeholder, we check if it exists
+    __uint(max_entries, MAX_CONCURRENT_REQUESTS);
+    __uint(pinning, OBI_PIN_INTERNAL);
+} handled_by_go SEC(".maps");
+
+static __always_inline void store_go_handled_connection_info(const connection_info_t *conn) {
+    if (conn) {
+        connection_info_t sorted_conn = *conn;
+        sort_connection_info(&sorted_conn);
+        bpf_map_update_elem(&handled_by_go_conn, &sorted_conn, &(bool){true}, BPF_ANY);
```
MrAlias, line 35: Is this ever cleared?

grcevski: We don't need to IMO, since it's very unlikely that two services are talking on the same connection with uprobe instrumented code and then they switch to non instrumented. Previously when I had just the ports it was possible to get a conflict, but not with the ip addresses included it's highly unlikely. If you are concerned I can follow-up with a PR that has probe on close and clean it up? Sorry I merged too soon.

bpf/gotracer/go_net.c
grcevski: Good idea, I'll follow-up with a PR.

bpf/gotracer/go_net.c
grcevski: We can't delete after since `handle_light_weight_thread_buf` doesn't return, it tail calls, so this is our last chance.

bpf/gotracer/maps/handled_by_go.h
MrAlias: It is an edge-case, indeed not common. I'll leave it up to you. If you think the cost of another uprobe is not outweigh the chance of this edge case occurring, we can leave it the way it is. :+1:

Issue #781, opened 2025-10-20
Docker image teardown fails on kernel 6.14


mmat11:
CI jobs failed with:

```
Container integration-obi-1  Error while Stopping
Error response from daemon: cannot stop container:
69332d321a802588509366dc706a460dddb490d0df437b112171f9d375997ad6: container 69332d321a80 PID 32462 is zombie and can not be killed.
Use the --init option when creating containers to run an init inside the container that forwards signals and reaps processes
```

A wrapper script which waits for the processed to end is being used as a workaround

NimrodAvni78:
@mariomac did #1782 fix this
can we close?

mariomac:
@NimrodAvni78 let's close!

PR #1782, opened 2026-04-09
Fix OBI being blocked on shutdown

mariomac:
## Summary

OBI is taking long to exit, mainly because the OpenTelemetry metrics exporters got blocked during their Shutdown. This PR takes a best-effort approach to OTEL metrics shutdown: we try to shut it down in a background goroutine, as it is not a critical operation (the worst case is that we loose few last data points in an OBI instance we are anyway shutting down).

The approach that was taken to fix it in the integration tests was to wrap the OBI process into a shell script, but that was hiding the blocking in production, and its real cause.

It also broke the coverage reporting of the integration tests, as the OBI process was killed before it had time to exit.

This should also fix this issue: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/781

## Validation


grcevski (approved):
LGTM! Nice! I was seeing these a lot on my new hardware.

mmat11 (approved):
thanks for fixing this!

PR #1784, opened 2026-04-09
Add support for JSON-RPC protocol in all languages

### LINKED ISSUE #1477: Support JSON-RPC in languages other than Go
This would allow instrumenting MCP clients and servers: https://modelcontextprotocol.info/docs/concepts/transports/#message-format.

Related: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/368

I haven't dug into how feasible this is to support, although jsonrpc.org says "It is designed to be simple!", so how hard can it be... I don't have cycles to look into this myself, but i'm opening the issue to see if this is something anyone else has looked at before, or is interested in.

NimrodAvni78:
## Summary

Resolves: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1477
Adds parsing support for JSON-RPC on top of http
next step is to add support for MCP protocol on top of JSON-RPC
follows [JSON-RPC semantic conventions](https://opentelemetry.io/docs/specs/semconv/rpc/json-rpc/)

## Validation


pkg/ebpf/common/http/jsonrpc.go
```diff
@@ -0,0 +1,192 @@
...
+	if resp != nil && resp.Body != nil {
+		respB, err := getResponseBody(resp)
+		if err == nil {
+			parseJSONRPCResponse(respB, result)
+		}
+	}
+
+	baseSpan.SubType = request.HTTPSubtypeJSONRPC
+	baseSpan.JSONRPC = result
+
```
grcevski, line 89: I think we need to set baseSpan.Method to the rpcReq.Method, otherwise the span name will be POST no? Maybe it's done differently than the Go specific implementation?

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -306,6 +306,26 @@ var (
...
+	rpc := span.JSONRPC
+	attrs := []attribute.KeyValue{
+		semconv.RPCSystemJSONRPC,
+		semconv.RPCMethod(rpc.Method),
+		attribute.String("jsonrpc.protocol.version", rpc.Version),
+	}
+	if rpc.RequestID != "" {
+		attrs = append(attrs, attribute.String("jsonrpc.request.id", rpc.RequestID))
+	}
+	if rpc.ErrorCode != 0 {
```
grcevski, line 322: I see you collect the error message when parsing the event, I think it would be good to add it here in the attributes if ErrorCode != 0.

grcevski (commented):
This looks great, I think we need to ensure the rpc method is in the HTTP method to match what Go does. Perhaps I misunderstand and this is done already.

pkg/export/otel/tracesgen/tracesgen.go
NimrodAvni78: im adding it to the status in `SpanStatusMessage()` function

pkg/ebpf/common/http/jsonrpc.go
NimrodAvni78: my bad🤦
fixing

pkg/appolly/app/request/span.go
```diff
@@ -760,6 +770,12 @@ func SpanStatusMessage(span *Span) string {
 		if span.SubType == HTTPSubtypeSQLPP && span.Status != 0 && span.DBError.Description != "" {
 			return span.DBError.Description
 		}
+	case EventTypeHTTP:
+		// handled below
+	}
+	// JSON-RPC spans can be either server (EventTypeHTTP) or client (EventTypeHTTPClient)
```
mmat11: can client span have an error message?

mmat11 (approved):

NimrodAvni78:
@grcevski open https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1799
will fix it separately

pkg/appolly/app/request/span.go
NimrodAvni78: yes
will fix

rafaelroquetto (approved):
LGTM pending other comments and conflicts 👍

grcevski (approved):
LGTM! I'm good with merging this and I'm good with making the span name consistent in a follow-up. I think we should also make sure these events end up as RPC metrics, hot HTTP, correct? I don't see a metrics test, but I imagine these should go as RPC.

NimrodAvni78:
@grcevski fixed
btw i think its the same issue in go, we emit http metrics instead of go rpc metrics
https://github.com/coralogix/opentelemetry-ebpf-instrumentation/blob/9026e282e8b87557d1fc3b2fc14c2273defb4f57/internal/test/integration/red_test.go#L316
iv'e created a [separate issue](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1821) to solve in another PR

PR #1792, opened 2026-04-10
[internal/tools] use native Go tool approach

florianl:
## Summary

Migrate tooling from `internal/tools/tools.go` to a more Go generic approach so that they can be used with `go tool ...`

If this change is accepted, I will look into migrating the rest of the tools.

## Validation

- [ ] ~~If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md)~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,14 +106,11 @@ $(TOOLS)/setup-envtest: PACKAGE=sigs.k8s.io/controller-runtime/tools/setup-envte
 KIND ?= $(TOOLS)/kind
 $(TOOLS)/kind: PACKAGE=sigs.k8s.io/kind
 
-GOLICENSES = $(TOOLS)/go-licenses
-$(TOOLS)/go-licenses: PACKAGE=github.com/google/go-licenses/v2
-
 MULTIMOD = $(TOOLS)/multimod
 $(TOOLS)/multimod: PACKAGE=go.opentelemetry.io/build-tools/multimod
 
 .PHONY: tools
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -148,9 +143,9 @@ prereqs: install-hooks
 	mkdir -p $(TEST_OUTPUT)/run
 
 .PHONY: fmt
-fmt: $(GOLANGCI_LINT)
+fmt:
 	@echo "### Formatting code and fixing imports"
-	$(GOLANGCI_LINT) fmt
+	go tool $(TOOLS_MODFILE) golangci-lint fmt
```
mariomac, line 133: So let me see if I understand: with your change, it still won't run any system-global golangci-lint but will use the version that is listed in the local TOOLS_MODFILE, right?

In other words, can I still have in my laptop multiple Go projects using different versions of golangci-lint without interfering each other?

florianl: > with your change, it still won't run any system-global golangci-lint but will use the version that is listed in the local TOOLS_MODFILE, right?

Yes - this is correct. the `go tool ...` command doesn't know about your other versions or system-globals of golangci-lint.

> can I still have in my laptop multiple Go projects using different versions of golangci-lint without interfering each other?

Yes you can.

mariomac (approved):
Amazing!

grcevski (approved):
LGTM!

mmat11 (approved):
lgtm!

MrAlias (approved):

PR #1806, opened 2026-04-13
Add `BatchMaxSize` and `QueueSize`

pinoOgni:
## Summary

This PR adds `BatchMaxSize` and `QueueSize` fields. Basically we were doing `batchCfg.MaxSize = int64(cfg.MaxQueueSize)` which it's misleading and now we are doing `batchCfg.MaxSize = int64(cfg.BatchMaxSize)`.

## Validation

### DIFF
--- a/pkg/export/otel/otelcfg/config_traces.go
+++ b/pkg/export/otel/otelcfg/config_traces.go
@@ -40,16 +40,7 @@ type TracesConfig struct {
 
 	// BatchMaxSize is the maximum number of spans that the batcher will accumulate
 	// before flushing a batch to the sending queue.
-	// It also accepts OTEL_EBPF_OTLP_TRACES_MAX_QUEUE_SIZE for backwards compatibility:
-	// that env var was historically the batch max size despite its misleading name.
-	BatchMaxSize int `yaml:"batch_max_size" env:"OTEL_EBPF_OTLP_TRACES_BATCH_MAX_SIZE,expand" envDefault:"${OTEL_EBPF_OTLP_TRACES_MAX_QUEUE_SIZE}"`
-
-	// DeprecatedMaxQueueSize accepts the legacy `max_queue_size` YAML key for
-	// backwards compatibility. It is copied into BatchMaxSize by
-	// NormalizeQueueConfig with a deprecation warning. Do not read this field
-	// directly; use BatchMaxSize. The env var equivalent is handled
-	// transparently via the envDefault expansion on BatchMaxSize above.
-	DeprecatedMaxQueueSize int `yaml:"max_queue_size" env:"-"`
+	BatchMaxSize int `yaml:"batch_max_size" env:"OTEL_EBPF_OTLP_TRACES_BATCH_MAX_SIZE"`
 
 	// QueueSize is the maximum number of spans that the sending queue will hold
 	// before applying back-pressure. It must be >= BatchMaxSize, otherwise the
@@ -84,11 +75,6 @@ type TracesConfig struct {
 }
 
 func (m *TracesConfig) NormalizeQueueConfig() error {
-	if m.BatchMaxSize == 0 && m.DeprecatedMaxQueueSize > 0 {
-		tlog().Warn("traces.max_queue_size is deprecated, use traces.batch_max_size instead",
-			"value", m.DeprecatedMaxQueueSize)
-		m.BatchMaxSize = m.DeprecatedMaxQueueSize
-	}
 	if m.QueueSize == 0 && m.BatchMaxSize > 0 {
 		// Queue capacity must be at least 2x max batch size to prevent "element size too large"
 		// errors and permanent data loss. We use a 4x multiplier to provide headroom for
@@ -108,8 +94,7 @@ func (m *TracesConfig) NormalizeQueueConfig() error {
 
 func (m TracesConfig) MarshalYAML() (any, error) {
 	omit := map[string]struct{}{
-		"endpoint":       {},
-		"max_queue_size": {},
+		"endpoint": {},
 	}
 	return omitFieldsForYAML(m, omit), nil
 }
--- a/pkg/export/otel/otelcfg/config_traces.go
+++ b/pkg/export/otel/otelcfg/config_traces.go
@@ -38,9 +38,26 @@ type TracesConfig struct {
 
 	SamplerConfig services.SamplerConfig `yaml:"sampler"`
 
-	// Configuration options below this line will remain undocumented at the moment,
-	// but can be useful for performance-tuning of some customers.
-	MaxQueueSize int           `yaml:"max_queue_size" env:"OTEL_EBPF_OTLP_TRACES_MAX_QUEUE_SIZE"`
+	// BatchMaxSize is the maximum number of spans that the batcher will accumulate
+	// before flushing a batch to the sending queue.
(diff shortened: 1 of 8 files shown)


pkg/export/otel/otelcfg/config_traces.go
```diff
@@ -38,9 +38,26 @@ type TracesConfig struct {
...
+	// It also accepts OTEL_EBPF_OTLP_TRACES_MAX_QUEUE_SIZE for backwards compatibility:
+	// that env var was historically the batch max size despite its misleading name.
+	BatchMaxSize int `yaml:"batch_max_size" env:"OTEL_EBPF_OTLP_TRACES_BATCH_MAX_SIZE,expand" envDefault:"${OTEL_EBPF_OTLP_TRACES_MAX_QUEUE_SIZE}"`
+
+	// DeprecatedMaxQueueSize accepts the legacy `max_queue_size` YAML key for
+	// backwards compatibility. It is copied into BatchMaxSize by
+	// NormalizeQueueConfig with a deprecation warning. Do not read this field
+	// directly; use BatchMaxSize. The env var equivalent is handled
+	// transparently via the envDefault expansion on BatchMaxSize above.
+	DeprecatedMaxQueueSize int `yaml:"max_queue_size" env:"-"`
```
mariomac: Since properties in this section are hidden/undocummented or just aimed for development fine-tuning, I would personally not care of providing backwards compatibility.

pinoOgni: sure, fine by me

mariomac (approved):

rafaelroquetto (commented):
From the naked eye code-wise it LGTM - but I saw that copilot many some suggestions that appear relevant IMHO.

I'm still a bit confused on how one is supposed to use these 2 settings together, since they are related. We have defaults, but what would be the guidelines to tune them up? (non-rhetorical question/just trying to better understand this corner of the code)

pinoOgni:
> From the naked eye code-wise it LGTM - but I saw that copilot many some suggestions that appear relevant IMHO.

Yes, I made some changes to remove `DeprecatedMaxQueueSize`, and I think I forgot to update something. I'll check it.

> I'm still a bit confused on how one is supposed to use these 2 settings together, since they are related. We have defaults, but what would be the guidelines to tune them up? (non-rhetorical question/just trying to better understand this corner of the code)

A few days ago we had a bug fixed here https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1761. From there, I noticed a naming issue, which is why I introduced these two.
- `BatchMaxSize` is the size of a batch
- `QueueSize` is the size of the queue. It must be greater or equal than `BatchMaxSize` and it is a hard constraint, otherwise the queue rejects every batch. The 4x default gives headroom for transient spikes, the minimum is 2x.

IMO, most users only need to touch `BatchMaxSize`.

pkg/export/otel/traces.go
```diff
@@ -301,32 +301,29 @@ func getTracesExporter(ctx context.Context, cfg otelcfg.TracesConfig, im imetric
...
-		if minQueue := int64(cfg.MaxQueueSize) * 4; queueConfig.QueueSize < minQueue {
-			queueConfig.QueueSize = minQueue
-		}
+		batchCfg.MaxSize = int64(cfg.BatchMaxSize)
 	}
 	if cfg.BatchTimeout > 0 {
 		batchSet = true
 		batchCfg.FlushTimeout = cfg.BatchTimeout
-		batchCfg.MinSize = int64(cfg.MaxQueueSize)
+		batchCfg.MinSize = int64(cfg.BatchMaxSize)
```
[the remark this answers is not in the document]
pinoOgni: The code sets `MinSize` only when `BatchTimeout > 0`, which should be the right guard. So, `MinSize = BatchMaxSize` means to no eagerly flush partial batches as data trickles in but wait to accumulate up to `BatchMaxSize` and if `FlushTimeout` fires first, flush whatever we have. This is my idea.

pkg/export/otel/otelcfg/config_traces.go
```diff
@@ -66,6 +74,24 @@ type TracesConfig struct {
 	InjectHeaders func(dst map[string]string) `yaml:"-" env:"-"`
 }
 
+func (m *TracesConfig) NormalizeQueueConfig() error {
+	if m.QueueSize == 0 && m.BatchMaxSize > 0 {
+		// Queue capacity must be at least 2x max batch size to prevent "element size too large"
+		// errors and permanent data loss. We use a 4x multiplier to provide headroom for
+		// transient latency spikes, ensuring brief collector slowdowns don't immediately
+		// back-pressure the eBPF reader.
```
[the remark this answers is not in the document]
pinoOgni: The queue needs to hold at least one full batch plus have room for the next one being assembled, 2x is the correct minimum. I missed `2*m.BatchMaxSize` in the code.

rafaelroquetto (approved):

pkg/export/otel/otelcfg/config_traces.go
```diff
@@ -38,9 +38,17 @@ type TracesConfig struct {
 
 	SamplerConfig services.SamplerConfig `yaml:"sampler"`
 
-	// Configuration options below this line will remain undocumented at the moment,
-	// but can be useful for performance-tuning of some customers.
-	MaxQueueSize int           `yaml:"max_queue_size" env:"OTEL_EBPF_OTLP_TRACES_MAX_QUEUE_SIZE"`
```
mmat11, line 43: I don't think this can just be removed, it's a breaking change

pinoOgni: Initially, `OTEL_EBPF_OTLP_TRACES_MAX_QUEUE_SIZE` was added like that:
```
BatchMaxSize int `yaml:"batch_max_size" env:"OTEL_EBPF_OTLP_TRACES_BATCH_MAX_SIZE,expand" envDefault:"${OTEL_EBPF_OTLP_TRACES_MAX_QUEUE_SIZE}"`
```
to support backward compatibility. After this [comment](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1806#discussion_r3072317829), I removed it. 
wdyt?

pinoOgni: Now I'm start to think that the comment was not related to `OTEL_EBPF_OTLP_TRACES_BATCH_MAX_SIZE` env. I can add it again.

mmat11: I hadn't read @mariomac 's comment, if having a breaking change in this section is ok, then fine by me

mariomac: I think we can just properly document it in the release notes, as this was an undocumented performance-tunning feature:
```
	// Configuration options below this line will remain undocumented at the moment,
	// but can be useful for performance-tuning of some customers.
```

pinoOgni: The code is ready on my end. 
If the plan is to document this in the upcoming release notes, feel free to merge whenever you're ready.

mmat11:
@pinoOgni there's a conflict which needs to be resolved, ready to merge after that

pinoOgni:
> @pinoOgni there's a conflict which needs to be resolved, ready to merge after that

Done.

mmat11 (approved):

PR #1808, opened 2026-04-13
Add coverage of privileged unit tests

mariomac:
## Summary

Uses a standard Go test selection mechanism to run/not-run privileged unit tests (go:build labels instead of runtime skipping via env vars).

Also amends some GH actions to run privileged and BPF verifier tests, and upload coverage data.

## Validation


rafaelroquetto (approved):

.github/workflows/pull_request.yml
```diff
@@ -282,4 +282,17 @@ jobs:
 
       - name: Run BPF verifier tests (host kernel)
         if: ${{ !matrix.use_vm }}
-        run: sudo env PATH=$PATH PRIVILEGED_TESTS=true go test -v -count=1 ./pkg/internal/ebpf/verifier/...
+        run: sudo env PATH=$PATH go test -v -count=1 -tags=bpf_verifier_tests ./pkg/internal/ebpf/verifier/... -coverprofile $(TEST_OUTPUT)/cover.all.txt
```
florianl: Did you consider something like this?
```suggestion
        run: go test -exec=sudo -v -count=1 -tags=bpf_verifier_tests ./pkg/internal/ebpf/verifier/... -coverprofile $(TEST_OUTPUT)/cover.all.txt
```

mariomac: Done. Thanks!

pkg/internal/ebpf/tcmanager/tcmanager_privileged_test.go
```diff
@@ -0,0 +1,339 @@
...
+SEC("tc_egress")
+int obi_egress(struct __sk_buff *skb) {
+    return 0;
+}
+
+clang -c -target bpf -O2 -o tc_program.o tc_program.c
+xxd --include tc_program.o > tc_program.c
+*/
+
+var tcProgram = []byte{
```
florianl, line 266: It might be simpler to handcraft a bpf program - see https://github.com/florianl/go-tc/blob/8fa73f9a1b38f5a099e072305d29539908b748e5/example_ebpf_test.go#L92-L107 as an example

florianl (approved):

pkg/internal/ebpf/tcmanager/tcmanager_privileged_test.go
mariomac: Ohh good point! This is just a moved test from another file so changing it might be out of the scope, but let me try your approach in another PR!

PR #1811, opened 2026-04-13
Context propagation support for Go generic requests

grcevski:
## Summary

This PR is a follow-up from https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1766. In the previous PR we added the ability to process Go protocol buffers using the generic handler, this one adds context propagation support.

Namely, we register the required information in `go_trace_map` and then we use the loop the gotracer uses to find parent goroutine with a parent server handler.

The tests I added test a combination of standard Go and non-standard Go, HTTP and TCP.
 
I also removed the goroutine checking code I had added in the previous PR, as I wrote these new tests I realised they were causing issues. Namely, you can have the same goroutine handle two different connections, e.g. standard Go HTTP server making a client call with non-standard Go library. In this case the goroutine is the same, so excluding buffers based on goroutines being seen as handling standard Go is not possible. Now we rely on the connection information.

## Validation

- [X] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](../SUPPORT_MATRIX.md) as needed.


mariomac (approved):
👏🏻

PR #1820, opened 2026-04-14
fix(gemini): fix GenAI support for Vertex AI

NameHaibinZhang:
## Summary

Enhance Gemini detection to support the googleapis/go-genai Go SDK,
which calls both the Gemini Developer API and Vertex AI backends.

Previously, Gemini requests were identified solely by the
X-Gemini-Service-Tier response header. The go-genai library, especially
when using the Vertex AI backend, may not return this header.

Changes:
- Add URL-based fallback detection for known Gemini hosts
  (generativelanguage.googleapis.com, *-aiplatform.googleapis.com)
- Extract operation name from URL path (e.g. generateContent →
  generate_content, embedContent → embed_content) instead of
  hardcoding "generate_content"
- Add Operation field to VendorGemini and OperationName() accessor
- Update tracesgen to use dynamic operation name

## Validation

- [x] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](../SUPPORT_MATRIX.md) as needed.

### DIFF
--- a/pkg/ebpf/common/http/gemini.go
+++ b/pkg/ebpf/common/http/gemini.go
@@ -22,8 +22,8 @@ const geminiModelPrefix = "/models/"
 // geminiHostPattern pairs a known hostname suffix with the URL path segment
 // that must be present for the request to be considered a Gemini API call.
 type geminiHostPattern struct {
-	hostSuffix    string
-	requiredPath  string
+	hostSuffix   string
+	requiredPath string
 }
 
 // geminiHostPatterns lists known Gemini API hosts and their required path
--- a/pkg/ebpf/common/http/gemini.go
+++ b/pkg/ebpf/common/http/gemini.go
@@ -156,7 +156,7 @@ func extractGeminiOperation(req *http.Request) string {
 	}
 	after := path[idx+len(geminiModelPrefix):]
 	colonIdx := strings.Index(after, ":")
-	if colonIdx < 0 {
+	if colonIdx < 0 || colonIdx+1 >= len(after) {
 		return request.DefaultGeminiOperation
 	}
 	return camelToSnake(after[colonIdx+1:])
--- a/pkg/ebpf/common/http/gemini.go
+++ b/pkg/ebpf/common/http/gemini.go
@@ -19,10 +19,20 @@ import (
 // in both the Gemini Developer API and Vertex AI URL layouts.
 const geminiModelPrefix = "/models/"
 
-// geminiHosts lists known hostnames used by the Gemini API and Vertex AI.
-var geminiHosts = []string{
-	"generativelanguage.googleapis.com",
-	"aiplatform.googleapis.com",
+// geminiHostPattern pairs a known hostname suffix with the URL path segment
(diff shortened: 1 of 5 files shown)


pkg/ebpf/common/http/gemini.go
```diff
@@ -14,12 +14,45 @@ import (
...
+	}
+
+	host := req.URL.Host
+	if host == "" {
+		host = req.Host
+	}
+
+	for _, h := range geminiHosts {
+		if host == h || strings.HasSuffix(host, "."+h) ||
+			strings.HasSuffix(host, "-"+h) {
```
mariomac: Wouldn't just:
```go
if strings.HasSuffix(host, h) {
```
suffice?

mariomac (commented):
Amazing! Thank you for your contribution. I have a small comment.

pkg/ebpf/common/http/gemini.go
NameHaibinZhang: Good point! strings.HasSuffix(host, h) does cover all the cases we need (exact match, subdomain with ., and region prefix with -). Simplified.

NameHaibinZhang:
@mariomac @rafaelroquetto  All comments have been resolved, thks.

pkg/ebpf/common/http/gemini.go
```diff
@@ -47,34 +107,73 @@ func GeminiSpan(baseSpan *request.Span, req *http.Request, resp *http.Response)
...
+	if req == nil || req.URL == nil {
+		return request.DefaultGeminiOperation
+	}
+	path := req.URL.Path
+	idx := strings.Index(path, geminiModelPrefix)
+	if idx < 0 {
+		return request.DefaultGeminiOperation
+	}
+	after := path[idx+len(geminiModelPrefix):]
+	colonIdx := strings.Index(after, ":")
```
grcevski, line 158: Can we please add check here to see if the ":" char isn't the last in the string? If it is we'll get a panic with the access `after[colonIdx+1:]`. A test for this as well please.

grcevski (commented):
Looks really great, thanks for adding this, I found one minor potential problem. Once we fix this we can merge.

pkg/ebpf/common/http/gemini.go
NameHaibinZhang: Good catch! Added a boundary check colonIdx+1 >= len(after) so that a trailing colon (e.g. /models/gemini-2.0-flash:) falls back to the default operation name instead of producing an empty string. Also added a "trailing colon with no operation" test case to cover this.

NameHaibinZhang:
@grcevski  all comment is fix, please review,thks

mariomac (approved):
LGTM! Thanks for addressing the comments.

Before merging, I'll wait for the rest of commenters to give their OK.

mmat11 (approved):
lgtm!

grcevski (approved):
LGTM!

PR #1825, opened 2026-04-14
Add TCP failed connections metric

pinoOgni:
## Summary

This PR adds TCP failed connections metric to Statsolly. 

Here's an example metric:
```
obi_stat_tcp_failed_connections{dst_address="192.168.x.x",dst_name="frontend-proxy-59cbb58fcd-fhxnj",dst_port="37868",dst_zone="***",k8s_cluster_name="",k8s_dst_name="frontend-proxy-59cbb58fcd-fhxnj",k8s_dst_namespace="default",k8s_dst_node_ip="192.168.x.x.",k8s_dst_node_name="node1",k8s_dst_owner_name="frontend-proxy",k8s_dst_owner_type="Deployment",k8s_dst_type="Pod",k8s_src_name="frontend-64765f9445-sjqrq",k8s_src_namespace="default",k8s_src_node_ip="192.168.x.x.",k8s_src_node_name="i-054820b1045823806",k8s_src_owner_name="frontend",k8s_src_owner_type="Deployment",k8s_src_type="Pod",obi_ip="192.168.x.x.",reason="unknown",service_name="frontend",service_namespace="opentelemetry-demo",service_peer_name="frontend-proxy",service_peer_namespace="opentelemetry-demo",src_address="192.168.x.x",src_name="frontend-64765f9445-sjqrq",src_port="0",src_zone="**"} 1
```
Note: The "reason" label is disabled by default.

To calculate this metric, `tracepoint/sock/inet_sock_set_state` was used. Therefore, I modified the Statsolly agent (stats_tracer) to handle tracepoints and also added support for multiple kprobes for the future.

## Validation

- [x] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](../SUPPORT_MATRIX.md) as needed.

### DIFF
--- a/bpf/statsolly/tp_tcp.c
+++ b/bpf/statsolly/tp_tcp.c
@@ -15,27 +15,43 @@
 #include <statsolly/types.h>
 #include <statsolly/maps/stats_events.h>
 
+#ifndef ECONNREFUSED
+#define ECONNREFUSED 111
+#endif
+#ifndef ECONNRESET
+#define ECONNRESET 104
+#endif
+#ifndef ETIMEDOUT
+#define ETIMEDOUT 110
+#endif
+#ifndef EHOSTUNREACH
+#define EHOSTUNREACH 113
+#endif
+#ifndef ENETUNREACH
+#define ENETUNREACH 101
+#endif
+
 enum tcp_fail_reason {
-    reason_unknown = 0,            // sk_err was 0, cause not determined
-    reason_connection_refused = 1, // ECONNREFUSED (111)
-    reason_connection_reset = 2,   // ECONNRESET (104)
-    reason_timed_out = 3,          // ETIMEDOUT (110)
-    reason_host_unreachable = 4,   // EHOSTUNREACH (113)
-    reason_net_unreachable = 5,    // ENETUNREACH (101)
-    reason_other = 255,            // anything else
+    reason_unknown = 0,
+    reason_connection_refused = 1,
+    reason_connection_reset = 2,
+    reason_timed_out = 3,
+    reason_host_unreachable = 4,
+    reason_net_unreachable = 5,
+    reason_other = 255,
 };
 
-static __always_inline u8 sk_err_to_reason(int err) {
+static __always_inline u8 sk_err_to_reason(const int err) {
     switch (err) {
-    case 111:
+    case ECONNREFUSED:
         return reason_connection_refused;
-    case 104:
+    case ECONNRESET:
         return reason_connection_reset;
-    case 110:
+    case ETIMEDOUT:
         return reason_timed_out;
-    case 113:
+    case EHOSTUNREACH:
         return reason_host_unreachable;
-    case 101:
+    case ENETUNREACH:
         return reason_net_unreachable;
     case 0:
         return reason_unknown;
@@ -76,8 +92,8 @@ int obi_tracepoint_inet_sock_set_state(struct trace_event_raw_inet_sock_set_stat
         return 0;
     }
 
-    int err = BPF_CORE_READ(sk, sk_err);
-    u8 reason = sk_err_to_reason(err);
+    const int err = BPF_CORE_READ(sk, sk_err);
+    const u8 reason = sk_err_to_reason(err);
 
     bpf_d_printk("tcp failed: oldstate=%d, reason=%d, s_port=%d, d_port=%d",
                  args->oldstate,
--- a/bpf/statsolly/tp_tcp.c
+++ b/bpf/statsolly/tp_tcp.c
@@ -1,0 +1,98 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build obi_bpf_ignore
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_helpers.h>
+#include <bpfcore/bpf_tracing.h>
+#include <bpfcore/bpf_core_read.h>
+
+#include <common/connection_info.h>
+#include <common/sockaddr.h>
+
+#include <logger/bpf_dbg.h>
+
+#include <statsolly/types.h>
+#include <statsolly/maps/stats_events.h>
+
+enum tcp_fail_reason {
+    reason_unknown = 0,            // sk_err was 0, cause not determined
+    reason_connection_refused = 1, // ECONNREFUSED (111)
+    reason_connection_reset = 2,   // ECONNRESET (104)
+    reason_timed_out = 3,          // ETIMEDOUT (110)
+    reason_host_unreachable = 4,   // EHOSTUNREACH (113)
+    reason_net_unreachable = 5,    // ENETUNREACH (101)
+    reason_other = 255,            // anything else
+};
+
+static __always_inline u8 sk_err_to_reason(int err) {
+    switch (err) {
+    case 111:
+        return reason_connection_refused;
+    case 104:
+        return reason_connection_reset;
+    case 110:
+        return reason_timed_out;
+    case 113:
+        return reason_host_unreachable;
+    case 101:
+        return reason_net_unreachable;
+    case 0:
+        return reason_unknown;
+    default:
+        return reason_other;
+    }
+}
+
+typedef struct tcp_failed_connection {
+    u8 flags; // Must be first, we use it to tell what kind of event we have on the ring buffer
+    u8 reason;
+    u8 _pad[2];
+    connection_info_t conn;
+} tcp_failed_connection_t;
+
+// Force tcp_failed_connection_t
+const tcp_failed_connection_t *unused_tcp_failed_connection __attribute__((unused));
+
+SEC("tracepoint/sock/inet_sock_set_state")
+int obi_tracepoint_inet_sock_set_state(struct trace_event_raw_inet_sock_set_state *args) {
+    if (args->protocol != IPPROTO_TCP)
+        return 0;
+
+    if (args->newstate != TCP_CLOSE)
+        return 0;
+
+    // These are normal completions, not failures
+    if (args->oldstate == TCP_LAST_ACK || args->oldstate == TCP_TIME_WAIT)
+        return 0;
+
+    struct sock *sk = (struct sock *)args->skaddr;
+
+    connection_info_t conn;
+    if (!parse_sock_info(sk, &conn)) {
+        return 0;
+    }
+
+    int err = BPF_CORE_READ(sk, sk_err);
+    u8 reason = sk_err_to_reason(err);
+
+    // TODO remove it
+    bpf_printk("tcp failed: oldstate=%d, reason=%d, s_port=%d, d_port=%d",
+               args->oldstate,
+               reason,
+               conn.s_port,
+               conn.d_port);
+
+    tcp_failed_connection_t *se = bpf_ringbuf_reserve(&stats_events, sizeof(*se), 0);
+    if (!se) {
+        return 0;
+    }
+
+    se->flags = k_event_stat_failed_connection;
+    se->reason = reason;
+    se->conn = conn;
+
+    bpf_ringbuf_submit(se, 0);
+
+    return 0;
+}
--- a/bpf/statsolly/tp_tcp.c
+++ b/bpf/statsolly/tp_tcp.c
@@ -102,7 +102,7 @@ int obi_tracepoint_inet_sock_set_state(struct trace_event_raw_inet_sock_set_stat
         return 0;
     }
 
-    se->flags = k_event_stat_failed_connection;
+    se->flags = k_event_stat_tcp_failed_connection;
     se->reason = reason;
     se->conn = conn;
 
--- a/bpf/statsolly/tp_tcp.c
+++ b/bpf/statsolly/tp_tcp.c
@@ -106,7 +106,7 @@ int obi_tracepoint_inet_sock_set_state(struct trace_event_raw_inet_sock_set_stat
     se->reason = reason;
     se->conn = conn;
 
-    bpf_ringbuf_submit(se, 0);
+    bpf_ringbuf_submit(se, stats_events_flags());
 
     return 0;
 }
--- a/bpf/statsolly/tp_tcp.c
+++ b/bpf/statsolly/tp_tcp.c
@@ -95,11 +95,7 @@ int obi_tracepoint_inet_sock_set_state(struct trace_event_raw_inet_sock_set_stat
     const int err = BPF_CORE_READ(sk, sk_err);
     const u8 reason = sk_err_to_reason(err);
 
-    bpf_d_printk("tcp failed: oldstate=%d, reason=%d, s_port=%d, d_port=%d",
-                 args->oldstate,
-                 reason,
-                 conn.s_port,
-                 conn.d_port);
+    bpf_d_printk("tcp failed: s_port=%d, d_port=%d, reason=%d", conn.s_port, conn.d_port, reason);
 
     tcp_failed_connection_t *se = bpf_ringbuf_reserve(&stats_events, sizeof(*se), 0);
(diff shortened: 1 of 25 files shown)


bpf/statsolly/tp_tcp.c
```diff
@@ -0,0 +1,100 @@
...
+#include <statsolly/maps/stats_events.h>
+
+enum tcp_fail_reason {
+    reason_unknown = 0,            // sk_err was 0, cause not determined
+    reason_connection_refused = 1, // ECONNREFUSED (111)
+    reason_connection_reset = 2,   // ECONNRESET (104)
+    reason_timed_out = 3,          // ETIMEDOUT (110)
+    reason_host_unreachable = 4,   // EHOSTUNREACH (113)
+    reason_net_unreachable = 5,    // ENETUNREACH (101)
+    reason_other = 255,            // anything else
```
rafaelroquetto: can we define those if they aren't defined?

something like

```
#ifndef ECONNREFUSED
#define ECONNREFUSED 111
#endif
```

and then use these on the sk_err_to_reason?

pkg/internal/statsolly/ebpf/stat.go
```diff
@@ -11,6 +11,31 @@ type StatType uint8
...
+	ConnectionReset   TCPFailReasonType = "reset"
+	TimedOut          TCPFailReasonType = "timed-out"
+	HostUnreachable   TCPFailReasonType = "host-unreachable"
+	NetUnreachable    TCPFailReasonType = "net-unreachable"
+	Other             TCPFailReasonType = "other"
+)
+
+type TCPFailReasonTypeCode uint8
+
+const (
```
rafaelroquetto, line 32: we should add a reference to the enum in the ebpf code here - i wonder if bpf2go can generate enum types that we could use here (probably not but just throwing the idea)

pkg/internal/statsolly/ebpf/stats_tracer.go
```diff
@@ -90,3 +119,55 @@ func (m *StatsFetcher) Close() error {
 func (m *StatsFetcher) StatsEventsMap() *ebpf.Map {
 	return m.statsEvents
 }
+
+func kprobes(log *slog.Logger, probes map[string]ebpfcommon.ProbeDesc) ([]io.Closer, error) {
```
rafaelroquetto, line 134: this kind of duplicates the code in instrumenter.go - I am not sure, but is it possible to collapse them somehow? (Otherwise its fine)

rafaelroquetto (commented):
Looking good overall - just a few minor comments (and I also didn't check for const-correctness but perhaps we can ask Claudio to do it).

pkg/export/attributes/names/attrs.go
```diff
@@ -272,3 +272,8 @@ const (
 	GenAIRequestModel    = Name(semconv.GenAIRequestModelKey)
 	GenAIResponseModel   = Name(semconv.GenAIResponseModelKey)
 )
+
+// Stat metrics
+const (
+	TCPFailedConnectionReason = Name("reason")
```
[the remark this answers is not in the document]
pinoOgni: I think that `reason` is actually clear and unambiguous since it's scoped by the metric name.

pkg/internal/statsolly/ebpf/stat.go
pinoOgni: AFAIK bpf2go can't generate Go types from C enums so I'll add a reference.

pkg/internal/statsolly/ebpf/stats_tracer.go
pinoOgni: Yes, I got it from there, but statsolly version is simpler; the `instrumenter.go` version has `i.metrics.InstrumentationError`, uses `i.closables` and so on. 
I think the duplication is minimal.

A possible idea: we may extract just the standalone `kprobes()` and `tracepoints()` functions (already in statsolly) and put it into `pkg/ebpf/common` as free functions, and have `instrumenter` call those too instead of its own methods. Not in this PR but if it makes sense I'll work on it.

grcevski (approved):
This looks great! LGTM!

pinoOgni:
I'm putting the PR in the draft because I'd like to change a couple of things

pinoOgni:
Okay, I had a slight existential doubt for a second :) 
I thought I could merge TCP failed connections and TCP successful connections (I'm working on that now) into a single metric (playing with labels), but apparently, semantically and technically, it's better to keep them separate.

Moreover, @grcevski, @rafaelroquetto  `features.metrics: stats` has been removed in favor of `features.metrics: stats_tcp_rtt, ...` with the list of desired stat metrics. Not in this PR but, WDYT about adding `stat_all, application_all, and network_all` to be able to select all metrics from statsolly, appolly, and nettolly, respectively (like a subset of `all`/`*`)

Update: We aligned ourselves with @mmat11 and agreed that removing `stats` is a breaking change so the solution is to use `stats` to select all stat metrics and then each stat metric can be selected with its name.

bpf/statsolly/types.h
```diff
@@ -5,5 +5,6 @@
 
 #pragma once
 enum {
-    k_event_stat_tcp_rtt = 1, // StatTypeTCPRtt
+    k_event_stat_tcp_rtt = 1,           // StatTypeTCPRtt
+    k_event_stat_failed_connection = 2, // StatTypeTCPFailedConnection
```
mmat11: should this be `StatTypeFailedConnection`? or the constant be named `k_event_stat_tcp_failed_connection`

it looks inconsistent now

pkg/internal/statsolly/ebpf/stats_tracer.go
```diff
@@ -90,3 +130,55 @@ func (m *StatsFetcher) Close() error {
...
+			closables = append(closables, krp)
+		}
+	}
+	return closables, nil
+}
+
+func tracepoints(log *slog.Logger, probes map[string]ebpfcommon.ProbeDesc) ([]io.Closer, error) {
+	var closables []io.Closer
+	for funcName, desc := range probes {
+		if desc.Start == nil {
```
mmat11: what's Start? should we error instead?

pkg/internal/statsolly/ebpf/stats_tracer.go
```diff
@@ -90,3 +130,55 @@ func (m *StatsFetcher) Close() error {
...
+		parts := strings.SplitN(funcName, "/", 2)
+		if len(parts) != 2 {
+			return closables, fmt.Errorf("invalid tracepoint %q: must be group/name", funcName)
+		}
+		tp, err := link.Tracepoint(parts[0], parts[1], desc.Start, nil)
+		if err != nil {
+			if desc.Required {
+				return closables, fmt.Errorf("tracepoint %s: %w", funcName, err)
+			}
+			log.Warn("tracepoint failed", "tracepoint", funcName, "error", err)
```
mmat11:
```suggestion
			log.Warn("tracepoint attachment failed", "tracepoint", funcName, "error", err)
```

pkg/internal/statsolly/ebpf/stats_tracer.go
```diff
@@ -90,3 +130,55 @@ func (m *StatsFetcher) Close() error {
...
+				log.Warn("kretprobe failed", "function", funcName, "error", err)
+				continue
+			}
+			closables = append(closables, krp)
+		}
+	}
+	return closables, nil
+}
+
+func tracepoints(log *slog.Logger, probes map[string]ebpfcommon.ProbeDesc) ([]io.Closer, error) {
```
mmat11: I know these functions are named this way in generictracer, but I think we should give these functions more explicit names here, eg. `attachTracepoints` or similar

pkg/internal/statsolly/ebpf/stats_tracer.go
```diff
@@ -90,3 +130,55 @@ func (m *StatsFetcher) Close() error {
...
+func kprobes(log *slog.Logger, probes map[string]ebpfcommon.ProbeDesc) ([]io.Closer, error) {
+	var closables []io.Closer
+	for funcName, desc := range probes {
+		if desc.Start != nil {
+			kp, err := link.Kprobe(funcName, desc.Start, nil)
+			if err != nil {
+				if desc.Required {
+					return closables, fmt.Errorf("kprobe %s: %w", funcName, err)
+				}
+				log.Warn("kprobe failed", "function", funcName, "error", err)
```
mmat11:
```suggestion
				log.Warn("kprobe attachment failed", "function", funcName, "error", err)
```

pkg/internal/statsolly/ebpf/stats_tracer.go
```diff
@@ -90,3 +130,55 @@ func (m *StatsFetcher) Close() error {
...
+			}
+			closables = append(closables, kp)
+		}
+		if desc.End != nil {
+			krp, err := link.Kretprobe(funcName, desc.End, nil)
+			if err != nil {
+				if desc.Required {
+					return closables, fmt.Errorf("kretprobe %s: %w", funcName, err)
+				}
+				log.Warn("kretprobe failed", "function", funcName, "error", err)
```
mmat11:
```suggestion
				log.Warn("kretprobe attachment failed", "function", funcName, "error", err)
```

pkg/internal/statsolly/ebpf/stats_tracer.go
```diff
@@ -58,20 +72,46 @@ func NewStatsFetcher(cfg *config.EBPFTracer) (*StatsFetcher, error) {
...
+	}
+	kpClosables, err := kprobes(tlog, kps)
 	if err != nil {
-		tlog.Error("opening %s: %s", "tcp_close", err)
-		return nil, fmt.Errorf("opening kprobe: %w", err)
+		closeAll(kpClosables)
+		return nil, err
+	}
+
+	tps := map[string]ebpfcommon.ProbeDesc{
```
mmat11: why are we using ProbeDesc (and forcing ourselves to use generictracer's code) while we could just do `lnk, err := link.Tracepoint(...)` for these 2 probes?

pkg/obi/config.go
```diff
@@ -631,10 +631,10 @@ func (c *Config) Validate() error {
 	}
 
 	if !c.Enabled(FeatureNetO11y) && !c.Enabled(FeatureAppO11y) && !c.Enabled(FeatureStatsO11y) {
-		return ConfigError("at least one of 'network', 'application' or 'stats' features must be enabled. " +
-			"Enable an OpenTelemetry or Prometheus metrics export, then enable any of the network*, application* or stats*" +
-			"features using the 'OTEL_EBPF_METRICS_FEATURES=network,application,stats' environment variable " +
-			"or 'meter_provider: { features: [network,application,stats] }' in the YAML configuration file. ")
+		return ConfigError("at least one feature must be enabled (network*, application* or stats* prefixed). " +
```
mmat11: I think we should leave this comment as-is if nothing changed in config

bpf/statsolly/types.h
pinoOgni: Ops! Yes it should be `k_event_stat_tcp_failed_connection`

pkg/internal/statsolly/ebpf/stats_tracer.go
pinoOgni: Nice catch! Start is the program to attach as tracepoint in this case but for tracepoints `End` doesn't exist, so `Start == nil` is a misconfiguration. We need to return an error!

pkg/internal/statsolly/ebpf/stats_tracer.go
pinoOgni: Fine by me

pkg/internal/statsolly/ebpf/stats_tracer.go
pinoOgni: Maybe you are right. I used `ProbeDesc` because I implemented `kprobes` first. I'll update it and this will automatically solve also the `if desc.Start == nil ...` issue.

pkg/obi/config.go
pinoOgni: IMO `at least one of 'network', 'application' or 'stats' features must be enabled` is misleading, ~moreover "stats" is not anymore a feature so it will be wrong.~
Reverted.

pkg/export/feature.go
```diff
@@ -38,18 +39,20 @@ const (
...
-	"application":               FeatureApplicationRED,
-	"application_span":          FeatureSpanLegacy,
-	"application_span_otel":     FeatureSpanOTel,
-	"application_span_sizes":    FeatureSpanSizes,
-	"application_service_graph": FeatureGraph,
-	"application_host":          FeatureApplicationHost,
-	"ebpf":                      FeatureEBPF,
-	"all":                       FeatureAll,
-	"*":                         FeatureAll,
+	"stats":                        FeatureStatsTCPRtt | FeatureStatsTCPFailedConnections,
```
pinoOgni: `stats` now enables all stat metrics. This is backwards-compatible because TCP RTT used to be the only stat metric, so anyone already using stats keeps the same behavior.

pkg/export/feature.go
```diff
@@ -23,7 +23,8 @@ const (
 	// zero value.
 	FeatureEmpty Features = 1 << iota
 	FeatureNetwork
-	FeatureStats
+	FeatureStatsTCPRtt
+	FeatureStatsTCPFailedConnections
```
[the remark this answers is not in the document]
pinoOgni: Just adding the `FeatureStats` const again so there is no breaking change for external consumers.

pkg/internal/statsolly/ebpf/stats_tracer.go
```diff
@@ -58,20 +71,50 @@ func NewStatsFetcher(cfg *config.EBPFTracer) (*StatsFetcher, error) {
...
+			closeAll(closables)
+			return nil, fmt.Errorf("invalid tracepoint %q: must be group/name", funcName)
+		}
+		tp, terr := link.Tracepoint(parts[0], parts[1], program, nil)
+		if terr != nil {
+			closeAll(closables)
+			return nil, fmt.Errorf("tracepoint attachment failed %s: %w", funcName, terr)
+		}
+		closables = append(closables, tp)
 	}
```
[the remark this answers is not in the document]
pinoOgni: 💯

pkg/export/prom/prom_stats.go
```diff
@@ -144,6 +163,14 @@ func (r *statMetricsReporter) observeTCPRtt(stat *ebpf.Stat) {
 	if r.tcpRtt == nil || stat.TCPRtt == nil {
 		return
 	}
-	r.tcpRtt.WithLabelValues(labelValues(stat, r.statsAttrs)...).
+	r.tcpRtt.WithLabelValues(labelValues(stat, r.tcpRttAttrs)...).
 		Metric.Observe(float64(stat.TCPRtt.SrttUs) / 1_000_000.0)
 }
+
+func (r *statMetricsReporter) observeTCPFailedConnections(stat *ebpf.Stat) {
```
mmat11, line 170: I'm not too much into prometheus best practices, but I _think_ this should be named differently. Maybe `incTCPFailedConnections` would work better since this is a counter?

mmat11 (approved):

pkg/export/prom/prom_stats.go
pinoOgni: I simply followed the netolly convention for `observeFlowBytes` and `observeInterZone`. They're counters, but the metric isn't incremented only by 1. I can change the name and update change `Metric.Add(1)` to `Metric.Inc()`.  I don't have any strong opinions on that. What do you think?

mmat11: If the same pattern is used somewhere else, let's keep as is for consistency

PR #1832, opened 2026-04-14
bpf: add gRPC/HTTP2 context propagation via sk_msg HPACK injection

### LINKED ISSUE #1095: gRPC context Propagation
like we have with http, allow to propagate context in http2/grpc headers

mmat11:
## Summary

This PR implements http2/grpc context propagation
resolves https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1095

## Validation


bpf/tpinjector/tpinjector.c
```diff
@@ -715,54 +814,48 @@ int obi_packet_extender(struct sk_msg_md *msg) {
...
-    bpf_dbg_printk("len=%d, s_port=%d", msg->size, msg->local_port);
-    bpf_dbg_printk("buf=[%s]", msg->data);
-    bpf_dbg_printk("ptr=%llx, end=%llx", ctx_msg_data(msg), ctx_msg_data_end(msg));
-    bpf_dbg_printk("BUF=[%s]", ctx_msg_data(msg));
-
-    init_tp_ctx_parent_tp(t_ctx);
-
-    bpf_tail_call_static(msg, &extender_jump_table, k_tail_find_existing_tp);
+    // HTTP/2 detection: known H2 socket or "PRI " preface.
+    if ((inject_flags & k_inject_http_headers) && msg->size >= k_h2_frame_header_len) {
```
grcevski, line 850: I think here we need to check if it's PRI or it's recorded in ongoing_http2_connections. PRI is only seen if it's the first communication between sides, but grpc clients tend to hold on to the same connection. When this happens we'll misclassify the request as TCP, we'll ship it to user-space and OBI user-space will detect it's HTTP2. Then from userspace we record the connection as HTTP2 into ongoing_http2_connections. This is how protocol_handler.h checks this. We must also check if it was HTTP2 SSL and not do anything, SSL and HTTP2 can easily be mixed up.

    http2_conn_info_data_t *h2g = bpf_map_lookup_elem(&ongoing_http2_connections, &args->pid_conn);
    if (h2g && (http2_flag_ssl(h2g->flags) == args->ssl)) {

grcevski (approved):
This looks great! LGTM! I think we don't need to add the suggestion I made in this PR, we can follow-up with the additional check. I'm sure the CI will give you hell with the verifier :).

bpf/generictracer/protocol_http2.h
```diff
@@ -56,47 +61,200 @@ static __always_inline u64 uniqueHTTP2ConnId(pid_connection_info_t *p_conn) {
     return random_id;
 }
 
+// Try to parse a traceparent value at the given pointer.
+// Returns 1 if a valid traceparent was found, 0 otherwise.
+static __always_inline u8 try_parse_tp_value(const unsigned char *val, tp_info_t *tp) {
```
rafaelroquetto, line 66: on the tpinjector code, we sanity check when parsing HTTP1 traceparent headers by checking for the existence of `-` - see https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/3583541ee601ad7e216e18b3ffedfe5b1a4ec714/bpf/tpinjector/tpinjector.c#L876

This is to avoid false positives. Would it be the case here as well? Can it be done?

bpf/generictracer/protocol_http2.h
```diff
@@ -56,47 +61,200 @@ static __always_inline u64 uniqueHTTP2ConnId(pid_connection_info_t *p_conn) {
...
+// name formats, matching sk_msg injection and Go uprobe injection respectively.
+static __always_inline u8 parse_hpack_traceparent(const unsigned char *data,
+                                                  u32 data_len,
+                                                  tp_info_t *tp) {
+    if (data_len < k_h2_tp_hpack_huffman_size) {
+        return 0;
+    }
+
+    const u32 max_pos = data_len - k_h2_tp_hpack_huffman_size;
+    const char tp_name[] = "traceparent";
```
rafaelroquetto, line 87: should this be defined in `tc_common.h` alongside the other constants (and in the same fashion)?

Doing that would also save stack space...

rafaelroquetto:
Just a heads up, I've started reviewing this, so if possible, hold on merging this one for a bit (just for another day).

bpf/generictracer/protocol_http2.h
```diff
@@ -56,47 +61,200 @@ static __always_inline u64 uniqueHTTP2ConnId(pid_connection_info_t *p_conn) {
...
+        bpf_map_delete_elem(&outgoing_trace_map, &sorted_e);
+    }
+}
+
 static __always_inline void http2_grpc_start(
     http2_conn_stream_t *s_key, void *u_buf, int len, u8 direction, u8 ssl, u16 orig_dport) {
     http2_grpc_request_t *existing = bpf_map_lookup_elem(&ongoing_http2_grpc, s_key);
     if (existing) {
         bpf_dbg_printk("already found existing grpcstart, ignoring this exchange");
+        if (existing->type == EVENT_HTTP_CLIENT) {
```
rafaelroquetto, line 155: just double checking - if we find an existing connection with a different traceparent, we override it/ last one wins?

bpf/tpinjector/tpinjector.c
```diff
@@ -685,7 +768,23 @@ static __always_inline void handle_existing_tp_pid(struct sk_msg_md *msg,
         if (inject_flags & k_inject_http_headers) {
             write_http_traceparent(msg, tp_pid);
         }
+        return;
```
rafaelroquetto, line 771: I think here we no longer clear_tp_info_pid if the inject flags condition does not eval to `true` and we never tail call in write_http_traceparent.

This may cause http_get_or_create_trace_info to misbehave

bpf/tpinjector/tpinjector.c
```diff
@@ -685,7 +768,23 @@ static __always_inline void handle_existing_tp_pid(struct sk_msg_md *msg,
...
+    }
+
+    // Not HTTP/1.1 — check for HTTP/2 (gRPC). For Go processes the
+    // Go uprobe sets tp_pid for ALL outgoing calls including gRPC.
+    // Without H2 detection here, gRPC sends skip injection entirely.
+    // Read socket flag BEFORE msg->data to avoid verifier rejection.
+    if (inject_flags & k_inject_http_headers) {
```
rafaelroquetto, line 778: I wonder if it is worth it making the condition on this line the outer loop, negated, and bail early, and if not, check for is_http or is_http2.


Another thing I will mention - perhaps  the is_h2_socket call should be inside protocol_detector, which then returns a protocol enum instead of is_http - the reason being that protocol_detector also sets up the `msg_buffers` map - so if anyone reorders this, and decides to check http2 before http, we may never set this buffer and that will fail downstream in generic tracer.

Putting everything in `protocol_detector` makes it a bit "atomic" (in very big quotes)

bpf/tpinjector/tpinjector.c
```diff
@@ -685,7 +768,23 @@ static __always_inline void handle_existing_tp_pid(struct sk_msg_md *msg,
...
+        const bool h2_flag = is_h2_socket(msg);
+        unsigned char *d = msg->data;
+        if (d && (void *)d + k_h2_frame_header_len <= msg->data_end) {
+            if (h2_flag || (d[0] == 'P' && d[1] == 'R' && d[2] == 'I' && d[3] == ' ')) {
+                bpf_tail_call_static(msg, &extender_jump_table, k_tail_detect_h2);
+            }
```
rafaelroquetto, line 784: and then, just to mirror and be consistent with the http case, I'd wrap this in `wrap_http2_traceparent`.

You may also do a bpf_msg_pull_data to the min http2 size or at least 4 to ensure the data is there (it will likely be, but who knows)

bpf/tpinjector/tpinjector.c
```diff
@@ -716,54 +815,48 @@ int obi_packet_extender(struct sk_msg_md *msg) {
...
-        // not enough data to detect anything, bail
         return SK_PASS;
     }
 
-    bpf_msg_pull_data(msg, 0, msg->size, 0);
-
-    // TODO: execute the protocol handlers here with tail calls, don't
-    // rely on tcp_sendmsg to do it and record these message buffers.
+    // Path 2: kprobe-instrumented processes — detect HTTP/1.1.
+    if (valid_pid(id)) {
```
rafaelroquetto, line 832: why is `valid_pid` only checked for HTTP?

the uprobes are handled in `handle_existing_tp_pid` so I'd expect that anything reaching this codepoint is a non-go, non-ssl "generic" connection that is tracked via valid_pid... Otherwise we will be picking all the outgoing messages of every process, no?

bpf/tpinjector/tpinjector.c
```diff
@@ -716,54 +815,48 @@ int obi_packet_extender(struct sk_msg_md *msg) {
...
-    init_tp_ctx_parent_tp(t_ctx);
-
-    bpf_tail_call_static(msg, &extender_jump_table, k_tail_find_existing_tp);
+    // HTTP/2 detection: known H2 socket or "PRI " preface.
+    if ((inject_flags & k_inject_http_headers) && msg->size >= k_h2_frame_header_len) {
+        if (is_h2_socket(msg)) {
+            bpf_tail_call_static(msg, &extender_jump_table, k_tail_detect_h2);
+        }
+        unsigned char *d = msg->data;
+        if (d && (void *)d + k_h2_preface_check_len <= msg->data_end && d[0] == 'P' &&
```
rafaelroquetto, line 855: here you should also pull data - this would be best put in a helper and shared with the handle_existing_tp_info codepath.

bpf/tpinjector/tpinjector.c
```diff
@@ -968,3 +1061,424 @@ int obi_packet_extender_create_tp(struct sk_msg_md *msg) {
...
+    u32 pos = 0;
+
+    // Check for HTTP/2 connection preface. Pull first 8 bytes for check.
+    // Consolidate the full message first. This ensures targeted pulls
+    // in the frame scan loop are pointer updates only (no splitting).
+    // Safe because detect_h2 is only called for confirmed H2 traffic.
+    bpf_msg_pull_data(msg, 0, msg->size, 0);
+
+    if (msg->size >= k_h2_preface_len + k_h2_frame_header_len) {
+        if (bpf_msg_pull_data(msg, 0, k_h2_preface_check_len, 0) == 0) {
```
rafaelroquetto, line 1091: this looks wrong - we already pull `msg->size` on line 1088 so this is redudant.

It would be best to pull the minimum data size actually, and only pull the full msg->size if required.

bpf/tpinjector/tpinjector.c
```diff
@@ -968,3 +1061,424 @@ int obi_packet_extender_create_tp(struct sk_msg_md *msg) {
...
+
+    // Scan up to 4 frames for a HEADERS frame with END_HEADERS.
+    // Each iteration pulls exactly 9 bytes at the current position,
+    // giving the verifier constant-offset access via msg->data[0..8].
+    for (u8 i = 0; i < k_h2_max_frame_scan; i++) {
+        if (pos + k_h2_frame_header_len > msg->size) {
+            return SK_PASS;
+        }
+
+        if (bpf_msg_pull_data(msg, pos, pos + k_h2_frame_header_len, 0) != 0) {
```
rafaelroquetto, line 1109: ditto - this is probably even confusing the verifier, as everytime this is called it invalidate the data pointers, which is why you need to reload `d` on line 1113

bpf/tpinjector/tpinjector.c
```diff
@@ -968,3 +1061,424 @@ int obi_packet_extender_create_tp(struct sk_msg_md *msg) {
...
+
+            // If the Go uprobe already injected traceparent on the wire for
+            // this stream (written=1), don't create a second child span.
+            // The wire and outgoing_trace_map already have the correct trace
+            // context; adopt_injected_trace will pick it up in the kprobe.
+            // Also inject HPACK so non-Go receivers (Node.js, Java, Python)
+            // can find the traceparent via parse_hpack_traceparent as a
+            // fallback when TCP options are not received in time.
+            tp_info_pid_t *go_tp = get_tp_info_pid(&t_ctx->e_key);
+            if (go_tp && go_tp->valid && go_tp->written) {
```
rafaelroquetto, line 1143: we shouldn't be doing this inside the loop - instead, we should cache this tailcall_ctx when we get it first in the beginning of obi_packet_extender, and use it here.

bpf/tpinjector/tpinjector.c
```diff
@@ -968,3 +1061,424 @@ int obi_packet_extender_create_tp(struct sk_msg_md *msg) {
...
+                if (inject_flags & k_inject_tcp_options) {
+                    schedule_write_tcp_option(msg, go_tp);
+                }
+                if (inject_flags & k_inject_http_headers) {
+                    tp_info_pid_t *tp_p = (tp_info_pid_t *)tp_info_mem();
+                    if (tp_p) {
+                        bpf_memcpy(tp_p, go_tp, sizeof(*tp_p));
+                        bpf_msg_pull_data(msg, 0, msg->size, 0);
```
rafaelroquetto, line 1151: this should be removed

bpf/tpinjector/tpinjector.c
```diff
@@ -968,3 +1061,424 @@ int obi_packet_extender_create_tp(struct sk_msg_md *msg) {
...
+            // context; adopt_injected_trace will pick it up in the kprobe.
+            // Also inject HPACK so non-Go receivers (Node.js, Java, Python)
+            // can find the traceparent via parse_hpack_traceparent as a
+            // fallback when TCP options are not received in time.
+            tp_info_pid_t *go_tp = get_tp_info_pid(&t_ctx->e_key);
+            if (go_tp && go_tp->valid && go_tp->written) {
+                if (inject_flags & k_inject_tcp_options) {
+                    schedule_write_tcp_option(msg, go_tp);
+                }
+                if (inject_flags & k_inject_http_headers) {
```
rafaelroquetto, line 1147: also this should probably be cached before the loop, and the conditions reversed so we first check for the flag, and then inspect go_tp

bpf/tpinjector/tpinjector.c
```diff
@@ -968,3 +1061,424 @@ int obi_packet_extender_create_tp(struct sk_msg_md *msg) {
...
+                    }
+                }
+                return SK_PASS;
+            }
+
+            // Re-consolidate the full message before tail-calling the
+            // injection chain. The targeted pulls in the frame scan may
+            // have fragmented the scatter-gather list. write_h2_tp needs
+            // a clean state for bpf_msg_push_data to work correctly.
+            bpf_msg_pull_data(msg, 0, msg->size, 0);
```
rafaelroquetto, line 1163: that's not true - once we do a bpf_msg_pull_data with msg->size the kernel will linerise everything once, it won't drop things

bpf/tpinjector/tpinjector.c
```diff
@@ -968,3 +1061,424 @@ int obi_packet_extender_create_tp(struct sk_msg_md *msg) {
...
+    }
+
+    const unsigned char *data = msg->data;
+    const unsigned char *end = msg->data_end;
+
+    if (!data) {
+        return 0;
+    }
+
+    const char tp_name[] = "traceparent";
```
rafaelroquetto, line 1216: as in protocol_http2 - I'd move this to tc_common and reuse

bpf/tpinjector/tpinjector.c
```diff
@@ -968,3 +1061,424 @@ int obi_packet_extender_create_tp(struct sk_msg_md *msg) {
...
+            schedule_write_tcp_option(msg, tp_p);
+        }
+
+        // Overwrite the span_id hex in the existing traceparent on the wire.
+        const long pull_err =
+            bpf_msg_pull_data(msg, span_id_offset, span_id_offset + SPAN_ID_CHAR_LEN, 0);
+        if (pull_err == 0) {
+            unsigned char *wire = msg->data;
+            if (wire && (void *)wire + SPAN_ID_CHAR_LEN <= msg->data_end) {
+                encode_hex(wire, tp_p->tp.span_id, SPAN_ID_SIZE_BYTES);
```
rafaelroquetto, line 1325: shouldn't this only write when we havek_inject_http_headers?

bpf/tpinjector/tpinjector.c
```diff
@@ -716,54 +815,48 @@ int obi_packet_extender(struct sk_msg_md *msg) {
...
-    bpf_dbg_printk("len=%d, s_port=%d", msg->size, msg->local_port);
-    bpf_dbg_printk("buf=[%s]", msg->data);
-    bpf_dbg_printk("ptr=%llx, end=%llx", ctx_msg_data(msg), ctx_msg_data_end(msg));
-    bpf_dbg_printk("BUF=[%s]", ctx_msg_data(msg));
-
-    init_tp_ctx_parent_tp(t_ctx);
-
-    bpf_tail_call_static(msg, &extender_jump_table, k_tail_find_existing_tp);
+    // HTTP/2 detection: known H2 socket or "PRI " preface.
+    if ((inject_flags & k_inject_http_headers) && msg->size >= k_h2_frame_header_len) {
```
rafaelroquetto, line 850: also, I suppose here we want to check for k_inject_tcp_options - otherwise we never inject them for HTTP2

rafaelroquetto (changes requested):
I did a first pass, in particular, I think tpinjector may have a few bugs that require some interations. If _think_ it's not injecting the tcp options at all in certain cases (i.e. when  `context_propagation: tcp` is set).

Also, we should probably pull data (msg->size) only once - it's ok to pull less than msg->size (i.e. min http 2) for smaller checks and pull the entire thing later, but once we do, we don't have to pull it further.

mmat11:
@grcevski @rafaelroquetto thanks for the review, latest rebase broke something and I'm investigating it -- I did most of the changes locally and I'll push as soon as the test passes reliably

bpf/tpinjector/tpinjector.c
mmat11: should be done now, `WITH_SSL` are skipped

bpf/generictracer/protocol_http2.h
mmat11: this should now be checked in `try_parse_tp_value`

bpf/generictracer/protocol_http2.h
mmat11: done

bpf/generictracer/protocol_http2.h
mmat11: not sure I get the question correctly, the override here happens only when written=1 (which was set previously either by uprobe or sk_msg)

bpf/tpinjector/tpinjector.c
mmat11: cleanups added

bpf/tpinjector/tpinjector.c
mmat11: the matching logic has been refactored a bit now, regarding the enum refactor, I would leave it as a follow up, WDYT?

bpf/tpinjector/tpinjector.c
mmat11: done

bpf/tpinjector/tpinjector.c
mmat11: for certain processes like nodejs libuv, the tracked pid doesn't match here, I added an additional check here https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1832/changes/8b2035310785fb4aadb2071b35e7045de8bc2893

bpf/tpinjector/tpinjector.c
mmat11: pull is now done in `wrap_http2_traceparent` before checking preface and tailcalling

bpf/tpinjector/tpinjector.c
mmat11: this has been refactored now

bpf/tpinjector/tpinjector.c
mmat11: also refactored

bpf/tpinjector/tpinjector.c
mmat11: it needs stream_id parsing, once it's found it exits the loop though

bpf/tpinjector/tpinjector.c
mmat11: this has een refactored as well

bpf/tpinjector/tpinjector.c
mmat11: full pull has been removed

bpf/tpinjector/tpinjector.c
mmat11: done

bpf/tpinjector/tpinjector.c
```diff
@@ -576,6 +624,43 @@ make_tp_string_skb(unsigned char *buf, const tp_info_t *tp, const unsigned char
     bpf_dbg_printk("tp_string=%s", tp_string);
 }
 
+static __always_inline void
+make_h2_tp_hpack(unsigned char *buf, const tp_info_t *tp, const unsigned char *end) {
```
rafaelroquetto, line 628: Perhaps here we could move this function after line 577 and then we can delete the definition on line 549, making this consistent with the rest of the file.

bpf/tpinjector/tpinjector.c
```diff
@@ -576,6 +624,43 @@ make_tp_string_skb(unsigned char *buf, const tp_info_t *tp, const unsigned char
...
+    buf = check_pkt_access(buf, k_h2_tp_hpack_size, end);
+
+    if (!buf) {
+        return;
+    }
+
+    // clang-format off
```
rafaelroquetto, line 635: nit: instead of disabling the linter, we could simply stick to the pattern and increment buf once per line, like we do elsewhere in the file

bpf/tpinjector/tpinjector.c
```diff
@@ -710,61 +824,57 @@ int obi_packet_extender(struct sk_msg_md *msg) {
...
-    if (tp_pid) {
-        handle_existing_tp_pid(msg, id, &conn, &e_key, tp_pid);
+    // H2 entries are keyed by actual stream_id; the stream=0 slot carries an
+    // unrelated tp written by server_or_client_trace. Tail-call detect_h2 for
+    // per-stream lookup. protocol_detector still runs so msg_buffers is
+    // populated for the kprobe fallback (nodejs CLIENT span emission needs it).
+    if (is_h2_socket(msg)) {
+        if (valid_pid(id)) {
+            bpf_msg_pull_data(msg, 0, msg->size, 0);
+            protocol_detector(msg, id, &conn, &e_key);
```
rafaelroquetto, line 839: since we are discarding the return value here, we are also performing unnecessary checks on a codepath that runs for every outgoing message - could we split protocol_detector so that the part that populates the msg_buffers is its own function that could be called both directly from here, and also from protocol_detector?

bpf/tpinjector/tpinjector.c
```diff
@@ -710,61 +824,57 @@ int obi_packet_extender(struct sk_msg_md *msg) {
...
-
-    // Higher-level uprobes have already set the tp_pid for us (either Go, or SSL)
-    if (tp_pid) {
-        handle_existing_tp_pid(msg, id, &conn, &e_key, tp_pid);
+    // H2 entries are keyed by actual stream_id; the stream=0 slot carries an
+    // unrelated tp written by server_or_client_trace. Tail-call detect_h2 for
+    // per-stream lookup. protocol_detector still runs so msg_buffers is
+    // populated for the kprobe fallback (nodejs CLIENT span emission needs it).
+    if (is_h2_socket(msg)) {
+        if (valid_pid(id)) {
```
rafaelroquetto, line 837: am I right to assume this codepath is for plaintext, non-GO, non-SSL HTTP2 traffic?

also, is it possible to have this check (is_h2_socket) after the `if (tp_pid)` check below? So first we check if there's already something setup by the uprobes, and if not, we fall back to the usual detection paths. Which ties to my next question: I don't understand the check to `valid_pid(id)` here - if it returns false, why do we want to proceed with `k_tail_detect_h2`? In this case, wouldn't we be processing unrelated traffic?

bpf/tpinjector/tpinjector.c
rafaelroquetto: I thought sk_msg always ran in the context of the sending process - I'd expect tests like nodejs-dist to fail otherwise. Are you sure about this? When it doesn't match, what PID do we have here instead?

bpf/tpinjector/tpinjector.c
```diff
@@ -710,61 +824,57 @@ int obi_packet_extender(struct sk_msg_md *msg) {
...
-    if (!is_http) {
-        return SK_PASS;
+    // Path 2: kprobe-instrumented processes — detect HTTP/1.1.
+    if (valid_pid(id)) {
+        bpf_dbg_printk("MSG=%llx:%d ->", conn.s_ip[3], conn.s_port);
+        bpf_dbg_printk("MSG TO=%llx:%d", conn.d_ip[3], conn.d_port);
+        bpf_dbg_printk("MSG SIZE=%u", msg->size);
+
+        bpf_msg_pull_data(msg, 0, msg->size, 0);
+        if (protocol_detector(msg, id, &conn, &e_key)) {
```
rafaelroquetto, line 867: nit: I'd keep the `is_http` variable as it tells the reader what protocol_detector is returning - this if is more obscure and we actually lost clarity here (which would contrast well with wrap_http2_traceparent)

bpf/tpinjector/tpinjector.c
```diff
@@ -710,61 +824,57 @@ int obi_packet_extender(struct sk_msg_md *msg) {
...
-    bpf_dbg_printk("len=%d, s_port=%d", msg->size, msg->local_port);
-    bpf_dbg_printk("buf=[%s]", msg->data);
-    bpf_dbg_printk("ptr=%llx, end=%llx", ctx_msg_data(msg), ctx_msg_data_end(msg));
-    bpf_dbg_printk("BUF=[%s]", ctx_msg_data(msg));
-
-    init_tp_ctx_parent_tp(t_ctx);
-
-    bpf_tail_call_static(msg, &extender_jump_table, k_tail_find_existing_tp);
-
+    // H2 detect runs regardless of valid_pid: Go procs aren't in the PID map
```
rafaelroquetto, line 875: but go procs should have been handled by `handle_existing_tp_pid`, they will never reach it here - this is why it happens before the original `valid_pid` check. So I am not sure about this.

`valid_pid` should also handled forked-off / child processes - it explicitly checks both the processes own pid and `ns_ppid`. Were it not the case, a lot of other things would be broken.

bpf/tpinjector/tpinjector.c
```diff
@@ -968,3 +1078,418 @@ int obi_packet_extender_create_tp(struct sk_msg_md *msg) {
...
+
+    // Mark socket on preface detection regardless of whether a frame follows
+    // in the same sendmsg. Go gRPC sends the 24-byte preface alone, then
+    // HEADERS in a separate write — gating the mark on preface+frame means
+    // the HEADERS write sees is_h2_socket=false and bypasses detect_h2.
+    if (msg_size >= k_h2_preface_check_len) {
+        if (bpf_msg_pull_data(msg, 0, k_h2_preface_check_len, 0) == 0) {
+            unsigned char *d = msg->data;
+            if (d && (void *)d + k_h2_preface_check_len <= msg->data_end && d[0] == 'P' &&
+                d[1] == 'R' && d[2] == 'I' && d[3] == ' ') {
```
rafaelroquetto, line 1107: this should be moved into its own inline helper (e.g. `is_http2_preface`) and reused by wrap_http2_traceparent, see https://github.com/rafaelroquetto/opentelemetry-ebpf-instrumentation/blob/179c435c24dcb619b4dfcd7bb9ffa76669ed0012/bpf/socktracer/http2.h#L20

rafaelroquetto (changes requested):
Thanks a lot for the changes. I've done another pass, but I stuck to parts of tpinjector only and I think I should probably list what I have found before digging deeper, so that I don't flood your PR and to allow us to iterate the higher level concerns before digging deeper.

A few notes:

- we should drop TCP options injection for HTTP2 (and document that) - it occurred to me that if we have more than one stream batched in the same sk_msg or tcp packet, we are toast, as we support only a single option per packet - I think it's just not possible to do properly.
- likewise, we can't do any injection for HTTP2/SSL at the sk_msg level (it needs to be done a the uprobe level) - with that in mind, the codepath for `get_tp_info_pid(&e_key) -> handle_existing_tp_pid(...)` only makes sense for non-HTTP2 packets that are actually not batched. This is either plaintext HTTP setup by the go uprobes, or SSL (HTTP) setup by the SSL uprobes.
- finally, we fall back to the regular plaintext processing - we check valid_pid(id) for the general case, but for the go stuff, the go uprobes setup a map keyed by pid_connection_info_t or something like that that tells us if this connection is to be processed.

```
sk_msg (outgoing packet)
        │
        ├─ get_tp_info_pid(e_key{stream_id=0})
        │       found ──► handle_existing_tp_pid → inject & return
        │
        ├─ find_go_conn_info(conn)
        │       found && is_ssl ──► bail (handled by SSL uprobes)
        │       found && is_go  ──► bail (uprobe already injected via bpf_probe_write_user)
        │
        ├─ !go_conn_info && !valid_pid ──► bail (unmonitored non-Go process)
        │
        └─ detect H1 / H2, generate span, inject & return

```

This also does away with another issue: grpc_pending_egress_trace - if I understood it correctly, this can be racy because `uprobe/transport_http2Client_NewStream` can be called concurrently, which would stomp the value in grpc_pending_egress_trace. 

With the approach above, we can get rid of it because go connections bail out early at the sk_msg level (is_go → bail), so sk_msg never needs to pick up a pending trace for them - the go uprobes handle injection directly via `bpf_probe_write_user`.

I haven't had the time to review the kprobes codepath yet.

rafaelroquetto:
I forgot to mention - it would also be good to add integration tests for the cases above, as I guess the existing tests don't cover the sk_msg HPACK injection path for non-Go clients.

mmat11:
@rafaelroquetto I think I addressed all comments

edit: regarding tests, the Go hops in the relay test are using Go uprobes (there's no SKIP_GO_SPECIFIC_TRACERS set); the full chain is:

```
// go-entry
// python-relay
// go-grpc-to-http
// go-http-to-grpc
// nodejs-relay
// java-relay
// go-terminal
```

mmat11:
@rafaelroquetto you think we can merge this or need to do other passes?

rafaelroquetto:
@mmat11 I haven't had the time to review the latest changes yet, will try to get to it today or tomorrow. Will keep you posted.

bpf/gotracer/go_grpc.c
```diff
@@ -592,6 +605,26 @@ int obi_uprobe_transport_http2Client_NewStream(struct pt_regs *ctx) {
                 wrapper.s_key.conn_ptr = (u64)conn_ptr_key;
 
                 bpf_map_update_elem(&transport_new_client_invocations, &g_key, &wrapper, BPF_ANY);
+
+                // Race-free: NewStream ENTRY runs before WriteHeaders control-buffer enqueue.
```
MrAlias: This fallback still looks racy on a reused HTTP/2 connection. `grpc_pending_egress_trace` is keyed only by `{s_port,d_port,stream_id=0}`, so two concurrent `NewStream` calls on the same connection can overwrite each other before `NewStream_ret` publishes the real stream ID.

When `grpcFramerWriteHeaders` later falls back to this map, it can pick up request A’s trace and republish it under request B’s actual stream ID. I think this needs a per-request key that stays unique across concurrent streams, or this fallback needs to be removed entirely.

bpf/tpinjector/tpinjector.c
```diff
@@ -968,3 +1102,396 @@ int obi_packet_extender_create_tp(struct sk_msg_md *msg) {
...
+            return SK_PASS;
+        }
+    } else {
+        // No existing tp: derive one locally. Go gRPC handles injection in
+        // grpcFramerWriteHeaders via bpf_probe_write_user.
+        init_tp_ctx_parent_tp(t_ctx);
+        if (!create_trace_info(t_ctx, tp_p)) {
+            return SK_PASS;
+        }
+        if (!t_ctx->has_parent_tp) {
```
MrAlias: This early return looks like a behavioral regression for top-level plaintext HTTP/2 clients. `create_trace_info()` already handles the no-parent case by generating a fresh trace ID and zero parent ID, which matches the existing HTTP/1 client path.

Returning on `!t_ctx->has_parent_tp` means a monitored non-Go process making a root gRPC/HTTP2 call won’t get a CLIENT span or injected trace context at all. I think this path should allow root-span creation, not bail out.

bpf/tpinjector/tpinjector.c
```diff
@@ -968,3 +1102,396 @@ int obi_packet_extender_create_tp(struct sk_msg_md *msg) {
...
+
+    const u32 hpack_start = t_ctx->h2_hpack_offset;
+    const u32 hpack_len = t_ctx->h2_hpack_len;
+
+    tp_info_pid_t *tp_p = (tp_info_pid_t *)tp_info_mem();
+    if (!tp_p) {
+        return SK_PASS;
+    }
+
+    const u32 span_id_offset = find_existing_h2_traceparent(msg, hpack_start, hpack_len, &tp_p->tp);
```
MrAlias, line 1333: This path parses an existing HTTP/2 `traceparent`, but it no longer applies the forwarded-header correction that the HTTP/1 path does via `assign_parent_tp()`.

In the HTTP/1 flow, if the outgoing header is just forwarding the parent span, we rewrite the span ID so the child span remains distinct. Here we compute `span_id_offset`, but we never use it to update the HPACK value when that forwarded-header case is detected. That can leave the child span reusing the parent’s span ID on the wire and collapse the lineage.

MrAlias (changes requested):
I took another pass through the current revision and I still see a few correctness issues in the HTTP/2 propagation path. The main remaining problems are one concurrency hazard in the Go fallback path and two behavior mismatches against the existing HTTP/1 semantics.

bpf/gotracer/go_grpc.c
mmat11: I thought I had removed all fallbacks 🤦🏻 , good catch

bpf/tpinjector/tpinjector.c
```diff
@@ -710,61 +851,54 @@ int obi_packet_extender(struct sk_msg_md *msg) {
...
     t_ctx->niter = 0;
 
-    tp_info_pid_t *tp_pid = get_tp_info_pid(&e_key);
-
-    // Higher-level uprobes have already set the tp_pid for us (either Go, or SSL)
-    if (tp_pid) {
-        handle_existing_tp_pid(msg, id, &conn, &e_key, tp_pid);
+    // Plaintext H2: per-stream HPACK. Go bails — uprobe already injected.
+    if (is_h2_socket(msg)) {
+        if (!valid_pid(id)) {
```
MrAlias: This early return still looks wrong for already-marked plaintext HTTP/2 sockets.

Once `is_h2_socket(msg)` is true, we now bail out immediately on `!valid_pid(id)`. That skips the later `wrap_http2_traceparent()` fallback path below, even though this function's own comment says we still need that path for cases where the `sk_msg` context PID does not match the tracked PID.

In that scenario, a plaintext H2 connection that was already identified earlier can stop getting HPACK adoption/injection on subsequent sends from the same connection, which looks like a behavioral regression. I think this path should still run the H2 detection/injection flow for already-marked sockets, rather than returning early just because `valid_pid(id)` is false. Right?

bpf/gotracer/go_grpc.c
```diff
@@ -723,9 +735,30 @@ int obi_uprobe_grpcFramerWriteHeaders(struct pt_regs *ctx) {
...
 
     grpc_client_func_invocation_t *invocation = bpf_map_lookup_elem(&ongoing_streams, &key);
+    u64 conn_key = (u64)conn_ptr;
+    connection_info_t *conn_info = bpf_map_lookup_elem(&grpc_conn_ptr_to_conn, &conn_key);
 
     if (invocation) {
         bpf_dbg_printk("Found invocation info: %llx", invocation);
+
+        // Publish per-stream trace for sk_msg adoption.
+        if (conn_info && valid_trace(invocation->tp.trace_id)) {
```
MrAlias, line 767: This still looks like a fresh-stream race on the first HEADERS write.

`grpcFramerWriteHeaders` only injects/publishes trace context when `ongoing_streams[{conn_ptr,stream_id}]` is already present. If `WriteHeaders` runs before `NewStream_ret` publishes that entry, this path does nothing, so the first HEADERS frame on a newly-created stream can go out without OBI's `traceparent`.

I realize the doc already calls this out as a known limitation, so I don't necessarily think this has to block the rest of the work here. But if we want to consider the feature complete rather than landing with a known gap, I think this race still needs either a fix here or an explicit follow-up plan.

bpf/common/trace_lifecycle.h
```diff
@@ -42,6 +42,7 @@ static __always_inline void delete_client_trace_info(pid_connection_info_t *pid_
         .d_port = pid_conn->conn.d_port,
         .s_port = pid_conn->conn.s_port,
     };
+    sort_egress_key(&e_key);
```
MrAlias, line 45: This cleanup path still deletes only the connection-scoped `outgoing_trace_map` entry with `stream_id=0`, but this PR now writes per-stream HTTP/2 entries with nonzero `stream_id`.

I agree the stale-adoption case here is probably uncommon in practice because it needs port reuse plus stream ID reuse before the LRU entry is evicted, so I don't view this as a high-probability production bug. Still, the lifecycle looks incomplete now that the key semantics have widened from connection-scoped to stream-scoped.

I think this should either grow a matching per-stream delete path, or there should be a clear reason these entries are expected to live until LRU eviction.

bpf/generictracer/protocol_http2.h
```diff
@@ -56,47 +61,194 @@ static __always_inline u64 uniqueHTTP2ConnId(pid_connection_info_t *p_conn) {
...
+        // Override with Go uprobe's trace if available.
+        adopt_injected_trace(s_key, &tp_p->tp);
+        if (valid_trace(tp_p->tp.trace_id)) {
+            found_tp = 1;
+        }
+    } else {
+        found_tp =
+            find_trace_for_server_request(&s_key->pid_conn.conn, &tp_p->tp, EVENT_HTTP_REQUEST);
+        if (!found_tp) {
+            const unsigned char *hpack = h2g_info->data + k_h2_frame_header_len;
```
MrAlias: This kprobe-side parser still assumes the HPACK block starts immediately after the 9-byte HEADERS frame header and spans the rest of the captured payload.

Unlike the `sk_msg` path, this does not account for PADDED or PRIORITY HEADERS flags. If either flag is present, the traceparent scan can read the wrong region and miss or misparse incoming context. This may be uncommon for current test traffic, but it still looks like a parser mismatch between the two HTTP/2 paths.

MrAlias (dismissed):
I took another full pass through the current revision. The relay integration coverage does appear to be passing in CI, and the earlier issues around root plaintext HTTP/2 clients and forwarded-header correction look addressed.

I still see a few correctness gaps in the HTTP/2/gRPC propagation path. Two of them look like real behavior mismatches, and the other two are lower-severity cleanup / edge-case issues, but I still think they are worth tightening up before merge.

bpf/tpinjector/tpinjector.c
rafaelroquetto: Adding to that, I'd like to further understand in which scenario the sk_msg context PID does not match the tracked PID. This has not been my experience when writing the original code - sk_msg was always running on the context of the application (not on kernel/softirq contexts) and we do track child processes correctly afaik.

bpf/tpinjector/tpinjector.c
```diff
@@ -710,61 +851,54 @@ int obi_packet_extender(struct sk_msg_md *msg) {
...
-    // Valid PID only works for kprobes since Go programs don't add their
-    // PIDs to the PID map (we instrument the binaries), handled in the
-    // previous check
-    if (!valid_pid(id)) {
+    tp_info_pid_t *tp_pid = get_tp_info_pid(&e_key);
+
+    // Uprobe-provided tp (Go net/http, SSL). Run before size guard: small
+    // H2 frames may be below MIN_HTTP_SIZE but tp is already valid.
+    if (tp_pid) {
+        handle_existing_tp_pid(msg, id, &t_ctx->p_conn, &e_key, tp_pid);
```
rafaelroquetto: handle_existing_tp_pid should _not_ be injecting HTTP2 headers. It can't do it reliably, exemplified by the fact that it discards tp_info_pid and looks it up again. The original idea of this codepath is to inject `tp_pid` and bail - that only works for non-multiplexed protocols like HTTP1.

bpf/tpinjector/tpinjector.c
```diff
@@ -710,61 +851,54 @@ int obi_packet_extender(struct sk_msg_md *msg) {
...
         return SK_PASS;
     }
```
rafaelroquetto, line 919: here is where we should have something like

```
go_conn_info_t *g_info = find_go_conn_info(conn);

if (g_info && (g_info->is_ssl || g_info->is_go)) {
    return SK_PASS; // this was handled by the uprobes, we can't inject
}

// at this point, this is neither a go process, nor ssl, so check valid_pid(id) and handle accordingly

if (msg->size <= MIN_HTTP_SIZE>) {
    return SK_PASS;
}

if (!valid_pid(id)) {
    return SK_PASS;
}

// handle http1 and http2 -> this is non-go, non-ssl, plaintext.
```

rafaelroquetto (changes requested):
So, I did anothr pass only on tpinjector to get this going.

I still feel we need to adjust the flow of things, as the current design is very hard to reason about. The routing logic is non-linear and scattered - valid_pid appears in 3 different places with different semantics, and there's no single place where you can answer 'should sk_msg inject for this connection - to the point the current code "hides" an actual bug, namely if a non-go, non-ssl H2 connection takes place, we end up hitting wrap_http2_traceparent() -> which itself has no valid_pid check, and we inject the traceparent for any connection (even the comment on line 899 says so).

We should instead strive to stick with a more linear flow:

```
sk_msg (outgoing packet)
        │
        ├─ get_tp_info_pid(e_key{stream_id=0})
        │       found ──► handle_existing_tp_pid → inject & return
        │
        ├─ find_go_conn_info(conn)
        │       found && (is_go || is_ssl) ──► bail (uprobe owns it)
        │
        ├─ !valid_pid ──► bail (unmonitored process)
        │
        ├─ is_h2_socket ──► detect_h2 (known H2, skip preface check)
        │
        └─ detect H1 / H2 from scratch, inject & return
```

`is_h2_socket` is optional - it's just a "fastpath". But with the flow above, there are linear branches/forks to take/reason about, no feedback loops.

The premise is that:
 1. for non-GO ssl, we don't have explicit support for HTTP2 here in tpinjector (which is a known/existing limitation). The uprobe has setup something in outgoing_trace_map and we used it to inject as TCP options only, so we can't inject HTTP2 there.
 2. for go HTTP1 -> we take the same path (handle_existing_tp_pid)
 3. we need a new map with something like `go_conn_info_t` to allow us to bail when the go uprobes are injecting H2 instead of us. The generictracer also appears to do it correctly (I haven't throughly checked it). It's a gap to the implementation here.
 4. finally, we hit a fallback branch that handles any plaintext HTTP1 or HTTP2, and do our thing (checking valid_pid once). 

In this scheme, each item above is linear.

Also, I'd like to point out that the injection code does not handle multiple multiplexed streams - this affects non-Go plaintext H2 clients (Python, Java gRPC with concurrent streams). For Go, the uprobe handles each stream correctly.

rafaelroquetto:
@mmat11 regarding the tests, I think we are mixing two things.

The relay chain tests cover non-Go via sk_msg, but only single-stream. The multiplexed test (`testGRPCMultiplexedContextPropagation`) covers concurrency, but only with a go client, so it does not exercise the same path, and we are not testing sk_msg under concurrency.

We do not have coverage for concurrent HTTP/2 streams from a non-Go client, which is where the tricky cases are.

mmat11:
@MrAlias @rafaelroquetto I tried to address all comments:

- added generic hops to multiplexed test
- added uprobes to gotracer for enriching with the correct stream_id (which fixes the "race" which was mentioned in devdocs)
- refactored obi_packet_extender to be less funky and more readable
- duplicated valid_pids map, the issue in tpinjector seemed to be that when going from Go hops to generic hops, the pid didn't match; added a tp_valid_pids map with duplicated userspace code (to be hopefully refactored as a followup) which adds pids with pidtype Go|Generic
- added PADDED/PRIORITY handling (and added 2 additional tail calls due to added complexity which make CI fails on 5.15)

grcevski:
I haven't looked in detail yet, but why do we need another PID map? What kind of issue did you hit? From Go we use a map to tell what the traceparent is in the HTTP case, so we consult that instead anything PID related. If Go is a complication, maybe we can scope it out for now and add it later?

mmat11:
@grcevski which map in particular?

grcevski:
> @grcevski which map in particular?

Sorry I meant the tp_valid_pids. I guess I forgot to name it :)

mmat11:
> > @grcevski which map in particular?
> 
> Sorry I meant the tp_valid_pids. I guess I forgot to name it :)

it was because the pid filter in tpinjector wasn't matching go pids and the tests were breaking in the go -> non-go hops, maybe I can try to refactor the duplicated valid_pid code in this PR for clarity?

grcevski:
> > > @grcevski which map in particular?
> > 
> > 
> > Sorry I meant the tp_valid_pids. I guess I forgot to name it :)
> 
> it was because the pid filter in tpinjector wasn't matching go pids and the tests were breaking in the go -> non-go hops, maybe I can try to refactor the duplicated valid_pid code in this PR for clarity?

I see. We had that issue for HTTP too, but we opted in to setup a map that will provide the Go information: `outgoing_trace_map`. Would similar approach work? We check to see if that map has any info for us and act on it before any PID related checks. This code in go_nethttp.c:

```
                if (tls_state) {
                    // Clone and mark it invalid for the purpose of storing it in the
                    // outgoing trace map, if it's an SSL connection
                    tp_info_pid_t tp_p_invalid = {0};
                    __builtin_memcpy(&tp_p_invalid, &tp_p, sizeof(tp_p));
                    tp_p_invalid.valid = 0;
                    bpf_map_update_elem(&outgoing_trace_map, &e_key, &tp_p_invalid, BPF_ANY);
                } else {
                    bpf_map_update_elem(&outgoing_trace_map, &e_key, &tp_p, BPF_ANY);
                }

```

mmat11:
> > > > @grcevski which map in particular?
> > > 
> > > 
> > > Sorry I meant the tp_valid_pids. I guess I forgot to name it :)
> > 
> > 
> > it was because the pid filter in tpinjector wasn't matching go pids and the tests were breaking in the go -> non-go hops, maybe I can try to refactor the duplicated valid_pid code in this PR for clarity?
> 
> I see. We had that issue for HTTP too, but we opted in to setup a map that will provide the Go information: `outgoing_trace_map`. Would similar approach work? We check to see if that map has any info for us and act on it before any PID related checks. This code in go_nethttp.c:
> 
> ```
>                 if (tls_state) {
>                     // Clone and mark it invalid for the purpose of storing it in the
>                     // outgoing trace map, if it's an SSL connection
>                     tp_info_pid_t tp_p_invalid = {0};
>                     __builtin_memcpy(&tp_p_invalid, &tp_p, sizeof(tp_p));
>                     tp_p_invalid.valid = 0;
>                     bpf_map_update_elem(&outgoing_trace_map, &e_key, &tp_p_invalid, BPF_ANY);
>                 } else {
>                     bpf_map_update_elem(&outgoing_trace_map, &e_key, &tp_p, BPF_ANY);
>                 }
> ```

maybe I could try and copy that approach for grpc too

bpf/tpinjector/tpinjector.c
```diff
@@ -710,61 +901,64 @@ int obi_packet_extender(struct sk_msg_md *msg) {
...
+    t_ctx->h2_scan_pos = 0;
+    t_ctx->h2_frames = 0;
+
+    // Go gRPC clients: uprobe injects HPACK in the user buffer. Bail so
+    // sk_msg doesn't overwrite the per-stream outgoing_trace_map entry the
+    // uprobe set. Schedule TCP option out-of-band
+    if (is_go_connection(&t_ctx->p_conn)) {
+        bpf_msg_pull_data(msg, 0, msg->size, 0);
+        fill_msg_buffers(msg, &t_ctx->p_conn, &e_key);
+        if (inject_flags & k_inject_tcp_options) {
```
rafaelroquetto: so, this whole block is redundant - for HTTP2/GRPC connections, we don't want to inject anything - this happens silently here just because `e_key` with a `stream_id = 0` key, in which case the `get_tp_info_pid` call will always return 0. Because is_go_connection seems only to be applied to http2/grpc, the entire call to get_tp_info_pid and schedule_wrte_tcp_option is dead code.

Also note: `is_go_connection` is misleading, because it's only marked as such for gRPC. Took me a bit to understand that - but see my next comment below.

bpf/tpinjector/tpinjector.c
```diff
@@ -710,61 +901,64 @@ int obi_packet_extender(struct sk_msg_md *msg) {
...
+        }
+        return SK_PASS;
+    }
 
-    tp_info_pid_t *tp_pid = get_tp_info_pid(&e_key);
-
-    // Higher-level uprobes have already set the tp_pid for us (either Go, or SSL)
-    if (tp_pid) {
-        handle_existing_tp_pid(msg, id, &conn, &e_key, tp_pid);
+    if (!tp_valid_pid(id)) {
```
rafaelroquetto: I am not sure this is the way. This is best solved by changing the ordering of things like I had proposed in the previous review iteration:

 ```
 sk_msg (outgoing packet)
        │
        ├─ get_tp_info_pid(e_key{stream_id=0})
        │       found ──► handle_existing_tp_pid → inject & return
        │
        ├─ find_go_conn_info(conn)
        │       found && (is_go || is_ssl) ──► bail (uprobe owns it)
        │
        ├─ !valid_pid ──► bail (unmonitored process)
        │
        ├─ is_h2_socket ──► detect_h2 (known H2, skip preface check)
        │
        └─ detect H1 / H2 from scratch, inject & return
```

if you follow this ordering, then what happens is:

 - get_tp_info_pid is the only place where we schedule tcp or ip options injection both for SSL and for go HTTP1, nothing changes here, business as usual, no pid map, central dispatch for this case. It's also easier to reason -> someone sets up outgoing_trace_map, it means inject that (based on `inject_flags`) and call it a day. No need to go deep into the code to understand all the non-orthogonal mechanics.
 - you now have  `is_go_connection` - instead, change that to be more generic - just a struct that carries whether this is a `go connection` or an `SSL connection`. If we arrived at this point, and either of these are true, we bail - if this had been Go HTTP1, or HTTP1/SSL, the `get_tp_info_pid` branch above would have dealt with it - so getting here means that we can't inject anything so we fill in the msg buffers and bail
 - finally, we do a regular valid_pid check - at this stage we are past any go stuff or SSL, so we fall back to regular plain-text handling, header scanning, etc.

File: bpf/tpinjector/tpinjector.c [hunk 37]
```diff
@@ -710,61 +901,64 @@ int obi_packet_extender(struct sk_msg_md *msg) {
...
-    const u8 is_http = protocol_detector(msg, id, &conn, &e_key);
+    bpf_dbg_printk("MSG=%llx:%d ->", conn.s_ip[3], conn.s_port);
+    bpf_dbg_printk("MSG TO=%llx:%d", conn.d_ip[3], conn.d_port);
+    bpf_dbg_printk("MSG SIZE=%u", msg->size);
 
-    // at this point, we can't handle anything other than HTTP, as we need to be able
-    // to tell whether this is the start of a new request
-    if (!is_http) {
+    bpf_msg_pull_data(msg, 0, msg->size, 0);
+    const bool is_http = protocol_detector(msg, id, &conn, &e_key);
```
rafaelroquetto: so, now that we did split the filling of the msg_buffers into fill_msg_buffer, we shouldn't call it inside protocol detector - instead we should call it once after the call to `valid_pid`, before `is_h2_socket`

That also means a single bpf_msg_pull_data(msg, 0, msg->size, 0) which, come to think of it, will simplify any downstream calls.

rafaelroquetto (changes requested):
I just reviewed the tpinjector side of things. Once we tackle those issues, I will do another pass to generictracer and hopefully we can merge it soon.

bpf/tpinjector/tpinjector.c
mmat11: tried this ordering but it breaks the test since outgoing_trace_map[stream_id=0] isn't always empty on plaintext h2; keeping is_h2_socket before handle_existing_tp_pid would prevent these issues:

```
//   obi_packet_extender (sk_msg entry)
//   │
//   ├── is_go_grpc_client_conn?      ─┐  Go gRPC: uprobe wrote HPACK in
//   │     └── fill_msg_buffers        │  user buffer; sk_msg bails
//   │     └── SK_PASS                 │
//   │                                 │
//   ├── is_h2_socket?                 │  Plaintext H2 (non-Go): per-stream
//   │     └─ tail-call detect_h2      │  HPACK chain (right column). Runs
//   │                                 │  before tp_pid so HEADERS frames
//   │                                 │  don't go through HTTP/1 path
//   │                                 │
//   ├── tp_pid present?               │  Go net/http, SSL: TCP option +
//   │     └── handle_existing_tp_pid  │  HTTP/1 inject. Bypasses valid_pid
//   │                                 │
//   ├── !valid_pid → SK_PASS          │
//   │                                 │
//   ├── HTTP/1 detected?              │  ── find_existing_tp ── create_tp ──
//   │                                 │     write_msg_traceparent
//   │                                 │
... h2 handling
```

and when is_h2_socket has been marked as true, the `valid_pid()` check has already been done earlier

bpf/tpinjector/tpinjector.c
mmat11: > That also means a single bpf_msg_pull_data(msg, 0, msg->size, 0) which, come to think of it, will simplify any downstream calls.

done

bpf/tpinjector/tpinjector.c
rafaelroquetto: >it breaks the test since outgoing_trace_map[stream_id=0] isn't always empty on plaintext h2

do you know why that is? because I feel by reordering things like you are proposing we are just maskign / working around the actual bug, so I'd like to better understand this.

There are only a handful of places in which we set `stream_id=0` for the egress key, and I am guessing the issue is in `http2_grpc_start` - if you look at it, it does the following:


```
    server_or_client_trace(
        EVENT_HTTP_CLIENT, &h2g_info->conn_info, k_lw_thread_none, tp_p, ssl, orig_dport);

    egress_key_t e_key = {
        .d_port = h2g_info->conn_info.d_port,
        .s_port = h2g_info->conn_info.s_port,
        .stream_id = s_key->stream_id,
    };
    sort_egress_key(&e_key);
    // BPF_NOEXIST: don't overwrite a Go uprobe's HPACK-injected span (written=1)
    bpf_map_update_elem(&outgoing_trace_map, &e_key, tp_p, BPF_NOEXIST);
```

i.e. it calls `server_or_client_trace`, which in turn calls for a client request:

```
        egress_key_t e_key = {
            .d_port = conn->d_port,
            .s_port = conn->s_port,
            .stream_id = 0, // <---- THIS!!
        };
        sort_egress_key(&e_key);

        if (ssl) {
            /* ... */
        } else {
            bpf_map_update_elem(&outgoing_trace_map, &e_key, tp_p, BPF_ANY);
            obi_ctx__set(id, &tp_p->tp);
        }
```

and then right afterwards when it returns we do:

```
bpf_map_update_elem(&outgoing_trace_map, &e_key, tp_p, BPF_NOEXIST);
```
so we insert twice - one for stream_id 0 and one for the right stream id, and we leave the stream_id=0 behind as potential garbage.

rafaelroquetto: I forgot to mention - it may make sense to actually change `server_or_client_trace` to receive `stream_id` and pass it a long, but don't quote me on that (other protocols like HTTP1 would just pass `0`)

mmat11: that worked, thanks a lot for the patience 🙏🏻

mmat11: updated flow -> https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1832/changes/2adbb69c723ee0122baa5593b8079380620cf389

bpf/tpinjector/tpinjector.c
```diff
@@ -710,61 +890,54 @@ int obi_packet_extender(struct sk_msg_md *msg) {
...
     t_ctx->p_conn.conn = conn;
+    sort_connection_info(&t_ctx->p_conn.conn);
     t_ctx->p_conn.pid = pid_from_pid_tgid(id);
     t_ctx->e_key = e_key;
     t_ctx->niter = 0;
+    t_ctx->h2_scan_pos = 0;
+    t_ctx->h2_frames = 0;
 
+    // Single pull + fill so downstream branches reuse the same buffer
+    bpf_msg_pull_data(msg, 0, msg->size, 0);
```
rafaelroquetto: just a minor nudge here: we should only do this when we are sure that we are tracking / instrumenting the process/socket - if we leave this here, we will pull data for _every_ `sk_msg_md` in the egress codepath and that can get _very slow_ (bpf_msg_pull_data causes the kernel to linearise a potentially scattered list of sk_buffers under the hood).

So instead, we should only pull data and fill the buffers when we are sure we are tracking the process - that would be when:

 - handle_existing_tp_pid returns true (before we just pull the necessary data size for the checks here)
 - is_go_grpc_client_conn returns true (I suppose, but double check)
 - after the check for `valid_pid(id)` on line 915 (i.e. beyonmd that we are tracking the process and we should fill in the buffers/pull data)

This avoids unnecessarily filling in the buffers for any other processes that cross this codepath.

bpf/tpinjector/tpinjector.c
```diff
@@ -710,61 +890,54 @@ int obi_packet_extender(struct sk_msg_md *msg) {
...
 
-    // At this stage, there were no previously TP information setup - it's the first
-    // time we are seeing this packet - so we need to detect whether this is the start
-    // of a new request and perform any injection if so.
-    // Valid PID only works for kprobes since Go programs don't add their
-    // PIDs to the PID map (we instrument the binaries), handled in the
-    // previous check
     if (!valid_pid(id)) {
         return SK_PASS;
     }
```
rafaelroquetto, line 918: so after this line, we can uncondiotnally pull data and fill the buffers

rafaelroquetto (commented):
This now looks good! Just one last nit and then I think we are good to go!

Thank you for all the effort and patience :)

bpf/tpinjector/tpinjector.c
mmat11: done in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1832/changes/947af1d7d021dcb6d3ef5058d23668d87ae706c1

rafaelroquetto (approved):
LGTM! Niceeeeeee!

grcevski:
:rocket:

NameHaibinZhang:
"Hi, is this PR ready to be merged? I've run into a gRPC context propagation issue and would appreciate it if we could get this merged into the main branch as soon as possible.

mmat11:
> "Hi, is this PR ready to be merged? I've run into a gRPC context propagation issue and would appreciate it if we could get this merged into the main branch as soon as possible.

this will probably get merged next week after the next release

PR #1834, opened 2026-04-14
[WIP - do not review] New socket tracer

rafaelroquetto:
## Summary

TODO

## Validation


grcevski:
Do you think we should continue to pursue this path? With the recent kernel bug we found related to the ingress path, we found out that the socket ingress path is very buggy, the kernel bug was introduced as an attempt to fix another bug. Maybe it's unrelated, but I wanted to ask. When I made a PR to attach a dummy ingress verdict program, many old kernel tests failed, anything 6.12+ seemed OK, but many below failed with random EOF file reached for various socket connections, especially on arm.

rafaelroquetto:
@grcevski not really, I agree with you. There's another issue that is very hard to tackle with this approach: pre-existing HTTP2/GRPC connections for which we have already missed the dynamic table.

There is good scrap metal in this PR when it comes to socket lifecyle handling, if it ever become useful in the future.

Closing.

PR #1849, opened 2026-04-15
fix(http): cap decompressed response bodies

MrAlias:
Caps decompressed HTTP response bodies at 64 KiB in the shared response parsing path.

This prevents compressed response payloads from expanding without bound when HTTP body enrichment or other HTTP response parsers read compressed bodies in userspace.

This has no impact on released versions, as the feature has yet to be released.

## Changes

- add a hard 64 KiB limit for decompressed HTTP response bodies
- return a clear error when decompressed content exceeds the limit
- keep the original response body readable after decompression errors
- add regression tests covering oversized compressed bodies

## Rationale

HTTP response capture is already bounded on the compressed side, but decompression previously used unbounded `io.ReadAll` calls for `gzip`, `zstd`, `deflate`, and `br`.

That allowed a decompression bomb to expand well beyond the configured HTTP body capture budget once body enrichment was enabled.

Keeping the decompressed limit aligned with the maximum captured HTTP body size closes that gap without changing the existing feature surface.

### DIFF
--- a/docs/config-schema.json
+++ b/docs/config-schema.json
@@ -341,7 +341,7 @@
         }
       },
       "type": "object",
-      "description": "Per-protocol maximum bytes to capture per request per direction, sent to userspace via large buffer events. The lte=65536 ceiling must match the k_large_buf_max_*_captured_bytes constants in bpf/common/large_buffers.h.  Default: 0 (disabled)."
+      "description": "Per-protocol maximum bytes to capture per request per direction, sent to userspace via large buffer events. MaxCapturedPayloadBytes must match the k_large_buf_max_*_captured_bytes constants in bpf/common/large_buffers.h. The validate tags below must stay aligned with MaxCapturedPayloadBytes.  Default: 0 (disabled)."
     },
     "EBPFTracer": {
       "properties": {
--- a/docs/config-schema.json
+++ b/docs/config-schema.json
@@ -341,7 +341,7 @@
         }
       },
       "type": "object",
-      "description": "Per-protocol maximum bytes to capture per request per direction, sent to userspace via large buffer events. MaxCapturedPayloadBytes must match the k_large_buf_max_*_captured_bytes constants in bpf/common/large_buffers.h. The validate tags below must stay aligned with MaxCapturedPayloadBytes.  Default: 0 (disabled)."
+      "description": "Per-protocol maximum bytes to capture per request per direction, sent to userspace via large buffer events. Values must stay aligned with MaxCapturedPayloadBytes and the k_large_buf_max_*_captured_bytes constants in bpf/common/large_buffers.h.  Default: 0 (disabled)."
     },
     "EBPFTracer": {
       "properties": {
--- a/pkg/config/ebpf_tracer_test.go
+++ b/pkg/config/ebpf_tracer_test.go
@@ -261,8 +261,6 @@ func TestEBPFTracer_CudaInstrumentationEnabled(t *testing.T) {
 }
 
 func TestEBPFBufferSizesValidateTagsMatchMaxCapturedPayloadBytes(t *testing.T) {
-	t.Helper()
-
 	expected := fmt.Sprintf("lte=%d", MaxCapturedPayloadBytes)
 	typ := reflect.TypeOf(EBPFBufferSizes{})
 
--- a/pkg/config/ebpf_tracer_test.go
+++ b/pkg/config/ebpf_tracer_test.go
@@ -4,6 +4,8 @@
 package config
 
 import (
+	"fmt"
+	"reflect"
 	"testing"
 )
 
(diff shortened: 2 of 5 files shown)


pkg/ebpf/common/http/responses.go
```diff
@@ -15,6 +15,16 @@ import (
 	"github.com/klauspost/compress/zstd"
 )
 
+// Keep the decompressed response cap aligned with the maximum captured HTTP body
+// size so body enrichment cannot expand a compressed payload beyond the configured
+// userspace budget.
+const maxDecompressedResponseBodyBytes = 64 * 1024
```
grcevski: This should match the maximum large buffer size we allow lte=65536, which it does here, but we need to somehow remember to change both. the lte=65536 is part of the config definition:
```
type EBPFBufferSizes struct {
	HTTP     uint32 `yaml:"http" env:"OTEL_EBPF_BPF_BUFFER_SIZE_HTTP" validate:"lte=65536"`
	MySQL    uint32 `yaml:"mysql" env:"OTEL_EBPF_BPF_BUFFER_SIZE_MYSQL" validate:"lte=65536"`
	Kafka    uint32 `yaml:"kafka" env:"OTEL_EBPF_BPF_BUFFER_SIZE_KAFKA" validate:"lte=65536"`
	Postgres uint32 `yaml:"postgres" env:"OTEL_EBPF_BPF_BUFFER_SIZE_POSTGRES" validate:"lte=65536"`
}
```

MrAlias: Updated this to reduce drift on the Go side.

Changes in this branch:
- introduced `config.MaxCapturedPayloadBytes` as the shared Go constant for the capture ceiling
- changed the HTTP decompression cap to derive from that constant instead of hardcoding `64 * 1024`
- added a unit test in `pkg/config/ebpf_tracer_test.go` that fails if any `EBPFBufferSizes` `validate:"lte=..."` tag drifts from `MaxCapturedPayloadBytes`

The test failure message also calls out the exact places that need to be updated together:
- the `EBPFBufferSizes` validate tag
- `MaxCapturedPayloadBytes`
- the matching `k_large_buf_max_*_captured_bytes` constant in `bpf/common/large_buffers.h`

The remaining literal is still the struct tag, since that validation mechanism cannot reference a constant directly, so I used the test to guard that coupling explicitly.

pkg/ebpf/common/http/responses.go
```diff
@@ -41,28 +52,52 @@ func getResponseBody(resp *http.Response) ([]byte, error) {
 // decompressBody decompresses b according to the Content-Encoding value.
 // Mirrors what http.Transport does for gzip, extended to cover zstd, deflate and brotli.
```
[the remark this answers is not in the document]
MrAlias: Updated. The `decompressBody` comment now documents the behavior change explicitly: unsupported encodings are returned unchanged, decompressed output is capped at `maxDecompressedResponseBodyBytes`, and the function can return `errResponseBodyTooLarge` when that cap is exceeded.

pkg/config/ebpf_tracer_test.go
```diff
@@ -257,3 +259,27 @@ func TestEBPFTracer_CudaInstrumentationEnabled(t *testing.T) {
 		})
 	}
 }
+
+func TestEBPFBufferSizesValidateTagsMatchMaxCapturedPayloadBytes(t *testing.T) {
+	t.Helper()
+
```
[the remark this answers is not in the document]
MrAlias: Updated. I removed the top-level `t.Helper()` from `TestEBPFBufferSizesValidateTagsMatchMaxCapturedPayloadBytes` so failure reporting stays on the test itself rather than being shifted as if it were a shared helper.

docs/config-schema.json
```diff
@@ -341,7 +341,7 @@
         }
       },
       "type": "object",
-      "description": "Per-protocol maximum bytes to capture per request per direction, sent to userspace via large buffer events. The lte=65536 ceiling must match the k_large_buf_max_*_captured_bytes constants in bpf/common/large_buffers.h.  Default: 0 (disabled)."
+      "description": "Per-protocol maximum bytes to capture per request per direction, sent to userspace via large buffer events. MaxCapturedPayloadBytes must match the k_large_buf_max_*_captured_bytes constants in bpf/common/large_buffers.h. The validate tags below must stay aligned with MaxCapturedPayloadBytes.  Default: 0 (disabled)."
```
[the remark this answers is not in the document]
MrAlias: Updated. I reworded the source comment so the generated schema no longer refers to Go-specific concepts like "validate tags below". The schema description now stays consumer-facing while still calling out that the values must remain aligned with the corresponding eBPF large-buffer limits.

pkg/ebpf/common/http/responses.go
```diff
@@ -39,30 +51,57 @@ func getResponseBody(resp *http.Response) ([]byte, error) {
...
 	}
+
+	if err != nil {
+		return nil, err
+	}
+	if closeFn != nil {
+		defer closeFn()
+	}
+
+	return readBodyWithLimit(reader, maxDecompressedResponseBodyBytes)
```
mmat11, line 93: question: in the case of http large buffer size set to 8192 in config, this will never error out, right? but the body will still be broken if the original payload was too big. Is this change driven by some performance optimization?

MrAlias: No, this prevents compressed response payloads from expanding without bound when HTTP body enrichment or other HTTP response parsers read compressed bodies in userspace.

HTTPEnricher currently reads response bodies via getResponseBody for generic HTTP body extraction; getResponseBody performs unbounded decompression, enabling decompression-bomb DoS against the agent when body enrichment is enabled. Because HTTP body enrichment can now be enabled for all JSON traffic, a remote attacker can serve a highly compressible JSON response (or proxy a response) that expands far beyond the captured size, causing excessive memory/CPU usage and potentially OOM-killing the privileged agent.

mmat11 (approved):

grcevski (approved):
LGTM!

PR #1855, opened 2026-04-16
fix(tests): use CollectT inside EventuallyWithT to fix flaky retries

skl:
## Summary

Several integration tests used the outer `*testing.T` for assertions inside `require.EventuallyWithT` callbacks instead of the `*assert.CollectT` parameter. This caused assertions to fail the test immediately instead of allowing `EventuallyWithT` to retry, which is the top source of test flakiness — the `exit-error` fingerprint accounts for 23 of 38 failure occurrences in the [daily CI analysis report](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/24460707528?pr=1807).

The fix changes `t` → `ct` for assertions in three files:
- `traces_test.go`: Python and Rails span checks in `testNestedHTTPTracesKProbes` (14 assertion calls)
- `multiprocess_test.go`: Python and Rails span checks in `testPartialLanguageHTTPProbes` (2 assertion calls)
- `test_python_elasticsearchclient.go`: Span count check in `assertElasticsearchOperation` (1 assertion call)

## Validation


NimrodAvni78 (approved):
NICE!

mariomac (approved):
Good catch! I guess we should have just invoked Eventually as:
```go
require.EventuallyWithT(t, func(t *assert.CollectT) {
```

and let inner `t *assert.CollectT` shadowing the outer `t *testing.T`.

mmat11 (approved):

PR #1857, opened 2026-04-16
feat(lint): add collectt analyzer to detect misuse of *testing.T in EventuallyWithT callbacks

skl:
## Summary

- Adds a `go/analysis`-based linter (`internal/test/analyzer/collectt/`) that flags `assert`/`require` calls inside `EventuallyWithT` callbacks that pass the outer `*testing.T` instead of the callback's `*assert.CollectT` parameter
- Integrated into `make lint` via a new `lint-collectt` Makefile target

## Context

Passing `*testing.T` instead of `*assert.CollectT` defeats the retry mechanism and causes immediate test failure — this was the root cause of 61% of flakiness in the `TestMultiProcess*` test variants (fixed in #1855). This analyzer prevents the bug from recurring.

> [!IMPORTANT]
> The `lint-collectt` check will fail until #1855 merges, as the existing violations it detects are fixed by that PR.

## Limitations

- Only checks package-level `assert.Foo(t, ...)` / `require.Foo(t, ...)` calls, not method calls on assertion objects (e.g. `a := assert.New(t); a.Equal(...)`)
- Only checks function literal callbacks, not named functions passed to `EventuallyWithT`

The codebase currently uses the package-level form.

Assisted-by: Claude:claude-opus-4-6


Makefile
```diff
@@ -152,6 +152,11 @@ lint-dependency-policy:
 		./scripts/lint-dependency-policy.sh; \
 	fi
 
+.PHONY: lint-collectt
+lint-collectt:
```
mmat11, line 156: should this be ran in CI as well?

skl: Yeah the existing lint job picks it up, you can see the failures before I merged main 😊

grcevski (approved):
LGTM!

mmat11 (approved):

PR #1866, opened 2026-04-16
Gracefully handle unavailable /proc/pid/exe

rafaelroquetto:
## Summary

On hardened systems with restrictive ptrace_scope or SELinux, reading `/proc/<pid>/exe` returns `EACCES`.

We now treat `os.ErrPermission` like `os.ErrNotExist` and continue with `exePath="unknown"`, so that the functionality provided by kprobes still work.

Also fixes Oracle Linux (ID="ol") not being recognized as RHEL-based, causing a spurious "kernel version not supported" error on 4.18 UEK kernels.

Closes https://github.com/grafana/beyla/issues/2712.

## Validation

### DIFF
--- a/internal/test/integration/docker-compose-exeperm.yml
+++ b/internal/test/integration/docker-compose-exeperm.yml
@@ -1,0 +1,50 @@
+services:
+  testserver:
+    build:
+      context: ../../..
+      dockerfile: internal/test/integration/components/nodejsserver/Dockerfile
+    image: hatest-testserver-node
+    ports:
+      - "3030:3030"
+
+  obi:
+    build:
+      context: ../../..
+      dockerfile: ./internal/test/integration/components/obi/Dockerfile
+    command:
+      - --config=/configs/obi-config-exeperm.yml
+    volumes:
+      - ./configs/:/configs
+      - ./system/sys/kernel/security_none:/sys/kernel/security
+      - /sys/fs/cgroup:/sys/fs/cgroup
+      - ../../../testoutput:/coverage
+      - ../../../testoutput/run-exeperm:/var/run/obi
+    image: hatest-obi
+    privileged: true
+    network_mode: "service:testserver"
+    pid: "service:testserver"
+    environment:
+      GOCOVERDIR: "/coverage"
+      OTEL_EBPF_TRACE_PRINTER: "text"
+      OTEL_EBPF_DISCOVERY_POLL_INTERVAL: 500ms
+      OTEL_EBPF_METRICS_INTERVAL: "10ms"
+      OTEL_EBPF_OTLP_TRACES_BATCH_TIMEOUT: "1ms"
+      OTEL_EBPF_BPF_BATCH_TIMEOUT: "10ms"
+      OTEL_EBPF_LOG_LEVEL: "DEBUG"
+      OTEL_EBPF_HOSTNAME: "obi"
+      OTEL_EBPF_SIMULATE_EXE_PERM_DENIED: "true"
+    depends_on:
+      testserver:
+        condition: service_started
+      jaeger:
+        condition: service_started
+
+  jaeger:
+    image: jaegertracing/all-in-one:1.60@sha256:4fd2d70fa347d6a47e79fcb06b1c177e6079f92cba88b083153d56263082135e
+    ports:
+      - "16686:16686"
+      - "4317"
+      - "4318"
+    environment:
+      - COLLECTOR_OTLP_ENABLED=true
+      - LOG_LEVEL=debug
--- a/pkg/appolly/discover/matcher.go
+++ b/pkg/appolly/discover/matcher.go
@@ -517,6 +517,10 @@ func logDeprecationAndConflicts(cfg *obi.Config) {
 	}
 }
 
+// simulateExePermDenied simulates EACCES on /proc/<pid>/exe for integration testing.
+// Set OTEL_EBPF_SIMULATE_EXE_PERM_DENIED=true to enable.
+var simulateExePermDenied = os.Getenv("OTEL_EBPF_SIMULATE_EXE_PERM_DENIED") == "true"
+
 // replaceable function to allow unit tests with faked processes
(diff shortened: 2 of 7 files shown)


grcevski (approved):
LGTM!

pkg/appolly/discover/matcher.go
```diff
@@ -517,21 +517,33 @@ func logDeprecationAndConflicts(cfg *obi.Config) {
 	}
 }
 
+// simulateExePermDenied simulates EACCES on /proc/<pid>/exe for integration testing.
+// Set OTEL_EBPF_SIMULATE_EXE_PERM_DENIED=true to enable.
+var simulateExePermDenied = os.Getenv("OTEL_EBPF_SIMULATE_EXE_PERM_DENIED") == "true"
+
```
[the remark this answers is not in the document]
rafaelroquetto: nope

pkg/appolly/discover/matcher.go
```diff
@@ -517,21 +517,33 @@ func logDeprecationAndConflicts(cfg *obi.Config) {
...
-	if err != nil {
-		if errors.Is(err, os.ErrNotExist) {
-			// this might happen if you query from the port a service that does not have executable path.
-			// Since this value is just for attributing, we set a default placeholder
-			exePath = "unknown"
-		} else {
-			return nil, fmt.Errorf("can't read /proc/<pid>/fd information: %w", err)
+	exePath := "unknown"
+	if !simulateExePermDenied {
+		exePath, err = proc.Exe()
```
MrAlias, line 533: The branch below keeps a process when `proc.Exe()` fails with `os.ErrPermission` by setting `ExePath` to `"unknown"`, but discovery still requires opening `/proc/<pid>/exe` later in `findExecElf()`. On hosts with restrictive `ptrace_scope` or LSM policy, `elf.Open("/proc/<pid>/exe")` will fail with the same permission error and `FilterClassify()` will drop the process before instrumentation.

This means, the main behavior change is not complete in real permission-denied environments. Also, the new warning message is accurate, but the service still does not survive the full discovery pipeline.

Adding a matching fallback in the ELF/typer path, or redesign the flow so port-only discovery does not require opening `/proc/<pid>/exe` when that path is unreadable should address this.

internal/test/integration/docker-compose-exeperm.yml
```diff
@@ -0,0 +1,50 @@
...
+    environment:
+      GOCOVERDIR: "/coverage"
+      OTEL_EBPF_TRACE_PRINTER: "text"
+      OTEL_EBPF_DISCOVERY_POLL_INTERVAL: 500ms
+      OTEL_EBPF_METRICS_INTERVAL: "10ms"
+      OTEL_EBPF_OTLP_TRACES_BATCH_TIMEOUT: "1ms"
+      OTEL_EBPF_BPF_BATCH_TIMEOUT: "10ms"
+      OTEL_EBPF_LOG_LEVEL: "DEBUG"
+      OTEL_EBPF_HOSTNAME: "obi"
+      OTEL_EBPF_SIMULATE_EXE_PERM_DENIED: "true"
```
MrAlias, line 35: The integration setup sets `OTEL_EBPF_SIMULATE_EXE_PERM_DENIED=true`, and the code uses that env var to skip the real `proc.Exe()` call entirely. That means the test validates a test-only code path rather than a real `/proc/<pid>/exe` permission failure. It also does not cover the later `findExecElf()` open, which is the remaining failure point.

Is there a way to remove the production env-var bypass and test this through a small injectable boundary around exe/ELF access, or reproduce the real permission-denied condition in integration? Ideally, this will add coverage for the full path through classification, not only matcher acceptance.

pkg/obi/os.go
```diff
@@ -42,7 +42,7 @@ func parseOSReleaseIsRHEL(data []byte) bool {
 		}
 		val = strings.Trim(val, "\"'")
 		if strings.Contains(val, "rhel") || strings.Contains(val, "centos") ||
-			strings.Contains(val, "rocky") || strings.Contains(val, "alma") {
+			strings.Contains(val, "rocky") || strings.Contains(val, "alma") || val == "ol" {
```
MrAlias, line 45: This seems unrelated to the permission issue being addressed in the PR. Should it be its own PR?

pkg/appolly/discover/matcher.go
```diff
@@ -517,21 +517,33 @@ func logDeprecationAndConflicts(cfg *obi.Config) {
 	}
 }
 
+// simulateExePermDenied simulates EACCES on /proc/<pid>/exe for integration testing.
+// Set OTEL_EBPF_SIMULATE_EXE_PERM_DENIED=true to enable.
+var simulateExePermDenied = os.Getenv("OTEL_EBPF_SIMULATE_EXE_PERM_DENIED") == "true"
```
mmat11, line 522: nit: should we name this something like `skipProcExeAttr` or similar?

pkg/appolly/discover/matcher.go
```diff
@@ -517,21 +517,33 @@ func logDeprecationAndConflicts(cfg *obi.Config) {
...
-		if errors.Is(err, os.ErrNotExist) {
-			// this might happen if you query from the port a service that does not have executable path.
-			// Since this value is just for attributing, we set a default placeholder
-			exePath = "unknown"
-		} else {
-			return nil, fmt.Errorf("can't read /proc/<pid>/fd information: %w", err)
+	exePath := "unknown"
+	if !simulateExePermDenied {
+		exePath, err = proc.Exe()
+		if err != nil {
```
mmat11, line 534: nit: maybe using a switch would make this more readable

mmat11 (approved):

rafaelroquetto:
Everyone, thank you for the reviews! Please hold on merging this as I am yet to address comments from @MrAlias and @mmat11. I've also decided to run a few tests on Oracle Linux to see if this suffices, so bear with me.

rafaelroquetto:
I am temporarily closing this as I need to collect more info on the relation issue, I might reopen it later.

Issue #1846, opened 2026-04-15
Use Generic tracer protocol parsing for Go programs


grcevski:
We have a number of Go instrumentations that are using the `uprobe` approach, where specific libraries are targeted, handled by reading information at the Go generated code level using field offset mappings. 

This approach works well in generally for Go, because a lot of frameworks use the Go standard libraries under the covers. However, there are Go frameworks out there that have re-implemented the protocol stack themselves, or the protocols are no in the Go standard library. 

Instead of implementing every Go library support with uprobes, which is labour intensive and error prone, we'd like to implement targeted Go library instrumentation when it makes sense, but for other cases rely on the protocol parsing that we already for for other languages. For example we don't want to implement support for every Go Kafka client library out there.

For this purpose we have to implement the following changes:

- [X] Allow use of the generic protocol handling though buffers captured by `netFD.Read` and `netFD.Write`.
- [X] Implement support for context propagation handling between the Go tracer and the generic tracer.
- [ ] Allow use of generic TLS protocol handling through buffers captured by `crypto/tls.(*Conn).Write` and `crypto/tls.(*Conn).Read`.
- [ ] Use large buffers for HTTP events captured by the existing Go probes, so that we can enable payload extraction.

PR #1867, opened 2026-04-17
feat(mcp): add mcp support

NameHaibinZhang:
## Summary

Add observability support for the **Model Context Protocol (MCP)** — the emerging standard for communication between AI agents and tool/resource servers.

MCP runs over HTTP using JSON-RPC 2.0. This change enables OBI to automatically detect, parse, and export MCP spans with rich semantic attributes following the OpenTelemetry MCP semantic conventions (semconv v1.40.0).

### Key changes

- **Protocol detection** (`pkg/ebpf/common/http/mcp.go`): Identifies MCP traffic by parsing the JSON-RPC request body and matching method names (`tools/call`, `resources/read`, `prompts/get`, `initialize`, etc.) with a `jsonrpc: "2.0"` version check. Falls back to the `Mcp-Session-Id` header for unknown extension methods. Extracts tool name, resource URI, prompt name, session ID, protocol version, request ID, and error information from both request and response bodies.

- **Span enrichment** (`pkg/appolly/app/request/span.go`): Adds `HTTPSubtypeMCP` subtype, `MCPCall` struct, and `OperationName()` for GenAI operation naming (`tools/call` → `execute_tool`). Extends `SpanStatusCode` and `SpanStatusMessage` to treat MCP JSON-RPC errors (non-zero error code in HTTP 200 responses) as span errors, matching existing JSON-RPC behavior.

- **HTTP transform integration** (`pkg/ebpf/common/http_transform.go`): Wires MCP detection into the HTTP span pipeline, gated by the MCP enabled flag. Placed before JSON-RPC detection so MCP-specific semantics take priority.

- **OTel attribute export** (`pkg/export/otel/tracesgen/tracesgen.go`, `pkg/export/attributes/names/attrs.go`): Emits `mcp.method.name`, `mcp.session.id`, `mcp.resource.uri`, `mcp.protocol.version`, `gen_ai.tool.name`, `gen_ai.prompt.name`, `gen_ai.operation.name`, `jsonrpc.request.id`, `rpc.response.status_code`, and `error.message` attributes.

- **Configuration** (`pkg/config/payload_extraction.go`, `pkg/obi/config.go`): Added `MCPConfig` with `OTEL_EBPF_HTTP_MCP_ENABLED` env var, disabled by default. `GenAIConfig.Enabled()` includes MCP.

- **Tests**: Comprehensive unit tests covering all MCP method types, error responses, session header fallback, non-MCP rejection (missing `jsonrpc: "2.0"`, unknown methods, non-POST, invalid JSON), span status/message for MCP errors, `GenAIConfig.Enabled()` coverage, and OTel trace generation for MCP success/error spans (server and client).

## Validation


NimrodAvni78:
@NameHaibinZhang we recently added JSONRPC detection, did you  consider using the same logic from there for parsing and detection?

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -86,22 +86,30 @@ func pidSegmentBit(k uint64) (uint32, uint32) {
 
 func (p *Tracer) buildPidFilter() []uint64 {
 	result := make([]uint64, maxConcurrentPids)
-	for nsid, pids := range p.pidsFilter.CurrentPIDs(ebpfcommon.PIDTypeKProbes) {
+	p.addPidsToFilter(result, ebpfcommon.PIDTypeKProbes)
+
+	// When payload extraction is enabled, also include Go PIDs so that
+	// kprobe events can supplement Go tracer data with body-based
+	// protocol detection (e.g. MCP, GenAI).
+	if p.cfg.EBPF.PayloadExtraction.Enabled() {
```
NimrodAvni78, line 94: adding the go pids to the pid filter will cause double spans being generated, since now traffic is tracked by the go tracer (uprobes) and the generic tracers (kprobes)
we have an [issue](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1846) that should unify the way we track go applications, but i belive for the scope of this PR we should focus on generic TCP path (like many protocols) and not touch any of the go tracers
if you want this to also work on go applications you can use `discovery.skip_go_specific_tracers: true`

pkg/export/attributes/names/attrs.go
```diff
@@ -11,6 +11,7 @@ import (
 
 	"go.opentelemetry.io/otel/attribute"
 	semconv "go.opentelemetry.io/otel/semconv/v1.38.0"
+	semconv_v140 "go.opentelemetry.io/otel/semconv/v1.40.0"
```
NimrodAvni78, line 14: not sure what i think about importing multiple semconv versions
i think we should either consider updating in the project or just use hardcoded strings until we do

NimrodAvni78 (changes requested):
thank you for the addition!
i think the scope of this PR is too wide, mainly with the go large buffers refactor and not reusing existing logic of JsonRPC parsing
I would also appreciate adding integration tests for this

rafaelroquetto:
I tend to agree with @NimrodAvni78, it would be great if we could split this into smaller PRs as per contributing guidelines.

NameHaibinZhang:
> I tend to agree with @NimrodAvni78, it would be great if we could split this into smaller PRs as per contributing guidelines.

ok

pkg/export/attributes/names/attrs.go
NameHaibinZhang: ok

pkg/internal/ebpf/generictracer/generictracer.go
NameHaibinZhang: ok

NameHaibinZhang:
> @NameHaibinZhang we recently added JSONRPC detection, did you consider using the same logic from there for parsing and detection?

ok

NameHaibinZhang:
@NimrodAvni78 @rafaelroquetto  I've minimized the MCP support code and reverted the changes to gotracer. It passed all tests, please review, thks.

pkg/ebpf/common/http/mcp.go
```diff
@@ -0,0 +1,207 @@
...
+	if !mcpMethods[rpcReq.Method] {
+		// Not a recognized MCP method. Check whether the session header
+		// was present — that still qualifies the request as MCP even if
+		// the method is unknown (e.g. a custom extension method).
+		if sessionID == "" {
+			return *baseSpan, false
+		}
+	}
+
+	slog.Debug("MCP", "method", rpcReq.Method, "request", string(reqB))
```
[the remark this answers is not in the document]
NameHaibinZhang: ok

pkg/ebpf/common/http/mcp.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	req.Body = io.NopCloser(bytes.NewBuffer(reqB))
+
+	reqB = bytes.TrimSpace(reqB)
+	if len(reqB) == 0 || reqB[0] != '{' {
+		return *baseSpan, false
+	}
+
+	var rpcReq mcpRequest
+	if err := json.Unmarshal(reqB, &rpcReq); err != nil {
```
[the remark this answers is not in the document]
NameHaibinZhang: Intentionally not supporting batch JSON-RPC requests here. MCP batch requests are extremely rare in practice — none of the major MCP SDKs use them today. The current code simply treats array payloads as non-MCP and skips them silently, which is safe. Adding batch support would significantly increase complexity (parsing, iterating, correlating responses) for a case we've never seen in the wild. Happy to revisit if batch usage becomes common.

NimrodAvni78:
@NameHaibinZhang can we add integration tests for this?

NameHaibinZhang:
> @NameHaibinZhang can we add integration tests for this?

sure

NameHaibinZhang:
@NimrodAvni78 I have already added integration tests, please review, thks.

internal/test/integration/components/pythonmcp/main.py
```diff
@@ -0,0 +1,163 @@
...
+    "initialize": handle_initialize,
+    "tools/list": handle_tools_list,
+    "tools/call": handle_tools_call,
+    "resources/read": handle_resources_read,
+    "prompts/get": handle_prompts_get,
+    "ping": handle_ping,
+}
+
+
+class Handler(BaseHTTPRequestHandler):
```
NimrodAvni78: nit: my preference for integration tests is to use the higher level sdk / libraries whenever possible, this allows us to more closely mimic real work usecases
in this case instead of implementing http server yourself which implements both JSONRPC and mcp on top of that, you could concider just using the [mcp python sdk](https://github.com/modelcontextprotocol/python-sdk)

pkg/ebpf/common/http/mcp.go
```diff
@@ -0,0 +1,211 @@
...
+
+	if len(rpcReq.ID) > 0 && string(rpcReq.ID) != "null" {
+		result.RequestID = rawIDString(rpcReq.ID)
+	}
+
+	parseMCPParams(rpcReq, result)
+
+	// Parse response for error and protocol version.
+	if resp != nil && resp.Body != nil {
+		respB, err := getResponseBody(resp)
```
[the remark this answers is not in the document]
NameHaibinZhang: ok

pkg/ebpf/common/http/mcp.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	if !mcpMethods[rpcReq.Method] {
+		// Not a recognized MCP method. Check whether the session header
+		// was present — that still qualifies the request as MCP even if
+		// the method is unknown (e.g. a custom extension method).
+		if sessionID == "" {
+			return *baseSpan, false
+		}
+	}
```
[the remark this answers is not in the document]
NameHaibinZhang: Good catch.

pkg/ebpf/common/http/mcp.go
```diff
@@ -0,0 +1,240 @@
...
+
+// mcpResponse is the JSON-RPC 2.0 response envelope used by MCP.
+type mcpResponse struct {
+	JSONRPC string          `json:"jsonrpc"`
+	ID      json.RawMessage `json:"id,omitempty"`
+	Result  json.RawMessage `json:"result,omitempty"`
+	Error   *mcpError       `json:"error,omitempty"`
+}
+
+type mcpError struct {
```
rafaelroquetto: `mcpError` is identical to `jsonRPCError` in the same package (`Code int`, `Message string`). `mcpResponse` could use `*jsonRPCError` directly and the `result.ErrorCode`/`result.ErrorMessage` extraction in `parseMCPResponse` would work unchanged. Introducing a parallel type for what is already defined here adds unnecessary duplication — see AGENTS.md: *"Do not introduce new implementations when equivalent functionality already exists."*

---
*Generated by Claude (claude-sonnet-4-6) via Claude Code*

pkg/ebpf/common/http/mcp.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	reqB, err := io.ReadAll(req.Body)
+	if err != nil {
+		return *baseSpan, false
+	}
+	req.Body = io.NopCloser(bytes.NewBuffer(reqB))
+
+	reqB = bytes.TrimSpace(reqB)
+	if len(reqB) == 0 || reqB[0] != '{' {
```
rafaelroquetto: MCP runs over JSON-RPC 2.0, which permits batch requests (`[...]`). `jsonrpc.go` in the same package already handles the array case in `parseJSONRPCRequest`. This check silently drops batch MCP requests without ever attempting to parse them. At minimum this gap should be documented; ideally `parseJSONRPCRequest` (or the batch-handling logic from it) should be reused here rather than duplicated later if/when batch support is added.

---
*Generated by Claude (claude-sonnet-4-6) via Claude Code*

pkg/ebpf/common/http/openai_test.go
```diff
@@ -89,7 +89,6 @@ func gzipBody(t *testing.T, body string) io.ReadCloser {
 	return io.NopCloser(&buf)
 }
 
-//nolint:unparam
 func makeRequest(t *testing.T, method, url, body string) *http.Request {
```
rafaelroquetto, line 93: Removing `//nolint:unparam` is an unrelated cleanup. Per AGENTS.md: *"Do not include unrelated edits, formatting changes, or cleanup."* Please revert this and keep the PR scoped to MCP support.

---
*Generated by Claude (claude-sonnet-4-6) via Claude Code*

pkg/ebpf/common/http/mcp.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	if !mcpMethods[rpcReq.Method] {
+		// Not a recognized MCP method. Check whether the session header
+		// was present — that still qualifies the request as MCP even if
+		// the method is unknown (e.g. a custom extension method).
+		if sessionID == "" {
+			return *baseSpan, false
+		}
+	} else if ambiguousMethods[rpcReq.Method] && sessionID == "" {
```
rafaelroquetto: `ambiguousMethods` currently contains `initialize` and `ping`, but the inner condition only names `initialize` explicitly — `ping` (and any future entry) relies implicitly on the outer branch: if session is absent, reject. This invariant isn't documented, so adding a new entry to `ambiguousMethods` requires understanding the assumption to avoid silently misclassifying it. Consider either making the per-method logic explicit, or documenting the invariant.

Separately: the existing code in this package uses `map[string]struct{}` for set membership (see `elasticsearch.go`). Both `mcpMethods` and `ambiguousMethods` should follow that pattern — per AGENTS.md: *"Follow existing code patterns and structure."*

---
*Generated by Claude (claude-sonnet-4-6) via Claude Code*

pkg/export/attributes/names/attrs.go
```diff
@@ -260,6 +260,16 @@ const (
 	RPCResponseStatusCode  = Name("rpc.response.status_code")
 )
 
+// MCP (Model Context Protocol) attributes
+const (
+	MCPMethodName      = Name("mcp.method.name")
+	MCPSessionID       = Name("mcp.session.id")
+	MCPProtocolVersion = Name("mcp.protocol.version")
+	MCPResourceURI     = Name("mcp.resource.uri")
+	GenAIToolName      = Name("gen_ai.tool.name")
```
rafaelroquetto: `GenAIToolName` and `GenAIPromptName` are generic GenAI attributes, not MCP-specific, but they are placed inside the `// MCP (Model Context Protocol) attributes` block. The file already has a `// GenAI` block at line 242 (`GenAIInput`, `GenAIOutput`, `GenAITools`, `GenAIOperationName`, etc.) — these two constants belong there. Per AGENTS.md: *"Follow existing code patterns and structure."*

---
*Generated by Claude (claude-sonnet-4-6) via Claude Code*

rafaelroquetto (commented):
hey @NameHaibinZhang thank you for all the changes. This is looking really good, and actually LGTM. There are a few auto-generated comments by Claude for your appreciation - some are nits, so up to you to decide what to do about them, but I do think it has a point with regards to the `ambiguousMethods` one, the `jsonRPCError` duplication and the batch requests ones.

Once we rebase this branch, I think we are good to go and I can approve it from my end (might wanna wait for @NimrodAvni78 to review it before merging it).

NameHaibinZhang:
> hey @NameHaibinZhang thank you for all the changes. This is looking really good, and actually LGTM. There are a few auto-generated comments by Claude for your appreciation - some are nits, so up to you to decide what to do about them, but I do think it has a point with regards to the `ambiguousMethods` one, the `jsonRPCError` duplication and the batch requests ones.
> 
> Once we rebase this branch, I think we are good to go and I can approve it from my end (might wanna wait for @NimrodAvni78 to review it before merging it).

"Got it, thanks. I'll review Claude's feedback and incorporate the relevant changes."

pkg/ebpf/common/http/mcp.go
NameHaibinZhang: fix

pkg/ebpf/common/http/openai_test.go
NameHaibinZhang: fix

pkg/ebpf/common/http/mcp.go
NameHaibinZhang: fix

pkg/export/attributes/names/attrs.go
NameHaibinZhang: fix

pkg/ebpf/common/http/mcp.go
NameHaibinZhang: fix

NameHaibinZhang:
@NimrodAvni78 @rafaelroquetto I've addressed all the comments and rebased onto the main branch. Please take another look, thanks!

internal/test/integration/components/pythonmcp/main.py
```diff
@@ -0,0 +1,163 @@
+"""
+MCP (Model Context Protocol) server for integration testing.
```
marctc, line 2: would it possible to test an end to end MCP system instead of this mock?

internal/test/integration/red_test_python_mcp.go
```diff
@@ -0,0 +1,202 @@
...
+		require.GreaterOrEqual(ct, len(traces), 1)
+
+		lastTrace := traces[len(traces)-1]
+		// The trace may contain child spans ("in queue", "processing");
+		// locate the MCP server span by its expected operation name.
+		res := lastTrace.FindByOperationName("execute_tool get-weather", "server")
+		require.GreaterOrEqual(ct, len(res), 1)
+		span := res[0]
+
+		tag, found := jaeger.FindIn(span.Tags, "mcp.method.name")
```
marctc: nit: can we these asserts in on single assert with `jaeger.Diff` like we do in other places?

pkg/appolly/app/request/span.go
```diff
@@ -609,6 +612,28 @@ func (b *VendorBedrock) GetStopReason() string {
 	return ""
 }
 
+// MCPCall holds parsed data from a Model Context Protocol request/response.
+type MCPCall struct {
+	Method       string `json:"method"`                 // mcp.method.name
```
marctc: nit: probably we can remove those comments on right side

pkg/ebpf/common/http/mcp.go
```diff
@@ -0,0 +1,240 @@
...
+		return *baseSpan, false
+	}
+	req.Body = io.NopCloser(bytes.NewBuffer(reqB))
+
+	reqB = bytes.TrimSpace(reqB)
+	// NOTE: JSON-RPC 2.0 also permits batch requests (arrays), but MCP
+	// batch support is not implemented yet. Only single-object requests
+	// are handled here; batch requests fall through to the generic
+	// JSON-RPC parser in jsonrpc.go.
+	if len(reqB) == 0 || reqB[0] != '{' {
```
NimrodAvni78: again, we already have jsonrpc parsing implemented and we are reimplementing it here, please try to reuse the same logic

pkg/ebpf/common/http/mcp.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// ambiguousMethods lists JSON-RPC method names shared with other protocols
+// (e.g. LSP). Each entry maps to a disambiguator function that returns true
+// when the request carries an MCP-specific signal beyond the method name.
+// The Mcp-Session-Id header is checked before consulting this map; entries
+// here only need to handle the no-session-header case.
+var ambiguousMethods = map[string]func(json.RawMessage) bool{
+	"initialize": hasMCPProtocolVersion,
+	"ping":       func(json.RawMessage) bool { return false },
```
NimrodAvni78: how is it ambiguous if it is always false?

internal/test/integration/components/pythonmcp/main.py
NameHaibinZhang: Good point. I'll switch the test server to use the [mcp Python SDK](https://github.com/modelcontextprotocol/python-sdk) so the integration test exercises a real MCP stack instead of our hand-rolled JSON-RPC implementation. Will update in the next revision.

pkg/ebpf/common/http/mcp.go
NameHaibinZhang: Done. I've extended jsonRPCRequest/jsonRPCResponse in jsonrpc.go with Params and Result fields, removed the duplicate mcpRequest/mcpResponse structs from mcp.go, and now MCPSpan calls parseJSONRPCRequest() directly to reuse the shared parsing logic.

pkg/ebpf/common/http/mcp.go
NameHaibinZhang: You're right — having ping in ambiguousMethods with an always-false function was confusing. I've separated it into a new sessionOnlyMethods set, which clearly expresses the intent: these method names overlap with other protocols and have no distinguishing params, so they are only recognized as MCP when the Mcp-Session-Id header is present.

pkg/appolly/app/request/span.go
NameHaibinZhang: Done, removed the inline comments from MCPCall struct fields.

internal/test/integration/components/pythonmcp/main.py
NameHaibinZhang: Yes, agreed — @NimrodAvni78 NimrodAvni78 also suggested the same. I'll rewrite the test server using the [mcp Python SDK](https://github.com/modelcontextprotocol/python-sdk) to exercise a real end-to-end MCP system. Will update in the next revision.

internal/test/integration/red_test_python_mcp.go
NameHaibinZhang: Done. Replaced the individual FindIn/assert.Equal calls with span.Diff(...) and span.DiffAsRegexp(...) (for the substring match on error.message) across all three test blocks, matching the pattern used elsewhere.

pkg/ebpf/common/http/mcp.go
```diff
@@ -0,0 +1,220 @@
...
+		return *baseSpan, false
+	}
+	req.Body = io.NopCloser(bytes.NewBuffer(reqB))
+
+	rpcReq, err := parseJSONRPCRequest(reqB, false)
+	if err != nil {
+		return *baseSpan, false
+	}
+
+	if _, known := mcpMethods[rpcReq.Method]; !known {
```
MrAlias, line 89: When both MCP and JSON-RPC extraction are enabled, this block still accepts shared method names like `tools/list` and `tools/call` as MCP based on the method name alone. That means plain JSON-RPC traffic using those method names can get classified as MCP before `JSONRPCSpan` has a chance to run, which drops the existing `rpc.system=jsonrpc` semantics.

I think the detection rule needs to be stricter here: for methods that are legal in both protocols, we should require MCP-specific evidence such as `Mcp-Session-Id` or method-specific MCP params, and otherwise fall back to JSON-RPC.

A sketch of the shape I mean:

```suggestion
if _, known := mcpMethods[rpcReq.Method]; !known {
	if sessionID == "" {
		return *baseSpan, false
	}
} else if sessionID == "" {
	disambiguate, ok := ambiguousMethods[rpcReq.Method]
	if ok && !disambiguate(rpcReq.Params) {
		return *baseSpan, false
	}
	if !ok {
		// Shared JSON-RPC method names should stay JSON-RPC unless we have
		// some MCP-specific signal beyond the method string itself.
		return *baseSpan, false
	}
}
```

If you go this route, I’d expect `ambiguousMethods` to cover the shared method names that can be disambiguated from params, while methods with no MCP-specific signal should require the session header.

pkg/ebpf/common/http/mcp.go
```diff
@@ -0,0 +1,220 @@
...
+	case "initialize":
+		var p mcpInitializeParams
+		if json.Unmarshal(rpcReq.Params, &p) == nil {
+			result.ProtocolVer = p.ProtocolVersion
+		}
+	}
+}
+
+// parseMCPResponse extracts error information and protocol version from the response.
+func parseMCPResponse(data []byte, result *request.MCPCall) {
```
MrAlias, line 194: This only handles object responses, so valid JSON-RPC batch replies (`[...]`) never populate MCP error fields or the negotiated protocol version for the matching request. In practice that means batched `initialize` / `tools/call` traffic can be exported as success even when the selected sub-response is an error.

Since `jsonrpc.go` already has batch matching logic, I’d strongly prefer reusing the same pattern here instead of treating MCP responses as object-only.

Something along these lines would preserve the selected sub-response:

```suggestion
func parseMCPResponse(data []byte, result *request.MCPCall) {
	data = bytes.TrimSpace(data)
	if len(data) == 0 {
		return
	}

	switch data[0] {
	case '{':
		var resp jsonRPCResponse
		if err := json.Unmarshal(data, &resp); err != nil {
			return
		}
		applyMCPResponse(resp, result)
	case '[':
		var batch []jsonRPCResponse
		if err := json.Unmarshal(data, &batch); err != nil {
			return
		}
		if resp, ok := matchResponse(batch, result.RequestID); ok {
			applyMCPResponse(resp, result)
		}
	}
}
```

with the MCP-specific field extraction moved into a small helper like `applyMCPResponse(...)` so the single-response and batch-response paths stay consistent.

MrAlias (commented):
Thanks for the revision here. I left two inline comments on remaining correctness issues in MCP detection and response parsing; once those are addressed, this looks much closer.

pkg/ebpf/common/http/mcp.go
NameHaibinZhang: Good catch. You're right that known MCP methods like tools/call and tools/list are currently classified as MCP based solely on the method name, without requiring any MCP-specific evidence when Mcp-Session-Id is absent. This could misclassify plain JSON-RPC traffic.Restructured the detection logic as suggested: when sessionID is empty, only methods in ambiguousMethods that pass their disambiguator are accepted. All other methods require the session header. This also makes sessionOnlyMethods redundant, so it's been removed.

pkg/ebpf/common/http/mcp.go
NameHaibinZhang: Makes sense. The current parseMCPResponse silently drops batch replies, so error fields and protocol version are never populated for batched traffic. Refactored to follow the same switch data[0] pattern used in parseJSONRPCResponse, reusing matchResponse for batch ID matching and extracting MCP-specific field assignment into an applyMCPResponse helper.

NameHaibinZhang:
@marctc @MrAlias @NimrodAvni78 @rafaelroquetto  Hi, I've finished all the requested changes and updated the tests for the MCP PR

marctc (approved):

NameHaibinZhang (commented):
ok

pkg/ebpf/common/http/mcp.go
```diff
@@ -0,0 +1,228 @@
...
+		}
+	}
+}
+
+// parseMCPResponse extracts error information and protocol version from the response.
+// It handles both single object and batch JSON-RPC responses, reusing
+// matchResponse for batch ID matching.
+func parseMCPResponse(data []byte, result *request.MCPCall) {
+	data = bytes.TrimSpace(data)
+	if len(data) == 0 {
```
mmat11, line 196: is this expected? should we return an error instead?

pkg/ebpf/common/http/mcp.go
```diff
@@ -0,0 +1,228 @@
...
+		}
+		applyMCPResponse(resp, result)
+	case '[':
+		var batch []jsonRPCResponse
+		if err := json.Unmarshal(data, &batch); err != nil {
+			return
+		}
+		if resp, ok := matchResponse(batch, result.RequestID); ok {
+			applyMCPResponse(resp, result)
+		}
```
mmat11, line 214: should we have a default branch returning error here? should we also return error when json.Unmarshal fails?

mmat11 (approved):

mariomac (approved):
LGTM!

pkg/ebpf/common/http/mcp.go
NameHaibinZhang: The design intent here is best-effort enrichment — MCP request detection has already succeeded upstream, and response parsing only extracts additional fields (error code, protocol version). If parsing fails, the span is still created with the information already gathered. This is consistent with how parseMCPParams handles unmarshal failures silently.

pkg/ebpf/common/http/mcp.go
NameHaibinZhang: Same as above

rafaelroquetto (approved):

pkg/ebpf/common/http_transform.go
```diff
@@ -204,6 +204,13 @@ func httpRequestResponseToSpan(parseCtx *EBPFParseContext, event *BPFHTTPInfo, r
 		}
 	}
 
+	if parseCtx != nil && parseCtx.payloadExtraction.HTTP.GenAI.MCP.Enabled {
```
NimrodAvni78: Since we try to parse mcp before JsonRPC, a JsonRPC request which is not mcp will be parsed twice, once mcp detection, which will fail in the stages of matching the method name, and after that in jsonrpc detection.
do you think it is possible to reverse the order?
first parsing JsonRPC and adding the JsonRPC field to the span struct, and after relying on this information (not parsing json again) trying to check if this data is an MCP call, if so creating an MCP span, and if not just creating a normal JsonRPC span
this has the downside of making mcp detection reliant on JsonRPC detection and it must run before it but i think it makes since in this context. especially if we mention it in a comment and do a validation check the JsonRPC field is set

NameHaibinZhang: > Since we try to parse mcp before JsonRPC, a JsonRPC request which is not mcp will be parsed twice, once mcp detection, which will fail in the stages of matching the method name, and after that in jsonrpc detection. do you think it is possible to reverse the order? first parsing JsonRPC and adding the JsonRPC field to the span struct, and after relying on this information (not parsing json again) trying to check if this data is an MCP call, if so creating an MCP span, and if not just creating a normal JsonRPC span this has the downside of making mcp detection reliant on JsonRPC detection and it must run before it but i think it makes since in this context. especially if we mention it in a comment and do a validation check the JsonRPC field is set

Agreed, this makes architectural sense — MCP is fundamentally a layer on top of JSON-RPC, so parsing JSON-RPC first and then checking for MCP signals is the more natural flow.

NameHaibinZhang: Good suggestion — refactored to parse JSON-RPC once via TryParseJSONRPC and reuse the result for both MCP and JSON-RPC detection, avoiding the double-parse. MCPSpanFromParsed and JSONRPCSpanFromParsed now accept pre-parsed data, while the original convenience wrappers are preserved for direct callers.

NimrodAvni78 (approved):

PR #1873, opened 2026-04-17
Fix race condition in netolly

pinoOgni:
## Summary

I was doing some tests with obi, and by running it and quickly stopping it, I noticed a nasty "panic: runtime error: invalid memory address or nil pointer dereference.".

I tried to reproduce it, but couldn't do it without putting a sleep command in the `TraceLoop` of the `MapTracer` (same thing for `RingBufTracer`). I ran a test, and here's the error.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x18dd024]

goroutine 185 [running]:
go.opentelemetry.io/obi/pkg/internal/netolly/ebpf.(*SockFlowFetcher).FlowPacketStatsMap(0x2540be400?)
	/tmp/lima/coralogix-github/opentelemetry-ebpf-instrumentation/pkg/internal/netolly/ebpf/sock_tracer.go:136 +0x4
go.opentelemetry.io/obi/pkg/netolly/agent.init.func1.(*MapTracer).TraceLoop.1({0x29a27d0, 0x4000176320})
	/tmp/lima/coralogix-github/opentelemetry-ebpf-instrumentation/pkg/internal/netolly/flow/tracer_map.go:81 +0x7c
go.opentelemetry.io/obi/pkg/pipe/swarm.(*Runner).Start.func2()
	/tmp/lima/coralogix-github/opentelemetry-ebpf-instrumentation/pkg/pipe/swarm/runner.go:82 +0x34
created by go.opentelemetry.io/obi/pkg/pipe/swarm.(*Runner).Start in goroutine 72
	/tmp/lima/coralogix-github/opentelemetry-ebpf-instrumentation/pkg/pipe/swarm/runner.go:81 +0x194
```

So, while it's still using sleep, but with the fix:
```
time=2026-04-17T17:23:50.618+02:00 level=DEBUG msg="Application O11y pipeline finished"
time=2026-04-17T17:24:00.569+02:00 level=DEBUG msg="can't run network metrics capture" error="failed to stop Flows agent: graceful shutdown has timed out while waiting for eBPF network infrastructure to finish"
time=2026-04-17T17:24:00.569+02:00 level=DEBUG msg="exiting trace loop due to context cancellation" component=flow.MapTracer
time=2026-04-17T17:24:00.570+02:00 level=ERROR msg="OpenTelemetry eBPF Instrumentation ran with errors" error="setupNetO11y: can't run network metrics capture: failed to stop Flows agent: graceful shutdown has timed out while waiting for eBPF network infrastructure to finish"
```

This PR adds a guard to `FlowFetcher/SockFlowFetcher `objects with a mutex so packet-stats reads and `Close()` can't race. It adds `LookupPacketStats()` that does the lookup under the lock and returns`ErrTracerTerminated` post-close and solve also the ownership issue of the eBPF map (previously `FlowPacketStatsMap()` handed the raw `*ebpf.Map` pointer to`PacketStats`, which kept using it after the fetcher had closed the underlying objects).

## Validation


pkg/internal/netolly/ebpf/tracer.go
```diff
@@ -185,6 +185,9 @@ func (m *FlowFetcher) Close() error {
 }
 
 func (m *FlowFetcher) FlowPacketStatsMap() *ebpf.Map {
+	if m.objects == nil {
+		return nil
+	}
 	return m.objects.FlowPacketStats
```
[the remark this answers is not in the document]
pinoOgni: Not sure about the mutex idea. I can add the local variable one. WDYT?
Update after the weekend: done.

pkg/internal/netolly/ebpf/tracer.go
```diff
@@ -175,16 +176,25 @@ func (m *FlowFetcher) Close() error {
 		}
 	}
 
-	if m.objects != nil {
```
mmat11, line 178: couldn't we have wrapped the old block in lock()/unlock()?

pkg/internal/netolly/ebpf/sock_tracer.go
```diff
@@ -133,6 +134,11 @@ func printVerifierErrorInfo(err error) {
 }
 
 func (m *SockFlowFetcher) FlowPacketStatsMap() *ebpf.Map {
+	m.objectsMu.Lock()
+	defer m.objectsMu.Unlock()
+	if m.objects == nil {
```
mmat11, line 143: did we check if the callers of this function expect a `nil` as a return value?

pkg/internal/netolly/ebpf/tracer.go
```diff
@@ -175,16 +176,25 @@ func (m *FlowFetcher) Close() error {
...
 		}
-		m.objects = nil
 	}
 	return errors.Join(errs...)
 }
 
 func (m *FlowFetcher) FlowPacketStatsMap() *ebpf.Map {
+	m.objectsMu.Lock()
+	defer m.objectsMu.Unlock()
+	if m.objects == nil {
```
mmat11, line 199: same here

pkg/internal/netolly/ebpf/tracer.go
pinoOgni: I did it then I thought to this solution because if we simply wrap it, it would hold the mutex for the entire duration of `obj.Close()`, while this pattern minimizes the critical section.

pkg/internal/netolly/ebpf/sock_tracer.go
pinoOgni: Yep, it is called here `packetStats, err := NewPacketStats(m.mapFetcher.FlowPacketStatsMap())` and the nil-return code was already there before the fix.

mmat11 (approved):
lgtm!

rafaelroquetto (changes requested):
I am not sure this is the way to go, it feels like we are just masking the issue.

First, there's no clear ownership of `FlowPacketStatsMap` - we are returning an internal pointer to a resource owned by MapFetcher.

The first workaround could be simply checking if that pointer is `nil` in `NewPacketStats` but that is also a band-aid IMHO.

The second issue caused by leaking the internal FlowPacketStatsMap is that if that map is closed, `packetStats` will try to operate on a closed map - which will cause the error returned by `fm.bpfPacketStats.Lookup(...)` to propagate up to the caller (instead of handling a `Close()` scenario gracefully). Not only that, it tries to `Close()` the map it does not own.

Finally, by storing the pointer, `packetStats` defeats the purpose of having the `[Sock]FlowFetcher::Close()` doing `		m.objects = nil` - since it holds the go object alive.


Perhaps better semantics would be abstract the `ebpf.Map` behind an interface (or even better, just extend the `mapFetcher` interface) to have a `PacketCount() ([]ebpf2.NetPacketCount, err)`.

This means that `PacketCount()` can check for `nil`, and any other errors and propagate them accordingly: for instance, it could propagate `ErrTracerTerminated` or something like that, which lets the caller deal with it differently.

I hope I am making sense, sorry for the wall of text.

pinoOgni:
> I am not sure this is the way to go, it feels like we are just masking the issue.
> 
> First, there's no clear ownership of `FlowPacketStatsMap` - we are returning an internal pointer to a resource owned by MapFetcher.
> 
> The first workaround could be simply checking if that pointer is `nil` in `NewPacketStats` but that is also a band-aid IMHO.
> 
> The second issue caused by leaking the internal FlowPacketStatsMap is that if that map is closed, `packetStats` will try to operate on a closed map - which will cause the error returned by `fm.bpfPacketStats.Lookup(...)` to propagate up to the caller (instead of handling a `Close()` scenario gracefully). Not only that, it tries to `Close()` the map it does not own.
> 
> Finally, by storing the pointer, `packetStats` defeats the purpose of having the `[Sock]FlowFetcher::Close()` doing ` m.objects = nil` - since it holds the go object alive.
> 
> Perhaps better semantics would be abstract the `ebpf.Map` behind an interface (or even better, just extend the `mapFetcher` interface) to have a `PacketCount() ([]ebpf2.NetPacketCount, err)`.
> 
> This means that `PacketCount()` can check for `nil`, and any other errors and propagate them accordingly: for instance, it could propagate `ErrTracerTerminated` or something like that, which lets the caller deal with it differently.
> 
> I hope I am making sense, sorry for the wall of text.

Hi @rafaelroquetto, thanks for the review and the detailed analysis! 

I initially implemented a quick fix to address the bug immediately, but I guess that your approach is the right one and it addresses ownership rather than masking it. I'll refactor the code following your suggestion and update the PR ASAP.

rafaelroquetto (approved):
Fantastic, looking great! Thank you for the changes!

PR #1875, opened 2026-04-18
Fix shared pidns disambiguation

### LINKED ISSUE #1871: Spans misattributed across pods sharing a PID namespace
## Summary

When multiple pods share a Linux PID namespace (most commonly via `hostPID: true`), `PodContainerByPIDNs` returns an arbitrary pod for the shared namespace inode, causing spans to be decorated with the wrong pod's k8s metadata (`k8s.pod.name`, `k8s.namespace.name`, `k8s.deployment.name`, `service.name`, `service.namespace`, `service.instance.id`, etc.).

This is especially severe on OpenShift clusters where many system DaemonSets run with `hostPID: true` by default (OVN-Kubernetes, machine-config-daemon, node-exporter, multus, tuned, etc.), creating a large collision pool under the host `init_pid_ns` inode (`4026531836`).

## Root Cause

`PodContainerByPIDNs` ([`pkg/kube/store.go`](pkg/kube/store.go)) looks up pod metadata using only the PID namespace inode:

```go
func (s *Store) PodContainerByPIDNs(pidns uint32) (*kube.CachedObjMeta, string) {
    if infos, ok := s.namespaces[pidns]; ok {
        for _, info := range infos {
            if om, ok := s.podsByContainer[info.ContainerID]; ok {
                // ...
                return om, containerName
            }
            break  // "the namespace is the same for all pids in the container"
        }
    }
    return nil, ""
}
```

The comment _"the namespace is the same for all pids in the container, we need to check one only"_ is correct for isolated PID namespaces, but **incorrect when `hostPID: true`** -- all such pods share the host's `init_pid_ns` inode. The function iterates a Go map (non-deterministic order), picks the first entry, and `break`s. This means every span whose PID lives in the shared namespace is attributed to whichever pod Go's map iteration happens to return first.

## Symptoms

- Spans from DaemonSet X show `service.name`, `service.namespace` of Deployment Y
- Particularly acute on OpenShift due to the large number of system DaemonSets with `hostPID: true`

## Reproduction

1. Deploy OBI as a DaemonSet (with `hostPID: true`, as required)
2. Deploy an application pod (Deployment, no `hostPID`)
3. Deploy a second DaemonSet with `hostPID: true` that serves HTTP (e.g., a node-proxy, node-exporter with HTTP endpoints, or any hostPID workload)
4. Configure OBI to instrument both workloads
5. Observe that spans from one workload carry the other's k8s metadata

## Environment

- Kubernetes/OpenShift with multiple `hostPID: true` DaemonSets
- OBI running as a DaemonSet with `hostPID: true`
- Any application workload instrumented by OBI

NimrodAvni78:
## Summary

- Fix `PodContainerByPIDNs` to use the host PID for disambiguation when multiple containers share a PID namespace (e.g. `hostPID: true` pods all map to the host `init_pid_ns` inode)
- Previously, the function iterated a Go map and returned the first entry, causing non-deterministic misattribution of k8s metadata across pods
- Add unit tests for the shared PID namespace scenario

Resolves open-telemetry/opentelemetry-ebpf-instrumentation#1871

## Testing
- Added unit tests for PodContainerByPIDNs
- Added integration tests for situations where an application and a daemonset are running together to make sure there is no misattribution
- Tested on openshift cluster and missattributions were gone

## Validation


pkg/kube/store.go
```diff
@@ -378,26 +378,62 @@ func (s *Store) PodByContainerID(cid string) *kube.CachedObjMeta {
...
+	// Fallback: only safe when every registered PID in this namespace points
+	// to the same container (i.e. the namespace is not actually shared).
+	var pick *container.Info
+	for _, info := range infos {
+		if pick == nil {
+			pick = info
+			continue
 		}
+		if info.ContainerID != pick.ContainerID {
+			s.log.Debug("cannot disambiguate shared PID namespace without matching host PID; skipping k8s decoration",
```
mmat11, line 415: should we increase log verbosity here? not sure between info/warn

mmat11 (approved):

rafaelroquetto (commented):
Non-rhetorical question: why can't we just rely on the host pid?

NimrodAvni78:
> Non-rhetorical question: why can't we just rely on the host pid?

I think it might work, i can try it out and test myself
maybe @grcevski @mariomac have any input here on why it was done using pidNs instead of hostPID

Edit: not sure we can only rely on hostPID, we do register child processes to the map, but it is done via polling, if an event from a child process of a normal pod (no hostPID:true) comes without the childPID being registered yet we wont find it, right now we will find the correct pidNs and attribute it to the correct pod

i think the only situation where there will be a problem with the newly implemented solution is for hostPID:true pods with child processes, and they have yet to be registered via the polling cycle. in that case we will hit the case where we see multiple processes from multiple containers under the same pidNS, we will see multiple infos from multiple containers without an exact hostPID match, so we decide to not enrich the span instead of picking a random container (what the previous solution has done)

tbh not sure how to fix that, unless we move to monitoring fork/exec via eBPF to not allow for periods where pids are not registred

MrAlias:
SIG meeting: this may be specific to deployment difference between sidecar and daemonset

NimrodAvni78:
@rafaelroquetto @MrAlias I just viewed the sig recording so ill try to clarify:
This issue came up when running OBI as a **daemonset**, when running with multiple other daemonsets (hostPID: true)
that do network calls, what happened is that those network calls got misclassified as a random container on that podNS (a random hostPID:true pod)

regarding why we can't rely on hostPID alone i explained [here](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1875#issuecomment-4287191604)
tldr we track new processes on an interval, and if a child process of a normal (no hostPID:true) emits a span and the processes in not yet cached we wont find it
with this solution we will see there is only 1 container in this podNS and we will find it

rafaelroquetto (approved):
@NimrodAvni78 thank you so much for all the clarifications, LGTM!

MrAlias:
I think the product behavior here is correct: when identity is ambiguous, we should prefer missing Kubernetes metadata over attaching the wrong pod/container metadata.

Misattribution is worse than under-attribution for OBI because it produces convincing but false telemetry. In this case, the right behavior is:
- decorate when the host PID matches exactly
- decorate when the PID namespace is unambiguous
- skip decoration when multiple containers share the namespace and we can’t match the process confidently

The one thing I’d still like is to make that tradeoff explicit in the PR or code comments, since this is a deliberate product decision, not just an implementation detail. If possible, adding observability for "skipped decoration due to ambiguous shared PID namespace" would also help operators understand coverage gaps.

grcevski (approved):
LGTM! This looks great, I don't see any edge cases, unless we somehow are not running in host pid mode. However, in that case we are in the process pid namespace and it should still work.

grcevski:
> skip decoration when multiple containers share the namespace and we can’t match the process confidently

I think this will be the case typically when there are multiple containers in a pod, but in this case the metadata is the same, so we should offer it back.

NimrodAvni78:
@MrAlias we have [this comment](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1875/changes#diff-a63c9a86285d3b585ad35713076397a23ecd2eedf57f2c60fa2aa62acd96ebd4R385) which gives the details in the function level, do you think it needs to be documented elsewhere?

MrAlias:
> @MrAlias we have [this comment](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1875/changes#diff-a63c9a86285d3b585ad35713076397a23ecd2eedf57f2c60fa2aa62acd96ebd4R385) which gives the details in the function level, do you think it needs to be documented elsewhere?

Ah, looks good. 👍

MrAlias (approved):

PR #1878, opened 2026-04-19
JavaAgent: Skip `$$Lambda` classes during agent loading

### LINKED ISSUE #1877: OBI Java agent causes NoClassDefFoundError on Java 8 when dynamically attached to applications with Hibernate
## Description

When the OBI Java agent is dynamically attached (via `agentmain`) to a Java 8 application that uses Hibernate, the application crashes with `NoClassDefFoundError` for lambda classes such as:

```
java.lang.NoClassDefFoundError: org/hibernate/boot/registry/selector/internal/StrategySelectorImpl$$Lambda$249
```

The issue does not occur on Java 15+ because lambda classes are implemented as hidden classes in those versions, which the JVM prevents from being retransformed.

## Steps to reproduce

**1. Build the agent JAR:**

```bash
# From the repository root — requires Docker
make java-docker-build

# The agent JAR is now at:
# pkg/internal/java/embedded/obi-java-agent.jar
```

**2. Create a minimal Java 8 Hibernate app that self-attaches the agent:**

`pom.xml`:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
                             http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>
    <groupId>io.opentelemetry.obi.test</groupId>
    <artifactId>java8-hibernate-lambda-test</artifactId>
    <version>1.0-SNAPSHOT</version>
    <packaging>jar</packaging>
    <properties>
        <maven.compiler.source>1.8</maven.compiler.source>
        <maven.compiler.target>1.8</maven.compiler.target>
    </properties>
    <dependencies>
        <dependency>
            <groupId>org.hibernate</groupId>
            <artifactId>hibernate-core</artifactId>
            <version>5.6.15.Final</version>
        </dependency>
        <dependency>
            <groupId>com.h2database</groupId>
            <artifactId>h2</artifactId>
            <version>2.2.224</version>
        </dependency>
        <dependency>
            <groupId>javax.xml.bind</groupId>
            <artifactId>jaxb-api</artifactId>
            <version>2.3.1</version>
        </dependency>
        <dependency>
            <groupId>com.sun</groupId>
            <artifactId>tools</artifactId>
            <version>1.8</version>
            <scope>system</scope>
            <systemPath>${java.home}/../lib/tools.jar</systemPath>
        </dependency>
    </dependencies>
    <build>
        <plugins>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-jar-plugin</artifactId>
                <version>3.3.0</version>
                <configuration>
                    <archive><manifest>
                        <mainClass>HibernateApp</mainClass>
                    </manifest></archive>
                </configuration>
            </plugin>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-dependency-plugin</artifactId>
                <version>3.6.1</version>
                <executions><execution>
                    <id>copy-dependencies</id>
                    <phase>package</phase>
                    <goals><goal>copy-dependencies</goal></goals>
                    <configuration>
                        <outputDirectory>${project.build.directory}/lib</outputDirectory>
                    </configuration>
                </execution></executions>
            </plugin>
        </plugins>
    </build>
</project>
```

`src/main/java/HibernateApp.java`:

```java
import com.sun.tools.attach.VirtualMachine;
import java.util.Properties;
import java.util.concurrent.*;
import org.hibernate.SessionFactory;
import org.hibernate.cfg.Configuration;

public class HibernateApp {

    private static volatile SessionFactory sessionFactory;

    public static void main(String[] args) throws Exception {
        System.out.println("Java version: " + System.getProperty("java.version"));

        String pid = java.lang.management.ManagementFactory
                .getRuntimeMXBean().getName().split("@")[0];

        // Phase 1: Bootstrap Hibernate — creates many lambda classes
        Configuration config = new Configuration();
        Properties props = new Properties();
        props.setProperty("hibernate.connection.driver_class", "org.h2.Driver");
        props.setProperty("hibernate.connection.url", "jdbc:h2:mem:testdb;DB_CLOSE_DELAY=-1");
        props.setProperty("hibernate.connection.username", "sa");
        props.setProperty("hibernate.connection.password", "");
        props.setProperty("hibernate.dialect", "org.hibernate.dialect.H2Dialect");
        props.setProperty("hibernate.hbm2ddl.auto", "create-drop");
        config.setProperties(props);
        sessionFactory = config.buildSessionFactory();
        System.out.println("SessionFactory created");

        // Phase 2: Exercise lambdas before attachment (baseline)
        System.out.println("[Pre-attach] Exercising lambdas...");
        exerciseLambdas();
        System.out.println("[Pre-attach] OK");

        // Phase 3: Self-attach the OBI Java agent
        String agentPath = System.getProperty("obi.agent.path");
        if (agentPath == null) {
            System.out.println("No obi.agent.path set, sleeping...");
            Thread.sleep(Long.MAX_VALUE);
            return;
        }
        System.out.println("Attaching agent: " + agentPath);
        VirtualMachine vm = VirtualMachine.attach(pid);
        vm.loadAgent(agentPath, "debug=true");
        vm.detach();
        System.out.println("Agent attached");

        // Phase 4: Exercise lambdas after attachment
        // This is where NoClassDefFoundError happens if lambdas were corrupted
        System.out.println("[Post-attach] Exercising lambdas...");
        try {
            exerciseLambdas();
            System.out.println("[Post-attach] OK");
            System.out.println("=== TEST PASSED ===");
        } catch (Throwable t) {
            System.err.println("=== TEST FAILED: " + t.getClass().getName()
                    + ": " + t.getMessage() + " ===");
            t.printStackTrace();
            System.exit(1);
        }
    }

    private static void exerciseLambdas() throws Exception {
        ExecutorService executor = Executors.newFixedThreadPool(2);
        try {
            for (int i = 0; i < 3; i++) {
                final int idx = i;
                executor.submit((Runnable) () -> sessionFactory.getStatistics()).get();
                Future<String> f = executor.submit(() -> "result-" + idx);
                f.get();
            }
        } finally {
            executor.shutdown();
            executor.awaitTermination(5, TimeUnit.SECONDS);
        }
    }
}
```

**3. Build and run on Java 8:**

```bash
mvn package -q -DskipTests
java -cp "target/java8-hibernate-lambda-test-1.0-SNAPSHOT.jar:target/lib/*:$JAVA_HOME/lib/tools.jar" \
     -Dobi.agent.path=/path/to/obi-java-agent.jar \
     HibernateApp
```

**4. Observe the error.** The app bootstraps Hibernate, exercises lambdas successfully, then self-attaches the agent. The agent logs show it retransforming `$$Lambda` classes:

```
Retransforming class org.hibernate.boot.registry.selector.internal.StrategySelectorImpl$$Lambda$4/508198356
Retransforming class org.hibernate.boot.registry.selector.internal.StrategySelectorImpl$$Lambda$3/1256728724
Retransforming class org.hibernate.engine.jdbc.connections.internal.DriverManagerConnectionProviderImpl$PoolState$$Lambda$31/51554940
```

After attachment, exercising lambdas crashes with `NoClassDefFoundError`:

```
[Post-attach] Exercising lambdas...
=== TEST FAILED: java.lang.NoClassDefFoundError: HibernateApp$$Lambda$62 ===
java.lang.NoClassDefFoundError: HibernateApp$$Lambda$62
	at HibernateApp$$Lambda$62/1658980982.get$Lambda(Unknown Source)
	at HibernateApp.exerciseLambdas(HibernateApp.java:74)
	at HibernateApp.main(HibernateApp.java:57)
Caused by: java.lang.ClassNotFoundException: HibernateApp$$Lambda$62
	at java.net.URLClassLoader.findClass(URLClassLoader.java:387)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
	... 3 more
```

> **Note:** The `-javaagent` flag (premain) does **not** trigger the issue because the agent loads before Hibernate's lambda classes exist. The bug is specific to dynamic attachment (agentmain), which is how OBI injects the agent in production.

## Environment

- Java 8 (tested with Eclipse Temurin 8u482)
- Hibernate 5.6.15.Final
- OBI Java agent (current main branch)
- Dynamic attachment via `agentmain` (not `premain`)

NimrodAvni78:
## Summary

Fixes open-telemetry/opentelemetry-ebpf-instrumentation#1877

Skip `$$Lambda` classes during dynamic agent attachment to prevent `NoClassDefFoundError` on Java 8.

## Root cause

This is triggered by [JDK-8145964](https://bugs.openjdk.org/browse/JDK-8145964), a HotSpot JVM bug in Java 8 where retransforming VM anonymous lambda classes corrupts their constant pool linkage to the host class. The bug was fixed in Java 9 and **never backported to Java 8**.

On Java 8, `Instrumentation.isModifiableClass()` incorrectly returns `true` for lambda classes (VM anonymous classes created via `Unsafe.defineAnonymousClass`). When `retransformClasses()` is called on them — even with a no-op transformer — the JVM's redefinition machinery corrupts the lambda's shared constant pool, causing `NoClassDefFoundError` on next use.

From [byte-buddy#219](https://github.com/raphw/byte-buddy/issues/219):
> *"The instrumentation, even if it is a no-op, resets the lambda class's patched constant pool causing this error. There is, unfortunately, nothing to be done here, this is a VM bug."*

On Java 9+, the fix added an `is_anonymous()` check to `is_modifiable_class()` in HotSpot, so the JVM refuses retransformation of lambda classes entirely. On Java 15+ (JEP 371), lambdas became hidden classes with the same protection.

| JDK | Lambda type | `isModifiableClass` | Behavior |
|-----|------------|-------------------|----------|
| 8 | VM anonymous | `true` (bug) | retransform corrupts class |
| 9-14 | VM anonymous | `false` (JDK-8145964 fix) | JVM refuses retransform |
| 15+ | hidden class | `false` (JEP 371) | JVM refuses retransform |

## Fix

In `Agent.java`, the dynamic-attachment retransformation loop now skips lambda classes:

```java
static void retransformLoadedClasses(Instrumentation inst) {
    for (Class<?> clazz : inst.getAllLoadedClasses()) {
        if (clazz.getName().contains("$$Lambda")) {
            continue;
        }
        // ... matchers + retransformClasses() ...
    }
}
```

The loop was extracted from `agentmain()` into a package-private `retransformLoadedClasses()` method so it can be exercised by a unit test (the test can't call `agentmain()` directly because `premain()` requires the JNI native library and bootstrap classloader injection, neither of which work in a test JVM).

Lambda classes never contain the SSL/socket/executor methods that the agent instruments, so skipping them has no functional impact.

### Helper

A small `testutil.LambdaFactory` is added under the test sources to create lambda instances outside the `io.opentelemetry.obi` package — `Agent.builder()` already ignores that prefix, so test lambdas in the same package would never be retransformed.

## Testing
[CI fails](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/24629445392/job/72014240311) prior to fix
## Validation


rafaelroquetto (approved):
LGTM but I must confess I am not too familar with the Java code - might be worth it waiting for @grcevski to confirm this is the way.

grcevski (approved):
LGTM! Great stuff. I didn't know about this bug.

PR #1880, opened 2026-04-20
Fix signed length handling in `netFdReadRet` uprobe

MrAlias:
Prevent negative `netFD.Read` return values from being cast to an unsigned type and sign-wrapped into huge lengths that could be forwarded into protocol parsing and cause out-of-bounds `bpf_probe_read` memory disclosure.

This treats the return value in `obi_uprobe_netFdReadRet` as signed by changing `len` from `u64` to `s64` and update the debug print format to use `%lld`, while preserving the existing `len > 0` guard before calling `handle_light_weight_thread_buf`.

The overflow is not something that is expected to happen, given the Go code this instruments, but this ensures safety and prevents bad writes from creeping in.

### DIFF
--- a/bpf/gotracer/go_net.c
+++ b/bpf/gotracer/go_net.c
@@ -193,9 +193,11 @@ int obi_uprobe_netFdReadRet(struct pt_regs *ctx) {
 
     void *buf = (void *)net_ptr->byte_ptr;
 
-    u64 len = (u64)GO_PARAM1(ctx);
-    bpf_dbg_printk("buf=%llx, len=%d === ", buf, len);
+    s64 len = (s64)GO_PARAM1(ctx);
+    bpf_dbg_printk("buf=%llx, len=%lld === ", buf, len);
     if (buf && len > 0) {
+        const int bytes_len = (len > __INT_MAX__) ? __INT_MAX__ : (int)len;
+
         dbg_print_http_connection_info(&net_ptr->p_conn.conn);
 
         u16 orig_dport = net_ptr->p_conn.conn.d_port;
@@ -210,7 +212,7 @@ int obi_uprobe_netFdReadRet(struct pt_regs *ctx) {
                                        (protocol_selector_t){.http = 1, .http2 = 0, .tcp = 1},
                                        &net_ptr->p_conn,
                                        buf,
-                                       len,
+                                       bytes_len,
                                        NO_SSL,
                                        TCP_RECV,
                                        orig_dport);
--- a/bpf/gotracer/go_net.c
+++ b/bpf/gotracer/go_net.c
@@ -238,8 +238,9 @@ int obi_uprobe_netFdWrite(struct pt_regs *ctx) {
 
     void *fd_ptr = GO_PARAM1(ctx);
     u8 *buf = GO_PARAM2(ctx);
-    u64 len = (u64)GO_PARAM3(ctx);
+    s64 len = (s64)GO_PARAM3(ctx);
     if (buf && len > 0) {
+        const int bytes_len = (int)min((s64)__INT_MAX__, len);
         pid_connection_info_t p_conn = {0};
 
         if (!get_conn_info_from_fd(fd_ptr, &p_conn.conn, false)) {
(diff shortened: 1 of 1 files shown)


bpf/gotracer/go_net.c
```diff
@@ -15,6 +15,8 @@
 
 //go:build obi_bpf_ignore
 
+#include <limits.h>
```
rafaelroquetto: we can't include system headers, they will directly or transiently include libc headers (ebpf programs do not link against libc)

bpf/gotracer/go_net.c
```diff
@@ -193,9 +193,11 @@ int obi_uprobe_netFdReadRet(struct pt_regs *ctx) {
 
     void *buf = (void *)net_ptr->byte_ptr;
 
-    u64 len = (u64)GO_PARAM1(ctx);
-    bpf_dbg_printk("buf=%llx, len=%d === ", buf, len);
+    s64 len = (s64)GO_PARAM1(ctx);
+    bpf_dbg_printk("buf=%llx, len=%lld === ", buf, len);
     if (buf && len > 0) {
+        const int bytes_len = (len > __INT_MAX__) ? __INT_MAX__ : (int)len;
```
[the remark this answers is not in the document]
rafaelroquetto: copilot is wrong here

bpf/gotracer/go_net.c
rafaelroquetto: This is why `__INT_MAX__` (the builtin) should be preferred.

bpf/gotracer/go_net.c
```diff
@@ -193,9 +193,11 @@ int obi_uprobe_netFdReadRet(struct pt_regs *ctx) {
 
     void *buf = (void *)net_ptr->byte_ptr;
 
-    u64 len = (u64)GO_PARAM1(ctx);
-    bpf_dbg_printk("buf=%llx, len=%d === ", buf, len);
+    s64 len = (s64)GO_PARAM1(ctx);
+    bpf_dbg_printk("buf=%llx, len=%lld === ", buf, len);
```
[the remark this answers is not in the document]
MrAlias: Addressed in `95ec5c2a`. The debug print now casts `buf` to `unsigned long long` and `len` to `long long` so the varargs types match `%llx` and `%lld`.

bpf/gotracer/go_net.c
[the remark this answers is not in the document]
MrAlias: Addressed in `95ec5c2a`. The clamp now uses `INT_MAX` with `<limits.h>`, and I verified `bpf/gotracer/go_net.c` still compiles directly with `clang -target bpf` in this repo's BPF toolchain.

bpf/gotracer/go_net.c
```diff
@@ -193,9 +195,12 @@ int obi_uprobe_netFdReadRet(struct pt_regs *ctx) {
 
     void *buf = (void *)net_ptr->byte_ptr;
 
-    u64 len = (u64)GO_PARAM1(ctx);
-    bpf_dbg_printk("buf=%llx, len=%d === ", buf, len);
+    s64 len = (s64)GO_PARAM1(ctx);
+    bpf_dbg_printk(
+        "buf=%llx, len=%lld === ", (unsigned long long)buf, (long long)len);
     if (buf && len > 0) {
+        const int bytes_len = (len > INT_MAX) ? INT_MAX : (int)len;
```
rafaelroquetto: nit: you can `#include <common/algorithm.h>` and use

```
const int bytes_len = min(INT_MAX, len);
```

bpf/gotracer/go_net.c
MrAlias: __INT_MAX__ formatting had me pretty confused for a second there 😆 

Thanks for the clarifying edit.

MrAlias: Thanks for the catch. I reverted the system header include in `4df62f4e`, so `bpf/gotracer/go_net.c` no longer pulls in `<limits.h>`. The clamp now stays on the builtin path, and the file still compiles cleanly with a direct `clang -target bpf` check.

bpf/gotracer/go_net.c
MrAlias: Thanks, I followed up with the `min(...)` style in `bd888114` by including `common/algorithm.h` and switching the clamp to `min((s64)__INT_MAX__, len)`. I kept the builtin instead of `INT_MAX` so we avoid reintroducing a system header in the BPF C path.

rafaelroquetto (approved):
LGTM! Good stuff, thanks for fixing this!

MrAlias:
hmm, I think we need to address `obi_uprobe_netFdWrite` as well. 👀

mmat11 (approved):
lgtm!

PR #1884, opened 2026-04-21
Add debug events to `netolly` and `statsolly` userspace

pinoOgni:
## Summary

This PR enables `netolly` and `statsolly` userspace to read from the shared `debug_events` map via a common reader.

I noticed that `bpf_dbg_printk` calls in netolly were lost in userspace. `statsolly` only uses `bpf_d_printk`, but it was also missing a userspace reader.

## Validation

### DIFF
--- a/pkg/internal/ebpf/logger/logger.go
+++ b/pkg/internal/ebpf/logger/logger.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"log/slog"
 
+	"github.com/cilium/ebpf"
 	"golang.org/x/sys/unix"
 
 	"go.opentelemetry.io/obi/pkg/appolly/app/request"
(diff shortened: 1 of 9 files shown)


pkg/internal/ebpf/logger/logger.go
```diff
@@ -85,12 +86,45 @@ func (p *BPFLogger) Run(ctx context.Context) {
...
+func (p *BPFLogger) processLogEvent(record *ringbuf.Record) (request.Span, bool, error) {
+	logDebugEvent(p.log, record)
 	return request.Span{}, true, nil
 }
+
+// ReadDebugEventsMap can be used by any subsystem that loads BPF programs including
+// bpf_dbg.h but doesn't go through the main appolly pipeline (e.g. statsolly, netolly).
+// This is a blocking function. Callers should invoke it with `go ReadDebugEventsMap(..)`.
+func ReadDebugEventsMap(ctx context.Context, debugEventsMap *ebpf.Map, log *slog.Logger) {
+	if debugEventsMap == nil {
```
mmat11, line 107: does this suffer from the same read/delete race of `StatsEventsMap()`? should this whole section be readlocked?

pinoOgni: not sure about the question but I don't think there's a race here, and I don't think `StatsEventsMap()` has one either.

mmat11: I had it confused with the netolly map

pkg/internal/ebpf/logger/logger.go
```diff
@@ -85,12 +86,45 @@ func (p *BPFLogger) Run(ctx context.Context) {
...
+	reader, err := ringbuf.NewReader(debugEventsMap)
+	if err != nil {
+		log.Error("failed to create debug events reader", "error", err)
+		return
+	}
+	stop := context.AfterFunc(ctx, func() { reader.Close() })
+	defer stop()
+
+	for {
+		record, err := reader.Read()
```
rafaelroquetto, line 120: just a quick suggestion: perhaps we should use `ReadInto()` to avoid reallocating the record internal buffers each time

rafaelroquetto (approved):

mmat11 (approved):

PR #1885, opened 2026-04-21
feat(embedding): add embedding support

NameHaibinZhang:
## Embedding Detection Architecture

Embedding monitoring is handled in two layers:

### Embedding requests already handled by existing provider detectors

- **OpenAI embeddings**: Detected by `OpenAISpan` via response headers; when the URL path contains `/v1/embeddings`, the operation name is automatically overridden to `"embeddings"`.
- **Qwen/DashScope embeddings**: Detected by `QwenSpan` via the `X-DashScope-Request-Id` response header; `extractQwenOperation()` maps the `/embeddings` URL path to the operation name `"embedding"`.

### Embedding requests handled by the standalone `embedding.go`

Covers only **Voyage AI**, **Cohere**, and **Jina AI** — three embedding-only providers.

| Reason | Detail |
|:-------|:-------|
| No existing vendor to belong to | These providers have no chat/completion API and do not fit into any existing vendor detector |
| No identifiable response headers | Unlike OpenAI (`Openai-Version`) or Anthropic (`Anthropic-Organization-Id`), these providers do not return vendor-specific headers; detection relies on hostname + URL path matching |
| Different API formats | e.g., Cohere uses `texts` instead of `input`, and token counts live under `meta.billed_units` rather than `usage`, requiring a dedicated `VendorEmbedding` data structure |

## Validation

### DIFF
--- a/devdocs/config/config-schema.json
+++ b/devdocs/config/config-schema.json
@@ -756,6 +756,16 @@
       },
       "type": "object"
     },
+    "EmbeddingProviderConfig": {
+      "properties": {
+        "enabled": {
+          "type": "boolean",
+          "description": "Enable generic embedding provider (Voyage AI, Cohere, Jina AI) payload extraction and parsing",
+          "x-env-var": "OTEL_EBPF_HTTP_EMBEDDING_ENABLED"
+        }
+      },
+      "type": "object"
+    },
     "EnrichmentConfig": {
       "properties": {
         "enabled": {
@@ -849,6 +859,10 @@
         "bedrock": {
           "$ref": "#/$defs/BedrockConfig",
           "description": "AWS Bedrock payload extraction and parsing"
+        },
+        "embedding": {
+          "$ref": "#/$defs/EmbeddingProviderConfig",
+          "description": "Generic embedding provider (Voyage AI, Cohere, Jina AI) payload extraction and parsing"
         },
         "gemini": {
           "$ref": "#/$defs/GeminiConfig",
--- a/devdocs/config/config-schema.json
+++ b/devdocs/config/config-schema.json
@@ -762,7 +762,7 @@
           "type": "boolean",
           "description": "Enable generic embedding provider (Voyage AI, Cohere, Jina AI) payload extraction and parsing",
           "default": false,
-          "x-env-var": "OTEL_EBPF_HTTP_EMBEDDING_ENABLED"
+          "x-env-var": "OTEL_EBPF_HTTP_GENAI_EMBEDDING_ENABLED"
         }
       },
       "type": "object"
--- a/devdocs/config/config-schema.json
+++ b/devdocs/config/config-schema.json
@@ -761,6 +761,7 @@
         "enabled": {
           "type": "boolean",
           "description": "Enable generic embedding provider (Voyage AI, Cohere, Jina AI) payload extraction and parsing",
+          "default": false,
           "x-env-var": "OTEL_EBPF_HTTP_EMBEDDING_ENABLED"
         }
       },
--- a/pkg/appolly/app/request/span.go
+++ b/pkg/appolly/app/request/span.go
@@ -1687,6 +1687,10 @@ func (s *Span) GenAIInputTokens() int {
 		return s.GenAI.Bedrock.Output.InputTokens
 	}
 
+	if s.GenAI.Embedding != nil {
+		return s.GenAI.Embedding.GetInputTokens()
+	}
+
 	return 0
 }
 
@@ -1714,6 +1718,8 @@ func (s *Span) GenAIOutputTokens() int {
 	if s.GenAI.Bedrock != nil {
 		return s.GenAI.Bedrock.Output.OutputTokens
 	}
+
+	// Embedding APIs do not produce output tokens.
 
 	return 0
 }
--- a/pkg/appolly/app/request/span.go
+++ b/pkg/appolly/app/request/span.go
@@ -748,6 +748,15 @@ func (e *VendorEmbedding) GetInputTokens() int {
 	return 0
 }
 
+// GetOutputTokens returns the output token count for embedding requests,
+// derived as total_tokens - prompt_tokens.
+func (e *VendorEmbedding) GetOutputTokens() int {
+	if e.Output.Usage.TotalTokens > 0 && e.Output.Usage.PromptTokens > 0 {
+		return e.Output.Usage.TotalTokens - e.Output.Usage.PromptTokens
+	}
+	return 0
+}
+
 // Span contains the information being submitted by the following nodes in the graph.
 // It enables comfortable handling of data from Go.
 // REMINDER: any attribute here must be also added to the functions SpanOTELGetters
@@ -1719,7 +1728,9 @@ func (s *Span) GenAIOutputTokens() int {
 		return s.GenAI.Bedrock.Output.OutputTokens
 	}
 
-	// Embedding APIs do not produce output tokens.
+	if s.GenAI.Embedding != nil {
+		return s.GenAI.Embedding.GetOutputTokens()
+	}
 
 	return 0
 }
--- a/pkg/appolly/app/request/span.go
+++ b/pkg/appolly/app/request/span.go
@@ -299,7 +299,17 @@ func (u *OpenAIUsage) GetOutputTokens() int {
 		return u.OutputTokens
 	}
 
-	return u.CompletionTokens
+	if u.CompletionTokens > 0 {
+		return u.CompletionTokens
+	}
+
+	// Embedding responses only report prompt_tokens and total_tokens.
+	// Derive output tokens from the difference.
+	if u.TotalTokens > 0 && u.PromptTokens > 0 {
+		return u.TotalTokens - u.PromptTokens
+	}
+
+	return 0
 }
 
 type OpenAIError struct {
--- a/pkg/appolly/app/request/span.go
+++ b/pkg/appolly/app/request/span.go
@@ -629,10 +629,10 @@ const EmbeddingOperationName = "embeddings"
 // VendorEmbedding represents a generic embedding API provider such as
 // Voyage AI, Cohere, or Jina AI.
 type VendorEmbedding struct {
-	Provider   string
-	Model      string
-	Input      EmbeddingRequest
-	Output     EmbeddingResponse
+	Provider string
+	Model    string
+	Input    EmbeddingRequest
+	Output   EmbeddingResponse
 }
 
 // OperationName returns the canonical embedding operation name.
@@ -646,7 +646,7 @@ type EmbeddingRequest struct {
 	Input      json.RawMessage `json:"input"`
 	Dimensions int             `json:"dimensions,omitempty"`
 	// Cohere uses "texts" instead of "input"
-	Texts      json.RawMessage `json:"texts,omitempty"`
+	Texts json.RawMessage `json:"texts,omitempty"`
 }
 
 // InputCount returns the number of input texts in the request.
--- a/pkg/appolly/app/request/span.go
+++ b/pkg/appolly/app/request/span.go
@@ -671,10 +671,10 @@ const EmbeddingOperationName = "embeddings"
 // VendorEmbedding represents a generic embedding API provider such as
 // Voyage AI, Cohere, or Jina AI.
 type VendorEmbedding struct {
-	Provider   string
-	Model      string
-	Input      EmbeddingRequest
-	Output     EmbeddingResponse
+	Provider string
+	Model    string
+	Input    EmbeddingRequest
+	Output   EmbeddingResponse
 }
 
 // OperationName returns the canonical embedding operation name.
@@ -688,7 +688,7 @@ type EmbeddingRequest struct {
 	Input      json.RawMessage `json:"input"`
 	Dimensions int             `json:"dimensions,omitempty"`
 	// Cohere uses "texts" instead of "input"
-	Texts      json.RawMessage `json:"texts,omitempty"`
+	Texts json.RawMessage `json:"texts,omitempty"`
 }
 
 // InputCount returns the number of input texts in the request.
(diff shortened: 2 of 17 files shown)


pkg/ebpf/common/http_transform.go
```diff
@@ -169,6 +169,13 @@ func httpRequestResponseToSpan(parseCtx *EBPFParseContext, event *BPFHTTPInfo, r
 		}
 	}
 
+	if parseCtx != nil && parseCtx.payloadExtraction.HTTP.GenAI.Embedding.Enabled {
+		span, ok := ebpfhttp.EmbeddingSpan(&httpSpan, req, resp)
+		if ok {
+			return span
+		}
+	}
```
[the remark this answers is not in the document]
NameHaibinZhang: ok

pkg/ebpf/common/http/openai.go
```diff
@@ -52,6 +53,11 @@ func OpenAISpan(baseSpan *request.Span, req *http.Request, resp *http.Response)
 
 	parsedResponse.Request = parsedRequest
 
+	// Override operation name for embedding requests.
+	if req.URL != nil && strings.Contains(req.URL.Path, "/v1/embeddings") {
+		parsedResponse.OperationName = request.EmbeddingOperationName
+	}
```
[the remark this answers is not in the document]
NameHaibinZhang: ok

pkg/appolly/app/request/span.go
```diff
@@ -1537,6 +1638,9 @@ func (s *Span) GenAIOperationName() string {
 	if s.GenAI.Bedrock != nil {
 		return "invoke_model"
 	}
+	if s.GenAI.Embedding != nil {
+		return s.GenAI.Embedding.OperationName()
+	}
```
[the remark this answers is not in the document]
NameHaibinZhang: ok

pkg/ebpf/common/http/embedding.go
```diff
@@ -0,0 +1,108 @@
...
+	if provider == "" {
+		return *baseSpan, false
+	}
+
+	reqB, err := io.ReadAll(req.Body)
+	if err != nil {
+		slog.Debug("EmbeddingSpan: failed to read request body", "provider", provider, "error", err)
+		return *baseSpan, false
+	}
+	req.Body = io.NopCloser(bytes.NewBuffer(reqB))
```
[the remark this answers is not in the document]
NameHaibinZhang: Good catch — the provider is already confirmed by hostname+path matching at this point, so a body read failure shouldn't cause the detection to fall through to downstream detectors (which could misclassify it). Updated to treat request body parsing as best-effort (matching the existing response body handling), so the span is still correctly classified as an embedding span with empty request-derived fields when the body is unavailable.

devdocs/config/config-schema.json
```diff
@@ -745,6 +745,17 @@
       },
       "type": "object"
     },
+    "EmbeddingProviderConfig": {
+      "properties": {
+        "enabled": {
+          "type": "boolean",
+          "description": "Enable generic embedding provider (Voyage AI, Cohere, Jina AI) payload extraction and parsing",
+          "default": false,
+          "x-env-var": "OTEL_EBPF_HTTP_EMBEDDING_ENABLED"
```
mariomac: "Embedding" is a concept that is so ubiquitous in computing that I had to google what does it mean in this context LOL

What do you think about adding an extra prefix to this set of environment variables so users get a sharper idea about what they refer to? For example, something like `OTEL_EBPF_RAG_HTTP_EMBEDDING_...` (or whatever else is meaningful and correct)

mariomac: Maybe `OTEL_EBPF_HTTP_GENAI_EMBEDDING` to be consistent with the `ebpf.payload_extraction.http.genai.embedding` property documented below?

pkg/ebpf/common/http_transform.go
```diff
@@ -169,6 +169,13 @@ func httpRequestResponseToSpan(parseCtx *EBPFParseContext, event *BPFHTTPInfo, r
 		}
 	}
 
+	if parseCtx != nil && parseCtx.payloadExtraction.HTTP.GenAI.Embedding.Enabled {
```
rafaelroquetto: should we be checking of `isClientEvent(event.Type)` here as well?  or do we wanna capture server side events as well?

pkg/ebpf/common/http/embedding.go
```diff
@@ -0,0 +1,115 @@
...
+	if len(respB) > 0 {
+		if err := json.Unmarshal(respB, &parsedResponse); err != nil {
+			slog.Debug("failed to parse embedding response", "provider", provider, "error", err)
+		}
+	}
+
+	model := parsedRequest.Model
+	if model == "" {
+		model = parsedResponse.Model
+	}
```
rafaelroquetto: I am not sure this is semantically correct - when the request has no model, then Embedding.Model will effectively be parsedResponse.Model. In this case, GetAIRequestModel() will return the _response_ model, rather than the request one. Is this intended? Same for `TraceName()`

pkg/ebpf/common/http/embedding.go
```diff
@@ -0,0 +1,115 @@
...
+func isEmbeddingProvider(req *http.Request) string {
+	if req == nil || req.URL == nil {
+		return ""
+	}
+
+	host := extractHostname(req)
+	path := req.URL.Path
+
+	for _, hp := range embeddingHostPatterns {
+		if strings.HasSuffix(host, hp.hostSuffix) &&
```
rafaelroquetto: I think this should be:
```
if (host == hp.hostSuffix || strings.HasSuffix(host, "."+hp.hostSuffix)) &&
    strings.HasSuffix(path, hp.pathSuffix) {
```

otherwise this will match `foobar.baz.ca` with suffix `bar.baz.ca` which does not sound right - but maybe I am misunderstanding the code

pkg/ebpf/common/http_transform.go
grcevski: +1

pkg/appolly/app/request/span.go
```diff
@@ -98,14 +98,16 @@ const (
...
 )
 
 func IsGenAISubtype(subtype int) bool {
 	return subtype == HTTPSubtypeOpenAI ||
 		subtype == HTTPSubtypeAnthropic ||
 		subtype == HTTPSubtypeGemini ||
 		subtype == HTTPSubtypeQwen ||
-		subtype == HTTPSubtypeAWSBedrock
+		subtype == HTTPSubtypeAWSBedrock ||
+		subtype == HTTPSubtypeEmbedding
```
grcevski, line 115: I think this will cause issues with the metrics generation. Essentially we say embedding is a GenAISubType, however the token count functions would not find anything and return 0. If I understand correctly this is just another OpenAI call, if there are tokens in the response we should adjust the code in span.go(line 1604...). If not we should probably create another helper function, something like GenAITokenCounterType(subtype int) which will not include embedding and it will be used in metrics.go and prom.go to guard the population of the token metrics.

grcevski (commented):
Looks great overall, I think we just need to address the comments and we should be good to merge.

pkg/ebpf/common/http/openai.go
```diff
@@ -52,6 +53,11 @@ func OpenAISpan(baseSpan *request.Span, req *http.Request, resp *http.Response)
 
 	parsedResponse.Request = parsedRequest
 
+	// Override operation name for embedding requests.
+	if req.URL != nil && strings.Contains(req.URL.Path, "/v1/embeddings") {
+		parsedResponse.OperationName = request.EmbeddingOperationName
```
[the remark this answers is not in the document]
NameHaibinZhang: Adopted the suggestion. Now trimming the trailing / and using an exact match (path == "/v1/embeddings") to avoid false positives on unintended paths like /v1/embeddingsX.

pkg/ebpf/common/http_transform.go
NameHaibinZhang: Good catch! Added the isClientEvent(event.Type) guard to align with the other GenAI detectors (OpenAI, Anthropic, Gemini, etc.), ensuring embedding detection only fires for client-side events.

pkg/appolly/app/request/span.go
NameHaibinZhang: Fixed. Added the Embedding case to GenAIInputTokens(), delegating to VendorEmbedding.GetInputTokens() for correct input token reporting. Since embedding APIs do not produce output tokens (no generated text), GenAIOutputTokens() intentionally returns 0 for embeddings, with a comment explaining this.

pkg/ebpf/common/http/embedding.go
NameHaibinZhang: You're right, this was semantically incorrect. Removed the model fallback logic — VendorEmbedding.Model now only holds the request model (empty if the request didn't specify one). The response model remains accessible via Output.Model. In tracesgen.go, GenAIRequestModel uses Input.Model (request model) and GenAIResponseModel uses Output.Model (response model), which is semantically correct.

pkg/ebpf/common/http/embedding.go
NameHaibinZhang: Good catch! Fixed by changing the match to host == hp.hostSuffix || strings.HasSuffix(host, "."+hp.hostSuffix), ensuring we only match exact hostnames or legitimate subdomains separated by .. This prevents foobar.baz.ca from falsely matching the suffix bar.baz.ca

devdocs/config/config-schema.json
NameHaibinZhang: Agreed, renamed the env var from OTEL_EBPF_HTTP_EMBEDDING_ENABLED to OTEL_EBPF_HTTP_GENAI_EMBEDDING_ENABLED to be consistent with the YAML path ebpf.payload_extraction.http.genai.embedding and avoid ambiguity around the generic term "embedding". Updated payload_extraction.go, config-schema.json, and CONFIG.md accordingly.

NameHaibinZhang:
@mariomac @rafaelroquetto @grcevski All comments have been addressed. Could you please take another look? Once the embedding support is merged, I will proceed with the remaining tasks, such as rerank and tools.

mariomac (approved):
LGTM! Waiting for the rest of commenters' OK.

marctc (approved):

pkg/appolly/app/request/span.go
```diff
@@ -296,7 +299,17 @@ func (u *OpenAIUsage) GetOutputTokens() int {
...
 
-	return u.CompletionTokens
+	if u.CompletionTokens > 0 {
+		return u.CompletionTokens
+	}
+
+	// Embedding responses only report prompt_tokens and total_tokens.
+	// Derive output tokens from the difference.
+	if u.TotalTokens > 0 && u.PromptTokens > 0 {
+		return u.TotalTokens - u.PromptTokens
```
grcevski, line 309: Nice!

grcevski (approved):
LGTM!

MrAlias (commented):
I found one remaining edge case in the generic embedding matcher that looks worth fixing before merge.

pkg/ebpf/common/http/embedding.go
```diff
@@ -0,0 +1,110 @@
...
+// isEmbeddingProvider checks whether the request targets a known embedding-only
+// provider by matching the hostname and URL path against embeddingHostPatterns.
+// Returns the provider name if matched, or empty string otherwise.
+func isEmbeddingProvider(req *http.Request) string {
+	if req == nil || req.URL == nil {
+		return ""
+	}
+
+	host := extractHostname(req)
+	path := req.URL.Path
```
MrAlias: This still misses trailing-slash variants for the generic embedding providers. A request like `https://api.voyageai.com/v1/embeddings/` or `https://api.cohere.com/v2/embed/` will not match here, so it falls back to plain HTTP or another detector and loses the GenAI attributes this PR adds.

I think the fix should mirror the normalization you already added in `openai.go`: trim a trailing `/` before matching, then add a regression test for the slash-suffixed form.

NameHaibinZhang: Good catch! Fixed by adding strings.TrimSuffix(path, "/") before matching, mirroring the normalization already applied in openai.go. Also added regression tests for the trailing-slash variants of all three providers

NameHaibinZhang:
@MrAlias All comments have been addressed. Could you please take another look?

pkg/ebpf/common/http/embedding.go
```diff
@@ -0,0 +1,110 @@
...
+	if req.Body != nil {
+		var err error
+		reqB, err = io.ReadAll(req.Body)
+		if err != nil {
+			slog.Debug("EmbeddingSpan: failed to read request body, continuing without it", "provider", provider, "error", err)
+		}
+		req.Body = io.NopCloser(bytes.NewBuffer(reqB))
+	}
+
+	// Response body parsing is best-effort: gzip-compressed or truncated
```
mmat11: compressed bodies are already handled in `getResponseBody`, truncated bodies should fail to get parsed earlier; I'd modify the comment

pkg/ebpf/common/http/embedding.go
```diff
@@ -0,0 +1,110 @@
...
+var embeddingHostPatterns = []embeddingHostPattern{
+	{"api.voyageai.com", "/v1/embeddings", "voyage"},
+	{"api.cohere.com", "/v2/embed", "cohere"},
+	{"api.jina.ai", "/v1/embeddings", "jina"},
+}
+
+// isEmbeddingProvider checks whether the request targets a known embedding-only
+// provider by matching the hostname and URL path against embeddingHostPatterns.
+// Returns the provider name if matched, or empty string otherwise.
+func isEmbeddingProvider(req *http.Request) string {
```
mmat11: nit: I would expect a bool as return given the function name, maybe we should name it "parseEmbeddingProvider" or similar

mmat11 (approved):
left a couple nits, lgtm!

pkg/ebpf/common/http/embedding.go
NameHaibinZhang: Good point — updated the comment to only mention truncated bodies, since gzip decompression is already handled inside getResponseBody.

pkg/ebpf/common/http/embedding.go
NameHaibinZhang: Makes sense — renamed to parseEmbeddingProvider to better reflect the string return type.

Issue #1330, opened 2026-02-18
Document k8s-cache


pellared:
SIG meeting notes:

Question: What is k8s-cache? I do not see it documented in [opentelemetry.io](http://opentelemetry.io/). I thought it was something used only for testing.

Answer: When obi runs as a daemonset, each obi instance opens a connection to the k8s API server. If a large k8s cluster has enough nodes, it will bring down the API server (even in GKE or other managed control planes). When this happens, the entire cluster becomes unusable. K8s-cache centralizes the API server connections to only a few replicas of a deployment (say 1 replica per 50 nodes), so it complements the daemonset deployment. The obi daemonsets then balance all of their connections to the k8s-cache replicas via a k8s-cache service, instead of directly hitting k8s API.

Action Item:
Ensure that is documented k8s-cache in (any that apply):
  - https://opentelemetry.io/docs/zero-code/obi/setup/kubernetes-helm/
  - https://opentelemetry.io/docs/zero-code/obi/setup/kubernetes/ 
  - https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-ebpf-instrumentation
  - https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/tree/main/devdocs

PR #1886, opened 2026-04-21
document k8s-cache

NimrodAvni78:
## Summary

Part of https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1330
This is more of an internal documentation for developers
a more high level documentation to [opentelemetry.io](http://opentelemetry.io) will come shortly

## Validation


MrAlias (commented):
Thanks for putting this together. This is useful internal documentation, especially the overview of the service, the code pointers, and the deployment guidance. I left a few comments where the doc appears to describe behavior or requirements more strongly than the current implementation supports.

devdocs/k8s-cache.md
```diff
@@ -0,0 +1,248 @@
...
+2. Exposes a gRPC streaming endpoint
+   (`informer.EventStreamService/Subscribe`) that clients connect to. On
+   subscription, the service replays the current snapshot and then keeps
+   pushing add/update/delete events in real time. An explicit
+   `SYNC_FINISHED` event marks the end of the initial replay, so the client
+   knows when it is safe to start decorating telemetry.
+
+Clients send a `FromTimestampEpoch` on `Subscribe`. On reconnect, OBI sends the
+timestamp of the last event it successfully processed so the cache can skip
+anything older and avoid a full snapshot replay.
```
MrAlias, line 47: This reads like reconnects can replay a true delta of what was missed, but the current implementation is narrower than that. `FromTimestampEpoch` is only used to filter the current in-memory snapshot in `meta.Informers.sortAndCut`; there is no persisted event log. That means a client can still miss deletes, and anything that disappeared before reconnect will not be replayed. I think this section should be softened so it does not over-promise recovery behavior.

devdocs/k8s-cache.md
```diff
@@ -0,0 +1,248 @@
...
+1. Defaults from `DefaultConfig` in
+   [pkg/kube/kubecache/config.go](../pkg/kube/kubecache/config.go).
+2. A YAML file passed via `--config` or `OTEL_EBPF_K8S_CACHE_CONFIG_PATH`.
+3. Environment variables.
+
+| YAML key                 | Env var                                                | Default        | Purpose                                                    |
+|--------------------------|--------------------------------------------------------|----------------|------------------------------------------------------------|
+| `log_level`              | `OTEL_EBPF_K8S_CACHE_LOG_LEVEL`                        | `info`         | `debug`/`info`/`warn`/`error`.                             |
+| `port`                   | `OTEL_EBPF_K8S_CACHE_PORT`                             | `50055`        | gRPC listen port.                                          |
+| `max_connections`        | `OTEL_EBPF_K8S_CACHE_MAX_CONNECTIONS`                  | `150`          | Max concurrent subscribing OBI clients.                    |
```
MrAlias, line 200: This wording sounds like `max_connections` is a total cap on subscribing OBI clients, but the server currently wires it into `grpc.MaxConcurrentStreams`, which limits streams per HTTP/2 transport rather than acting as a global client limit. Since each OBI instance creates its own gRPC connection in `cache_svc_client.connect`, I think this should be clarified.

devdocs/k8s-cache.md
```diff
@@ -0,0 +1,248 @@
...
+| `profile_port`           | `OTEL_EBPF_K8S_CACHE_PROFILE_PORT`                     | `0` (disabled) | If non-zero, starts a `net/http/pprof` listener.           |
+| `informer_resync_period` | `OTEL_EBPF_K8S_CACHE_INFORMER_RESYNC_PERIOD`           | `30m`          | Full informer resync interval. Increase to lower API load. |
+| `informer_send_timeout`  | `OTEL_EBPF_K8S_CACHE_INFORMER_SEND_TIMEOUT`            | `10s`          | Drops a subscriber that does not drain an event in time.   |
```
MrAlias, line 203: I think this is documenting behavior that is not implemented yet. `pkg/kube/kubecache/service/service.go` stores `sendTimeout` on the connection, but `handleMessagesQueue` never uses it and never calls `MessageTimeout()`. As written, readers will expect slow subscribers to be dropped after a per-message deadline, and the metrics section later suggests the same. This should either be removed for now or rewritten to match the current behavior.

devdocs/k8s-cache.md
```diff
@@ -0,0 +1,248 @@
...
+## Minimum RBAC permissions
+
+The cache needs `list` and `watch` on the same resources OBI would otherwise
+watch itself. The integration tests use the service account in
+[internal/test/integration/k8s/manifests/01-serviceaccount.yml](../internal/test/integration/k8s/manifests/01-serviceaccount.yml):
+
+```yaml
+rules:
+  - apiGroups: [ "apps" ]
+    resources: [ "replicasets" ]
```
MrAlias, line 221: The minimum-RBAC section currently includes `replicasets`, but `pkg/kube/kubecache/meta.InitInformers` only creates Pod, Node, and Service informers. Since this section is framed as the minimum required permissions, I think it should avoid granting access that the service does not currently use.

devdocs/k8s-cache.md
NimrodAvni78: yeah you are right
will remove this comment
opened a [separate issue](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1903) on this to be fixed separately

devdocs/k8s-cache.md
NimrodAvni78: fixed

devdocs/k8s-cache.md
```diff
@@ -0,0 +1,239 @@
...
+   `OTEL_EBPF_KUBE_INFORMERS_SYNC_TIMEOUT` expires, so decoration does not
+   start with partial metadata.
+3. On disconnect, sleeps for `ReconnectInitialInterval` and retries.
+
+The wire types are defined in
+[proto/informer.proto](../proto/informer.proto). Any change to
+the event schema must stay backwards-compatible with already-deployed OBI
+instances that connect to a newer cache (and vice versa).
+
+## How to deploy
```
mariomac, line 95: I'd add a first paragraph saying something like:

If you are using our [OBI Helm chart](https://opentelemetry.io/docs/zero-code/obi/setup/kubernetes-helm/), you just have to provide a non-zero value for the
`k8sCache > replicas` configuration option in [values.yaml](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-ebpf-instrumentation/examples/with-k8s-cache/values.yaml).

MrAlias (commented):
This mostly looks good. I found two places where the doc drifted from the updated implementation on `main` after the recent merge, plus one reconnect description that still reads a bit stronger than the code guarantees.

MrAlias:
Github is giving me issues 😞 . There are supposed to be 3 inline comments for that review that it is failing to resolve/render. I'm looking into it.

MrAlias:
Things are all jacked. I'll just post the feedback as comments. Sorry for the mess that is incoming.

MrAlias:
lines 62-63:

nit: the code only cuts the snapshot at `StatusTimeEpoch >= fromTimestampEpoch`. Since `status_time_epoch` is only second-granularity, reconnects can still replay items from the same timestamp window.

```suggestion
- **Cheaper reconnects.** OBI passes the timestamp of its last event, so the
  snapshot replay on resubscribe can skip older entries. This reduces replay
  volume, but it is not a durable event log and reconnects can still replay
  entries from the same timestamp window.
```

MrAlias:
line 206:

`informer_send_timeout` is still part of the updated implementation on main, so I think it should stay in the config table.

```suggestion
| `informer_resync_period` | `OTEL_EBPF_K8S_CACHE_INFORMER_RESYNC_PERIOD`           | `30m`          | Full informer resync
interval. Increase to lower API load.                |
| `informer_send_timeout`  | `OTEL_EBPF_K8S_CACHE_INFORMER_SEND_TIMEOUT`            | `10s`          | Per-message send deadline
before a slow subscriber connection is closed.  |
```

MrAlias:
lines 242-244:

The updated implementation on main still emits the timeout label here, so I think the doc should mention it.

```suggestion
- `*_kube_cache_client_messages_total{status=submit|success|timeout|error}` —
  outcome of events forwarded to clients. Growing `timeout` means a slow
  subscriber hit `informer_send_timeout`; growing `error` means the stream
  failed during `Send`.
```

NimrodAvni78:
@MrAlias done

mmat11 (approved):
lgtm

MrAlias (approved):
🚀

PR #1888, opened 2026-04-21
notices: run notices-update in a container, run for amd64 and arm64

### LINKED ISSUE #1224: go-notices-update should run on both amd64 and arm64
when running `make go-notices-update` which runs go-licenses, it runs only on the specific OS you are running the image on
our CI runs on amd64. which generates the notices and makes sure the work tree is clean.
this creates 2 issues:
1. we don't save arm64 specific notices
2. running `make go-notices-update` on arm64 creates diffs from when our CI runs this

also relates to https://github.com/google/go-licenses/issues/187

proposed solution:
change go-licenses to run go-licenses tool with both amd64 and arm64 GOARCH, add both notices (not supported by go-licenses)

mmat11:
## Summary

Fixes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1224

This PR generates per-arch notices in NOTICES/amd64 and NOTICES/arm64. Most licenses are now duplicated, but this makes checking for diffs easier. Any custom merge approach would make the check phase more convoluted, I'm not sure it's worth it

## Validation


MrAlias (commented):
Thanks for updating the notice generation to keep Go notices separated per architecture. There is one release-packaging gap that still needs to be addressed.

The new layout stores notices under `NOTICES/<arch>`, but the release packaging step still copies the entire `NOTICES/` directory into every per-architecture archive. That means an archive like `obi-*-linux-amd64.tar.gz` can now include arm64-only notice content as well, for example files under `NOTICES/arm64/...`.

This is a correctness issue for released artifacts: each architecture-specific archive should only ship the notices for its own target architecture. As written, the archive contents no longer cleanly match the binary being released, and the CycloneDX SBOM generated from unpacking that archive can end up describing files that do not actually belong to that build.

Please update the release packaging flow so each per-architecture archive includes only the matching notice tree for that architecture.

mmat11:
@MrAlias done

grcevski (approved):
LGTM!

Issue #1854, opened 2026-04-16
Proposal AI Agent Observability Support


NameHaibinZhang:
## Summary

This proposal introduces AI Agent observability capabilities to OpenTelemetry eBPF Instrumentation (OBI). As AI applications evolve from simple LLM calls into complex multi-step Agent workflows — involving tool calls, MCP protocol interactions, vector embeddings, reranking, and retrieval-augmented generation (RAG) — there is a growing need for transparent, zero-code observability into these systems. OBI's eBPF-based approach is uniquely positioned to monitor Agent behavior without requiring any application code changes.

### Current State

OBI currently supports basic LLM call monitoring for three GenAI providers:

| Provider | Detection Method | Supported Operations |
|:---------|:----------------|:---------------------|
| OpenAI | Response header matching (`Openai-Version`, etc.) | chat completion, response |
| Anthropic | Response header / hostname matching | message (including SSE streaming) |
| Google Gemini | Response header / URL pattern matching | generateContent, embedContent, countTokens |

These capabilities cover foundational LLM inference calls but **cannot capture higher-level Agent behaviors**, such as:

- How the Agent orchestration layer schedules multi-turn LLM calls and tool invocations
- Tool execution through MCP (Model Context Protocol) servers
- Embedding, retrieval, and reranking operations against vector databases
- Inter-agent communication and coordination

### Why AI Agent Observability Matters

1. **Agent architectures are becoming mainstream**: Frameworks like LangChain, CrewAI, AutoGen, Dify, and Coze have seen widespread adoption. Agent workflows — not single LLM calls — are the primary pattern for AI applications.

2. **Debugging complexity grows exponentially**: A single Agent task may involve dozens of LLM calls, tool invocations, and data retrievals. Without end-to-end observability, troubleshooting becomes extremely difficult.

3. **Explosive growth of MCP**: The Model Context Protocol has become the de facto standard for Agent-tool interaction, with support from Anthropic, OpenAI, Google, and other major providers.

4. **RAG pipeline performance bottlenecks**: Embedding and reranking are core operations in RAG systems; their latency and quality directly impact end-user experience.

[...]

PR #1892, opened 2026-04-22
feat(qwen): add qwen support

NameHaibinZhang:
## Summary

- Add Qwen (DashScope) GenAI HTTP parsing support across the end-to-end span pipeline.
- Detect Qwen traffic for both compatible-mode and DashScope generation APIs, and map parsed data to GenAI semantic attributes/metrics.
- Add integration coverage (OATS + mock server + test app) and update feature/support docs to include Qwen.

## Validation


marctc:
Thanks for your this! in #1854 you opened a proposal what to add to OBI regarding AI observability but I cant see any reference of qwen. 

I think it would be good for maintainers if in that issue you break down the issue with sub-issues to understand the scope of the project. Perhaps maintainers want to discuss if we want support for specific models/agents before adding support for all ones existing out there.

NameHaibinZhang:
> Thanks for your this! in #1854 you opened a proposal what to add to OBI regarding AI observability but I cant see any reference of qwen.
> 
> I think it would be good for maintainers if in that issue you break down the issue with sub-issues to understand the scope of the project. Perhaps maintainers want to discuss if we want support for specific models/agents before adding support for all ones existing out there.

Got it. I'll add the Qwen-related details to that issue

NameHaibinZhang:
@rafaelroquetto @marctc All comments have been addressed, and the branch has been rebased onto main. Qwen is an open-source large language model developed by Alibaba and widely adopted in China. Adding support for it in our project will enable Chinese users to better collect AI-related monitoring data. please review, thks.

pkg/appolly/app/request/span.go
```diff
@@ -252,6 +254,7 @@ type GenAI struct {
 	OpenAI    *VendorOpenAI
 	Anthropic *VendorAnthropic
 	Gemini    *VendorGemini
+	Qwen      *VendorOpenAI
```
marctc: is this intentional? or it's because share same format?

NameHaibinZhang: Yes, this is intentional. Qwen's DashScope compatible-mode API returns the exact same JSON structure as OpenAI (id, object, model, choices, usage). The native DashScope generation API uses slightly different field names (request_id, output, input_tokens/output_tokens), but VendorOpenAI already accommodates both naming conventions via its GetInputTokens()/GetOutputTokens() methods and the Output raw JSON field. Creating a separate VendorQwen would duplicate VendorOpenAI with no structural differences.

pkg/ebpf/common/http/qwen.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	slog.Debug("Qwen", "request", string(reqB), "response", string(respB))
+
+	var parsedRequest request.OpenAIInput
+	if err := json.Unmarshal(reqB, &parsedRequest); err != nil {
+		slog.Debug("failed to parse Qwen request", "error", err)
+	}
+	if parsedRequest.Model == "" {
+		if matches := modelFieldRegexp.FindSubmatch(reqB); len(matches) == 2 {
```
marctc: could this regex capture something that's in the prompt that has something like `"model":"gpt-4`?

NameHaibinZhang: Good catch. This regex is a best-effort fallback that only runs when json.Unmarshal fails on truncated request bodies (eBPF buffer limits). Since FindSubmatch returns the first match and the model field is conventionally serialized before messages, the risk is low in practice. However, I'll add a comment to clarify the limitation. Alternatively, I could restrict the regex search to the first N bytes to reduce the risk further

pkg/appolly/app/request/span.go
marctc: Maybe we could add a comment to clarify that. Readers could be confused

marctc (approved):
LGTM!

pkg/appolly/app/request/span.go
mariomac: +1 to add a comment.

Also, what about reusing the `OpenAI` field instead of duplicating it in `Qwen`? We already have the `SubType` field to differentiate both providers, and I guess it would avoid some code duplication.

Nevermind, if you consider that it would be cleaner to keep them in separate fields, I'm Ok with it.

mariomac (commented):
Amazing!! Thank you a lot for this contribution. LGTM. I have a minor, optional comment.

pkg/appolly/app/request/span.go
NameHaibinZhang: > +1 to add a comment.
> 
> Also, what about reusing the `OpenAI` field instead of duplicating it in `Qwen`? We already have the `SubType` field to differentiate both providers, and I guess it would avoid some code duplication.
> 
> Nevermind, if you consider that it would be cleaner to keep them in separate fields, I'm Ok with it.

Thanks for the suggestions! I've added a comment on the Qwen field explaining why it reuses VendorOpenAI.Regarding reusing the OpenAI field: I prefer keeping them as separate fields because multiple methods (GenAIProviderName(), GenAIResponseID(), GenAIResponseModel(), etc.) use s.GenAI.Qwen != nil / s.GenAI.OpenAI != nil to route provider-specific logic. Merging them would require checking SubType everywhere, which is less self-documenting. Separate fields also allow future divergence without refactoring if Qwen's API evolves away from OpenAI's structure.

mariomac (approved):

PR #1894, opened 2026-04-22
Fix Java Agent injection

mariomac:
## Summary

Some container users got their Java instrumentation broken because OBI was trying to write a copy of the agent JAR in `/.cache` (no permissions).

While working on a fix, I realized it was unnecessary caching the agent in disk because it's already in the executable memory, so it is injected directly from memory to the target filesystem. It simplifies a bit the code.

## Validation

### DIFF
--- a/pkg/internal/java/java_inject.go
+++ b/pkg/internal/java/java_inject.go
@@ -49,16 +49,16 @@ type JavaInjector struct {
 	cfg *obi.Config
 }
 
-func NewJavaInjector(cfg *obi.Config) (*JavaInjector, error) {
+func NewJavaInjector(cfg *obi.Config) *JavaInjector {
 	if !cfg.Java.Enabled {
-		return nil, nil
+		return nil
 	}
 	ensureEmbeddedAgent()
 
 	return &JavaInjector{
 		cfg: cfg,
 		log: slog.With("component", "javaagent.Injector"),
-	}, nil
+	}
 }
 
 func tempDirPath(root, dir string) (string, bool) {
--- a/pkg/internal/java/java_inject.go
+++ b/pkg/internal/java/java_inject.go
@@ -9,7 +9,6 @@ import (
 	"bufio"
 	"bytes"
 	"context"
-	"crypto/sha256"
 	_ "embed"
 	"errors"
 	"fmt"
@@ -35,14 +34,7 @@ const (
 )
 
 //go:embed embedded/obi-java-agent.jar
-var embeddedJavaAgentRaw []byte
-
-// Aliases used for testing.
-var (
-	embeddedJavaAgentBytes = embeddedJavaAgentRaw
-	userCacheDir           = os.UserCacheDir
-	renameFile             = os.Rename
-)
+var embeddedJavaAgentBytes []byte
 
 type JavaInjectError struct {
 	Message string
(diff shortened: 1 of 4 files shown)


marctc (approved):

pkg/internal/java/java_inject.go
```diff
@@ -53,25 +45,19 @@ func (e *JavaInjectError) Error() string {
...
 func NewJavaInjector(cfg *obi.Config) (*JavaInjector, error) {
 	if !cfg.Java.Enabled {
 		return nil, nil
 	}
-
-	agentPath, err := ensureEmbeddedAgentInCache()
-	if err != nil {
-		return nil, fmt.Errorf("unable to extract embedded OBI java agent jar: %w", err)
-	}
+	ensureEmbeddedAgent()
```
pellared, line 56: Why are we panicking instead of returning an error?
I think it would be better to have something like

```go
	if err := validateEmbeddedAgent(); err != nil {
		return nil, err
	}
```

Note that otherwise `NewJavaInjector` never returns an error.

Also from https://go.dev/doc/effective_go#panic

> real library functions should avoid panic. If the problem can be masked or worked around, it's always better to let things continue to run rather than taking down the whole program.

pkg/internal/java/java_inject.go
```diff
@@ -211,92 +197,10 @@ func (i *JavaInjector) NewExecutable(ie *ebpf.Instrumentable) error {
...
-		// checksum-addressed file between our initial stat and rename.
-		// If the target is now present and valid, treat this as success.
-		if info, statErr := os.Stat(targetPath); statErr == nil {
-			if !info.IsDir() && info.Size() == int64(len(embeddedJavaAgentBytes)) {
-				return targetPath, nil
-			}
-		}
-
-		return "", fmt.Errorf("unable to move java agent into cache: %w", err)
+		panic("embedded OBI java agent artifact is missing; run `make java-docker-build`")
```
pellared: Maybe something like

```suggestion
		// Make sure to run `make java-docker-build` to build the Java Agent
		// so that it can be embedded during build.
		return errors.New("embedded OBI java agent artifact is missing") 
```

pkg/internal/java/java_inject.go
mariomac: This error should be never returned to the final user, as it only happens if OBI is wrongly compiled (unrecoverable error). I think in this case is better to panic for the earliest notice of the unrecoverable error.

I'll remove the error from the `NewJavaInjector` signature.

pkg/internal/java/java_inject.go
```diff
@@ -211,92 +197,12 @@ func (i *JavaInjector) NewExecutable(ie *ebpf.Instrumentable) error {
...
-	if err != nil {
-		return "", fmt.Errorf("unable to resolve user cache directory: %w", err)
-	}
-
-	cacheDir := filepath.Join(cacheRoot, "obi", "java")
-	if err := os.MkdirAll(cacheDir, 0o755); err != nil {
-		return "", fmt.Errorf("unable to create java agent cache directory: %w", err)
+		// Make sure to run `make java-docker-build` to build the Java Agent
+		// so that it can be embedded during build.
+		return errors.New("embedded OBI java agent artifact is missing")
```
pellared:
```suggestion
		panic("embedded OBI java agent artifact is missing") 
```

if you want the build to pass

grcevski (approved):
LGTM!

pkg/internal/java/java_inject.go
MrAlias: This was an incorrect assumption. This panic does indeed get raised for end users. If they are running malformed binary or incompatible java agent this introduces a crash to their system and is a regression in behavior. We should not do this. I am submitting a fix.

PR #1898, opened 2026-04-22
fix(otel): clean up missing metric instances on teardown

MrAlias:
## Summary
This PR fixes missing OTEL metric cleanup on `Metrics` teardown and strengthens the regression coverage around that cleanup path.

## Problem
`(*Metrics).cleanupAllMetricsInstances()` did not clean up all metric expirer instances created for a `Metrics` set.

The original issue in this PR was that `genAIOutputTokenUsage` was skipped. While broadening the regression coverage, the test also exposed that `gpuGraphCallsTotal` was missing from teardown cleanup as well.

That left stale metric attribute sets behind after teardown for those instruments.

## Changes
- add the missing cleanup call for `genAIOutputTokenUsage`
- add the missing cleanup call for `gpuGraphCallsTotal`
- factor the metric-expirer setup used by `newMetricSet()` into a shared helper
- replace the narrow regression test with one that builds metrics through the shared setup path, seeds every created expirer, and verifies teardown clears them all

## Testing
- `go test ./pkg/export/otel -run TestCleanupAllMetricsInstances_RemovesAllMetrics -count=1`

### DIFF
--- a/pkg/export/otel/metrics_cleanup_test.go
+++ b/pkg/export/otel/metrics_cleanup_test.go
@@ -5,95 +5,213 @@ package otel
 
 import (
 	"context"
-	"sync"
+	"reflect"
+	"strings"
 	"testing"
 	"time"
+	"unsafe"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric/embedded"
 
 	"go.opentelemetry.io/obi/pkg/appolly/app/request"
+	"go.opentelemetry.io/obi/pkg/export"
 	"go.opentelemetry.io/obi/pkg/export/attributes"
+	attr "go.opentelemetry.io/obi/pkg/export/attributes/names"
+	"go.opentelemetry.io/obi/pkg/export/instrumentations"
 	instrument "go.opentelemetry.io/obi/pkg/export/otel/metric/api/metric"
+	"go.opentelemetry.io/obi/pkg/export/otel/otelcfg"
+	"go.opentelemetry.io/obi/pkg/export/otel/perapp"
 )
 
-func TestCleanupAllMetricsInstances_RemovesGenAIOutputTokenUsage(t *testing.T) {
+func TestCleanupAllMetricsInstances_RemovesAllMetrics(t *testing.T) {
 	t.Parallel()
 
 	ctx := t.Context()
-	inputMetric := newTestFloat64Histogram()
-	outputMetric := newTestFloat64Histogram()
+	metrics := &Metrics{ctx: ctx}
+	meter := &testMeter{}
+	reporter := newCleanupTestMetricsReporter(ctx)
 
-	metrics := &Metrics{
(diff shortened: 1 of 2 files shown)


pkg/export/otel/metrics_cleanup_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric/embedded"
+
+	"go.opentelemetry.io/obi/pkg/appolly/app/request"
+	"go.opentelemetry.io/obi/pkg/export/attributes"
+	instrument "go.opentelemetry.io/obi/pkg/export/otel/metric/api/metric"
+)
+
+func TestCleanupAllMetricsInstances_RemovesGenAIOutputTokenUsage(t *testing.T) {
```
marctc: I guess the intention of this test is to prove that removal of GenAI metrics works, right? I wonder if this test could fail when we forgot to add some of the metric instances to cleanupAllMetricsInstances

mmat11 (approved):
lgtm

pkg/export/otel/metrics_cleanup_test.go
MrAlias: I can take a look at building this out to fail when any registered metric instances is missed. It is going to increase the scope of the PR though.

MrAlias: I have a test I am refining. Looks like a good change given `gpuGraphCallsTotal` was also not cleaned up 😄

MrAlias: I updated this to avoid mirroring the `Metrics` field list in the test. The cleanup test now builds the metric expirer set through the same production setup path used by `newMetricSet()`, then seeds every created expirer reflectively and verifies each one is empty after `cleanupAllMetricsInstances()`. That means a newly added expirer created by setup should now be covered automatically, and this also exposed a missing cleanup for `gpuGraphCallsTotal`, which is fixed in this PR.

pkg/export/otel/metrics_cleanup_test.go
```diff
@@ -0,0 +1,217 @@
...
+func newCleanupTestMetricsReporter(ctx context.Context) *MetricsReporter {
+	return &MetricsReporter{
+		ctx: ctx,
+		cfg: &otelcfg.MetricsConfig{
+			TTL: time.Minute,
+		},
+		jointMetricsCfg: &perapp.MetricsConfig{
+			Features: export.FeatureApplicationRED | export.FeatureSpanOTel | export.FeatureSpanSizes,
+		},
+		is: instrumentations.NewInstrumentationSelection([]instrumentations.Instrumentation{
```
marctc, line 150: Can we use InstrumentationAll instead? i'm afraid we will forget to upgrade this list, or its intentended that all instrumentations have to be explicit?

MrAlias: Updated: 59e3db31

marctc (approved):
LGTM! thanks for this!

skl (approved):

PR #1909, opened 2026-04-23
fix(netolly): honor configured port guesser for protocol inference

MrAlias:
## Summary
- make `network.protocol` use the configured server port guesser instead of always using ordinal guessing
- add regression coverage for the disabled guess mode and the inferred protocol value

## Why
The netolly attribute getter for `network.protocol` ignored the configured port guess strategy and always used ordinal guessing to choose the server port. That made `guess_ports: false` ineffective for protocol inference even though the rest of the port-based getters respected the setting.

## Impact
This keeps protocol inference consistent with the configured port guessing behavior and avoids deriving application protocol names from a guessing mode the operator explicitly disabled.

## Validation
- `go test ./pkg/internal/netolly/ebpf`


mariomac (approved):

grcevski (approved):
LGTM!

PR #1914, opened 2026-04-24
[OTEP 4719] Add processcontext enrichement

florianl:
Implement OTEP 4719 to enable sharing of process context information with external readers. This allows services to export resource attributes and metadata via the OTEL_CTX environment mapping without direct instrumentation.

The ProcessContextDecorator enriches discovered processes by:
- Locating the OTEL_CTX mapping in /proc/<pid>/maps
- Reading process context information from remote memory
- Merging resource attributes into service metadata

This is a proof-of-concept with known limitations:
- cilium/ebpf upgraded to v0.21.0 to resolve dependency conflicts
- Process context proto definitions pending merge: https://github.com/open-telemetry/opentelemetry-proto/pull/783
- No SDK implementations of OTEP 4719 yet available

For testing, use the reference implementation at:
https://github.com/open-telemetry/sig-profiling/tree/main/process-context/c-and-cpp

See: https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/profiles/4719-process-ctx.md

## Summary

Describe the change briefly.

## Validation


NimrodAvni78:
This is a great idea!
I you think this can help us (once sdks start implementing this) with detecting if a service is instrumented to avoid double instrumentation (we currently rely on detecting http/grpc exporting telemetry)
can we know from this some information like which instrumentations are registered? or only if a process is instrumented or not and its resource attributes

florianl:
> Do you think this can help us with detecting if a service is instrumented to avoid double instrumentation

Yes - I think this will be one benefit.

> can we know from this some information like which instrumentations are registered? or only if a process is instrumented or not and its resource attributes

At the moment, if this information is available we just know, that the process was instrumented. But I think, at some point it will be possible to share also information about the instrumentation scope.

pkg/appolly/discover/processcontext_decorator.go
```diff
@@ -0,0 +1,166 @@
...
+	out *msg.Queue[[]Event[ebpf.Instrumentable]]
+	log *slog.Logger
+}
+
+func (pcd *processContextDecorator) decorate(ctx context.Context) {
+	defer pcd.out.Close()
+	swarms.ForEachInput(ctx, pcd.in, pcd.log.Debug, func(instrumentables []Event[ebpf.Instrumentable]) {
+		for i := range instrumentables {
+			ev := &instrumentables[i]
+			if ev.Type == EventCreated {
```
mmat11: this is triggered on process creation, but the process context should be present only when the SDK registers it -- if the SDK registers it with some delay we never see it, right?

should we have an initial delay instead, to give some time to the SDK to register itself?

pkg/appolly/discover/processcontext_decorator.go
```diff
@@ -0,0 +1,166 @@
...
+				pcd.enrichEvent(ev)
+			}
+		}
+		pcd.out.SendCtx(ctx, instrumentables)
+	})
+}
+
+func (pcd *processContextDecorator) enrichEvent(ev *Event[ebpf.Instrumentable]) {
```
mmat11: this is fired one-shot, but, correct me if I am wrong, I _think_ the OTEP intended readers to watch for changes to the context mapping (https://github.com/open-telemetry/opentelemetry-specification/pull/4719/changes#diff-c02a091b4a954164a90867c19771116b26d717419c16b715bce8b56d80ba49edR192) via polling or another mechanism, then:

- process the new values if `monotonic_published_at_ns` is non-zero
- wait/do nothing if `monotonic_published_at_ns` is zero

mmat11:
it would be nice to have some tests with some SDK that implements the OTEP already (or some custom binary if that's not available)

pkg/appolly/discover/processcontext_decorator.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1914/commits/e65b953678bbb16b84cc0b745572a01526bd2c17

pkg/appolly/discover/processcontext_decorator.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1914/commits/e65b953678bbb16b84cc0b745572a01526bd2c17

pkg/appolly/discover/processcontext_decorator_test.go
```diff
@@ -0,0 +1,327 @@
...
+	*(*uint32)(unsafe.Pointer(&buf[8])) = otelCtxVersion
+	*(*uint32)(unsafe.Pointer(&buf[12])) = payloadSize
+	*(*uint64)(unsafe.Pointer(&buf[16])) = publishedAt
+	*(*uint64)(unsafe.Pointer(&buf[24])) = payloadPtr
+}
+
+// setupOTELCTXMapping creates a memfd named "OTEL_CTX", writes a valid header
+// pointing at payload, and maps it into the current process.  The caller must
+// call the returned cleanup function when done.
+func setupOTELCTXMapping(t *testing.T, ctx *processcontextpb.ProcessContext) (cleanup func()) {
```
florianl, line 55: Once https://github.com/open-telemetry/opentelemetry-go/issues/8730 lands, this custom code can be dropped and removed with the upstream OTel SDK version.

pkg/appolly/discover/processcontext_decorator.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+}
+
+// handleCreated attempts an immediate enrichment on process creation and registers
+// the process for ongoing polling to handle SDK startup delays and context updates.
+func (pcd *processContextDecorator) handleCreated(ev *Event[ebpf.Instrumentable]) {
+	pid := ev.Obj.FileInfo.Pid()
+	entry := &processEntry{fi: ev.Obj.FileInfo}
+
+	if addr, ok := pcd.findOTELContextMapping(pid); ok {
```
mmat11: maybe we can make this block work both with first scan and updates and extract it into a function so we don't have to repeat it in poll(), WDYT?

pkg/export/otel/metrics_runtime_histograms_test.go
```diff
@@ -14,6 +14,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
```
mmat11, line 17: nit: unneeded whiteline

pkg/appolly/services/criteria.go
```diff
@@ -123,6 +123,12 @@ type DiscoveryConfig struct {
 	// Min process age to be considered for discovery.
 	MinProcessAge time.Duration `yaml:"min_process_age" env:"OTEL_EBPF_MIN_PROCESS_AGE" validate:"gte=0"`
 
+	// ProcessContextPollInterval controls how often OBI re-reads the OTEL_CTX
+	// mapping of each discovered process.  Polling handles both SDKs that
+	// publish the mapping after startup and context updates published later.
+	// 0 is treated as a default (1s).
+	ProcessContextPollInterval time.Duration `yaml:"process_context_poll_interval" env:"OTEL_EBPF_PROCESS_CONTEXT_POLL_INTERVAL" validate:"gte=0"`
```
mmat11, line 130: is there a way to completely disable it?

mmat11 (approved):
left a couple nits, lgtm!

I would also leave a couple of lines somewhere in devdocs so it's easier to understand what the config option is for

pkg/export/otel/metrics_runtime_histograms_test.go
florianl: `make fmt` makes sure this whiteline is kept :|

pkg/appolly/discover/processcontext_decorator.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1914/commits/4e22407625372563aeda74d878a9d8da6ec60f5f

pkg/appolly/services/criteria.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1914/commits/4e22407625372563aeda74d878a9d8da6ec60f5f

pkg/export/otel/metrics_runtime_histograms_test.go
mmat11: oh ok, nvm then

grcevski (approved):
LGTM!

PR #1921, opened 2026-04-24
Add support for AMQP 1.0 protocol

### LINKED ISSUE #1142: Instrument and collect telemetry for AMQP

marctc:
## Summary

This PR adds support to capture events from AMQP protocol v1.0

Fixes #1142 

## Validation


marctc:
Dear reviewers, I know the PR is big and scary, but important places to review are:

- pkg/ebpf/common/amqp_detect_transform.go
- pkg/ebpf/common/tcp_detect_transform.go
- pkg/internal/ebpf/amqpparser/frame.go
- pkg/internal/ebpf/amqpparser/parser.go
- pkg/internal/ebpf/amqpparser/protocol.go

pkg/ebpf/common/amqp_detect_transform.go
```diff
@@ -0,0 +1,138 @@
...
+		End:           int64(trace.EndMonotimeNs),
+		Status:        0,
+		TraceID:       trace.Tp.TraceId,
+		SpanID:        trace.Tp.SpanId,
+		ParentSpanID:  trace.Tp.ParentId,
+		TraceFlags:    trace.Tp.Flags,
+		Pid: request.PidInfo{
+			HostPID:   app.PID(trace.Pid.HostPid),
+			UserPID:   app.PID(trace.Pid.UserPid),
+			Namespace: trace.Pid.Ns,
```
rafaelroquetto, line 93: _(From Claude)_ - makes sense IMHO

`swapConnInfo` and `oppositeDirection` duplicate logic that already exists in `reverseTCPEvent` ([tcp_detect_transform.go:430-443](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/99c46a7a770d676947be40fe972f080632714bed/pkg/ebpf/common/tcp_detect_transform.go#L430-L443)).

`reverseTCPEvent` already swaps S/D addresses and ports in-place and flips the direction bit. `swapConnInfo` is the value-copy variant of the same swap, and `oppositeDirection` is the direction-flip extracted as a standalone function.

`amqpSpanEndpoints` could take a local copy of `trace.ConnInfo` and mutate it directly (same pattern used elsewhere), eliminating `swapConnInfo`. And if `reverseTCPEvent` extracted its direction flip into a shared one-liner, `oppositeDirection` would not be needed at all.

Per the repo guidelines: *"Do not introduce new implementations when equivalent functionality already exists."*

pkg/ebpf/common/tcp_detect_transform.go
```diff
@@ -360,6 +363,42 @@ func matchNATS(parseCtx *EBPFParseContext, event *TCPRequestInfo, requestBuffer,
 	return TCPToNATSToSpan(event, info), false, true, nil
 }
 
+func matchAMQP(parseCtx *EBPFParseContext, event *TCPRequestInfo, requestBuffer, responseBuffer *largebuf.LargeBuffer) (request.Span, bool, bool, error) { //nolint:unparam
+	reqLooks, reqErr := isAMQP(requestBuffer)
```
rafaelroquetto: in addition to what claude said below, we should probably change "parser.go" to use the large buffer reader  API instead of UnsafeView() - this will also mitigate double parsing the buffer.

For example, something like:

```
func parseFrameHeader(r *largebuf.LargeBufferReader) (frameHeader, error) {
    size, err := r.ReadU32BE()
    // ...
    doff, err := r.ReadU8()
    // ...
    ft, err := r.ReadU8()
    // ...
    _, err = r.Skip(2) // channel, unused
    // ...
}
```

--- _(Claude)_  ---

`matchAMQP` double-parses each buffer. `isAMQP(requestBuffer)` and `isAMQP(responseBuffer)` each call `amqpparser.Parse`. Then `ProcessPossibleAMQPEvent` is called, which internally calls `processAMQPBuffer` twice — each of which calls `Parse` again. Every event therefore parses each buffer twice and the first `Result` is discarded.

The `isAMQP` calls here are redundant: `processAMQPBuffer` already decides whether a buffer looks like AMQP and whether it has transfers. `matchAMQP` could work directly from the `(bool, []AMQPInfo, error)` triple returned by `ProcessPossibleAMQPEvent` and skip the preliminary check entirely.

pkg/ebpf/common/tcp_detect_transform.go
```diff
@@ -360,6 +363,42 @@ func matchNATS(parseCtx *EBPFParseContext, event *TCPRequestInfo, requestBuffer,
...
+	respLooks, respErr := isAMQP(responseBuffer)
+	if !reqLooks && !respLooks {
+		return request.Span{}, false, false, nil
+	}
+
```
rafaelroquetto: `slog.Warn` is inconsistent with the rest of the heuristic detection layer. The equivalent MQTT failure path (tcp_detect_transform.go:384) uses `slog.Debug`. A parse failure on AMQP-shaped traffic is a normal operational event — it means the heuristic fired on something that turned out not to be valid AMQP — not a condition the operator needs to act on. Use `slog.Debug` here (and at the second `slog.Warn` a few lines below) to match the convention already established by MQTT.

pkg/appolly/app/request/span_getters.go
```diff
@@ -217,6 +219,9 @@ func spanOTELGetters(name attr.Name) (attributes.Getter[*Span, attribute.KeyValu
 			if span.Type == EventTypeNATSClient || span.Type == EventTypeNATSServer {
 				return semconv.MessagingDestinationName(span.Path)
 			}
+			if span.Type == EventTypeAMQPClient {
+				return semconv.MessagingDestinationName(span.Path)
```
rafaelroquetto, line 220: _Claude_
Separate `if` block is unnecessary — this should extend the NATS condition immediately above it:

```go
if span.Type == EventTypeNATSClient || span.Type == EventTypeNATSServer || span.Type == EventTypeAMQPClient {
    return semconv.MessagingDestinationName(span.Path)
}
```

All other messaging types in this function use a single condition per attribute, not one `if` per type. The current form duplicates the return value and diverges from the established pattern.

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -826,6 +829,18 @@ func TraceAttributesSelector(span *request.Span, optionalAttrs map[attr.Name]str
...
 			attrs = append(attrs, request.PeerService(request.PeerServiceFromSpan(span)))
 		}
+	case request.EventTypeAMQPClient:
+		operation := request.MessagingOperationType(span.Method)
+		attrs = []attribute.KeyValue{
+			request.ServerAddr(request.HostAsServer(span)),
+			request.ServerPort(span.HostPort),
+			messagingSystemAMQP,
+			semconv.MessagingDestinationName(span.Path),
+			semconv.MessagingClientID(span.Statement),
```
rafaelroquetto: _Claude_
`span.Statement` is never populated by the AMQP parser — the limitation is explicitly documented: *"No destination extraction: link target/source and other addressing details inside the TRANSFER performative are not decoded."*

`MessagingClientID(span.Statement)` will therefore always emit an empty `messaging.client.id` attribute. The NATS and Kafka cases at lines 789 and 810 populate `Statement` in at least some paths; AMQP does not. Either guard this with `if span.Statement != ""` or omit it until the parser can actually extract a client ID.

pkg/ebpf/common/amqp_detect_transform_test.go
```diff
@@ -0,0 +1,322 @@
...
+			StartMonotimeNs: 100,
+			EndMonotimeNs:   200,
+			Direction:       directionSend,
+			ConnInfo: BpfConnectionInfoT{
+				S_port: 54321,
+				D_port: 5672,
+			},
+		}
+
+		span := tcpToAMQPToSpan(trace, AMQPInfo{Direction: directionRecv})
```
rafaelroquetto: _Claude_
The "process uses broker as server" test sets `trace.Direction = directionSend` and `info.Direction = directionRecv` — which exercises the swap branch inside `amqpSpanEndpoints` (`trace.Direction != direction`). However, it does not assert the resulting `PeerPort` or `HostPort` values, so the swap logic is not actually verified.

```go
assert.Equal(t, 54321, span.PeerPort)  // ephemeral port = peer
assert.Equal(t, 5672,  span.HostPort)  // broker port = host
```

Without those assertions, a swap inversion (e.g. ports exchanged) would go undetected.

rafaelroquetto (commented):
**EDIT: this was Claude accidentally posting comments automagically when I was using it to review the PR - I didn't have a chance to curate them, so if you see any edits, this is me reviewing it.** 

Overall this is a solid addition — the parser package is well-layered (`protocol.go` / `frame.go` / `parser.go`), the test coverage of adversarial inputs and truncation paths is thorough, and the integration into the span/metrics pipeline follows the established messaging pattern correctly.

Left inline comments on the issues that need attention before merge. In order of priority:

1. **`swapConnInfo` / `oppositeDirection` duplicate `reverseTCPEvent`** — existing function, no new code needed.
2. **`matchAMQP` double-parses each buffer** — `isAMQP` result is immediately discarded when `ProcessPossibleAMQPEvent` re-parses the same data.
3. **`slog.Warn` for heuristic parse failures** — should be `slog.Debug` to match MQTT.
4. **`span_getters.go` separate `if` for AMQP destination** — should extend the NATS condition.
5. **`MessagingClientID(span.Statement)` always emits empty** — guard or remove until the parser populates it.
6. **`TestTCPToAMQPSpan` "process" case missing port assertions** — the swap path is exercised but not verified.

rafaelroquetto (commented):
I think the biggest blocker is the `UnsafeView()` -> pivoting to the large buffers reader API. Otherwise, Claude and Copilot left a few comments and LGTM

rafaelroquetto (approved):
Groovy!

grcevski (approved):
LGTM!

pkg/ebpf/common/amqp_detect_transform.go
```diff
@@ -0,0 +1,115 @@
...
+	reqLooks, reqInfos, reqErr := processAMQPBuffer(pkt, event.Direction)
+	respLooks, respInfos, respErr := processAMQPBuffer(rpkt, reverseDirection(event.Direction))
+
+	infos := reqInfos
+	infos = append(infos, respInfos...)
+	if len(infos) > 0 {
+		return infos, false, nil
+	}
+
+	if reqErr != nil && respErr != nil {
```
mmat11, line 29: should we check for errors first and then return infos if available?

pkg/ebpf/common/amqp_detect_transform.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+		return result.LooksLikeAMQP, nil, err
+	}
+	if !result.LooksLikeAMQP {
+		return false, nil, nil
+	}
+
+	infos := make([]AMQPInfo, 0, result.TransferCount)
+	for i := 0; i < result.TransferCount; i++ {
+		infos = append(infos, AMQPInfo{Direction: direction})
```
mmat11, line 65: it's not clear to me why we create multiple equal spans with different direction?

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -826,6 +829,20 @@ func TraceAttributesSelector(span *request.Span, optionalAttrs map[attr.Name]str
 		if span.Type == request.EventTypeNATSClient {
 			attrs = append(attrs, request.PeerService(request.PeerServiceFromSpan(span)))
 		}
+	case request.EventTypeAMQPClient:
+		operation := request.MessagingOperationType(span.Method)
+		attrs = []attribute.KeyValue{
+			request.ServerAddr(request.HostAsServer(span)),
+			request.ServerPort(span.HostPort),
+			messagingSystemAMQP,
+			semconv.MessagingDestinationName(span.Path),
```
mmat11: span.Path is never filled in amqp_detect_transform

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -826,6 +829,20 @@ func TraceAttributesSelector(span *request.Span, optionalAttrs map[attr.Name]str
...
+			operation,
+		}
+		if span.Statement != "" {
+			attrs = append(attrs, semconv.MessagingClientID(span.Statement))
```
mmat11: same here, span.Statement is never filled

pkg/ebpf/common/amqp_detect_transform.go
marctc: If one side produced a valid transfer, we should emit it even if the other buffer failed to parse. We only surface the parser error when neither side gives us a usable transfer.

pkg/ebpf/common/amqp_detect_transform.go
marctc: added comment

pkg/ebpf/common/amqp_detect_transform.go
mmat11: thanks!

mmat11 (approved):
lgtm!

PR #1929, opened 2026-04-24
fix(testdata): remove committed Rails credentials

MrAlias:
## Summary
- remove the committed Rails credentials bundle from the nginx-rails-sql integration test component
- remove the committed Rails master key from the same test fixture

## Why
The integration test fixture included Rails credential material that should not live in the repository. In the current fixture, those files are also not used by the test stack anymore: the compose setup does not provide `RAILS_MASTER_KEY`, Rails production config does not enable `config.require_master_key`, and the remaining references to `Rails.application.credentials` in `storage.yml` are commented examples rather than active configuration.

## Impact
This removes unused committed credential files from version control without changing the current runtime behavior of the nginx-rails-sql integration test fixture.

## Validation
- verified the change set only deletes the committed credential files
- verified the fixture does not actively use Rails credentials or require a master key in its current configuration


grcevski (approved):
LGTM!

NimrodAvni78 (approved):

PR #1931, opened 2026-04-24
Java Agent multi architecture build

grcevski:
## Summary

We currently embed the java agent in the OBI binary, but the agent is platform dependent, amd64 and aarch64. We get away with this because we build OBI from scratch, including the java agent along with OBI with `docker buildx`, however this will not work for the collector. They'll simply import the dependency and it will be whatever platform the committer updated the agent last.

I'm changing the build to perform compilation for both platforms and the java agent now chooses what .so file to load on start.

TODO:
- [x] More extensive tests on aarch64.

## Validation


MrAlias:
This makes the default local Java build path depend on producing both native JNI libraries, not just the host one.

I agree the shipped embedded JAR needs to stay multi-arch, especially since downstream consumers pick up the source-generated release artifact rather than building from a full Git checkout. My concern is narrower: wiring both native builds into `shadowJar` means ordinary local flows like `gradle build`, `make java-build`, and `make java-test` now also require the non-native cross-toolchain.

Is it possible to separate those concerns in the task graph? For example, keep the multi-arch packaging on the path that produces the shipped embedded JAR, but leave the default local Gradle build/test path host-only, or gate the multi-arch packaging behind a dedicated task/property used by Docker/release. That would preserve the release artifact requirement without broadening the default local toolchain requirements.

grcevski:
> Is it possible to separate those concerns in the task graph? For example, keep the multi-arch packaging on the path that produces the shipped embedded JAR, but leave the default local Gradle build/test path host-only, or gate the multi-arch packaging behind a dedicated task/property used by Docker/release. That would preserve the release artifact requirement without broadening the default local toolchain requirements.

Yeah I think I can do this, good idea.

grcevski:
Hey @MrAlias, I added an option to the gradle and I made sure gradle can natively build aarch64 just like the docker files. By default `make java-build` now builds a single architecture, the native only, no cross-compilation.

mariomac (approved):
I haven't locally tested it but the code LGTM

MrAlias (approved):
Thanks for addressing this. The latest change does fix the `make java-build` path by making it host-only by default.

I think there’s still a smaller follow-up remaining, though: `make java-test` and plain `gradle test` / `gradle build` still appear to go through the default `shadowJar` path without `-PnativeOnly=true`, and the root Java test task still depends on `copyLoaderJar` -> `:agent:shadowJar`.

I don’t think that needs to block this PR, since the release/source-consumer path is the more important constraint here and that still looks correct. We can follow-up to make the default local test/build flows host-only end-to-end as well.

grcevski:
> I don’t think that needs to block this PR, since the release/source-consumer path is the more important constraint here and that still looks correct. We can follow-up to make the default local test/build flows host-only end-to-end as well.

Good point, I'll follow-up with a fix for those.

PR #1934, opened 2026-04-26
fix(k8s_cache): informer_send_timeout is not enforced

### LINKED ISSUE #1903: k8s_cache: informer_send_timeout is not enforced
marked in [here](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1886#discussion_r3119977646)

informer_send_timeout is stored on the connection connection, but is never used

itsprdp:
## Summary
The `kubecache` service streams Kubernetes object updates to subscribers over gRPC. Each subscriber connection has a configurable `informer_send_timeout`, the maximum time allowed for a single `Send()` call before the connection is considered stalled and dropped.

**bug**: `informer_send_timeout` was stored on the connection struct but never actually used. `Send()` was called directly with no timeout, so a slow or unresponsive client could block the `goroutine` forever, preventing other events from being delivered.

## PR includes:
1. Adds a regression test - a `blockingStream` fake whose `Send()` blocks indefinitely is wired up, and the test asserts that `handleMessagesQueue` returns within `200ms` when the timeout is `50ms`.
2. Fixes the bug - introduces `sendWithTimeout()`, which:
    - If no timeout is configured, calls `Send()` directly (same behavior as before).
    - If a timeout is configured, runs `Send()` in a `goroutine` and races it against a `time.Timer`. Whichever fires first wins: if the timer  fires, the connection is logged, metered, and dropped.
      The timer is created once outside the loop and reset on each iteration to avoid allocating a new timer per message.

## Validation

Fixes #1903 

## AI
This contribution was prepared with the assistance of an AI development tool.


NimrodAvni78 (approved):
LGTM

pkg/kube/kubecache/service/service.go
```diff
@@ -147,6 +147,11 @@ func (o *connection) On(event *informer.Event) error {
 }
 
 func (o *connection) handleMessagesQueue(ctx context.Context) {
+	var timer *time.Timer
+	if o.sendTimeout > 0 {
```
mmat11: should we have a default timeout here if sendTimeout is unset?

pkg/kube/kubecache/service/service.go
```diff
@@ -157,12 +162,52 @@ func (o *connection) handleMessagesQueue(ctx context.Context) {
 			if event == nil {
 				return
 			}
-			if err := o.server.Send(event); err != nil {
-				o.log.Debug("Error sending message. Closing client connection", "clientID", o.ID(), "error", err)
-				o.metrics.MessageError()
+			if err := o.sendWithTimeout(ctx, timer, event); err != nil {
 				return
 			}
 			o.metrics.MessageSucceed()
```
mmat11: I think we should update this metric inside sendWithTimeout, for consistency

pkg/kube/kubecache/service/service.go
```diff
@@ -157,12 +162,52 @@ func (o *connection) handleMessagesQueue(ctx context.Context) {
...
 		}
 	}
 }
+
+// sendWithTimeout sends event and, if timer is non-nil, drops the connection
+// when Send blocks longer than o.sendTimeout. Returns a non-nil error whenever
+// the caller should stop processing the queue.
+func (o *connection) sendWithTimeout(ctx context.Context, timer *time.Timer, event *informer.Event) error {
+	if timer == nil {
```
mmat11: if we set a default timeout, I think we can avoid this check as timer will never be nil

pkg/kube/kubecache/service/service.go
mmat11, line 165: is the sendTimeout supposed to be per-`Send()` or per-batch?

pkg/kube/kubecache/service/service.go
itsprdp: good idea! I'll initialize the default timeout value for connection itself if it's unset so that we can skip the check here

pkg/kube/kubecache/service/service.go
itsprdp: It is per-`Send()` and each event would get the full `SendTimeout` window. I'll update the comments to reflect that.

pkg/kube/kubecache/service/service.go
itsprdp: addressed this in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1934/commits/f900b5e89379a7da6b60fc5c79d5bc8cf9ae67a9

pkg/kube/kubecache/service/service.go
itsprdp: addressed this in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1934/commits/f900b5e89379a7da6b60fc5c79d5bc8cf9ae67a9

pkg/kube/kubecache/service/service.go
itsprdp: addressed this in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1934/commits/f900b5e89379a7da6b60fc5c79d5bc8cf9ae67a9

pkg/kube/kubecache/service/service.go
```diff
@@ -157,12 +162,52 @@ func (o *connection) handleMessagesQueue(ctx context.Context) {
...
+			o.log.Debug("Error sending message. Closing client connection", "clientID", o.ID(), "error", err)
+			o.metrics.MessageError()
+			return err
+		}
+		return nil
+	}
+
+	sendErr := make(chan error, 1)
+	go func() { sendErr <- o.server.Send(event) }()
+
```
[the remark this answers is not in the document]
itsprdp: `Send()` doesn't take a context so we can't cancel it. The goroutine exits when gRPC closes the stream after `Subscribe()` returns. `sendErr` is a buffered `chan` so it won't block. Added a comment in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1934/commits/9312b28ccfea2e825cc5d7068a44af6ea297858a

pkg/kube/kubecache/service/service.go
```diff
@@ -157,12 +162,52 @@ func (o *connection) handleMessagesQueue(ctx context.Context) {
...
+		if err != nil {
+			o.log.Debug("Error sending message. Closing client connection", "clientID", o.ID(), "error", err)
+			o.metrics.MessageError()
+			return err
+		}
+		return nil
+	case <-timer.C:
+		o.log.Warn("Send timed out. Closing client connection", "clientID", o.ID(), "timeout", o.sendTimeout)
+		o.metrics.MessageTimeout()
+		return context.DeadlineExceeded
```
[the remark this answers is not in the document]
itsprdp: same as https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1934#discussion_r3151795154

mmat11 (approved):
lgtm

pkg/kube/kubecache/service/service.go
```diff
@@ -157,12 +166,44 @@ func (o *connection) handleMessagesQueue(ctx context.Context) {
...
+		case <-timer.C:
+		default:
+		}
+	}
+	timer.Reset(o.sendTimeout)
+
+	select {
+	case err := <-sendErr:
+		if err != nil {
+			o.log.Debug("Error sending message. Closing client connection", "clientID", o.ID(), "error", err)
```
mmat11: just one last thing: should this be a warn instead?

pkg/kube/kubecache/service/service_test.go
```diff
@@ -129,3 +133,162 @@ func TestRunStopsServerOnContextCancellationWithActiveStream(t *testing.T) {
...
+	}()
+
+	<-sendCalled // wait until Send is blocking before cancelling
+	cancel()
+
+	select {
+	case <-done:
+	case <-time.After(200 * time.Millisecond):
+		t.Fatal("handleMessagesQueue did not return within 200ms after context cancellation")
+	}
```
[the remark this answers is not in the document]
itsprdp: switched this to 2s instead

pkg/kube/kubecache/service/service_test.go
```diff
@@ -129,3 +133,162 @@ func TestRunStopsServerOnContextCancellationWithActiveStream(t *testing.T) {
...
+			go func() {
+				o.handleMessagesQueue(context.Background())
+				close(done)
+			}()
+
+			select {
+			case <-done:
+			case <-time.After(200 * time.Millisecond):
+				t.Fatalf("handleMessagesQueue did not return within 200ms")
+			}
```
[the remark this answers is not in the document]
itsprdp: switched this to 2s instead

pkg/kube/kubecache/service/service.go
itsprdp: switched it to warn in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1934/commits/0e6ef54cd8b196ef533670922a1946cc54ceb1cc

grcevski (approved):
LGTM!

PR #1939, opened 2026-04-27
Fix overriding of cloud_host_id in prometheus exporter

mariomac:
## Summary

Even if CloudHostID variable could be overridden from any software component vendoring OBI, the change did not take effect because the old value was still in the `hostInfoLabelNames` variable. 

## Validation


grcevski (approved):
LGTM!

MrAlias (commented):
Missing test coverage for the regression this PR is fixing.

The code change in `pkg/export/prom/prom.go` looks correct, but I don’t see a test that actually overrides `prom.CloudHostIDKey` and verifies that `traces_host_info` is created with the overridden label name. The existing assertion in `pkg/export/prom/prom_test.go` still hard-codes the default `cloud_host_id`, so it wouldn’t catch the package-init caching bug this PR is addressing if it regressed later.

I’d prefer adding a focused regression test that temporarily sets `CloudHostIDKey` to a non-default value, constructs the reporter/exporter, and asserts the exported metric uses that overridden label key.

mariomac:
@MrAlias good point! I'll add the tests once this PR is merged https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1942

mmat11 (approved):

PR #1941, opened 2026-04-27
feat: Go user-space HTTP server request body extraction

### LINKED ISSUE #1938: Use large buffers for HTTP events captured by the existing Go probes, so that we can enable payload extraction.

skl:
## Summary

This PR implements HTTP server request body extraction in user-space for Go programs. Response body will follow in a separate PR.

Based on `internal/test/integration/http_go_enrichment_body_test.go`, the new attributes that can be added to Go spans include:

- `http.request.header.content-type`
- `http.request.body.content`

As shown in the obfuscation example below:

```json
{
  "name": "POST /rolldice/:id",
  "kind": "SERVER",
  "attributes": {
    "http.request.method":       "POST",
    "url.path":                  "/rolldice/50",
    "http.route":                "/rolldice/:id",
    "http.response.status_code": 200,
    "server.address":            "testserver",
    "server.port":               8080,
    "client.address":            "10.0.0.5",

    "http.request.header.content-type": "application/json",
    "http.request.body.content":        "{\"username\":\"alice\",\"password\":\"***\",\"secret\":\"***\",\"email\":\"alice@test.com\"}"
  }
}
```

Fixes #1938

## Validation


pkg/ebpf/common/common.go
```diff
@@ -431,7 +431,9 @@ func ReadBPFTraceAsSpan(parseCtx *EBPFParseContext, cfg *config.EBPFTracer, reco
 		return request.Span{}, true, err
```
rafaelroquetto, line 562: Looking at this code, it just occurred to me that the `ReinterpretCast` above is unconditionally casting to `HTTPRequestTrace` without even checking `eventType`. This tangent to this PR but we should fix it, either as part of it os as a separate pre-condition PR.

pkg/ebpf/common/common.go
```diff
@@ -431,7 +431,9 @@ func ReadBPFTraceAsSpan(parseCtx *EBPFParseContext, cfg *config.EBPFTracer, reco
...
 	}
 
-	return HTTPRequestTraceToSpan(event), false, nil
+	span := HTTPRequestTraceToSpan(event)
```
rafaelroquetto: I feel the enrichment should happen inside `HTTPRequestTraceToSpan`. This would be inline with what we do in `HTTPInfoEventToSpan`. If you look there, you will see how it looks up the large buffers for the request and the response, and passes them down to the enricher. I believe we should do the same here. (Maybe a pattern emerge that lets us reuse some code, but even if not).

Then we can rename (or inline/remove it) the enrichGOHttpServerSpan. On that topic, even though it comes as a result for uprobes sending HTTPRequestTrace, there's nothing go specific to it afaik, it's just enrinching based on large buffers, so I don't think it should have GO on the name

bpf/gotracer/go_nethttp.c
```diff
@@ -499,6 +505,68 @@ int obi_uprobe_readContinuedLineSliceReturns(struct pt_regs *ctx) {
...
+    if (bpf_probe_read_user(large_buf->buf, avail, arr) != 0) {
+        return;
+    }
+    large_buf->len = avail;
+
+    u32 payload_size = avail < sizeof(void *) ? sizeof(void *) : avail;
+    bpf_clamp_umax(payload_size, k_large_buf_payload_max_size);
+    u32 total_size = sizeof(tcp_large_buffer_t) + payload_size;
+    bpf_clamp_umax(total_size, k_large_buf_max_size);
+    bpf_ringbuf_output(&events, large_buf, total_size, get_flags());
```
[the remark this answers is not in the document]
rafaelroquetto: Copilot is right here. We should consider the feasibility of moving large_bbuf_emit_chunks from `protocol_common.h` into somehwere in `common` and reuse it here.

There are a few values at play here:
 - k_large_buf_max_http_captured_bytes -> this is the maximum number of bytes we are allowed to capture from a request (64K)
 - k_large_buf_payload_max_size -> this is the maximum number of bytes a _single_ large buffer can carry as a payload (16K)

 So if your k_large_buf_max_http_captured_bytes is greater than k_large_buf_payload_max_size, you will require multiple large buffers -> which is the logic in emit_chunks.

rafaelroquetto (changes requested):
I think there are 2 issues that need to be resolved:

 - the flow into `ReadBPFTraceAsSpan` -> HTTPRequestTraceToSpan needs to be streamlined and made agnostic (no need to cite go, all it cares about is a combination of HTTPRequestTrace and large buffers)
- emitting the chunks as copilot suggests at the ebpf layer

pkg/ebpf/common/common.go
skl: Done in b375cbc1c

bpf/gotracer/go_nethttp.c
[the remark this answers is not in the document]
skl: @rafaelroquetto done in f1cd99868e549f608c535b0d001c47a087e15957

pkg/ebpf/common/common.go
skl: Done in f1cd99868e549f608c535b0d001c47a087e15957

bpf/common/large_buf_emit.h
```diff
@@ -0,0 +1,63 @@
...
+    k_large_buf_read_user = 1,
+};
+
+static __always_inline u32 large_buf_emit_chunks(tcp_large_buffer_t *large_buf,
+                                                 const void *u_buf,
+                                                 u32 available_bytes,
+                                                 enum large_buf_read_mode mode) {
+    const unsigned char *p = (const unsigned char *)u_buf;
+
+    bpf_clamp_umax(available_bytes, k_large_buf_max_http_captured_bytes);
```
rafaelroquetto: this could have been a pre-existing bug but this caught my attention here - now that this is truly "common/shared", we can't rely on a protocol specific constant.

Instead, since they are all 64K, my recommendation is to collapse all of the k_large_buf_max_XXX_captured_bytes into a single constant.

The reason for this constant is to help the verifier be bounded. The configured limits in the settings may never be greater than this constant.

pkg/ebpf/common/http_transform.go
```diff
@@ -21,6 +21,27 @@ import (
...
+	conn BpfConnectionInfoT,
+	isClient bool,
+) (*http.Request, bool) {
+	buffer, ok := extractTCPLargeBuffer(parseCtx, traceID, packetTypeRequest,
+		directionByPacketType(packetTypeRequest, isClient), conn, ProtocolTypeHTTP)
+	if !ok {
+		return nil, false
+	}
+
+	reqReader := buffer.NewReader()
```
rafaelroquetto: the epilogue of HTTPRequestToSpan does something similar - is it possible to collapse them so that this helper is used?

rafaelroquetto:
> _Note: this comment was written by **Claude** (Anthropic's coding agent), posted on behalf of @rafaelroquetto._

A few concerns about how `ship_server_request_body` reads the request bytes out of the `*bufio.Reader`. The reader pointer is cached at header-parse time (`ongoing_server_bufr`) and `buf[0:w]` is read back later in `serve_http_returns`. That gap, plus reading from offset `0`, has three consequences:

**1. The capture is bounded by the bufio buffer size (~4KB), so the 64KB limit and the chunking path are effectively unreachable here.** `blen` is `_buf_reader_w_pos`, the reader's write index, which is bounded by `len(buf)`. The reader is `conn.bufr`, which net/http creates with the default 4096-byte buffer, so `avail = min(blen, cfg_max) <= 4096` regardless of `http_max_captured_bytes`. As a result `large_buf_emit_chunks` only ever emits a single sub-16KB chunk on this path — the multi-chunk loop added by the `large_buf_emit_chunks` generalisation is never actually exercised by the Go path it was added for.

**2. It reads from `buf[0]` instead of the reader's `r` offset.** The raw request lives at `buf[r:w]`, but we capture `buf[0:w]`. This happens to work when the request starts at offset 0 (fresh connection, or non-pipelined keep-alive where `fill()` slides `r` back to 0). With HTTP/1.1 pipelining, though, request 2's bytes are already buffered when request 1 finishes (`r > 0`), so `buf[0:w]` is `<stale request 1> ++ <request 2>` and `http.ReadRequest` parses request 1 — enriching the span with the *wrong* request's data. Because that parse succeeds, the error fallback doesn't catch it.

**3. The entry→return gap lets a body read slide the buffer.** If the handler reads a body large enough to trigger a `fill()` while `r > 0`, bufio slides (`copy(buf, buf[r:w])`) and overwrites the request line/headers. At return, `buf[0:w]` no longer starts with a valid request line, so `http.ReadRequest` fails and enrichment is silently skipped — i.e. precisely the larger requests this is meant for tend not to be captured even when they'd fit the configured limit.

The graceful fallback (parse error → skip enrichment) handles (3) and most of (2) cleanly by degrading to "no enrichment." The exception is the pipelined case in (2), which is silently *wrong* rather than silently *absent*. And (1) is the structural ceiling: as wired, this is best-effort capture of the `<=4KB` currently sitting in `conn.bufr`, not full body extraction. Capturing the full body would need a source that actually holds it (the body reader, or a copy taken at read time) and tracking the request's start offset rather than assuming `0`.

rafaelroquetto:
@skl thanks for the changes, I just had a minor comment regarding the HTTP constant. Then there's this big wall of text from Claude above - I think you know better about the go internals to than myself to judge whether what it is saying is sound or BS, so will leave it up to you.

bpf/common/large_buf_emit.h
```diff
@@ -0,0 +1,64 @@
...
+
+    u32 consumed_bytes = 0;
+
+    for (u32 b = 0; b < niter; b++) {
+        const u32 offset = b * k_large_buf_payload_max_size;
+
+        u32 read_size = min(available_bytes, k_large_buf_payload_max_size);
+        bpf_clamp_umax(read_size, k_large_buf_payload_max_size);
+
+        const long read_err = (mode == k_large_buf_read_user)
```
rafaelroquetto, line 39: is there any reason why you explicitly went for this, instead of keeping `bpf_probe_read`?

If so, then we need `bpf_probe_read_kernel` instead of `bpf_probe_read`

bpf/gotracer/go_large_buffer.h
```diff
@@ -0,0 +1,146 @@
...
+
+static __always_inline void
+send_http_large_buffers_if_needed(connection_info_t *conn, void *buf, s64 len, u8 direction) {
+
+    if (conn) {
+        u8 event_type = k_lb_event_type_unknown;
+
+        ship_large_request(buf, len, conn, event_type, direction);
+    }
+}
```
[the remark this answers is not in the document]
rafaelroquetto: copilot is right here

bpf/gotracer/go_net_common.h
```diff
@@ -36,6 +36,8 @@
 
 #include <pid/pid_helpers.h>
 
+//SCRATCH_MEM_TYPED(http_client_conn, connection_info_t);
```
rafaelroquetto: stray code?

rafaelroquetto (approved):
LGTM - very minor nit, and one copilot comment, but looks good!

bpf/gotracer/go_large_buffer.h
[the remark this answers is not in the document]
grcevski: good catch.

bpf/common/large_buf_emit.h
grcevski: All reads from the Go side are done with probe_read_user, but the kernel side reads can be either. I'd prefer to keep the Go use read_user for security reasons.

rafaelroquetto (approved):

Issue #1935, opened 2026-04-26
malformed MIME header in LargeBuffer parsing


esara:
When parsing the http response headers with large buffers with TLS - the TLS record gets spliced onto the plaintext header, breaking the headerparsing of the response in
https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/ebpf/common/http_transform.go#L280

```
time=2026-04-25T15:16:58.660Z level=DEBUG msg=Event traceID="[228 246 241 195 97 150 131 230 229 62 120 154 236 4 244 223]" conn="{_:{_:{}} S_addr:[0 0 0 0 0 0 0 0 0 0 255 255 192 168 57 120] D_addr:[0 0 0 0 0 0 0 0 0 0 255 255 162 159 140 245] S_port:49290 D_port:443}" buf="POST /v1/chat/completions HTTP/1.1\r\nAccept: application/json\r\nAuthorization: Bearer sk-proj-XXXr\nContent-Type: application/json\r\nUser-Agent: langchain4j-openai\r\ncontent-length: 3088\r\nhost: api.openai.com\r\n\r\n{\n  "
[...]

PR #1944, opened 2026-04-27
extract larger buffer from java

esara:
## Summary

extract 16k buffer from java instead of 1k when using large buffer
AND
strip of the tls encrypted bytes

address https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1935

Screenshot 2026-04-27 at 4 13 35 PM

## Validation

### DIFF
--- a/pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/instrumentations/util/ByteBufferExtractor.java
+++ b/pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/instrumentations/util/ByteBufferExtractor.java
@@ -9,7 +9,7 @@ import java.nio.ByteBuffer;
 import java.util.Arrays;
 
 public class ByteBufferExtractor {
-  public static final int MAX_SIZE = 1024;
+  public static final int MAX_SIZE = 16 * 1024;
   public static final int MAX_KEY_SIZE = 64;
 
   // Made for JDK8 support
--- a/pkg/internal/java/agent/src/test/java/io/opentelemetry/obi/java/instrumentations/util/ByteBufferExtractorTest.java
+++ b/pkg/internal/java/agent/src/test/java/io/opentelemetry/obi/java/instrumentations/util/ByteBufferExtractorTest.java
@@ -151,13 +151,14 @@ class ByteBufferExtractorTest {
 
   @Test
   void testFlattenUsedByteBufferArray_LenGreaterThanMaxSize() {
-    ByteBuffer buf = ByteBuffer.allocate(2000);
-    for (int i = 0; i < 2000; i++) {
+    int n = ByteBufferExtractor.MAX_SIZE + 1000;
+    ByteBuffer buf = ByteBuffer.allocate(n);
+    for (int i = 0; i < n; i++) {
       buf.put((byte) (i % 128));
     }
     ByteBuffer[] buffers = new ByteBuffer[] {buf};
-    ByteBuffer result = ByteBufferExtractor.flattenUsedByteBufferArray(buffers, 2000);
-    assertEquals(2000, buf.position());
+    ByteBuffer result = ByteBufferExtractor.flattenUsedByteBufferArray(buffers, n);
+    assertEquals(n, buf.position());
 
     assertEquals(ByteBufferExtractor.MAX_SIZE, result.capacity());
 
@@ -360,8 +361,9 @@ class ByteBufferExtractorTest {
 
   @Test
   void testFlattenFreshByteBufferArray_LimitGreaterThanMaxSize() {
-    ByteBuffer buf1 = ByteBuffer.allocate(2000);
-    for (int i = 0; i < 2000; i++) {
+    int n = ByteBufferExtractor.MAX_SIZE + 1000;
+    ByteBuffer buf1 = ByteBuffer.allocate(n);
+    for (int i = 0; i < n; i++) {
       buf1.put((byte) (i % 128));
     }
     buf1.flip();
@@ -374,14 +376,14 @@ class ByteBufferExtractorTest {
       assertEquals((byte) (i % 128), result.get());
     }
     assertEquals(0, buf1.position());
-    assertEquals(2000, buf1.limit());
+    assertEquals(n, buf1.limit());
   }
 
   @Test
   void testFlattenFreshByteBufferArray_MultipleBuffers_ExceedMaxSize() {
-    ByteBuffer buf1 = ByteBuffer.allocate(800);
-    ByteBuffer buf2 = ByteBuffer.allocate(800);
-    for (int i = 0; i < 800; i++) {
+    ByteBuffer buf1 = ByteBuffer.allocate(10240);
+    ByteBuffer buf2 = ByteBuffer.allocate(10240);
+    for (int i = 0; i < 10240; i++) {
       buf1.put((byte) (i + 1));
       buf2.put((byte) (i + 101));
     }
@@ -392,16 +394,16 @@ class ByteBufferExtractorTest {
     ByteBuffer result = ByteBufferExtractor.flattenFreshByteBufferArray(srcs);
     assertEquals(ByteBufferExtractor.MAX_SIZE, result.capacity());
     result.flip();
-    for (int i = 1; i < 800; i++) {
+    for (int i = 1; i < 10240; i++) {
       assertEquals((byte) (i + 1), result.get());
     }
-    for (int i = 0; i < 224; i++) {
+    for (int i = 0; i < 6145; i++) {
       assertEquals((byte) (i + 101), result.get());
     }
     assertEquals(1, buf1.position());
-    assertEquals(800, buf1.limit());
+    assertEquals(10240, buf1.limit());
     assertEquals(0, buf2.position());
-    assertEquals(800, buf2.limit());
+    assertEquals(10240, buf2.limit());
   }
 
   @Test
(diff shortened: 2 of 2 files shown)


pkg/ebpf/common/tcp_large_buffer.go
```diff
@@ -23,6 +23,29 @@ const (
...
 )
 
+// stripLeadingTLSRecords removes one or more contiguous TLS 1.x records from the start of b.
+// Each record is: type (1) + legacy version 0x03xx (2) + length (2, big-endian) + payload.
+// If a record header is present but the chunk ends before the full record, b is discarded
+// (cannot splice safely). If nothing looks like TLS, b is returned unchanged.
+func stripLeadingTLSRecords(b []byte) []byte {
+	for len(b) >= 5 {
+		switch b[0] {
+		case 20, 21, 22, 23: // change_cipher_spec, alert, handshake, application_data
```
mariomac: Can we define this as constants?

pkg/ebpf/common/tcp_large_buffer.go
```diff
@@ -23,6 +23,29 @@ const (
 	largeBufferActionAppend
 )
 
+// stripLeadingTLSRecords removes one or more contiguous TLS 1.x records from the start of b.
+// Each record is: type (1) + legacy version 0x03xx (2) + length (2, big-endian) + payload.
+// If a record header is present but the chunk ends before the full record, b is discarded
+// (cannot splice safely). If nothing looks like TLS, b is returned unchanged.
+func stripLeadingTLSRecords(b []byte) []byte {
+	for len(b) >= 5 {
```
mariomac: I'd replace this `5` by:
```
const tlsRecordHeaderLen = 5
```

pkg/ebpf/common/tcp_large_buffer.go
```diff
@@ -23,6 +23,29 @@ const (
...
+		switch b[0] {
+		case 20, 21, 22, 23: // change_cipher_spec, alert, handshake, application_data
+			if b[1] != 0x03 {
```
mariomac: I'd also replace this `0x03` by a constant named something like `tlsRecordLegacyVersion`

pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/instrumentations/util/ByteBufferExtractor.java
```diff
@@ -9,7 +9,7 @@
 import java.util.Arrays;
 
 public class ByteBufferExtractor {
-  public static final int MAX_SIZE = 1024;
+  public static final int MAX_SIZE = 16 * 1024;
```
mariomac, line 12: Can this (and later buffer size increases) involve some performance penalty? Should we make it configurable?

mariomac (commented):
Thanks for your contribution!! I have few suggestions and a question

pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/instrumentations/util/ByteBufferExtractor.java
esara: yes - I am sure it does, also I think 16k is the max, but maybe @grcevski have more precise answer

ps: this is running in the injected byte code, so making it configurable may be bit harder, because we do not reinject, if already injected

pkg/ebpf/common/tcp_large_buffer.go
esara: yes, done

pkg/ebpf/common/tcp_large_buffer.go
esara: yes, thanks!

pkg/ebpf/common/tcp_large_buffer.go
esara: of course, thanks!

pkg/ebpf/common/tcp_large_buffer.go
```diff
@@ -23,6 +23,41 @@ const (
...
+	tlsLegacyRecordVersionMajor uint8 = 0x03
+
+	tlsRecordHeaderLen = 5
+)
+
+// stripLeadingTLSRecords removes one or more contiguous TLS 1.x records from the start of b.
+// Each record is: type (1) + legacy version 0x03xx (2) + length (2, big-endian) + payload.
+// If a record header is present but the chunk ends before the full record, b is discarded
+// (cannot splice safely). If nothing looks like TLS, b is returned unchanged.
+func stripLeadingTLSRecords(b []byte) []byte {
```
mmat11: I have 2 concerns with this function:

- I think it can generate false positives too often and strip potentially good data on a TCP buffer
- I don't think it belongs in this layer, what is sent here is also sent through the normal event buffer so we should either filter where we get the data (ebpf) or where we use it (the various *_transform.go files)

esara: good points, I was also first thinking about doing this in http_transform
having said that the large buffer exist to parse things out of the body and I am not sure how TLS records can help there... (for any transform)
lets discuss

mmat11: I tried to ask Claude some context on this issue and as far as I understood, it can be solved java-agent side, correct?

disclaimer: I don't know java and this is fully AI-generated, I also tried to ask for a fix and this came out: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1950

would it be an alternative to this implementation? do you have a reproducer locally so we can validate it works?

pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/instrumentations/util/ByteBufferExtractor.java
grcevski: I think it's possible that it will be slower, but unfortunately it's necessary to capture the HTTP payloads. I think we can make this configurable and have OBI inject the configuration when it sends the agent over to the JVM, but it's a more complex change and we should do it separately.

pkg/ebpf/common/tcp_large_buffer.go
grcevski: I think maybe we should leave out the TLS removal for now and implement a different solution. I got BPF logs from Endre and I'll take a look at why we send the TLS buffer. Essentially, there are two possibilities:
1. The java agent has a bug and it sent the encrypted bytes too.
2. We mark the traffic as SSL in the java_tls.c code, but maybe we didn't do it right and the kprobes sent TCP large buffers (with the TLS bytes) on the same connection as the HTTP. This can happen in rare circumstances. Having said this, I think for TCP large buffers we send them over only if we know the protocol, so I'm thinking it's maybe 1. 

@esara how about we amend the PR with just the Java change for now and then we figure out the remaining bug with the TLS buffer being sent?

grcevski: The suggested fix by Claude might be right, but I think with Endre we saw this:

Large buffer: 1024 bytes (part of the headers, cut off by the max limit of 1K)
Large buffer: 2048 bytes (SSL)
Large buffer: 1024 bytes (body)

grcevski: So I think it can't be stale bytes past the end of the buffer, the 2048 confuses me, it's like we sent another buffer and it was exactly 2048 and we thought it was unencrypted.

mmat11: ok, I'll close it since we have more context, just wanted to make sure we don't delete unwanted data here :D

grcevski: OK, so I confirmed the issue is number 2 scenario. From the java agent side we send only good buffers, however, after the first unecrypted payload is sent from the Java side, we see this event:

```
time=2026-04-27T21:14:07.407Z level=DEBUG msg="=== kprobe buf=[\x17\x03\x03\x05jz\xe7\x82\x13*\xc3~-\x18rbh@\xc5R\x9d\x04'\x12\a], pid=3012682, len=1391 === [obi_hand" component=BPFLogger pid=3012835 comm=vert.x-eventloo
time=2026-04-27T21:14:07.407Z level=DEBUG msg="mysql_parse_fixup_header: failed to parse mysql header [mysql_parse_fixup_heade" component=BPFLogger pid=3012835 comm=vert.x-eventloo
time=2026-04-27T21:14:07.407Z level=DEBUG msg="is_mysql: failed to parse mysql header [is_mysql]" component=BPFLogger pid=3012835 comm=vert.x-eventloo
time=2026-04-27T21:14:07.407Z level=DEBUG msg="is_postgres: message length mismatch: message_size=50529643 data_len=1391 [is_p" component=BPFLogger pid=3012835 comm=vert.x-eventloo
time=2026-04-27T21:14:07.407Z level=DEBUG msg="possible invalid message_size: 386073349 [kafka_read_message_size]" component=BPFLogger pid=3012835 comm=vert.x-eventloo
time=2026-04-27T21:14:07.407Z level=DEBUG msg="failed to parse kafka request header [kafka_parse_fixup_request_header]" component=BPFLogger pid=3012835 comm=vert.x-eventloo
time=2026-04-27T21:14:07.407Z level=DEBUG msg="possible invalid message_size: 386073349 [kafka_read_message_size]" component=BPFLogger pid=3012835 comm=vert.x-eventloo
time=2026-04-27T21:14:07.407Z level=DEBUG msg="failed to parse kafka response header [kafka_parse_fixup_response_header]" component=BPFLogger pid=3012835 comm=vert.x-eventloo
time=2026-04-27T21:14:07.407Z level=DEBUG msg="failed to parse kafka response header [is_kafka]" component=BPFLogger pid=3012835 comm=vert.x-eventloo
time=2026-04-27T21:14:07.407Z level=DEBUG msg="large buffer consumed 1391 bytes [http_send_large_buffer]" component=BPFLogger pid=3012835 comm=vert.x-eventloo

```

So it seems that:

1. We failed to correctly mark the connection as SSL from the Java sys_ioctl probe side
2. We sent a large TCP buffer even though it failed all protocol checks? I think Endre's build doesn't have MSSQL.

grcevski: Fix for the bad SSL buffer: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1952

esara: reverting the large buffer change in favor of https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1952

grcevski (approved):
LGTM! We can follow-up later with a fix to make the maximum buffer size configurable if it ends up being an problem. The maximum size we'd hold here could be 150 MB in the worst case scenario, but an application handling 10,000 concurrent connections will use a lot of memory.

mmat11 (approved):
lgtm!

mariomac (approved):

PR #1947, opened 2026-04-28
[exporter/otel] replace panic with proper error handling

florianl:
## Summary

Replace the panic with proper error handling.

## Validation

### DIFF
--- a/pkg/export/otel/metrics.go
+++ b/pkg/export/otel/metrics.go
@@ -301,7 +301,7 @@ func newMetricsReporter(
 			mr.attrGetters, mr.attributes.For(attributes.GenAIClientOperationDuration))
 	}
 
-	mr.reporters = otelcfg.NewReporterPool[*svc.Attrs, *Metrics](cfg.ReportersCacheLen, cfg.TTL, timeNow,
+	mr.reporters, err = otelcfg.NewReporterPool[*svc.Attrs, *Metrics](cfg.ReportersCacheLen, cfg.TTL, timeNow,
 		func(id svc.UID, v *Metrics) {
 			llog := log.With("service", id)
 			llog.Debug("evicting metrics reporter from cache")
@@ -317,6 +317,9 @@ func newMetricsReporter(
 				}
 			}()
 		}, mr.newMetricSet)
+	if err != nil {
+		return nil, fmt.Errorf("creating metrics reporters pool: %w", err)
+	}
 	// Instantiate the OTLP HTTP or GRPC metrics exporter
 	exporter, err := ctxInfo.OTELMetricsExporter.Instantiate(ctx)
 	if err != nil {
--- a/pkg/export/otel/metrics_svc_graph.go
+++ b/pkg/export/otel/metrics_svc_graph.go
@@ -128,7 +128,8 @@ func newSvcGraphMetricsReporter(
 		log:              log,
 	}
 
-	mr.reporters = otelcfg.NewReporterPool[*svc.Attrs, *SvcGraphMetrics](cfg.ReportersCacheLen, cfg.TTL, timeNow,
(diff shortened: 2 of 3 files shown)


mariomac (changes requested):
This panic addresses a unrecoverable error derived from a software bug, like a null pointer or an array out of bounds access. If that happened, I don't think we should leak a message error to the final users (and let OBI gracefully/silently continue) but just crash so the bug is detected as early as possible.

florianl:
While I agree we shouldn't silently ignore unrecoverable errors, I’d argue for proper error handling over panic(). We can still terminate the process to ensure the bug is detected, but doing so through a controlled shutdown allows us to log the incident properly without leaking a full stack trace to the end user.

NimrodAvni78 (approved):

rafaelroquetto:
I just want to chime in - non-rhetorical question? What's the problem with displaying the stack trace to the end user? In my experience, it has actually been generally useful when users report issues/provide logs containing a stack trace.

florianl:
In my experience, people in operations and analytics/data engineering favor a graceful approach to such situations, while people with a development background often care less of the format. As the app terminates in both scenarios, whether through error propagation or the use of panic(), the outcome will be the same.

MrAlias (approved):

grcevski (commented):
This looks good! Do you know if OBI will terminate when this happens, I think it would if we cannot create the metrics pipeline component, or it will sit idle broken? I mean, with the change to send the error back?

mariomac (commented):
I still strongly prefer the "non-graceful" shutdown, given that this error does not derive from anything external, like a wrong configuration or a failed connection, but just a critical code error introduced by us right in the code (I don't see it different from a nil pointer exception). A panic treats this unrecoverable, pretendedly-impossible situation cleanly and doesn't require to pollute the function signatures and code logic with an error path that must never happen.

But given that most the people seem Ok with the change, I won't oppose to merge it.

rafaelroquetto:
I must say I am with @mariomac on this one. Panics are easier to spot, the stack traces are valuable, and - well I make his words mine.

rafaelroquetto:
I also wonder, wouldn't it just be better to validate this like this:

```
ReportersCacheLen int `yaml:"reporters_cache_len" env:"OTEL_EBPF_METRICS_REPORT_CACHE_LEN" validate:"min=1"`
```

and keep the panic? It's a genuine bug not some graceful error.

florianl:
I didn't want to start a conflict :sweat_smile: so, I will be fine with closing it as well. (And sorry that I can't attend todays meeting :pray:)

MrAlias:
Do we have tests that will exercise the panic path in our CI if it becomes possible so this is actually "caught early"? Or, is it possible to have the panic be accessible and release that code to users?

mariomac:
@florianl 
> I didn't want to start a conflict 😅 

No conflicts from my side!! Just an exchange of opposed opinions :D

rafaelroquetto:
> No conflicts from my side!! Just an exchange of opposed opinions :D

Same from me!

florianl:
> I didn't want to start a conflict 😅

My apologies :pray: I used the wrong wording there! I didn't mean to say "conflict", I was actually referring to a technical discussion regarding coding style.

mmat11 (approved):
I also slighlty prefer panics over error propagation in the case of unrecoverable errors but:

- this cache is created with a user supplied value (`ReportersCacheLen`) which is not validated so it's something that can be fixed with a config change
- the error is propagated back and OBI will not start (the errgroup in `RunWithContextInfo` should cancel the context because of this cache init going wrong)
- there are already some approvals

I vote for merging it

mariomac (approved):

PR #1952, opened 2026-04-28
Delay cleaning the SSL requests until we finish HTTP

grcevski:
## Summary

We were deleting the SSL marking on a connection too early for HTTP requests, meaning we'd delete as soon as we saw the first response packet. If the SSL response was large including large buffers, we would send TLS data to userspace mixed with non-TLS, making payload extraction fail.

Relates to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1944

## Validation

### DIFF
--- a/bpf/generictracer/protocol_http.h
+++ b/bpf/generictracer/protocol_http.h
@@ -238,6 +238,7 @@ static __always_inline void cleanup_http_request_data(pid_connection_info_t *pid
 static __always_inline void terminate_http_request_if_needed(pid_connection_info_t *pid_conn) {
     http_info_t *info = bpf_map_lookup_elem(&ongoing_http, pid_conn);
     cleanup_http_request_data(pid_conn, info);
+    bpf_map_delete_elem(&active_ssl_connections, pid_conn);
 }
 
 static __always_inline void process_http_request(http_info_t *info,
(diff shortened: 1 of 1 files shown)


mmat11 (approved):

MrAlias (commented):
The delayed-finish move makes sense for the large TLS response case, but I think we still need to preserve cleanup on the connection-teardown path. I left one inline comment on the specific regression I found.

bpf/generictracer/protocol_http.h
```diff
@@ -231,7 +233,6 @@ static __always_inline void cleanup_http_request_data(pid_connection_info_t *pid
             delete_client_trace_info(pid_conn);
         }
     }
```
MrAlias, line 235: This still needs to clear `active_ssl_connections` on the teardown path. In the `tcp_close` flow we call `terminate_http_request_if_needed()`, but after this change that path only deletes trace state and no longer removes the SSL marker.

That means connections that close without a graceful `SSL_shutdown` can leave stale TLS state behind. At minimum that leaks entries in the LRU map, and if the same process later reuses the same 4-tuple we can keep classifying plaintext traffic as SSL and skip the normal HTTP parsing.

I think the fix should keep the new delayed cleanup in `finish_http()`, but also preserve SSL cleanup when we are tearing the request/connection down from the close path.

grcevski: Good catch!

grcevski: Fixed!

MrAlias (approved):
🚀

PR #1959, opened 2026-04-29
Add NetO11y, AppO11y and StatsO11y metrics doc

pinoOgni:
## Summary

While implementing StatsO11y, I learned a few things about OBI metrics. 

This PR adds NetO11y, AppO11y, and StatsO11y metrics doc with a list of metric examples (some from real runs, some not).

## Validation


devdocs/metrics.md
```diff
@@ -0,0 +1,287 @@
...
+In [pkg/export/attributes/attr_defs.go](../pkg/export/attributes/attr_defs.go) some `AttrReportGroup` type structures are defined for application metrics in both the k8s and non-k8s environment: `appAttributes` and `appKubeAttributes`. Here too, ad hoc attributes such as `httpCommon`, `httpClientInfo`, and so on are added for each metric. There are attributes that default to true and others to false, but which can be enabled by the user during configuration.
+
+In the following methods:
+
+- `setupOtelMeters` in [pkg/export/otel/metrics.go](../pkg/export/otel/metrics.go) for OTEL
+- `newReporter` in [pkg/export/prom/prom.go](../pkg/export/prom/prom.go) for Prometheus
+
+the actual metrics are created using the names defined in [pkg/export/attributes/metric.go](../pkg/export/attributes/metric.go) with the attributes defined and added in [pkg/export/attributes/attr_defs.go](../pkg/export/attributes/attr_defs.go).
+
+Below is a list of application metrics supported by OBI with an example metric in Prometheus format:
```
marctc: Is this list relevant for devs? I'm not sure what value provides. Also, prone to be outdated over time.

pinoOgni: > Is this list relevant for devs? I'm not sure what value provides. 

When I was learning how metrics flow through OBI from the eBPF probe all the way to the exporter, I was reading `metric.go` where you can see the names, but it didn't show what an actual scrape looks like with all the attribute groups composed (appAttributes + httpCommon + httpClientInfo, and so on). Having all the metrics in one place with a fully labeled example was just what I was looking for.

> Also, prone to be outdated over time.

Yes, are right. To address it I can:
- Add a step to the `Add a new metric` recipes asking contributors to update this list, so the doc lives next to the code change.
- Add a rule to `.github/copilot-instructions.md` so that, if a PR adds or renames an entry in `pkg/export/attributes/metric.go`, Copilot suggests adding a matching example to `devdocs/metrics.md`.

WDYT?

marctc: > > Is this list relevant for devs? I'm not sure what value provides.
> 
> When I was learning how metrics flow through OBI from the eBPF probe all the way to the exporter, I was reading `metric.go` where you can see the names, but it didn't show what an actual scrape looks like with all the attribute groups composed (appAttributes + httpCommon + httpClientInfo, and so on). Having all the metrics in one place with a fully labeled example was just what I was looking for.

In many prometheus exporters there's a list of metrics (some also expose the labels as well in table format) usually in README, but here are we are just dumping an example that can be see if you look at localhost:9090/metrics when running OBI locally.

> > Also, prone to be outdated over time.
> 
> Yes, are right. To address it I can:
> 
>     * Add a step to the `Add a new metric` recipes asking contributors to update this list, so the doc lives next to the code change.
> 
>     * Add a rule to `.github/copilot-instructions.md` so that, if a PR adds or renames an entry in `pkg/export/attributes/metric.go`, Copilot suggests adding a matching example to `devdocs/metrics.md`.
> 
> 
> WDYT?

We have a list of metrics in OBI docs, so this would lead to duplicity and complexity for reviewers/developers. If we want to mantain a list of metrics per metrics component, I'd prefer to have them in a central place only. You can link here OBI docs if you want.

pinoOgni: > In many prometheus exporters there's a list of metrics (some also expose the labels as well in table format) usually in README, but here are we are just dumping an example that can be see if you look at localhost:9090/metrics when running OBI locally.

If you know how to trigger the metrics calculation, yes, but it's not always easy and in any case this involves running obi with a certain configuration just to get an example metric.

> We have a list of metrics in OBI docs, so this would lead to duplicity and complexity for reviewers/developers. If we want to mantain a list of metrics per metrics component, I'd prefer to have them in a central place only. You can link here OBI docs if you want.

Oh, I didn't realize we already had this so well documented: https://opentelemetry.io/docs/zero-code/obi/metrics/. I'll remove the list from this doc and link the OBI docs instead.

marctc (approved):
LGTM thanks!

grcevski (approved):
LGTM!

Issue #1799, opened 2026-04-10
Fix Go JSON-RPC instrumentation to follow semantic conventions


NimrodAvni78:
for example in this [test](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/cb7b623a4a12daad483fc19a16fdc424b2385f5c/internal/test/integration/k8s/daemonset_multi_node/k8s_daemonset_multi_node_traces_test.go#L74), we expect the span name to be `Arith.T /jsonrpc`, but it actually should be `Arith/Multiply` from this [endpoint](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/fe783ea2de28d4b370da9939b3f7e2fb26483a60/internal/test/integration/components/testserver/jsonrpc/jsonrpc.go#L32)

it needs to follow to [jsonrpc semantic conventions](https://opentelemetry.io/docs/specs/semconv/rpc/json-rpc/)

apatniv:
Hi, I'd like to work on this issue.

NimrodAvni78:
thank you @apatniv! im assigning this to you

apatniv:
@NimrodAvni78 

The [test](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/cb7b623a4a12daad483fc19a16fdc424b2385f5c/internal/test/integration/k8s/daemonset_multi_node/k8s_daemonset_multi_node_traces_test.go#L74)  mentioned seems to make a request to [`http://localhost:38080/gotracemetoo`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/internal/test/integration/k8s/daemonset_multi_node/k8s_daemonset_multi_node_traces_test.go#L43)[1].

Request[1] ends up being processed [here](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/internal/test/integration/components/testserver/std/std.go#L236-L249)[2]. 

Request[2] ends up making a http json rpc call to `http://testserver:8088/jsonrpc`[3] for method `Arith.Traceme` 

```
http.Post(requestURL, "application/json", bytes.NewReader([]byte(`{"jsonrpc":"2.0","method":"Arith.Traceme","params":[{"A":1,"B":2}],"id":1}`)))
```



[...]

PR #1967, opened 2026-04-30
Fix: GO JSONRPC Span Name

apatniv:
## Summary

Fixes the GO JSONRPC span name. 

## Validation

Address the following issue https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1799


bpf/gotracer/go_nethttp.c
```diff
@@ -1400,15 +1400,16 @@ int obi_uprobe_jsonrpcReadRequestHeaderReturns(struct pt_regs *ctx) {
         return 0;
     }
 
+    bpf_memcpy(invocation->method, "JSONRPC", 7);
```
apatniv: I am not sure about the best way to share a string literal across ebpf and golang code. If there are any examples let me know.

grcevski: This is fine, but it would be nicer to define this constant earlier in the file, like this:

```
const char JSON_RPC[] = "JSONRPC";
const u32 JSON_RPC_SIZE = sizeof(JSONRPC) - 1;
```

And later:

```
bpf_memcpy(invocation->method, JSON_RPC, JSON_RPC_SIZE);
```

bpf/gotracer/go_nethttp.c
```diff
@@ -1400,15 +1400,16 @@ int obi_uprobe_jsonrpcReadRequestHeaderReturns(struct pt_regs *ctx) {
...
     if (!read_go_str("JSON-RPC method",
                      (void *)rpc_request_addr,
                      go_offset_of(ot, (go_offset){.v = _jsonrpc_request_header_service_method_pos}),
-                     invocation->method,
-                     k_method_max_len)) {
+                     invocation->pattern,
+                     k_pattern_max_len)) {
         bpf_dbg_printk("Failed to read JSON-RPC method from: %llx", rpc_request_addr);
         return 0;
     }
```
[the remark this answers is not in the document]
grcevski: This comment is valid, we can safely move setting the method after we successfully read in the actual JSON RPC method.

pkg/ebpf/common/spanner.go
```diff
@@ -21,6 +22,14 @@ func HTTPRequestTraceToSpan(trace *HTTPRequestTrace) request.Span {
 	scheme := cstr(trace.Scheme[:])
 	origHost := cstr(trace.Host[:])
 
+	var overrideTraceName string
+	if method == "JSONRPC" {
+		// json rpc signal
+		method = http.MethodPost
```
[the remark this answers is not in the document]
grcevski: Not an issue, we do this all over the place. I'd just move the "JSONRPC" constant as a global const in the file.

pkg/ebpf/common/spanner.go
```diff
@@ -21,6 +22,14 @@ func HTTPRequestTraceToSpan(trace *HTTPRequestTrace) request.Span {
...
+		overrideTraceName = pattern
+		pattern = path
+	}
```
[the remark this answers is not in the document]
grcevski: I'm not sure this is true.

grcevski (commented):
This looks great! I just have few small nit comments and we should be good to go.

grcevski:
The integration tests are failing, they might be expecting the old value. https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/25166659656/job/73801270554?pr=1967

apatniv:
@grcevski Thanks for the feedback. Addressed your comments.

grcevski (approved):
LGTM!

pkg/ebpf/common/spanner.go
```diff
@@ -13,6 +14,8 @@ import (
 	"go.opentelemetry.io/obi/pkg/internal/sqlprune"
 )
 
+const jsonRpc = "JSONRPC" // must remain in sync with value used in obi_uprobe_jsonrpcReadRequestHeaderReturns
```
MrAlias:
```suggestion
const jsonRPC = "JSONRPC" // must remain in sync with value used in obi_uprobe_jsonrpcReadRequestHeaderReturns
```

and updated where used ...

pkg/ebpf/common/spanner.go
```diff
@@ -21,6 +24,14 @@ func HTTPRequestTraceToSpan(trace *HTTPRequestTrace) request.Span {
 	scheme := cstr(trace.Scheme[:])
 	origHost := cstr(trace.Host[:])
 
+	var overrideTraceName string
+	if method == jsonRpc {
+		// json rpc signal
+		method = http.MethodPost
+		overrideTraceName = pattern
+		pattern = path
```
MrAlias: This fixes the displayed trace name, but it still leaves the span as plain HTTP. In this pipeline, the JSON-RPC-specific export path is gated on `span.SubType == HTTPSubtypeJSONRPC` and a populated `span.JSONRPC`, and this code never sets either of those.

That means Go JSON-RPC server spans will still miss the JSON-RPC/RPC semantics downstream, including `rpc.system=jsonrpc`, `rpc.method`, and the RPC metrics path. Since `#1799` is about following JSON-RPC semantic conventions, I think this needs to construct an actual JSON-RPC span shape here, not just override `span.name`.

MrAlias (commented):
The span name change looks right, but I think there’s still one semantic-convention gap in the Go JSON-RPC path.

pkg/ebpf/common/spanner.go
apatniv: Ok. Will fix lint issues

pkg/ebpf/common/spanner.go
apatniv: @MrAlias Will look more into this.

bpf/gotracer/go_nethttp.c
```diff
@@ -48,6 +48,9 @@
 
 #include <shared/obi_ctx.h>
 
+const char JSON_RPC[] = "JSONRPC";
```
rafaelroquetto: I believe this should be defined in nethttp.h - we should probably static_assert that the sizeof jsonrpc == k_method_len (or <=).

Look for the existing pattern in nethttp.h for the traceparent string and stick to it.


You don't need JSON_RPC_SIZE - you can use sizeof(invocation->method)

pkg/ebpf/common/spanner.go
```diff
@@ -13,6 +14,8 @@ import (
 	"go.opentelemetry.io/obi/pkg/internal/sqlprune"
 )
 
+const jsonRPC = "JSONRPC" // must remain in sync with value used in obi_uprobe_jsonrpcReadRequestHeaderReturns
```
rafaelroquetto: We shouldn't introduce hidden coupling between this and the eBPF layer.

Instead of having a string, maybe pass a boolean up from the ebpf layer indicating this is a json rpc flag.

pkg/ebpf/common/spanner.go
```diff
@@ -21,6 +24,14 @@ func HTTPRequestTraceToSpan(trace *HTTPRequestTrace) request.Span {
 	scheme := cstr(trace.Scheme[:])
 	origHost := cstr(trace.Host[:])
 
+	var overrideTraceName string
+	if method == jsonRPC {
```
rafaelroquetto: So why do we need this explicit heuristic? Can't we just ship the right info from the ebpf layer?

Since you already know this is a JSONRPC method in ebpf, you can just set `invocation->method = "POST"` I suppose, and then here do something like:

```
if trace.IsJsonrpc != 0 {
    span.SubType = request.HTTPSubtypeJSONRPC
    span.JSONRPC = &request.JSONRPC{Method: cstr(trace.Pattern[:])}
}
```

so you no longer to keep the string in userspace and also in sync, and it's more robust overall

pkg/ebpf/common/spanner.go
```diff
@@ -68,7 +79,8 @@ func HTTPRequestTraceToSpan(trace *HTTPRequestTrace) request.Span {
 			UserPID:   app.PID(trace.Pid.UserPid),
 			Namespace: trace.Pid.Ns,
 		},
-		Statement: schemeHost,
+		Statement:         schemeHost,
+		OverrideTraceName: overrideTraceName,
 	}
```
rafaelroquetto: If you implement the suggestion above, this is no longer necessary and I think it can be removed.

rafaelroquetto (changes requested):

pkg/ebpf/common/spanner.go
apatniv: Only reason this was done is to avoid bloating up the existing `server_http_func_invocation_t` with additional field. Just in case if there a future need for other fields. 
Relevant discussion regarding the concern: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1799#issuecomment-4243594627

bpf/gotracer/go_nethttp.c
apatniv: Nice.

pkg/ebpf/common/spanner.go
apatniv: 100% agree with you that it creates hidden coupling.

pkg/ebpf/common/spanner.go
apatniv: Don't we need to populate other fields in `request.JSONRPC` like `RequestID` at least as it would be useful for correlating. 

Even if this needs to be done in the future, we need either additional fields in ebpf layer. So wondering if all these fields need to go existing struct or use struct composition (or maps).

I think this ties with what @MrAlias was commented about.

pkg/ebpf/common/spanner.go
apatniv: Yes agreed. See other comments regarding what to do other fields in the future.

pkg/ebpf/common/spanner.go
rafaelroquetto: @apatniv exactly - an additional `bool` field in `server_http_func_invcation_t` and also in `http_request_trace_t` should be fine, as we have plenty of padding in those structs we can leverage.

rafaelroquetto:
Hey @apatniv - thanks for your comments.

I was re-evaluating this PR and the issue with a bit of more depth and I think this would be worth revisiting.

So if I understand this correctly, to bring everything up on par with the json rpc semantic conventions, we need:

 - `rpc.system` (constant "jsonrpc")
 - `rpc.service` and `rpc.method` (split from the JSON-RPC method, e.g. "Arith.Multiply" becomes "Arith" / "Multiply")
 - `jsonrpc.protocol.version` (e.g. "2.0")
 - `jsonrpc.request.id` (the request id)
 - `jsonrpc.error_code` / `jsonrpc.error_message` (on error responses)

they all require a few different set of changes to the code, so it is up to you if you want to open multiple smaller PRs or one comprehensive one (if you do open smaller ones, we should not resolve https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1799 until everything is merged).

I think this ties to what @MrAlias was saying about missing fields but he can correct me here otherwise.

Now, a couple of points regarding the implementation:

 - your fix in `go_nethttp.c` redirecting the service method to `invocation->pattern` is correct - that keeps  the HTTP method untainted (e.g. `POST)`
 - the use of a sentinel value `"JSONRPC"` is incorrect - apart from what we discussed with regards to coupling and the fragility of the implementation, it forces the userspace to *undo* it and bring back the actual HTTP method
 - instead, you should use a proper `bool` that the userspace can check
 - userspace should not rely on `OverrideTraceName` - instead, we already have code that correctly deals with it in the `TraceName()` function in `span.go` (look for `JSONRPC` there) - instead, we should leverage this codepath by setting `span.SubType = HTTPSubtypeJSONRPC` and populate `JSONRPC{} ` so that everything downstream handles it accordingly, otherwise the span ships as plain HTTP: no rpc.system/rpc.method/jsonrpc.* attributes, and metrics land in HTTP buckets instead of the RPC ones.

pkg/ebpf/common/spanner.go
apatniv: See discussion with @rafaelroquetto . It should address your concern.

bpf/gotracer/go_nethttp.c
apatniv: Would no longer be implemented using hard coded string literals. So, closing this conversation.

apatniv:
@rafaelroquetto 
>we should not resolve https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1799 until everything is merged

Not a problem 

----

>userspace should not rely on OverrideTraceName - instead, we already have code that correctly deals with it in the TraceName() function in span.go (look for JSONRPC there) - instead, we should leverage this codepath by setting span.SubType = HTTPSubtypeJSONRPC and populate JSONRPC{} 

Yes.

rafaelroquetto:
hey @apatniv just wondering if you are still planning on following up with this PR.

apatniv:
Should be able to send in a patch today. Running into intermittent test
failures locally.

apatniv:
Will do additional changes related to requestId and error object in follow up PRs.

bpf/common/common.h
```diff
@@ -77,7 +77,8 @@ typedef struct http_request_trace {
     u16 status;
     unsigned char method[k_method_max_len];
     unsigned char scheme[k_scheme_max_len];
-    u8 _pad1[11];
+    bool is_jsonrpc;
```
rafaelroquetto: tiny nit: we can move this in the place of _pad0 and then get rid of that member entirely

rafaelroquetto:
@apatniv you may need to rebase your branch onto main for the tests

apatniv:
@rafaelroquetto Merged latest main changes.

rafaelroquetto:
@apatniv looks like the tests are genuinely failing this time

apatniv:
Let me a take a look. Running into some docker issues locally to reproduce.

apatniv:
The failing red_test should be fixed now.

apatniv:
I was looking at the logs and from what it appears, it could be because of docker images cache in github workflows. 

```
obi-1                   | time=2026-05-27T15:55:50.074Z level=DEBUG msg="tp=00-c1ce063326fb2804a803af7352482f2a-672604ec6af8c1a8-01 [serve_http_returns]" component=BPFLogger pid=18645 comm=testserver
obi-1                   | time=2026-05-27T15:55:50.074Z level=DEBUG msg="method=Arith.Mhttp [serve_http_returns]" component=BPFLogger pid=18645 comm=testserver
obi-1                   | time=2026-05-27T15:55:50.074Z level=DEBUG msg="path=/jsonrpc [serve_http_returns]" component=BPFLogger pid=18645 c
omm=testserver
obi-1                   | time=2026-05-27T15:55:50.074Z level=DEBUG msg="pattern= [serve_http_returns]" component=BPFLogger pid=18645 comm=testserver
obi-1                   | time=2026-05-27T15:55:50.086Z level=DEBUG msg="goroutine_addr=2eedf9728d20 [serve_http_returns]" component=BPFLogger pid=18646 comm=testserver
obi-1                   | time=2026-05-27T15:55:50.086Z level=DEBUG msg="method=Arith.Mhttp [serve_http_returns]" component=BPFLogger pid=18646 comm=testserver
obi-1                   | time=2026-05-27T15:55:50.086Z level=DEBUG msg="path=/jsonrpc [serve_http_returns]" component=BPFLogger pid=18646 comm=testserver
obi-1                   | time=2026-05-27T15:55:50.086Z level=DEBUG msg="pattern= [serve_http_returns]" component=BPFLogger pid=18646 comm=testserve
```
This is the still the old way of extracting the json rpc method into method field: `/Arith.M` 
Also, can't see the occurrence of `is_jsonrpc=` anywhere in the logs.

apatniv:
@rafaelroquetto Can you take a look now as all tests are passing. Thank you.

rafaelroquetto (approved):
one minor nit, but LGTM otherwise. Approving.

Congratulations on your first contribution @apatniv! Thank you for that!

PR #1975, opened 2026-05-01
Split ring buffer reader and parser into separate goroutines, reduce mutex overhead

rafaelroquetto:
## Summary

Holding a mutex while doing work is an anti-pattern. We were holding `rbf.access` across the entire parse call on every event (and paying the context of locking unlocking the mutex).

This PR splits the work into two concurrent goroutines: a reader that drains the ring buffer into a record pool, and a parser that decodes and accumulates spans. The lock is now held only for the O(1) slice append, not during parse.

## Benchmark

`BenchmarkForwardRingbuf` vs main (`-benchtime=5s -count=10`, `benchstat`):

| parse cost | old | new | Δ |
|---|---|---|---|
| 1µs (HTTP trace decode) | 430k events/s | 727k events/s | **+69%** |
| 10µs (GoTracer + LRU lookup) | 49k events/s | 96k events/s | **+97%** |
| 100µs (unknown TCP protocol) | 5k events/s | 10k events/s | **+99%** |

p=0.000, n=10.

## Validation

### DIFF
--- a/pkg/ebpf/common/ringbuf.go
+++ b/pkg/ebpf/common/ringbuf.go
@@ -273,55 +273,73 @@ func (rbf *ringBufForwarder[T]) parserLoop(
 	workIdx <-chan int,
 	out *msg.Queue[[]T],
 ) {
+	pending := make([]int, 0, cap(workIdx))
+	parsed := make([]T, 0, cap(workIdx))
+
 	for {
-		var i int
-		var ok bool
+		pending = pending[:0]
+		parsed = parsed[:0]
+
+		// Block until at least one record is ready.
 		select {
 		case <-ctx.Done():
 			return
-		case i, ok = <-workIdx:
+		case i, ok := <-workIdx:
 			if !ok {
 				return
 			}
+			pending = append(pending, i)
+		}
+
+		// Drain any additional records that are already waiting.
+		for {
+			select {
+			case i, ok := <-workIdx:
+				if ok {
+					pending = append(pending, i)
+					continue
+				}
+			default:
+			}
+			break
 		}
 
 		if depth := len(workIdx); depth == cap(workIdx)-1 {
 			rbf.logger.Debug("parser falling behind: work queue full", "depth", depth+1)
 		}
 
-		rbf.parseAndAccumulate(ctx, i, records, freeIdx, out)
-	}
-}
-
-func (rbf *ringBufForwarder[T]) parseAndAccumulate(
-	ctx context.Context,
-	i int,
-	records []ringbuf.Record,
-	freeIdx chan<- int,
-	out *msg.Queue[[]T],
-) {
-	item, ignore, err := rbf.parse(&records[i])
-	freeIdx <- i
-
-	if err != nil {
-		rbf.logger.Debug("error parsing perf event", "error", err)
-		return
-	}
-	if ignore {
-		return
-	}
-
-	rbf.access.Lock()
-	rbf.items[rbf.itemsLen] = item
-	rbf.itemsLen++
-	if rbf.itemsLen == rbf.cfg.BatchLength {
-		rbf.logger.Debug("submitting batch (full)", "len", rbf.itemsLen)
-		rbf.flushEvents(ctx, out)
-		if rbf.ticker != nil {
-			rbf.ticker.Reset(rbf.cfg.BatchTimeout)
+		// Parse outside the lock, return each slot to the pool immediately.
+		for _, i := range pending {
+			item, ignore, err := rbf.parse(&records[i])
+			freeIdx <- i
+			if err != nil {
+				rbf.logger.Debug("error parsing perf event", "error", err)
+				continue
+			}
+			if !ignore {
+				parsed = append(parsed, item)
+			}
 		}
+
+		if len(parsed) == 0 {
+			continue
+		}
+
+		// Lock once to enqueue the whole batch.
+		rbf.access.Lock()
+		for _, item := range parsed {
+			rbf.items[rbf.itemsLen] = item
+			rbf.itemsLen++
+			if rbf.itemsLen == rbf.cfg.BatchLength {
+				rbf.logger.Debug("submitting batch (full)", "len", rbf.itemsLen)
+				rbf.flushEvents(ctx, out)
+				if rbf.ticker != nil {
+					rbf.ticker.Reset(rbf.cfg.BatchTimeout)
+				}
+			}
+		}
+		rbf.access.Unlock()
 	}
-	rbf.access.Unlock()
 }
 
 func (rbf *ringBufForwarder[T]) storeLastReadAt(t time.Time) {
--- a/pkg/ebpf/common/ringbuf.go
+++ b/pkg/ebpf/common/ringbuf.go
@@ -293,6 +293,10 @@ func (rbf *ringBufForwarder[T]) parserLoop(
 			pending = append(pending, i)
 		}
 
+		if depth := len(workIdx); depth == cap(workIdx)-1 {
+			rbf.logger.Debug("parser falling behind: work queue full", "depth", depth+1)
+		}
+
 		// Drain any additional records that are already waiting.
 		for {
 			select {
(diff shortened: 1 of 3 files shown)


grcevski (commented):
This looks great, do you think it would be possible to add some sort of validation tests that we get all events? I think we do, but I just want to make sure we don't break something as concurrent issues are hard to reason about. 

What I'm thinking is, if not possible to use the ring buffer directly, can we extract the actual ring buffer usage in a way that we can replace it with a bounded array for the tests? Then we can write tests for this code.

grcevski:
I think if you tried to decode unknown TCP protocol, it should be even a greater win, don't you think? Because that's where the true cost is.

pkg/ebpf/common/ringbuf.go
```diff
@@ -185,44 +179,166 @@ func (rbf *ringBufForwarder[T]) flushOnAvailableBytes(ctx context.Context, event
...
-	// 4. When the length of the batch slice reaches cfg.BatchLength,
-	//    submit it to the next stage of the pipeline
+	rbf.items = make([]T, rbf.cfg.BatchLength)
+	rbf.itemsLen = 0
+
+	poolSize := 2 * rbf.cfg.BatchLength
+	records := make([]ringbuf.Record, poolSize)
+	freeIdx := make(chan int, poolSize)
+	workIdx := make(chan int, poolSize)
+	for i := range poolSize {
```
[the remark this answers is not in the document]
rafaelroquetto: done

pkg/ebpf/common/ringbuf.go
```diff
@@ -185,44 +179,166 @@ func (rbf *ringBufForwarder[T]) flushOnAvailableBytes(ctx context.Context, event
...
+		freeIdx <- i
+		switch {
+		case errors.Is(err, ringbuf.ErrFlushed):
+			rbf.logger.Debug("ring buffer already flushed")
+			return true
+		case errors.Is(err, ringbuf.ErrClosed):
+			rbf.logger.Debug("ring buffer is closed")
+			return false
+		default:
 			rbf.logger.Error("error reading from perf reader", "error", err)
```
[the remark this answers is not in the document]
rafaelroquetto: done

pkg/ebpf/common/ringbuf_bench_test.go
```diff
@@ -0,0 +1,115 @@
...
+		slog.New(slog.NewTextHandler(io.Discard, nil)),
+		&metricsReporter{},
+	)
+
+	b.ResetTimer()
+	go fwd(ctx, out)
+
+	for range batches {
+		<-sub
+	}
```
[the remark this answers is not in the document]
rafaelroquetto: done

grcevski (approved):
LGTM!

rafaelroquetto:
@grcevski  nice, thanks for the feedback! I added a new 100µs latency to the benchmark to simulate a slow unknown TCP packet parsing (and updated the table on the description), as expected this shows the greatest improvement, since at that cost everything is busy the whole time.

I also added a test that checks N events in -> N events out to make sure we aren't losing anything.

grcevski (approved):

pkg/ebpf/common/ringbuf_bench_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+func benchForwardRingbuf(b *testing.B, latency time.Duration) {
+	b.Helper()
+
+	// Round up to a full batch so every flush is triggered by BatchLength events,
+	// not by a timeout. This avoids a stall on the last partial batch and keeps
+	// the measurement boundary clean.
+	n := ((b.N + benchBatchLen - 1) / benchBatchLen) * benchBatchLen
+
+	rb := &benchReader{n: n, latency: latency}
```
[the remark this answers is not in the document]
rafaelroquetto: done

mariomac (approved):
Good stuff!

pkg/ebpf/common/ringbuf.go
```diff
@@ -185,44 +179,168 @@ func (rbf *ringBufForwarder[T]) flushOnAvailableBytes(ctx context.Context, event
...
+				if ok {
+					pending = append(pending, i)
+					continue
+				}
+			default:
+			}
+			break
+		}
+
+		if depth := len(workIdx); depth == cap(workIdx)-1 {
```
mmat11: should we log this before draining remaining data?

mmat11 (approved):
🚀

pkg/ebpf/common/ringbuf.go
rafaelroquetto: Done

MrAlias (commented):
One non-blocking concern around shutdown behavior. The throughput refactor and added coverage look good overall, but I think the new reader/parser split slightly weakens draining semantics during cancellation.

pkg/ebpf/common/ringbuf.go
```diff
@@ -185,44 +179,168 @@ func (rbf *ringBufForwarder[T]) flushOnAvailableBytes(ctx context.Context, event
...
+	parsed := make([]T, 0, cap(workIdx))
+
+	for {
+		pending = pending[:0]
+		parsed = parsed[:0]
+
+		// Block until at least one record is ready.
+		select {
+		case <-ctx.Done():
+			return
```
MrAlias, line 288: This is a non-blocking concern, but I think the cancellation semantics changed here. In the old serial loop, `ctx.Done()` could only interrupt the current read / parse path. With the split reader/parser design, we can now return here while `workIdx` still contains records that were already read from the kernel ring buffer and while `rbf.items` may already hold parsed-but-unflushed items. Since those records have already advanced the ring buffer consumer position, they are dropped on shutdown instead of being drained.

If preserving already-read events during graceful shutdown matters for OBI, I’d prefer to let the reader stop first, then have the parser drain the remaining `workIdx` entries and flush any partial batch before exiting.

grcevski: We use the same pattern I think for all exporters, so even if we drained the events, the exporters would terminate based on the same signal ctx.Done() and we cannot guarantee that the events will be exported. I think if OBI is shutting down, it's normal for events to be missed. I think there would be very little difference to the stop being issued 100ms before.

MrAlias: Yeah, and I get that point of view. That's why this isn't per se blocking. 

Just this kind of inconsistency is the kind of thing that leads to bugs. Both in user expectations, and testing. 

Users/Devs that expect a shutdown to flush all telemetry they expect OBI to have seen and then not see it leads to false-negatives and frustration.

rafaelroquetto: > Just this kind of inconsistency is the kind of thing that leads to bugs. Both in user expectations, and testing.

But there's no bug or regression here. The behaviour is consistent - before we also dropped events, it's just now you are able to X-ray what used to be a "single pipe".

So the previous code had:

```
read -> parse -> append to rbf.items -> if full, flush
```

even tho there was no explicit `Done()`, it worked implicitly here:

```
func (rbf *ringBufForwarder[T]) bgListenContextCancelation(ctx context.Context, eventsReader ringBufReader) {
	<-ctx.Done()
	rbf.logger.Debug("context is cancelled. Closing events reader")
	_ = eventsReader.Close()
}
```

i.e. it closed the ring buffer, causing the records sitting on the ring buffer and also non-flushed events in `rbf.items` to never be flushed.

The problem is that there's no way to draw a reliable line of what should be flushed vs what shouldn't - there will always be events sitting somewhere in the pipeline, from the actual ringbuffer, to anything meant to be flushed forward, to events in the decorators. 


Even if I implement your suggestion, it doesn't tackle the fact that the entire pipeline is being shutdown - so even if the drained events make it to the next stage, they will be dropped because that stage is shutting down - actually, `queue.SendCtx` bails on `Done()` as well, so the events will die there. There's no point to this change, if we want to ensure that any event on the pipeline gets drained (and not just here) we need to rethink all of our shutdown logic - which honestly is not worth it/non-issue. So yeah, any change here won't fix anything.

PR #1987, opened 2026-05-04
refactor(ebpf): extend HTTP request and protocol-args structs for chunked traceparent scanner

smehboub:
# refactor(ebpf): extend HTTP request and protocol-args structs for chunked traceparent scanner

## Summary

Preparatory data-structure changes for the chunked traceparent scanner
(follow-up PR). No behaviour change: all new fields are zeroed on every
existing allocation path and no existing code reads or writes them yet.

This PR can be reviewed independently ("are the new fields correctly sized,
aligned, and named?") before the scanner logic lands.

### `bpf/common/http_info.h`

**`original_trace_id[16]`** — stable large-buffer correlation key.

When the eBPF scanner later finds a traceparent in a large HTTP request, it
overwrites `tp.trace_id`. The `http_send_large_buffer` helper must still be
able to look up the original buffer by the key that was stored at request
start. `original_trace_id` is initialised once from `tp.trace_id` when the
request is first seen; it is never modified by subsequent traceparent
discoveries. Both the eBPF and Go sides will use this field as the map key
for large-buffer correlation.

Padding updated to maintain 8-byte struct alignment.

### `bpf/common/http_types.h`

New fields added to `call_protocol_args_t`, the scratch structure threaded
through every tail-call in the protocol parsing pipeline:

| Field | Purpose |
|---|---|
| `is_append` | Set on the still\_reading (large-buffer append) path, clear on the initial request parse path |
| `u_buf_is_user` | `u_buf` points into userspace (SSL/Java-TLS/Go uprobe); callers must use `bpf_probe_read_user` |
| `niter` | Chunked-scanner iteration counter, needed for the verifier's loop-bound proof in `bpf_loop` |
| `orig_buf` | Original `ITER_UBUF` or single-segment `ITER_IOVEC` buffer pointer |
| `full_bytes_len` | Actual payload length when `orig_buf` is set (equals `bytes_len` otherwise) |

Padding updated for alignment.

## Testing

No test changes. These are struct layout additions; alignment correctness is
validated by the `make build` step via `go vet` and the BPF verifier.

```
make build
```

## How to review

- Verify that `sizeof(http_info_t)` padding is correct (16 + 3 + _pad[7] = 8-byte aligned block).
- Verify that `call_protocol_args_t` fields preserve natural alignment.
- Confirm no existing code sets or reads the new fields (they are only wired up in the follow-up PR).

## Depends on

- fix(tracing): correct trace-id not-found sentinel and use Copy() for dummy BPF program specs

### DIFF
--- a/bpf/common/http_info.h
+++ b/bpf/common/http_info.h
@@ -23,6 +23,7 @@ typedef struct http_info {
     u64 req_monotime_ns;
     u64 extra_id;
     tp_info_t tp;
+    unsigned char original_trace_id[16];
     pid_info pid;
     u32 len;
     u32 resp_len;
@@ -34,6 +35,7 @@ typedef struct http_info {
     u8 has_large_buffers;
     u8 direction;
     u8 submitted;
+    u8 _pad2[3];
     enum event_source_type event_source;
-    u8 _pad[2];
+    u8 _pad[7];
 } http_info_t;
--- a/bpf/common/http_types.h
+++ b/bpf/common/http_types.h
@@ -51,13 +51,26 @@ typedef struct call_protocol_args {
     protocol_selector_t protocols;
     u8 skip_tp_parsing;
     u8 use_bpf_loop;
-    u8 pad[1];
+    u8 pad[5];
+    u32 niter;
     int bytes_len;
     u16 orig_dport;
-    u16 _pad2;
+    // is_append: set when this invocation originates from the large-buffer
+    // still_reading path (obi_handle_buf_with_args / __obi_protocol_http),
+    // as opposed to the initial request parse path. Unrelated to HTTP
+    // chunked transfer encoding.
+    u8 is_append;
+    u8 u_buf_is_user;
+    u8 pad_align_ubuf[8];
     u64 u_buf;
     u64 self_ref_parent_id;
     lw_thread_t lw_thread;
+    u64 orig_buf;
+    // full_bytes_len is the actual payload length when orig_buf is set
+    // (ITER_UBUF or single-segment ITER_IOVEC paths in return_recvmsg).
+    // It equals bytes_len in all other cases.
+    u32 full_bytes_len;
+    u32 _pad2;
 } call_protocol_args_t;
 
 // Here we keep information on the packets passing through the socket filter
--- a/bpf/common/trace_util.h
+++ b/bpf/common/trace_util.h
@@ -18,6 +18,8 @@ struct callback_ctx {
     u8 _pad[4];
 };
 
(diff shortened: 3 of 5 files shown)


grcevski:
> When the eBPF scanner later finds a traceparent in a large HTTP request, it
overwrites tp.trace_id. The http_send_large_buffer helper must still be
able to look up the original buffer by the key that was stored at request
start. original_trace_id is initialised once from tp.trace_id when the
request is first seen; it is never modified by subsequent traceparent
discoveries. Both the eBPF and Go sides will use this field as the map key
for large-buffer correlation.

I don't think this is how things work, what exactly is trying to do? The overwrite needs to happen.

rafaelroquetto (changes requested):
See comments on https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1988

smehboub:
> > When the eBPF scanner later finds a traceparent in a large HTTP request, it
> > overwrites tp.trace_id. The http_send_large_buffer helper must still be
> > able to look up the original buffer by the key that was stored at request
> > start. original_trace_id is initialised once from tp.trace_id when the
> > request is first seen; it is never modified by subsequent traceparent
> > discoveries. Both the eBPF and Go sides will use this field as the map key
> > for large-buffer correlation.
> 
> I don't think this is how things work, what exactly is trying to do? The overwrite needs to happen.

Thanks for the review @grcevski. 

You are right to question original_trace_id. I will address the rationale directly in [#1988](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1988) where the field is actually used, and rework or drop it based on the discussion there.

smehboub:
I will close this PR because the struct changes in http_info.h and http_types.h are already part of [#1988](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1988). They can only be properly evaluated alongside the code that uses them.

smehboub:
> > When the eBPF scanner later finds a traceparent in a large HTTP request, it
> > overwrites tp.trace_id. The http_send_large_buffer helper must still be
> > able to look up the original buffer by the key that was stored at request
> > start. original_trace_id is initialised once from tp.trace_id when the
> > request is first seen; it is never modified by subsequent traceparent
> > discoveries. Both the eBPF and Go sides will use this field as the map key
> > for large-buffer correlation.
> 
> I don't think this is how things work, what exactly is trying to do? The overwrite needs to happen.

Hi,

I answered in this comment: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1988#issuecomment-4480759173

PR #1988, opened 2026-05-04
feat(tracing): add chunked tail-call traceparent scanner for large HTTP headers

### LINKED ISSUE #1381: Bug: traceparent header not detected when located beyond the eBPF capture buffer (large headers offset)
## Description
In scenarios where an incoming HTTP request contains a large volume of header data (e.g., extensive Cookie or User-Agent headers), the eBPF instrumentation fails to extract the existing traceparent.

If the traceparent is located deep in the header block at an offset exceeding the number of bytes captured by the eBPF program, the instrumentation fails to find the context and generates a new trace_id. This creates a "trace break" in distributed tracing.

## Steps to Reproduce
- Send an HTTP request to an instrumented service.
- Populate the request with several large headers (e.g., a massive Cookie) before the traceparent header.
- Ensure the cumulative size of headers preceding traceparent exceeds the internal capture buffer (e.g., several thousand bytes).
- Observe the exported spans.

## Expected Behavior
The eBPF agent should be able to scan the full header block or provide a configurable offset/buffer size to ensure traceparent discovery, even in requests with large metadata.

## Actual Behavior
The agent only inspects the first N bytes of the request. If traceparent is located at an offset greater than N bytes, it is ignored, and a new trace is started.

## Environment
OS: Ubuntu 22.04.5 LTS - Kernel 6.8.0-101-generic
Agent Version: v0.5.0

smehboub:
## Summary

Find `traceparent` headers buried after large HTTP headers.

Fixes: #1381

---

## Description

As described in the related issue, when a service places large headers (such as auth tokens or cookies) before traceparent, the tracing context can land beyond the first ~1 KB that OBI scans. OBI then misses it and generates a new trace ID, silently breaking the distributed trace.

Following @grcevski's suggestion in [comment in #1381](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1381#issuecomment-3987386476) to implement tail calls as in tpinjector, this PR extends the eBPF tracer to walk the HTTP request buffer in successive windows.

---

## Design Overview

To bypass the 1 KB buffer limitation without hitting eBPF bytecode complexity bounds, the parsing logic is split into successive windows using eBPF tail calls.

- **Windowed Scanning:** A shared helper (`__tp_chunk_scan`) processes the buffer in 956-byte steps until traceparent is found or the end of HTTP headers (`\r\n\r\n`) is reached. The 956-byte window ensures that headers straddling a standard 1024-byte boundary are captured entirely in a single read. It never reads beyond the configured budget (`OTEL_EBPF_BPF_MAX_REQUEST_TP_PARSE_SIZE_KB`, default 4 KB, max 27 KB).
- **Logic Splitting:** The scanning path is divided into two separate programs in the jump_table to handle different tp_p sources:
  - **Program 14 (obi_parse_traceparent_http):** Handles the initial ingress path.
  - **Program 15 (obi_parse_traceparent_http_append):** Handles the large-buffer append path.
- **State Management:** No new BPF maps are introduced. The iteration counter (niter), offsets, and buffer pointers are tracked across tail calls using four new fields in the existing per-CPU scratchpad (`call_protocol_args_t`).

### State machine

tp_state_machine

### Requirements & Compatibility

- **Kernel:** Requires Linux kernel 5.17+ for full tail-call support.
- **Fallback:** Automatically falls back to the legacy single-chunk behavior on older kernels.
- **Activation:** Active only when `OTEL_EBPF_CONTEXT_PROPAGATION` or `OTEL_EBPF_TRACK_REQUEST_HEADERS` is enabled.

---

## Relationship with tpinjector

`tpinjector` and the chunked scanner sit on **opposite traffic directions**:

- **Chunked scanner** — kretprobe on `return_recvmsg` (ingress). Extracts the
  incoming `traceparent` via `set_trace_info_for_connection`.
- **tpinjector** — sk_msg / sockops on the egress path. Reads from
  `outgoing_trace_map` to inject a `traceparent` into outgoing requests.

The two maps are distinct and do not interfere. In the proxy scenario, the scanner
extracts the incoming trace ID, OBI creates a child span, and the injector propagates
it on egress — standard W3C Trace Context behaviour, unchanged by this PR.

---

## Limitations (out of scope for this PR)

| Case | Reason |
| --- | --- |
| HTTP/2 | Headers are HPACK-compressed; `traceparent` does not appear as plaintext |
| HTTP/3 / QUIC | Transport-layer encryption |
| TLS without userspace offloading | Socket buffer is ciphertext at the kprobe |
| Multi-segment TCP (traceparent in a later recv call) | Scanner covers one `recvmsg` call; if `traceparent` arrives in a separate segment OBI does not see it |
| Scatter-gather I/O, kernel 5.17–5.x | Chunked rescan requires a single-segment buffer (`nr_segs == 1`); does not apply on kernel ≥ 6.0 |
| `traceparent` beyond configured budget | By design — bounded to cap per-request CPU cost. Configurable up to 27 KB. |

---

## Performance

Load tests run against openresty in a local kind cluster (kernel 6.17,
`kubectl port-forward`, 100 connections, 30 s). Absolute latency includes
port-forward overhead; the relative differences between scenarios are what matter.

### oha results

Kernel 6.17.0-19-generic, kind cluster, `kubectl port-forward`, oha 1.14.0,
100 connections, 30 s, `OTEL_EBPF_BPF_MAX_REQUEST_TP_PARSE_SIZE_KB=27`.

| Scenario | req size | p50 | p95 | p99 | req/sec | avg | success |
| --- | --- | --- | --- | --- | --- | --- | --- |
| main baseline | ~200 B | 7.9 ms | 14.7 ms | 19.6 ms | 11,617 | 8.6 ms | 100% |
| feat: baseline (scanner not triggered) | ~200 B | 7.4 ms | 14.2 ms | 18.7 ms | 12,338 | 8.1 ms | 100% |
| feat: chunked (2.5 KB filler, ~3 chunks) | ~2.6 KB | 20.2 ms | 30.4 ms | 37.0 ms | 4,636 | 21.6 ms | 100% |
| feat: boundary (traceparent at byte 956) | ~1.1 KB | 15.4 ms | 22.5 ms | 28.7 ms | 6,114 | 16.4 ms | 100% |

`main` and `feat: baseline` are equivalent. The higher latency in the chunked and
boundary scenarios tracks the larger request sizes, not BPF overhead.

### BPF program runtime (bpf_stats)

`obi_parse_traceparent_http` (prog 14) and `obi_parse_traceparent_http_append`
(prog 15) only run on requests where `traceparent` was not found in the first ~1 KB.
Their average runtime per invocation is measured below.

dashboard

dashboard

dashboard

When the scanner does not trigger, the only added cost is the guard check in
`__obi_continue_protocol_http_tp`. When it activates, the cost is one
`bpf_probe_read_user` of 1024 bytes per 956-byte step.

---

## Test environments

All tests were run with OBI deployed as a DaemonSet in a local single-node kind
cluster, instrumenting an openresty application.

| Kernel | Distribution | Chunked scanner | `TestTraceparentExtraction` |
| --- | --- | --- | --- |
| 5.15.0-91-generic | Ubuntu 22.04.3 LTS | disabled (< 5.17, legacy path) | 3/5 PASS |
| 5.19.17-051917-generic | Ubuntu 22.04.3 LTS | enabled | 5/5 PASS |
| 6.17.0-19-generic | Ubuntu 24.04.4 LTS | enabled | 5/5 PASS |

On 5.15, `FixupSpec` swaps `obi_continue_protocol_http` and
`obi_parse_traceparent_http{,_append}` for stubs. The three baseline tests pass; the
two chunked tests fail as expected.

## Validation


grcevski:
This PR contains number of unrelated changes, if you want to implement a fix for this scenario the fix needs to be scoped to the smallest change possible. Also, the PR doesn't provide a test which we can use to prove the changes are good.

You've provided an example with large payload, but not an example where the traceparent can be split between chunks.

rafaelroquetto:
On top of what @grcevski said, whilst we are not against the usage of AI, I encourage you to review our contributing guidelines in relation to code ownership and understanding.

I will highlight the 2 important points here:

> Contributors must review, test, and understand all changes before submitting a PR. This applies equally to manually written and tool-generated code. Use of AI or other tools does not transfer responsibility — the contributor is fully accountable for the final patch.

and 

> ### Code Ownership
>
>AI tools are permitted, but they do not change what is expected of contributors. Every line in a PR is your responsibility, regardless of how it was produced. If you cannot explain a change, do not submit it.
>
>Reviewers will ask you to walk through your changes. Inability to explain the rationale, the approach, or the details of any part of a PR is grounds for rejection. This includes changes generated or suggested by AI tools.
>
>**Avoid reimplementing existing code.** AI tools frequently generate new implementations of functionality that already exists in the codebase. Before introducing any new utility, helper, abstraction, or pattern, search the codebase first. Reviewers will reject code that duplicates existing functionality.
>
>**Vet AI-generated plans and issue reports before filing.** If you use an AI tool to draft an issue, design proposal, or implementation plan, read it critically before submitting. Check that it accurately reflects the codebase, does not contradict existing architecture, and does not propose work that is already done. Unvetted AI output creates noise and wastes reviewer time.

This PR description, the related issue, and the description of other related PRs are all AI generated. The code is mostly AI generated as well - but it does not follow the guidelines set by `AGENTS.md` - I encourage you to point your AI agent there - and also at `copilot-instructions.md`, `ebpf.instructions.md` and friends.

In particular, I'd highly encourage you to actually handcraft the PR descriptions, as that help convey the changes being proposed are well-understood - otherwise, it gets really difficult for us on the reviewing end to establish whether we are reviewing a pre-vetted code by the author, or just some AI slop.

bpf/generictracer/protocol_http.h
```diff
@@ -365,6 +385,10 @@ static __always_inline int http_send_large_buffer(http_info_t *req,
     large_buf->conn_info = req->conn_info;
     large_buf->action = action;
     large_buf->tp = req->tp;
+    // Use the stable original trace ID for large buffers so userspace can correlate
+    // chunks even if the request's trace ID is later updated by a chunked traceparent.
```
rafaelroquetto: can you illustrate what's going on here?

bpf/generictracer/protocol_http.h
```diff
@@ -398,6 +422,14 @@ static __always_inline int __obi_continue2_protocol_http(struct pt_regs *ctx,
         tp_info_pid_t *tp_p = trace_info_for_connection(&args->pid_conn.conn, type);
         if (tp_p) {
             info->tp = tp_p->tp;
+            const u64 *orig_id = (const u64 *)info->original_trace_id;
```
rafaelroquetto: also this, it seems relatred, but it would be helpful to provide some context on the actual mechanism behind this - I don't undersdtand the concept of `original_trace_id` - this seems off.

bpf/common/http_types.h
```diff
@@ -51,13 +51,26 @@ typedef struct call_protocol_args {
     protocol_selector_t protocols;
     u8 skip_tp_parsing;
     u8 use_bpf_loop;
-    u8 pad[1];
+    u8 pad[5];
```
rafaelroquetto: the padding here is wrong and chaotic

rafaelroquetto (changes requested):
Aside from my previous comment, I gave this a real look and I can't review it fairly in its current state. the scope and structure need to be addressed first.

`obi_parse_traceparent_http` alone is a good example of what I mean (it's overly complicated, there's a flag controlling I guess 5 different branches, etc...).

smehboub:
> This PR contains number of unrelated changes, if you want to implement a fix for this scenario the fix needs to be scoped to the smallest change possible. Also, the PR doesn't provide a test which we can use to prove the changes are good.
> 
> You've provided an example with large payload, but not an example where the traceparent can be split between chunks.

Thanks for the feedback!

I agree that the scope is too broad. I'm working on splitting the PR into smaller, focused changes.

As for the test, I added ̀`testWithTraceparentAtChunkBoundary` to target byte 956, which is where the eBPF scanner transitions from chunk 0 to 1.

bpf/common/http_types.h
smehboub: Thanks for the correction! That pad was a bit clumsy. I restored the original pad and cleaned up the remaining padding.

smehboub:
Hi @rafaelroquetto @grcevski

Sorry, the PR description was not clear enough about what `original_trace_id` was actually doing.

`original_trace_id` was a workaround for a map key mismatch that only happens when `OTEL_EBPF_BPF_BUFFER_SIZE_HTTP > 0` (`http_max_captured_bytes`).

When the large-buffer path is active, `http_send_large_buffer` writes body chunks into a Go map keyed on `largeBufferKey{traceID, ...}`, using `tp.trace_id` as it exists at the time the chunk is sent to the ring buffer. Later, when the chunked scanner finds the traceparent in the HTTP headers, it overwrites `tp.trace_id` with the client's value. By the time the final HTTP event arrives on the Go side, `event.Tp.TraceId` no longer matches the key that was used to store the chunks, so the lookup fails and OBI logs:

```
DEBUG missing large buffer for HTTP request traceID=<client_trace_id> ...
```

`original_trace_id` was a stable copy of `tp.trace_id` captured before the scanner ran, so both the BPF write and the Go lookup used the same key. It never affected which trace ID was propagated because `tp.trace_id` was always overwritten unconditionally at decode_hex(tp_p->tp.trace_id, ...) in both paths.

I hit this because I had `OTEL_EBPF_BPF_BUFFER_SIZE_HTTP` set in my test environment, which is why I introduced it.

Since this path is only reachable when that variable is explicitly configured (disabled by default), I removed `original_trace_id` entirely to keep this PR focused.

bpf/generictracer/protocol_http.h
smehboub: I answered in this comment: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1988#issuecomment-4480759173

bpf/generictracer/protocol_http.h
smehboub: I answered in this comment: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1988#issuecomment-4480759173

smehboub:
> Aside from my previous comment, I gave this a real look and I can't review it fairly in its current state. the scope and structure need to be addressed first.
> 
> `obi_parse_traceparent_http` alone is a good example of what I mean (it's overly complicated, there's a flag controlling I guess 5 different branches, etc...).

`obi_parse_traceparent_http` is now split into two programs (index 14 for the initial-request path, index 15 for the append path), each with a single tp source and no branches on mode. Both delegate to `__tp_chunk_scan`.

rafaelroquetto:
Hi @smehboub, just before I dive into it, would you mind explaining to me how exactly this works (please no AI generated prompts)?

(One feedback I can give you upfront - the comments are too verbose - please see our guidelines and the agents instructions).

smehboub:
> Hi @smehboub, just before I dive into it, would you mind explaining to me how exactly this works (please no AI generated prompts)?
> 
> (One feedback I can give you upfront - the comments are too verbose - please see our guidelines and the agents instructions).

Hi @rafaelroquetto, I updated the description to make it clearer with these explanations and removed the obsolete comments.

rafaelroquetto:
@smehboub I saw the updated description, but it is still not enough for reviewing this change.

The main issue is that maintainers still have to reverse-engineer the state machine, memory model, and tradeoffs from the diff before we can even start reviewing the implementation. This PR is large and touches core HTTP parsing logic, so the design needs to be clear upfront.

What is missing is the actual design explanation. The description says the request is scanned in "successive windows" and mentions the new limit, but it does not explain the mechanism clearly enough: what data is scanned, where it comes from, when it is copied, what state is carried between scans, and which boundary cases this actually handles.

For example, "split across windows" can mean several different things here: scan chunks, BPF scratch buffers, application buffers, send/recv calls, TCP segments, or the large-buffer append path. These are not equivalent, and the PR needs to say which cases are handled and which are not.

Please also explain the performance side. If this copies request data while scanning, describe from where to where, how often, and whether it was profiled. This is a hot path.

Finally, I'd also like to understand how this relates to the existing traceparent handling in the tpinjector path. I didn't see that being touched.

The goal here is not to add process for its own sake. We need the high-level design before line-by-line review.

smehboub:
> @smehboub I saw the updated description, but it is still not enough for reviewing this change.
> 
> The main issue is that maintainers still have to reverse-engineer the state machine, memory model, and tradeoffs from the diff before we can even start reviewing the implementation. This PR is large and touches core HTTP parsing logic, so the design needs to be clear upfront.
> 
> What is missing is the actual design explanation. The description says the request is scanned in "successive windows" and mentions the new limit, but it does not explain the mechanism clearly enough: what data is scanned, where it comes from, when it is copied, what state is carried between scans, and which boundary cases this actually handles.
> 
> For example, "split across windows" can mean several different things here: scan chunks, BPF scratch buffers, application buffers, send/recv calls, TCP segments, or the large-buffer append path. These are not equivalent, and the PR needs to say which cases are handled and which are not.
> 
> Please also explain the performance side. If this copies request data while scanning, describe from where to where, how often, and whether it was profiled. This is a hot path.
> 
> Finally, I'd also like to understand how this relates to the existing traceparent handling in the tpinjector path. I didn't see that being touched.
> 
> The goal here is not to add process for its own sake. We need the high-level design before line-by-line review.

Thank you very much for this feedback @rafaelroquetto. The DoD and your expectations are much clearer to me now. 

I completely agree that a high-level design explanation is necessary before a line-by-line review, especially for a hot path touching core HTTP parsing. 

I am going to convert this PR into a draft for now (I have a few personal constraints over the next few days). I will put in the necessary effort to address all your points (the state machine, memory model, performance impact, and the tpinjector relationship).

Thanks again for guiding me through the process!

rafaelroquetto:
@smehboub thank you! Just make sure you don't spend time/effort implementing something before we have the chance to discuss and vet it. Because you are touching a core part of the code, it's important that we are all on the same page about the approach chosen. It's best to hash out the high-level details now, and assess the feasibility of what is being proposed, than you spending time on something that may not take off.

Feel free to ask any questions here - happy to answer them.

marctc:
@smehboub I didn't review the code either, but I read the title and clicked thru the issue you are trying to fix. I'm not aware of this piece of the code, but to me feels that solve the problem i'd great to see in the original issue different proposals on how to fix this, pros/cons and high level detail of each, rather than creating this gigantic PR, hard to reason for anyone

smehboub:
> @smehboub I didn't review the code either, but I read the title and clicked thru the issue you are trying to fix. I'm not aware of this piece of the code, but to me feels that solve the problem i'd great to see in the original issue different proposals on how to fix this, pros/cons and high level detail of each, rather than creating this gigantic PR, hard to reason for anyone

Hi @marctc,

Absolutely, that makes complete sense. I would rather we get the design right than rush something into a core path.

To give you a bit of context on where I'm coming from: I work in an e-commerce business unit that does end-to-end tracing, from frontends down to data services, using higher-level libraries and agents. I see a lot of potential in OBI precisely because of its language-agnostic nature. The problem is that on the frontend side, headers tend to be large by default (partner cookies, auth tokens, etc.), which means this limitation effectively blocks end-to-end tracing for our use cases. That's what motivated me to file the bug/limitation in the first place.

I should also be transparent: I'm exploring this on my own time, out of personal interest and curiosity. I'm not paid or mandated by my employer to work on this. My only real goal is that this use case gets covered, regardless of whether this particular implementation is the one that lands.

I will work on putting together a proper design doc covering the state machine, memory model, performance side and the tpinjector relationship before touching any more code. If there are alternative approaches you think are worth considering before I go further, I'm very open to hearing them.

Thanks in advance.

smehboub:
> @smehboub I saw the updated description, but it is still not enough for reviewing this change.
> 
> The main issue is that maintainers still have to reverse-engineer the state machine, memory model, and tradeoffs from the diff before we can even start reviewing the implementation. This PR is large and touches core HTTP parsing logic, so the design needs to be clear upfront.
> 
> What is missing is the actual design explanation. The description says the request is scanned in "successive windows" and mentions the new limit, but it does not explain the mechanism clearly enough: what data is scanned, where it comes from, when it is copied, what state is carried between scans, and which boundary cases this actually handles.
> 
> For example, "split across windows" can mean several different things here: scan chunks, BPF scratch buffers, application buffers, send/recv calls, TCP segments, or the large-buffer append path. These are not equivalent, and the PR needs to say which cases are handled and which are not.
> 
> Please also explain the performance side. If this copies request data while scanning, describe from where to where, how often, and whether it was profiled. This is a hot path.
> 
> Finally, I'd also like to understand how this relates to the existing traceparent handling in the tpinjector path. I didn't see that being touched.
> 
> The goal here is not to add process for its own sake. We need the high-level design before line-by-line review.

Hi @rafaelroquetto

I updated the PR description to focus on the design first, before another line-by-line review.

It now explains the scan path end to end: what buffer is being scanned, how bytes are copied, what state is preserved across tail calls, which cases are handled versus out of scope, how the append path fits in, and how this interacts with the state machine, memory model, performance characteristics, and the existing tpinjector path.

I also tightened the wording around older kernels and scatter-gather I/O so it matches the current code, tests, and measured behavior.

If there is a part you would like me to make more explicit before code-level review, I’m happy to do that.

Thanks in advance.

smehboub:
Hi @rafaelroquetto,

Just a gentle ping on this one. I wanted to make sure the updated design documentation addresses your concerns.

Thanks in advance!

bpf/generictracer/k_tracer.c
```diff
@@ -989,6 +991,18 @@ static __always_inline int return_recvmsg(void *ctx, struct sock *in_sock, u64 i
             goto done;
         }
 
+        if (bpf_core_enum_value_exists(enum iter_type___dummy, ITER_UBUF) &&
```
rafaelroquetto, line 994: this does not look right - see read_iovec_ctx below, this seems to be duplicating the code (and reading it again).

Why have you decided to read only the first segment?

Also, passing the iovec as an u_buf seems wrong or misleading at best

bpf/generictracer/k_tracer.c
```diff
@@ -1042,8 +1056,16 @@ static __always_inline int return_recvmsg(void *ctx, struct sock *in_sock, u64 i
...
-                handle_buf_with_connection(
-                    ctx, &info, buf, copied_len, NO_SSL, TCP_RECV, orig_dport);
+                handle_buf_with_connection_ext(ctx,
+                                               &info,
+                                               buf,
+                                               copied_len,
+                                               NO_SSL,
+                                               TCP_RECV,
+                                               orig_dport,
+                                               orig_ubuf,
```
rafaelroquetto, line 1066: what's the difference between the current iovec ctx and this orig_ubuf?

bpf/generictracer/k_tracer_defs.h
```diff
@@ -60,6 +60,10 @@ static __always_inline call_protocol_args_t *make_protocol_args(const pid_connec
     args->u_buf = (u64)u_buf;
     args->lw_thread = lw_thread;
     args->protocols = protocols;
+    args->orig_buf = 0;
+    args->full_bytes_len = 0;
+    args->niter = 0;
+    args->u_buf_is_user = 0;
```
rafaelroquetto, line 66: u_buf is always user - the fact that is appears to be doubling as an iovec suggests there is some misunderstanding with the design - I suggest you revisit `read_iovec_ctx` - I would expect that we don't need to re-read it, let alone reinvent it. If there is a genuine reason, though, feel free to lay it out.

devdocs/config/config-schema.json
```diff
@@ -717,6 +717,14 @@
...
...stops as soon as it detects the end-of-headers boundary (CRLFCRLF) within a chunk, so it does not read into the request body within a single received buffer. For multi-segment TCP requests where headers and body arrive in separate kernel receive calls, OBI cannot distinguish a body-only segment from a header segment; in that case the body is not scanned because the traceparent has already been found or the headers-complete event was observed in an earlier segment.  Minimum is 4 KB because the chunked scanner starts at the second chunk (niter=1, offset=956 bytes): the first ~1 KB is already scanned in the initial BPF pass. A budget below ~1 KB would never trigger a second chunk and the setting would have no effect. 4 KB provides at least three additional 956-byte windows and covers the common case of a single large header (e.g. a 2-3 KB auth token or cookie) before traceparent. Maximum is 27 KB to stay within the tail-call budget (tightest path: 3 preceding TCs + 29 invocations of prog 14 = 32 = BPF_MAX_TAIL_CALL_CNT; at niter=28, offset=28×956=26,768; budget cap of 27×1024=27,648 limits the last read to 880 bytes; next chunk at 29×956=27,724 exceeds the budget and is never read).",
```
rafaelroquetto, line 724: this is a bit excessive

rafaelroquetto (changes requested):

rafaelroquetto:
@smehboub I truly appreciate the effort and all the month you've put in - so please don't get me wrong, I am fully aware this isn't about willingness, but there are some fundamental gaps here that keep resurfacing - too many fundamental issues that suggest there's no fully understanding of how this code works. I understand our eBPF code is dense and not straightfoward, but the kind of clarifications we'd need to discuss here overflow the role of a pull request and of a maintainer/reviewer. Therefore, I am going to close it.

We have a SIG that happens every Wednesday, you are more than welcome to join. Some of us have floated the idea of office hours for those contributors willing to learn more about the OBI codebase - there's not anything official yet but I am sure we could set up some time aside if you are interested.

A few issues that I have eyeballed:
- The orig_buf / u_buf path reinvents read_iovec_ctx and reads the buffer a second time -  u_buf is already normalized for you. This machinery also does nothing at the default config
- The two scanner programs handle the trace differently (tp_info_mem() vs trace_info_for_connection()) which is a correctness risk, not a style nit
- Unexplained scope creep (e.g. the gotracer.go uretprobe change) unrelated to the feature
- The boundary test doesn't actually assert the chunked path was taken, it seems to only cover the happy path
- The PR description and config docs are walls of generated text that don't match the code (fallback window, index numbers) - these comments do more harm than good and make it really hard for the reader

So yeah, my suggestion is for you to step back and really try to understand the machinery behind the codepath taken from the kprobe entrypoint (e.g. `tcp_recvmsg`), the arguments setup -> the retprobe -> buffer handling. Try to follow the HTTP1 path only, for starters. Hopefully that will make more sense.

Again, thanks for all the effort and feel free to reach out.

PR #1989, opened 2026-05-04
test(tracing): add TLS and Go uprobe integration tests for large-header traceparent

### LINKED ISSUE #1381: Bug: traceparent header not detected when located beyond the eBPF capture buffer (large headers offset)

smehboub:
# test(tracing): add TLS and Go uprobe integration tests for large-header traceparent

## Summary

Extends `TestTraceparentExtraction` with two sub-tests that cover the chunked
traceparent scanner on code paths not exercised by the HTTP/kprobe sub-test
added in the preceding PR.

Both sub-tests use distinct static trace IDs to prevent cross-contamination:
Jaeger may return a trace by ID without filtering by service, so sharing the
same trace ID across sub-tests risks surfacing stale spans from a previous run.

### `testWithHugeHeadersTraceparentTLS` — port 7000 (HTTPS)

- Sends a 2500-byte `big-filler` header before `Traceparent` over a TLS
  connection to the existing `tpclient-a/b/c` Node.js chain.
- Validates the SSL uprobe path (`obi_uretprobe_ssl_read` family) feeds the
  chunked scanner correctly.
- Uses `staticHugeTLSTraceID` as the trace identifier.

### `testWithHugeHeadersTraceparentGoUprobe` — port 8000 (HTTP, Go services)

- Sends a 2500-byte `big-filler` header to the new `tpclient-go-a/b/c` chain,
  which are plain Go `net/http` servers.
- Validates the `readContinuedLineSlice` uretprobe path (gotracer), which fires
  per header line — a distinct instrumentation path from the kprobe-based Node
  services in the other sub-tests.
- Uses `staticGoTraceID` as the trace identifier.

## Infrastructure

| File | Change |
|---|---|
| `tpclient/service.js` | Add HTTPS server (port + 1000), `/with-huge-tp-tls` endpoint, `httpsAgent`/`axiosInstance` for TLS-capable chain calls |
| `tpclient/Dockerfile` | Install `openssl`, generate self-signed cert, expose TLS ports 7000–7002 |
| `docker-compose-tpclient.yml` | Add TLS ports to tpclient; add `tpclient-go` service (ports 8000–8002); set `OTEL_EBPF_BPF_MAX_REQUEST_TP_PARSE_SIZE_KB` |
| `configs/obi-config-tpclient.yml` | Add `tpclient-go-a/b/c` service discovery entries |
| `tpclient-go/` | New Go chain service mirroring the Node.js tpclient for the gotracer path |

## Testing

```
go test -v -run TestTraceparentExtraction/with_huge_headers_traceparent_tls -timeout 10m ./internal/test/integration/
go test -v -run TestTraceparentExtraction/with_huge_headers_traceparent_go_uprobe -timeout 10m ./internal/test/integration/
```

## Depends on

- fix(tracing): correct trace-id not-found sentinel and use Copy() for dummy BPF program specs
- refactor(ebpf): extend HTTP request and protocol-args structs for chunked traceparent scanner
- feat(tracing): add chunked tail-call traceparent scanner for large HTTP headers (fixes #1381)


rafaelroquetto (changes requested):
This needs to be split and reworked - see my comments on https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1988

PR #1990, opened 2026-05-04
pkg/docker: Add event-based Docker container info caching

florianl:
## Summary

Implement in-process caching of Docker container metadata in ContainerStore with event-driven invalidation. When containers die or are destroyed, Docker events trigger immediate cache eviction, eliminating stale metadata and reducing repeated API calls.

This reduces CPU consumption in span enrichment pipeline by avoiding redundant Docker API calls. Flamegraph showed ~13.4% CPU in dockerEnricher; caching eliminates most calls for containers that host multiple processes.

## Validation


grcevski (approved):
LGTM! Pretty cool! I'll let @mariomac review this since I'm only partially familiar with this code, but looks great from my side.

pkg/docker/docker_api_client.go
```diff
@@ -43,14 +53,21 @@ type ContainerMeta struct {
...
+
+	cacheMu sync.RWMutex
+	byPID   map[app.PID]ContainerMeta
+	byID    map[string][]app.PID
 }
 
 func NewStore() *ContainerStore {
 	return &ContainerStore{
-		log: cmlog(),
+		log:   cmlog(),
```
MrAlias: This still misses the cache entry in the normal case. `byID` is populated from `meta.ID`, which is the shortened 12-character ID, but Docker events report the full container ID in `msg.Actor.ID`.

In practice that means `die` / `destroy` events won't evict anything, so stale metadata stays cached. I think the internal cache key needs to stay as the full ID, and only the exported `ContainerMeta.ID` should remain abbreviated if we still want the short form in attributes.

pkg/docker/docker_api_client_test.go
```diff
@@ -0,0 +1,124 @@
...
+func TestContainerStoreInvalidation(t *testing.T) {
+	t.Run("invalidation_clears_all_pids_for_container", func(t *testing.T) {
+		store := NewStore()
+		pid1 := app.PID(1111)
+		pid2 := app.PID(2222)
+		containerID := "abc123def456"
+
+		// Manually populate cache to simulate a successful ContainerInfo call
+		meta := ContainerMeta{
+			ID:             containerID[:12],
```
MrAlias: This test data is reinforcing the same assumption as the production bug: it uses a 12-character container ID, so `invalidateContainer(meta.ID)` matches even though real Docker events deliver the full actor ID.

I'd prefer this test to use a full-length container ID and exercise the actual event/invalidation path, so it fails if the cache is keyed by the abbreviated ID instead of the full one.

MrAlias (dismissed):
The caching direction makes sense, but I found one correctness issue in the invalidation path that looks blocking.

`ContainerStore` stores cache entries in `byID` using the shortened 12-character container ID derived from `inspectInfo.ID`, but the Docker event stream delivers the full container ID in `msg.Actor.ID`. In the current code that means the invalidation lookup never finds the cached entry, so `die` / `destroy` events do not actually evict stale metadata.

I think the fix should be to key invalidation with the full container ID internally, and only keep the 12-character form for exported metadata if that abbreviated form is still desired.

mariomac:
Nice! Thank you!

In principle LGTM! But I have a question: are the performance tests taken in production/dev scenarios, or are they micro-benchmarks ran with `go test -bench`?

I would expect that a cache at this level is not needed because the `docker_decorator.go` file already has a `containerByPID` cache that is populated when the container data is found, and removed when an `EventDeleted` event is received. Also keeping the cache there removes the need for any mutex.

pkg/docker/docker_api_client.go
florianl: sorry for the delay, as I was off-desk last week. I have updated the PR to always use the full ID.

florianl:
> But I have a question: are the performance tests taken in production/dev scenarios, or are they micro-benchmarks ran with go test -bench?

I noticed this issue, when experimenting with OBI and ebpf-profiler in the OTel demo using something like this config:

```
receivers:
  profiling:
    obi_process_ctx: true
    samples_per_second: 73
    off_cpu_threshold: 0.1
  obi:
    open_port: "1-65535"
```

So it's not a micro-benchmark but more a learning from a demo environment.

pkg/docker/docker_api_client.go
MrAlias: d516805 fixes the cache-key part of my earlier comment: `byID` is now using the full Docker ID, so `die` / `destroy` events can invalidate the cached entries correctly.

I wasn’t intending to change the emitted `ContainerMeta.ID` as part of that, though. The earlier comment was about using the full ID internally for invalidation; the exported value has historically stayed in the 12-character abbreviated form.

I’d keep the abbreviated exported value here and use the full ID only for the internal cache key. As written, this also changes the 12-character expectations in `internal/test/integration/container_meta_suites_test.go`.

pkg/docker/docker_api_client.go
```diff
@@ -167,3 +193,69 @@ func ContainerMetadata[T ~string](dst map[T]string, ci *ContainerMeta, stringer
...
+	}
+}
+
+func (s *ContainerStore) invalidateContainer(containerID string) {
+	s.cacheMu.Lock()
+	defer s.cacheMu.Unlock()
+	for _, pid := range s.byID[containerID] {
+		delete(s.byPID, pid)
+	}
+	delete(s.byID, containerID)
```
MrAlias: This shared cache now owns PID entries beyond the lifetime of the decorator-local caches, but it is only evicted on container `die` / `destroy` events. If a process exits and the kernel later reuses that PID for a different container before the old entry is removed, `ContainerInfo()` will return the stale `byPID` entry without re-reading `/proc`.

I think the store needs a PID-scoped eviction helper as well, and then the existing process-lifecycle handlers should call it from `pkg/appolly/discover/docker_decorator.go` on `EventDeleted` and `pkg/transform/transform_docker.go` on `ProcessEventTerminated` so `byPID` and `byID` stay in sync.

```suggestion
func (s *ContainerStore) invalidatePID(pid app.PID) {
	s.cacheMu.Lock()
	defer s.cacheMu.Unlock()

	meta, ok := s.byPID[pid]
	if !ok {
		return
	}
	delete(s.byPID, pid)

	pids := s.byID[meta.ID][:0]
	for _, cachedPID := range s.byID[meta.ID] {
		if cachedPID != pid {
			pids = append(pids, cachedPID)
		}
	}

	if len(pids) == 0 {
		delete(s.byID, meta.ID)
		return
	}
	s.byID[meta.ID] = pids
}

func (s *ContainerStore) invalidateContainer(containerID string) {
	s.cacheMu.Lock()
	defer s.cacheMu.Unlock()
	for _, pid := range s.byID[containerID] {
		delete(s.byPID, pid)
	}
	delete(s.byID, containerID)
}
```

MrAlias (commented):
One follow-up correctness concern on the shared Docker metadata cache: the new store-level PID cache can outlive the process lifecycle unless it also gets explicit PID-scoped eviction. The container-event invalidation path makes sense, but I think the store still needs a helper for process-deletion paths so a reused PID cannot pick up stale metadata from a previous container. After that, the existing lifecycle handlers in `pkg/appolly/discover/docker_decorator.go` on `EventDeleted` and `pkg/transform/transform_docker.go` on `ProcessEventTerminated` should call it so the shared cache stays aligned with PID lifetime.

pkg/docker/docker_api_client_test.go
```diff
@@ -0,0 +1,124 @@
...
+			mockDockerClientBase: &mockDockerClientBase{
+				eventsChan: eventsChan,
+				errsChan:   errsChan,
+			},
+		}
+		store.docker = mockClient.mockDockerClientBase
+
+		// Verify store can be initialized with mocked client
+		assert.NotNil(t, store.docker)
+	})
```
MrAlias: Can we add a regression test for the PID-lifecycle part too? The current tests cover container-ID invalidation, but they won't fail if a dead PID remains cached in `byPID` and later gets reused by another process.

Something like this would exercise the new helper directly while keeping the existing event-stream smoke test.

```suggestion
	t.Run("pid_invalidation_removes_pid_from_both_indexes", func(t *testing.T) {
		store := NewStore()
		pid1 := app.PID(1111)
		pid2 := app.PID(2222)
		containerID := "abc123def456789abc123def456789abc123def456789abc123def456789abcdef"

		meta := ContainerMeta{
			ID:   containerID,
			Name: "test-container",
		}

		store.cacheMu.Lock()
		store.byPID[pid1] = meta
		store.byPID[pid2] = meta
		store.byID[containerID] = []app.PID{pid1, pid2}
		store.cacheMu.Unlock()

		store.invalidatePID(pid1)

		store.cacheMu.RLock()
		_, ok1 := store.byPID[pid1]
		_, ok2 := store.byPID[pid2]
		pids := store.byID[containerID]
		store.cacheMu.RUnlock()

		assert.False(t, ok1)
		assert.True(t, ok2)
		assert.Equal(t, []app.PID{pid2}, pids)
	})

	t.Run("event_stream_can_be_created", func(t *testing.T) {
		store := NewStore()
		eventsChan := make(chan events.Message, 1)
		errsChan := make(chan error, 1)
		errsChan <- io.EOF

		mockClient := &struct {
			*mockDockerClientBase
		}{
			mockDockerClientBase: &mockDockerClientBase{
				eventsChan: eventsChan,
				errsChan:   errsChan,
			},
		}
		store.docker = mockClient.mockDockerClientBase

		assert.NotNil(t, store.docker)
	})
```

pkg/docker/docker_api_client.go
```diff
@@ -43,14 +53,21 @@ type ContainerMeta struct {
...
-	docker    client.ContainerAPIClient
-	log       *slog.Logger
+	initMutex      sync.Mutex
+	docker         dockerClient
+	log            *slog.Logger
+	watcherStarted sync.Once
+
+	cacheMu sync.RWMutex
+	byPID   map[app.PID]ContainerMeta
+	byID    map[string][]app.PID
```
MrAlias: I think `ContainerStore` still needs one more internal index here so PID-scoped eviction can recover the full Docker ID without relying on the exported `ContainerMeta.ID`.

Concretely, I’d keep the 12-character ID in `ContainerMeta` for emitted attributes and add a separate internal mapping for cache bookkeeping. A `pid -> fullID` map is probably the smallest change here.

```suggestion
	cacheMu sync.RWMutex
	byPID   map[app.PID]ContainerMeta
	pidToID map[app.PID]string
	byID    map[string][]app.PID
```

pkg/docker/docker_api_client.go
```diff
@@ -167,3 +198,93 @@ func ContainerMetadata[T ~string](dst map[T]string, ci *ContainerMeta, stringer
...
+		if cachedPID != pid {
+			pids = append(pids, cachedPID)
+		}
+	}
+
+	if len(pids) == 0 {
+		delete(s.byID, meta.ID)
+		return
+	}
+	s.byID[meta.ID] = pids
```
MrAlias: This helper should use the internal full ID, not `meta.ID`.

Right now `byID` is keyed by the full Docker ID from `ContainerInfo()`, but this path looks it up via the exported 12-character ID. So the invalidation here is only partial: `byPID` is cleared, but the stale PID stays attached to the old container's `byID` bucket. If that PID is later reused for a different container, a `die` or `destroy` event for the old container can still evict the new container's cached metadata through `invalidateContainer()`.

If we add the extra `pid -> fullID` state above, the change here should be straightforward: read the full ID from that map, update `byID[fullID]`, and clear the PID from both indexes under the same lock.

pkg/docker/docker_api_client.go
```diff
@@ -167,3 +198,93 @@ func ContainerMetadata[T ~string](dst map[T]string, ci *ContainerMeta, stringer
...
+	for {
+		if err := s.eventsLoop(ctx, fltrs); err != nil && !errors.Is(err, context.Canceled) {
+			s.log.Debug("docker event stream error", "error", err)
+			select {
+			case <-time.After(time.Second):
+			case <-ctx.Done():
+				return
+			}
+		}
+	}
```
MrAlias, line 339: With `sync.Once` on `Start()`, this watcher only gets one chance to come up. If Docker is unavailable during startup, `watchContainerEvents()` returns here and the shared cache never gets event-driven invalidation for the rest of the OBI lifetime, even though later `IsEnabled()` calls can still enable Docker enrichment.

I think this path needs to keep retrying initialization so the watcher matches the rest of OBI's lazy Docker detection behavior.

```suggestion
func (s *ContainerStore) watchContainerEvents(ctx context.Context) {
	for {
		s.initMutex.Lock()
		s.initialize(ctx)
		docker := s.docker
		s.initMutex.Unlock()

		if docker == nil {
			select {
			case <-time.After(time.Second):
			case <-ctx.Done():
				return
			}
			continue
		}

		fltrs := make(client.Filters).
			Add("type", string(events.ContainerEventType)).
			Add("event", string(events.ActionDie), string(events.ActionDestroy))

		if err := s.eventsLoop(ctx, fltrs); err != nil && !errors.Is(err, context.Canceled) {
			s.log.Debug("docker event stream error", "error", err)
		}

		select {
		case <-time.After(time.Second):
		case <-ctx.Done():
			return
		}
	}
}
```

pkg/docker/docker_api_client_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+		meta := ContainerMeta{
+			ID:   containerID,
+			Name: "test-container",
+		}
+
+		store.cacheMu.Lock()
+		store.byPID[pid1] = meta
+		store.byPID[pid2] = meta
+		store.byID[containerID] = []app.PID{pid1, pid2}
```
MrAlias: This fixture still does not match the production cache shape. `ContainerInfo()` stores the exported `ContainerMeta.ID` as the 12-character abbreviation, while the reverse index is keyed by the full Docker ID.

Using the full ID for both here lets `InvalidatePID()` pass even if it is still indexing `byID` with the wrong key. I’d tighten the test so it fails on the real bug.

```suggestion
		meta := ContainerMeta{
			ID:   containerID[:12],
			Name: "test-container",
		}

		store.cacheMu.Lock()
		store.byPID[pid1] = meta
		store.byPID[pid2] = meta
		store.byID[containerID] = []app.PID{pid1, pid2}
		store.cacheMu.Unlock()
```

MrAlias (commented):
Two correctness issues still look unresolved on the current head, and I left one test suggestion to make the remaining PID-lifecycle bug reproducible.

pkg/docker/docker_api_client.go
mariomac: I'd add: To improve safety against someone mixing fullID and short ID in the future, I'd either:
* Remove `ID` and use only FullID. And use a `FullID[:12]` when you need the short form.
* Create a type for FullID and use it as map key, so we would prevent anyone using the short ID in the map:
```go
type ContainerID string
type ContainerMeta struct {
	ID             string // short form ID limited to abbreviationLength
	FullID       ContainerID
}
type ContainerStore struct {
	biID map[ContainerID]ContainerMeta
}
```

florianl: With https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1990/changes/2e4dbef758bd4d5b8d4909bf7286c25c966a9733#diff-7c7d0a1e7ad9f1ccf1cad293a2297fa9b55a7cd2340b22b3e1ed0658e4285daaR37-R39 I have update the struct to this to accommodate this case:
```
type ContainerMeta struct {
	// TODO: add other fields https://opentelemetry.io/docs/specs/semconv/resource/container/
	ID             string // short form ID limited to abbreviationLength
	FullID         string // Full length ID as provided by the docker API
```

florianl: Added `ContainerID` with https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1990/changes/5450ba48cb709119fa58ec438898484cfc7cc34f.

pkg/docker/docker_api_client.go
```diff
@@ -167,3 +206,101 @@ func ContainerMetadata[T ~string](dst map[T]string, ci *ContainerMeta, stringer
 	out[stringer(attr.ContainerID)] = ci.ID
```
mariomac, line 278: for data consistency, I'd remove `ID` field and replace it here by `attr.ContainerId[:abbreviationLength]`

mariomac (commented):
LGTM despite I'd still remove the `ID string` field to ensure data consistency, as it's just a substring of `FullID`

pkg/docker/docker_api_client.go
```diff
@@ -93,6 +118,13 @@ func (s *ContainerStore) initialize(ctx context.Context) {
 // ContainerInfo returns the ContainerMeta that is associated to the provided PID.
 // It also returns true if the ContainerMeta was found for the provided PID. False otherwise
 func (s *ContainerStore) ContainerInfo(ctx context.Context, pid app.PID) (ContainerMeta, bool) {
+	s.cacheMu.RLock()
+	if ci, ok := s.byPID[pid]; ok {
```
MrAlias, line 138: One remaining cache-efficiency gap: `ContainerInfo()` still only reuses entries by PID. After `osInfoForPID()` discovers that a new PID belongs to a container we have already inspected, this path still calls `ContainerInspect()` again instead of reusing metadata by full container ID.

That means a container with multiple instrumented processes still pays one Docker API call per PID, which undercuts the PR’s stated goal of avoiding repeated calls for containers hosting multiple processes. I think this should either add a fullID -> metadata cache, with PID entries linked to it for invalidation, or narrow the PR description to the PID-scoped cache behavior.

MrAlias (commented):
One remaining non-blocking cache-efficiency concern inline.

pkg/transform/transform_docker.go
```diff
@@ -121,6 +121,7 @@ func DockerProcessEventDecoratorProvider(
 					}
 				case exec.ProcessEventTerminated:
 					delete(containerByPID, ev.File.Pid())
+					containers.InvalidatePID(ev.File.Pid())
```
MrAlias, line 117: This invalidates the shared `ContainerStore` on process termination, but the span enricher above still keeps its own `dockerEnricher.containerByPID` cache. In a PID-reuse scenario, the shared store can evict the old PID while the span path keeps returning stale metadata from its local map before consulting the store.

I think the fix should make `ContainerStore` the only Docker metadata cache for this path too, for example by removing `dockerEnricher.containerByPID` and having `containerInfo()` always call `dd.docker.ContainerInfo(ctx, pid)`.

pkg/docker/docker_api_client.go
```diff
@@ -119,12 +173,20 @@ func (s *ContainerStore) ContainerInfo(ctx context.Context, pid app.PID) (Contai
...
 		ID:             containerID,
+		FullID:         ContainerID(inspectInfo.ID),
 		ComposeService: composeSvcName,
-	}, true
+	}
+
+	s.cacheMu.Lock()
+	s.byPID[pid] = meta
+	s.byContainerID[ContainerID(inspectInfo.ID)] = containerEntry{meta: meta, pids: []app.PID{pid}}
+	s.cacheMu.Unlock()
```
MrAlias, line 236: This write can lose PID links when two PIDs from the same uncached container race through `ContainerInfo()`. Both goroutines can miss `byContainerID`, both call `ContainerInspect()`, and the later write replaces the first `containerEntry` with a single-PID entry. After that, a Docker `die` / `destroy` event only evicts whichever PID survived in `byContainerID`, leaving stale metadata in `byPID` for the other PID.

I think the fix should merge with any entry that appeared while `ContainerInspect()` was in flight instead of overwriting it.

```suggestion
	s.cacheMu.Lock()
	if entry, ok := s.byContainerID[meta.FullID]; ok {
		meta = entry.meta
		seen := false
		for _, cachedPID := range entry.pids {
			if cachedPID == pid {
				seen = true
				break
			}
		}
		if !seen {
			entry.pids = append(entry.pids, pid)
		}
		s.byPID[pid] = meta
		s.byContainerID[meta.FullID] = entry
		s.cacheMu.Unlock()
		return meta, true
	}
	s.byPID[pid] = meta
	s.byContainerID[meta.FullID] = containerEntry{meta: meta, pids: []app.PID{pid}}
	s.cacheMu.Unlock()
```

MrAlias (commented):
Two follow-up cache lifecycle comments inline. One is an adjacent stale-cache concern in the span path; the other is a current race in the new fullID cache write path.

pkg/transform/transform_docker.go
florianl: resolving with https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1990/changes/d00ddfaa2933c14acb538c4cb22c7a400427bbac

pkg/docker/docker_api_client.go
```diff
@@ -93,11 +125,34 @@ func (s *ContainerStore) initialize(ctx context.Context) {
...
+		s.cacheMu.RUnlock()
+		s.cacheMu.Lock()
+		s.byPID[pid] = meta
+		entry = s.byContainerID[ContainerID(osCntInfo.ContainerID)]
+		entry.pids = append(entry.pids, pid)
+		s.byContainerID[ContainerID(osCntInfo.ContainerID)] = entry
+		s.cacheMu.Unlock()
+		return meta, true
+	}
+	s.cacheMu.RUnlock()
```
MrAlias: This fast path still publishes a new `byPID` entry after checking container membership outside the write lock. If the PID exits, or a Docker event invalidates this container, between the `/proc` read above and this write, `InvalidatePID()` can be a no-op and this path can cache metadata for a dead or reused PID. There is also a smaller consistency bug here: after dropping the read lock, line 148 reads `byContainerID` again without checking that the entry still exists, so a concurrent container invalidation can be overwritten with a zero-value `containerEntry`.

I think this path needs to re-check the PID/container relationship while holding `cacheMu`, before writing either index. If holding the cache lock across `osInfoForPID` is too broad, an equivalent generation/tombstone check would work too; the important part is ordering validation and cache writes against invalidation.

```suggestion
	fullContainerID := ContainerID(osCntInfo.ContainerID)
	s.cacheMu.Lock()
	if entry, ok := s.byContainerID[fullContainerID]; ok {
		currentInfo, err := osInfoForPID(pid)
		if err != nil || ContainerID(currentInfo.ContainerID) != fullContainerID {
			s.cacheMu.Unlock()
			return ContainerMeta{}, false
		}

		meta := entry.meta
		seen := false
		for _, cachedPID := range entry.pids {
			if cachedPID == pid {
				seen = true
				break
			}
		}
		if !seen {
			entry.pids = append(entry.pids, pid)
			s.byContainerID[fullContainerID] = entry
		}
		s.byPID[pid] = meta
		s.cacheMu.Unlock()
		return meta, true
	}
	s.cacheMu.Unlock()
```

MrAlias (commented):
Sorry for the review churn here. I should have caught these cache lifecycle issues in earlier passes, and I appreciate your patience as we get this tightened up. I found two remaining races that can leave stale Docker metadata in the shared store, plus one lower-severity watcher enablement issue.

pkg/docker/docker_api_client.go
```diff
@@ -119,12 +173,37 @@ func (s *ContainerStore) ContainerInfo(ctx context.Context, pid app.PID) (Contai
...
+			entry.pids = append(entry.pids, pid)
+		}
+		s.byPID[pid] = meta
+		s.byContainerID[meta.FullID] = entry
+		s.cacheMu.Unlock()
+		return meta, true
+	}
+	s.byPID[pid] = meta
+	s.byContainerID[meta.FullID] = containerEntry{meta: meta, pids: []app.PID{pid}}
+	s.cacheMu.Unlock()
```
MrAlias, line 236: This write can cache stale metadata if the process exits while `ContainerInspect()` is in flight. In that interleaving, the process lifecycle handler calls `InvalidatePID(pid)` before `byPID[pid]` exists, so it removes nothing; then this code stores the old container metadata for that PID. If the kernel later reuses the PID, the early `byPID` lookup returns stale Docker metadata without re-reading `/proc`.

I think the store should validate that the PID still belongs to the inspected full container ID while holding `cacheMu`, so lifecycle invalidation cannot race past the cache write.

```suggestion
	s.cacheMu.Lock()
	currentInfo, err := osInfoForPID(pid)
	if err != nil || ContainerID(currentInfo.ContainerID) != meta.FullID {
		s.cacheMu.Unlock()
		return ContainerMeta{}, false
	}
	if entry, ok := s.byContainerID[meta.FullID]; ok {
		meta = entry.meta
		seen := false
		for _, cachedPID := range entry.pids {
			if cachedPID == pid {
				seen = true
				break
			}
		}
		if !seen {
			entry.pids = append(entry.pids, pid)
		}
		s.byPID[pid] = meta
		s.byContainerID[meta.FullID] = entry
		s.cacheMu.Unlock()
		return meta, true
	}
	s.byPID[pid] = meta
	s.byContainerID[meta.FullID] = containerEntry{meta: meta, pids: []app.PID{pid}}
	s.cacheMu.Unlock()
```

pkg/instrumenter/instrumenter.go
```diff
@@ -237,6 +237,7 @@ func BuildCommonContextInfo(
 	)
 
 	ctxInfo.DockerMetadata = docker.NewStore()
+	ctxInfo.DockerMetadata.Start(ctx)
```
MrAlias: Starting the Docker watcher here makes it run for every OBI process as soon as common context is built. In Kubernetes mode, or on hosts where Docker metadata is not otherwise used, `watchContainerEvents()` still retries Docker initialization once per second even though the Docker decorators later bypass themselves.

I think this should be started lazily only after Docker metadata is actually selected/enabled, or guarded by the same Docker-vs-Kubernetes enablement decision used by the decorators.

florianl: Added `K8sInformer.IsKubeEnabled()`, like it is used in other places. Hope this is fine.

florianl:
It seems like, the failing CI tests are related to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2304.

grcevski:
Hi @florianl, do you mind syncing with main, the test failures have been resolved.

pkg/instrumenter/instrumenter.go
```diff
@@ -246,6 +246,9 @@ func BuildCommonContextInfo(
 	)
 
 	ctxInfo.DockerMetadata = docker.NewStore()
+	if !ctxInfo.K8sInformer.IsKubeEnabled() {
```
MrAlias, line 249: This still misses one Docker fallback path. If Kubernetes is initially enabled here, `Start(ctx)` is skipped; later `setupKubernetes()` can fail and `ForceDisable()` the informer, after which the Docker decorators can enable Docker metadata but the event watcher was never started.

I think the watcher should start when Docker metadata is actually selected after Kubernetes setup has finalized, or be started/retried from the Kubernetes failure fallback. Otherwise the shared `ContainerStore` can cache Docker metadata without `die` / `destroy` event invalidation. Right?

pkg/docker/docker_api_client.go
```diff
@@ -167,3 +271,115 @@ func ContainerMetadata[T ~string](dst map[T]string, ci *ContainerMeta, stringer
...
+		select {
+		case <-time.After(time.Second):
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+func (s *ContainerStore) eventsLoop(ctx context.Context, fltrs client.Filters) error {
+	result := s.docker.Events(ctx, client.EventsListOptions{Filters: fltrs})
```
MrAlias: Reconnecting here starts a new real-time stream with an empty `Since`. If the Docker event stream closes or errors, `watchContainerEvents()` waits one second before calling this again, so a container can emit `die` / `destroy` during that gap and the cache never sees it. If the corresponding process-lifecycle invalidation is also missed or races, `byPID` keeps stale metadata and a reused PID is decorated as the old container.

I think the watcher should carry an event checkpoint across reconnects and set `EventsListOptions.Since` (with a small overlap, since invalidation is idempotent), or reconcile the cached container IDs before resuming the live stream. The initial subscription needs the same protection so `Start()` cannot return before the watcher can cover container exits.

MrAlias (approved):

PR #1993, opened 2026-05-04
Fix OTEL job and instance attribute keys

MrAlias:
## Summary
- map job and instance attributes to the correct OTEL resource attribute keys
- keep the rest of the attribute export path unchanged while correcting the emitted key names

## Why
The exporter was assigning job and instance values to the wrong OTEL attribute keys. That caused emitted resource attributes to use incorrect semantic-convention names even though the underlying values were present.

## Impact
This restores the expected OTEL attribute naming for job and instance resource fields, improving downstream compatibility and attribute interpretation without changing the underlying values.


grcevski (approved):
LGTM! OMG!

PR #1995, opened 2026-05-04
fix discover proc stat buffer race

MrAlias:
## Summary
- remove the shared proc stat reader buffer race from the discovery path
- preserve the existing proc stat reading behavior while making concurrent access safe

## Why
The discovery proc stat reader reused shared buffer state across calls in a way that introduced a data race risk under concurrent access. That made the proc stat parsing path depend on unsynchronized shared mutable state.

## Impact
This keeps proc stat reading behavior the same while removing the concurrent buffer-sharing risk, making the discovery path safer under parallel use.

### DIFF
--- a/pkg/appolly/discover/watcher_proc.go
+++ b/pkg/appolly/discover/watcher_proc.go
@@ -36,7 +36,16 @@ import (
 const (
 	defaultPollInterval = 5 * time.Second
 	emptyDuration       = time.Duration(0)
+	// procStatBufSize safely fits /proc/<pid>/stat for our parsing path.
+	procStatBufSize = 4096
 )
+
+// procStatBufPool amortizes stat buffer allocations across all procStatReader callers.
+var procStatBufPool = sync.Pool{
+	New: func() any {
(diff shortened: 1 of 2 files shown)


rafaelroquetto (commented):
Looks good, but is this fixing a theoretical race-condition or does it really happen?

I thought `getProcStatField` being re-entrant sufficed.

MrAlias:
@rafaelroquetto `getProcStatField` being re-entrant is not sufficient here, because the previous `ProcessAgeFunc()` implementation returned a method bound to a single `procStatReader`, and that reader reused one mutable `[4096]byte` buffer across calls. If that closure were ever used concurrently, it would race on the shared buffer.

That said, from the current call graph, the existing production code path does not appear to exercise that race today. The call sites I checked invoke it serially, so this is best described as fixing a latent goroutine-safety issue rather than an observed race in the current production path.

rafaelroquetto (approved):
Cool thanks for the clarifications!

grcevski (approved):
LGTM!

PR #1996, opened 2026-05-04
fix(appolly): make export mode enables idempotent

MrAlias:
## Summary
- make `AllowTraces`, `AllowMetrics`, and `AllowLogs` clear their block bits instead of toggling them
- keep YAML and text parsing of export modes idempotent even when entries are repeated
- add regression coverage for repeated allow calls and duplicate config entries

## Why
The export mode helpers and parsing paths used XOR to remove blocked signals. That worked for a single occurrence, but a repeated call or duplicate config entry could flip the bit back on and silently disable a signal again.

## Impact
This makes export mode enabling idempotent. Repeated helper calls or duplicate `exports` entries now keep traces, metrics, and logs enabled as intended instead of accidentally turning them back off.

## Validation

### DIFF
--- a/pkg/appolly/services/export.go
+++ b/pkg/appolly/services/export.go
@@ -99,15 +99,15 @@ func (modes ExportModes) CanExportLogs() bool {
 //  modes.AllowMetrics() // export metrics only
 
 func (modes *ExportModes) AllowTraces() {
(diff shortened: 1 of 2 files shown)


grcevski (approved):
LGTM!

rafaelroquetto (approved):

PR #1997, opened 2026-05-04
fix(ci): correct workflow line continuation

MrAlias:
## Summary
- fix the shell line continuation in the testing-image publish workflow
- restore correct image tag parsing by removing the stray trailing whitespace after the continuation backslash

## Why
A workflow line in the image-tag parsing pipeline had a trailing space after the shell continuation backslash. That breaks the continuation and causes the downstream tag parsing command chain to be interpreted incorrectly.

## Impact
This restores the intended shell pipeline for image tag parsing in the workflow without changing the surrounding publish logic.


rafaelroquetto (approved):

grcevski (approved):
LGTM!

PR #1998, opened 2026-05-04
Handle unknown protocol TCP large tcp buffers

grcevski:
## Summary

There are number of protocols we don't handle in eBPF which can benefit from large buffers. Namely, one protocol like that is FastCGI, which produces HTTP events, but in certain cases we cannot even recognize the protocol as FastCGI, because of how the headers may be laid out.

I also made a fix in the large buffer support to also key off the process ID as well as the other attributes. With FastCGI in the example, we often handle the client (nginx) and the server call, so the large buffer support would confuse the direction of the buffers, since they are overwritten by the other side. This change typically doesn't change anything for any other calls, but it will help us tell the client or server call when we instrument both sides.

Reported issue in Beyla https://github.com/grafana/beyla/issues/2526

## Validation


bpf/generictracer/protocol_common.h
```diff
@@ -84,10 +84,7 @@ static __always_inline u32 task_netns() {
     return (u32)BPF_CORE_READ(task, nsproxy, net_ns, ns.inum);
 }
 
-static __always_inline u8 infer_packet_type(u8 direction, u16 port) {
```
grcevski, line 87: small optimisation to avoid calling the is_listening again, since we know the server status in the caller.

bpf/generictracer/protocol_tcp.h
```diff
@@ -325,19 +370,19 @@ static __always_inline void handle_unknown_tcp_connection(pid_connection_info_t
...
+        if (existing->len > 0 && existing->len < (k_tcp_max_len / 2)) {
+            // Attempt to append one more packet. I couldn't convince the verifier
+            // to use a variable (k_tcp_max_len-existing->len). If needed we may need
+            // to try harder. Mainly needed for userspace detection of missed gRPC, where
+            // the protocol may sent a RST frame after we've done creating the event, so
+            // the next event has an RST frame prepended.
+            u32 off = existing->len;
+            bpf_clamp_umax(off, (k_tcp_max_len / 2));
+            bpf_probe_read(existing->buf + off, (k_tcp_max_len / 2), u_buf);
+        }
```
grcevski, line 393: I think we can clean-up one of the two, len and req_len now. I don't know why we had two.

bpf/generictracer/protocol_tcp.h
```diff
@@ -325,19 +370,19 @@ static __always_inline void handle_unknown_tcp_connection(pid_connection_info_t
...
         existing->len += bytes_len;
         existing->req_len = existing->len;
-        existing->protocol_type = protocol_type;
```
grcevski, line 339: @mmat11 I see that `protocol_type` and `is_server` were being repeatedly written, I'm not sure why so I wanted to ask if I'm going to break something here. I think switching the server/client and the type in the middle of the request can be unexpected? The large buffers would get sent with different type and we'd miss them.

internal/test/integration/components/php/nginx.conf
```diff
@@ -13,8 +13,42 @@ http {
         location / {
             fastcgi_pass php-fpm:9000;
             # fastcgi_pass unix:/run/php-fpm/www.sock
-            include fastcgi_params;
             fastcgi_index index.php;
+
+            # Tracing / correlation headers (placed first so kprobes with small buffers see them)
```
grcevski, line 18: I modified our FastCGI example to force all these headers to happen before the ones we use for actual protocol detection.

grcevski:
Obviously I broke bunch of stuff, good tests! I'm looking at the failures.

pkg/ebpf/common/tcp_large_buffer.go
```diff
@@ -12,10 +12,21 @@ import (
 	"go.opentelemetry.io/obi/pkg/internal/largebuf"
 )
 
+type largeBufferKind uint8
+
+// must match the table large_buf_kind in common.h
+const (
```
rafaelroquetto, line 18: I haven't had the chance to review this yet, but here's just an idea: my understanding is that `kind` is being used to tell apart buffers being sent from the `kprobe` layer vs those from the `uprobe` layer, avoiding them to be mixed up (otherwise you end up with junk in the case of SSL connections, for instance, as the kprobes will have seen the encrypted traffic and shipped the buffers before they had been decrypted). Is that the case?

So I am wondering if we can model this in a more protocol-agnostic way (that is easier to maintain): think "same connection" but "different layers/streams" - something like:

```

enum large_buf_layer : u8 {
    k_large_buf_layer_wire, // <--- wire format as seen by kprobes or socket ingress programs/xdp
    k_large_buf_layer_app,  // <--- as seen by the app layer, e.g. ssl uprobes or sk_msg on engress
};
```


This simplify the logic and the maintenance when adding new protocols. We could (but do not have to) make easy assumptions when looking up a buffer in userspace: can try to lookup a `k_large_buf_layer_app` buffer and fallback to `k_large_buf_layer_wire` otherwise - then knowing the buffer type upfront becomes a mere short-circuit/optimisation to avoid a potential double look-up.

As a side effect, this would also collapse `k_large_buf_kind_tcp_known` as Kafka, MySQL, Postgres etc... all come from kprobes seeing the same wire bytes, so they'd just be layer_wire along with the generic TCP case and `protocolToLargeBufferKind()` would become dead code.

bpf/generictracer/protocol_kafka.h
```diff
@@ -367,9 +367,13 @@ static __always_inline int kafka_send_large_buffer(tcp_req_t *req,
     lb->type = EVENT_TCP_LARGE_BUFFER;
     lb->packet_type = PACKET_TYPE_RESPONSE;
     lb->action = action;
+    lb->kind = k_large_buf_kind_tcp_known;
     lb->direction = direction;
     lb->conn_info = pid_conn->conn;
     lb->tp = req->tp;
+    const u64 pid_tid = bpf_get_current_pid_tgid();
+    const u32 pid = pid_from_pid_tgid(pid_tid);
+    lb->pid = pid;
```
mmat11: this can probably be taken from `pid_conn->pid` to save some stack space

bpf/generictracer/protocol_mssql.h
```diff
@@ -198,9 +198,13 @@ static __always_inline void mssql_send_large_buffer(tcp_req_t *req,
         large_buf->type = EVENT_TCP_LARGE_BUFFER;
         large_buf->packet_type = packet_type;
         large_buf->action = action;
+        large_buf->kind = k_large_buf_kind_tcp_known;
         large_buf->direction = direction;
         large_buf->conn_info = req->conn_info;
         large_buf->tp = req->tp;
+        const u64 pid_tid = bpf_get_current_pid_tgid();
+        const u32 pid = pid_from_pid_tgid(pid_tid);
+        large_buf->pid = pid;
```
mmat11: same here, should already be available in `req->pid_info->host_pid`

bpf/generictracer/protocol_mysql.h
```diff
@@ -148,9 +148,13 @@ static __always_inline int mysql_send_large_buffer(tcp_req_t *req,
     lb->type = EVENT_TCP_LARGE_BUFFER;
     lb->packet_type = packet_type;
     lb->action = action;
+    lb->kind = k_large_buf_kind_tcp_known;
     lb->direction = direction;
     lb->conn_info = pid_conn->conn;
     lb->tp = req->tp;
+    const u64 pid_tid = bpf_get_current_pid_tgid();
+    const u32 pid = pid_from_pid_tgid(pid_tid);
+    lb->pid = pid;
```
mmat11: same here, from `pid_conn`

bpf/generictracer/protocol_postgres.h
```diff
@@ -66,9 +66,13 @@ static __always_inline int postgres_send_large_buffer(tcp_req_t *req,
     large_buf->type = EVENT_TCP_LARGE_BUFFER;
     large_buf->packet_type = packet_type;
     large_buf->action = action;
+    large_buf->kind = k_large_buf_kind_tcp_known;
     large_buf->direction = direction;
     large_buf->conn_info = req->conn_info;
     large_buf->tp = req->tp;
+    const u64 pid_tid = bpf_get_current_pid_tgid();
+    const u32 pid = pid_from_pid_tgid(pid_tid);
+    large_buf->pid = pid;
```
mmat11: same here

bpf/generictracer/protocol_tcp.h
```diff
@@ -145,14 +146,61 @@ static __always_inline void finish_ongoing_tcp_req(pid_connection_info_t *pid_co
...
+    lb->type = EVENT_TCP_LARGE_BUFFER;
+    lb->packet_type = packet_type;
+    lb->action = action;
+    lb->kind = k_large_buf_kind_tcp;
+    lb->direction = direction;
+    lb->conn_info = pid_conn->conn;
+    lb->tp = req->tp;
+    const u64 pid_tid = bpf_get_current_pid_tgid();
+    const u32 pid = pid_from_pid_tgid(pid_tid);
+    lb->pid = pid;
```
mmat11: same here

bpf/generictracer/protocol_tcp.h
mmat11: the `is_server` was added here https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1034/changes#r2635303604 and I see some other comments related to that; I can't remember the context /cc @pinoOgni

pkg/ebpf/common/tcp_large_buffer.go
```diff
@@ -12,10 +12,21 @@ import (
...
+	LBKindTCP      largeBufferKind = 0
+	LBKindHTTP     largeBufferKind = 1
+	LBKindTCPKnown largeBufferKind = 2
+)
+
 type largeBufferKey struct {
 	traceID               [16]uint8
 	packetType, direction uint8
 	connInfo              BpfConnectionInfoT
+	pid                   uint32
```
mmat11: question: I understand why `kind` might be necessary, why do we need `pid`?

grcevski: So here's the scenario I ran into:

```
Client call  1.2.3.4:12345 -> 5.6.7.8:8080 (we track large buffers per connection as a client)
Server call 1.2.3.4:12345 -> 5.6.7.8:8080 (we track large buffers per connection as a server)
```

If OBI tracks both services, the large buffers are duplicated except for the first one, since the first one is add, the others are append. Imagine we have this scenario:

1. Large buffer send (Hello) (client), OP: add
    State: Hello
2. Large buffer receive (Hello) (server), OP: add
    State: Hello
3. Large buffer send (World) (client), OP: append
    State: Hello World
4. Large buffer receive (World) (server), OP: append
    State: Hello World World

It should only be "Hello World" but because we track per connection, we record everything but the first packet twice.

bpf/generictracer/protocol_kafka.h
grcevski: Oh I didn't check if we had pid_conn, I'll verfify. I cannot use req->pid_info->host_pid because it's not set on the first request for TCP. I had it like that, but I broke bunch of tests because we missed the first large buffer packet.

pkg/ebpf/common/tcp_large_buffer.go
grcevski: Yeah you are right, I can reduce the 3 options into two, wire and app. Good idea.

pkg/ebpf/common/tcp_large_buffer.go
mmat11: shouldn't that be already covered by the (packet_type, direction) tuple?

grcevski: Hmmm, good point. Let me take a look, maybe the issue I was hitting was something else.

pkg/ebpf/common/tcp_large_buffer.go
grcevski: Updated.

pkg/ebpf/common/tcp_large_buffer.go
grcevski: I removed the PID, my issue was with not determining correctly is_server, which is now resolved.

bpf/common/common.h
```diff
@@ -53,6 +53,13 @@ enum large_buf_action : u8 {
     k_large_buf_action_append = 1,
 };
 
+enum large_buf_kind : u8 {
+    k_large_buf_layer_wire =
+        0, // <--- wire format as seen by kprobes or socket ingress programs/xdp
+    k_large_buf_layer_app =
+        1, // <--- as seen by the app layer, e.g. ssl uprobes, detected protocols in eBPF or sk_msg on engress
```
mmat11: typo, engress

pkg/ebpf/common/tcp_large_buffer.go
```diff
@@ -69,33 +79,75 @@ func appendTCPLargeBuffer(parseCtx *EBPFParseContext, record *ringbuf.Record) (r
...
+		fallthrough
+	case ProtocolTypePostgres:
+		fallthrough
+	case ProtocolTypeMSSQL:
+		return KindLayerApp
+	}
+	// No large buffers for MQTT
+	return KindLayerWire
+}
+
```
mmat11, line 129: I think we can just call extractTCPLargeBuffer instead of having these wrapper functions. We could extend `protocolToLargeBufferKind` to also have `ProtocolTypeHTTP` (which is currently unused, but could be hardcoded caller-side) and translate to `kind` in the function itself (not the wrapper)

grcevski: Good idea!

grcevski: Updated! Thanks!

mmat11 (approved):
lgtm!

rafaelroquetto (approved):

PR #2005, opened 2026-05-05
fix: align AI instrumentations with GenAI OTel semconv

ishanjainn:
## Summary

- Align GenAI span attributes (gen_ai.input.messages, gen_ai.system_instructions, gen_ai.output.messages, gen_ai.tool.definitions) with the [OpenTelemetry GenAI semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/) JSON schema (parts-based format)
- Normalize all provider-specific message formats (OpenAI, Anthropic, Gemini, Qwen, Bedrock) into the unified semconv structure: [{"role":"...","parts":[{"type":"text","content":"..."}]}]
- Derive gen_ai.operation.name from URL path rather than response body "object" field for OpenAI (/v1/chat/completions → "chat") and Qwen (same path pattern → "chat")
- Add support for tool call normalization across providers (function calls, tool_use, functionDeclarations)
- Fix OATS integration test expectations to match the new normalized attribute formats

## Validation


marctc:
Hey @ishanjainn, do you mind add a description to the PR? also, you reckon that the PR can be splitted? is kinda big

pkg/ebpf/common/http/openai.go
```diff
@@ -53,11 +53,18 @@ func OpenAISpan(baseSpan *request.Span, req *http.Request, resp *http.Response)
...
 	parsedResponse.Request = parsedRequest
 
-	// Override operation name for embedding requests.
+	// Override operation name and derive API type from URL path.
 	if req.URL != nil {
 		path := strings.TrimSuffix(req.URL.Path, "/")
-		if path == "/v1/embeddings" {
+		switch path {
+		case "/v1/chat/completions":
+			parsedResponse.OperationName = "chat"
```
grcevski: Can we make these constants like `request.EmbeddingOperationName`?

pkg/appolly/app/request/genai_normalize.go
```diff
@@ -0,0 +1,638 @@
...
+
+import (
+	"encoding/json"
+)
+
+// Semconv-compliant message types per:
+// https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-input-messages.json
+// https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-output-messages.json
+
+type normalizedPart struct {
```
grcevski, line 14: Can we split these normalize functions into separate files, specific to the AI vendor? Also can we add unit tests for all edge cases, e.g. failures to parse, not enough parts etc.

pkg/ebpf/common/http/bedrock.go
```diff
@@ -101,3 +105,34 @@ func extractBedrockModel(req *http.Request) string {
...
+func isBedrockStream(req *http.Request) bool {
+	if req == nil || req.URL == nil {
+		return false
+	}
+	return strings.Contains(req.URL.Path, "invoke-with-response-stream")
+}
+
+// extractBedrockGuardrailID extracts the guardrail identifier from the
+// response header or the request JSON body.
+func extractBedrockGuardrailID(req *http.Request, resp *http.Response) string {
```
grcevski, line 120: Can we please add dedicated unit tests for this function?

grcevski (commented):
Great addition! I added a few comments of tests and some minor refactorings I think that should be added.

pkg/ebpf/common/http/bedrock.go
ishanjainn: Updated

pkg/ebpf/common/http/openai.go
ishanjainn: Addressed

pkg/appolly/app/request/genai_normalize.go
ishanjainn: Addressed

mmat11 (approved):
lgtm, just one minor comment: I'm not sure, but probably we should move all `genai_*.go` stuff into a `genai` subdirectory, WDYT?

ishanjainn:
@mmat11 

> we should move all genai_*.go stuff into a genai subdirectory, WDYT?

I agree, Ill create a follow up PR for that

pkg/appolly/app/request/genai_normalize_openai.go
```diff
@@ -0,0 +1,176 @@
...
+	if len(content) == 0 {
+		return nil
+	}
+
+	var s string
+	if err := json.Unmarshal(content, &s); err == nil {
+		return []normalizedPart{{Type: "text", Content: s}}
+	}
+
+	return []normalizedPart{{Type: "text", Content: string(content)}}
```
MrAlias: This still doesn’t normalize structured OpenAI `content` payloads. When `content` is an array like `[{"type":"text",...},{"type":"image_url",...}]`, we currently emit a single semconv part of type `text` whose content is the raw JSON string.

In practice that means chat and multimodal requests/responses lose their structure, so this PR only normalizes the string-only cases. I think the fix should parse the known OpenAI content block shapes and map them into semconv parts directly, instead of falling back to `string(content)` for every non-string payload.

MrAlias (dismissed):
This is close, but I found two correctness gaps that I think should be fixed before merge.

First, the new OpenAI normalizer still drops structured `content` into a single text blob instead of preserving semconv parts, so multimodal and structured chat payloads are not actually normalized.

Second, this normalization work only landed in `tracesgen`. The shared getter path still emits raw provider payloads for several GenAI attributes, which leaves the repo with two inconsistent export paths depending on which attribute builder is used.

pkg/appolly/app/request/genai_normalize_openai.go
grcevski: Maybe this can be a follow-up, when multiple structured parts are in the content? Or do you think what's here now will break the multi response content array?

MrAlias: I’d prefer not to defer this to a follow-up.

The issue is broader than the case where `content` contains multiple structured parts. In the current implementation, any non-string OpenAI `content` falls back to a single semconv `text` part containing the raw JSON payload. That means even a single structured content array is normalized incorrectly, and mixed arrays like `[{"type":"text",...},{"type":"image_url",...}]` lose both their structure and their per-part types.

So I see this as a correctness gap in the current PR rather than an incremental improvement we can safely defer. Since the purpose of this change is to align these attributes with the GenAI semconv shape, I think we should parse the known OpenAI content block variants and map them into semconv parts directly before merging.

Also, this thread only covers one part of my top-level review. The other correctness issue I called out there, about normalization only landing in `tracesgen` while the shared getter path still emits raw provider payloads for some GenAI attributes, would still need to be addressed as well.

ishanjainn: @MrAlias @grcevski 
Made changes as requested. Lemme know if anything else needs to be updated

NameHaibinZhang:
LGTM!

ishanjainn:
@MrAlias @grcevski can you take another pass at the changes?

MrAlias (commented):
I reran `go test ./pkg/export/otel -run TestGenerateTracesAttributes`, `go test ./pkg/export/otel/...`, and `go test ./pkg/appolly/app/request ./pkg/ebpf/common/http ./pkg/export/otel/...` after generation; they passed locally. I did my best to review this as-is, but this is a large PR spanning multiple providers and export paths, and it would likely be easier to validate and maintain if it were split into smaller pieces. The remaining blocking issues are inline.

pkg/appolly/app/request/genai_normalize.go
```diff
@@ -0,0 +1,151 @@
...
+	Type      string          `json:"type"`
+	Content   string          `json:"content,omitempty"`
+	Response  any             `json:"response,omitempty"`
```
MrAlias: The semconv output schema uses `result` for `tool_call_response` parts. Emitting `response` here leaves every normalized tool response object off-schema, and this field is populated by the OpenAI, Anthropic, and Gemini normalizers.
```suggestion
	Response  any             `json:"result,omitempty"`
```

pkg/appolly/app/request/genai_normalize_openai.go
```diff
@@ -0,0 +1,274 @@
...
+// array to the semconv output messages schema.
+func normalizeOpenAIResponsesOutput(raw json.RawMessage) string {
+	var items []struct {
+		Role    string `json:"role"`
+		Status  string `json:"status"`
+		Content []struct {
+			Type string `json:"type"`
+			Text string `json:"text"`
+		} `json:"content"`
+	}
```
MrAlias, line 227: This decoder only models message-shaped output items. OpenAI Responses output can also contain tool/function-call items, and those currently fall through here as zero-value structs and get emitted as empty normalized messages instead of `tool_call` parts.

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -566,30 +577,84 @@ func TraceAttributesSelector(span *request.Span, optionalAttrs map[attr.Name]str
...
+			if ai.APIType != "" {
+				attrs = append(attrs, openAIAPITypeKey.String(ai.APIType))
+			}
 			if _, ok := optionalAttrs[attr.GenAIInput]; ok {
 				attrs = append(attrs, semconv.GenAIInputMessagesKey.String(ai.Request.GetInput()))
 			}
 			if _, ok := optionalAttrs[attr.GenAIOutput]; ok {
-				attrs = append(attrs, semconv.GenAIOutputMessagesKey.String(ai.GetOutput()))
+				if ai.OperationName != request.EmbeddingOperationName {
+					attrs = append(attrs, semconv.GenAIOutputMessagesKey.String(ai.GetOutput()))
```
MrAlias, line 661: This fixes the tracesgen path for embeddings, but the shared getter path still unconditionally returns OpenAI/Qwen output messages. That leaves `gen_ai.output.messages` behavior inconsistent depending on which attribute builder is used.

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -604,22 +669,59 @@ func TraceAttributesSelector(span *request.Span, optionalAttrs map[attr.Name]str
 			}
 			attrs = append(attrs, semconv.GenAIRequestModel(ai.Input.Model))
 			attrs = append(attrs, semconv.GenAIResponseModel(ai.Output.Model))
-			attrs = append(attrs, semconv.GenAIUsageInputTokens(ai.Output.Usage.InputTokens))
+			// Per Anthropic semconv: input_tokens excludes cached tokens.
+			// Compute total: input_tokens + cache_read + cache_creation.
+			anthropicInputTokens := ai.Output.Usage.InputTokens +
+				ai.Output.Usage.CacheReadInputTokens +
+				ai.Output.Usage.CacheCreationInputTokens
+			attrs = append(attrs, semconv.GenAIUsageInputTokens(anthropicInputTokens))
```
MrAlias: This brings the trace attribute into line with Anthropic semconv, but the shared `Span.GenAIInputTokens()` helper still returns only `Usage.InputTokens`, and metrics records that helper directly. So after this change traces and metrics still disagree for Anthropic cache usage.

pkg/appolly/app/request/genai_normalize.go
```diff
@@ -0,0 +1,151 @@
...
+				Type:        "function",
+				Name:        fd.Name,
+				Description: fd.Description,
+				Parameters:  fd.Parameters,
+			})
+		}
+		return out
+	}
+
+	nt := normalizedTool{Type: "function"}
```
MrAlias: This normalization path still hardcodes everything to `type:"function"`, and unsupported items end up normalizing to `null` (as codified by `TestNormalizeToolDefinitions_InvalidItem`). For semconv output, it would be better to drop unsupported entries or preserve the original JSON than emit invalid tool-definition objects.

pkg/appolly/app/request/genai_normalize_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+func TestNormalizeToolDefinitions_ParseFailure(t *testing.T) {
+	invalid := json.RawMessage(`not json`)
+	assert.Equal(t, "not json", NormalizeToolDefinitions(invalid))
+}
+
+func TestNormalizeToolDefinitions_InvalidItem(t *testing.T) {
+	raw := json.RawMessage(`[123]`)
+	result := NormalizeToolDefinitions(raw)
+	assert.Equal(t, "null", result)
```
MrAlias: This test is currently locking in `"null"` as the normalized output for an unsupported tool-definition item. That makes it harder to fix the invalid-output behavior above, and it also codifies a result that is not semconv-valid for `gen_ai.tool.definitions`.

pkg/appolly/app/request/span.go
```diff
@@ -681,8 +739,14 @@ type JSONRPC struct {
 
 // Generic embedding provider types (Voyage AI, Cohere, Jina AI)
 
-// EmbeddingOperationName is the canonical operation name for embedding spans.
-const EmbeddingOperationName = "embeddings"
+// GenAI operation name constants aligned with OTel semantic conventions.
+const (
+	ChatOperationName        = "chat"
+	CompletionOperationName  = "completion"
```
MrAlias: The semconv well-known operation name here is `text_completion`, so wiring `/completions` through `completion` still leaves the operation-name mapping off-spec.
```suggestion
	CompletionOperationName  = "text_completion"
```

grcevski (approved):
LGTM! There seems to be one unit test failure because of the recent change from `completion` to `text_completion`.  TestExtractQwenOperation/completions

Issue #1821, opened 2026-04-14
Go JSON-RPC emits http metrics and not RPC metrics


NimrodAvni78:
as we can see from this [test case](https://github.com/coralogix/opentelemetry-ebpf-instrumentation/blob/9026e282e8b87557d1fc3b2fc14c2273defb4f57/internal/test/integration/red_test.go#L316) we expect that go jsonrpc calls will emit http metrics, instead of the [RPC metrics](https://opentelemetry.io/docs/specs/semconv/rpc/rpc-metrics/).

krakenlgk:
Hi maintainers -- first-time contributor here, would like to pick this up.

mmat11:
@krakenlgk thanks!

krakenlgk:
Hi @mmat11 Looks like the Go JSON-RPC routing piece of this issue is already being addressed in #1967 by @apatniv, which adds the `"JSONRPC"` sentinel on the BPF side and carries the method name through. I had opened #2007 before spotting that overlap and closed it to avoid duplicate work.

Happy to defer to #1967. Once it merges, if there's any follow-up scope left on this issue, I'd like to pick it up as a smaller follow-up PR. Otherwise feel free to close this once #1967 lands, or reassign as you see fit.

Thanks for the opportunity to contribute first time on an OBI issue and the feedback has been useful.

NimrodAvni78:
solved by https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1967

PR #2007, opened 2026-05-05
fix(ebpf): emit RPC metrics for Go JSON-RPC

krakenlgk:
## Summary

Refs #1821.

Go `net/rpc/jsonrpc` traces were being exported as HTTP metrics with the JSON-RPC procedure name leaking into `http.request.method` (e.g. `Arith.M` after the 7-byte uprobe truncation). This violates OTel semantic conventions and produces unbounded `http.request.method` cardinality. After the fix, the same traffic surfaces as `rpc_server_duration_seconds` with `rpc_system="jsonrpc"`.

## Root cause

Two JSON-RPC detection paths exist:

| Path | Trigger | Sets `SubType=JSONRPC`? |
|------|---------|-------------------------|
| **A** — kprobe body parsing (`pkg/ebpf/common/http/jsonrpc.go`) | `OTEL_EBPF_HTTP_JSONRPC_ENABLED=true`, language-agnostic, payload-driven | Yes |
| **B** — Go uprobe (`bpf/gotracer/go_nethttp.c`) | Always-on for Go, overwrites `method` field with the procedure name | No (this PR) |

The exporter routing in `pkg/export/otel/metrics.go` and `pkg/export/prom/prom.go` already dispatches on `SubType==HTTPSubtypeJSONRPC` correctly — Path B was simply never setting it.

## Approach

Detect the JSON-RPC case in `HTTPRequestTraceToSpan` via a **negative allowlist**:

- HTTP verbs are a closed set (`GET, HEAD, POST, PUT, DELETE, CONNECT, OPTIONS, TRACE, PATCH`) plus the empty string for non-HTTP traces.
- The Go uprobe truncates `method` to `k_method_max_len` (7 bytes) — every standard verb fits intact.
- Any captured value outside that closed set therefore marks a JSON-RPC procedure name.

When detected, the span is promoted to `HTTPSubtypeJSONRPC`, the `JSONRPC` struct is populated, and `Method` is reset to `"POST"` so the wire-level HTTP method (rather than the leaked procedure name) appears in trace attributes.

## Files

| File | Change |
|------|--------|
| `pkg/ebpf/common/spanner.go` | Add `isJSONRPCMethod` helper, promote spans in `HTTPRequestTraceToSpan` |
| `pkg/ebpf/common/spanner_test.go` | Unit tests for the helper (14 cases) and the spanner promotion path (6 cases) |
| `internal/test/integration/red_test.go` | Flip `testREDMetricsForJSONRPCHTTP` from asserting the buggy HTTP output to asserting `rpc_server_duration_seconds` + a negative assertion that the procedure name no longer leaks into `http.request.method` |

## Out of scope

- **#1799 (span name uses `Arith.Multiply` instead of `Arith` / `Multiply`)** — that's an `RPCService` getter concern, orthogonal to the metric routing fix here. Already assigned to @apatniv.
- **MCP detection on the Go path** — MCP requires payload parsing (Path A only). The Go uprobe surfaces MCP traffic as plain JSON-RPC, which is the closest accurate classification possible without parsing the body.

## Test plan

- [x] Unit tests pass locally (`go test ./pkg/ebpf/common/...`)
- [x] `go vet`, `go build`, `golangci-lint`, `gofmt` clean
- [ ] Integration test (`testREDMetricsForJSONRPCHTTP`) passes against running stack — CI will validate
- [ ] CLA signed (will action on EasyCLA bot prompt)

## Questions for maintainers

1. **Heuristic vs structural flag:** I went with the negative allowlist for simplicity. An alternative would be adding an `is_jsonrpc` flag to the BPF struct (`bpf/gotracer/types/nethttp.h`) and setting it in the uprobe. Happy to switch if you prefer the explicit signal.
2. **`rpc.jsonrpc.version`:** I default to `"2.0"` because OTel semconv references the JSON-RPC spec version. The Go `net/rpc/jsonrpc` package speaks 1.0 on the wire. Open to changing to `""` or `"1.0"` if you'd rather match wire reality.
3. **Negative assertion strictness:** The integration test asserts that `http_server_request_duration_seconds` is NOT emitted with the procedure-name as `http.request.method`. If you'd prefer a softer check, I can relax it.

First-time CNCF contributor — guidance very welcome on any of the above.

### DIFF
--- a/internal/test/integration/red_test.go
+++ b/internal/test/integration/red_test.go
@@ -317,6 +317,8 @@ func testREDMetricsForJSONRPCHTTP(t *testing.T, url, svcName, svcNs string) {
 	jsonBody, err := os.ReadFile(path.Join(pathRoot, "internal", "test", "integration", "components", "testserver", "jsonrpc", "body", "formated.json"))
 	require.NoError(t, err)
 	urlPath := "/jsonrpc"
+	// The Go uprobe truncates the captured method to k_method_max_len (7 bytes),
+	// so "Arith.Multiply" surfaces as "Arith.M" on the metric label.
 	expectedMethod := "Arith.M"
 
 	for i := 0; i < 4; i++ {
@@ -328,22 +330,27 @@ func testREDMetricsForJSONRPCHTTP(t *testing.T, url, svcName, svcNs string) {
 	var results []promtest.Result
 	require.EventuallyWithT(t, func(ct *assert.CollectT) {
 		var err error
-		results, err = pq.Query(`http_server_request_duration_seconds_count{` +
-			`http_request_method="` + expectedMethod + `",` +
-			`http_response_status_code="200",` +
+		results, err = pq.Query(`rpc_server_duration_seconds_count{` +
+			`rpc_method="` + expectedMethod + `",` +
+			`rpc_system="jsonrpc",` +
 			`service_namespace="` + svcNs + `",` +
-			`service_name="` + svcName + `",` +
-			`url_path="` + urlPath + `"}`)
+			`service_name="` + svcName + `"}`)
 		require.NoError(ct, err)
 		enoughPromResults(ct, results)
 		val := totalPromCount(ct, results)
 		assert.LessOrEqual(ct, 3, val)
-		if len(results) > 0 {
-			res := results[0]
-			addr := res.Metric["client_address"]
-			assert.NotNil(ct, addr)
-		}
 	}, testTimeout, 100*time.Millisecond)
+
+	// Negative assertion: the JSON-RPC traffic must not surface as plain HTTP
+	// metrics carrying a procedure-name as http.request.method (the original
+	// bug from issue #1821).
+	leakResults, err := pq.Query(`http_server_request_duration_seconds_count{` +
+		`http_request_method="` + expectedMethod + `",` +
+		`service_namespace="` + svcNs + `",` +
+		`service_name="` + svcName + `",` +
+		`url_path="` + urlPath + `"}`)
+	require.NoError(t, err)
+	assert.Empty(t, leakResults, "JSON-RPC procedure must not leak into http.request.method")
 }
 
 func testREDMetricsForHTTPLibrary(t *testing.T, url, svcName, svcNs string) {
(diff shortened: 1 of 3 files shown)


rafaelroquetto:
@krakenlgk I haven't reviewed the code yet, but could you please point your AI agent to AGENTS.md and go.instructions.md at least, to make sure it respects the AI generated code conventions and check against that.

Also, I'd like to bring your attention to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1967 - this may overlap with your proposal.

rafaelroquetto:
@krakenlgk please check our contributing guidelines. They explicit state that:

> Contributors must review, test, and understand all changes before submitting a PR. This applies equally to manually written and tool-generated code. Use of AI or other tools does not transfer responsibility — the contributor is fully accountable for the final patch.


and most importantly here:

> **Vet AI-generated plans and issue reports before filing.** If you use an AI tool to draft an issue, design proposal, or implementation plan, read it critically before submitting. Check that it accurately reflects the codebase, does not contradict existing architecture, and does not propose work that is already done. Unvetted AI output creates noise and wastes reviewer time.

Since you responded with an AI generated plan.

We can't review a PR that is entirely AI generated, so if you'd like this to move forward, I'd encourage you to update the PR description with your own words, and respond to any questions posed by a reviewer without using AI.

krakenlgk:
@rafaelroquetto thanks — read AGENTS.md and the Go instructions, and looked at #1967. There's clear overlap: #1967 covers #1799 (span name + method-leak) but doesn't route to RPC metrics, which is what #1821 asks for. Plan is to wait for #1967 to land, rebase on top, drop my heuristic in favor of the explicit `JSONRPC` sentinel it introduces, and just set `SubType=HTTPSubtypeJSONRPC` + populate the JSONRPC struct from `pattern`. Marking this draft in the meantime.

rafaelroquetto:
> Going to mark this as draft until https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1967 lands. If the simplified plan above sounds off, happy to adjust before I rework it.

It's best to close the PR and reopen it/create a new one once you think it's in a reviwable state. From our guidelines:

> PRs must be ready for review when submitted, not ready for validation. The contributor is responsible for verifying correctness before asking for a reviewer's time.

Issue #1759, opened 2026-04-06
Define the OBI v1 telemetry contract


MrAlias:
As part of `v1.0`, OBI needs an explicit decision about which emitted telemetry is stable.

Today, telemetry behavior exists in code, tests, and docs, but the `v1` release needs a clear contract for what users can rely on across future releases.

## Goal

Define and document the emitted telemetry surface that is part of the OBI `v1` compatibility contract.

## Scope

This should cover at least:

- which spans and metrics are part of the stable `v1` surface
- which names, attributes, and units are stable
- what telemetry remains experimental or outside the `v1` contract
- known caveats or limitations that must be documented for `v1`

This issue is broader than telemetry schema adoption alone.

## Tasks

- [ ] Define which emitted telemetry is stable at `v1`
- [ ] Define which names, attributes, units, and semantics are part of the `v1` contract
- [ ] Identify telemetry that remains experimental or outside the `v1` contract
- [ ] Document known telemetry limitations that are acceptable for `v1`
- [ ] Align or explicitly relate this work to telemetry schema work
- [ ] Publish the resulting telemetry contract in the appropriate user-facing location(s)
- [ ] Link the published contract from the `v1.0` release epic

## Related

- #1148

## Acceptance Criteria

- [ ] The emitted telemetry that is stable at `v1` is explicitly documented
- [ ] Experimental or out-of-contract telemetry is clearly distinguished
- [ ] Known telemetry caveats for `v1` are documented
- [ ] Maintainers agree this defines the telemetry compatibility baseline for `v1.0`

MrAlias:
SIG decision: do not treat the #1092 network attribute rename as part of the v1 contract.

The v1 telemetry contract should explicitly exclude the current network flow attribute names (`src.*`, `dst.*`, `transport`, `iface`, `direction`) from the stable surface, so a later alignment with OpenTelemetry semantic conventions is not an accidental v1 break.

MrAlias:
@NimrodAvni78, thanks for driving the telemetry registry work. I want to confirm the remaining scope of this issue. My understanding is:

- This issue defines and publishes the telemetry compatibility guarantees for v1.
- #2562 separately tracks exhaustive registry coverage and deterministic validation of everything OBI can emit.
- This issue should only be blocked by a #2562 coverage gap when that gap affects telemetry proposed for the stable v1 surface.

From what I see, the remaining concrete deliverables are:

[...]

PR #2008, opened 2026-05-05
[CI] Weaver full coverage of integration tests

NimrodAvni78:
## Summary

Follow up of https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2059

Wires the [`weaver`](https://github.com/open-telemetry/weaver) semantic-convention validator into nearly all integration tests so every metric / span OBI
emits is checked against the OBI registry under `schemas/obi/` (and through it
the upstream OpenTelemetry semantic conventions).

### Architecture

```
                              ┌─ metrics        → prometheus exporter ─→ tests assert
OBI ──OTLP──→ otelcol (otlp) ─┼─ traces         → jaeger
                              ├─ metrics/weaver → otlp/weaver ─→ weaver:4317 → Rego live-check
                              └─ traces/weaver  → otlp/weaver ─→ weaver:4317 → Rego live-check
```

- New `weaver` container (`otel/weaver:v0.23.0`, digest-pinned) runs alongside
  otelcol in each weaver-wired compose. Shared definition at
  `internal/test/integration/components/weaver/service.yml`, pulled in via
  Docker Compose `extends:` (no per-compose duplication).
- otelcol fans OTLP onto **separate** weaver pipelines (`metrics/weaver`,
  `traces/weaver`). Bounded queue, `block_on_overflow: false`,
  `retry_on_failure: false` — a slow weaver can never back-pressure prom or
  jaeger.
- `runWeaverValidation(t)`:
  1. POST `/stop` to weaver's admin port (`4320`).
  2. `docker wait` for the container to flush.
  3. `docker cp /tmp/live_check.json` out of the container.
  4. Parse the report and assert zero actionable violations.

  Skips parsing on prior subtest failure so the real bug stays the headline.
  Falls back to `docker rm -f` on any post-`/stop` step error so the next run
  starts clean.
- Equivalent helper `setupContainerWeaver` in `dockerutil_test.go` for the
  programmatic-setup tests (`cloud_*`, `http_go_otel_*`) that don't use
  docker-compose.

### Advice on how to review
This PR touches many files so its intimidating to review, but most of the changes are repeating across tests and fall into the same categories

- `internal/test/integration/configs/otelcol-config*weaver*.yml` - new collector config containing weaver for different types of setup
- `internal/test/integration/configs/obi-config*.yml` - Adding weaver to other collector configuration 
- `internal/test/integration/*_test.go` - added `runWeaverValidation(t)` as a final assertion to each test, for manual docker api test we also add 

```go
	setupContainerWeaver(t, network)
	setupContainerCollector(t, network, "otelcol-config-weaver.yml")
```
- `internal/test/integration/docker-compose*.yml` - bumped`OTEL_EBPF_METRICS_INTERVAL`: `10ms` → `1s`, appended `weaver` for collector config, add `weaver` component via extends
- `internal/test/integration/weaver.go` - add `weaverIgnoredSignals` and `weaverIgnoredAdviceMessages`

## Performance overhead

Measured against equivalent runs on `main`:

|                          | main    | weaver branch |
| ------------------------ | ------- | ------------- |
| Median shard duration    | ~19 min | ~22 min       |
| Slowest shard            | 20 min  | 26 min        |

Roughly **+15–20% per shard**. Adding around 20-40 minutes of total CI time
This is after some rounds of optimizations, if we think this is not acceptable we can do more optimizations around it

## Tests **without** weaver validation (and why)

| Test | Why no weaver |
| --- | --- |
| `TestInstrumentationErrors` | Runs OBI **without privileges** to force instrumentation errors. Zero application telemetry is produced by design. |
| `TestAvoidedServicesMetrics` | OBI internal-metrics test only, no application telemetry to validate. |
| `TestExistingSocketsDetection` (keepalive) | Protocol-detection test, no application telemetry. |
| K8s suite (`internal/test/integration/k8s/...`) | Different test infrastructure (manifests, not compose). Out of scope for this PR. |

## `application_span` → `application_span_otel`

In `dockerutil_test.go:471-472`, the obi container default for both
`OTEL_EBPF_METRICS_FEATURES` and `OTEL_EBPF_PROMETHEUS_FEATURES` switched from

```
application,application_span         # legacy: traces_spanmetrics_calls_total, traces_spanmetrics_latency
```

to

```
application,application_span_otel    # canonical: traces_span_metrics_calls_total, traces_span_metrics_duration
```

The OBI registry only declares the canonical names — leaving the legacy feature
on produced hundreds of "Metric does not exist in the registry" violations per
test.

A handful of compose files intentionally **keep** `application_span` in
`OTEL_EBPF_PROMETHEUS_FEATURES` specifically:

- `docker-compose-client.yml`
- `docker-compose-other-grpc.yml`
- `docker-compose-go-otel-grpc.yml`
- `docker-compose-python.yml`

These tests still exercise OBI's legacy prom output for downstream
back-compat. `OTEL_EBPF_METRICS_FEATURES` on the same composes is
`application_span_otel`, so the OTel / weaver path stays clean while the
prom-scrape path still emits the legacy names for prom assertions.

## Weaver exceptions

Two small allow-lists live in `internal/test/integration/weaver.go`.

### `weaverIgnoredSignals`

```
metric:rpc.server.duration
metric:rpc.client.duration
```

OBI emits these with `unit: "s"`, but semconv **v1.38** (the version pinned in
`schemas/obi/manifest.yaml`) specifies `unit: "ms"`. Live-check resolves
against upstream and flags every data point. A registry override doesn't help
— live-check rejects duplicate IDs. The unit reverts to `s` in
**semconv v1.40.0**, so these entries drop once the manifest is bumped (TODO
in the source comment).

### `weaverIgnoredAdviceMessages`

```
Namespace 'server' collides with existing attribute 'server.address'
Namespace 'server' collides with existing attribute 'server.port'
Namespace 'client' collides with existing attribute 'client.address'
Namespace 'client' collides with existing attribute 'client.port'
Namespace 'iface'  collides with existing attribute 'iface.direction'
```

- **`server` / `client`** — OTel-contrib's `servicegraphconnector` emits bare
  `server` / `client` labels (matched in OBI's `service_graph.yaml`), but
  upstream semconv reserves `server.*` / `client.*` as namespace prefixes
  (`server.address`, `server.port`, …). Weaver flags every signal that
  touches an upstream `server.*` attribute, even ones that don't use the bare
  label. The contract is owned by the connector spec, not OBI.
- **`iface`** — OBI's netolly emits `iface` (interface name) **and**
  `iface.direction`. Same shape: a leaf attribute that's also a namespace.
  Mirrors the older network-flow exporter convention, fixed for back-compat.

## Metrics export interval: `10ms` → `1s`

Bumped on **56 weaver-wired composes**. The historical `10ms`
`OTEL_EBPF_METRICS_INTERVAL` overwhelmed weaver's per-entity Rego pass — the
`otlp/weaver` queue overflowed within seconds, every batch in flight was
dropped, and `runWeaverValidation` consistently raced against an empty
report. `1s` drains cleanly, with no observable change in what the test
assertions can see (the production pipeline still polls Prometheus / Jaeger
on the same timing it always did).

The **4 composes intentionally left at `10ms`** are non-weaver-wired or
otherwise low-volume:

```
docker-compose-go-stat-metrics.yml
docker-compose-large-http-req.yml
docker-compose-log-enricher.yml
docker-compose-go-otel-grpc.yml
```

## Validation


mariomac (approved):

mmat11:
not sure it's doable in all instances, instead of adding `runWeaverValidation()` to all tests, maybe there's a way to do that per-package/suite?
something like:
```
func TestMain(m *testing.M) {
    r := m.Run()
    runWeaverValidation(m)
    os.Exit(r)
}
```

mmat11:
> not sure it's doable in all instances, instead of adding `runWeaverValidation()` to all tests, maybe there's a way to do that per-package/suite? something like:
> 
> ```
> func TestMain(m *testing.M) {
>     r := m.Run()
>     runWeaverValidation(m)
>     os.Exit(r)
> }
> ```

this probably needs a bigger refactor since Run() will also teardown the compose containers so disregard my comment for now

skl (approved):
lgtm, can always optimise ci perf later

PR #2009, opened 2026-05-05
Add `network.tcp.handshake.role` attribute to TCP failed connection metrics

pinoOgni:
## Summary

This PR adds a `network.tcp.handshake.role` attribute to the `obi_stat_tcp_failed_connections` metric, indicating whether the local endpoint was acting as a TCP client or server at the time of the failure.

Note: the `network.tcp.handshake.role` attribute name is taken from proposal issue https://github.com/open-telemetry/semantic-conventions/issues/3682 so it may change in the future (for example with `network.tcp.connection.role` or `network.tcp.role`).

## Validation

### DIFF
--- a/bpf/statsolly/tp_tcp.c
+++ b/bpf/statsolly/tp_tcp.c
@@ -14,6 +14,7 @@
 
 #include <statsolly/types.h>
 #include <statsolly/maps/stats_events.h>
+#include <statsolly/maps/sock_role.h>
 
 #ifndef ECONNREFUSED
 #define ECONNREFUSED 111
@@ -42,16 +43,10 @@ enum tcp_fail_reason {
 };
 
 enum tcp_handshake_role {
+    role_unknown = 0,
     role_client = 1,
     role_server = 2,
 };
-
-struct {
-    __uint(type, BPF_MAP_TYPE_LRU_HASH);
-    __type(key, void *);
-    __type(value, u8);
-    __uint(max_entries, 1 << 16);
-} sock_role SEC(".maps");
 
 static __always_inline u8 sk_err_to_reason(const int err) {
     switch (err) {
@@ -89,7 +84,7 @@ int obi_tracepoint_inet_sock_set_state(struct trace_event_raw_inet_sock_set_stat
         return 0;
     }
 
-    struct sock *sk = (struct sock *)args->skaddr;
+    struct sock *const sk = (struct sock *)args->skaddr;
 
     if (args->oldstate == TCP_SYN_SENT || args->oldstate == TCP_SYN_RECV) {
         if (args->newstate == TCP_ESTABLISHED) {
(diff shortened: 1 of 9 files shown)


bpf/statsolly/tp_tcp.c
grcevski: Should we delete the map element here?

grcevski, line 128: Also do we need to delete here as well? if we cannot find place on the ring buffer should we clean up?

pinoOgni: oh yes

pinoOgni: yes again, thanks!

bpf/statsolly/tp_tcp.c
```diff
@@ -41,6 +41,18 @@ enum tcp_fail_reason {
...
 };
 
+enum tcp_handshake_role {
+    role_client = 1,
+    role_server = 2,
+};
+
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, void *);
```
mmat11: nit: can we add a comment signaling what kind of pointer this is? should we use `struct sock *` for clarity?

mmat11: also, does this belong in bpf/statsolly/maps?

bpf/statsolly/tp_tcp.c
```diff
@@ -100,20 +123,34 @@ int obi_tracepoint_inet_sock_set_state(struct trace_event_raw_inet_sock_set_stat
...
 
+    const u8 *role_ptr = bpf_map_lookup_elem(&sock_role, &sk);
+    if (role_ptr) {
+        se->role = *role_ptr;
+    } else if (args->oldstate == TCP_SYN_SENT) {
+        se->role = role_client;
+    } else if (args->oldstate == TCP_SYN_RECV) {
+        se->role = role_server;
+    } else {
+        se->role = 0;
```
mmat11: what's role `0`? should we add another constant like `unknown`/`unset`?

bpf/statsolly/tp_tcp.c
mmat11, line 113: I see the map is not cleaned up in some places, if we need to clean it up in every branch like a `defer` statement, I would add a `cleanup:` label at the end of the function and `goto` there instead

bpf/statsolly/tp_tcp.c
pinoOgni: > nit: can we add a comment signaling what kind of pointer this is? should we use `struct sock *` for clarity?

Oh yes!

> also, does this belong in bpf/statsolly/maps?

It could go there for consistency with stats_events map, but since it's only used within here and acts more like a per-tracepoint program map I just wrote it here. 
Update: I'll move in maps because I'll need it in another probe.

bpf/statsolly/tp_tcp.c
pinoOgni: Nice catch! I wrote it only in userspace.

grcevski (approved):
LGTM!

mmat11 (approved):
lgtm!

PR #2059, opened 2026-05-06
Improvments to weaver schema + tooling

NimrodAvni78:
## Summary

Split some of the work In preparation for https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2008 to make it easier to review, allowing us to have full coverage of telemetry schema on all of our integration tests

## Changes in the tooling:
1. Added `lint-schema` make command that validates our telemetry schema (the schema itself, not that data matches the schema), using weaver and making sure it passes without warnings. added it to both github CI, and dependencies image 
2. Added `fetch-upstream-semconv` make command that pre-fetches the otel semantic convention registry and caches it. this significantly improves CI time for weaver validation tests, by saving each test needing to fetch the schema from github, with this change this only happens once per shard. this also happens locally and is git ignored,  part of `prereqs` when preparing the project locally and on CI
3. Weaver now emits the output to a file to a mounted path, and we read it at the end of the test, instead of using `docker logs` and reading the entire large output. Saving a lot of CI time
4. Add a shared `weaver.yaml` docker-compose component to reuse in all tests

## Changes the schema itself:
1. Removed `intenal_spans` section, even though obi does emit the `in_queue` and `processing` spans, they do not have any attributes. Weaver's registry-check requires every group declaration to have either attributes or that it extends another group, not the case for us. If we decide to leave it we just will get many warnings every live-check validation
2. Added sctions for`netolly` (`network.yaml`) and`statsolly` (`stats.yaml`) metrics
3. Changed `service_graph` and `span_metrics` metric units to match what OBI is exporting
4. Add missing attributes obi is exporting like `obi.revision` and `obi.version`, also remove the temporary `rpc.server.duration` since it is not really working to re-declare it, we will just have conflicts. the solution is to ignore it in the weaver tests until we bump the semconv version we export.

####  currently in `python-redis` test, this will be the most of the changes in the upcoming PR
1. Increase `OTEL_EBPF_METRICS_INTERVAL` `"10ms" -> "1s"`. this generates dramatically less datapoints (as expected) which has the effect of cutting schema validation time by many minutes in CI. Plus does not negatively effect the tests.
2. Use the `weaver.yaml` docker-compose component
3. (not in this test), Switch all tests with `application_span` feature to `application_span_otel`. `application_span` is deprecated and the schema does not declare those metrics, we only declare the metrics exported by `application_span_otel`, so those are the ones we are testing

## Validation


grcevski (approved):
LGTM! Nice!

mmat11 (approved):
lgtm

PR #2065, opened 2026-05-07
Fix service name resolution

### LINKED ISSUE #1893: Attribute `service.name` is not deduced according to spec
Right now, the attribute `service.name` is not deduced according to the official otel recommendation. Source:

* https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#how-servicename-should-be-calculated

As a workaround, I am sending all collected signals to a local otelcol daemonset pod, and enrich with k8s metadata again.

### LINKED ISSUE #2057: Inconsistent service.name calculation between OBI and k8sattributes processor
### Component

OBI (eBPF instrumentation)

### What happened?

I'm trying to correlate my telemetry, but I'm running into an issue where `service.name` is calculated differently between OBI and the OTel collector's `k8sattributes` processor. 

In my cluster, I use OBI for eBPF telemetry alongside an OTel collector that uses the `filelog` receiver for pod logs (enriched by the `k8sattributes` processor). While looking at our dashboards, I noticed that for the exact same pod, the two tools are outputting different service names. 

The specific resource I noticed this on is `prometheus-k8s` in the `openshift-monitoring` namespace (the default managed Prometheus in OpenShift). I inspected the labels on the statefulset, and it has these two:
- `app.kubernetes.io/instance=k8s`
- `app.kubernetes.io/name=prometheus`

It looks like the `k8sattributes` processor evaluates the `instance` label first, setting the service name to `k8s`. OBI, however, seems to prioritize the `name` label and sets it to `prometheus`. Because this is a managed OpenShift resource, I can't just change the labels on the pod to work around it (also it is managed by a different team), and the mismatch breaks our ability to correlate logs and eBPF data.

I expected both tools to calculate the `service.name` the exact same way - ideally strictly following the [semconv docs on how service.name should be calculated](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#how-servicename-should-be-calculated). 

Honestly, I prefer OBI's calculation here (`prometheus` is a much more useful service name than `k8s`), but the core issue is that they don't match. It would be great if OBI aligned with how the collector handles the fallback order, or if there was at least a way to configure the label priority in OBI.

### Steps to reproduce

1. Run OBI to collect eBPF telemetry alongside an OTel collector using the `k8sattributes` processor for logs.
2. Observe a workload that has both instance and name labels where the values differ (like the default OpenShift `prometheus-k8s` statefulset).
3. Check the generated telemetry for that specific pod from both sources.
4. Notice that the `service.name` attribute differs between the two (e.g., OBI reports `prometheus`, while the collector reports `k8s`).

### OBI configuration

```yaml

```

### Relevant logs

```shell

```

### Additional context

_No response_

### Version

0.8.0

### Kernel version

_No response_

### OS / Distribution

OpenShift

### Deployment

Kubernetes

Orenico10:
## Summary

Changed the default configuration for pod labels to use for resolving service name from to be according to [the semantic conventions guidelines](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#how-servicename-should-be-calculated). Resolve #2057 & Resolve #1893

## Validation

### DIFF
--- a/pkg/kube/store.go
+++ b/pkg/kube/store.go
@@ -72,7 +72,9 @@ const (
 var DefaultResourceLabels = ResourceLabels{
 	// If a user sets useLabelsForResourceAttributes: false in its OTEL operator config, is the task of the
 	// OTEL operator to provide empty values for this.
-	"service.name":      []string{"app.kubernetes.io/name"},
+	// app.kubernetes.io/instance has higher precedence than app.kubernetes.io/name per the OTel spec:
+	// https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#how-servicename-should-be-calculated
+	"service.name":      []string{"app.kubernetes.io/instance", "app.kubernetes.io/name"},
 	"service.namespace": []string{"app.kubernetes.io/part-of"},
 	"service.version":   []string{"app.kubernetes.io/version"},
 }
--- a/pkg/kube/store_test.go
+++ b/pkg/kube/store_test.go
@@ -499,6 +499,74 @@ func TestServiceInfo(t *testing.T) {
 	assert.Equal(t, "namespaceA", k8sNamespace)
 }
 
+func TestServiceNameLabelPriority(t *testing.T) {
+	newPod := func(labels map[string]string) informer.ObjectMeta {
+		return informer.ObjectMeta{
+			Name:      "podA",
+			Namespace: "ns",
+			Ips:       []string{"1.2.3.4"},
+			Kind:      "Pod",
+			Labels:    labels,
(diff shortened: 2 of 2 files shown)


NimrodAvni78 (approved):
nice change
LGTM

grcevski (approved):
LGTM! Cool I also checked and the OTel Operator follows this algorithm as well:

https://github.com/open-telemetry/opentelemetry-operator/blob/f0bb76b5b8f27c1922be20c8e6d63be279d79a41/README.md?plain=1#L823

PR #2066, opened 2026-05-07
Make the database error response optional and off by default

grcevski:
## Summary

Adds new attribute similar to db.query.text called db.response.error which determines if we'll store database error responses in the `otel.span_description` field. By default this is off similar to the db.query.text field. To help users understand that we purposefully didn't add the field, I've added a placeholder text to instruct the users to enable the `db.response.error` field if they want to see more details on database errors.

## Validation

### DIFF
--- a/pkg/appolly/app/request/span.go
+++ b/pkg/appolly/app/request/span.go
@@ -80,6 +80,10 @@ const (
 	SchemeHostSeparator = ";"
 )
 
+const (
+	DBErrorMessagePlaceholder = "enable the db.response.error attribute for details"
+)
+
 type SQLKind uint8
 
 const (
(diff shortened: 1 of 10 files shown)


rafaelroquetto (approved):
LGTM

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -170,7 +171,7 @@ func GenerateTracesWithAttributes(
 		// Set status code
 		statusCode := CodeToStatusCode(request.SpanStatusCode(span))
 		s.Status().SetCode(statusCode)
-		statusMessage := request.SpanStatusMessage(span)
+		statusMessage := request.SpanStatusMessage(span, traceAttrs)
```
MrAlias: `GroupSpans()` has already materialized the per-span exported attributes before we get here, so using `traceAttrs` again for `status.message` introduces a second source of truth at this API boundary. That looks fine in OBI's current call path, but it does mean embedders now need to keep the selector map and the resolved span attributes in sync manually. I think this layer would be safer if it derived the decision from the resolved per-span data instead.

pkg/appolly/app/request/span.go
```diff
@@ -80,6 +80,10 @@ const (
 	SchemeHostSeparator = ";"
 )
 
+const (
+	DBErrorMessagePlaceholder = "enable the db.response.error attribute for details"
```
MrAlias: Please add the corresponding user-facing docs for `db.response.error` as part of this change, e.g. in `devdocs/`, examples, and any matching `opentelemetry.io` docs if this is surfaced there.

MrAlias (commented):
Thanks for tightening up the default behavior here. Making DB error text opt-in makes sense, and the new tests cover the selected vs unselected cases well.

I have two concerns:

1. `GenerateTracesWithAttributes` now depends on both the resolved per-span attributes and the original `traceAttrs` selector map. The in-repo call path keeps those aligned today, so I don't think this is a current behavior bug in the main pipeline, but it does make the public API easier to misuse for embedders because emission now depends on two representations of the same policy decision.

2. The new placeholder text tells users to enable `db.response.error`, so this change should also add user-facing documentation for that attribute.

marctc (approved):
LGTM!

marctc:
>     2. The new placeholder text tells users to enable `db.response.error`, so this change should also add user-facing documentation for that attribute.

Isn't documentation something that we do always after merging the PR? Or in this case has to be resolved in this PR?

MrAlias:
> Isn't documentation something that we do always after merging the PR? Or in this case has to be resolved in this PR?

I'm fine if that is the case. I would like to have a plan worked out before we merge as to what we plan to do for documentation though.

grcevski:
> > Isn't documentation something that we do always after merging the PR? Or in this case has to be resolved in this PR?
> 
> I'm fine if that is the case. I would like to have a plan worked out before we merge as to what we plan to do for documentation though.

I pushed an update for the duplicate map concern and I made sure this attribute isn't added to the traces, since it's not in the spec. it only affects the existence of the db.error in the span status message. We'll need to document that. I'm going to create an issue for this.

MrAlias (approved):
🥇

PR #2068, opened 2026-05-08
Fix slice bounds panic in Postgres BIND parser on truncated frame

keunwooo:
## Summary

Fix `slice bounds out of range` panic in `handlePostgres` BIND case when the
BPF ring buffer captures a Postgres BIND message truncated mid-portal.

## Stack trace (observed in production)

```
panic: runtime error: slice bounds out of range [13:12]

goroutine 497 [running]:
go.opentelemetry.io/obi/pkg/ebpf/common.handlePostgres(...)
  /src/vendor/go.opentelemetry.io/obi/pkg/ebpf/common/sql_detect_postgres.go:330
go.opentelemetry.io/obi/pkg/ebpf/common.dispatchPostgres(...)
  tcp_detect_transform.go:151
go.opentelemetry.io/obi/pkg/ebpf/common.ReadTCPRequestIntoSpan(...)
  tcp_detect_transform.go:61
```

## Root cause

`handlePostgres` BIND case (sql_detect_postgres.go:328-330):

```go
portal := strings.Clone(unix.ByteSliceToString(msg.data))
portalLen := len(portal) + 1   // +1 for the null terminator
stmtName := strings.Clone(unix.ByteSliceToString(msg.data[portalLen:]))
```

`unix.ByteSliceToString(msg.data)` reads to the end of the buffer when no null
terminator is present. When the BPF ring buffer truncates the BIND payload
mid-portal (TCP segmentation splits the BIND body across kernel reads), the
extracted portal occupies the entire buffer, so `portalLen = len(portal)+1`
becomes greater than `len(msg.data)`. The next slice expression
`msg.data[portalLen:]` panics.

## Trigger

Beyla 3.14.0 (current OBI HEAD as of 2026-05) on EKS amzn2023, kernel 6.12,
with a Rust application using ` sqlx`/`tokio-postgres` issuing prepared-statement
BIND messages to PostgreSQL. Observed crash rate ~1/min on the Beyla pod
co-located with the Postgres master node. Beyla pods on other nodes (with no
Postgres master, or with Postgres replicas only) did not panic.

## Fix

Add an explicit bounds check before the slice. When the captured frame contains
only the portal name (`portalLen >= len(msg.data)`), log a debug message and
`continue` with the next iterator message — there is no statement name to map
into `postgresPortals` because the body was truncated.

## Test

`TestPostgresHandleBindTruncatedDoesNotPanic` reproduces the truncated frame
(12-byte BIND body with no null terminator) and asserts that `handlePostgres`
does not panic. The test fails on `main` without the fix and passes with it.

## Files

- `pkg/ebpf/common/sql_detect_postgres.go` — 8-line bounds check
- `pkg/ebpf/common/sql_detect_postgres_test.go` — regression test

### DIFF
--- a/pkg/ebpf/common/sql_detect_postgres.go
+++ b/pkg/ebpf/common/sql_detect_postgres.go
@@ -327,6 +327,14 @@ Loop:
 		case "BIND":
 			portal := strings.Clone(unix.ByteSliceToString(msg.data))
 			portalLen := len(portal) + 1 // +1 for the null terminator
+			if portalLen >= len(msg.data) {
+				// BIND message truncated by BPF ring buffer: portal name fills the
+				// captured buffer, leaving no room for the statement name. Skip the
+				// statement-name lookup rather than panicking on the slice expression.
+				slog.Debug("Postgres BIND message truncated; skipping portal mapping",
+					"portal_len", portalLen, "buf_len", len(msg.data))
+				continue
+			}
 			stmtName := strings.Clone(unix.ByteSliceToString(msg.data[portalLen:]))
 
 			parseCtx.postgresPortals.Add(postgresPortalsKey{
--- a/pkg/ebpf/common/sql_detect_postgres.go
+++ b/pkg/ebpf/common/sql_detect_postgres.go
@@ -325,16 +325,14 @@ Loop:
 
 			continue
 		case "BIND":
-			portal := strings.Clone(unix.ByteSliceToString(msg.data))
-			portalLen := len(portal) + 1 // +1 for the null terminator
+			portalStr := unix.ByteSliceToString(msg.data)
+			portalLen := len(portalStr) + 1 // +1 for the null terminator
 			if portalLen >= len(msg.data) {
-				// BIND message truncated by BPF ring buffer: portal name fills the
-				// captured buffer, leaving no room for the statement name. Skip the
-				// statement-name lookup rather than panicking on the slice expression.
 				slog.Debug("Postgres BIND message truncated; skipping portal mapping",
 					"portal_len", portalLen, "buf_len", len(msg.data))
 				continue
 			}
+			portal := strings.Clone(portalStr)
 			stmtName := strings.Clone(unix.ByteSliceToString(msg.data[portalLen:]))
 
 			parseCtx.postgresPortals.Add(postgresPortalsKey{
--- a/pkg/ebpf/common/sql_detect_postgres_test.go
+++ b/pkg/ebpf/common/sql_detect_postgres_test.go
@@ -134,45 +134,23 @@ func TestPostgresMessagesIteratorNoAllocs(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


pkg/ebpf/common/sql_detect_postgres_test.go
```diff
@@ -132,3 +133,46 @@ func TestPostgresMessagesIteratorNoAllocs(t *testing.T) {
 		t.Errorf("MessageIterator allocated %v allocs per run; want 0", allocs)
 	}
 }
+
+// TestPostgresHandleBindTruncatedDoesNotPanic reproduces the production crash
```
marctc: nit: i dont think we need these big comments here. I'd personally remove them

mariomac: I agree with mark, instead of justifying this test from a very detailed bug that has happened in the past, I would just add a shorter and general comment like:

```
This test makes sure that if the BIND message is truncated at the eBPF side, OBI won't crash after trying to read beyond the buffer limits.
```

mariomac (commented):
LGTM! But I also recommend addressing @marctc comments before merging.

pkg/ebpf/common/sql_detect_postgres.go
```diff
@@ -327,6 +327,14 @@ Loop:
 		case "BIND":
 			portal := strings.Clone(unix.ByteSliceToString(msg.data))
 			portalLen := len(portal) + 1 // +1 for the null terminator
+			if portalLen >= len(msg.data) {
+				// BIND message truncated by BPF ring buffer: portal name fills the
```
mmat11: I would also remove this comment, it's clear from the debug message what's happening

rafaelroquetto:
@keunwooo thanks for your contribution - as you noticed, you got a bit of feedback regarding the excessive comments. For future reference (or if you would like to do another round here), you can point your AI agent to AGENTS.md - it has an explicit section about comments:

```
Comments must be minimal:

- Do not add comments that restate the code.
- Prefer clearer code over explanatory comments.
- Add comments only when they provide necessary context, explain non-obvious behavior, or document verifier, kernel, or ABI constraints.
```

pkg/ebpf/common/sql_detect_postgres.go
```diff
@@ -327,6 +327,14 @@ Loop:
...
+				// captured buffer, leaving no room for the statement name. Skip the
+				// statement-name lookup rather than panicking on the slice expression.
+				slog.Debug("Postgres BIND message truncated; skipping portal mapping",
+					"portal_len", portalLen, "buf_len", len(msg.data))
+				continue
```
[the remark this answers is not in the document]
rafaelroquetto: Yes please!

keunwooo:
Closing — the same bug was fixed on `main` in 8ac96fa5 with a more comprehensive solution (`parsePostgresBindNames` helper that validates both the portal and statement NUL terminators, plus unit tests covering the missing-portal-terminator case I was targeting here). No need to duplicate it.

Thanks to everyone who took the time to review.

PR #2069, opened 2026-05-08
fix(2043): fix #2043 url full path

### LINKED ISSUE #2043: HTTP client url.full includes query strings by default
## Summary

- HTTP client span generation now preserves query-bearing request URLs in `Span.FullPath` and uses that field to populate `url.full`. As a result, outbound HTTP query strings are exported by default in trace telemetry rather than being dropped from the URL field.

## Impact

- Secrets, tokens, signed URLs, or PII embedded in outbound query strings can be copied into telemetry backends.
- The issue affects current product code, not just test or example paths.
- The impact is limited to metadata exposure in telemetry and depends on applications putting sensitive values into query strings.

## Environment

- First identified in: 14b7e12e49abc22db7955a5853c662976ec230a8

## Evidence

- [pkg/ebpf/common/http_transform.go](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/pkg/ebpf/common/http_transform.go#L111-L116) stores the full request URI, including the query string, in `Span.FullPath`.
- [pkg/ebpf/common/http2grpc_transform.go](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/pkg/ebpf/common/http2grpc_transform.go#L404-L446) preserves the same full-path value in the HTTP/2 and gRPC-derived client span path.
- [pkg/export/otel/tracesgen/tracesgen.go](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/pkg/export/otel/tracesgen/tracesgen.go#L393-L407) exports `Span.FullPath` as `url.full` without redaction or query stripping.

## Steps to reproduce

1. From the repository root on `main`, create `/tmp/url_full_poc.go` with these exact contents:
   ```go
   package main
   import (
   	"fmt"
   	"go.opentelemetry.io/obi/pkg/appolly/app/request"
   	"go.opentelemetry.io/obi/pkg/export/otel/tracesgen"
   )
   func main() {
   	span := request.Span{Type: request.EventTypeHTTPClient, Method: "GET", Path: "/search", FullPath: "/search?token=secret&user=alice", Host: "example.com", HostPort: 443, Status: 200}
   	attrs := tracesgen.TraceAttributesSelector(&span, map[attr.Name]struct{}{})
   	for _, kv := range attrs {
   		if string(kv.Key) == string(attr.HTTPUrlPath) { fmt.Printf("url.path=%s\n", kv.Value.AsString()) }
   		if string(kv.Key) == string(attr.HTTPUrlFull) { fmt.Printf("url.full=%s\n", kv.Value.AsString()) }
   	}
   }
   ```
2. Build it with `GOTOOLCHAIN=auto go build -o /tmp/url_full_poc /tmp/url_full_poc.go`.
3. Run `/tmp/url_full_poc`.
4. Observe that the output shows `url.path=/search` but `url.full=/search?token=secret&user=alice`, proving the query string is exported in `url.full`.

## Suggested Fix Direction

- Redact or drop query strings from `url.full` by default, or
- make inclusion of query strings an explicit opt-in behavior, or
- add a first-class configuration knob for sanitizing URL query parameters before trace export.

## Acceptance Criteria

- Default HTTP client `url.full` export does not include query parameters.
- If query inclusion is supported, it is explicitly opt-in and clearly documented.
- Regression coverage verifies that `url.full` no longer leaks query strings by default.

## Note

NameHaibinZhang:
## Summary

fix #2043 

## Validation

### DIFF
--- a/pkg/export/attributes/names/attrs.go
+++ b/pkg/export/attributes/names/attrs.go
@@ -34,7 +34,7 @@ const (
 	HTTPURLScheme          = Name(semconv.URLSchemeKey)
 	HTTPUrlPath            = Name(semconv.URLPathKey)
 	HTTPUrlFull            = Name(semconv.URLFullKey)
-	HTTPUrlQuery           = Name("url.query")
+	HTTPUrlQuery           = Name(semconv.URLQueryKey)
 	ClientAddr             = Name(semconv.ClientAddressKey)
 	ServerAddr             = Name(semconv.ServerAddressKey)
 	ServerPort             = Name(semconv.ServerPortKey)
--- a/pkg/export/attributes/names/attrs.go
+++ b/pkg/export/attributes/names/attrs.go
@@ -34,6 +34,7 @@ const (
 	HTTPURLScheme          = Name(semconv.URLSchemeKey)
 	HTTPUrlPath            = Name(semconv.URLPathKey)
 	HTTPUrlFull            = Name(semconv.URLFullKey)
+	HTTPUrlQuery           = Name("url.query")
 	ClientAddr             = Name(semconv.ClientAddressKey)
 	ServerAddr             = Name(semconv.ServerAddressKey)
 	ServerPort             = Name(semconv.ServerPortKey)
--- a/pkg/export/otel/traces_test.go
+++ b/pkg/export/otel/traces_test.go
@@ -1641,6 +1641,33 @@ func TestGenerateTracesAttributes(t *testing.T) {
 		ensureTraceStrAttr(t, attrs, semconv.URLFullKey, "https://api.example.com/external/api?foo=bar")
 		ensureTraceStrAttr(t, attrs, semconv.URLQueryKey, "foo=bar")
 	})
+	t.Run("test HTTP client url.full includes query through selector config opt-in", func(t *testing.T) {
+		tr := makeTracesTestReceiver([]instrumentations.Instrumentation{instrumentations.InstrumentationHTTP})
+		tr.selectorCfg = &attributes.SelectorConfig{
+			SelectionCfg: attributes.Selection{
+				"traces": attributes.InclusionLists{
+					Include: []string{"url.query"},
+				},
+			},
+		}
+
+		traces := generateTracesForSpans(t, tr, []request.Span{{
+			Type:      request.EventTypeHTTPClient,
+			Method:    "GET",
+			Path:      "/external/api",
+			FullPath:  "/external/api?foo=bar",
+			Status:    200,
+			Host:      "api.example.com",
+			HostPort:  443,
+			Statement: "https;api.example.com",
+		}})
+
+		require.Len(t, traces, 1)
+
+		attrs := traces[0].ResourceSpans().At(0).ScopeSpans().At(0).Spans().At(0).Attributes()
+		ensureTraceStrAttr(t, attrs, semconv.URLFullKey, "https://api.example.com/external/api?foo=bar")
+		ensureTraceStrAttr(t, attrs, semconv.URLQueryKey, "foo=bar")
+	})
 	t.Run("test HTTP client url.full falls back to Path when FullPath is empty", func(t *testing.T) {
 		span := request.Span{
 			Type:      request.EventTypeHTTPClient,
--- a/pkg/export/otel/traces_test.go
+++ b/pkg/export/otel/traces_test.go
@@ -1619,6 +1619,25 @@ func TestGenerateTracesAttributes(t *testing.T) {
 		traces := tracesgen.GenerateTracesWithAttributes(cache, &span.Service, []attribute.KeyValue{}, hostID, groupFromSpanAndAttributes(&span, tAttrs), reporterName)
 
 		attrs := traces.ResourceSpans().At(0).ScopeSpans().At(0).Spans().At(0).Attributes()
+		ensureTraceStrAttr(t, attrs, semconv.URLFullKey, "https://api.example.com/external/api")
+	})
+	t.Run("test HTTP client url.full includes query with opt-in", func(t *testing.T) {
+		span := request.Span{
+			Type:      request.EventTypeHTTPClient,
+			Method:    "GET",
+			Path:      "/external/api",
+			FullPath:  "/external/api?foo=bar",
+			Status:    200,
+			Host:      "api.example.com",
+			HostPort:  443,
+			Statement: "https;api.example.com",
+		}
+
+		optionalAttrs := map[attr.Name]struct{}{attr.HTTPUrlQuery: {}}
+		tAttrs := tracesgen.TraceAttributesSelector(&span, optionalAttrs)
+		traces := tracesgen.GenerateTracesWithAttributes(cache, &span.Service, []attribute.KeyValue{}, hostID, groupFromSpanAndAttributes(&span, tAttrs), reporterName)
+
+		attrs := traces.ResourceSpans().At(0).ScopeSpans().At(0).Spans().At(0).Attributes()
 		ensureTraceStrAttr(t, attrs, semconv.URLFullKey, "https://api.example.com/external/api?foo=bar")
 	})
 	t.Run("test HTTP client url.full falls back to Path when FullPath is empty", func(t *testing.T) {
@@ -1651,7 +1670,7 @@ func TestGenerateTracesAttributes(t *testing.T) {
 		traces := tracesgen.GenerateTracesWithAttributes(cache, &span.Service, []attribute.KeyValue{}, hostID, groupFromSpanAndAttributes(&span, tAttrs), reporterName)
 
 		attrs := traces.ResourceSpans().At(0).ScopeSpans().At(0).Spans().At(0).Attributes()
-		ensureTraceStrAttr(t, attrs, semconv.URLFullKey, "https://upstream.example.com/external/api?foo=bar")
+		ensureTraceStrAttr(t, attrs, semconv.URLFullKey, "https://upstream.example.com/external/api")
 	})
 	t.Run("test JSON-RPC server span with error", func(t *testing.T) {
 		span := request.Span{
--- a/pkg/export/otel/traces_test.go
+++ b/pkg/export/otel/traces_test.go
@@ -1639,6 +1639,7 @@ func TestGenerateTracesAttributes(t *testing.T) {
 
 		attrs := traces.ResourceSpans().At(0).ScopeSpans().At(0).Spans().At(0).Attributes()
 		ensureTraceStrAttr(t, attrs, semconv.URLFullKey, "https://api.example.com/external/api?foo=bar")
+		ensureTraceStrAttr(t, attrs, semconv.URLQueryKey, "foo=bar")
 	})
 	t.Run("test HTTP client url.full falls back to Path when FullPath is empty", func(t *testing.T) {
 		span := request.Span{
--- a/pkg/export/otel/tracesgen/tracesgen.go
+++ b/pkg/export/otel/tracesgen/tracesgen.go
@@ -465,8 +465,10 @@ func TraceAttributesSelector(span *request.Span, optionalAttrs map[attr.Name]str
 		}
 		// Strip query parameters from url.full by default to avoid leaking
 		// sensitive data (tokens, PII). Include them only when explicitly opted in.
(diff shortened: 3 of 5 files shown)


marctc (approved):

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -462,6 +463,13 @@ func TraceAttributesSelector(span *request.Span, optionalAttrs map[attr.Name]str
 		if span.FullPath != "" {
 			urlPath = span.FullPath
 		}
+		// Strip query parameters from url.full by default to avoid leaking
+		// sensitive data (tokens, PII). Include them only when explicitly opted in.
+		if _, ok := optionalAttrs[attr.HTTPUrlQuery]; !ok {
+			if idx := strings.IndexByte(urlPath, '?'); idx > 0 {
```
mariomac: Are we stripping the URL query only if `url.query` attribute is opted in? Shouldn't we strip it always?

mariomac (commented):
Thanks for the contribution!

pkg/export/otel/tracesgen/tracesgen.go
NameHaibinZhang: @mariomac  The logic is actually the inverse: we always strip query parameters by default (when url.query is not opted in). We only keep them in url.full when the user explicitly opts in to url.query.The rationale is: opting in to url.query signals that the user explicitly accepts the privacy implications of exposing query parameters, so it's consistent to also include them in url.full (per OTel semantic conventions, url.full should represent the complete URL). If we always stripped from url.full even when opted in, it would be inconsistent — the user asked for query data but url.full wouldn't reflect it.

pkg/export/attributes/names/attrs.go
```diff
@@ -34,6 +34,7 @@ const (
 	HTTPURLScheme          = Name(semconv.URLSchemeKey)
 	HTTPUrlPath            = Name(semconv.URLPathKey)
 	HTTPUrlFull            = Name(semconv.URLFullKey)
+	HTTPUrlQuery           = Name("url.query")
```
mmat11: you can use `semconv.URLQueryKey`

pkg/export/otel/tracesgen/tracesgen.go
mmat11: reading https://opentelemetry.io/docs/specs/semconv/http/http-spans/ I think we should also populate `url.query` when opted-in and always redact sensitive data:

```
Sensitive content provided in url.full SHOULD be scrubbed when instrumentations can identify it.
Development Query string values for the following keys SHOULD be redacted by default and replaced by the value REDACTED: *see below

This list is subject to change over time.
Matching of query parameter keys against the sensitive list SHOULD be case-sensitive.
```

  [AWSAccessKeyId](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html#RESTAuthenticationQueryStringAuth)
  [Signature](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html#RESTAuthenticationQueryStringAuth)
  [sig](https://learn.microsoft.com/azure/storage/common/storage-sas-overview#sas-token)
  [X-Goog-Signature](https://cloud.google.com/storage/docs/access-control/signed-urls)
***

also, when handling `url.query` it's written we may give users a way to signal additional sensitive parameters:
```
Instrumentation MAY provide a way to override this list via declarative configuration. If so, it SHOULD use the sensitive_query_parameters property (an array of case-sensitive strings with minimum items 0) under .instrumentation/development.general.sanitization.url. This list is a full override of the default sensitive query parameter keys, it is not a list of keys in addition to the defaults.

When a query string value is redacted, the query string key SHOULD still be preserved, e.g. q=OpenTelemetry&sig=REDACTED.
```
I think this should be a follow-up

NameHaibinZhang: Thanks for digging into the spec! You're right — the spec recommends selective redaction of known sensitive keys (preserving the key, replacing the value) rather than full stripping. Our current approach is more conservative (strip everything by default), which is safe but less informative.I agree this should be a follow-up: implementing per-key redaction with the default sensitive list (AWSAccessKeyId, Signature, sig, X-Goog-Signature) and also populating url.query as a separate attribute when opted in.For now, the full strip provides a safe baseline. I'll open a follow-up issue to track the spec-compliant redaction behavior.

pkg/export/attributes/names/attrs.go
NameHaibinZhang: Done. Updated to use semconv.URLQueryKey.

pkg/export/otel/tracesgen/tracesgen.go
NameHaibinZhang: @mmat11 #2073

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -462,6 +463,13 @@ func TraceAttributesSelector(span *request.Span, optionalAttrs map[attr.Name]str
 		if span.FullPath != "" {
 			urlPath = span.FullPath
 		}
+		// Strip query parameters from url.full by default to avoid leaking
+		// sensitive data (tokens, PII). Include them only when explicitly opted in.
+		if _, ok := optionalAttrs[attr.HTTPUrlQuery]; !ok {
```
MrAlias: This check makes sense, but I think the opt-in is still unreachable through the normal config path.

`optionalAttrs` comes from `UserSelectedAttributes -> NewAttrSelector(...).For(attributes.Traces)`, and that selector only returns attributes registered in the trace definitions table. As written, `attr.HTTPUrlQuery` is checked here, but it is not added to the trace selector definitions, so including `url.query` in `attributes.select.traces` will never put it into `optionalAttrs`.

In practice that means `url.full` will still always be stripped, even when the user tries to opt in. I think this PR also needs the trace attribute registry updated so `url.query` can actually be selected.

pkg/export/otel/traces_test.go
```diff
@@ -1618,6 +1618,25 @@ func TestGenerateTracesAttributes(t *testing.T) {
...
+			Method:    "GET",
+			Path:      "/external/api",
+			FullPath:  "/external/api?foo=bar",
+			Status:    200,
+			Host:      "api.example.com",
+			HostPort:  443,
+			Statement: "https;api.example.com",
+		}
+
+		optionalAttrs := map[attr.Name]struct{}{attr.HTTPUrlQuery: {}}
```
MrAlias, line 1636: This test passes because it manually constructs `optionalAttrs` with `attr.HTTPUrlQuery` and passes that directly into `TraceAttributesSelector`.

That bypasses the real config path, where `optionalAttrs` is built by `UserSelectedAttributes -> NewAttrSelector(...).For(attributes.Traces)`. Since `url.query` is not registered in the trace selector definitions, users still cannot produce this state through `attributes.select`, so this test currently proves the local branch logic but not the end-to-end opt-in behavior.

MrAlias (changes requested):
This looks good overall and it fixes a real issue around `url.full` preferring the captured full path. I do think there’s one blocking gap in the attribute-selection plumbing, though: the new `attr.HTTPUrlQuery` opt-in is used to control whether the query string is preserved, but it isn’t wired into the trace attribute selector definitions yet, so users still can’t actually enable it through `attributes.select`.

mmat11:
one last thing is not clear to me: if a user opts-in for `url.query`, shouldn't we fill that attribute?

NameHaibinZhang:
> one last thing is not clear to me: if a user opts-in for `url.query`, shouldn't we fill that attribute?

Good point! I've added the url.query attribute population when opted in. Now when a user opts in to url.query, we both preserve the query string in url.full and emit url.query as a separate attribute with just the query string portion.For the selective redaction of sensitive keys (AWSAccessKeyId, Signature, sig, X-Goog-Signature) as described in the spec, I'll track that in a follow-up issue since it's a larger change.

pkg/export/otel/tracesgen/tracesgen.go
NameHaibinZhang: Good point. I've added attr.HTTPUrlQuery (with default false) to the Traces.Section definition in attr_defs.go, so it's now registered in the trace attribute registry and can be selected through attributes.select.traces.

pkg/export/otel/traces_test.go
NameHaibinZhang: You're right that this test only validates the local branch logic. Now that attr.HTTPUrlQuery is registered in the trace selector definitions (added in the latest push), I'll also add an end-to-end test that exercises the real config path through UserSelectedAttributes to verify the opt-in works as expected.

pkg/export/otel/traces_test.go
```diff
@@ -1618,6 +1618,25 @@ func TestGenerateTracesAttributes(t *testing.T) {
...
+		tAttrs := tracesgen.TraceAttributesSelector(&span, optionalAttrs)
+		traces := tracesgen.GenerateTracesWithAttributes(cache, &span.Service, []attribute.KeyValue{}, hostID, groupFromSpanAndAttributes(&span, tAttrs), reporterName)
+
 		attrs := traces.ResourceSpans().At(0).ScopeSpans().At(0).Spans().At(0).Attributes()
 		ensureTraceStrAttr(t, attrs, semconv.URLFullKey, "https://api.example.com/external/api?foo=bar")
```
mmat11, line 1641: can we add an assertion for the url query attribute?

NameHaibinZhang: Done — added ensureTraceStrAttr(t, attrs, semconv.URLQueryKey, "foo=bar") to assert that the url.query attribute is correctly populated when opted in.

pkg/export/otel/traces_test.go
```diff
@@ -1618,8 +1618,28 @@ func TestGenerateTracesAttributes(t *testing.T) {
...
+		ensureTraceStrAttr(t, attrs, semconv.URLQueryKey, "foo=bar")
 	})
```
MrAlias, line 1643: Could we add an end-to-end variant here that goes through `UserSelectedAttributes` / `selectorCfg` instead of constructing `optionalAttrs` directly?

```suggestion
	t.Run("test HTTP client url.full includes query through selector config opt-in", func(t *testing.T) {
		tr := makeTracesTestReceiver([]instrumentations.Instrumentation{instrumentations.InstrumentationHTTP})
		tr.selectorCfg = &attributes.SelectorConfig{
			SelectionCfg: attributes.Selection{
				"traces": attributes.InclusionLists{
					Include: []string{"url.query"},
				},
			},
		}

		traces := generateTracesForSpans(t, tr, []request.Span{{
			Type:      request.EventTypeHTTPClient,
			Method:    "GET",
			Path:      "/external/api",
			FullPath:  "/external/api?foo=bar",
			Status:    200,
			Host:      "api.example.com",
			HostPort:  443,
			Statement: "https;api.example.com",
		}})

		require.Len(t, traces, 1)

		attrs := traces[0].ResourceSpans().At(0).ScopeSpans().At(0).Spans().At(0).Attributes()
		ensureTraceStrAttr(t, attrs, semconv.URLFullKey, "https://api.example.com/external/api?foo=bar")
		ensureTraceStrAttr(t, attrs, semconv.URLQueryKey, "foo=bar")
	})
```

MrAlias (commented):
One remaining test gap is the end-to-end selector-config path for `url.query`.

pkg/export/otel/traces_test.go
MrAlias: Or, if you don't want to replace the test, add the proposed one in addition.

mmat11 (approved):
lgtm

pkg/export/otel/traces_test.go
NameHaibinZhang: Good idea! I've added the end-to-end test alongside the existing one. The new test ("test HTTP client url.full includes query through selector config opt-in") exercises the full config path through makeTracesTestReceiver → selectorCfg → generateTracesForSpans → UserSelectedAttributes, confirming that including "url.query" in attributes.select.traces correctly flows through to optionalAttrs and preserves both url.full with query and the url.query attribute.

MrAlias (approved):
🚀

PR #2070, opened 2026-05-08
refactor: add private field routeMatcher with atomic pointer

### LINKED ISSUE #2026: Delayed Java matcher update races with span routing
## Summary

Java route harvesting now updates `Service.HarvestedRouteMatcher` asynchronously after instrumentation begins. The delayed callback writes into shared `*svc.Attrs` state while other goroutines can concurrently copy and read the same service attributes during PID filtering and route decoration.

## Impact

This can produce inconsistent route attribution or crash the OBI agent through a timing-dependent data race. The impact is limited to telemetry integrity and agent availability, so `low` severity is appropriate.

## Environment

- First identified in: 7ed8b95f48191c025d1b364ef4f95e57bbfb9f66

## Evidence

- `pkg/appolly/discover/attacher.go` uses `time.AfterFunc` and later calls `ie.FileInfo.Service.SetHarvestedRoutes(m)`.
- `pkg/appolly/app/svc/svc.go` assigns `HarvestedRouteMatcher` without synchronization.
- `pkg/ebpf/common/pids.go` copies `*info.service` while sharing the underlying `*svc.Attrs`.
- `pkg/transform/routes.go` reads `s.Service.HarvestedRouteMatcher` during route decoration.

## Steps to reproduce

1. From the repository root on `main`, create `/tmp/racepoc_test.go` with this exact test:
   ```go
   package racepoc
   import (
       "sync"
       "testing"
       "go.opentelemetry.io/obi/pkg/appolly/app/svc"
       "go.opentelemetry.io/obi/pkg/internal/transform/route"
   )
   func TestHarvestedRoutesRace(t *testing.T) {
       attrs := &svc.Attrs{}
       var wg sync.WaitGroup
       wg.Add(2)
       go func() {
           defer wg.Done()
           for i := 0; i < 10000; i++ {
               attrs.SetHarvestedRoutes(route.NewMatcher([]string{"/users/:id"}))
           }
       }()
       go func() {
           defer wg.Done()
           for i := 0; i < 10000; i++ {
               if attrs.HarvestedRouteMatcher != nil {
                   _ = attrs.HarvestedRouteMatcher.Find("/users/123")
               }
           }
       }()
       wg.Wait()
   }
   ```
2. Run the race PoC with `GOTOOLCHAIN=auto go test -race /tmp/racepoc_test.go`.
3. Let the test repeatedly call `SetHarvestedRoutes` while another goroutine reads the same matcher.
4. Observe that the Go race detector reports a data race on `HarvestedRouteMatcher`, proving the unsynchronized update path.

## Suggested Fix Direction

Make harvested route matcher updates synchronized with readers or publish a replacement service snapshot instead of mutating shared `*svc.Attrs` in place.

## Acceptance Criteria

- Delayed route harvesting no longer mutates shared route matcher state unsafely.
- Concurrent span processing cannot race with harvested route matcher publication.
- Tests cover the delayed Java harvest path and concurrent routing reads.

## Note

rupeshkoushik07:
## Summary
Fixes [#2026](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2026#event-25270826003)

Replace the plain `route.Matcher` field with `*atomic.Pointer[route.Matcher]` so concurrent writes from the Java harvest timer and reads from `PIDsFilter.Filter` are synchronized without changing how `svc.Attrs` is used elsewhere.

## Changes Made

**`pkg/appolly/app/svc/svc.go`**
Replaced the plain `route.Matcher` field with `routeMatcher *atomic.Pointer[route.Matcher]` because a Go interface is two words and copying it while another goroutine writes causes a torn read. The atomic pointer makes the load and store indivisible. Added `InitHarvestedRoutes`, `SetHarvestedRoutes` and `HarvestedRouteMatcher` to manage access safely.

**`pkg/appolly/discover/attacher.go`**
Added `InitHarvestedRoutes()` before `harvestRoutes` on both code paths because the atomic container must exist before the harvest timer goroutine can write to it. Without this, `SetHarvestedRoutes` would hit a nil pointer.

**`pkg/transform/routes.go`**
Updated to call `HarvestedRouteMatcher()` instead of reading the field directly because the field is now unexported. The method returns the same `route.Matcher` type so the behaviour is unchanged.

**`pkg/ebpf/common/pids_race_test.go`**
Added a regression test that runs `SetHarvestedRoutes` and `PIDsFilter.Filter` concurrently to prove the race no longer fires under `go test -race`.

## Validation

### DIFF
--- a/pkg/appolly/app/svc/svc.go
+++ b/pkg/appolly/app/svc/svc.go
@@ -4,6 +4,8 @@
 package svc // import "go.opentelemetry.io/obi/pkg/appolly/app/svc"
 
 import (
+	"sync/atomic"
+
 	"go.opentelemetry.io/otel/sdk/trace"
 	semconv "go.opentelemetry.io/otel/semconv/v1.38.0"
 
@@ -129,7 +131,12 @@ type Attrs struct {
 
 	CustomInRouteMatcher  route.Matcher
 	CustomOutRouteMatcher route.Matcher
-	HarvestedRouteMatcher route.Matcher
+	// harvestedRouteMatcher is written by the Java route-harvest timer goroutine and
+	// read concurrently by the span-filter goroutine. Storing it behind an
+	// *atomic.Pointer keeps Attrs copy-safe (only the 8-byte pointer is copied) while
+	// making the actual load/store atomic. Callers must use InitHarvestedRoutes before
+	// the first SetHarvestedRoutes call, and HarvestedRoutes to read.
+	harvestedRouteMatcher *atomic.Pointer[route.Matcher]
 	PathTrie              *clusterurl.PathTrie
 }
 
@@ -188,8 +195,31 @@ func (i *Attrs) ExportsOTelTraces() bool {
 	return i.getFlag(exportsOTelTraces)
 }
 
+// InitHarvestedRoutes allocates the atomic storage for the harvested route
+// matcher. It must be called once on the owning *Attrs before any concurrent
+// access (i.e. before the process is registered with the PIDsFilter).
+func (i *Attrs) InitHarvestedRoutes() {
+	i.harvestedRouteMatcher = &atomic.Pointer[route.Matcher]{}
+}
+
+// SetHarvestedRoutes atomically stores the harvested route matcher.
+// It is safe to call concurrently with HarvestedRouteMatcher.
 func (i *Attrs) SetHarvestedRoutes(matcher route.Matcher) {
-	i.HarvestedRouteMatcher = matcher
+	if i.harvestedRouteMatcher != nil {
+		i.harvestedRouteMatcher.Store(&matcher)
+	}
+}
+
+// HarvestedRouteMatcher atomically loads the harvested route matcher.
+// Returns nil if no matcher has been set yet.
+func (i *Attrs) HarvestedRouteMatcher() route.Matcher {
+	if i.harvestedRouteMatcher == nil {
+		return nil
+	}
+	if p := i.harvestedRouteMatcher.Load(); p != nil {
+		return *p
+	}
+	return nil
 }
 
 func (i *Attrs) SetCustomRoutes(config *services.CustomRoutesConfig) {
--- a/pkg/appolly/app/svc/svc.go
+++ b/pkg/appolly/app/svc/svc.go
@@ -131,13 +131,11 @@ type Attrs struct {
 
 	CustomInRouteMatcher  route.Matcher
 	CustomOutRouteMatcher route.Matcher
-	// routeMatcher is written by the Java route-harvest timer goroutine and
-	// read concurrently by the span-filter goroutine. Storing it behind an
-	// *atomic.Pointer keeps Attrs copy-safe (only the 8-byte pointer is copied) while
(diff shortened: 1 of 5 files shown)


marctc:
hey @rupeshkoushik07, are you agents reading AGENTS.md? The comments provided here are long and verbose (and even not necessary for tests), do you mind fix it?

rafaelroquetto (changes requested):
thanks for the fix, the atomic.Pointer does solve the underlying torn-read race, but `InitHarvestedRoutes` itself is actually not thread-safe: the nil-check-then-assign isn't synchronised, it only works because today it happens to be called from the single-threaded discovery path before anyone else can see the struct. feels fragile, and the panic in `SetHarvestedRoutes` kind of proves the point.

If we take a step back, I reckon the real issue is that nobody really owns `svc.Attrs`, it's shared between discovery, `PIDsFilter`, the harvest timer and the span pipeline, with field-level mutations from multiple goroutines. I had a quick grep and the harvest is actually the only post-publication writer to the shared struct, everything else either runs before AllowPID or operates on a copy - given that, would it make more sense to make Attrs immutable after publication and route the harvest through a CoW update via a single `atomic.Pointer[svc.Attrs]`?  That protects every field for free, gets rid of the Init/panic dance, and the migration is small since there's only one writer to move.

rupeshkoushik07:
> thanks for the fix, the atomic.Pointer does solve the underlying torn-read race, but `InitHarvestedRoutes` itself is actually not thread-safe: the nil-check-then-assign isn't synchronised, it only works because today it happens to be called from the single-threaded discovery path before anyone else can see the struct. feels fragile, and the panic in `SetHarvestedRoutes` kind of proves the point.
> 
> If we take a step back, I reckon the real issue is that nobody really owns `svc.Attrs`, it's shared between discovery, `PIDsFilter`, the harvest timer and the span pipeline, with field-level mutations from multiple goroutines. I had a quick grep and the harvest is actually the only post-publication writer to the shared struct, everything else either runs before AllowPID or operates on a copy - given that, would it make more sense to make Attrs immutable after publication and route the harvest through a CoW update via a single `atomic.Pointer[svc.Attrs]`? That protects every field for free, gets rid of the Init/panic dance, and the migration is small since there's only one writer to move.

Thanks for the review and Yes, I agree that fixing the matcher with `atomic.Pointer` removes the torn read when Filter copies `svc.Attrs`, but it doesn’t really establish ownership or make `InitHarvestedRoutes` correctness part of the type system; it depends on callers getting order right, checked only by a panic.

And `svc.Attrs `shouldn’t be a shared mutable after we publish it. Immutable published attrs + updates via CoW `(atomic.Pointer[svc.Attrs])` seems to be much clearer, fixes the whole struct read/copy story, and drops the Init/panic if every post-publish change goes through that path.

rafaelroquetto:
Closing. There are multiple issues with the proposed code (e.g. `Attrs` referencing itself or race conditions that still remain). Happy to reopen if you can walk me through an intended and sane design.

I'd like to take the opportunity to invite you to revisit our [Contributing Guidelines](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/CONTRIBUTING.md#contribution-guidelines)

PR #2080, opened 2026-05-09
[PoC] Integrate Config v2

MrAlias:
Follow up to #1351


MrAlias:
Concern is expressed about this languishing. We need to progress this.

@skl is going to take a look at the config v2 parity checker and add that as part of the CI to help new development of the configs progress in parallel.

MrAlias:
Tracking integration of this PoC in #2251

PR #2081, opened 2026-05-10
Fix stats RTT otel buckets + Make stats test use native OTLP export

NimrodAvni78:
## Summary

Related to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2008

Currently when exporting stats RTT metrics via otlp the buckets dont really make sense for this ms operation.

Changes Made:
- Match Otel histogram buckets to match promethues buckets for TCP RTT metric
- Changed integration test to evaluate against otlp exported metrics instead of prometheus metrics
- Changed the test not to rely on an exact le=0.1, since it is fragile and depends on the histogras exact buckets, changed it to avg

## Validation

### DIFF
--- a/internal/test/integration/suites_stat_test.go
+++ b/internal/test/integration/suites_stat_test.go
@@ -14,7 +14,7 @@ import (
 
 func TestStat_GoStatMetrics(t *testing.T) {
 	compose, err := docker.ComposeSuite("docker-compose-go-stat-metrics.yml", path.Join(pathOutput, "test-suite-go-stat-metrics.log"))
-	compose.Env = append(compose.Env, `TEST_SERVICE_PORTS=8381:8080`, `OTEL_EBPF_CONFIG_SUFFIX=-go-stat-metrics`, `PROM_CONFIG_SUFFIX=-promscrape`)
+	compose.Env = append(compose.Env, `TEST_SERVICE_PORTS=8381:8080`, `OTEL_EBPF_CONFIG_SUFFIX=-go-stat-metrics`, `PROM_CONFIG_SUFFIX=-promscrape-otel`)
 	require.NoError(t, err)
 	require.NoError(t, compose.Up())
 	t.Run("Go Stat Metrics TCP RTT tests", testStatMetricsTCPRttGo)
--- a/pkg/export/otel/metrics_stats.go
+++ b/pkg/export/otel/metrics_stats.go
@@ -135,7 +135,11 @@ func newStatMetricsExporter(
 	if cfg.CommonCfg.Features.StatsTCPRtt() {
 		log := log.With("metricFamily", "StatsTCPRtt")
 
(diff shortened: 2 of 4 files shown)


pkg/export/otel/metrics_stats.go
```diff
@@ -135,7 +135,11 @@ func newStatMetricsExporter(
 	if cfg.CommonCfg.Features.StatsTCPRtt() {
 		log := log.With("metricFamily", "StatsTCPRtt")
 
-		tcpRtt, err := ebpfEvents.Float64Histogram(attributes.StatTCPRtt.OTEL, metric2.WithUnit("s"))
+		tcpRtt, err := ebpfEvents.Float64Histogram(
+			attributes.StatTCPRtt.OTEL,
+			metric2.WithUnit("s"),
+			metric2.WithExplicitBucketBoundaries(0.0005, 0.001, 0.002, 0.005, 0.010, 0.025, 0.050, 0.100, 0.250, 0.500, 1.0),
+		)
```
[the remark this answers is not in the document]
NimrodAvni78: i think this requires a bigger change of syncronizing between otlp and prom metrics export, prefer to keep it duplicate for now

internal/test/integration/suites_stat_test.go
```diff
@@ -14,7 +14,7 @@ import (
 
 func TestStat_GoStatMetrics(t *testing.T) {
 	compose, err := docker.ComposeSuite("docker-compose-go-stat-metrics.yml", path.Join(pathOutput, "test-suite-go-stat-metrics.log"))
-	compose.Env = append(compose.Env, `TEST_SERVICE_PORTS=8381:8080`, `OTEL_EBPF_CONFIG_SUFFIX=-go-stat-metrics`, `PROM_CONFIG_SUFFIX=-promscrape`)
+	compose.Env = append(compose.Env, `TEST_SERVICE_PORTS=8381:8080`, `OTEL_EBPF_CONFIG_SUFFIX=-go-stat-metrics`, `PROM_CONFIG_SUFFIX=-promscrape-otel`)
```
pinoOgni, line 17: If it is not needed anymore I think we can delete `prometheus-config-promscrape.yml`

pinoOgni (approved):
lgtm! I left only a comment

mariomac (approved):

internal/test/integration/suites_stat_test.go
skl: Why not have both and two tests, one for prometheus export and one for OTLP export? Otherwise we lose the existing prometheus export coverage.

NimrodAvni78: @skl I think for most tests we use otel export and i think it makse sense to keep it since we meant to be otlp first
otherwise we need to duplicate each test, or at least duplicate the assertions

PR #2086, opened 2026-05-11
fix(#2076): make db.response.error attribute selection consistent with other optional attributes

### LINKED ISSUE #2076: Document the functionality of db.response.error
As per the comments in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2066#issuecomment-4409649260. We should document the `db.response.error` optional attribute and how it behaves.

NameHaibinZhang:
## Description

When `db.response.error` is not included in `attributes.select.traces`, the span status message is now left empty instead of showing a placeholder hint. This makes the behavior consistent with how other optional attributes (e.g. `db.query.text`) work — when not selected, they are simply omitted.

## Changes

- Remove the `DBErrorMessagePlaceholder` constant
- Update `DBResponseErrorAttr()` to return nil when the attribute is not selected
- Update config schema description to clarify that `attributes.select` applies to both metrics and traces
- Update tests to reflect the new behavior

Fixes #2076


devdocs/db-response-error.md
```diff
@@ -0,0 +1,39 @@
+# Trace Selection and `db.response.error`
```
NimrodAvni78: I would not have db.response.error in the title, i would just rename the devdocs to something more general like `Attribute selection`

devdocs/attribute-selection.md
```diff
@@ -0,0 +1,39 @@
+# Trace Selection and `db.response.error`
+
+## Trace selection (`traces`)
+
+For exported OpenTelemetry traces, use the `traces` key (not a metric name) under
+`attributes.select`. It controls optional trace decoration such as `db.query.text`,
+`url.query`, GenAI payload attributes, and **`db.response.error`**.
+
+## `db.response.error`
```
NimrodAvni78: 2 questions regarding implementation rather then the docs, before we write it
@grcevski 

1. why do we call it `db.response.error`? this can apply to all types of spans not only database spans, maybe we should call it `status.message` instead?
2. why do we add a placeholder for the attribute `enable the db.response.error attribute for details`, won't it just produce unnecessary data for customers, we should just leave it empty

devdocs/db-response-error.md
NameHaibinZhang: Done. Renamed the file to `devdocs/attribute-selection.md` and updated the title to "Attribute Selection" to keep it general.

devdocs/attribute-selection.md
NameHaibinZhang: Good questions. Some context on the current implementation:

1. **Naming**: The name `db.response.error` is intentionally scoped to database spans. In `tracesgen.go`, the status message logic is gated by `span.IsDBSpan()` — for non-DB spans a separate `SpanStatusMessage` path is used. So the configuration flag accurately reflects its scope. That said, if we want to generalize error detail opt-in to other span types in the future, renaming to something like `status.message` could make sense — deferring to @grcevski on whether that's the direction we want.

2. **Placeholder**: The placeholder `enable the db.response.error attribute for details` was added as an intentional hint so users can see that richer error info is available but gated. I agree it adds data to every failed DB span even when the user hasn't opted in. Whether to keep the hint or leave it empty is a trade-off between discoverability and data minimalism — happy to remove it if that's the consensus.

Both are design decisions on the existing implementation rather than this documentation PR. Deferring to @grcevski for the final call on these.

NimrodAvni78: regarding 2, for example when users don't select the `db.query.text` we don't add a hint saying to enable it in order to get the full payload, so I think it should be the same here

NameHaibinZhang: Good point — you're right that this should be consistent with how other optional attributes behave. When `db.query.text` isn't selected we simply omit it, so `db.response.error` should follow the same pattern.

Fixed in 8197e9b: removed the placeholder message from `DBResponseErrorAttr`. When `db.response.error` is not selected, the span status message is now left empty instead of being set to a hint string. Updated all related tests accordingly.

marctc (changes requested):
Unless I'm missing something, this should be documented in official OBI docs, like this https://github.com/open-telemetry/opentelemetry.io/pull/9753
devdocs IMHO should be only used to documented stuff that's helpful for the developers of this project.

NameHaibinZhang:
Thanks @marctc — you're right. Looking at the `devdocs/README.md` ("documentation that is not useful for our users but might be useful for developers") and the pattern in `devdocs/metrics.md` (which explicitly links to the user-facing page on opentelemetry.io and labels itself a "developer-internal companion"), the `attribute-selection.md` content is clearly user-facing and belongs in the official OBI docs.

I'll remove the devdocs file from this PR and open a follow-up PR against opentelemetry.io to document the `db.response.error` trace-selection behavior there. The code changes (removing the placeholder constant, making the opt-in behavior consistent with `db.query.text`) and the config-schema description updates should still be valid independently.

Let me know if you'd also like the config-schema description change reverted, or if that can stay since it's developer-facing reference.

marctc:
>The code changes (removing the placeholder constant, making the opt-in behavior consistent with `db.query.text`) and the config-schema description updates should still be valid independently.

sounds good, thanks!

NameHaibinZhang:
@marctc doc: https://github.com/open-telemetry/opentelemetry.io/pull/9887

marctc (commented):
I'm happy to unblock the PR but the description and title doesn't match the PR anymore. Would you mind update it?

NameHaibinZhang:
@marctc  Done! I've updated the PR title and description to reflect the current scope (code behavior change only, documentation moved to opentelemetry.io in a separate PR). Thanks for flagging this!

marctc (approved):
LGTM

grcevski (approved):
LGTM!

PR #2087, opened 2026-05-11
feat(retrieval): add vector retrieval observability for RAG pipelines

NameHaibinZhang:
## Summary

Add vector retrieval (similarity search) observability for GenAI RAG pipelines. `RetrievalSpan` detects vector-store calls and enriches the HTTP span with GenAI semantic-convention attributes.

Supported providers (host-anchored): Pinecone, Qdrant, Milvus, Zilliz, Chroma, Weaviate. For self-hosted / custom domains, a generic path + body heuristic is used with `gen_ai.provider.name = generic`.

Highlights:

- New `HTTPSubtypeRetrieval` path in `pkg/ebpf/common/http/retrieval.go`, emitting `gen_ai.operation.name=retrieval`, `gen_ai.provider.name`, `gen_ai.data_source.id`, `gen_ai.response.id`, `gen_ai.usage.input_tokens`.
- Weaviate GraphQL detection accepts both `Get {` and `Get{` formatting.
- Cheap method/host/path pre-check short-circuits non-retrieval HTTP traffic before any body read.
- Debug logging redacts full request/response bodies; only provider/path/byte-sizes are logged.

#1854 

## Validation


pkg/ebpf/common/http/retrieval.go
```diff
@@ -0,0 +1,337 @@
...
+
+	path := strings.TrimSpace(req.URL.Path)
+	if path == "" {
+		return ""
+	}
+	if path == "/" {
+		return "/"
+	}
+	return strings.TrimSuffix(path, "/")
+}
```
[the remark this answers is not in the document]
NameHaibinZhang: Fixed. `normalizedRetrievalPath` now delegates to the shared `requestPath(req)` helper (from responses.go) so that opaque URLs and RequestURI fallbacks are handled consistently with the rerank and Qwen detectors.

pkg/ebpf/common/http/retrieval.go
```diff
@@ -0,0 +1,337 @@
...
+		return 0
+	}
+
+	count := 0
+	for _, signal := range retrievalBodySignals {
+		if bytes.Contains(body, []byte(signal)) {
+			count++
+		}
+	}
+	return count
```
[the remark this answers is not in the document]
NameHaibinZhang: Fixed. Added a package-level `retrievalBodySignalBytes` (precomputed via an IIFE from `retrievalBodySignals`) and updated `retrievalBodySignalCount` to iterate over the `[][]byte` directly, eliminating per-call allocations.

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -835,6 +835,29 @@ func TraceAttributesSelector(span *request.Span, optionalAttrs map[attr.Name]str
 			}
 		}
 
+		if span.SubType == request.HTTPSubtypeRetrieval && span.GenAI != nil && span.GenAI.Retrieval != nil {
+			ai := span.GenAI.Retrieval
+			attrs = append(attrs, semconv.GenAIProviderNameKey.String(ai.Provider))
+			attrs = append(attrs, semconv.GenAIOperationNameKey.String(ai.OperationName()))
+			if ai.Input.Model != "" {
```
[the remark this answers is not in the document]
NameHaibinZhang: Thanks for the observation. The PR description lists the full intended attribute set for the retrieval feature, but this initial PR deliberately focuses on the core GenAI semantic convention attributes (`gen_ai.provider.name`, `gen_ai.operation.name`, `gen_ai.request.model`, `gen_ai.response.model`, `gen_ai.response.id`, `gen_ai.usage.input_tokens`, `gen_ai.data_source.id`). The provider-specific attributes (`db.vector.query.top_k`, `db.vector.query.result_count`, `pinecone.usage.read_units`) are planned for a follow-up PR once the core detection infrastructure is merged. I'll update the PR description to clarify this phasing.

pkg/appolly/app/request/span.go
```diff
@@ -849,6 +852,82 @@ type RerankError struct {
...
+}
+
+// RetrievalResponse captures the common fields from vector search response
+// bodies.
+type RetrievalResponse struct {
+	ID    string         `json:"id,omitempty"`
+	Model string         `json:"model,omitempty"`
+	Usage RetrievalUsage `json:"usage,omitempty"`
+}
+
```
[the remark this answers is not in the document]
NameHaibinZhang: Good point — this ties to the same phasing as the exporter comment above. The current `RetrievalRequest`/`RetrievalResponse` structs capture the fields needed for the core semantic convention attributes shipped in this PR. Extending them with `top_k`, `result_count`, and provider-specific usage fields (e.g., Pinecone read units) is planned for the follow-up PR that adds the corresponding exporter attributes. Keeping the structs minimal for now avoids carrying dead fields that aren't yet wired to output.

pkg/ebpf/common/http/retrieval.go
```diff
@@ -0,0 +1,344 @@
...
+	{"qdrant.io", "/points/query", "qdrant"},
+	// Milvus / Zilliz
+	{"milvus.io", "/vector/search", "milvus"},
+	{"milvus.io", "/entities/search", "milvus"},
+	{"zillizcloud.com", "/vector/search", "zilliz"},
+	{"zillizcloud.com", "/entities/search", "zilliz"},
+	// Chroma
+	{"trychroma.com", "/query", "chroma"},
+	// Weaviate (GraphQL-based similarity search)
+	{"weaviate.io", "/v1/graphql", "weaviate"},
```
mmat11: will this get matched by the graphql parser before getting here?

pkg/ebpf/common/http/retrieval.go
```diff
@@ -0,0 +1,344 @@
...
+	`"collection"`,
+	`"collectionName"`,
+	`"collection_name"`,
+	`"query_embeddings"`,
+}
+
+// retrievalBodySignalBytes is the precomputed []byte form of
+// retrievalBodySignals, avoiding repeated string→[]byte conversions
+// during request classification.
+var retrievalBodySignalBytes = func() [][]byte {
```
mmat11: if those are only used via byte functions, why not hardcode them as []byte instead of string?

pkg/ebpf/common/http/retrieval.go
```diff
@@ -0,0 +1,344 @@
...
+	out := make([][]byte, len(retrievalBodySignals))
+	for i, s := range retrievalBodySignals {
+		out[i] = []byte(s)
+	}
+	return out
+}()
+
+const genericRetrievalProvider = "generic"
+
+var weaviateGraphQLRetrievalSignals = []string{
```
mmat11, line 107: are these referenced somewhere in some documentation? if yes, I would leave a reference here

pkg/ebpf/common/http/retrieval.go
```diff
@@ -0,0 +1,344 @@
...
+
+func parseRetrievalProvider(req *http.Request) string {
+	if req == nil || req.URL == nil {
+		return ""
+	}
+
+	host := extractHostname(req)
+	path := normalizedRetrievalPath(req)
+
+	for _, hp := range retrievalHostPatterns {
```
grcevski: What do you think about making this a map, such that we retrieve the valid mappings for the host and then just suffix on the path? What I mean is something like this:

1. We have a host `something.pinecone.io`, so the extractHostname can just take the last two components, e.g. `pinecone.io`
2. We can get the valid paths like this then `paths := retrievalHostPatterns[host]`
3. Finally we do `for _, p := range paths { if strings.HasSuffix(path, p) return ...}`

pkg/ebpf/common/http/retrieval.go
```diff
@@ -0,0 +1,344 @@
...
+}
+
+func retrievalBodySignalCount(body []byte) int {
+	if len(body) == 0 {
+		return 0
+	}
+
+	count := 0
+	for _, signal := range retrievalBodySignalBytes {
+		if bytes.Contains(body, signal) {
```
grcevski, line 177: This can be a pretty expensive check I imagine for a large body... I don't know if there's an easier way to perform this check. If it's not possible, then we should at least pass in an argument saying how many minimum we are looking for and break when we find the minimum required.

grcevski: Aho Corrasick algorithm can be an effective single linear scan method to check for the occurences. I searched now to see if there's an implementation in go and there seems to be one by claudflare (missing package though) and this one "github.com/pgavlin/aho-corasick".

grcevski (commented):
Great addition as usual!!! I left a few comments on efficiency of some of the checks. I'm particularly concerned about the overhead of the body scan, so it would be great to improve that. If you choose to use the Aho Corrasick algorithm, do you mind adding a benchmark with that approach and the original loop?

pkg/ebpf/common/http/retrieval.go
NameHaibinZhang: Good question. No — `GraphQLSpan` runs on server-side events (`!isClientEvent` in http_transform.go L151), while `RetrievalSpan` runs exclusively on client-side events (`isClientEvent` at L186). They never evaluate the same HTTP event, so there is no risk of the generic GraphQL parser pre-empting retrieval detection for Weaviate outbound calls.

pkg/ebpf/common/http/retrieval.go
NameHaibinZhang: Good point. Simplified — removed the intermediate `[]string` slice and the IIFE, now defining `retrievalBodySignalBytes` directly as a `[][]byte` literal.

pkg/ebpf/common/http/retrieval.go
NameHaibinZhang: Done. Added a doc comment with a reference to the Weaviate similarity search documentation (https://weaviate.io/developers/weaviate/search/similarity).

pkg/ebpf/common/http/retrieval.go
NameHaibinZhang: Done. Refactored to a `map[string][]retrievalHostEntry` keyed by registrable domain (last two labels). A new `registrableDomain()` helper extracts the key, giving O(1) host lookup followed by a short path-suffix scan over only the matching entries.

pkg/ebpf/common/http/retrieval.go
NameHaibinZhang: Good catch. Added a `minRequired` parameter to `retrievalBodySignalCount` — the function now returns as soon as it finds the threshold (currently 2). With only 9 short patterns and early-break at 2, worst-case is 2 linear scans of the body before returning. For the Aho-Corasick suggestion: given the small pattern set and the early-break optimization, I believe the overhead is acceptable for now. If profiling later shows this is still a hotspot, I am happy to introduce Aho-Corasick with a benchmark in a follow-up.

NameHaibinZhang:
Thanks for the positive feedback and the efficiency suggestions! I've addressed the concerns in the latest push:

1. **Host lookup**: Refactored to a `map[string][]retrievalHostEntry` keyed by registrable domain for O(1) lookup instead of linear scan.
2. **Body scan**: Added early-break optimization — `retrievalBodySignalCount` now accepts a `minRequired` threshold and returns immediately once hit, avoiding unnecessary scans on large payloads. Also switched from runtime string→byte conversion to pre-defined `[][]byte` literals to eliminate per-call allocations.

Regarding Aho-Corasick: with only 9 short patterns and early-break at threshold 2, worst-case is ~2 linear passes over the body before returning. I'd like to keep this approach for now since it's simple and efficient for the current signal set. If the pattern list grows or profiling shows this is still a hotspot, I'll implement Aho-Corasick with a benchmark comparison as a follow-up. Happy to discuss further!

marctc (approved):
LGTM! what do you think add a devdoc with more info for developers on how to work with the GenAI codebase? my question is what we would need to do if, for example, we have to deal with v2 of weviate?

NameHaibinZhang:
@marctc  Thanks for the approval and great suggestion! I agree that a devdoc on how to extend the GenAI codebase would be very helpful for contributors. It could cover things like:

- How to add support for a new vector database (e.g. Weaviate v2)
- The detection pattern structure (host matching, body signal scanning)
- How to add new retrieval signals and integrate with the existing framework

I'll create a follow-up PR with a `devdocs/genai-extension-guide.md` to document this. Thanks again for the feedback!

marctc:
> I'll create a follow-up PR with a `devdocs/genai-extension-guide.md` to document this. Thanks again for the feedback!

cool, thanks! more than extension, can be also architecture and so. But up to you!

grcevski (approved):
LGTM! Thanks for making the fixes. I agree, with small number of values to look for and  hit count, it's likely that it's a wash with Aho Corrsick.

mmat11 (approved):
🚀

PR #2088, opened 2026-05-11
fix(metrics): remove DNS question name from default metric labels

### LINKED ISSUE #2028: Raw DNS question names are exported as default metric labels
## Summary

OBI copies the first DNS question name into `span.Path`, exposes it as `DNSQuestionName`, and enables that attribute by default for DNS lookup duration metrics. This causes raw queried domains to be emitted into metrics and lets attacker-influenced hostnames drive label cardinality.

## Impact

Telemetry consumers can see raw DNS names, and many unique names can grow Prometheus or OTLP metric state indefinitely. The issue is limited to observability outputs and operational memory pressure, so `low` severity is appropriate.

## Environment

- First identified in: 764826e266eeb2bacabe0774acc7e9d2044a8cad

## Evidence

- `pkg/ebpf/common/dns_request_transform.go` stores the first question name in `span.Path`.
- `pkg/appolly/app/request/span_getters.go` maps `DNSQuestionName` to `span.Path`.
- `pkg/export/attributes/attr_defs.go` enables `DNSQuestionName` for DNS lookup duration by default.
- `pkg/export/prom/prom.go` emits DNS metrics with attribute-derived label values.

## Steps to reproduce

1. From the repository root on `main`, create `/tmp/dns_label_poc.go` with this exact program:
   ```go
   package main
   import (
       "fmt"
       "golang.org/x/net/dns/dnsmessage"
       "go.opentelemetry.io/obi/pkg/appolly/app/request"
   )
   func main() {
       questionName := dnsmessage.MustNewName("unique-123.attacker.example.")
       msg := dnsmessage.Message{
           Questions: []dnsmessage.Question{{
               Name:  questionName,
               Type:  dnsmessage.TypeA,
               Class: dnsmessage.ClassINET,
           }},
       }
       packed, err := msg.Pack()
       if err != nil { panic(err) }
       var parsed dnsmessage.Message
       if err := parsed.Unpack(packed); err != nil { panic(err) }
       q := parsed.Questions[0]
       span := request.Span{Path: q.Name.String()}
       kv := request.DNSQuestionName(span.Path)
       fmt.Printf("question=%q span.Path=%q metric_label=%q\n", q.Name.String(), span.Path, kv.Value.AsString())
   }
   ```
2. Run it with `GOTOOLCHAIN=auto go run /tmp/dns_label_poc.go`.
3. Let the PoC print the original DNS question, the derived `span.Path`, and the metric label value.
4. Observe that the output shows the attacker-controlled DNS name copied unchanged into the metric label, proving both raw-name disclosure and per-name series growth.

## Suggested Fix Direction

Do not enable raw DNS question names by default for metrics, or normalize/hash/truncate them behind an explicit opt-in.

## Acceptance Criteria

- Raw DNS question names are not emitted as default metric labels.
- DNS metrics include bounded or redacted naming behavior when detailed labels are enabled.
- Tests cover repeated unique DNS names and default exporter output.

## Note

NameHaibinZhang:
## Summary

- Remove `dns.question.name` from default attributes for DNSLookupDuration metrics to prevent unbounded cardinality caused by highly variable or malicious DNS domains
- Make DNSQuestionName span attribute conditional in `TraceAttributesSelector` - only emitted when explicitly selected via `optionalAttrs`
- Update integration test to not expect `dns_question_name` label by default

## Problem

Raw DNS question names are exported as default metric labels for the DNS lookup duration metric (`dns_lookup_duration_seconds`). Since DNS queries can contain highly variable domains (including traffic patterns or malicious domains), this triggers **unbounded metric cardinality**, potentially causing monitoring backends to experience operational memory pressure as unique time series accumulate indefinitely.

## Solution

- **Default behavior**: DNS metrics no longer include the raw domain name label. The `error_type` label is retained for error classification (e.g., NXDomain).
- **Optional enablement**: Users can still explicitly enable `DNSQuestionName` via attribute configuration if they need domain-level granularity, but should be aware of cardinality implications.

## Changed Files

| File | Change |
|------|--------|
| `pkg/export/attributes/attr_defs.go` | Remove `attr.DNSQuestionName` from default attributes for `DNSLookupDuration.Section` |
| `pkg/export/otel/tracesgen/tracesgen.go` | Make `DNSQuestionName` span attribute conditional on `optionalAttrs` |
| `internal/test/integration/red_test_python.go` | Update DNS metric queries to not expect `dns_question_name` label |

Fixes #2028


grcevski:
This is actually a good idea, maybe we shouldn't close it :)

NameHaibinZhang:
@grcevski  Thanks for the support! I've reopened the PR. The unbounded cardinality from raw DNS question names in metrics can be a real operational concern, so making it opt-in while keeping `error_type` as the default label seems like a reasonable balance between observability and stability.

Happy to address any further feedback on the implementation.

pkg/export/attributes/attr_defs.go
```diff
@@ -381,6 +381,7 @@ func getDefinitions(
 		},
 		Traces.Section: {
 			Attributes: map[attr.Name]Default{
+				attr.DNSQuestionName:   false,
```
grcevski: I think we can make it true here, this is for traces and it would be useful for it to be there.

pkg/export/attributes/attr_defs.go
```diff
@@ -419,10 +420,17 @@ func getDefinitions(
 		DNSLookupDuration.Section: {
 			SubGroups: []*AttrReportGroup{&appAttributes},
 			Attributes: map[attr.Name]Default{
-				attr.DNSQuestionName: true,
-				attr.ErrorType:       true,
+				attr.ErrorType: true,
+				// Note: DNSQuestionName is intentionally NOT included as a default
```
grcevski: I think we should add it as `attr.DNSQuestionName: false`. This was it would be possible to add it via config.

internal/test/integration/red_test_python.go
```diff
@@ -82,21 +82,23 @@ func testREDMetricsDNSForPython(t *testing.T, url, comm, namespace string) {
...
 		results, err = pq.Query(`dns_lookup_duration_seconds_count{` +
-			`dns_question_name="opentelemetry.io.",` +
 			`service_namespace="` + namespace + `",` +
 			`service_name="` + comm + `"}`)
 		require.NoError(ct, err)
 		enoughPromResults(ct, results)
 		val := totalPromCount(ct, results)
 		assert.LessOrEqual(ct, 1, val)
 
+		// Verify error-labeled DNS metrics still work (NXDomain responses)
```
grcevski: With my suggestion of making the attr.DNSQuestionName: false in the attributes by default, but exiting there we can renable these checks again. I think our config uses `*` to include all attributes. All we need to is disable it by default from metrics. Trace attributes are fine, I don't think the DNS name is sensitive information, which is the reason we might want to hide it from traces.

grcevski (commented):
Thanks for fixing this, small comments and we should merge this.

internal/test/integration/red_test_python.go
NameHaibinZhang: Right, since the test config uses `include: ["*"]`, all attributes including `DNSQuestionName` will be enabled. Restored the `dns_question_name` checks in the Prometheus queries. Also updated the unit test in `tracesgen_test.go` to verify that `DNSQuestionName` defaults to `true` for traces.

pkg/export/attributes/attr_defs.go
NameHaibinZhang: Good point! Changed it to `true` — DNS question name is indeed useful in traces and doesn't have the cardinality concern that metrics do.

pkg/export/attributes/attr_defs.go
NameHaibinZhang: Makes sense. Added `attr.DNSQuestionName: false` so it stays configurable via the attribute selection. Users who want it can enable it through `include` config.

grcevski (approved):
LGTM!

PR #2095, opened 2026-05-11
refactor(tracing): use explicit sentinel for traceparent position in bpf_strstr_tp_loop

smehboub:
## Summary

Refine traceparent detection to support matches at offset 0.

## Description

This PR modifies the bpf_strstr_tp_loop logic to distinguish between a "header not found" state and a match located at the very beginning of the buffer.

Currently, the helper uses 0 as the initial value and checks for success via if (data.pos). This creates an ambiguity where a match occurring at offset 0 is indistinguishable from a "not found" result.

While this offset might not occur in the current HTTP request line parsing context, using a sentinel value (0xffffffff) makes the helper more generic for future cases where a match could occur at the very start of a buffer.

## Validation

### DIFF
--- a/bpf/common/trace_util.h
+++ b/bpf/common/trace_util.h
@@ -18,7 +18,7 @@ struct callback_ctx {
     u8 _pad[4];
 };
 
-enum { k_tp_pos_not_found = 0xffffffffu };
+enum : u32 { k_tp_pos_not_found = 0xFFFFFFFFU };
 
 static unsigned char *hex = (unsigned char *)"0123456789abcdef";
 static unsigned char *reverse_hex =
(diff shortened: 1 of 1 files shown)


bpf/common/trace_util.h
```diff
@@ -18,6 +18,8 @@ struct callback_ctx {
     u8 _pad[4];
 };
 
+enum { k_tp_pos_not_found = 0xffffffffu };
```
rafaelroquetto:
```suggestion
enum : u32 { k_tp_pos_not_found = 0xffffffffu };
```

mmat11 (approved):
lgtm, linter is failing

bpf/common/trace_util.h
smehboub: Thanks for the suggestion! Done.

rafaelroquetto (approved):

PR #2113, opened 2026-05-14
Fix #2038: use basename for instrumentation_errors_total process_name label

### LINKED ISSUE #2038: instrumentation_errors_total uses full exec paths as labels
## Summary

The low-level instrumentation error metric still uses `ie.FileInfo.CmdExePath` as `process_name`. Each unique executable path therefore creates a new metric series when required attach failures occur.

## Impact

This can grow internal metric state without bound and exposes full filesystem paths in observability outputs. The issue is local and operational, so `low` severity is appropriate.

## Environment

- First identified in: 2410419e3475b26b3a42d4f907cded30025d2cb2

## Evidence

- `pkg/ebpf/tracer_linux.go` still sets `processName: ie.FileInfo.CmdExePath`.
- `pkg/ebpf/instrumenter.go` still reports required attach failures with `i.processName`.
- `pkg/export/imetrics/iprom.go` still keys `instrumentation_errors_total` by `process_name` and `error_type`.

## Steps to reproduce

1. From the repository root on `main`, create `/tmp/instrumentation_error_metric_poc.go` with this exact program:
   ```go
   package main
   import (
       "fmt"
       "github.com/prometheus/client_golang/prometheus"
       "go.opentelemetry.io/obi/pkg/components/imetrics"
   )
   func main() {
       reg := prometheus.NewRegistry()
       reporter := imetrics.NewPrometheusReporter(&imetrics.PrometheusConfig{}, nil, reg)
       for i := 0; i < 5; i++ {
           processName := fmt.Sprintf("/tmp/attacker/bin-%d", i)
           reporter.InstrumentationError(processName, "attach_uprobe")
       }
       mfs, err := reg.Gather()
       if err != nil { panic(err) }
       for _, mf := range mfs {
           if mf.GetName() == "obi_instrumentation_errors_total" {
               fmt.Printf("metric=%s series=%d\n", mf.GetName(), len(mf.Metric))
               for _, m := range mf.Metric {
                   for _, l := range m.Label {
                       if l.GetName() == "process_name" {
                           fmt.Printf("label process_name=%s\n", l.GetValue())
                       }
                   }
               }
           }
       }
   }
   ```
2. Run it with `GOTOOLCHAIN=auto go run /tmp/instrumentation_error_metric_poc.go`.
3. Let the PoC simulate instrumentation failures with distinct executable paths.
4. Observe that it creates one `obi_instrumentation_errors_total` series per unique `process_name`, demonstrating both raw-path exposure and unbounded label growth.

## Suggested Fix Direction

Use a bounded, normalized process identifier instead of full executable paths, or aggregate error metrics across a safer key.

## Acceptance Criteria

- Instrumentation error metrics no longer use full executable paths as unbounded labels.
- Repeated failures across many unique paths do not create unbounded internal series growth.
- Tests cover instrumentation-error labeling behavior.

## Note

NameHaibinZhang:
## Summary

Fixes #2038

The `instrumentation_errors_total` metric was using full executable paths as the `process_name` label, which could lead to unbounded cardinality growth and filesystem path exposure.

## Changes

- Use `FileInfo.ExecutableName()` to extract only the executable basename for the `process_name` label in the instrumenter struct
- This is consistent with how `attacher.go` already reports process names for `InstrumentProcess` and `InstrumentationError` calls
- Bounds the label cardinality and avoids exposing full filesystem paths

## Testing

- CI: Generate and checks pass, Unit tests pass
- The change uses the existing `ExecutableName()` method already used throughout `attacher.go` for the same purpose

### DIFF
--- a/pkg/ebpf/tracer_linux.go
+++ b/pkg/ebpf/tracer_linux.go
@@ -329,7 +329,7 @@ func (pt *ProcessTracer) NewExecutable(exe *link.Executable, ie *Instrumentable)
 		offsets:     ie.Offsets, // this is needed for the function offsets, not fields
 		modules:     map[uint64]struct{}{},
 		metrics:     pt.metrics,
-		processName: ie.FileInfo.CmdExePath,
+		processName: ie.FileInfo.ExecutableName(),
 	}
 
 	for _, p := range pt.Programs {


mariomac (approved):
Good point! For consistency with the OTEL semantics, we should use the `.name` to spot the executable name instead of the path: https://opentelemetry.io/docs/specs/semconv/registry/attributes/process/#process-executable-name

Thanks for fixing!

PR #2114, opened 2026-05-14
Fix #2036: prevent nil map panic in SpanNameLimiter for zero-value service key

### LINKED ISSUE #2036: SpanNameLimiter panics on zero-valued service key
## Summary

`SpanNameLimiter.aggregate` initializes `lastCount` with a nil `routes` map and only allocates a map when the service key changes. If the first span in a batch has an empty name/namespace and therefore the zero `ServiceNameNamespace` key, the code writes into the nil map and panics.

## Impact

This can crash the metrics path or the OBI process when span metrics are enabled. The bug is edge-case and configuration-gated, so `low` severity is appropriate.

## Environment

- First identified in: 3520a9e1c48aac5bd33f24377f2a2563598b0d9e

## Evidence

- `pkg/transform/span_name_limiter.go` starts with `var lastKey svc.ServiceNameNamespace` and `lastCount := &routesCount{}`.
- The map is only allocated when `span.Service.UID.NameNamespace() != lastKey`.
- The code still writes `lastCount.routes[span.TraceName()] = struct{}{}` unconditionally afterward.

## Steps to reproduce

1. From the repository root on `main`, add `pkg/transform/span_name_limiter_poc_test.go` with this exact test:
   ```go
   package transform
   import (
       "testing"
       "time"
       "go.opentelemetry.io/obi/pkg/app/request"
       "go.opentelemetry.io/obi/pkg/components/helpers/cache"
   )
   func TestSpanNameLimiterNilMap(t *testing.T) {
       l := &spanNameLimiter{
           limit:          1,
           spanNamesCount: cache.NewExpirableLRU[svc.ServiceNameNamespace, *routesCount](time.Minute),
       }
       spans := []request.Span{{}}
       l.aggregate(spans)
   }
   ```
2. Run `GOTOOLCHAIN=local go test ./pkg/transform -run TestSpanNameLimiterNilMap -count=1`.
3. Observe that the test panics with `assignment to entry in nil map`, proving the first empty service key crashes the limiter.

## Suggested Fix Direction

Initialize `lastCount.routes` defensively before the first write, or special-case the zero-value key the same way as any other service key.

## Acceptance Criteria

- `SpanNameLimiter` cannot write to a nil `routes` map for any input batch.
- Empty or missing service identity does not panic the limiter.
- Tests cover zero-value service keys and first-span edge cases.

## Note

NameHaibinZhang:
## Summary

Fixes #2036

`SpanNameLimiter.aggregate` panics when the first span in a batch has a zero-value `ServiceNameNamespace` key because the `routes` map is nil.

## Root Cause

- `lastCount` is initialized with `&routesCount{}` which has a nil `routes` map
- The map is only allocated when the service key changes
- If the first span's service key matches the zero-value `lastKey`, the map is never allocated
- Writing to the nil map causes a panic

## Fix

Initialize the `routes` map in `routesCount` at creation time to ensure it is never nil when written to.

## Testing

- Added regression test `TestSpanNameLimiter_ZeroValueServiceKey` covering the zero-value service key edge case
- Verified existing tests still pass

### DIFF
--- a/pkg/transform/span_name_limiter.go
+++ b/pkg/transform/span_name_limiter.go
@@ -95,7 +95,7 @@ func (l *spanNameLimiter) aggregate(spans []request.Span) []request.Span {
 	// assuming many spans from the same service could come in a row
 	// we can slightly optimize by avoiding the cache lookup for each span
 	var lastKey svc.ServiceNameNamespace
-	lastCount := &routesCount{}
+	lastCount := &routesCount{routes: map[string]struct{}{}}
 
 	output := spans
 	alreadyCopying := false
--- a/pkg/transform/span_name_limiter.go
+++ b/pkg/transform/span_name_limiter.go
@@ -95,13 +95,13 @@ func (l *spanNameLimiter) aggregate(spans []request.Span) []request.Span {
 	// assuming many spans from the same service could come in a row
 	// we can slightly optimize by avoiding the cache lookup for each span
 	var lastKey svc.ServiceNameNamespace
-	lastCount := &routesCount{routes: map[string]struct{}{}}
+	var lastCount *routesCount
 
 	output := spans
 	alreadyCopying := false
 	for i := 0; i < len(output); i++ {
 		span := &output[i]
-		if key := span.Service.UID.NameNamespace(); key != lastKey {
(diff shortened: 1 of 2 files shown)


mariomac (approved):
Thanks! 👏🏻

PR #2118, opened 2026-05-14
feat(tool): add LLM tool call detection for OpenAI, Anthropic, Gemini, and Qwen

NameHaibinZhang:
## Summary

Adds tool call detection for four major LLM providers, addressing part of the AI Agent Observability Support (issue #1854).

### Data Model
- Added `ToolCall` struct with `ID` and `Name` fields
- Extended `VendorOpenAI`, `VendorAnthropic`, `VendorGemini` with `ToolCalls` field

### Detection Logic
- **OpenAI**: Extracts `tool_calls` from `choices[].message.tool_calls`
- **Anthropic**: Extracts `tool_use` blocks from response content (both streaming and non-streaming)
- **Gemini**: Extracts `functionCall` from `candidates[].content.parts[]`
- **Qwen**: Reuses OpenAI-compatible `extractToolCalls` for `choices[].message.tool_calls`

### Trace Attributes
- Generates `gen_ai.tool.name` (comma-separated for multiple calls)
- Generates `gen_ai.tool.call.id` (when single call with ID)

## Validation

### DIFF
--- a/pkg/ebpf/common/http/anthropic.go
+++ b/pkg/ebpf/common/http/anthropic.go
@@ -16,6 +16,35 @@ import (
 	"go.opentelemetry.io/obi/pkg/appolly/app/request"
 )
 
+type anthropicContentBlock struct {
+	Type string `json:"type"`
+	ID   string `json:"id"`
+	Name string `json:"name"`
+}
+
+func extractAnthropicToolCalls(content json.RawMessage) []request.ToolCall {
+	if len(content) == 0 {
+		return nil
+	}
+
+	var blocks []anthropicContentBlock
+	if err := json.Unmarshal(content, &blocks); err != nil {
+		return nil
+	}
+
+	var result []request.ToolCall
+	for i := range blocks {
+		if blocks[i].Type != "tool_use" || blocks[i].Name == "" {
+			continue
+		}
+		result = append(result, request.ToolCall{
+			ID:   blocks[i].ID,
+			Name: blocks[i].Name,
+		})
+	}
+	return result
+}
+
 func isAnthropic(hdr http.Header) bool {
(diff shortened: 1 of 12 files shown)


pkg/ebpf/common/http/anthropic.go
```diff
@@ -73,22 +102,30 @@ func AnthropicSpan(baseSpan *request.Span, req *http.Request, resp *http.Respons
 	}
 
 	var parsedResponse request.AnthropicResponse
+	var streamToolCalls []request.ToolCall
 	if len(respB) > 0 && respB[0] == '{' {
 		if err := json.Unmarshal(respB, &parsedResponse); err != nil {
 			slog.Debug("failed to parse Anthropic response", "error", err)
 		}
```
mmat11, line 109: question: why don't we define `toolCalls []request.ToolCall` outside the if branch and then parse the response toolcalls in the response branch and the stream toolcalls in the stream branch? this way we can avoid the block on lines 118-122

disregard my comment if response toolCalls always have precedence and should be preferred if available

mmat11 (approved):
left one nit, the rest lgtm!

pkg/ebpf/common/http/anthropic.go
NameHaibinZhang: Good point! Refactored to define `toolCalls` outside the branches and assign directly within each — `extractAnthropicToolCalls(parsedResponse.Content)` in the JSON branch and the stream's tool calls in the stream branch. This eliminates the `streamToolCalls` variable and the fallback block. Thanks for the suggestion!

rafaelroquetto (approved):

grcevski (approved):
LGTM!

PR #2120, opened 2026-05-14
Add regression tests for OBI exclusion scope

### LINKED ISSUE #1254: Decrease matching scope of  example exclude regex
The example exclude regex `.*obi.*` is likely too broad (it will match any executable path containing the substring "obi", potentially excluding unrelated processes). Consider anchoring the match to the binary name/path segment (e.g., `(?:^|/)obi$` or a similar bounded pattern) while still excluding otelcol.
```suggestion
#        - exe_path: "(?:^|/)(?:obi|otelcol)$"
```

_Originally posted by @Copilot in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1252#discussion_r2776153307_

stellhub:
Fixes #1254.

## What changed
- Added regression coverage for the default `exclude_instrument` glob so paths that only contain `obi` as a substring are not excluded.
- Added the same coverage for the legacy `default_exclude_services` regex path selector, including `/usr/bin/obi/test` and `obi` substring cases.

## Why
The default OBI exclusion should target the `obi` binary/path segment without excluding unrelated binaries whose paths merely contain the substring `obi`.

## Validation
- `gofmt -w pkg/obi/config_test.go`
- `git diff --check`

Note: `go test ./pkg/obi` does not build in my local Windows environment because this package pulls Linux/eBPF-only symbols and generated BPF bindings that are unavailable there.

### DIFF
--- a/pkg/obi/config_test.go
+++ b/pkg/obi/config_test.go
@@ -747,6 +747,9 @@ func TestDefaultExclusionFilter(t *testing.T) {
 	assert.True(t, c[0].Path.MatchString("otelcol-contrib"))
 
 	assert.False(t, c[0].Path.MatchString("/usr/bin/obi/test"))
+	assert.False(t, c[0].Path.MatchString("myobi"))
+	assert.False(t, c[0].Path.MatchString("/usr/bin/myobi"))
+	assert.False(t, c[0].Path.MatchString("/usr/bin/obi-helper"))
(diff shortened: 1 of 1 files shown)


stellhub:
EasyCLA is now passing. The GitHub Actions runs are currently in action_required state, and I don't have repository admin rights to rerun or approve them from my fork. Could a maintainer please approve the workflow runs when convenient?

rafaelroquetto (approved):
LGTM

PR #2125, opened 2026-05-14
Refactor FileInfo onwership semantics

### LINKED ISSUE #2026: Delayed Java matcher update races with span routing

rafaelroquetto:
## Summary

https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2026 got me wondering about the ownership and lifecycle of `svc.Attr` - it's suffering from race conditions, and whilst the instinctive reaction is to try to synchronise access to it at every usage site, taking a step back I realised this is an ownership semantics issue. Who owns that object? Where is it materialised? So I went looking because from the code it's hard to tell. 

It turns out it was not owned by anyone - different entities held references and performed access to this object at different times - like a true unchecked shared pointer.

So I went on to map its lifecycle - in a nutshell, it is first born in `FilterClassify`, and then gets forwarded to `findExecElf` as sink, wherein it gets stored inside an `exec.FileInfo` object. From that point onwards, the `exec.FileInfo` object gets passed down and then shared multiple times.

So the problem is: there's no ownership control of `exec.FileInfo`, and consequentially, of its `svc.Attr` data member - which gets copied and shared itself.

In short: anyone who holds a reference to `exec.FileInfo` or `svc.Attr` has the potential to mutate any of those, at any given time, and `svc.Attr` at all times.

This PR seeks to reinstate sane ownership semantics to `FileInfo`: it's now simple to reason about: _it's now treated as an explicitly shared, thread-safe object_
 - it's materialised once, and shared across the board
 - as such, we guarantee thread-safety by encapsulating access to data members via thread-safe accessors/methods.

_A simple as that._ There's no more need to reason about who owns the object, or when to synchronise it, by design. Any one that needs to store `svc.Attr` now gets their own copy, for instance. Everything is now gated, the semantics are clear and linear.

Fixes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2026

## Validation


rafaelroquetto:
This PR accidentally includes logenricher tests (which also need changes) I will rebase once https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2126 gets merged.

skl (approved):
💯

pkg/appolly/discover/exec/file.go
```diff
@@ -25,7 +36,207 @@ type FileInfo struct {
...
 	return parts[len(parts)-1]
 }
+
+func (fi *FileInfo) ServiceAttrs() svc.Attrs {
+	fi.mu.RLock()
+	defer fi.mu.RUnlock()
+
+	out := fi.service
+	out.Metadata = maps.Clone(fi.service.Metadata)
+	out.EnvVars = maps.Clone(fi.service.EnvVars)
```
mariomac, line 89: Can this slowdown some loops that continuously ask for service attributes of each data point they process?

mariomac: I just noticed that the `maps.Clone` function is now efficient unlike `maps.Copy`. Cool!

mariomac (approved):

PR #2136, opened 2026-05-15
Introduce health check endpoint

rafaelroquetto:
## Summary

Introduce a new health check endpoint (`http://<obi-host>:54545/healthz`) that can be used by watchdogs to quickly verify if OBI is running and its internal pipeline is "flowing"/not stuck.

This is meant to be a simple and straightforward API, so there's nothing fancy here, just heart beats. The endpoint schema is as follows:

```
{
  "schema_version": 1,
  "now_unix_ns": 1747300000000000000,
  "process_uptime_ns": 12345678,
}
```

hopefully it's self-explanatory.

## Validation

### DIFF
--- a/devdocs/config/config-schema.json
+++ b/devdocs/config/config-schema.json
@@ -1258,14 +1258,9 @@
     },
     "HealthCheckConfig": {
       "properties": {
-        "enabled": {
-          "type": "boolean",
-          "default": false,
-          "x-env-var": "OTEL_EBPF_HEALTH_CHECK_ENABLED"
-        },
         "port": {
           "type": "integer",
-          "default": 54545,
+          "default": 0,
           "x-env-var": "OTEL_EBPF_HEALTH_CHECK_PORT"
         }
       },
--- a/devdocs/config/config-schema.json
+++ b/devdocs/config/config-schema.json
@@ -1256,6 +1256,21 @@
         "response"
       ]
     },
+    "HealthCheckConfig": {
+      "properties": {
+        "enabled": {
+          "type": "boolean",
+          "default": false,
+          "x-env-var": "OTEL_EBPF_HEALTH_CHECK_ENABLED"
+        },
+        "port": {
+          "type": "integer",
+          "default": 54545,
+          "x-env-var": "OTEL_EBPF_HEALTH_CHECK_PORT"
+        }
+      },
+      "type": "object"
+    },
     "HostIDConfig": {
       "properties": {
         "override": {
@@ -2808,6 +2823,9 @@
     },
     "filter": {
       "$ref": "#/$defs/AttributesConfig"
+    },
+    "health_check": {
+      "$ref": "#/$defs/HealthCheckConfig"
     },
     "internal_metrics": {
       "$ref": "#/$defs/InternalMetricsConfig"
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -1,12 +1,8 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-// Package health exposes a watchdog-style health endpoint for OBI.
-//
-// Each long-running subsystem registers a name and receives a Tick function
-// that it must call from inside its own work-loop select. A parent supervisor
-// polls /healthz, compares now_unix_ns to each subsystem's last_tick_unix_ns,
-// and decides on its own threshold whether OBI is stuck.
+// Package health exposes a /healthz endpoint reporting that the OBI process
+// is reachable and its scheduler is alive.
 package health // import "go.opentelemetry.io/obi/pkg/health"
 
 import (
@@ -16,17 +12,11 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
-	"os"
-	"sync"
-	"sync/atomic"
-	"syscall"
 	"time"
 )
 
 const (
-	// Path is the URL path served by the health endpoint.
-	Path = "/healthz"
-
+	path          = "/healthz"
 	schemaVersion = 1
 )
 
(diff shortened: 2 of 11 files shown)


NimrodAvni78 (approved):

devdocs/config/config-schema.json
```diff
@@ -1256,6 +1256,21 @@
         "response"
       ]
     },
+    "HealthCheckConfig": {
+      "properties": {
+        "enabled": {
+          "type": "boolean",
+          "default": false,
+          "x-env-var": "OTEL_EBPF_HEALTH_CHECK_ENABLED"
+        },
```
mariomac: I think we can remove this option and enable health check iif `OTEL_EBPF_HEALTH_CHECK_PORT` is not zero.

This simplifies the configuration and gives consistency with other configuration options that can be enabled just by defining a port (e.g. prometheus export)

pkg/internal/appolly/appolly.go
```diff
@@ -69,6 +70,10 @@ func New(ctx context.Context, ctxInfo *global.ContextInfo, config *obi.Config) (
 
 	swi := &swarm.Instancer{}
 
+	swi.Add(swarm.DirectInstance(func(ctx context.Context) {
+		health.TickEvery(ctx, ctxInfo.AppO11yHeartbeat, 5*time.Second)
```
mariomac: Since `TickEvery` runs in its own goroutine independently from everything else, I think it would always report "health" even if the whole pipeline has crashed. I would remove tickers by now (to avoid misleading info) and we can consider in the future wether a [more standard, pull-based health check](https://metrics.dropwizard.io/4.2.0/manual/healthchecks.html) is worth implementing

mariomac (commented):
I'd really love to get health check implemented in OBI!! Thanks for starting it. I have a couple of non-blocking comments.

pkg/health/health.go
```diff
@@ -0,0 +1,196 @@
...
+	"os"
+	"sync"
+	"sync/atomic"
+	"syscall"
+	"time"
+)
+
+const (
+	// Path is the URL path served by the health endpoint.
+	Path = "/healthz"
```
mmat11: does this need to be exported?

pkg/health/health.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+		return nil
+
+	case err := <-srvErr:
+		if err == nil {
+			return nil
+		}
+		l.Error("health endpoint exited unexpectedly", "err", err)
+		// match prommgr behavior: interrupt for graceful shutdown
+		if kerr := syscall.Kill(os.Getpid(), syscall.SIGINT); kerr != nil {
```
mmat11: will this kill OBI if the healthcheck port is busy?

mmat11: I was thinking, would it be enough to log an error?

devdocs/config/config-schema.json
rafaelroquetto: done

pkg/internal/appolly/appolly.go
rafaelroquetto: Ahhh good point - so I have just removed the tickets in the manner you suggested. Eventually I will introduce a more "pull based" check - something similar to Alloy's "component health" I reckon

pkg/health/health.go
rafaelroquetto: Nope, unexported.

pkg/health/health.go
rafaelroquetto: yes, fixed

pkg/health/health.go
```diff
@@ -0,0 +1,85 @@
...
+	now := time.Now()
+	resp := struct {
+		SchemaVersion   int   `json:"schema_version"`
+		NowUnixNs       int64 `json:"now_unix_ns"`
+		ProcessUptimeNs int64 `json:"process_uptime_ns"`
+	}{
+		SchemaVersion:   schemaVersion,
+		NowUnixNs:       now.UnixNano(),
+		ProcessUptimeNs: now.Sub(e.start).Nanoseconds(),
+	}
```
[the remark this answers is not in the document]
rafaelroquetto: Already done

pkg/health/health.go
```diff
@@ -0,0 +1,85 @@
...
+			l.Warn("error closing health endpoint", "err", err)
+		}
+		return nil
+
+	case err := <-srvErr:
+		if err != nil {
+			l.Error("health endpoint exited unexpectedly", "err", err)
+		}
+		return nil
+	}
```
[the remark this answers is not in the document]
rafaelroquetto: @mmat11 wdyt here? I changed it to silently continue not to bring it down, but maybe Copilot has a point here. The health check and always be disabled anyway, but perhaps if it is enabled and not coming up properly, we should indeed fail, or?

pkg/obi/config.go
```diff
@@ -421,6 +424,12 @@ type Config struct {
 
 	NodeJS NodeJSConfig `yaml:"nodejs"`
 	Java   JavaConfig   `yaml:"javaagent"`
+
+	HealthCheck HealthCheckConfig `yaml:"health_check"`
+}
+
+type HealthCheckConfig struct {
+	Port int `yaml:"port" env:"OTEL_EBPF_HEALTH_CHECK_PORT"`
 }
```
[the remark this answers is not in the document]
rafaelroquetto: Intended. This is the OBI way.

pkg/health/health.go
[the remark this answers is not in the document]
mmat11: mm not sure I agree with copilot, we are not swallowing the error, we are logging it and avoid killing OBI if _something_ goes wrong on the healthcheck server (which is not critical for OBI to continue working). Propagating the failure would lead to the same scenario of killing it directly, wdyt?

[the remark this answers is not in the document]
mmat11: I think we should decide whether the health endpoint, when enabled, is a critical part of OBI and handle the error based on that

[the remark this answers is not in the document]
rafaelroquetto: I really don't have a strong opinion on this. Perhaps a slight inclination towards failing if it is enabled, but honestly I can't tell.

Let's discuss this offline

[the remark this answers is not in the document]
rafaelroquetto: As discussed offline, will leave this as is - providing mechanism, not policy. Anyone watching the endpoint can decide what to do best.

pkg/health/health_test.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func freePort(t *testing.T) int {
+	t.Helper()
+	l, err := net.Listen("tcp", "127.0.0.1:0")
+	require.NoError(t, err)
+	port := l.Addr().(*net.TCPAddr).Port
+	require.NoError(t, l.Close())
+	return port
+}
```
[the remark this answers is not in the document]
rafaelroquetto: Fixed

mmat11 (approved):
lgtm!

mariomac (approved):
Amazing! Thanks!

PR #2145, opened 2026-05-19
fix(2032): restore bounded eviction when metric TTL is zero

### LINKED ISSUE #2032: TTL=0 disables metric eviction entirely
## Summary

`ExpiryMap.DeleteExpired` now treats `ttl == 0` as "disable expiration completely." When operators set metric TTL to zero, every new label combination remains in memory indefinitely.

## Impact

If attacker-influenced traffic or metadata drives high cardinality, metric state can grow without bound and eventually pressure or exhaust agent memory. The issue is configuration-gated and availability-only, so `low` severity is appropriate.

## Environment

- First identified in: 012d05ef580c9ca36262a8c77b52a3e91340e272

## Evidence

- `pkg/export/expire/expiry_map.go` returns early from `DeleteExpired` when `ttl == 0`.
- `GetOrCreate` continues inserting a new entry for each unique label set.
- `pkg/export/expire/expiry_map_test.go` explicitly asserts the zero-TTL no-expiration behavior.

## Steps to reproduce

1. From the repository root on `main`, create `/tmp/expiry_poc.go` with this exact program:
   ```go
   package main
   import (
       "fmt"
       "runtime"
       "time"
       "go.opentelemetry.io/obi/pkg/export/expire"
   )
   func main() {
       now := time.Now()
       clock := func() time.Time { return now }
       em := expire.NewExpiryMap[int](clock, 0)
       for i := 0; ; i++ {
           lbls := []string{fmt.Sprintf("label_%d", i)}
           em.GetOrCreate(lbls, func() int { return i })
           if i%100000 == 0 {
               now = now.Add(1 * time.Hour)
               em.DeleteExpired()
               var ms runtime.MemStats
               runtime.ReadMemStats(&ms)
               fmt.Printf("i=%d heap=%d\n", i, ms.HeapAlloc)
           }
       }
   }
   ```
2. Build it with `env GOTOOLCHAIN=auto go build -o /tmp/expirymap_poc /tmp/expiry_poc.go`.
3. Run it with a small memory ceiling such as `ulimit -v 65536 && /tmp/expirymap_poc`.
4. Observe steadily increasing memory usage and eventual `runtime: out of memory`, proving that `TTL=0` disables eviction and allows unbounded cache growth.

## Suggested Fix Direction

Restore bounded behavior for zero TTL or require an explicit separate flag to disable expiration entirely.

## Acceptance Criteria

- `ttl == 0` does not silently disable eviction unless that behavior is explicitly intended and separately gated.
- Metric caches remain bounded under documented default and edge-case TTL settings.
- Tests cover high-cardinality label growth with zero TTL.

## Note

NameHaibinZhang:
## Summary

Fix #2032 

- Remove the `ttl == 0` early return in `ExpiryMap.DeleteExpired()` that silently disabled all metric eviction
- `TTL=0` now means "expire immediately" (zero time-to-live), preventing unbounded memory growth under high-cardinality label sets
- Replace the old "disable expiration" test with tests verifying immediate expiration and high-cardinality eviction

## Validation

### DIFF
--- a/pkg/export/expire/expiry_map.go
+++ b/pkg/export/expire/expiry_map.go
@@ -69,11 +69,6 @@ func (ex *ExpiryMap[T]) GetOrCreate(lbls []string, instancer func() T) T {
 
 // DeleteExpired entries and return their label set
 func (ex *ExpiryMap[T]) DeleteExpired() []T {
-	// If TTL is 0, disable expiration completely
-	if ex.ttl == 0 {
-		return nil
-	}
-
 	var delKeys []string
 	var delEntries []T
 	ex.mt.RLock()
--- a/pkg/export/expire/expiry_map_test.go
+++ b/pkg/export/expire/expiry_map_test.go
@@ -4,40 +4,51 @@
 package expire
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
 )
 
-func TestExpiryMap_DisableExpiration(t *testing.T) {
-	// Mock clock that we can control
+func TestExpiryMap_ZeroTTLExpiresImmediately(t *testing.T) {
 	now := time.Now()
 	clock := func() time.Time { return now }
 
-	// Create expiry map with TTL=0 (should disable expiration)
 	em := NewExpiryMap[string](clock, 0)
 
-	// Add some entries
-	val1 := em.GetOrCreate([]string{"label1", "value1"}, func() string { return "entry1" })
-	val2 := em.GetOrCreate([]string{"label2", "value2"}, func() string { return "entry2" })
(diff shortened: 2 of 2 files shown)


NameHaibinZhang:
#2032

grcevski (commented):
I think it's fine either way, I don't think it's a big issue. Anyone can set these settings to whatever they like, so if a user sets TTL to 1000 years, they will see unbounded growth.

I'm OK to approve this, but I think the issue report is maybe not realistic?

mariomac:
Discussed/agreed during OBI SIG: users have always the chance to footgun themselves (e.g. by setting the TTL to 2 years). Instead of fixing this for the TTL==0 case, let's modify the documentation warning users about possible memory leaks if they explicitly set TTL== 0.

PR #2146, opened 2026-05-19
feat: add Go openai-go v3 Chat Completions uprobe instrumentation

NameHaibinZhang:
## Description

Add eBPF uprobe-based instrumentation for the [openai-go v3 SDK](https://github.com/openai/openai-go) (`github.com/openai/openai-go/v3`) Chat Completions API. This enables automatic collection of GenAI semantic convention attributes from Go applications using the official OpenAI Go client library, without requiring HTTP body capture.

## Motivation

Go applications using `openai-go v3` previously relied solely on HTTP-layer detection for GenAI observability. The uprobe approach provides:
- **Reliable model attribution**: Reads the request model directly from SDK memory, unaffected by HTTP compression or chunked transfer encoding
- **Token usage**: Extracts `prompt_tokens` and `completion_tokens` directly from the `CompletionUsage` struct returned by the SDK
- **Response metadata**: Captures `response.id` and `response.model` from the `ChatCompletion` struct without parsing response bodies

## Validation

#1854


grcevski:
Hi @NameHaibinZhang, I wonder if we just need to finish this item https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1941 and then we should be able to use the existing detection. Maybe check in with Stephen if you'd like to see if he needs help to get this through the finish line?

NameHaibinZhang:
Hi @grcevski, thanks for the pointer to #1941 — I took a closer look at it (and at #1938).

My understanding is that #1941 enables generic HTTP body extraction at the Go HTTP layer, which is a great long-term foundation for SDK-agnostic coverage. I do think it and the uprobe path here are complementary rather than overlapping, for a few reasons specific to the OpenAI Go SDK:

1. **Compressed / chunked / streaming responses.** The OpenAI Go SDK negotiates `Content-Encoding: gzip` (or `br`) on most non-stream responses, and uses SSE / chunked transfer for `stream=true`. Reconstructing valid JSON from the raw HTTP bytes in those cases is non-trivial. Reading from the SDK's already-decoded `ChatCompletion` / `CompletionUsage` structs gives us deterministic `response.id`, `response.model`, `prompt_tokens`, `completion_tokens` regardless of wire encoding.
2. **Avoiding large-buffer payload reconstruction.** The HTTP-layer path (per #1938) needs enlarged event buffers and JSON reassembly across multiple events. The uprobe path sidesteps that overhead entirely for the openai-go SDK case.
3. **Streaming usage.** In SSE chat completions, `usage` is only emitted on the final chunk (and only when `stream_options.include_usage=true`). The uprobe observes the final assembled struct the SDK returns, which is simpler and more robust than stream parsing.

That said, I fully agree #1941 is the right direction for broader, non-Go-SDK coverage, and I'd be glad to sync with @skl and help move #1941 / #1938 forward.

Would you be open to landing this uprobe path as an **SDK-specific complement** (gated behind openai-go offset detection, zero impact when the SDK isn't present), while #1941 progresses for the generic HTTP-layer story? I'm also happy to adjust scope, split commits, or hold this PR until #1941 lands — whichever you prefer.

grcevski (commented):
This is really good work! 

I only have one general concern, I think (I haven't tried this myself) that as it's implemented now, we'll get one OpenAI event and one HTTP (client) event for every request, or multiple HTTP clients calls?

I think we need to stop the HTTP client call from happening. We've done this before, we need to check in the HTTP client code if there's an OpenAI map for the same request, e.g. query `ongoing_openai_requests`. If there's something there we can simply refuse to handle the HTTP client.

Another thing that we need to figure out is the connection information. Everything in OBI relies on the connection information and that needs to be added to the `request.Span`. To do that there are two options:

1. In the HTTP client call, when you find that you have an entry in `ongoing_openai_requests`, write the connection information before you reject to make the request.
2. Add an if statement for `ongoing_openai_requests` in the handler for netFD.Read. (go_net.c)
3. The connection info might be a field in some of the OpenAI structs, so it's possible maybe to pull from there.

I think option 3. and then 1. should be preferred if it's possible to do.

bpf/common/common.h
```diff
@@ -281,3 +284,22 @@ typedef struct dns_req {
     unsigned char buf[k_dns_max_len];
     u8 _pad3[4];
 } dns_req_t;
+
+typedef struct openai_go_req {
```
rafaelroquetto, line 289: could you please double check the padding? I feel this can be improved/squashed.

bpf/gotracer/maps/ongoing_openai_requests.h
```diff
@@ -0,0 +1,25 @@
...
+#pragma once
+
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_helpers.h>
+
+#include <common/common.h>
+#include <common/go_addr_key.h>
+#include <common/map_sizing.h>
+
+struct {
```
rafaelroquetto, line 14: you should pin these maps to OBI_PIN_INTERNAL

rafaelroquetto (commented):
Quick pass, will do another one.

bpf/gotracer/maps/ongoing_openai_requests.h
NameHaibinZhang: Done. Added `__uint(pinning, OBI_PIN_INTERNAL)` to both `ongoing_openai_requests` and `openai_req_mem`.

bpf/common/common.h
NameHaibinZhang: Done. Moved `input_message_role` up next to `type` to squash the trailing 7-byte pad — net saving of 8 bytes (816 → 848 bytes after also adding the new `connection_info_t conn` field requested in the architecture review).

NameHaibinZhang:
@grcevski 

Addressed both concerns from the architecture review:

1. **Event deduplication**: Added a check in `obi_uprobe_roundTripReturn` — if the goroutine key exists in `ongoing_openai_requests`, we skip emitting the HTTP client span and jump to `done` for proper cleanup.

2. **Connection info** (option 1): Before skipping, we look up `ongoing_client_connections` and memcpy the connection info into the OpenAI request entry. The Go-side `ReadGoOpenAIRequestIntoSpan` now extracts Peer/Host from the `conn` field.

bpf/gotracer/maps/openai.h
```diff
@@ -0,0 +1,28 @@
...
+    __uint(pinning, OBI_PIN_INTERNAL);
+} ongoing_openai_requests SEC(".maps");
+
+struct {
+    __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+    __type(key, u32);
+    __type(value, openai_go_req_t);
+    __uint(max_entries, 1);
+    __uint(pinning, OBI_PIN_INTERNAL);
+} openai_req_mem SEC(".maps");
```
rafaelroquetto: please use SCRATCH_MEM and rename this file to ongoing_openai_requests.h

bpf/gotracer/go_nethttp.c
```diff
@@ -711,6 +712,18 @@ int obi_uprobe_roundTripReturn(struct pt_regs *ctx) {
...
     go_addr_key_from_id(&g_key, goroutine_addr);
 
+    // If this HTTP client request belongs to an in-flight OpenAI SDK call,
+    // attach the connection info to the OpenAI entry and skip emitting an
+    // HTTP client span (the OpenAI uprobe will emit a GenAI span instead).
+    openai_go_req_t *openai_req = bpf_map_lookup_elem(&ongoing_openai_requests, &g_key);
+    if (openai_req) {
+        connection_info_t *conn_info = bpf_map_lookup_elem(&ongoing_client_connections, &g_key);
+        if (conn_info) {
+            __builtin_memcpy(&openai_req->conn, conn_info, sizeof(connection_info_t));
```
rafaelroquetto: nit:

it should suffice to do `openai_req->conn = *conn_info;`

rationale: readability and no size mismatch

bpf/common/common.h
```diff
@@ -281,3 +284,23 @@ typedef struct dns_req {
...
+    u8 type;               // Must be first
+    u8 input_message_role; // 0=user, 1=system, 2=assistant, 3=developer, 4=tool, 5=function
```
rafaelroquetto: instead of the comment, you should define the openai_role_enum in a separate header (perhaps together with the other enums), and use it here instead of a plain `u8`

bpf/gotracer/go_openai.c
```diff
@@ -0,0 +1,318 @@
...
+#include <gotracer/go_common.h>
+#include <gotracer/go_str.h>
+
+#include <gotracer/maps/openai.h>
+
+#include <logger/bpf_dbg.h>
+
+#include <shared/obi_ctx.h>
+
+// Stack offset for the body parameter in the Go register ABI calling convention.
```
rafaelroquetto: Most of these comments could/should be removed as they are just restating the code.

bpf/gotracer/go_openai.c
```diff
@@ -0,0 +1,318 @@
...
+//   ctx context.Context        -> BX  (GO_PARAM2) type, CX (GO_PARAM3) data
+//   body ChatCompletionNewParams -> stack at SP+8 (too large for remaining registers)
+//   opts ...option.RequestOption -> stack (variadic slice)
+SEC("uprobe/openai_chat_new")
+int obi_uprobe_openai_chat_new(struct pt_regs *ctx) {
+    void *goroutine_addr = GOROUTINE_PTR(ctx);
+    bpf_dbg_printk("=== uprobe/openai_chat_new goroutine_addr=%lx ===", goroutine_addr);
+
+    u32 zero = 0;
+    openai_go_req_t *req = bpf_map_lookup_elem(&openai_req_mem, &zero);
```
rafaelroquetto: here is wehere you can use SCRATCH_MEM (you can declare it at the top of this file)

bpf/gotracer/go_openai.c
```diff
@@ -0,0 +1,318 @@
...
+    req->input_message_role = k_openai_role_unknown;
+
+    client_trace_parent(goroutine_addr, &req->tp);
+
+    const u64 model_off = go_offset_of(ot, (go_offset){.v = _openai_chat_params_model_pos});
+
+    // Pre-offset the base pointer so read_go_str can consume a wide offset
+    // (its `offset` parameter is u8). body_ptr + model_off then points at the
+    // Go string header (ptr + len) we want to dereference.
+    read_go_str("openai request model",
```
rafaelroquetto, line 192: I beleive the proper approach here would be to change the `offset` type to `u32`.

bpf/gotracer/go_openai.c
```diff
@@ -0,0 +1,318 @@
...
+
+// github.com/openai/openai-go/v3.(*ChatCompletionService).New
+// func (r *ChatCompletionService) New(ctx context.Context, body ChatCompletionNewParams,
+//                                     opts ...option.RequestOption) (*ChatCompletion, error)
+//
+// Go 1.17+ register ABI (AMD64):
+//   r   *ChatCompletionService -> AX  (GO_PARAM1)
+//   ctx context.Context        -> BX  (GO_PARAM2) type, CX (GO_PARAM3) data
+//   body ChatCompletionNewParams -> stack at SP+8 (too large for remaining registers)
+//   opts ...option.RequestOption -> stack (variadic slice)
```
rafaelroquetto: This comment is misleading as it implies this only works on AMD64 but we also support arm.

In general, could you please reivisit all of this comments and trim them to the bare minimum. Obvious comments restating the code are a maintenanance burden and source of  confusion, because they tend to drift.

Overall, I wouldn't bother restating the go calling conventions, anyone touching these uprobes should be aware. A simple "body" spills to the stack suffice if you must

bpf/gotracer/go_openai.c
```diff
@@ -0,0 +1,318 @@
...
+        bpf_dbg_printk("openai prompt_tokens=%lld completion_tokens=%lld",
+                       req->prompt_tokens,
+                       req->completion_tokens);
+
+        read_first_output_choice(ot, resp_ptr, req);
+    }
+
+    openai_go_req_t *trace = bpf_ringbuf_reserve(&events, sizeof(openai_go_req_t), 0);
+    if (trace) {
+        __builtin_memcpy(trace, req, sizeof(openai_go_req_t));
```
rafaelroquetto: since you are copiying, you can just use `bpf_ringbuf_output` here anyway

bpf/gotracer/go_openai.c
```diff
@@ -0,0 +1,318 @@
...
+                                usage_ptr + comp_tokens_off) != 0) {
+            bpf_dbg_printk("can't read openai completion_tokens");
+        }
+        if (bpf_probe_read_user(&req->prompt_tokens,
+                                sizeof(req->prompt_tokens),
+                                usage_ptr + prompt_tokens_off) != 0) {
+            bpf_dbg_printk("can't read openai prompt_tokens");
+        }
+
+        bpf_dbg_printk("openai response_id=[%s] model=[%s]", req->response_id, req->response_model);
```
rafaelroquetto, line 261: I am not familiar with the protocol, is it possible  that we are leaking sensitive tokens here to the log?

bpf/gotracer/go_openai.c
```diff
@@ -0,0 +1,318 @@
...
+        bpf_dbg_printk("can't read openai choices len");
+        return;
+    }
+
+    if (!choices_arr || choices_len == 0) {
+        return;
+    }
+
+    // Content lives inside the first Choice's Message struct.
+    void *content_hdr = (unsigned char *)choices_arr + choice_message_off + message_content_off;
```
rafaelroquetto:
```suggestion
    unsigned char *content_hdr = (unsigned char *)choices_arr + choice_message_off + message_content_off;
```

bpf/common/common.h
rafaelroquetto: The alignment becomes more obvious when we have similar-sized types grouped together, so could you please change this to:

```
typedef struct openai_go_req {
    u8 type;               // Must be first
    u8 input_message_role; // 0=user, 1=system, 2=assistant, 3=developer, 4=tool, 5=function
    u8 _pad[6];
    u64 start_monotime_ns;
    u64 end_monotime_ns;
    s64 prompt_tokens;
    s64 completion_tokens;
    pid_info pid;
    u8 _pad2[4];
    tp_info_t tp;
    connection_info_t conn;
    unsigned char request_model[k_openai_model_max_len];
    unsigned char response_model[k_openai_model_max_len];
    unsigned char response_id[k_openai_response_id_max_len];
    unsigned char input_message_content[k_openai_msg_content_max_len];
    unsigned char output_message_content[k_openai_msg_content_max_len];
    u8 _pad3[4];
} openai_go_req_t;
```

pkg/ebpf/common/openai_detect_transform.go
```diff
@@ -0,0 +1,206 @@
...
+//	    u8 _pad3[4];
+//	    unsigned char request_model[64];
+//	    unsigned char response_model[64];
+//	    unsigned char response_id[64];
+//	    s64 prompt_tokens;
+//	    s64 completion_tokens;
+//	    unsigned char input_message_content[256];
+//	    unsigned char output_message_content[256];
+//	} openai_go_req_t;
+type openaiGoReqT struct {
```
rafaelroquetto: We must not duplicate auto-generated structs (they will drift, this is too error prone) - these are already auto-generated by bpf2go - see BpfOpenaiGoReqT

bpf/gotracer/maps/ongoing_openai_requests.h
```diff
@@ -0,0 +1,28 @@
...
+#include <common/map_sizing.h>
+#include <common/pin_internal.h>
+
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, go_addr_key_t);     // key: goroutine id
+    __type(value, openai_go_req_t); // the in-flight Chat Completion request
+    __uint(max_entries, MAX_CONCURRENT_REQUESTS);
+    __uint(pinning, OBI_PIN_INTERNAL);
+} ongoing_openai_requests SEC(".maps");
```
rafaelroquetto, line 20: This is a huge map - together with the other protocols (which we should probably revisit, I will open an issue), we should only load this (and anything related to openai) on demand - otherwise we end up pilling up used memory (this alone is about 8.5 MB).

Look up fixupSpec in the code for how to do it. You will need something similar.

rafaelroquetto (changes requested):

bpf/gotracer/maps/ongoing_openai_requests.h
rafaelroquetto: For future reference: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2159

bpf/gotracer/go_nethttp.c
NameHaibinZhang: Fixed. Changed to `openai_req->conn = *conn_info;`.

bpf/common/common.h
NameHaibinZhang: Fixed. Reordered the struct fields as suggested: moved prompt_tokens/completion_tokens after end_monotime_ns and before char arrays; moved _pad3 to the end. Also replaced the plain `u8` with `enum openai_role`.

bpf/gotracer/maps/openai.h
NameHaibinZhang: Fixed. Renamed the file to `ongoing_openai_requests.h` and switched to `SCRATCH_MEM_TYPED` macro declared in `go_openai.c` (consistent with go_redis.c pattern).

bpf/gotracer/go_openai.c
NameHaibinZhang: Fixed. Replaced the manually defined percpu array map with `SCRATCH_MEM_TYPED(openai_scratch, openai_go_req_t)` and now obtain the scratch buffer via `openai_scratch_mem()`.

bpf/common/common.h
NameHaibinZhang: Fixed. Created `bpf/common/openai_enums.h` defining `enum openai_role : u8` and used it in the struct instead of a plain `u8`.

bpf/gotracer/go_openai.c
NameHaibinZhang: Fixed. Removed comments that merely restate the code logic, keeping only necessary function documentation.

bpf/gotracer/maps/openai.h
NameHaibinZhang: ok

bpf/gotracer/go_openai.c
NameHaibinZhang: ok

bpf/gotracer/go_openai.c
NameHaibinZhang: Fixed. Switched to `bpf_ringbuf_output(&events, req, sizeof(openai_go_req_t), get_flags())` directly, eliminating the reserve+memcpy+submit pattern.

bpf/gotracer/go_openai.c
NameHaibinZhang: Fixed. Changed the `offset` parameter type from `u8` to `u32` in `read_go_str` and `peek_go_str_len`, and now pass the offset directly instead of pre-offsetting the base pointer.

bpf/gotracer/go_openai.c
NameHaibinZhang: Fixed. Removed all misleading comments about AMD64 register ABI and other redundant comments that restate the code.

bpf/gotracer/go_openai.c
NameHaibinZhang: Thanks for raising the security concern! The log here outputs the response_id (e.g. `chatcmpl-abc123`) and model name (e.g. `gpt-4o-mini`), neither of which is sensitive. The OpenAI API key resides in the HTTP Authorization header, which is not captured by our uprobe (it hooks the SDK's ChatCompletionService.New method whose parameters only contain the request body, not transport-layer headers). Additionally, these logs are only emitted when `bpf_debug` is enabled.

bpf/gotracer/go_openai.c
NameHaibinZhang: Fixed. Changed to `unsigned char *content_hdr = ...`.

pkg/ebpf/common/openai_detect_transform.go
NameHaibinZhang: Fixed. Removed the manually defined `openaiGoReqT` struct and added a type alias `GoOpenAIInfo BpfOpenaiGoReqT` in `common.go`, directly using the bpf2go auto-generated type.

bpf/gotracer/maps/ongoing_openai_requests.h
NameHaibinZhang: Thanks for flagging this! I see you've opened issue #2159 to track it. The on-demand loading implementation involves the gotracer's fixupSpec/conditional loading mechanism which is fairly broad in scope — I'll follow up in a separate PR.

PR #2147, opened 2026-05-19
CI: polish workflows, refactor VM integration testing

mmat11:
## Summary

- migrated built-in kernels to a subset of the lvh catalog (see: https://github.com/cilium/little-vm-helper)
- bpf-verifier tests run on all of these kernels while VM integration tests run on only 2 (conditionally on all with a user supplied command `/all-vm-tests`)
- VM integration testing coverage increased from only context propagation to context propagation + log enrichment + large buffers
- decoupled workflows, in current main, verifier tests and other checks were done in an umbrella job
- added some modular actions like `free-disk`/`generate-bpf` which were hardcoded in every workflow
- polished jobs/workflows names for consistency
- fixed gotracer loading in 5.10; extended `FixupSpec` to dummy-out `obi_uprobe_readMimeHeader` when `bpf_loop` unsupported, mirroring the existing http/http_legacy pattern

TODO in another PR: 

fix RHEL failures, 1M instructions limit is hit easier on those kernels, needs deeper investigation

## Validation


rafaelroquetto:
@mmat11 is it possible to split the PR or are they all somehow related? (I haven't looked into it yet)

mmat11:
> @mmat11 is it possible to split the PR or are they all somehow related? (I haven't looked into it yet)

I split everything in separate commits

internal/test/vm/Makefile
```diff
@@ -13,31 +24,37 @@ RUN_NUMBER ?= 1
 TEST_OUTPUT = $(REPO_ROOT)/testoutput
 MNT ?= mnt
 IMG_SIZE ?= 20G
-QEMU_BIN = qemu-system-$(ARCH)
+QEMU_BIN = qemu-system-x86_64
```
rafaelroquetto: I'd keep the arch linux as it was, makes it easier to switch in the future

internal/test/vm/Makefile
```diff
@@ -13,31 +24,37 @@ RUN_NUMBER ?= 1
...
 # Ensure test output directory exists
 $(TEST_OUTPUT):
 	mkdir -p $(TEST_OUTPUT)
 
 $(ROOT_IMG): Dockerfile
-	# Build the Docker image and create a tarball
+	# Build the Docker image and create a tarball. Test args (TARGET_TEST,
+	# TEST_PATTERN, RUN_NUMBER) are no longer baked here; they get written to
+	# testoutput/test-env at launchvm time, letting the qcow2 be cached
+	# across kernel/test matrix entries.
```
rafaelroquetto: I think this comment should not mention TARGET_TEST, etc since they are not even part of this file anymore - this is just claude giving meaningless context

rafaelroquetto (commented):
Overall it looks good to me, just one minimal comment and one question: what's the reason why we no longer have clang-format and clang-tidy as independent workflows? I _personally_ find having separate workflows easier to maintain - and it also enables us to trigger re-runs and what not independently.

internal/test/vm/Makefile
mmat11: he does that no matter how many times I try to tell him to be concise with comments 😞

rafaelroquetto: Bad bad Claudio! No donnut for him!

rafaelroquetto (approved):

PR #2148, opened 2026-05-19
Improve TCP unmatched parsing performance

grcevski:
## Summary

Users have reported that on high request volume they are missing some amount of traces and metrics. After investigations we discovered that we have a lot of events in the BPF logs where we cannot get enough data on the ring buffer. This means that we are unable to drain the ring buffer in userspace fast enough to keep up with the volume of data. 

This proved to be a relatively recent regression after we started adding all the new protocol detectors. From the linked Beyla issue https://github.com/grafana/beyla/issues/2707 we saw that the majority of the events are TCP, rather than HTTP which would explain why things got slower.

I added a benchmark that tests the worst case scenario for TCP, by creating random stream of bytes and passing it down the protocol detectors. The main overhead were allocations and missed opportunities to easily discard not matching protocols. One thing we have to pay attention to in the future is that doing `fmt.Errorf` and 'errors.New` on regular protocol parsing is allocating new error every time. Since most of the time the protocols don't match at least some of the detectors, this adds up over time.

Heuristic SQL detection is also very slow, since we iterate over all of the buffers looking for the SQL statements, but this is off by default.

Here are some benchmark results before and after my changes:

Before:
```
BenchmarkReadTCPRequestIntoSpan_RandomGarbage-24       	 9618645	      3560 ns/op	   28376 B/op	      27 allocs/op
```
image

After (around 6x improvement):
```
BenchmarkReadTCPRequestIntoSpan_RandomGarbage-24    	53958400	       624.9 ns/op	     190 B/op	       4 allocs/op
```
image

I've split the protocol by protocol changes in separate commits. Most of the changes are effectively replacing `fmt.Errorf` and `errors.New` with global constants.

More work is needed in the future to look into specific protocols and investigate how much overhead does the actual parsing take for various protocols.

## Validation

### DIFF
--- a/pkg/ebpf/common/fast_cgi_detect_transform.go
+++ b/pkg/ebpf/common/fast_cgi_detect_transform.go
@@ -122,17 +122,23 @@ func maybeFastCGI(b *largebuf.LargeBuffer) bool {
 	if b.Len() <= fastCGIRequestHeaderLen {
 		return false
 	}
-	view := b.UnsafeView()
 	// FastCGI 1.0: every record starts with version=1 and a record type in
 	// 1..11. Cheap 2-byte check that filters ~99.98% of non-FastCGI payloads
 	// before the more expensive REQUEST_METHOD substring scan.
-	if view[0] != fcgiVersion1 {
+
+	ver, err := b.U8At(0)
+
+	if err != nil || ver != fcgiVersion1 {
 		return false
 	}
-	if view[1] < fcgiFrameTypeBeginReq || view[1] > fcgiFrameTypeUnknown {
+
+	frameType, err := b.U8At(1)
+
+	if err != nil || frameType < fcgiFrameTypeBeginReq || frameType > fcgiFrameTypeUnknown {
 		return false
 	}
-	return bytes.Contains(view, []byte(requestMethodKey))
+
+	return bytes.Contains(b.UnsafeView(), []byte(requestMethodKey))
 }
 
 func parseHeader(b *largebuf.LargeBuffer) ([]byte, error) {
--- a/pkg/ebpf/common/fast_cgi_detect_transform.go
+++ b/pkg/ebpf/common/fast_cgi_detect_transform.go
@@ -7,7 +7,6 @@ import (
 	"bytes"
 	"encoding/binary"
 	"errors"
-	"fmt"
 	"io"
 	"strconv"
 	"unsafe"
@@ -25,7 +24,17 @@ const (
 	responseStatusKey       = "Status: "
 )
 
-const fcgiFrameTypeParams = 4
+const (
+	fcgiVersion1          = 1
+	fcgiFrameTypeBeginReq = 1
+	fcgiFrameTypeUnknown  = 11
+	fcgiFrameTypeParams   = 4
+)
+
+var (
+	errFastCGIPayloadTooShort  = errors.New("payload too short")
+	errFastCGIHeaderReadFailed = errors.New("failed to read FastCGI header")
+)
 
 // fastCGIHeader represents the structure of a FastCGI header
 type fastCGIHeader struct {
@@ -43,7 +52,7 @@ func readFastCGIHeader(b []byte) (*fastCGIHeader, error) {
 	// FastCGI header is always 8 bytes
 	headerBytes := make([]byte, fastCGIRequestHeaderLen)
 	if _, err := io.ReadFull(reader, headerBytes); err != nil {
-		return nil, fmt.Errorf("failed to read FastCGI header: %w", err)
+		return nil, errFastCGIHeaderReadFailed
 	}
 
 	// Parse the header
@@ -51,22 +60,22 @@ func readFastCGIHeader(b []byte) (*fastCGIHeader, error) {
 	buffer := bytes.NewReader(headerBytes)
 
 	if err := binary.Read(buffer, binary.BigEndian, &header.Version); err != nil {
-		return nil, fmt.Errorf("failed to read version: %w", err)
+		return nil, errFastCGIHeaderReadFailed
 	}
 	if err := binary.Read(buffer, binary.BigEndian, &header.Type); err != nil {
-		return nil, fmt.Errorf("failed to read type: %w", err)
+		return nil, errFastCGIHeaderReadFailed
 	}
 	if err := binary.Read(buffer, binary.BigEndian, &header.RequestID); err != nil {
-		return nil, fmt.Errorf("failed to read request ID: %w", err)
+		return nil, errFastCGIHeaderReadFailed
 	}
 	if err := binary.Read(buffer, binary.BigEndian, &header.ContentLength); err != nil {
-		return nil, fmt.Errorf("failed to read content length: %w", err)
+		return nil, errFastCGIHeaderReadFailed
 	}
 	if err := binary.Read(buffer, binary.BigEndian, &header.PaddingLength); err != nil {
-		return nil, fmt.Errorf("failed to read padding length: %w", err)
+		return nil, errFastCGIHeaderReadFailed
 	}
 	if err := binary.Read(buffer, binary.BigEndian, &header.Reserved); err != nil {
-		return nil, fmt.Errorf("failed to read reserved byte: %w", err)
+		return nil, errFastCGIHeaderReadFailed
 	}
 
 	return header, nil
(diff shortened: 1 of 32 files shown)


pkg/ebpf/common/fast_cgi_detect_transform.go
```diff
@@ -113,34 +122,44 @@ func maybeFastCGI(b *largebuf.LargeBuffer) bool {
...
+	// FastCGI 1.0: every record starts with version=1 and a record type in
+	// 1..11. Cheap 2-byte check that filters ~99.98% of non-FastCGI payloads
+	// before the more expensive REQUEST_METHOD substring scan.
+	if view[0] != fcgiVersion1 {
+		return false
+	}
+	if view[1] < fcgiFrameTypeBeginReq || view[1] > fcgiFrameTypeUnknown {
+		return false
+	}
+	return bytes.Contains(view, []byte(requestMethodKey))
```
rafaelroquetto:
```suggestion
	// FastCGI 1.0: every record starts with version=1 and a record type in
	// 1..11. Cheap 2-byte check that filters ~99.98% of non-FastCGI payloads
	// before the more expensive REQUEST_METHOD substring scan.
	
	ver, err := b.U8At(0)
	
	if err || ver != fcgiVersion1 {
		return false
	}
	
	frameType, err := b.U8At(1)
	
	if err || frameType < fcgiFrameTypeBeginReq || frameType > fcgiFrameTypeUnknown {
		return false
	}
	return bytes.Contains(b.UnsafeView(), []byte(requestMethodKey))
```

rationale: `UnsafeView()` is expensive, because it can linerialise data - so if we are returning early in 99.98% of the cases, we can do without materialising it and just use `U8At()`

pkg/ebpf/common/memcached_detect_transform.go
```diff
@@ -22,6 +22,23 @@ const (
...
+var memcachedFirstByte = [256]bool{
+	'a': true, 'c': true, 'd': true, 'f': true, 'g': true,
+	'i': true, 'm': true, 'p': true, 'r': true, 's': true,
+	't': true, 'v': true,
+	'V': true, 'E': true, 'S': true, 'N': true, 'D': true,
+	'T': true, 'O': true, 'C': true, 'H': true, 'M': true,
+	'0': true, '1': true, '2': true, '3': true, '4': true,
+	'5': true, '6': true, '7': true, '8': true, '9': true,
+}
+
```
rafaelroquetto:
```suggestion
var memcachedFirstByteBitmap = func() [4]uint64 {
	var m [4]uint64
	for _, c := range []byte("acdfgimprstvVESNDTOCHM0123456789") {
		m[c>>6] |= 1 << (c & 63)
	}
	return m
}()

func isMemcachedFirstByte(b byte) bool {
	return memcachedFirstByteBitmap[b>>6]&(1<<(b&63)) != 0
}
```

rationale: the above bitmap is 8 times smaller - for the hotpath, it spans a single cache line (64 bytes), so in the hot path it keeps the cache warm (the [256]bool spans 4 cache lines on x64)

pkg/ebpf/common/mqtt_detect_transform.go
```diff
@@ -48,11 +50,20 @@ func packetTypeToMethod(packetType mqttparser.PacketType) string {
 // Otherwise, returns MQTTInfo with the processed data. The ignore bool indicates whether the event
 // should be ignored for span creation (e.g., control packets like CONNECT).
 func ProcessPossibleMQTTEvent(event *TCPRequestInfo, pkt *largebuf.LargeBuffer, rpkt *largebuf.LargeBuffer) (*MQTTInfo, bool, error) {
-	m, ignore, err := ProcessMQTTEvent(pkt.UnsafeView())
+	pktView := pkt.UnsafeView()
+	rpktView := rpkt.UnsafeView()
+
+	// Cheap prefilter: if neither buffer has a plausible MQTT fixed-header
+	// byte 0, skip the variable-length parsing entirely.
+	if !mqttparser.IsLikelyMQTT(pktView) && !mqttparser.IsLikelyMQTT(rpktView) {
```
rafaelroquetto: same as the other comment above, UnsafeView() is relatively expensive - I checked IsLikelyMQTT and it only cares about the first byte and the packet length, so perhaps you could write this as:

```
isLikelyMQTT := func(pkt *largebuf.LargeBuffer) {
    startByte, err := pkt.U8At(0)

    return err == nil && mqttparse.IsLikelyMQTT(startByte, pkt.Len())
}

if !isLikelyMQTT(pkt) || !isLikelyMQTT(rpkt) {
    return nil, true, errNotMQTT
}

````

this also restores the original behaviour in which we only materialised the rpkt UnsafeView() if required.

File: pkg/ebpf/common/nats_detect_transform.go [hunk 4]
```diff
@@ -15,6 +15,25 @@ import (
...
+	'P': true, 'p': true,
+	'I': true, 'i': true,
+	'C': true, 'c': true,
+	'S': true, 's': true,
+	'U': true, 'u': true,
+	'M': true, 'm': true,
+	'H': true, 'h': true,
+	// needed for NATS server protocol if we see any
+	'R': true, 'r': true,
+}
```
rafaelroquetto:
```suggestion
var natsCmdFirstByteBitmap = func() [4]uint64 {
	var m [4]uint64
	for _, c := range []byte("+-PpIiCcSsUuMmHhRr") {
		m[c>>6] |= 1 << (c & 63)
	}
	return m
}()

func isNATSCmdFirstByte(b byte) bool {
	return natsCmdFirstByteBitmap[b>>6]&(1<<(b&63)) != 0
}
```

as above

pkg/ebpf/common/nats_detect_transform.go
```diff
@@ -91,7 +141,15 @@ func ProcessPossibleNATSEvent(event *TCPRequestInfo, pkt *largebuf.LargeBuffer,
...
 	if pkt == nil || pkt.Len() == 0 {
-		return nil, true, errors.New("packet too short for NATS")
+		return nil, true, errPacketTooShortForNATS
+	}
+
+	// Cheap prefilter: every NATS frame starts with the first letter of one of
+	// the protocol commands (+OK, -ERR, PING/PONG/PUB, INFO, CONNECT, SUB,
+	// UNSUB, MSG, HPUB/HMSG). Reject anything else without invoking the
+	// frame parser.
+	if !natsCmdFirstByte[pkt.UnsafeView()[0]] {
```
rafaelroquetto: here we should use `pkt.U8At(0)`

pkg/internal/ebpf/amqpparser/protocol.go
```diff
@@ -44,6 +45,50 @@ func startsWithMagic(r *largebuf.LargeBufferReader) bool {
 	return err == nil && bytes.Equal(data, amqpMagic)
 }
 
+// IsLikelyAMQP is an O(1) prefilter that reports whether buf could plausibly
+// start an AMQP 1.0 conversation. It accepts either the connection preamble
+// ("AMQP" magic) or a structurally valid frame header. The full Parse should
+// still be invoked to confirm — this only filters out obviously non-AMQP
+// payloads cheaply.
+// https://docs.oasis-open.org/amqp/core/v1.0/csprd01/amqp-core-transport-v1.0-csprd01.html
+func IsLikelyAMQP(buf []byte) bool {
```
rafaelroquetto: also here, this is triggering an undesirable call to `UnsafeView()`. 

I reckon the `largebuffer.Reader` API is a better fit here, because it can directly read without materialising the view.

So either we need a dependency on large buffer here, or if that is not possible, I'd just move this function inside amqp_detect_transform.go

rafaelroquetto (commented):
Nice that you found this and optimised the hell out of it! In hindsight (it's always easy in hindsight) this is all so obvious now! Looking good, I just left a few comments, mostly in relation to the usage of UnsafeView(), which we should avoid / defer until the last moment.

pkg/ebpf/common/memcached_detect_transform.go
```diff
@@ -50,6 +67,14 @@ func isMemcachedBuf(buf *largebuf.LargeBuffer) bool {
 		return false
 	}
 
+	// Cheap prefilter: every memcached text frame starts with a command
```
mmat11, line 72: doesn't this match _a lot_ of false positives? I wonder if it isn't just faster to proceed past this regardless of the first byte

mmat11: (same for NATS and others)

mmat11 (commented):
nice findings!

claude tells me some errors have been missed:

File | Per-call err allocs | Notes
-- | -- | --
pkg/ebpf/common/sql_detect_postgres.go | 14 | PR patched sql_detect_transform.go + the mssql test, skipped Postgres entirely. Every short-buffer / mid-parse short-read does errors.New("too short"). These fire on every non-match.
pkg/ebpf/common/kafka_detect_transform.go | 3 | errors.New("unsupported Kafka API key") (×2), errors.New("no response buffer for metadata request"). PR patched kafkaparser/common.go but not the dispatcher.
pkg/internal/ebpf/kafkaparser/fetch.go | 2 | "offset exceeds packet size", "no Topics found in fetch request"
pkg/internal/ebpf/kafkaparser/metadata.go | 1 | "no Topics found in metadata response"
pkg/internal/ebpf/kafkaparser/produce.go | 1 | "no Topics found in produce request"
pkg/internal/ebpf/mqttparser/packet_subscribe.go | 4 | "insufficient data", "failed to read topic filter", "failed to read subscription options", "must contain at least one subscription"
pkg/internal/ebpf/mqttparser/packet_publish.go | 1 | "insufficient data for PUBLISH packet"


File:line | Pattern
-- | --
pkg/internal/ebpf/mqttparser/packet_connect.go:41 | errors.New("insufficient data for CONNECT packet") — simple, same fix
pkg/internal/ebpf/mqttparser/packet_connect.go:186 | fmt.Errorf("%q is not a recognized mqtt protocol name", name) — formats dynamic input; either drop the dynamic name (use static msg "unknown mqtt protocol name") or define a sentinel var ErrUnknownMQTTProtoName = errors.New(...) and stop carrying the offending bytes upstream
pkg/internal/ebpf/mqttparser/packet_connect.go:211 | fmt.Errorf("%d is not a recognized mqtt protocol level", level) — same

pkg/ebpf/common/memcached_detect_transform.go
grcevski: fixed

pkg/ebpf/common/memcached_detect_transform.go
grcevski: I think every little bit helps, I kept iterating and I saw small improvements in the profile. I'm testing with garbage bytes, so in this case it's 32/256 false positive = 12.5%. I think it's worth it.

pkg/ebpf/common/nats_detect_transform.go
grcevski: fixed.

pkg/ebpf/common/nats_detect_transform.go
grcevski: fixed.

pkg/internal/ebpf/amqpparser/protocol.go
grcevski: Fixed. I didn't realise UnsafeView() had that cost, good to know.

pkg/ebpf/common/fast_cgi_detect_transform.go
grcevski: fixed.

pkg/ebpf/common/mqtt_detect_transform.go
grcevski: fixed.

grcevski:
> nice findings!
> 
> claude tells me some errors have been missed:
> 

Nice! I've fixed those too now.

rafaelroquetto (approved):
This is gnarly!

mmat11 (approved):
🚀

PR #2156, opened 2026-05-21
Fix #2035: Apply unresolved-host renaming to addr attributes

### LINKED ISSUE #2035: Unresolved-host renaming does not cover server.addr
## Summary

OBI's unresolved-host renaming logic still rewrites only the `client` and `server` name-style attributes. Address-style attributes such as `server.addr` continue to export raw unresolved IP values.

## Impact

Internal IP addresses can still appear in metrics, and address-based label cardinality remains unbounded despite unresolved-host renaming being enabled. The exposure is limited to telemetry outputs, so `low` severity is appropriate.

## Environment

- First identified in: 26a5ffac9ffc88f6382478a5688dd8c8cd0976c6

## Evidence

- `pkg/appolly/app/request/span_getters_unresolved.go` only rewrites `attr.Client` and `attr.Server`.
- `pkg/appolly/app/request/span_getters.go` derives `ServerAddr` and `ClientAddr` directly from raw host/peer values.
- `pkg/export/attributes/attr_defs.go` still enables `server.addr` in default metric groups.

## Steps to reproduce

1. From the repository root on `main`, create `/tmp/unresolved_poc.go` with this exact program:
   ```go
   package main
   import (
       "fmt"
       "go.opentelemetry.io/obi/pkg/app/request"
   )
   func main() {
       span := request.Span{
           Type:      request.EventTypeHTTP,
           HostName:  "192.168.1.1",
           Host:      "10.0.0.1",
           PeerName:  "192.168.1.2",
           Peer:      "10.0.0.2",
           Statement: "http;192.168.1.3:8080",
       }
       getters := request.SpanOTELGetters("unresolved")
       for _, name := range []attr.Name{attr.Client, attr.ClientAddr, attr.Server, attr.ServerAddr} {
           getter, ok := getters(name)
           if !ok {
               fmt.Printf("%s: no getter\n", name)
               continue
           }
           kv := getter(&span)
           fmt.Printf("%s=%s\n", name, kv.Value.AsString())
       }
   }
   ```
2. Run it with `GOTOOLCHAIN=auto go run /tmp/unresolved_poc.go`.
3. Let the PoC print both the renamed host attributes and the address attributes.
4. Observe that `client` and `server` become `unresolved` while `client.address` and `server.address` still print raw IPs, proving the address fields remain exposed.

## Suggested Fix Direction

Apply unresolved-host renaming consistently to address-style attributes, or stop exporting raw addr labels by default when unresolved-host replacement is enabled.

## Acceptance Criteria

- Unresolved IP replacement covers the exported addr attributes or those attributes are disabled by default in that mode.
- Metrics no longer expose raw unresolved IPs when unresolved-host renaming is enabled.
- Tests cover both name and addr attributes.

## Note

NameHaibinZhang:
## Summary

Fixes open-telemetry/opentelemetry-ebpf-instrumentation#2035

OBI's unresolved-host renaming logic only covered `client` and `server` name-style attributes. Address-style attributes (`client.address`, `server.address`) continued to export raw unresolved IP values.

## Changes

**`pkg/appolly/app/request/span_getters_unresolved.go`**:
- Extended the `otelUnresolvedHostGetters` switch cases to include `attr.ClientAddr` alongside `attr.Client`, and `attr.ServerAddr` alongside `attr.Server`
- Applied the same change to `promUnresolvedHostGetters`

**`pkg/appolly/app/request/span_getters_unresolved_test.go`**:
- Updated expected values for `expectedClientAddr` and `expectedServerAddr` across all four test suites (OTEL ServerSide, OTEL ClientSide, Prom ServerSide, Prom ClientSide) to reflect that addr attributes are now properly renamed when unresolved-host renaming is enabled

## Behavior

- When a `ClientAddr` or `ServerAddr` value is a parseable IP address, it is now replaced with the configured unresolved name (e.g., `incoming`, `outgoing`, `unknown`)
- Values that are not pure IPs (e.g., `hostname:port`) are left unchanged, since `net.ParseIP` correctly rejects them
- No impact on rename-disabled mode (all values pass through unchanged)

## Validation

- All `TestRenameUnresolved_*` tests pass locally


mariomac:
Thank you for the PR @NameHaibinZhang . I wonder that, instead of changing the behavior of `client.address` or `server.address` to something outside of the OTEL semantic convention, we should just remove the attributes or require some explicit flag to let the users being aware of the consequences of enabling these attributes. Some extra discussion here: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2035#issuecomment-4508534527

rafaelroquetto (approved):
LGTM!

rafaelroquetto:
Hey @NameHaibinZhang I wanted to merge this today but I noticed that @mariomac left a few comments. I am going to wait for the discussion to be resolved before merging this one.

For any maintainers reading this: **do not merge**

NameHaibinZhang:
Thanks @mariomac for the thoughtful feedback — the semconv compliance point is well taken.

To clarify what this PR does today: OBI already rewrites `client` / `server` (the OBI-specific name attributes) to the configured `unresolved-host` placeholder when the value is a raw IP. This PR only extends that *existing* behavior to the addr-shaped counterparts (`client.address` / `server.address`), because users who opt into `unresolved-host` today reasonably expect it to apply consistently — that inconsistency is the root cause of #2035.

That said, I agree that strictly speaking, `client.address` / `server.address` per semconv should carry the original IP/domain value. Rewriting them to a placeholder is pragmatic but technically non-compliant.

I'd lean toward **option B**: when `unresolved-host` renaming is enabled and the value would be a raw IP, simply **omit `client.address` / `server.address`** from the exported span rather than masking them with a placeholder. This:
- Keeps semconv compliance (no non-standard values in spec-defined attributes)
- Still solves the cardinality explosion from #2035 (attribute not exported = zero cardinality cost)
- Requires minimal code change in the same getter functions

For completeness, the other paths I see:
- **(A)** Keep this PR as-is — pragmatic consistency with the existing `client`/`server` renaming, accepting the semconv deviation under the explicit `unresolved-host` opt-in.
- **(C)** Close this PR and pursue the broader footgun-flag design (`OTEL_EBPF_ALLOW_FOOTGUNS` or similar) as a separate effort, leaving #2035 open until that lands.

I'm happy to rework this PR toward (B) if that aligns with your preference. Would appreciate confirmation from you and @rafaelroquetto before pushing further changes. Thanks!

mariomac:
Thank you a lot for your work and explanations @NameHaibinZhang . During May 27th SIG meeting, we discussed together this issue and we decided to close it as "not an issue".

`client.address` and `server.address` attributes are optional attributes that are not enabled by default, so they are enabled by users at their own responsibility.

Modify their behavior outside of the semantic convention could break some observability pipelines (for example, an OTEL collector requiring this information to apply redirections/transformations).

I'm sorry because you already provided a fix, but thank you again for your valuable contributions.

mariomac:
On June 3rd SIG meeting, we decided that we should address this issue by un-defaulting `server.address` and replacing it by `server`, which has the wanted behavior this PR looked for `server.address`.

https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2248

mariomac:
I see that attr.Server is not anymore in the semantic conventions and metrics explicitly set attr.ServerAddr as recommended attributes.

So even if we don't fulfill 100% the semantic conventions here, I think this is the way to go to avoid unmanageable cardinality.

mariomac (approved):

grcevski (approved):
LGTM!

mariomac:
Thanks @NameHaibinZhang

PR #2157, opened 2026-05-21
Add TCP IO metric (bytes transferred at socket layer)

pinoOgni:
## Summary

This PR adds `obi_stat_tcp_io` counter tracking TCP bytes sent and received at the socket layer, differentiated by a `network.io.direction` label (transmit/receive), following the [semconv issue proposal](https://github.com/open-telemetry/semantic-conventions/issues/3682).

- The probes that calculate the tcp io metric, unlike other stat metrics that fire once per connection event, fire on every send/recv call. Volume and overhead caveat is documented.
- In the bpf side, bytes are batched in a per-socket hash map (up to 10 per direction) before being emitted to the ring buffer, significantly reducing event volume versus emitting on every call.
- On kernels older than ~6.5, traffic sent via `sendfile()` is not captured because it went through `tcp_sendpage` rather than `tcp_sendmsg`. 
- Renames all existing StatsO11y BPF programs to the `obi_stats_{probe_type}_{kernel_func}[_{purpose}]` convention, and documents the convention in `devdocs/metrics.md`.

## Validation


bpf/statsolly/k_tcp.c
```diff
@@ -29,8 +29,18 @@ typedef struct tcp_rtt {
     connection_info_t conn;
 } tcp_rtt_t;
 
-// Force tcp_rtt_t
+typedef struct tcp_io {
+    u8 flags; // Must be first, we use it to tell what kind of event we have on the ring buffer
+    u8 direction;
+    u8 _pad[6];
+    u64 bytes;
```
mmat11: I think this can be u32, then we can shrink the padding

bpf/statsolly/k_tcp.c
```diff
@@ -73,3 +83,60 @@ int BPF_KPROBE(obi_kprobe_tcp_close_srtt, struct sock *sk) {
 
     return 0;
 }
+
+SEC("kprobe/tcp_sendmsg")
+int BPF_KPROBE(obi_kprobe_tcp_sendmsg_bytes_transmit,
```
mmat11:
```suggestion
int BPF_KPROBE(obi_kprobe_tcp_sendmsg,
```
unless we have multiple kprobes on this hook, the convention is to prepend obi_kprobe_ to the hook name

bpf/statsolly/k_tcp.c
```diff
@@ -29,8 +29,18 @@ typedef struct tcp_rtt {
     connection_info_t conn;
 } tcp_rtt_t;
 
-// Force tcp_rtt_t
+typedef struct tcp_io {
+    u8 flags; // Must be first, we use it to tell what kind of event we have on the ring buffer
+    u8 direction;
```
mmat11: shouldn't this be `enum network_io_direction`?

bpf/statsolly/k_tcp.c
```diff
@@ -73,3 +83,60 @@ int BPF_KPROBE(obi_kprobe_tcp_close_srtt, struct sock *sk) {
...
+
+    tcp_io_t *se = bpf_ringbuf_reserve(&stats_events, sizeof(*se), 0);
+    if (!se) {
+        return 0;
+    }
+
+    se->flags = k_event_stat_tcp_io;
+    se->direction = direction_transmit;
+    // size gives us the bytes submitted to the send path by the application
+    // which may overcount in cases of failure of tcp_sendmsg
```
mmat11: I'm not familiar with these kind of metrics and use cases. Would it be more useful to report all the bytes that go through tcp_sendmsg or just the succesful ones? In case it's the latter, we should probably go through the kretprobe and check for errors first

pkg/export/feature.go
```diff
@@ -38,7 +39,10 @@ const (
 )
 
 // FeatureStats enables all stat metrics.
-const FeatureStats = FeatureStatsTCPRtt | FeatureStatsTCPFailedConnections | FeatureStatsTCPRetransmits
+// Note: FeatureStatsTCPIo fires on every tcp_sendmsg and tcp_cleanup_rbuf call — much
```
mmat11: I was thinking if we could aggregate or sample these events before they even get to userspace, for example: we see 10 tcp_sendmsg, the first 9 we just append the size value to an array, the tenth time we fire the event; we cleanup and fire a final event in the case of a closed connection too

pkg/internal/statsolly/ebpf/stat.go
```diff
@@ -59,6 +60,22 @@ const (
 	CodeRoleServer  NetworkTCPHandshakeRoleCode = 2
 )
 
+type NetworkIoDirectionType string
+
+const (
+	DirectionUnknown  NetworkIoDirectionType = "unknown"
```
mmat11: I think direction unknown doesn't exist

bpf/statsolly/k_tcp.c
pinoOgni: Yes, `obi_kprobe_tcp_cleanup_rbuf`/`obi_kprobe_tcp_sendmsg` are already used in `k_tracer.c`. I can follow (more or less) the convention used in statsolly and name it `obi_kprobe_tcp_sendmsg_io`/`obi_kprobe_tcp_sendmsg_tcp_io` like `obi_kprobe_tcp_close_srtt`/`obi_tp_inet_sock_set_state_tcp_failed_conn`. WDYT?

bpf/statsolly/k_tcp.c
pinoOgni: Before implementing the solution I searched around a bit and finally I chose to report all the bytes that go through `tcp_sendmsg`.

If we want to do the latter, I think we need a kprobe that saves the sock pointer in a map, and the kretprobe reads the actual return value from `tcp_sendmsg` which are the bytes accepted by the TCP stack, or negative on error. 

Actually, this should be the most correct solution, because we only count bytes that made it into the kernel send path.

pkg/export/feature.go
pinoOgni: I have tho think about it. Not sure if it has to be done in this PR or in a future one.

bpf/statsolly/k_tcp.c
mmat11: imho it would make more sense to namespace these names to the subsystem, eg. `obi_stats_kprobe_tcp_sendmsg` and implicitly assume that if the namespace is missing, it means `generic`, not sure though

pkg/export/feature.go
mmat11: if it's something problematic in the sense that it could keep the ringbuffer busy for other events, I think it should be addressed in this PR

bpf/statsolly/k_tcp.c
pinoOgni: Fine by me, I can rename all probes with `obi_stats_...` in this PR.

bpf/statsolly/k_tcp.c
```diff
@@ -29,8 +29,24 @@ typedef struct tcp_rtt {
     connection_info_t conn;
 } tcp_rtt_t;
 
-// Force tcp_rtt_t
+typedef struct tcp_io {
+    u8 flags; // Must be first, we use it to tell what kind of event we have on the ring buffer
+    u8 _pad[3];
+    enum network_io_direction direction;
```
rafaelroquetto, line 37: we should change all of the enums in the types.h header to have a `u8` underlying type - then you can drop the padding here to `_pad[2]`  - same for other analogous sites

bpf/statsolly/k_tcp.c
```diff
@@ -29,8 +29,24 @@ typedef struct tcp_rtt {
...
+    u32 bytes;
+    connection_info_t conn;
+} tcp_io_t;
+
+// Force structs into the ELF for automatic creation of Golang struct
 const tcp_rtt_t *unused_tcp_rtt __attribute__((unused));
+const tcp_io_t *unused_tcp_io __attribute__((unused));
+
+struct {
```
rafaelroquetto: should this go on its own header in `maps/`?

rafaelroquetto (commented):
Minor nits

bpf/statsolly/k_tcp.c
pinoOgni: > should this go on its own header in `maps/`?

Yep, thanks!

bpf/statsolly/maps/stats_events.h
```diff
@@ -10,7 +10,7 @@
 
 struct {
     __uint(type, BPF_MAP_TYPE_RINGBUF);
-    __uint(max_entries, 1 << 16);
+    __uint(max_entries, 1 << 20);
```
rafaelroquetto, line 13: With this being increased to 1MB, do you think it makes sense to change the threshold in which we wake up the ring buffer below? I wonder if we should set it to half of the size of the ring buffer, assuming that the userspace is consuming this swifitly. May be out of scope for this PR as it may need some profiling to compare different scenarios

rafaelroquetto (approved):

rafaelroquetto (approved):
LGTM

bpf/statsolly/maps/stats_events.h
pinoOgni: Yes, nice catch! I will run some tests.

bpf/statsolly/maps/tcp_io_accum.h
```diff
@@ -0,0 +1,27 @@
...
+
+typedef struct tcp_io_accum {
+    u32 bytes[k_tcp_io_batch_size];
+    u8 count;
+    u8 _pad[3];
+} tcp_io_accum_t;
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __uint(max_entries, 20480);
```
mmat11: question: why exactly 20480?

bpf/statsolly/k_tcp.c
```diff
@@ -23,18 +25,95 @@ enum {
...
+    }
+    connection_info_t conn;
+    if (!parse_sock_info(sk, &conn)) {
+        bpf_ringbuf_discard(se, 0);
+        return;
+    }
+    se->flags = k_event_stat_tcp_io;
+    se->direction = direction;
+    se->count = accum->count;
+    __builtin_memcpy(se->bytes, accum->bytes, sizeof(se->bytes));
```
mmat11: use `bpf_memcpy`

bpf/statsolly/maps/tcp_sendmsg_sock.h
```diff
@@ -0,0 +1,13 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//go:build obi_bpf_ignore
+#pragma once
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_helpers.h>
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __uint(max_entries, 10240);
+    __type(key, u64);
```
mmat11:
```suggestion
    __type(key, u64); // pid_tgid
```

bpf/statsolly/maps/tcp_sendmsg_sock.h
```diff
@@ -0,0 +1,13 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+//go:build obi_bpf_ignore
+#pragma once
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_helpers.h>
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
```
mmat11: should we use an LRU here to avoid random leaks?

bpf/statsolly/k_tcp.c
```diff
@@ -73,3 +152,45 @@ int BPF_KPROBE(obi_kprobe_tcp_close_srtt, struct sock *sk) {
...
+    const u64 pid_tgid = bpf_get_current_pid_tgid();
+    struct sock *const *skp = bpf_map_lookup_elem(&tcp_sendmsg_sock, &pid_tgid);
+
+    if (!skp) {
+        return 0;
+    }
+    struct sock *const sk = *skp;
+    bpf_map_delete_elem(&tcp_sendmsg_sock, &pid_tgid);
+
+    const long sent = PT_REGS_RC(ctx);
```
mmat11: is this the return value? if yes, you can add it to the function signature and you don't need to read registers manually

bpf/statsolly/maps/tcp_io_accum.h
pinoOgni: I was doing some tests and I left it (5 * 4096), I can write `1 << 14` or `1 << 15`. WDYT?

bpf/statsolly/maps/tcp_sendmsg_sock.h
pinoOgni: Do you think that we should also use `BPF_MAP_TYPE_LRU_HASH` for `tcp_io_accum` map?

mmat11 (approved):

Issue #2119, opened 2026-05-14
gRPC server-side context propagation fails for non-Go services due to tcp_recvmsg/sock_recvmsg kretprobe race condition


NameHaibinZhang:
## What happened?

When a Go gRPC client calls a non-Go gRPC server (e.g., .NET), spans generated on the server side are not correlated with the client-side spans. The server produces a brand-new `trace_id` instead of extracting the one from the received `traceparent` header. This effectively breaks distributed tracing for any non-Go gRPC service instrumented by OBI.

tcpdump confirms the Go client successfully injects `traceparent` (66-byte HPACK-encoded header visible in HTTP/2 HEADERS frames). The issue lies in the server-side receive path within the generictracer.

## What did you expect to happen?

The server-side span should inherit the `trace_id` and `parent_span_id` from the `traceparent` header sent by the client, resulting in a properly correlated distributed trace.

mmat11:
Hi, was the gRPC connection already established before OBI startup? If yes, this is a known limitation, see: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/grpc-context-propagation.md#persistent-connections-established-before-obi

NameHaibinZhang:
> Hi, was the gRPC connection already established before OBI startup? If yes, this is a known limitation, see: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/grpc-context-propagation.md#persistent-connections-established-before-obi

@mmat11  During my testing, I first installed OBI, then restarted both gRPC services.I found an issue：

This happens because the OBI eBPF
 agent's sk_msg program (obi_packet_extender) bails early for Go gRPC client connections without scheduling TCP options — the only cross-language, cross-node context propagation mechanism.

 The Go uprobe already injects traceparent via HPACK into user-space buffers and stores trace context in outgoing_trace_map (keyed with per-stream stream_id). But the sk_msg program:
 1. Returns early when it detects is_go_grpc_client_conn, skipping the H2 detection chain entirely
 2. Even if it didn't bail, the initial egress key uses stream_id=0, which won't match the per-stream entries stored by the Go uprobe

NameHaibinZhang:
this test on the main branch

NimrodAvni78:
@NameHaibinZhang @mmat11 can we close this?
Is the devdoc update enough to close this or is there a follow up?

mmat11:
[...]

PR #2165, opened 2026-05-25
fix(2119): enable TCP options propagation for Go gRPC client connections

NameHaibinZhang:
#2119 

The sk_msg program (obi_packet_extender) was returning early for Go gRPC client connections without entering the H2 detection chain. This meant TCP options — the cross-language context propagation mechanism — were never scheduled, breaking distributed tracing when a Go gRPC client calls a non-Go server.

Two changes:
1. After filling msg buffers for Go gRPC client connections, tail-call into the H2 frame scanner instead of returning immediately.
2. In the H2 detector, when a Go uprobe trace parent is found (valid and written), schedule TCP options so the trace context reaches the remote non-Go service via the kernel TCP option path.

The H2 scanner also resolves the stream_id from the frame header, fixing the mismatch where the initial egress key used stream_id=0 which could not match per-stream entries stored by the Go uprobe.

## Summary

Describe the change briefly.

## Validation

### DIFF
--- a/bpf/tpinjector/tpinjector.c
+++ b/bpf/tpinjector/tpinjector.c
@@ -910,6 +910,7 @@ int obi_packet_extender(struct sk_msg_md *msg) {
     if (is_go_grpc_client_conn(&t_ctx->p_conn)) {
         bpf_msg_pull_data(msg, 0, msg->size, 0);
         fill_msg_buffers(msg, &t_ctx->p_conn, &e_key);
+        bpf_tail_call_static(msg, &extender_jump_table, k_tail_detect_h2);
         return SK_PASS;
     }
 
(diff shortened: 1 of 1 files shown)


mmat11:
gRPC context propagation only works via headers injection, see https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1832#pullrequestreview-4150849428

I realized I've left some stale documentation on TCP options, opened https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2166 to fix it

rafaelroquetto:
Hey thanks for this but this doesn't look right. The early return for GRPC connections is intentional - there's a comment at the top of tpinjector that explains the flow and why.

In particular, we can't inject tcp options for grpc connections as they are semantically incompatible - a single grpc connection packet may contain multiple streams, which cannot be modelled with the use of tcp options.

I haven't looked deeply but the bug you are seeing is probably meant to be addressed at the generic tracer level on the receiving end, not here.

Closing the PR as it actively introduces a regression.

PR #2166, opened 2026-05-25
devdocs: update gRPC context propagation docs to explicitly state TCP options are not a supported mechanism

mmat11:
## Summary

update gRPC context propagation docs to explicitly state TCP options are not a supported mechanism

/cc @NameHaibinZhang 

## Validation

### DIFF
--- a/bpf/tpinjector/tpinjector.c
+++ b/bpf/tpinjector/tpinjector.c
@@ -847,9 +847,8 @@ static __always_inline void wrap_http2_traceparent(struct sk_msg_md *msg,
     }
 }
 
-// HTTP/1 + TCP option only. Returns false so caller can try H2 — H2 needs
-// per-stream keys, not the connection-scoped tp_pid. Helper owns clearing
-// the tp_pid in every exit path
+// HTTP/1 only. Caller must skip this for H2 sockets — connection-scoped tp_pid
+// carries the wrong context for multiplexed streams.
 static __always_inline bool handle_existing_tp_pid(struct sk_msg_md *msg,
                                                    u64 id,
                                                    const pid_connection_info_t *p_conn,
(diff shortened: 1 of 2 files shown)


mariomac (approved):
🚀

bpf/tpinjector/tpinjector.c
```diff
@@ -841,13 +841,16 @@ static __always_inline void wrap_http2_traceparent(struct sk_msg_md *msg,
...
+// socket NOT being H2: gRPC/H2 multiplexes streams per connection so the
+// connection-scoped tp_pid would carry the wrong context for all but
+// one stream — HPACK injection is the only correct CP path for H2.
 static __always_inline bool handle_existing_tp_pid(struct sk_msg_md *msg,
                                                    u64 id,
                                                    const pid_connection_info_t *p_conn,
                                                    const egress_key_t *e_key,
                                                    tp_info_pid_t *tp_pid) {
-    if (inject_flags & k_inject_tcp_options) {
+    if ((inject_flags & k_inject_tcp_options) && !is_h2_socket(msg)) {
```
rafaelroquetto: is this really required? if so, we must update the comment above, but otherwise I'd prefer if we left this out, for the following reason: `get_tp_info_pid` should not hit for http2 sockets - if it does, for whatever reason, then it could either be legitimate or not, in which case we should verify at the source - this is just a bandaid that lets it fail silently

NameHaibinZhang:
Thanks for the documentation clarification, @mmat11 . This helps prevent future confusion about the propagation mechanisms available for gRPC.I agree that TCP options are not suitable for gRPC due to the multiplexed stream nature of HTTP/2 — a single TCP connection carries multiple concurrent streams, and TCP options cannot disambiguate per-stream trace context.Regarding @rafaelroquetto's comment on the !is_h2_socket(msg) guard in tpinjector.c: I agree this is worth discussing. If the intent is purely a documentation PR, perhaps the code change should be split out or dropped to keep the scope clean. If it stays, the comment above the condition should be updated to explain why H2 sockets are excluded (i.e., TCP options cannot model multiplexed streams).For issue #2119 (Go gRPC client → non-Go server context propagation), I'll look into addressing it at the generic tracer receiving side as suggested by @rafaelroquetto .

bpf/tpinjector/tpinjector.c
mmat11: > get_tp_info_pid should not hit for http2 sockets

it does, at the start of packet extender; I can move the check there

rafaelroquetto: Why does it tho? are we populating outgoing_trace_map for http2 sockets? I was under the impression we should not be

mmat11: I remember we discussed this somewhere in the grpc PR, can't find the discussion though; the map is stream_id indexed, why is that an issue?

rafaelroquetto: maybe it was in the case of when a connection is upgraded from http1 to http2? to be honest I don't quite recall it either

rafaelroquetto (approved):
LGTM

PR #2172, opened 2026-05-26
Update to LLVM22

rafaelroquetto:
## Summary

Update the generator image to LLVM22.

LLVM22 generates code that trips 5.10. I traced this to a few `bpf_dbg_printk` inside loops, which I removed: bpf_dbg_printk inside loops are a bad idea, because they themselves carry branching, pointer tracking and ringbuffer flushes - none of which is desirable when a loop is being unrolled.

I'd avoid this pattern in the general case anyway, as it's not LLVM22 exclusive.

## Validation

### DIFF
--- a/generator.Dockerfile
+++ b/generator.Dockerfile
@@ -9,7 +9,10 @@ ENV PROTOC_AARCH_64_SHA256="56af3fc2e43a0230802e6fadb621d890ba506c5c17a1ae1070f6
 
 ARG TARGETARCH
 
-RUN apk add clang llvm20 wget unzip curl make bash git
+RUN echo "https://dl-cdn.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositories
+RUN echo "https://dl-cdn.alpinelinux.org/alpine/edge/community" >> /etc/apk/repositories
+
+RUN apk add clang22 llvm22 wget unzip curl make bash git
 RUN apk cache purge
 
 COPY internal/tools/generator/ internal/tools/generator/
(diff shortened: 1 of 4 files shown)


generator.Dockerfile
```diff
@@ -39,7 +42,7 @@ RUN --mount=type=cache,target=/go/pkg \
 
 RUN cat <<EOF > /generate.sh
 #!/bin/sh
-export PATH="/usr/lib/llvm20/bin:\$PATH"
+export PATH="/usr/lib/llvm22/bin:\$PATH"
 export BPF2GO=/go/bin/bpf2go
 export BPF_CLANG=clang
```
mmat11: I checked the generator image we use in CI locally and found out it uses the clang version this points to:

```
docker run --rm --entrypoint clang ghcr.io/open-telemetry/obi-generator:0.2.13 --version                                                                                                                                                                                     

Alpine clang version 21.1.2
Target: aarch64-alpine-linux-musl
Thread model: posix
InstalledDir: /usr/lib/llvm21/bin
Configuration file: /etc/clang21/aarch64-alpine-linux-musl.cfg
```

this is also what this branch CI used.

I _think_ we should hardcode the clang version we want here in the Dockerfile, publish a new image and rerun the CI with the updated workflow

rafaelroquetto:
@mmat11 pushed a new updated image version, updated the Makefile and retriggered CI with the latest push.

mmat11:
I see a verifier error in 5.10 -> https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/26545128365/job/78196812910?pr=2172

rafaelroquetto:
@mmat11 good catch - yeah I think we'll need to split something inside gotracer... I will get back to this PR later

generator.Dockerfile
rafaelroquetto: Done

mmat11 (approved):
lgtm!

PR #2179, opened 2026-05-27
Add trace pipe only BPF debug mode

### LINKED ISSUE #1295: Enhance OTEL_EBPF_BPF_DEBUG to allow printing to trace_pipe only
This is more like a request for comments/feedback: often times I find it useful to just print bpf debug information to `trace_pipe` rather than to the userspace stdout:
1. it keeps the application output clean
2. likewise, it keeps the eBPF logs separated from userspace junk
3. there are no character limits

My proposal is to change the semantics of `OTEL_EBPF_BPF_DEBUG` so that:
1. `default == true == 1 (fallback)`: current behaviour of printing to both `trace_pipe` and submitting the log messages to the userspace ring buffer
2. `trace_pipe`: only prints to trace pipe (skips the ring buffer)

In the `trace_pipe` case (and perhaps in every case, whatever is simpler), we should also aim to print the current function name or context, like we do for  the userspace messages:

```
bpf_printk("%s: "fmt, __FUNCTION__, ##args);
```

caveat: limits the printk arguments back to 2 rather than 3 in certain kernels.

ethical-buddy:
## Summary

Closes #1295.

Adds `bpf_debug_mode` so eBPF debug output can be routed to `trace_pipe`, userspace debug events, or both. The existing `bpf_debug` boolean remains supported as the deprecated compatibility option.

The BPF side now uses a single `g_bpf_debug_flags` value, matching the flag-style approach requested in review.

## Validation

Local validation:

- `go test ./pkg/config ./pkg/obi ./pkg/internal/statsolly/ebpf ./pkg/internal/ebpf/logger`
- `make docker-generate`
- `make generate-config-schema`
- `make check-config-schema`
- `make lint-markdown`
- `make lint`
- `make compile`
- `git diff upstream/main...HEAD --check`

The privileged verifier test was attempted locally, but this environment cannot run it without sudo password access.


devdocs/config/config-schema.json
```diff
@@ -109,6 +109,26 @@
...
+          "type": "boolean",
+          "description": "false disables eBPF debug logging; true enables trace_pipe and userspace ring buffer debug logging"
+        },
+        {
+          "type": "string",
+          "enum": [
+            "default",
+            "false",
+            "trace_pipe",
+            "true"
```
mariomac: Mixing boolean with strings sounds a bit confusing to me. Maybe we can replace this by an additive list of enums, like:
```yaml
bpf_debug_mode: ["trace_pipe", "ringbuffer"]
```
empty debug mode would mean "disabled".

Then we mark `bpf_debug` as deprecated but we keep supporting it for backwards-compatibility. Its value would be set as:

* `true` if `bpf_debug: true` (explicitly set) or `bpf_debug_mode` is not empty
* `false` if `bpf_debug: false` (explicitly set) of `bpf_debug_mode` is empty.

mariomac (commented):
Thank you for your contribution @ethical-buddy ! I have one comment w.r.t the configuration

rafaelroquetto (commented):
hey @ethical-buddy piggy-backing into @mariomac suggestion, I believe a more appropriate way to implement it would be doing something analogous to `inject_flags` used by `tpinjector` - so instead of having individual bools, we have a single `flags` variable we can test - this avoid fragmentation and hardens the concept.

bpf/common/globals.h
```diff
@@ -5,7 +5,14 @@
 
 #include <bpfcore/vmlinux.h>
 
-volatile const bool g_bpf_debug = false;
+enum bpf_debug_flags {
+    BPF_DEBUG_TRACE_PIPE = 1 << 0,
+    BPF_DEBUG_RINGBUF = 1 << 1,
```
rafaelroquetto: nit: this is our enum style

```suggestion
    k_bpf_debug_trace_pipe = 1 << 0,
    k_bpf_debug_ringbuf = 1 << 1,
```

bpf/common/globals.h
```diff
@@ -5,7 +5,14 @@
...
+};
+
+volatile const u32 g_bpf_debug_flags = 0;
 volatile const bool g_bpf_traceparent_enabled = false;
 volatile const bool g_bpf_header_propagation = false;
-volatile const bool g_bpf_loop_enabled = false;
\ No newline at end of file
+volatile const bool g_bpf_loop_enabled = false;
+
+#define g_bpf_debug (g_bpf_debug_flags != 0)
```
rafaelroquetto: lets not use macros like this, if you must, define a static __always inline bool wrapper

bpf/logger/bpf_dbg.h
```diff
@@ -66,8 +71,8 @@ enum bpf_func_id___x {
         bpf_ringbuf_submit(__trace__, 0);                                                          \
     } while (0)
 
-#define bpf_dbg_enter() bpf_dbg_printk("%s entered", __FUNCTION__)
-#define bpf_dbg_return() bpf_dbg_printk("%s returning", __FUNCTION__)
+#define bpf_dbg_enter() bpf_dbg_printk("entered")
+#define bpf_dbg_return() bpf_dbg_printk("returning")
```
rafaelroquetto: this removed the printing of the function name, I believe this is incorrect

devdocs/config/config-schema.json
```diff
@@ -109,6 +109,38 @@
...
+        }
+      ],
+      "title": "BPF Debug Mode",
+      "description": "Controls eBPF debug logging. Deprecated: use bpf_debug_mode instead."
+    },
+    "BPFDebugModeList": {
+      "items": {
+        "type": "string",
+        "enum": [
+          "ringbuffer",
```
rafaelroquetto: `userspace` is a better term than ring_buffer, which is just implementation detail

ethical-buddy:
Updated based on the review feedback: added bpf_debug_mode as an additive list while keeping bpf_debug deprecated-compatible, moved the BPF side to a single debug flags constant, and kept explicit trace_pipe-only verifier coverage. I also removed the extra printk function-name argument so older kernels stay within verifier limits.

pkg/config/ebpf_tracer.go
```diff
@@ -20,7 +20,23 @@ type RedisDBCacheConfig struct {
...
+type (
+	BPFDebugMode     uint8
+	BPFDebugOutput   string
+	BPFDebugModeList []BPFDebugOutput
+)
+
 const (
+	BPFDebugDisabled BPFDebugMode = iota
+	BPFDebugDefault
+	BPFDebugTracePipe
```
rafaelroquetto: iota is the wrong approach here - these are flags, should another value be added, this will break as `3` should represent `default | trace_pipe`

pkg/config/ebpf_tracer.go
```diff
@@ -34,6 +50,135 @@ const (
...
 
+func (m BPFDebugMode) Enabled() bool {
+	return m != BPFDebugDisabled
+}
+
+func (m BPFDebugMode) RingbufEnabled() bool {
+	return m == BPFDebugDefault
+}
+
+func (m BPFDebugMode) Flags() uint32 {
```
rafaelroquetto: There is no need for such method if we treat BPFDebugMode as flags. BPFDebugMode itsels should be a flag type just like we do for `ContextPropagationMode`

rafaelroquetto (changes requested):
The type iself should have the semantics of a flag type. You should just mirror/stay consistent to what we've done with `ContextPropagationMode` - including the name of the string variables prefixed with `Str` and the absence of a list type

ethical-buddy:
Thanks for the detailed review, and sorry for the extra iteration here.\n\nI reworked this to follow the existing flag-style pattern more closely:\n- kept `bpf_debug` as the deprecated boolean compatibility option\n- made `bpf_debug_mode` use the single flag-style `BPFDebugMode`, similar to `ContextPropagationMode`\n- switched the user-facing output name from `ringbuffer` to `userspace`\n- kept the BPF side on one `g_bpf_debug_flags` constant\n- replaced the macro with an inline helper and updated the BPF enum naming style\n- restored function names for the enter/return debug helpers\n\nI also regenerated the config docs/schema and reran the focused validation locally.

bpf/logger/bpf_dbg.h
```diff
@@ -41,10 +41,15 @@ enum bpf_func_id___x {
 
 #define bpf_dbg_printk(fmt, args...)                                                               \
     do {                                                                                           \
-        if (!g_bpf_debug) {                                                                        \
+        if (!bpf_debug_enabled()) {                                                                \
```
rafaelroquetto: I think we can remove this block, since it's already covered by the subsequent ifs below

pkg/config/ebpf_tracer.go
```diff
@@ -32,8 +43,126 @@ const (
...
+		default:
+			return fmt.Errorf("invalid value for bpf_debug_mode: '%s' (valid: disabled, trace_pipe, userspace, all)", part)
+		}
+	}
+
+	*m = result
+	return nil
+}
+
+func (m *BPFDebugMode) UnmarshalYAML(value *yaml.Node) error {
```
rafaelroquetto: If we want to support the list syntax, then we need to extend it to context propagation mode. Otherwise, we should drop this.

pkg/config/ebpf_tracer.go
```diff
@@ -32,8 +43,126 @@ const (
...
+func (m BPFDebugMode) Flags() uint32 {
+	var flags uint32
+	if m&BPFDebugTracePipe != 0 {
+		flags |= BPFDebugFlagTracePipe
+	}
+	if m&BPFDebugUserspace != 0 {
+		flags |= BPFDebugFlagUserspace
+	}
+	return flags
+}
```
rafaelroquetto: This is requivalent to `return uint32(m)`.  There's no need to recompose the flags (and it's error prone to do so)

pkg/config/ebpf_tracer.go
```diff
@@ -20,7 +21,17 @@ type RedisDBCacheConfig struct {
...
+type BPFDebugMode uint8
+
 const (
+	BPFDebugDisabled  BPFDebugMode = 0
+	BPFDebugTracePipe BPFDebugMode = 1 << 0
+	BPFDebugUserspace BPFDebugMode = 1 << 1
+	BPFDebugDefault   BPFDebugMode = BPFDebugTracePipe | BPFDebugUserspace
+
+	BPFDebugFlagTracePipe uint32 = 1 << 0
+	BPFDebugFlagUserspace uint32 = 1 << 1
```
rafaelroquetto: These are not required, see my comment on `Flags()`

pkg/config/ebpf_tracer.go
```diff
@@ -32,8 +43,126 @@ const (
 	StrContextPropagationHeaders  = "headers"
 	StrContextPropagationHTTP     = "http"
 	StrContextPropagationTCP      = "tcp"
+	StrBPFDebugDisabled           = "disabled"
+	StrBPFDebugTracePipe          = "trace_pipe"
+	StrBPFDebugUserspace          = "userspace"
+	StrBPFDebugAll                = "all"
 )
 
+func (m BPFDebugMode) Enabled() bool {
```
```suggestion
func (m BPFDebugMode) IsEnabled() bool {
```
rationale: consistency

pkg/config/ebpf_tracer.go
```diff
@@ -32,8 +43,126 @@ const (
...
+	return m != BPFDebugDisabled
+}
+
+func (m BPFDebugMode) UserspaceEnabled() bool {
```
```suggestion
func (m BPFDebugMode) IsUserspaceEnabled() bool {
```

pkg/internal/ebpf/logger/logger.go
```diff
@@ -43,7 +43,7 @@ func New(cfg *obi.Config) *BPFLogger {
 }
 
 func (p *BPFLogger) LoadSpecs() ([]*ebpfcommon.SpecBundle, error) {
-	if p.cfg.EBPF.BpfDebug {
+	if p.cfg.EBPF.BpfDebugMode.UserspaceEnabled() {
```
rafaelroquetto: We need to update the error message to something like `BPF userspace debug is not enabled`

pkg/config/ebpf_tracer_test.go
```diff
@@ -9,6 +9,39 @@ import (
 	"testing"
 )
 
+func TestBPFDebugMode_UnmarshalText(t *testing.T) {
```
rafaelroquetto, line 11: We need to add tests for the other accessors as well - see the `ContextPropagationMode` tests

pkg/config/ebpf_tracer.go
```diff
@@ -32,8 +43,126 @@ const (
...
+		flags |= BPFDebugFlagTracePipe
+	}
+	if m&BPFDebugUserspace != 0 {
+		flags |= BPFDebugFlagUserspace
+	}
+	return flags
+}
+
+func (m *BPFDebugMode) UnmarshalText(text []byte) error {
+	str := strings.ToLower(strings.TrimSpace(string(text)))
```
rafaelroquetto: We don't do ToLower (again, inconsistent with ContextPropagationMode).

Rationale: makes semantics/validation harder.

pkg/obi/config.go
```diff
@@ -818,6 +819,12 @@ func registerCustomValidations(validate *validator.Validate, customValidations C
 
 // normalizeConfig normalizes user input to a common set of assumptions that are global to OBI
 func (c *Config) normalize() {
+	if c.EBPF.BpfDebugMode.Enabled() {
+		c.EBPF.BpfDebug = true
+	} else if c.EBPF.BpfDebug {
+		c.EBPF.BpfDebugMode = config.BPFDebugDefault
+	}
```
rafaelroquetto: Instead of doing this, we should reject if both settings are set in Validate(). We should never backfill BpfDebug from BpfDebugMode

bpf/logger/bpf_dbg.h
```diff
@@ -71,7 +76,7 @@ enum bpf_func_id___x {
 
 #define bpf_d_printk(fmt, args...)                                                                 \
     do {                                                                                           \
-        if (!g_bpf_debug) {                                                                        \
+        if (!bpf_debug_enabled()) {                                                                \
```
rafaelroquetto: This should be gated on k_bpf_debug_trace_pipe.

rafaelroquetto (changes requested):

ethical-buddy:
Thanks for the detailed review. I updated the branch to follow the suggested flag-based approach more closely, removed the YAML list handling for bpf_debug_mode, kept the deprecated bpf_debug behavior separate, and added validation/tests for the new accessors and config cases.

I also updated the branch with the latest upstream main and reran the targeted validation locally:
- go test ./pkg/config ./pkg/obi ./pkg/internal/statsolly/ebpf
- make docker-generate
- make compile
- make check-config-schema
- make lint-markdown
- git diff --check

Please let me know if there is anything else I should adjust.

rafaelroquetto:
hey @ethical-buddy looks like some tests are still failing, do you plan to continue working on this PR?

ethical-buddy:
> hey @ethical-buddy looks like some tests are still failing, do you plan to continue working on this PR?

yes yes , I just need some time to get back , I will be fixing it soon .

ethical-buddy:
Thanks for the patience here. I updated the branch with the latest `main` and fixed the failures from the previous CI run that I could reproduce locally: the static lint issues, the BPF debug helper clang-tidy issue, and the verifier constants after the upstream merge.

I also regenerated the BPF artifacts and config docs/schema, and refreshed the PR description to match the current flag-based `bpf_debug_mode` implementation.

Local validation completed:

- `go test ./pkg/config ./pkg/obi ./pkg/internal/statsolly/ebpf ./pkg/internal/ebpf/logger`
- `make docker-generate`
- `make generate-config-schema`
- `make check-config-schema`
- `make lint-markdown`
- `make lint`
- `make compile`
- `git diff upstream/main...HEAD --check`

I tried to run the privileged verifier test locally as well, but this environment requires a sudo password, so I’ll watch the CI verifier result on the updated branch.

bpf/common/connection_info.h
```diff
@@ -72,7 +72,7 @@ typedef struct connection_info_part {
 } connection_info_part_t;
 
 static __always_inline void dbg_print_http_connection_info(const connection_info_t *info) {
-    if (!g_bpf_debug) {
+    if (!bpf_debug_enabled()) {
```
mmat11: question: can we still check g_bpf_debug instead of another function? it should still resolve to `false` if no flags are added right?

bpf/common/globals.h
```diff
@@ -5,7 +5,16 @@
...
-volatile const bool g_bpf_header_propagation = false;
-volatile const bool g_bpf_loop_enabled = false;
\ No newline at end of file
+enum bpf_debug_flags {
+    k_bpf_debug_trace_pipe = 1 << 0,
+    k_bpf_debug_userspace = 1 << 1,
+};
+
+volatile const u32 g_bpf_debug_flags = 0;
+volatile const bool g_bpf_traceparent_enabled = (bool)0;
```
mmat11: why are this changed?

bpf/common/globals.h
```diff
@@ -5,7 +5,16 @@
...
-volatile const bool g_bpf_traceparent_enabled = false;
-volatile const bool g_bpf_header_propagation = false;
-volatile const bool g_bpf_loop_enabled = false;
\ No newline at end of file
+enum bpf_debug_flags {
+    k_bpf_debug_trace_pipe = 1 << 0,
+    k_bpf_debug_userspace = 1 << 1,
+};
+
+volatile const u32 g_bpf_debug_flags = 0;
```
mmat11: can we keep the name `g_bpf_debug` so we don't need to change the code everywhere else?

bpf/rdns/rdns_xdp.c
```diff
@@ -7,8 +7,6 @@
 #include <bpfcore/bpf_builtins.h>
 #include <bpfcore/bpf_endian.h>
 
-#include <logger/bpf_dbg.h>
```
mmat11, line 10: why is this code being deleted?

pkg/appolly/discover/watcher_proc.go
```diff
@@ -151,7 +151,7 @@ func (pa *pollAccounter) run(ctx context.Context) {
 		return
 	}
 
-	if pa.cfg.EBPF.BpfDebug {
+	if pa.cfg.EBPF.DebugMode().IsEnabled() {
```
mmat11: I think you need to check `BPFDebugUserspace` explicitly here, not just `IsEnabled()`

pkg/config/ebpf_tracer.go
```diff
@@ -20,7 +20,14 @@ type RedisDBCacheConfig struct {
 	MaxSize int  `yaml:"max_size" env:"OTEL_EBPF_BPF_REDIS_DB_CACHE_MAX_SIZE" validate:"gt=0"`
 }
 
+type BPFDebugMode uint8
+
 const (
+	BPFDebugDisabled  BPFDebugMode = 0
+	BPFDebugTracePipe BPFDebugMode = 1 << 0
+	BPFDebugUserspace BPFDebugMode = 1 << 1
+	BPFDebugDefault   BPFDebugMode = BPFDebugTracePipe | BPFDebugUserspace
```
mmat11: nit: I'd name this BPFDebugAll or similar, for clarity

pkg/config/ebpf_tracer.go
```diff
@@ -43,9 +143,12 @@ type MapsConfig struct {
 
 // EBPFTracer configuration for eBPF programs
 type EBPFTracer struct {
-	// Enables logging of eBPF program events
+	// Deprecated: Use bpf_debug_mode instead.
 	BpfDebug bool `yaml:"bpf_debug" env:"OTEL_EBPF_BPF_DEBUG" validate:"boolean"`
 
+	// Enables logging of eBPF program events to the selected outputs.
+	BpfDebugMode BPFDebugMode `yaml:"bpf_debug_mode" env:"OTEL_EBPF_BPF_DEBUG_MODE" validate:"-"`
```
mmat11, line 146: do we need `oneof` validation here?

edit: or drop the validate tag?

pkg/internal/ebpf/verifier/bpf_verifier_test.go
```diff
@@ -167,9 +167,12 @@ func TestBPFVerifierWithConstants(t *testing.T) {
 		t.Skipf("cannot remove memlock limit (insufficient privileges?): %v", err)
 	}
 
+	debugFlags := []any{uint32(0), uint32(1), uint32(3)}
```
mmat11: can we use the constants instead?

pkg/internal/ebpf/verifier/bpf_verifier_test.go
```diff
@@ -167,9 +167,12 @@ func TestBPFVerifierWithConstants(t *testing.T) {
...
+	tracePipeDebugFlags := []any{uint32(0), uint32(1)}
```
mmat11: what's the difference between debugFlags and tracePipeDebugFlags?

mmat11 (changes requested):

mmat11:
@ethical-buddy thanks for addressing most of the review comments -- I've resolved the ones that were fixed and left the remaining ones open. Once those are addressed/answered as well, we can merge this

ethical-buddy:
Thanks for the patience here. I updated the branch with latest `main` and addressed the remaining open review items I could verify:

- restored the rdns debug logging that had been removed by mistake
- removed the extra `BPFDebugMode.Flags()` helper and now pass the flag value directly where BPF constants are loaded
- kept `bpf_debug_mode` validation in `UnmarshalText` with `validate:"-"`, following the existing config pattern
- kept the debug mode accessor tests and cleaned them up after removing `Flags()`
- reduced the tpinjector verifier debug matrix to avoid spending time on duplicate userspace debug combinations, while keeping userspace debug coverage in the dedicated logger and smaller programs

Local validation completed:

- `make docker-generate`
- focused `go test` for the touched packages
- `make check-config-schema`
- `make fmt`
- `make lint`
- `make compile`
- `make lint-markdown`
- `git diff --check`

I also ran the tagged verifier test locally; it builds but skips here because this environment cannot raise the memlock limit, so CI is still the source of truth for the full kernel verifier run.

Please let me know if there is anything else I should adjust.

pkg/config/ebpf_tracer_test.go
mmat11: @ethical-buddy was this comment addressed?

pkg/config/ebpf_tracer.go
mmat11: @ethical-buddy this needs to be addressed

pkg/export/otel/metric/provider_test.go
```diff
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr/funcr"
+
```
mmat11, line 16: unnecessary newline

pkg/export/otel/metrics_runtime_histograms_test.go
```diff
@@ -14,6 +14,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
```
mmat11, line 17: unnecessary newline

pkg/internal/ebpf/verifier/bpf_verifier_test.go
```diff
@@ -167,9 +168,25 @@ func TestBPFVerifierWithConstants(t *testing.T) {
 		t.Skipf("cannot remove memlock limit (insufficient privileges?): %v", err)
 	}
 
+	debugFlags := []any{
+		uint32(config.BPFDebugDisabled),
+		uint32(config.BPFDebugTracePipe),
+		uint32(config.BPFDebugAll),
+	}
+	// Some programs only use trace_pipe debug output, so userspace ring buffer
```
mmat11: I see we load the constant everywhere with `uint32(p.cfg.EBPF.DebugMode())` so I believe we should still test all flags

pkg/internal/ebpf/verifier/bpf_verifier_test.go
```diff
@@ -224,38 +241,36 @@ func TestBPFVerifierWithConstants(t *testing.T) {
...
 	forEachCombination(t, "logenricher/Bpf", logenricherbpf.LoadBpf, []constOption{
-		{"g_bpf_debug", []any{true, false}},
+		{"g_bpf_debug", debugFlags},
 	})
 
 	// rdns xdp
-	forEachCombination(t, "rdns/xdp/Bpf", rdnsxdpbpf.LoadBpf, []constOption{
-		{"g_bpf_debug", []any{true, false}},
-	})
+	loadAndVerify(t, "rdns/xdp/Bpf", rdnsxdpbpf.LoadBpf)
```
mmat11: this needs to be tested with combinations as well

pkg/obi/config.go
```diff
@@ -558,6 +559,9 @@ func stringSliceToTextUnmarshalerHookFunc() mapstructure.DecodeHookFunc {
 		}
 
 		if slice, ok := data.([]any); ok {
+			if to == reflect.TypeOf(config.BPFDebugMode(0)) {
```
mmat11: this is a generic function, I don't think we should validate specific config fields in here

pkg/obi/config.go
```diff
@@ -572,6 +576,9 @@ func stringSliceToTextUnmarshalerHookFunc() mapstructure.DecodeHookFunc {
 
 		// Handle []string directly
 		if slice, ok := data.([]string); ok {
+			if to == reflect.TypeOf(config.BPFDebugMode(0)) {
```
mmat11: this is a generic function, I don't think we should validate specific config fields in here

mmat11 (changes requested):

pkg/config/ebpf_tracer_test.go
ethical-buddy: Yes, this is covered now by `TestBPFDebugMode_Accessors`. It checks disabled, trace_pipe, userspace, and all modes for both accessors.

pkg/export/otel/metrics_runtime_histograms_test.go
ethical-buddy: Removed this extra newline as well.

pkg/export/otel/metric/provider_test.go
ethical-buddy: Removed this extra newline.

bpf/rdns/rdns_xdp.c
ethical-buddy: This deletion was accidental. I restored the rdns debug include and the debug logging blocks in the current branch.

pkg/config/ebpf_tracer.go
ethical-buddy: Addressed. I dropped the numeric validation and kept `validate:"-"` here, so the accepted values are handled by `UnmarshalText` instead of a `oneof` tag.

pkg/internal/ebpf/verifier/bpf_verifier_test.go
ethical-buddy: Addressed. The verifier values now use the `config.BPFDebug*` constants instead of raw numeric values.

mmat11 (approved):

MrAlias:
@ethical-buddy can you update with `main` and address the failing CI?

pkg/internal/ebpf/verifier/bpf_verifier_test.go
```diff
@@ -167,9 +168,19 @@ func TestBPFVerifierWithConstants(t *testing.T) {
 		t.Skipf("cannot remove memlock limit (insufficient privileges?): %v", err)
 	}
 
+	debugFlags := []any{
+		uint32(config.BPFDebugDisabled),
+		uint32(config.BPFDebugTracePipe),
+		uint32(config.BPFDebugAll),
+	}
+	tracePipeDebugFlags := []any{
```
mmat11, line 176: IIRC these were removed earlier and are not reintroduced?

pkg/internal/ebpf/verifier/bpf_verifier_test.go
```diff
@@ -240,38 +251,38 @@ func TestBPFVerifierWithConstants(t *testing.T) {
...
 	// logenricher
 	forEachCombination(t, "logenricher/Bpf", logenricherbpf.LoadBpf, []constOption{
-		{"g_bpf_debug", []any{true, false}},
+		{"g_bpf_debug", debugFlags},
 	})
 
 	// rdns xdp
 	forEachCombination(t, "rdns/xdp/Bpf", rdnsxdpbpf.LoadBpf, []constOption{
-		{"g_bpf_debug", []any{true, false}},
+		{"g_bpf_debug", tracePipeDebugFlags},
```
mmat11, line 280:
```suggestion
		{"g_bpf_debug", debugFlags},
```

mmat11 (changes requested):
all verifier tests are failing

MrAlias:
@ethical-buddy can you take a look at the failing CI?

PR #2184, opened 2026-05-28
fix metrics exporter memory leak

### LINKED ISSUE #2138: Suspected memory leak with the otel.MetricsReporter
### Component

OBI (eBPF instrumentation)

### What happened?

OBI (running as a daemonset) in a semi-large k8s cluster (32 nodes) has a specific pod with usage of 180Gi of memory (while the rest use 2-12Gi).
There is an INFO log that is being written multiple time per ms saying the following:
```log
level=INFO msg="failed to upload metrics: context deadline exceeded"
```
And this WARN log being written every now and then (lower frequency compared to the INFO log):
```log
level=WARN msg="error flushing evicted metrics provider" component=otel.MetricsReporter service="{Name:Kube-rbac-proxy Namespace: Instance:REDACTED}" error="context deadline exceeded"
```

Metrics being sent to the upstream collector and no memory leaks ;)

### Steps to reproduce

Not sure.

### OBI configuration

```yaml
network:
  enable: true
  protocol: ["TCP"]
ebpf:
  enable: true
  context_propagation: all
  buffer_sizes:
    http: 8192
    mysql: 8192
    kafka: 8192
    postgres: 8192
    tcp: 8192
attributes:
  select:
    span:
      include:
        - "net.transport"
        - "net.sock.peer.addr"
        - "net.sock.peer.port"
        - "net.host.port"
otel_metrics_export:
  endpoint: "http://my-collector.svc.cluster:4318"
  protocol: http/protobuf
otel_traces_export:
  endpoint: "http://my-collector.svc.cluster:4318"
  protocol: http/protobuf
discovery:
  exclude_instrument:
    - exe_path: '{*ebpf-instrument*,*otelcol*}'
    - k8s_namespace: 'obi'
  instrument:
    - k8s_namespace: '*'
    - exe_path: '/bin/kubelet'
```

### Relevant logs

```shell

```

### Additional context

Memory usage graph of the last 2 days:
Image

### Version

0.9.0

### Kernel version

_No response_

### OS / Distribution

OpenShift 4.16

### Deployment

Kubernetes

NimrodAvni78:
## Summary

Fixes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2138

`PeriodicReader` spawns a goroutine that is never closed, TTL eviction also calls `ForceFlush` and not `Shutdown`,  which doesn't cancel the context and leaves the goroutine dangling

The underlying exporter is shared across multiple `MeterProviders`, we can't shut it down directly or else other calls will fail with an error of the exporter is already shut down. Instead we wrap the each of them in a `sharedExporter` which noops the shutdown, still allowing the context to cancel and the goroutine to exit, but not actually shutting down the shared exporter

the `MetricsReporter.Close` method still shuts down the underlying exporter

## Validation

### DIFF
--- a/pkg/export/otel/metrics.go
+++ b/pkg/export/otel/metrics.go
@@ -312,8 +312,8 @@ func newMetricsReporter(
 			}
 
 			go func() {
-				if err := v.provider.ForceFlush(ctx); err != nil {
-					llog.Warn("error flushing evicted metrics provider", "error", err)
+				if err := v.provider.Shutdown(ctx); err != nil {
+					llog.Warn("error shutting down evicted metrics provider", "error", err)
 				}
 			}()
 		}, mr.newMetricSet)
@@ -699,7 +699,7 @@ func (mr *MetricsReporter) newMetricsInstance(service *svc.Attrs) Metrics {
 
 	opts := []metric.Option{
 		metric.WithResource(resources),
-		metric.WithReader(metric.NewPeriodicReader(mr.exporter,
+		metric.WithReader(metric.NewPeriodicReader(sharedExporter{mr.exporter},
 			metric.WithInterval(mr.cfg.Interval))),
 	}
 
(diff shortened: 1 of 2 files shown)


grcevski (approved):
LGTM! Nice, I like it.

mmat11 (approved):

PR #2187, opened 2026-05-28
ci: use crun instead of runc

### LINKED ISSUE #872: Find a way to use latest runc in VM workflows
Since [CVE-2025-52565](https://advisories.gitlab.com/pkg/golang/github.com/opencontainers/runc/CVE-2025-52565/), the OBI VM workflows started failing with `unsafe procfs detected`. A workaround was implemented in #870 which effectively pinned runc to an older version, by running the `alpine:3:20` image - this is not ideal because we no longer receive updates, and there's a risk this older image might become patched, or deleted.

A longer term solution is needed so that we can run the latest post-patch version of runc, or some other way to run the nested virtualisation workflow on older kernels.

Some ideas:

- Await documented workarounds/flags to disable unsafe procfs detection
- Try recompiling the kernels with different configuration
- Try a different container engine
- Somehow avoid nested virtualisation
- Offload the build to a remote/self-hosted agent/service

Relates to:

- #870

skl:
## Summary

The VM integration tests are currently locked to an older Alpine version to avoid an error `unsafe procfs detected` from runc.

As KVM is [not available](https://cloud-native.slack.com/archives/C08P4HUFQ6M/p1779989273662129?thread_ts=1779988812.462079&cid=C08P4HUFQ6M) on the CNCF ARM GitHub self-hosted runners, this PR instead swaps out the OCI runtime from `runc` to `crun` (an alternative runtime created by RedHat).

This should allow us to continue to run different kernels in CI, whilst also bringing in the latest version of Alpine for security updates.

Fixes #872.

## Validation

### DIFF
--- a/internal/test/vm/Dockerfile
+++ b/internal/test/vm/Dockerfile
@@ -38,10 +38,15 @@ RUN mkdir -p /etc/docker && cat <<'EOF' > /etc/docker/daemon.json
 }
 EOF
 
-# Default iptables symlink to legacy; modules-load may relink to nft on
-# RHEL kernels that lack CONFIG_IP_NF_NAT modules.
-RUN ln -sf /sbin/iptables-legacy /sbin/iptables && \
-    ln -sf /sbin/ip6tables-legacy /sbin/ip6tables
+# Default iptables to the legacy backend; modules-load may relink to nft on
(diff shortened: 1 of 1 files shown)


skl:
Closing as `crun` is isolating the host pid namespace to only obi itself for some reason. Will raise an alternative approach in a separate PR using CNCF hosted runners with KVM.

Issue #2174, opened 2026-05-27
GenAI payload parsing fails with EOF when the large buffer is smaller then the contentlegth


esara:
### Component

OBI (eBPF instrumentation)

### What happened?

While testing GenAI instrumentation (request payload parsing), it fails with EOF and does not generate a trace, for example if I set the ebpf.buffer_size.http to 8192, but the request payload is larger

This error is silently eaten by
https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/ebpf/common/http/openai.go#L70

### What did you expect to happen?

I would expect a proper trace to be generated with genai semantic conventions

### Steps to reproduce

I can easily reproduce this by a simple python rag application with openai
https://github.com/esara/rag-llamaindex-wikipedia

### OBI configuration

```yaml
ebpf:
      buffer_sizes:
        http: 8192
      payload_extraction:
        http:
          genai:
            mcp:
              enabled: true
            openai:
              enabled: true
```

### Relevant logs

```shell
Adding a small local change to print the request and continue even with the EOF:

time=2026-05-27T01:16:16.806Z level=INFO msg="OpenAISpan: truncated request body, continuing with partial data" error="unexpected EOF" bytes=7332 contentLength=9343 spanContentLength=9343

[...]

PR #2191, opened 2026-05-29
best effort parsing of large buffers

esara:
## Summary

Use besteffort parsing of the genai request and response, similar to how qwen was already handled, but consistently across all parsing. Even if the buffer size is smaller then the request/response - parse it as much of the attributes (like operation and request model) as possible.

addressing https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2174

## Validation

### DIFF
--- a/pkg/ebpf/common/http/openai.go
+++ b/pkg/ebpf/common/http/openai.go
@@ -4,9 +4,7 @@
 package ebpfcommon // import "go.opentelemetry.io/obi/pkg/ebpf/common/http"
 
 import (
-	"bytes"
 	"encoding/json"
-	"io"
 	"log/slog"
 	"net/http"
 	"strings"
@@ -66,27 +64,26 @@ func OpenAISpan(baseSpan *request.Span, req *http.Request, resp *http.Response) 
 		return *baseSpan, false
 	}
 
-	reqB, err := io.ReadAll(req.Body)
-	if err != nil {
+	reqB, ok := readHTTPRequestBody("OpenAISpan", req, baseSpan, "headers", resp.Header)
+	if !ok {
 		return *baseSpan, false
 	}
-	req.Body = io.NopCloser(bytes.NewBuffer(reqB))
 
-	respB, err := getResponseBody(resp)
-	if err != nil && len(respB) == 0 {
+	respB, ok := readHTTPResponseBody("OpenAISpan", resp, baseSpan, "headers", resp.Header)
+	if !ok {
 		return *baseSpan, false
 	}
 
 	slog.Debug("OpenAI", "request", string(reqB), "response", string(respB))
 
-	var parsedRequest request.OpenAIInput
-	if err := json.Unmarshal(reqB, &parsedRequest); err != nil {
-		slog.Debug("failed to parse OpenAI request", "error", err)
-	}
+	parsedRequest := parseOpenAIInput(reqB)
+	parsedResponse := parseVendorOpenAI(respB)
 
-	var parsedResponse request.VendorOpenAI
-	if err := json.Unmarshal(respB, &parsedResponse); err != nil {
-		slog.Debug("failed to parse OpenAI response", "error", err)
+	if parsedResponse.ResponseModel == "" {
+		parsedResponse.ResponseModel = parsedRequest.Model
+	}
+	if parsedRequest.Model == "" {
+		parsedRequest.Model = parsedResponse.ResponseModel
 	}
 
 	parsedResponse.Request = parsedRequest
--- a/pkg/ebpf/common/http/partial_json.go
+++ b/pkg/ebpf/common/http/partial_json.go
@@ -1,0 +1,227 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package ebpfcommon // import "go.opentelemetry.io/obi/pkg/ebpf/common/http"
+
+import (
+	"bytes"
+	"io"
+	"log/slog"
+	"net/http"
+	"regexp"
+	"strings"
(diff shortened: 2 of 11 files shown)


mariomac:
As a first implementation, LGTM! But I think that the code could be simplified (and get a more efficient code) if we provide a more graceful way to parse incomplete JSON. If we want to stay in the standard library `json.Decoder` would do the job but it's complex to use, so I'd suggest just replacing standard `json.Unmarshall` by JSONIter, which is easy to use, [way faster](https://jsoniter.com/benchmarks/go-benchmark.png), and gracefully handles incomplete JSONs. 

For example:
```go
package main
import (
	"fmt"
	jsoniter "github.com/json-iterator/go"
)
type Thing struct {
	FirstName string
	Surname   string
	Address   string
}
func main() {
	for i, input := range []string{
		`{"FirstName":"Foo","Surname":"Bar","Address":"Somewhere"}`,
		`{"FirstName":"Foo","Surname":"Bar","Address":"Somewhere"`,
		`{"FirstName":"Foo","Surname":"Bar","Address":"Some`,
		`{"FirstName":"Foo","Surname":"B`,
	} {
		t := Thing{}
		jsoniter.Unmarshal([]byte(input), &t)
		fmt.Printf("%d: %+v\n", i, t)
	}
}
```
Output:
```
0: {FirstName:Foo Surname:Bar Address:Somewhere}
1: {FirstName:Foo Surname:Bar Address:Somewhere}
2: {FirstName:Foo Surname:Bar Address:}
3: {FirstName:Foo Surname: Address:}
```

esara:
> As a first implementation, LGTM! But I think that the code could be simplified (and get a more efficient code) if we provide a more graceful way to parse incomplete JSON. If we want to stay in the standard library `json.Decoder` would do the job but it's complex to use, so I'd suggest just replacing standard `json.Unmarshall` by JSONIter, which is easy to use, [way faster](https://jsoniter.com/benchmarks/go-benchmark.png), and gracefully handles incomplete JSONs.
> 
> For example:
> 
> ```go
> package main
> import (
> 	"fmt"
> 	jsoniter "github.com/json-iterator/go"
> )
> type Thing struct {
> 	FirstName string
> 	Surname   string
> 	Address   string
> }
> func main() {
> 	for i, input := range []string{
> 		`{"FirstName":"Foo","Surname":"Bar","Address":"Somewhere"}`,
> 		`{"FirstName":"Foo","Surname":"Bar","Address":"Somewhere"`,
> 		`{"FirstName":"Foo","Surname":"Bar","Address":"Some`,
> 		`{"FirstName":"Foo","Surname":"B`,
> 	} {
> 		t := Thing{}
> 		jsoniter.Unmarshal([]byte(input), &t)
> 		fmt.Printf("%d: %+v\n", i, t)
> 	}
> }
> ```
> 
> Output:
> 
> ```
> 0: {FirstName:Foo Surname:Bar Address:Somewhere}
> 1: {FirstName:Foo Surname:Bar Address:Somewhere}
> 2: {FirstName:Foo Surname:Bar Address:}
> 3: {FirstName:Foo Surname: Address:}
> ```

thanks @mariomac this was a fairly easy change, I just squashed into the same commit

mariomac (approved):
LGTM! Waiting for a second pair of eyes before merging

pkg/ebpf/common/http/partial_json.go
```diff
@@ -0,0 +1,160 @@
...
+// unmarshalJSONBestEffort unmarshals body into v using json-iterator, which
+// populates fields seen before truncation even when the JSON is incomplete.
+func unmarshalJSONBestEffort(body []byte, v any) {
+	if len(body) == 0 {
+		return
+	}
+	_ = jsonBestEffort.Unmarshal(body, v)
+}
+
+func readAllBestEffort(r io.Reader) ([]byte, error) {
```
rafaelroquetto: I don't believe we should have this function in its current shape - all it's doing is obsfuscating the real call to `io.ReadAll(r)` from the reader at the call sites - it gives the impression that it does something else it actually doesn't.

pkg/ebpf/common/http/openai.go
```diff
@@ -66,27 +66,35 @@ func OpenAISpan(baseSpan *request.Span, req *http.Request, resp *http.Response)
 		return *baseSpan, false
 	}
 
-	reqB, err := io.ReadAll(req.Body)
-	if err != nil {
+	reqB, err := readAllBestEffort(req.Body)
```
rafaelroquetto: See my comment at the readAllBestEffor definition site.

Perhaps instead of doing without `readAllBestEffort` altogether, we could simply encapsulate the repeated code down below into it, to make things clearer.

If I understood it correctly, best effort means:

- try to read
- if err == nil -> good
- else if len == 0 -> log and drop
- else -> log truncated body and continue


perhaps `readAllBestEffort` can encapsulate the pattern above?

rafaelroquetto (changes requested):
I think we should use readAllBestEffort to encapsulate the behaviour that seems to span across the different AI parsers. Or remove it altogether.

pkg/ebpf/common/http/partial_json.go
esara: @rafaelroquetto you are absolutely right, I did not like my own code - I force pushed an update, that is slightly easier to read and common across the genai jsons

pkg/ebpf/common/http/openai.go
esara: I tried to encapsulate, let me know - happy to improve further

grcevski (approved):
LGTM!

rafaelroquetto (approved):
LGTM

PR #2194, opened 2026-05-29
ci: use latest alpine in nested virt tests

### LINKED ISSUE #872: Find a way to use latest runc in VM workflows

skl:
## Summary

This PR unblocks the version pinning on Alpine 3.20, which was in place to avoid newer versions of `runc`.

It seems the latest Alpine 3.23.4 no longer has the `unsafe procfs` issue that prevented the upgrade previously.

Fixes #872.

## Validation


mmat11 (approved):
lgtm if CI pass

skl:
@mmat11 I think CI is blocked in main by #2196

internal/test/vm/Dockerfile
mmat11, line 91: seems like there's an issue with RH kernels -> https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/26643059518/job/78522259401, claude suggests to add these 2 lines here:

```
    modprobe iptable_raw 2>/dev/null || true
    modprobe iptable_mangle 2>/dev/null || true
```

skl: Done in 644cd2ebb

mmat11: seems to be working 🥳

rafaelroquetto (approved):

PR #2198, opened 2026-05-29
Make stats ring buffer wakeup threshold configurable

pinoOgni:
## Summary

This PR adds `stats_wakeup_data_bytes` following the same pattern used for `wakeup_data_bytes` in the generic tracer.

TL;DR 

After Rafael's comment in #2157 about the threshold used in `stats_events_flags()`, I ran some tests to see if the 4096-byte threshold was appropriate (from my tests it was, at least locally and on a k8s node with various pods running). 
Then I thought that in a node with much more traffic, with the addition of more stat metrics with different overhead and different event sizes, it would be nice to have a configurable threshold.

## Validation


bpf/statsolly/maps/stats_events.h
```diff
@@ -14,7 +14,14 @@ struct {
     __uint(pinning, OBI_PIN_INTERNAL);
 } stats_events SEC(".maps");
 
+// To be injected from userspace during eBPF program load & initialization.
+// When 0, every submission wakes up userspace immediately.
```
rafaelroquetto, line 18: I'd remove this comment entirely, it's just stating the obvivous (this is a well established ebpf idiom)

rafaelroquetto (approved):
I think this is ok, so approving, but I have a couple of questions that piggy-back on the configuration creep we had been discussing in the past weeks:
1. what's the use case for this? are we hitting any scenarios that require this kind of tuning?
2. how do we explain this to the user?

Let's go with it if this is solving a problem, but I think we should discuss on the next SIG an approach to the ring buffers that maybe allows us to strip all of these wakeup_len options for something smarter (wondering if we can adjust this dynamically based on the ring buffer pressure - something for another day)

pinoOgni:
> I think this is ok, so approving, but I have a couple of questions that piggy-back on the configuration creep we had been discussing in the past weeks:
> 
> 1. what's the use case for this? are we hitting any scenarios that require this kind of tuning?
> 2. how do we explain this to the user?

I think the answer to both questions lies in why we introduced `wakeup_len` for the generictracer. 
Statsolly hasn't had any issues requiring a threshold change yet, and I've tried to prevent that by introducing this configuration. In particular, the `tcp_io` metric introduces a significant overhead which, as you pointed out in the review, could require a change in the threshold, but instead of being hardcoded it is configurable.

> Let's go with it if this is solving a problem, but I think we should discuss on the next SIG an approach to the ring buffers that maybe allows us to strip all of these wakeup_len options for something smarter 

Fine by me.

> (wondering if we can adjust this dynamically based on the ring buffer pressure - something for another day)
This would be really cool!

rafaelroquetto:
@pinoOgni makes sense, perfetto! Thanks for the clarifications :)

bpf/statsolly/maps/stats_events.h
pinoOgni: Fine by me, I just followed more or less the same comment here https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/1cf541ccac7318e393f9ad973098c85cfe2e5436/bpf/common/ringbuf.h#L25

WDYT?

rafaelroquetto: I should have pointed out it's just a nit (sorry, I don't mean to bike-shed). The comment you are proposing is fine too! As you please :)

mariomac (approved):
Looks great! Ready to merge as soon as the failing CI task is fixed.

mmat11 (approved):
lgtm, unit tests are failing though

PR #2204, opened 2026-05-30
fix: skip SSL_read buffer parsing on non-positive return values

### LINKED ISSUE #2029: Failed SSL reads still flow into protocol parsing
## Summary

The libssl read probes still forward raw `SSL_read` return values into `handle_ssl_buf`, and the parser path no longer rejects non-positive lengths before handing the buffer to downstream protocol handlers.

## Impact

When SSL reads fail or hit EOF, OBI can still parse stale application buffers and export bytes that were not freshly read. The effect is limited to telemetry disclosure and parser work on error paths, so `low` severity is appropriate.

## Environment

- First identified in: d1760a3b3c77772172d82b6c08c65d76ab3ebbe8

## Evidence

- `bpf/generictracer/libssl.c` passes the raw `ret` value from `SSL_read` to `handle_ssl_buf`.
- `bpf/generictracer/ssl_defs.h` no longer requires `bytes_len > 0` before calling `handle_buf_with_connection`.
- Downstream TCP parsing still consumes the supplied length as parse input.

## Steps to reproduce

1. From the repository root on `main`, build the project with `make generate` and the normal agent build so the SSL BPF programs are available, then run the instrumenter on a TLS client or server pair.
3. Force `SSL_read` to return `0` or `-1` by closing the TLS connection mid-read or otherwise triggering an SSL error.
4. Observe in BPF ring-buffer events or debug logs that the parser still reads and exports bytes from the reused user buffer despite the non-positive SSL read result, proving stale-buffer exposure on the error path.

## Suggested Fix Direction

Reject `SSL_read` results that are `<= 0` before any parsing or cleanup path that assumes fresh buffer contents.

## Acceptance Criteria

- Failed or zero-length SSL reads do not enter payload parsing.
- Error-path cleanup does not consume application buffer contents.
- Tests cover `SSL_read` error and EOF returns.

## Note

M-sasank:
Fixes #2029

## Summary

Validated the issue and confirmed the behavior of ret <=0 error against SSL docs([here](https://docs.openssl.org/1.1.1/man3/SSL_read/)) 

**Why this issue occured?**
A preexisting guard was removed as a side-effect of `d1760a3b`, which dropped the `bytes_len > 0` check from `handle_ssl_buf` to allow connection-tracking setup to run even on zero-length calls. This exposed this bug for which the fix is a single guard in the SSL_read uretprobe.

`SSL_read` returns `> 0` on success (bytes written to buf), `0` on EOF, and `< 0` on error. In the EOF and error cases the application buffer is untouched, so forwarding it to `handle_ssl_buf` exposes stale data to the protocol parser.

This PR adds the equivalent check directly in the `SSL_read` uretprobe before `handle_ssl_buf` is called. I also verified the `SSL_read_ex` version already has the correct guard (`ret != 1`); and I aligned `SSL_read` with the same pattern.

## Changes

- `bpf/generictracer/libssl.c`: add `if (ret <= 0) { return 0; }` guard in `obi_uretprobe_ssl_read` after the map delete, before `handle_ssl_buf`
- `bpf/tests/bpf_ssl_read_guard.c`: plain-C unit test (stubs BPF map helpers and `handle_ssl_buf`) verifying the guard on error, EOF, and success paths

## Verification

Verified with live BPF debug logs (`OTEL_EBPF_BPF_DEBUG=true`, reading `/sys/kernel/tracing/trace_pipe`) against an openssl TLS server. Reproduce: send `GET / HTTP/1.0\r\n\r\n`, then Ctrl+C to force EOF.

**Before (main — bug):**
```
uretprobe SSL_read ret=15  →  SSL_buf        ✓  real data forwarded
uretprobe SSL_read ret=1   →  SSL_buf        ✓  real data forwarded
uretprobe SSL_read ret=0   →  SSL_buf        ✗  EOF: stale buffer forwarded to parser
```

**After (this fix):**
```
uretprobe SSL_read ret=15  →  SSL_buf        ✓  real data forwarded
uretprobe SSL_read ret=1   →  SSL_buf        ✓  real data forwarded
uretprobe SSL_read ret=0   →  (nothing)      ✓  guard returned early, parser not reached
```

  ## Validation

  - [x] I have read and followed the [contributing
  guidelines](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/CONTRIBUTING.md)
  - [x] If this enhances / fixes / changes a core feature, I have updated the [features
  documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support
   matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed.

### DIFF
--- a/bpf/generictracer/libssl.c
+++ b/bpf/generictracer/libssl.c
@@ -65,6 +65,10 @@ int BPF_URETPROBE(obi_uretprobe_ssl_read, int ret) {
 
     bpf_map_delete_elem(&active_ssl_read_args, &id);
 
+    if (ret <= 0) {
+        return 0;
+    }
+
     // must be last in the function, doesn't return
     handle_ssl_buf(ctx, id, args, ret, TCP_RECV);
     return 0;
--- a/bpf/tests/bpf_ssl_read_guard.c
+++ b/bpf/tests/bpf_ssl_read_guard.c
@@ -1,0 +1,194 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Tests for the SSL_read uretprobe return-value guard in
+ * bpf/generictracer/libssl.c.
+ *
+ * SSL_read returns > 0 on success (bytes written to buf), 0 on EOF, and
+ * < 0 on error.  In the EOF and error cases the application buffer has not
+ * been touched, so forwarding its contents to the protocol parser would
+ * expose stale data.  The uretprobe must bail out before calling
+ * handle_ssl_buf whenever ret <= 0.
+ *
+ * The guard was absent after commit d1760a3b removed the bytes_len > 0
+ * check from handle_ssl_buf (needed for large-payload connection tracking).
+ * SSL_read_ex already guards with `ret != 1`; these tests cover the
+ * equivalent guard that was added to SSL_read.
+ *
+ * Pattern mirrors bpf/tests/bpf_kafka_large_buffer.c: the BPF map helpers
+ * and handle_ssl_buf are stubbed in userspace so the handler logic can be
+ * exercised without a kernel.
+ */
+
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+typedef uint8_t u8;
+typedef uint64_t u64;
+
+#define TCP_RECV 0
+
+// Matches ssl_args_t in bpf/common/ssl_args.h
+typedef struct ssl_args {
+    u64 ssl;
+    u64 buf;
+    u64 len_ptr;
+    u64 flags;
+} ssl_args_t;
+
+// --- active_ssl_read_args map stub ---------------------------------------
+
+#define MAP_SIZE 8
+
+typedef struct {
+    u64 key;
+    ssl_args_t value;
+    int used;
+} ssl_read_args_entry_t;
+
+static ssl_read_args_entry_t active_ssl_read_args[MAP_SIZE];
+
+static ssl_args_t *ssl_read_args_lookup(u64 id) {
+    for (int i = 0; i < MAP_SIZE; i++) {
+        if (active_ssl_read_args[i].used && active_ssl_read_args[i].key == id) {
+            return &active_ssl_read_args[i].value;
+        }
+    }
+    return NULL;
+}
+
+static void ssl_read_args_update(u64 id, const ssl_args_t *val) {
+    for (int i = 0; i < MAP_SIZE; i++) {
+        if (!active_ssl_read_args[i].used) {
+            active_ssl_read_args[i].used = 1;
+            active_ssl_read_args[i].key = id;
+            active_ssl_read_args[i].value = *val;
+            return;
+        }
+    }
+}
+
+static int ssl_read_args_delete_count;
+
+static void ssl_read_args_delete(u64 id) {
+    for (int i = 0; i < MAP_SIZE; i++) {
+        if (active_ssl_read_args[i].used && active_ssl_read_args[i].key == id) {
+            active_ssl_read_args[i].used = 0;
+            ssl_read_args_delete_count++;
+            return;
+        }
+    }
+}
+
+// --- handle_ssl_buf stub -------------------------------------------------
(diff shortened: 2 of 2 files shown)


grcevski:
The important bit is here:

> The guard was removed as a side-effect of [d1760a3](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/commit/d1760a3b3c77772172d82b6c08c65d76ab3ebbe8), which dropped the bytes_len > 0 check from handle_ssl_buf to allow connection-tracking setup to run even on zero-length calls.

There is no point in adding anything to the read path, it's not possible for read to be involved in failed connection establishing.

grcevski:
Sorry, I was confused by that comment, the send path doesn't have to do anything with this fix. 

Can you explain this part:

> In the EOF and error cases the application buffer is untouched, so forwarding it to handle_ssl_buf exposes stale data to the protocol parser.

if the data length was 0 what would we read as stale?

M-sasank:
> if the data length was 0 what would we read as stale?

The previous `SSL_read`'s data. When a prior `SSL_read` on that buffer succeeded, the terminating `<= 0` call still holds that last successful read's plaintext (`SSL_read` doesn't touch the buffer on `0`/`< 0`). So by "stale", I meant the previous successful read's decrypted bytes, still at the same address.

Even setting aside what exactly leaks, on every `ret <= 0` we still classify and run the buffer through the protocol parsers, an unnecessary parser work on close/error paths. What follows is the specifics of *which* bytes get read.

There are different cases of what exact data is read as stale. Specifically the following:

#### When ret == 0 (EOF) two unconditional stale reads:
1. `make_protocol_args` does `bpf_probe_read(small_buf, MIN_HTTP2_SIZE /*24*/, u_buf)` with no `bytes_len` check, then `is_http(small_buf, MIN_HTTP_SIZE, …)` classifies it. Hence, parsing and classifying stale data at `small_buf`.
2. 256-byte event capture: on the new-request branch, `bpf_probe_read(info->buf, FULL_BUF_SIZE /*256*/, u_buf)` fills `info->buf`, the field later emitted to userspace as the request's captured payload , again no `bytes_len` check, so up to 256 stale bytes are staged for export.

#### ret < 0 (error) , same reads, plus a signedness hazard:
3. Same stale reads as above fire on `ret == -1` too.
4. Signedness hazard. `bytes_len` is a signed `int`, but `http_send_large_buffer` takes `u32`. So `-1` reinterprets to *4294967295*, which bypasses its `bytes_len == 0` short-circuit and stages stale bytes for the large-buffer emit.


All of the above I verified live on a debug build with the guard removed, watching `trace_pipe`:
```
=== uretprobe SSL_read id=... ret=0 ===
CLAIM1 classify ssl=1 len=0 buf=[GET / HTTP/1.0
CLAIM2 newreq_export len=0 info_buf=[GET / HTTP/1.0

=== uretprobe SSL_read id=... ret=-1 ===
CLAIM3 largebuf_proceed bytes_len_u32=4294967295 ptype=1
```

As reflected by the severity in the original issue, the chances of leaking parseable data are not high. Bounded scope:
- Disclosure is ≤256 bytes via `info->buf` in the default config (no extra flags). The separate large-buffer path (point 4 above) can stage more than that, but only when body capture is explicitly enabled via `OTEL_EBPF_BPF_BUFFER_SIZE_HTTP`. Either way, it only fires when the reused buffer's leading bytes still parse as a protocol, i.e. the previous read's first bytes are valid parseable HTTP. This goes off for both `ret == 0` and `ret == -1`.
- Net effect of the issue reported is wasted parser work on stale data on close/error paths. This is why the original issue is at "low" severity.

grcevski:
Closing this as per our AI policy https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/AI-POLICY.md

M-sasank:
Hey, Sorry for this mess
I didn't read the AI Policy beforehand. I admit my mistake on using AI for communication and descriptions since i consider that as "refined", despite the work being validated by me manually.

Will ensure this doesn't happen with future PRs. You can keep this one closed

PR #2210, opened 2026-06-01
Add SunRPC (ONC RPC) protocol support

damemi:
## Summary

SunRPC (also ONC RPC, Open Network Computing RPC) is the RPC framework behind classic Unix services — NFS, mount, portmapper/rpcbind, NIS, etc. Defined in [RFC 5531](https://datatracker.ietf.org/doc/html/rfc5531).

Relevant semconv attributes from: https://opentelemetry.io/docs/specs/semconv/registry/attributes/onc-rpc/

Conceptually:

* A client sends a CALL to a program (e.g. NFS = 100003), version, and procedure (method number).
* A server responds with a REPLY (accepted/rejected, plus optional result data).
* Over TCP, messages are framed with record marking (length + last-fragment bit), then XDR-encoded RPC bodies.

It predates gRPC but fills a similar role in legacy/enterprise Linux infrastructure.

### Protocol wire format

Per RFC 5531, a TCP SunRPC stream looks like:

```
┌─────────────────────────────────────┐
│ Record marking (4 bytes, BE)        │  high bit = last fragment; low 31 bits = length
├─────────────────────────────────────┤
│ RPC message body (XDR)              │
│   xid (4)                           │  transaction ID
│   msg_type (4)                      │  0 = CALL, 1 = REPLY
│   ...                               │
└─────────────────────────────────────┘
```

**CALL body** (after xid + msg_type):
```
rpcvers (=2) | prog | vers | proc | credential (opaque_auth) | verifier (opaque_auth) | [procedure args...]
```

**REPLY body** (accepted reply):
```
reply_stat | verifier (opaque_auth) | accept_stat | [results...]
```

Each `opaque_auth` is: `flavor (4) | length (4) | body (padded to 4-byte boundary)`.

Standard auth flavors: NULL (0), UNIX (1), SHORT (2), DES (3), KERB (4), RSA (5), RPCSEC_GSS (6).

### How we detect it

**1. Kernel (primary) — `bpf/generictracer/protocol_sunrpc.h`**

On the first packet of an unknown TCP connection, `is_sunrpc()` checks whether the buffer starts with a valid SunRPC record:

1. **Record marking**: last-fragment bit set, length 24–1 MiB, full record present in buffer.
2. **Message type**: CALL or REPLY.
3. **CALL validation**: `rpcvers == 2`, program in a known range (100000–101000 or registered high range), `vers != 0`, `proc <= 0xffff`, two valid `opaque_auth` blocks (credential + verifier) with recognized auth flavors.
4. **REPLY validation**: accepted reply with valid verifier and `accept_stat <= 5`.

If it passes, the connection is tagged `ProtocolTypeSunRPC` and cached in `protocol_cache` so later packets skip re-classification.

**2. Userspace fallback — `pkg/ebpf/common/sunrpc_detect_transform.go`**

If OBI attaches mid-connection and the kernel missed the first RPC, `matchSunRPC()` in the TCP heuristic path re-runs the same logic via `sunrpcparser.IsLikelySunRPC()`.

### How we decode it

**Userspace parser — `pkg/internal/sunrpcparser/`**

`Parse()` walks TCP record marking (including multi-fragment reassembly), then for each record:

| Extracted | Skipped |
|-----------|---------|
| `xid`, `prog`, `vers`, `proc`, auth flavor | Auth credential/verifier bodies |
| REPLY `accept_stat` (when paired by xid) | Procedure argument XDR |
| | Reply result XDR |

Procedure arguments are **not** decoded for any auth flavor — only header fields are extracted. That would require per-program XDR schemas (NFS vs mount vs portmapper). Procedure *names* are not mapped yet either; spans use the numeric procedure ID (e.g. `portmapper/0`).

**Span construction**

- Prefer CALL from either request or response buffer (handles ephemeral-port ambiguity, commented in code).
- Pair CALL + REPLY by matching `xid` to populate `rpc.response.status_code`.
- Map known program numbers to names (`nfs`, `mount`, `portmapper`, …).
- Client/server span kind from TCP direction + CALL presence.

Ambiguity between CALL/REPLY can occur for example if OBI is enabled mid-connection and misses the initial CALL, it can then misinterpret server REPLYs as a "client request" when relying solely on packet direction. To fix this we inspect the messages to determine if they are actually a server REPLY.

### Limitations

- **TCP only** (no UDP SunRPC).
- Kernel classification needs a complete single-fragment record in the capture buffer.
- No distributed context propagation on SunRPC.
- NFS pseudo-flavors (e.g. krb5i at 390004) not in the base 0–6 auth allowlist.
- Procedure names unmapped — numbers only unless extended later.

---

[AI policy](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/cb4a989f314eb4f01dbdde749aabf6b98a724a66/AI-POLICY.md): This PR was developed with the help of Cursor to assist me in understanding the SunRPC protocol, its implementation, and how to best integrate it with OBI (particularly in consideration of current patterns and practices for existing protocol support).

## Validation


shivanshuraj1333 (approved):

pkg/appolly/app/request/span_getters.go
```diff
@@ -100,7 +109,51 @@ func spanOTELGetters(name attr.Name) (attributes.Getter[*Span, attribute.KeyValu
 			return semconv.RPCService("")
 		}
 	case attr.RPCGRPCStatusCode:
-		getter = func(s *Span) attribute.KeyValue { return semconv.RPCGRPCStatusCodeKey.Int(s.Status) }
+		getter = func(s *Span) attribute.KeyValue {
+			if s.Type == EventTypeSunRPCClient || s.Type == EventTypeSunRPCServer {
```
grcevski: I think we need to look to see what the right semconv attribute should be used here, because the attr.RPCGRPCStatusCode will write `RPCGRPCStatusCodeKey = attribute.Key("rpc.grpc.status_code")`, which I think is not correct for SunRPC.

grcevski: I see we have RPCResponseStatusCode below, do we end up getting here for this protocol and that's why you needed to add this if SunRPC statement?

pkg/ebpf/common/sunrpc_detect_transform.go
```diff
@@ -0,0 +1,235 @@
...
+	case reqErr == nil || respErr == nil:
+		// Framing looked plausible but no CALL/REPLY extracted; not worth an error span.
+		return nil, true, nil
+	}
+
+	if errors.Is(reqErr, sunrpcparser.ErrNotSunRPC) && errors.Is(respErr, sunrpcparser.ErrNotSunRPC) {
+		return nil, true, sunrpcparser.ErrNotSunRPC
+	}
+
+	return nil, true, errors.Join(reqErr, respErr)
```
grcevski: Does errors.Join allocate a new error each time? If it does, can we make this a hardcoded error please? Since these parses run on unidentified protocols any allocations slow down the pipeline a lot.

pkg/ebpf/common/sunrpc_detect_transform.go
```diff
@@ -0,0 +1,232 @@
...
+		ProgramName: progName,
+		Method:      sunrpcparser.ProcedureLabel(call.Program, call.Procedure),
+		AuthFlavor:  sunrpcparser.AuthFlavorName(call.AuthFlavor),
+	}
+
+	if reply != nil && reply.MatchCallXid {
+		if reply.AcceptStat != sunrpcAcceptSuccess {
+			info.Status = int(reply.AcceptStat) + 1
+		}
+	}
```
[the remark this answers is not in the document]
damemi: This is a total hallucination, the [ONC RPC Semconv](https://opentelemetry.io/docs/specs/semconv/registry/attributes/onc-rpc/) does not say anything about span.status holding the actual accept_stat value. booo

pkg/export/prom/prom.go
```diff
@@ -511,7 +516,7 @@ func newReporter(
 				NativeHistogramMinResetDuration: cfg.NativeHistogram.MinResetDuration,
 			}, labelNames(attrHTTPClientDuration)).MetricVec, clock.Time, cfg.TTL)
 		}),
-		grpcDuration: optionalHistogramProvider(is.GRPCEnabled(), func() *Expirer[prometheus.Histogram] {
+		grpcDuration: optionalHistogramProvider(is.GRPCEnabled() || is.SunRPCEnabled(), func() *Expirer[prometheus.Histogram] {
```
grcevski, line 540: Should these be in the grpc duration metrics or new metric series? The OTel metrics exporter seems to use new series RPCServerDuration, I think we should do the same with the prom export.

grcevski (commented):
This is looking great overall!  

I left a few comments that I think need to be addressed. Since for Prometheus export I think the data would have ended up in the grpc metrics, instead of in the sunrpc metrics, do you mind extending the integration test to use the prom exporter too? We have examples for that with the HTTP go test programs.

bpf/generictracer/protocol_sunrpc.h
```diff
@@ -0,0 +1,183 @@
...
+
+static __always_inline u8 sunrpc_valid_auth_flavor(u32 flavor) {
+    switch (flavor) {
+    case 0: // AUTH_NULL
+    case 1: // AUTH_UNIX
+    case 2: // AUTH_SHORT
+    case 3: // AUTH_DES
+    case 4: // AUTH_KERB
+    case 5: // AUTH_RSA
+    case 6: // RPCSEC_GSS
```
rafaelroquetto: can we make these constants, instead of comments? i.e.

```
enum sunrpc_auth_flavor : u8 {
 k_auth_null = 0,
 /* ... */
};
```

pkg/ebpf/common/sunrpc_detect_transform.go
```diff
@@ -0,0 +1,232 @@
...
+	Status      int
+}
+
+// ProcessPossibleSunRPCEvent parses both TCP capture buffers and picks the best SunRPC
+// metadata for span creation. The ignore bool means drop quietly; ErrNotSunRPC means
+// this event is not SunRPC.
+func ProcessPossibleSunRPCEvent(event *TCPRequestInfo, pkt, rpkt *largebuf.LargeBuffer) (*SunRPCInfo, bool, error) {
+	// ignore=true when a buffer is empty, not SunRPC, or inconclusive — see processSunRPCBuffer.
+	reqInfo, reqIgnore, reqErr := processSunRPCBuffer(pkt)
+	respInfo, respIgnore, respErr := processSunRPCBuffer(rpkt)
```
rafaelroquetto, line 36: non-rhetorical question: are we parsing both buffers here when we only need either? Or do we always need to parse both unconditionally?

pkg/appolly/app/request/span_getters.go
damemi: yeah it's kind of messy because, as I can tell, we define `rpc.grpc.status_code` as one of the default attrs on `RPCClientDuration` and `RPCServerDuration` metrics.

to me, SunRPC should definitely be included in those generic duration metrics. I think maybe we shouldn't have included gRPC.status_code as a default attribute because now it's included in any RPC system that wants to use the generic duration metrics. This gate avoids writing the sunrpc status as a gRPC status (which both have totally different semantics -- and sunrpc status isn't even defined in semconv)

We could give SunRPC its own metrics, or just not report metrics for it if we want to keep this strictly to grpc. Alternatively, for SunRPC cases we can remove the `grpc.status_code` from the attribute names list for sun metrics

pkg/export/prom/prom.go
damemi: I think this is really more of a misnomer on the `grpcDuration` variable. Its a similar problem to your other comment... these aren't strictly "grpc" metrics; they're `rpc_server_duration` and `rpc_client_duration`.

imo just renaming this variable to `rpcDuration` makes it clearer that this is for any rpc system. but you have the same problem with `grpc.status_code` showing up in the base RPC metrics

pkg/ebpf/common/sunrpc_detect_transform.go
damemi: yeah we can do that thanks

damemi:
> Since for Prometheus export I think the data would have ended up in the grpc metrics, instead of in the sunrpc metrics, do you mind extending the integration test to use the prom exporter too? We have examples for that with the HTTP go test programs.

addressed inline, but I think the main takeaway is that these _aren't_ strictly gRPC metrics. They're base [RPC client/server metrics](https://opentelemetry.io/docs/specs/semconv/rpc/rpc-metrics/), but OBI currently includes `grpc.status` by default. So SunRPC actually should use these metrics.

Maybe we should remove `grpc.status` from the default attributes and only _insert_ it for grpc? Then we can have a cleaner split between RPC systems (in this case SunRPC/gRPC)

pkg/ebpf/common/sunrpc_detect_transform.go
```diff
@@ -0,0 +1,232 @@
...
+	if pkt == nil || pkt.IsEmpty() {
+		return nil, true, sunrpcparser.ErrNotSunRPC
+	}
+
+	reader := pkt.NewReader()
+	if !sunrpcparser.IsLikelySunRPC(&reader) {
+		return nil, true, sunrpcparser.ErrNotSunRPC
+	}
+
+	reader = pkt.NewReader()
```
rafaelroquetto: I believe you can call `reader.Reset()` here to avoid allocating a new one

rafaelroquetto (commented):
Man this is really nice code! Just a few very minor comments.

pkg/ebpf/common/sunrpc_detect_transform.go
damemi: Yeah from what I learned the sunrpc CALL might be in either buffer (in cases like attaching mid connection) and it's not always just based on the direction of the packet, that's what the switch case below this is for

bpf/generictracer/protocol_sunrpc.h
damemi: yup good idea

pkg/ebpf/common/sunrpc_detect_transform.go
damemi: thanks

rafaelroquetto (approved):
LGTM!

bpf/generictracer/protocol_tcp.h
```diff
@@ -215,6 +215,7 @@ static __always_inline int tcp_send_large_buffer(tcp_req_t *req,
         return 0;
     case k_protocol_type_http:
     case k_protocol_type_mqtt:
+    case k_protocol_type_sunrpc:
         break;
```
MrAlias, line 225: SunRPC is classified before entering the generic TCP handler, but this new case falls through without emitting large-buffer chunks. The ring-buffer event still only carries `k_tcp_max_len`/`k_tcp_res_len`, so valid SunRPC records larger than the small capture are parsed from truncated data and dropped. Treating SunRPC like generic TCP large-buffer capture keeps the existing wire-format parser usable for oversized records.

```suggestion
    case k_protocol_type_http:
    case k_protocol_type_mqtt:
        break;
    case k_protocol_type_sunrpc:
        unknown_send_large_buffer(req, pid_conn, u_buf, bytes_len, packet_type, direction, action);
        break;
```

bpf/generictracer/protocol_sunrpc.h
```diff
@@ -0,0 +1,194 @@
...
+    const u32 flavor = sunrpc_read_u32_be(data + *off);
+    const u32 length = sunrpc_read_u32_be(data + *off + 4);
+    if (!sunrpc_valid_auth_flavor(flavor)) {
+        return -1;
+    }
+
+    const u32 padded = (length + 3) & ~3U;
+    if (*off + 8 + padded > data_len) {
+        return -1;
+    }
```
MrAlias: This length is payload-controlled, so `(length + 3)` can wrap before the bounds check. A large opaque-auth length can produce a small `padded` value and falsely validate a non-SunRPC payload. Bound `length` against the remaining record first, then compute the padded length.

```suggestion
    if (*off > data_len || data_len - *off < 8) {
        return -1;
    }

    const u32 flavor = sunrpc_read_u32_be(data + *off);
    const u32 length = sunrpc_read_u32_be(data + *off + 4);
    if (!sunrpc_valid_auth_flavor(flavor)) {
        return -1;
    }

    const u32 remaining = data_len - *off - 8;
    if (length > remaining) {
        return -1;
    }

    const u32 padded = (length + 3) & ~3U;
    if (padded > remaining) {
        return -1;
    }
```

bpf/generictracer/protocol_sunrpc.h
```diff
@@ -0,0 +1,194 @@
...
+
+    if (sunrpc_read_u32_be(rpc + 4) != k_sunrpc_msg_reply) {
+        return 0;
+    }
+
+    const u32 reply_stat = sunrpc_read_u32_be(rpc + 8);
+    if (reply_stat != k_sunrpc_reply_accepted) {
+        return reply_stat <= 1;
+    }
```
MrAlias: The userspace parser now validates denied replies, but the BPF classifier still accepts any `reply_stat == 1` without checking `reject_stat`, the RPC-mismatch payload, or the auth-error payload. That can cache a non-SunRPC connection as SunRPC before userspace has a chance to reject it.

pkg/ebpf/common/sunrpc_detect_transform.go
```diff
@@ -0,0 +1,234 @@
...
+		ProgramName: progName,
+		Method:      sunrpcparser.ProcedureLabel(call.Program, call.Procedure),
+		AuthFlavor:  sunrpcparser.AuthFlavorName(call.AuthFlavor),
+	}
+
+	if reply != nil && reply.MatchCallXid && !reply.Denied {
+		if reply.AcceptStat != sunrpcAcceptSuccess {
+			info.Status = int(reply.AcceptStat) + 1
+		}
+	}
```
MrAlias: Denied replies are parsed, but they never set `Status`, so `SpanStatusCode` exports them as unset/success. The earlier status-code discussion covered the accepted-reply mapping; this is the missing `MSG_DENIED` error path.

```suggestion
	if reply != nil && reply.MatchCallXid {
		switch {
		case reply.Denied:
			info.Status = 1
		case reply.AcceptStat != sunrpcAcceptSuccess:
			info.Status = int(reply.AcceptStat) + 1
		}
	}
```

pkg/ebpf/common/sunrpc_detect_transform.go
```diff
@@ -0,0 +1,234 @@
...
+	var subType int
+	if data.Version > 0 && data.Version <= 255 {
+		subType = int(data.Version)
+	}
+
+	return request.Span{
+		Type:         spanType,
+		Method:       data.Method,
+		Path:         data.ProgramName,
+		Route:        strconv.FormatUint(uint64(data.Procedure), 10),
```
MrAlias: For reply-only spans, `Procedure` is still the zero value, but this converts it into `Route == "0"`. Exporters then emit `onc_rpc.procedure.number=0`, which means the NULL procedure was observed even though the CALL was not captured and the procedure is unknown. The previous synthetic `procedure.name` issue is handled, but the procedure number still leaks the same unknown data.

pkg/export/prom/prom.go
```diff
@@ -358,11 +358,16 @@ func newReporter(
...
-	if is.GRPCEnabled() {
-		attrGRPCDuration = attributes.PrometheusGetters(attributeGetters,
-			attrsProvider.For(attributes.RPCServerDuration))
-		attrGRPCClientDuration = attributes.PrometheusGetters(attributeGetters,
-			attrsProvider.For(attributes.RPCClientDuration))
+	rpcServerAttrs := attrsProvider.For(attributes.RPCServerDuration)
+	rpcClientAttrs := attrsProvider.For(attributes.RPCClientDuration)
+	if is.SunRPCEnabled() {
+		rpcServerAttrs = attributes.AppendUniqueNames(rpcServerAttrs, attributes.OncRPCMetricAttributes)
+		rpcClientAttrs = attributes.AppendUniqueNames(rpcClientAttrs, attributes.OncRPCMetricAttributes)
```
MrAlias: This is related to the existing Prometheus RPC metrics discussion, but the concern here is label-schema compatibility. When SunRPC is enabled, these ONC-specific labels are appended to the shared `rpc.server.duration` and `rpc.client.duration` histograms. With the default `InstrumentationALL`, existing gRPC users get extra `onc_rpc_*` labels with empty values, changing the Prometheus series shape and breaking dashboards or recording rules that depend on the previous RPC label set.

pkg/export/prom/prom.go
```diff
@@ -511,7 +516,7 @@ func newReporter(
 				NativeHistogramMinResetDuration: cfg.NativeHistogram.MinResetDuration,
 			}, labelNames(attrHTTPClientDuration)).MetricVec, clock.Time, cfg.TTL)
 		}),
-		grpcDuration: optionalHistogramProvider(is.GRPCEnabled(), func() *Expirer[prometheus.Histogram] {
+		grpcDuration: optionalHistogramProvider(is.GRPCEnabled() || is.SunRPCEnabled(), func() *Expirer[prometheus.Histogram] {
 			return NewExpirer[prometheus.Histogram](prometheus.NewHistogramVec(prometheus.HistogramOpts{
 				Name:                            attributes.RPCServerDuration.Prom,
 				Help:                            "duration of RCP service calls from the server side, in seconds",
```
MrAlias: Small cleanup from the earlier help-text thread: this should say RPC, not RCP.

```suggestion
					Help:                            "duration of RPC service calls from the server side, in seconds",
```

MrAlias (commented):
A few remaining issues to address before this lands. Some of these are adjacent to earlier review threads: the userspace rejected-reply parser issue appears fixed, but the BPF classifier still has the same permissive denied-reply behavior; the Prometheus comments are related to the existing RPC metrics discussion, with one concrete label-schema concern and two small help-text cleanups.

pkg/export/prom/prom.go
```diff
@@ -521,7 +526,7 @@ func newReporter(
 				NativeHistogramMinResetDuration: cfg.NativeHistogram.MinResetDuration,
 			}, labelNames(attrGRPCDuration)).MetricVec, clock.Time, cfg.TTL)
 		}),
-		grpcClientDuration: optionalHistogramProvider(is.GRPCEnabled(), func() *Expirer[prometheus.Histogram] {
+		grpcClientDuration: optionalHistogramProvider(is.GRPCEnabled() || is.SunRPCEnabled(), func() *Expirer[prometheus.Histogram] {
 			return NewExpirer[prometheus.Histogram](prometheus.NewHistogramVec(prometheus.HistogramOpts{
 				Name:                            attributes.RPCClientDuration.Prom,
 				Help:                            "duration of GRPC service calls from the client side, in seconds",
```
MrAlias: This metric is now shared by gRPC and SunRPC, so the help text should be protocol-generic.

```suggestion
					Help:                            "duration of RPC service calls from the client side, in seconds",
```

pkg/export/prom/prom.go
damemi: true, this is the problem I was talking about with @grcevski and you're right, it goes both ways (grpc labels affect sunrpc + sunrpc labels affect grpc)

how do you think we should handle that? I'm wondering if `rpc.grpc.status_code` should even be in the metrics at all now, because looking at the [gRPC metrics semconv](https://opentelemetry.io/docs/specs/semconv/rpc/grpc/#metrics) it just says gRPC metrics should follow the [general RPC semconv](https://opentelemetry.io/docs/specs/semconv/rpc/rpc-metrics/) which just use the generic `rpc.response.status_code`, which would work for all RPC systems

still though, that doesn't answer the reverse of how to handle sunrpc labels. looking more at it, the [sunrpc attributes](https://opentelemetry.io/docs/specs/semconv/registry/attributes/onc-rpc/) I'm using aren't actually part of that main OTel spec semconv (linked above), so maybe we just shouldn't export any additional RPC system metrics besides the base default? or add a config knob for other systems

imo, I think it seems like we should:

1. stop using `rpc.grpc.status_code` and use `rpc.response.status_code` (though I'm aware this is breaking, from what I can tell it's what the actual semconv is for rpc metrics)
2. don't export any additional RPC metric attributes by default, but allow things like `onc_rpc.*` or `rpc.grpc.*` if configured (again not sure how we can avoid the break in removing default `rpc.grpc.status_code`)

Alternatively, for this PR I can just not export any additional SunRPC metric labels (safest) and keep the gated logic to treat `rpc.grpc.status_code` as `0` for SunRPC (weird but need to do something with it)

this is I think the biggest question this PR brings up so I'm curious what people think

bpf/generictracer/protocol_sunrpc.h
```diff
@@ -0,0 +1,230 @@
...
+};
+
+static __always_inline u32 sunrpc_read_u32_be(const unsigned char *p) {
+    u32 v = 0;
+    bpf_probe_read(&v, sizeof(v), p);
+    return bpf_ntohl(v);
+}
+
+static __always_inline u8 sunrpc_valid_program(u32 prog) {
+    if (prog == 0 || prog >= 0x40000000) {
```
mmat11: this first `if` is redundant

bpf/generictracer/protocol_sunrpc.h
```diff
@@ -0,0 +1,230 @@
...
+    const u32 proc = sunrpc_read_u32_be(rpc + 20);
+    if (!sunrpc_valid_program(prog) || vers == 0 || proc > 0xffff) {
+        return 0;
+    }
+
+    u32 off = 24;
+    if (sunrpc_skip_opaque_auth(rpc, rpc_len, &off) != 0) {
+        return 0;
+    }
+    if (sunrpc_skip_opaque_auth(rpc, rpc_len, &off) != 0) {
```
mmat11, line 127: just checking, are there 2 auth blocks that needs skipping? or is this block repeated by mistake?

bpf/generictracer/protocol_sunrpc.h
```diff
@@ -0,0 +1,230 @@
...
+                                    const unsigned char *data,
+                                    u32 data_len,
+                                    enum protocol_type *protocol_type) {
+    if (*protocol_type == k_protocol_type_sunrpc) {
+        return 1;
+    }
+    if (*protocol_type != k_protocol_type_unknown) {
+        return 0;
+    }
+    if (!sunrpc_record_looks_valid(data, data_len)) {
```
mmat11, line 227: how easy is it for this to match false positive buffers? if we mark this as sunrpc for the connection, all the other packets get misclassified

bpf/generictracer/protocol_tcp.h
```diff
@@ -216,6 +216,9 @@ static __always_inline int tcp_send_large_buffer(tcp_req_t *req,
     case k_protocol_type_http:
     case k_protocol_type_mqtt:
         break;
+    case k_protocol_type_sunrpc:
```
mmat11, line 223: I think this should be marked as `k_large_buf_layer_app` and not `wire`; we should probably refactor these methods into a single one, but for now maybe we need some code duplication

bpf/generictracer/protocol_sunrpc.h
damemi: yeah in the call message there are 2 auth blocks, called `cred` and `verf`. they're the same XDR format, and the `skip_opaque_auth` function walks the offset forward to the next block when checking a call (`cred`->`verf`). that checks both blocks in a call to look like valid sunrpc. will add some comments to explain the duplication

bpf/generictracer/protocol_sunrpc.h
damemi: it's possible, but it comes down to the heuristic similar to the other protocols. The biggest risk I think is the wide program number range, which a custom binary with `msg_type=0`, `rpcvers=2`, and ` prog` in that range could hit, but it would still need 2 valid `opaque_auth` blocks (the checks mentioned above)

That risk is probably most likely during a mid-stream attach where something could falsely be identified as a REPLY. Those have the weakest checks, though the message would still have to be aligned with the sunrpc format which is possible with a 1 followed by a length of 0s to pass the REPLY, ACCEPTED, flavor, length, and accept_stat sections.

We could probably tighten this, like with an xid check like Kafka. But to try to keep this to a minimum I added sunrpc at the bottom of all the other protocols

bpf/generictracer/protocol_tcp.h
damemi: hm interesting what effect does this have? Looking into it it seems like this is just a way to classify the bytes? I agree this is more of an "app" and not unknown "wire" traffic but what are the implications of that in OBI?

pkg/export/prom/prom.go
MrAlias: Yeah, I think the safest scope for this PR is the alternative you called out: don’t add the `onc_rpc.*` labels to the shared Prometheus RPC duration metrics by default.

The `rpc.grpc.status_code` issue is real, but changing that to `rpc.response.status_code` is a broader metrics compatibility decision and probably should be handled separately from adding SunRPC support. For this PR, preserving the existing gRPC Prometheus label schema seems more important than exposing the extra ONC RPC metric dimensions immediately.

So my preference would be:

1. keep SunRPC in the shared `rpc.server.duration` / `rpc.client.duration` metrics
2. remove the default `onc_rpc.*` Prometheus labels from those histograms
3. leave any `rpc.grpc.status_code` -> `rpc.response.status_code` migration and configurable RPC-specific metric labels for a follow-up design/change

That keeps this PR from breaking existing Prometheus users while still adding the core SunRPC metrics/spans.

pkg/internal/sunrpcparser/parser.go
```diff
@@ -0,0 +1,384 @@
...
+	if !validAuthFlavor(flavor) {
+		return nil, false, nil
+	}
+	off += n
+
+	_, n, err = readOpaqueAuth(body[off:])
+	if err != nil {
+		return nil, false, err
+	}
+	off += n
```
MrAlias: This validates the CALL credential auth flavor, but the verifier returned by the second `readOpaqueAuth` is never checked. That makes the userspace heuristic accept CALL-shaped buffers with a valid `cred` flavor and an arbitrary `verf` flavor, while the BPF classifier rejects the same shape because `sunrpc_skip_opaque_auth` validates both blocks.

I think the second flavor should be checked with `validAuthFlavor` as well, so userspace fallback and kernel classification keep the same acceptance criteria.

bpf/generictracer/protocol_sunrpc.h
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    const u32 rm = sunrpc_read_u32_be(data);
+    if (!(rm & k_sunrpc_rm_last_frag)) {
+        return 0;
+    }
+
+    const u32 rec_len = rm & k_sunrpc_rm_frag_len_mask;
+    if (rec_len < 24 || rec_len > k_sunrpc_max_record) {
+        return 0;
```
MrAlias, line 200: This top-level record-length gate rejects a minimal valid `MSG_DENIED` / `AUTH_ERROR` reply before the denied-reply validator can run. Per RFC 5531, that reply can be 20 bytes at the RPC message layer (`xid`, `msg_type`, `reply_stat`, `reject_stat`, and `auth_stat`), so `rec_len < 24` excludes it before `sunrpc_parse_reply_msg` sees the message.

The CALL parser still needs a larger minimum, but the common record gate should allow the minimal REPLY size and let `sunrpc_parse_call_msg` / `sunrpc_parse_reply_msg` enforce their own message-specific bounds.

pkg/export/prom/prom.go
damemi: sounds good to me, I agree that's safest. Just didn't know if you'd prefer to solve the whole issue here or break it into its own discussion (I agree with the latter)

pkg/internal/sunrpcparser/parser.go
damemi: sounds good thanks

bpf/generictracer/protocol_sunrpc.h
damemi: 👍

bpf/generictracer/protocol_tcp.h
mmat11: sorry for getting back now, I was on vacation. IIRC this was recently added to avoid mixing up unknown tcp bytes with already classified protocol packets, see: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/f1ec1b2965e7420ecd5288c3378e6194fc7c1dce/pkg/ebpf/common/tcp_large_buffer.go#L89

grcevski (approved):
LGTM!

grcevski:
@damemi if you can resolve the conflict, I think all comments are addressed and we should merge this IMO.

pkg/export/prom/prom.go
damemi: Working on this in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2295
Removed the SunRPC attributes from this PR, and when we update the semconv attributes I will also be able to remove the weird `grpc.status_code` handling here for sun

damemi:
@grcevski thanks! I do still want to get more to @mmat11's comment (https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2210#discussion_r3355338541) about this being classified as wire/app, not sure I understand if there is any functional implication of that change or if it's more code-organizational. Do you have any thoughts on that?

Also I think #2293 will make it so I can drop the `rpc.grpc.status_code` handling I had to do for Sun here. There will probably be some minor conflicts in that but probably just around the metrics.

So I just want to resolve those 2 points and then I agree, will merge after that

damemi:
#2293 merged, rebased and updated to report sunrpc status through the metrics

MrAlias (approved):

PR #2213, opened 2026-06-01
Enrichment status code #2178

### LINKED ISSUE #2178: Support HTTP response status code matching in HTTP body extraction enrichment rules
### Component

OBI (eBPF instrumentation)

### Problem

Currently, the HTTP body extraction enrichment rules match block supports filtering by HTTP method and URL path patterns, but not by HTTP response status code. This makes it impossible to conditionally capture the HTTP response body only for error responses (e.g. 5xx) while still emitting all traces.

  The only workaround is to either:
  - Capture the HTTP response body for all responses 
  - Use the filter section with numeric operators (PR #1323) - but this drops entire traces, not just the HTTP body extraction

  Add status code matching support to the HTTP body extraction enrichment rules match block:

```
  ebpf:
    payload_extraction:
      http:
        enrichment:
          enabled: true
          policy:
            default_action:
              body: exclude
          rules:
            - action: include
              type: body
              match:
                http_response_status_code:
                  greater_equals: 500
                  less_equals: 599
```
  Use Case

  Extract HTTP request and response body (if possible to define scope: all)  only for 5xx errors of response for debugging purposes, while:
  - Still emitting all traces regardless of status code

hectar-glitches:
## Summary

Add HTTP response status code matching to enrichment rules

Closes #2178

Add http_response_status_code to the enrichment rule block, allowing rules to filter by response status code range. This lets users capture HTTP body data only for specific responses (e.g. 5xx errors) without dropping entire traces.

## Validation


pkg/config/payload_extraction.go
```diff
@@ -45,6 +45,13 @@ type HTTPConfig struct {
 	Enrichment EnrichmentConfig `yaml:"enrichment"`
 }
 
+type HTTPResponseStatusCodeRange struct {
```
mmat11: question: should we reuse `MatchDefinition` in `pkg/filter/matcher.go` which is more complete and already implemented?

pkg/config/payload_extraction.go
```diff
@@ -325,24 +342,29 @@ type HTTPParsingMatch struct {
 	URLPathPatterns []services.GlobAttr `yaml:"url_path_patterns"`
 	// Methods is a list of HTTP methods this rule applies to (shared). Empty means all methods.
 	Methods []HTTPMethod `yaml:"methods"`
+	// HTTPResponseStatusCode filters this rule to only when the response code matches teh given
```
mmat11:
```suggestion
	// HTTPResponseStatusCode filters this rule to only when the response code matches the given
```

pkg/config/payload_extraction.go
```diff
@@ -325,24 +342,29 @@ type HTTPParsingMatch struct {
...
+	// range
+	HTTPResponseStatusCode *HTTPResponseStatusCodeRange `yaml:"http_response_status_code"`
```
mmat11: nit: this is already part of an "http_"-named struct, so I would omit the `http_` prefix from the field name and yaml tag

devdocs/config/CONFIG.md
```diff
@@ -765,6 +765,23 @@ HTTPParsingMatch defines matching criteria for an HTTP parsing rule. Header rule
...
+| `http_response_status_code` | [`HTTPResponseStatusCodeRange`](#httpresponsestatuscoderange) |  | Filters this rule to only apply when the response status code matches the given range. If unset, the rule applies regardless of status code. (body rules only) |
+
+### HTTPResponseStatusCodeRange
+
+HTTPResponseStatusCodeRange defines an inclusive numeric range to match HTTP response status codes. Both bounds are optional — omitting one leaves that end open.
+
+| Field | Type | Values | Description |
+|---|---|---|---|
+| `greater_equals` | `integer` |  | Matches if the response status code is greater than or equal to this value |
+| `less_equals` | `integer` |  | Matches if the response status code is less than or equal to this value |
```
damemi: maybe just call these `min` and `max`? or is there prior convention on the `_equals` pattern?

pkg/config/payload_extraction.go
hectar-glitches: I will use `MatchDefinition,` i had not noticed it earlier

pkg/config/payload_extraction.go
hectar-glitches: Removed the http_ prefix — field and yaml tag are now response_status_code.

devdocs/config/CONFIG.md
hectar-glitches: Kept _equals since MatchDefinition in pkg/filter/matcher.go already establishes that convention in this codebase.

pkg/config/payload_extraction.go
mmat11: seems like there's an import cycle, we should probably move `MatchDefinition` somewhere else (maybe `config`?)

hectar-glitches: on it

grcevski:
@hectar-glitches if you run `make lint` on the command line you should be able to see the lint errors. I think all other comments are resolved.

matofeder:
@hectar-glitches ping

rafaelroquetto:
@hectar-glitches do you still plan on working on this PR?

hectar-glitches:
yes, the code is done but it failed once i tried to resolve a merge conflict; there is a linting issue with the config file @matofeder, @rafaelroquetto

rafaelroquetto:
@hectar-glitches thanks for the update. I'll check back once the conflicts have been solved.

hectar-glitches:
@rafaelroquetto ! solved the conflicts

mmat11 (approved):
thanks!

PR #2216, opened 2026-06-01
fix: preserve exe_path_regexp during metadata normalization

### LINKED ISSUE #2042: Deprecated exe_path_regexp is bypassed by metadata normalization
## Summary

Regex selector normalization still synthesizes `exe_path="."` for metadata-only selectors without checking whether the deprecated `exe_path_regexp` field is already set. Because executable matching prefers `exe_path` over `exe_path_regexp`, the deprecated path filter is effectively ignored.

## Impact

OBI can instrument more executables than intended inside matching pods or namespaces, leading to telemetry over-collection. The bug depends on a specific deprecated configuration pattern, so `low` severity is appropriate.

## Environment

- First identified in: f42c1a037d3c79815955e7f6e1cf904a26baf3b3

## Evidence

- `pkg/appolly/services/attr_regex.go` still defines deprecated `PathRegexp`.
- `pkg/appolly/discover/matcher.go` still synthesizes `Path="."` in `normalizeRegexCriteria`.
- `pkg/appolly/discover/matcher.go` still prefers `Path` over `PathRegexp` in `matchByExecutable`.

## Steps to reproduce

1. From the repository root on `main`, configure a discovery selector that uses `exe_path_regexp: "^/bin/foo$"` together with Kubernetes metadata criteria but no `exe_path`.
2. Run the matcher against a process in matching metadata scope whose executable is `/bin/notfoo`.
3. Run `go test -tags poc ./pkg/appolly/discover -run TestNormalizeRegexCriteria_ExePathRegexpWithMetadata`.
4. Observe that the test fails because a non-matching executable such as `/bin/notfoo` still matches after normalization, proving `exe_path_regexp` is bypassed when metadata is present.

## Suggested Fix Direction

Treat `PathRegexp` as an existing path constraint during normalization, or stop supporting the deprecated field without an explicit migration path.

## Acceptance Criteria

- Metadata-only normalization does not override deprecated `exe_path_regexp` constraints.
- Executable matching remains scoped to the intended path regex when deprecated config is used.
- Tests cover selectors that use `exe_path_regexp` with metadata.

## Note

MrAlias:
## Summary
- Preserve deprecated `exe_path_regexp` during regex selector metadata normalization.
- Add regression coverage for metadata selectors that use `exe_path_regexp`.

## Root Cause
Metadata-only regex selector normalization synthesized `exe_path="."` without checking whether the deprecated path regexp was already set. Since executable matching prefers `exe_path` over `exe_path_regexp`, that normalized wildcard bypassed the intended regexp constraint.

## Validation
- `go test ./pkg/appolly/discover -run TestCriteriaMatcher_MetadataWithDeprecatedPathRegexp -count=1`
- `go test ./pkg/appolly/discover -count=1`
- `go test -race ./pkg/appolly/discover -count=1`
- `go test ./pkg/appolly/... -count=1`
- `gofmt -l pkg/appolly/discover/matcher.go pkg/appolly/discover/matcher_test.go`
- `make go-mod-tidy`
- `make license-header-check`

Fixes #2042

### DIFF
--- a/pkg/appolly/discover/matcher.go
+++ b/pkg/appolly/discover/matcher.go
@@ -357,7 +357,7 @@ func normalizeRegexCriteria(finderCriteria services.RegexDefinitionCriteria) []s
 	criteria := make([]services.Selector, 0, len(finderCriteria))
 	for i := range finderCriteria {
 		fc := &finderCriteria[i]
-		if !fc.Path.IsSet() && fc.OpenPorts.Len() == 0 && (len(fc.Metadata) > 0 || len(fc.PodLabels) > 0 || len(fc.PodAnnotations) > 0) {
+		if !fc.Path.IsSet() && !fc.PathRegexp.IsSet() && fc.OpenPorts.Len() == 0 && (len(fc.Metadata) > 0 || len(fc.PodLabels) > 0 || len(fc.PodAnnotations) > 0) {
 			// match any executable path
 			if err := fc.Path.UnmarshalText([]byte(".")); err != nil {
 				panic("bug! " + err.Error())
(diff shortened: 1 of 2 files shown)


grcevski (approved):
LGTM!

rafaelroquetto (approved):

PR #2217, opened 2026-06-01
fix(2013): reject mixed glob and numeric attribute filters

### LINKED ISSUE #2013: Mixed glob and numeric conditions silently drop numeric checks
## Summary

`pkg/filter.MatchDefinition` accepts configurations that combine `match` or `not_match` with numeric operators such as `greater_equals` or `less_than`, but `buildMatcher()` only wires numeric comparisons in the numeric-only branch. If a glob clause is present, the numeric clauses are silently ignored.

## Impact

This makes attribute filters more permissive than the operator configured. In practice that can allow extra telemetry to be exported when the operator expected numeric thresholds to exclude it. The impact is limited to data-minimization and policy-enforcement failure in configured telemetry export, so `low` severity is appropriate.

## Environment

- First identified in: 5d6239f252a1e0670e0277bff7500dd538845dd3

## Evidence

- [pkg/filter/matcher.go](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/pkg/filter/matcher.go#L90-L99) validates mixed glob-plus-numeric definitions without error.
- [pkg/filter/attribute.go](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/pkg/filter/attribute.go#L106-L140) prioritizes `match` and `not_match` branches and only sets numeric comparison fields in the `default` branch.
- [pkg/filter/matcher.go](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/pkg/filter/matcher.go#L33-L71) only runs numeric comparisons if those fields were populated.

## Steps to reproduce

1. From the repository root on `main`, create `pkg/filter/mixed_matcher_poc_test.go` with these exact contents:
   ```go
   package filter
   import (
   	"testing"
   	"go.opentelemetry.io/obi/pkg/export/attributes"
   )
   type dummyRecord struct{ val string }
   func dummyGetters(name attr.Name) (attributes.Getter[dummyRecord, string], bool) {
   	if name == attr.HTTPResponseStatusCode { return func(r dummyRecord) string { return r.val }, true }
   	return nil, false
   }
   func TestMixedGlobAndNumericIgnored(t *testing.T) {
   	gt := 100
   	def := MatchDefinition{Match: "*", GreaterThan: &gt}
   	matcher, err := buildMatcher[dummyRecord](dummyGetters, attr.HTTPResponseStatusCode, &def)
   	if err != nil { t.Fatalf("buildMatcher error: %v", err) }
   	if matcher.Matches(dummyRecord{val: "50"}) { t.Fatalf("expected numeric comparison to reject 50, but match was true (numeric ignored)") }
   }
   ```
3. Run `GOTOOLCHAIN=auto go test ./pkg/filter -run TestMixedGlobAndNumericIgnored -count=1`.
4. Observe that the test fails because the value `50` still matches when both `match: "*"` and `greater_equals: 500` are set, proving the numeric comparison is ignored.

## Suggested Fix Direction

Reject mixed glob-plus-numeric matcher definitions during validation, or explicitly define and implement combined semantics instead of silently dropping one side.

## Acceptance Criteria

- Mixed `match`/`not_match` plus numeric comparison configurations are either rejected or applied consistently.
- Tests cover the mixed-config case so the silent-drop behavior cannot regress back in.

## Note

MrAlias:
## Summary

Reject attribute filter configurations that combine `match`/`not_match` glob clauses with numeric comparisons.

## Root Cause

`MatchDefinition` validation previously allowed mixed glob-plus-numeric definitions, but `buildMatcher` only populated numeric comparison fields in the numeric-only branch. When a glob clause was present, numeric comparisons were silently dropped.

## Impact

Mixed configurations now fail validation instead of behaving more permissively than configured. This makes the invalid configuration explicit and prevents accidental export of telemetry that numeric thresholds were expected to exclude.

Fixes #2013.

## Validation

- `go test ./pkg/filter`
- `gofmt -l pkg/filter/matcher.go pkg/filter/matcher_test.go pkg/filter/attribute_test.go`

### DIFF
--- a/pkg/filter/attribute_test.go
+++ b/pkg/filter/attribute_test.go
@@ -586,6 +586,8 @@ func TestAttributeFilter_VerificationError(t *testing.T) {
 		{"obi.ip": MatchDefinition{}},
 		// valid attribute with double match definition
 		{"obi.ip": MatchDefinition{Match: "foo", NotMatch: "foo"}},
+		// valid attribute with invalid mixed glob and numeric comparisons
+		{"http.response.status_code": MatchDefinition{Match: "2*", GreaterEquals: intPtr(200)}},
 	}
 	for _, tc := range testCases {
 		t.Run(fmt.Sprintf("%v", tc), func(t *testing.T) {
--- a/pkg/filter/matcher.go
+++ b/pkg/filter/matcher.go
@@ -89,13 +89,17 @@ type MatchDefinition struct {
 
 func (md *MatchDefinition) Validate() error {
 	hasGlob := md.Match != "" || md.NotMatch != ""
-	hasNumeric := md.GreaterThan != nil || md.LessThan != nil || md.Equals != nil || md.NotEquals != nil || md.GreaterEquals != nil || md.LessEquals != nil
+	hasNumeric := md.GreaterThan != nil || md.LessThan != nil || md.Equals != nil ||
+		md.NotEquals != nil || md.GreaterEquals != nil || md.LessEquals != nil
 
 	if !hasGlob && !hasNumeric {
 		return errors.New("attribute must include a match/not_match clause or numeric comparison")
 	}
 	if md.Match != "" && md.NotMatch != "" {
 		return errors.New("attribute can't include bot match or not_match clauses")
+	}
+	if hasGlob && hasNumeric {
+		return errors.New("attribute can't combine match/not_match clauses with numeric comparisons")
 	}
 	return nil
 }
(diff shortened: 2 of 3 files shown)


grcevski (approved):
LGTM!

mariomac (approved):

PR #2218, opened 2026-06-01
fix(java,bpf): limit SSL receive packets to bytes read

### LINKED ISSUE #2019: SSLSocket read(byte[]) copies the full buffer instead of bytesRead
## Summary

The Java `SSLSocketStreamInst` advice for `InputStream.read(byte[])` sizes and writes the native IOCTL packet using `b.length` rather than the actual return value from `read()`. This causes the entire array to be forwarded even when only part of it contains fresh data.

## Impact

Unused trailing bytes from the application buffer can be exported through OBI's telemetry path. The exposure is limited to buffer-resident data that becomes visible only to telemetry consumers, so `low` severity is appropriate.

## Environment

- First identified in: 61113471f1430b2f0230cce0d2c983f2262a8b3a

## Evidence

- [pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/instrumentations/SSLSocketStreamInst.java](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/instrumentations/SSLSocketStreamInst.java#L107-L109) allocates and copies using `b.length`.
- [pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/instrumentations/SSLSocketStreamInst.java](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/instrumentations/SSLSocketStreamInst.java#L144-L147) shows the offset-based path correctly using `bytesRead`.
- [pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/ebpf/IOCTLPacket.java](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/ebpf/IOCTLPacket.java#L59-L60) confirms the no-offset overload copies `buf.length` bytes.

## Steps to reproduce

1. From the repository root on `main`, build the Java agent classes and JNI support from `pkg/internal/java/agent`.
2. Run a Java TLS client or server under the agent and arrange for `InputStream.read(byte[])` to fill only the first byte of a larger preinitialized buffer.
3. Run a Java program that performs a partial `read(byte[])` on an instrumented TLS socket while the agent is active.
4. Observe on the native side or in debug output that the captured payload length is the full buffer size and includes bytes beyond the actual `len`, proving stale data is forwarded.

## Suggested Fix Direction

Use the actual bytes-read count (`len`) when sizing the packet prefix and copying buffer contents for `read(byte[])`, mirroring the offset-based advice.

## Acceptance Criteria

- `read(byte[])` only forwards the bytes actually read.
- SSL receive instrumentation no longer copies trailing stale buffer contents.
- Tests cover partial-read behavior for the no-offset advice path.

## Note

MrAlias:
## Summary

- Use the actual `read(byte[])` return value when sizing and copying `SSLSocketStreamInst` SSL receive packets.
- Apply the same partial-read fix to the `ProxyInputStream.read(byte[])` SSL receive path.
- Bound BPF protocol sniff-buffer reads to the reported payload length so short Java IOCTL packets do not read beyond the copied payload.
- Add focused Java tests for partial-buffer receive packet construction.

## Why

`InputStream.read(byte[])` can return fewer bytes than the destination buffer length. The previous Java agent code used the full buffer length when building native IOCTL packets, which could cause stale trailing bytes from the application buffer to be treated as payload by OBI. BPF has only the packet length supplied by the Java agent, so it cannot recover the true bytes-read count downstream.

The BPF sniff-buffer change keeps protocol detection bounded to the payload length that Java reports, avoiding inspection past short Java IOCTL payloads.

Fixes #2019

## Validation

- `make java-clean`
- `make java-verify`
- `make generate`
- `make compile`
- `git diff --check`

`make clang-tidy` was also run, but the repository-wide target currently fails on pre-existing unrelated diagnostics in files such as `bpf/common/common.c`, `bpf/common/sockaddr.h`, and `bpf/tpinjector/tpinjector.c`; it processed the touched `bpf/generictracer/k_tracer_defs.h` without reporting a diagnostic there.

### DIFF
--- a/bpf/generictracer/k_tracer_defs.h
+++ b/bpf/generictracer/k_tracer_defs.h
@@ -62,7 +62,7 @@ static __always_inline call_protocol_args_t *make_protocol_args(const pid_connec
     __builtin_memset(args->small_buf, 0, sizeof(args->small_buf));
     if (bytes_len > 0) {
         u32 small_buf_len = (u32)bytes_len;
-        bpf_clamp_umax(small_buf_len, MIN_HTTP2_SIZE);
+        bpf_clamp_umax(small_buf_len, sizeof(args->small_buf));
         bpf_probe_read(args->small_buf, small_buf_len, (void *)args->u_buf);
     }
 
--- a/bpf/generictracer/k_tracer_defs.h
+++ b/bpf/generictracer/k_tracer_defs.h
@@ -44,6 +44,10 @@ static __always_inline call_protocol_args_t *make_protocol_args(const pid_connec
                                                                 u8 ssl,
                                                                 u8 direction,
                                                                 u16 orig_dport) {
+    if (bytes_len <= 0) {
+        return 0;
+    }
+
     call_protocol_args_t *args = protocol_args();
     if (!args) {
         return 0;
@@ -60,11 +64,9 @@ static __always_inline call_protocol_args_t *make_protocol_args(const pid_connec
 
     args->pid_conn = *info;
     __builtin_memset(args->small_buf, 0, sizeof(args->small_buf));
-    if (bytes_len > 0) {
-        u32 small_buf_len = (u32)bytes_len;
-        bpf_clamp_umax(small_buf_len, sizeof(args->small_buf));
-        bpf_probe_read(args->small_buf, small_buf_len, (void *)args->u_buf);
-    }
+    u32 small_buf_len = (u32)bytes_len;
+    bpf_clamp_umax(small_buf_len, sizeof(args->small_buf));
+    bpf_probe_read(args->small_buf, small_buf_len, (void *)args->u_buf);
 
     return args;
 }
--- a/bpf/generictracer/k_tracer_defs.h
+++ b/bpf/generictracer/k_tracer_defs.h
@@ -59,7 +59,12 @@ static __always_inline call_protocol_args_t *make_protocol_args(const pid_connec
     args->protocol_type = protocol_type_for_conn_info(info);
 
     args->pid_conn = *info;
-    bpf_probe_read(args->small_buf, MIN_HTTP2_SIZE, (void *)args->u_buf);
+    __builtin_memset(args->small_buf, 0, sizeof(args->small_buf));
+    if (bytes_len > 0) {
+        u32 small_buf_len = (u32)bytes_len;
+        bpf_clamp_umax(small_buf_len, MIN_HTTP2_SIZE);
+        bpf_probe_read(args->small_buf, small_buf_len, (void *)args->u_buf);
+    }
 
     return args;
 }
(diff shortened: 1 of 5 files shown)


MrAlias:
Additional fixes:

- `fix(java): limit proxy SSL reads to bytes read`
  - Applies the same `bytesRead` sizing/copying fix to `ProxyInputStream.read(byte[])`.
  - Adds `ProxyInputStreamTest` coverage for partial-buffer receive packets.
- `fix(bpf): bound protocol sniff buffer reads`
  - Clears `small_buf` and only reads up to `min(bytes_len, MIN_HTTP2_SIZE)` before protocol sniffing, so short Java IOCTL packets do not cause BPF to inspect bytes beyond the reported payload.

Validation run:

- `make java-clean`
- `make java-verify`
- `make generate`
- `make compile`

bpf/generictracer/k_tracer_defs.h
```diff
@@ -59,7 +59,12 @@ static __always_inline call_protocol_args_t *make_protocol_args(const pid_connec
     args->protocol_type = protocol_type_for_conn_info(info);
 
     args->pid_conn = *info;
-    bpf_probe_read(args->small_buf, MIN_HTTP2_SIZE, (void *)args->u_buf);
+    __builtin_memset(args->small_buf, 0, sizeof(args->small_buf));
```
grcevski, line 66: should we just return and bail if (bytes_len <=0) ? This way we can avoid the unnecessary memset.

grcevski (approved):
LGTM!

One minor comment about avoiding the memset to 0 when the len <= 0.

bpf/generictracer/k_tracer_defs.h
```diff
@@ -59,7 +59,12 @@ static __always_inline call_protocol_args_t *make_protocol_args(const pid_connec
...
+    if (bytes_len > 0) {
+        u32 small_buf_len = (u32)bytes_len;
+        bpf_clamp_umax(small_buf_len, MIN_HTTP2_SIZE);
```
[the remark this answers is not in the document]
MrAlias: Addressed in ef31c6fac by clamping `small_buf_len` with `sizeof(args->small_buf)`, keeping the read bound tied to the actual destination buffer size.

bpf/generictracer/k_tracer_defs.h
MrAlias: Addressed in 494d89832 by returning from `make_protocol_args` when `bytes_len <= 0`, before allocating protocol args or clearing `small_buf`.

PR #2224, opened 2026-06-02
Add UDS support for the OTLP exporter and the health check endpoint

rafaelroquetto:
## Summary

Allow unix socket endpoints to be specified as valid OTLP export URLs. The particular use case for this is to be able to drive OBI as a subprocess and communicate with it without paying the cost of a round trip inside the kernel TCP stack when operating in loopback, but that's just one use-case and is not limited to it.

## Validation


mariomac (commented):
Amazing feature! I'm only wondering about the need of implement the health check via unix socket.

mariomac (approved):

PR #2226, opened 2026-06-02
Fix stale Kafka traceparent contaminating later trace context (#2046)

### LINKED ISSUE #2046: Stale Kafka goroutine traceparent contaminates later trace context
## Summary

The Go Kafka instrumentation stores a traceparent in `produce_traceparents_by_goroutine` when a produce operation begins, but never removes that goroutine-keyed entry when the operation completes. Later `runtime.casgstatus` events can still find the stale entry and write it into the shared `traces_ctx_v1` map, causing unrelated work on the same goroutine to inherit an old trace ID/span ID.

## Impact

This is a real telemetry integrity bug in OBI product code. It can misattribute spans or log enrichment to the wrong request and can leak trace context across requests handled by the same instrumented process. The leaked data is limited to trace IDs and span IDs, so `low` severity is appropriate.

## Environment

- First identified in: 4353b7e5ac081988088dd5370bd62557e6fdc183

## Evidence

- [bpf/gotracer/go_kafka_go.c](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/bpf/gotracer/go_kafka_go.c#L50-L51) inserts `produce_traceparents_by_goroutine`.
- [bpf/gotracer/go_kafka_go.c](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/bpf/gotracer/go_kafka_go.c#L101) deletes only `produce_traceparents`.
- [bpf/gotracer/go_kafka_go.c](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/bpf/gotracer/go_kafka_go.c#L214-L217) deletes `ongoing_produce_messages` and `produce_requests`, but not `produce_traceparents_by_goroutine`.
- [bpf/gotracer/go_runtime.c](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/bpf/gotracer/go_runtime.c#L276-L279) looks up `produce_traceparents_by_goroutine` and calls `obi_ctx__set_()`.
- [bpf/shared/obi_ctx.h](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/bpf/shared/obi_ctx.h#L19-L40) shows that `obi_ctx__set_()` writes into the pinned shared `traces_ctx_v1` map.

## Steps to reproduce

1. From the repository root on `main`, build and load the Kafka Go instrumentation with `make generate && make compile`, then start the built agent with the Go Kafka tracer enabled and enough privileges to load BPF.
2. Run a Go Kafka producer workload once so `uprobe/writer_write_messages` inserts a traceparent into `produce_traceparents_by_goroutine`.
3. After the produce finishes, inspect the pinned BPF maps with `bpftool map dump` and confirm the goroutine key is still present in `produce_traceparents_by_goroutine`.
4. Reuse the same goroutine for unrelated work and observe that `runtime.casgstatus` rehydrates the stale traceparent into the shared trace context, contaminating subsequent activity.

## Suggested Fix Direction

Delete `produce_traceparents_by_goroutine` on all Kafka produce completion and abandonment paths, and verify there is no path where `runtime.casgstatus` can observe a stale goroutine-keyed traceparent after the request lifecycle ends.

## Acceptance Criteria

- Kafka produce completion removes the goroutine-keyed traceparent entry.
- Aborted or partially completed Kafka produce paths also clean up the goroutine-keyed entry.
- Shared trace context is not reinstalled from stale Kafka state after the produce lifecycle ends.

## Note

lucacavenaghi97:
Fixes #2046.

The Go Kafka instrumentation stores a traceparent in
`produce_traceparents_by_goroutine` on `WriteMessages` but never removes it, so
after the produce ends `runtime.casgstatus` can still reinstall it into the
shared `traces_ctx_v1` and unrelated work on the same goroutine inherits the
old trace context.

The change adds a return uprobe on `(*Writer).WriteMessages` that deletes the
goroutine-keyed entry and clears the per-thread context on return.

A note on the approach: the issue suggested cleaning up on the produce/roundTrip
paths. I tried a single return probe instead, since as far as I could tell those
paths can run on a different goroutine than the `WriteMessages` caller, which
would make a goroutine-keyed delete there miss the entry (including the async
writer). Deleting on the `WriteMessages` return also lines up with the
set-on-entry/delete-on-return pattern the other Go instrumentations use.

I also added an integration test that produces and then logs from a worker
goroutine, checking via the log enricher that the log line does not pick up the
produce trace_id.

### DIFF
--- a/bpf/gotracer/go_kafka_go.c
+++ b/bpf/gotracer/go_kafka_go.c
@@ -57,6 +57,23 @@ int obi_uprobe_writer_write_messages(struct pt_regs *ctx) {
     return 0;
 }
 
+SEC("uprobe/writer_write_messages_ret")
+int obi_uprobe_writer_write_messages_ret(struct pt_regs *ctx) {
+    void *goroutine_addr = (void *)GOROUTINE_PTR(ctx);
+    bpf_dbg_printk("=== uprobe/writer_write_messages_ret ===");
+    bpf_dbg_printk("goroutine_addr=%llx", goroutine_addr);
+
+    go_addr_key_t g_key = {};
+    go_addr_key_from_id(&g_key, goroutine_addr);
+
+    // Drop the goroutine-keyed traceparent so casgstatus can't re-install this
+    // produce's context after the request ends (issue #2046).
+    bpf_map_delete_elem(&produce_traceparents_by_goroutine, &g_key);
+    obi_ctx__del(bpf_get_current_pid_tgid());
+
+    return 0;
+}
+
 SEC("uprobe/writer_produce")
 int obi_uprobe_writer_produce(struct pt_regs *ctx) {
     void *goroutine_addr = (void *)GOROUTINE_PTR(ctx);
--- a/internal/test/integration/components/gokafka-seg/main.go
+++ b/internal/test/integration/components/gokafka-seg/main.go
@@ -51,6 +51,41 @@ func producerHandlerWithTopic(kafkaWriter *kafka.Writer, topic string) func(http
 	})
 }
 
+// Marker log lines for the issue #2046 traceparent-contamination regression test.
+const (
+	kafkaTPPositiveControlMsg = "kafka-tp-positive-control"
+	kafkaTPAfterProduceMsg    = "kafka-tp-after-produce"
+)
+
+func logJSON(message string) {
+	fmt.Printf(`{"message":"%s","level":"INFO"}`+"\n", message)
+}
+
+// traceparentProbeHandler logs once from the request goroutine (positive control),
+// then runs a Kafka produce followed by a log line on a worker goroutine that has
+// no request context of its own (issue #2046).
+func traceparentProbeHandler(kafkaWriter *kafka.Writer) func(http.ResponseWriter, *http.Request) {
+	return http.HandlerFunc(func(wrt http.ResponseWriter, _ *http.Request) {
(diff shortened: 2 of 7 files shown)


grcevski:
The log enricher failures look real to me, I'll re-run the tests. But I wonder if the actual issue is a real problem. Yes we will temporary have stale data for the goroutine, but as soon as another relevant operation starts it will associate the goroutine with the new id. I guess the log operations in between may get wrong trace, span id, rather than empty one?

lucacavenaghi97:
Yeah, a wrong id rather than empty. It doesn't always get overwritten though: if the goroutine only logs after the produce (background publisher, pooled worker) nothing replaces obi_ctx and casgstatus keeps re-installing the produce's tp. The other goroutine-keyed tracers (grpc/http/sql/redis/mongo) already delete their entry on return, produce is the only one that doesn't, so I'd align it with them

grcevski (approved):
LGTM! Thanks for fixing this!

PR #2232, opened 2026-06-02
fix: clear stale traceparent parent ids

### LINKED ISSUE #2020: Reused traceparent path leaves parent_id stale
## Summary

When OBI parses an existing `traceparent` header in the sock_msg tpinjector path, it reuses a scratch `tp_info_pid_t` object and stores it without clearing `parent_id` first. If `assign_parent_tp()` does not set a parent, the old `parent_id` can persist.

## Impact

This can leak a stale parent span ID into the current request's telemetry and corrupt trace lineage. The impact is limited to observability metadata, so `low` severity is appropriate.

## Environment

- First identified in: 521cdce4ccb41ef29ccc323f0533b5894248d146

## Evidence

- [bpf/tpinjector/tpinjector.c](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/bpf/tpinjector/tpinjector.c#L804-L823) does not clear `parent_id` on early returns.
- [bpf/tpinjector/tpinjector.c](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/bpf/tpinjector/tpinjector.c#L898-L912) stores the parsed trace info after calling `assign_parent_tp()` without prior zeroing.
- [bpf/tpinjector/tpinjector.c](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/bpf/tpinjector/tpinjector.c#L279-L280) shows the separate new-trace path explicitly zeroing `parent_id`.
- [bpf/common/http_info.h](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/bpf/common/http_info.h#L15-L26) embeds `tp_info_t` in emitted HTTP info.

## Steps to reproduce

1. From the repository root on `main`, run `GOTOOLCHAIN=auto go test ./internal/test/integration -run TestTraceparentExtraction -tags=integration` to start the traceparent extraction integration stack.
2. Send one HTTP request with a `Traceparent` header that establishes a parent span relationship, then send a second request with a different `Traceparent` header in a path where no matching parent trace exists in the current process.
3. Collect the exported spans from the integration backend or inspect the ring-buffer output for the emitted `parent_id` values.
4. Observe that the second request can inherit the earlier `parent_id` even though no matching parent exists for it, proving the stale scratch-slot reuse.

## Suggested Fix Direction

Clear `parent_id` before parsing or before calling `assign_parent_tp()` in the reused-header path so the stored trace info is always fully initialized.

## Acceptance Criteria

- The reused traceparent parse path always initializes `parent_id`.
- Stale parent span IDs cannot survive from prior scratch-buffer users.
- Tests cover the no-parent and mismatched-parent cases.

## Note

MrAlias:
## Summary

Fixes stale `parent_id` propagation in existing-traceparent paths. The original sock_msg tpinjector path now clears `parent_id` before parent assignment so reused scratch state cannot leak a previous parent when no matching parent trace exists.

While adding behavioral regression coverage for the review feedback, the test exposed the same stale-parent class in the generic HTTP client existing-traceparent path: a parsed client `traceparent` can switch trace IDs after generated in-process parent state already exists. That path now clears `parent_id` when the parsed client trace ID differs from the previous in-process trace ID.

Resolves #2020.

## Changes

- Clear `tp_p->tp.parent_id` before calling `assign_parent_tp()` in `obi_packet_extender_find_existing_tp()`.
- Clear the generic HTTP client `parent_id` when an existing `traceparent` changes the trace ID away from the current in-process trace.
- Extend `TestTraceparentExtraction` with parent relationship assertions for the reused-traceparent path, including the mismatched-parent case and the matching downstream parent case.

## Validation

- `make generate/all`
- `go test -count=1 ./pkg/internal/ebpf/tpinjector`
- `go test -count=1 -tags=bpf_verifier_tests ./pkg/internal/ebpf/verifier`
- `go test -v -count=1 -run TestTraceparentExtraction -tags=integration -timeout 20m ./internal/test/integration`
- True-positive check: temporarily removing the generic-tracer fix makes `TestTraceparentExtraction/with_traceparent` fail on the new parent assertion, then it passes again with the fix restored.

### DIFF
--- a/bpf/generictracer/protocol_http.h
+++ b/bpf/generictracer/protocol_http.h
@@ -495,11 +495,22 @@ __obi_continue_protocol_http_tp(struct pt_regs *ctx,
                 unsigned char *t_id = extract_trace_id(res);
                 unsigned char *s_id = extract_span_id(res);
                 unsigned char *f_id = extract_flags(res);
+                const bool is_client = meta && meta->type == EVENT_HTTP_CLIENT;
+                unsigned char previous_trace_id[TRACE_ID_SIZE_BYTES] = {};
+
+                if (is_client) {
+                    __builtin_memcpy(
+                        previous_trace_id, tp_p->tp.trace_id, sizeof(previous_trace_id));
+                }
 
                 decode_hex(tp_p->tp.trace_id, t_id, TRACE_ID_CHAR_LEN);
                 decode_hex((unsigned char *)&tp_p->tp.flags, f_id, FLAGS_CHAR_LEN);
                 if (meta && meta->type != EVENT_HTTP_CLIENT) {
                     decode_hex(tp_p->tp.parent_id, s_id, SPAN_ID_CHAR_LEN);
+                } else if (is_client && bpf_memcmp(previous_trace_id,
+                                                   tp_p->tp.trace_id,
+                                                   sizeof(previous_trace_id)) != 0) {
+                    __builtin_memset(tp_p->tp.parent_id, 0, sizeof(tp_p->tp.parent_id));
                 }
 
                 if (g_bpf_debug) {
--- a/bpf/generictracer/protocol_http.h
+++ b/bpf/generictracer/protocol_http.h
@@ -17,6 +17,7 @@
 #include <common/lw_thread.h>
 #include <common/ringbuf.h>
 #include <common/runtime.h>
+#include <common/scratch_mem.h>
 #include <common/trace_helpers.h>
 #include <common/trace_lifecycle.h>
 #include <common/trace_parent.h>
@@ -36,6 +37,8 @@
 #include <maps/tp_char_buf_mem.h>
 
 volatile const u32 high_request_volume;
+
+SCRATCH_MEM_SIZED(http_previous_trace_id, TRACE_ID_SIZE_BYTES);
 
 // empty_http_info zeroes and return the unique percpu copy in the map
 // this function assumes that a given thread is not trying to use many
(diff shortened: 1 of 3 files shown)


rafaelroquetto (approved):
Nice catch!

grcevski (approved):
LGTM!

bpf/tpinjector/tpinjector.c
```diff
@@ -1084,6 +1084,7 @@ int obi_packet_extender_find_existing_tp(struct sk_msg_md *msg) {
             // if we got to this point, we managed to parse a valid
             // 'Traceparent: ...' header that we can utilise
 
+            bpf_memset(tp_p->tp.parent_id, 0, sizeof(tp_p->tp.parent_id));
             assign_parent_tp(t_ctx, &tp_p->tp, span_id);
```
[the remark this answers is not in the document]
MrAlias: Addressed in `40dbef3e2`.

I added behavioral coverage in `TestTraceparentExtraction` for the reused-traceparent parent relationship: `tpclient-a` now asserts that a client span using the static incoming `traceparent` does not inherit a generated in-process parent from a different trace, while `tpclient-b` still asserts the matching downstream parent relationship is preserved.

While adding that test, it exposed the same stale-parent class in the generic HTTP client existing-`traceparent` path: parsing the header can switch trace IDs after generated parent state already exists. I increased the scope slightly to clear `parent_id` there when the parsed client trace ID differs from the previous in-process trace ID. I also updated the PR title/body to make that expanded scope explicit.

Validation included the full `TestTraceparentExtraction` integration test, and I confirmed the new assertion is a true positive by temporarily removing the generic-tracer fix and seeing `TestTraceparentExtraction/with_traceparent` fail on the new parent assertion.

MrAlias:
@rafaelroquetto PTAL at https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2232/commits/40dbef3e2848dec5904b36f187056a5b5f4db493. In addressing the copilot feedback, I uncovered that stale parent span ID use as well.

bpf/generictracer/protocol_http.h
```diff
@@ -495,11 +495,22 @@ __obi_continue_protocol_http_tp(struct pt_regs *ctx,
                 unsigned char *t_id = extract_trace_id(res);
                 unsigned char *s_id = extract_span_id(res);
                 unsigned char *f_id = extract_flags(res);
+                const bool is_client = meta && meta->type == EVENT_HTTP_CLIENT;
+                unsigned char previous_trace_id[TRACE_ID_SIZE_BYTES] = {};
```
rafaelroquetto: We should use SCRATCH_MEM here, this is too heavy on the stack.

Then, I would gate the memcpy only if `tp_p->tp.trace_id` is populated, so that we don't do it for every call.

rafaelroquetto (changes requested):
Marking as request changes just to prevent it from being accidentally merged since it's been approved. Good catch with the protocol_http.h changes as well - we should just avoid the previous tp on the stack because we are on the verge or running out of stack (one of the reasons we have so many tailcalls)

bpf/generictracer/protocol_http.h
MrAlias: Updated. PTAL at https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2232/commits/a4696cc7073b25ac63ae819eb9e39a62dbe34fa5

rafaelroquetto (approved):
Awesome, thanks for the changes!

PR #2233, opened 2026-06-02
fix(2017): bound Go readMimeHeader traceparent scan

### LINKED ISSUE #2017: Go readMimeHeader can reuse stale bytes as traceparent
## Summary

The Go net/http `readMimeHeader` uprobe copies only the current header length into the shared traceparent scratch buffer and then scans that buffer for the `traceparent:` prefix without clearing old bytes first.

## Impact

This can cause stale trace IDs/span IDs from a prior request to be reused for the current request, producing cross-request telemetry misattribution and limited trace-context disclosure. The impact is limited to observability data, so `low` severity is appropriate.

## Environment

- First identified in: a7f8ea757f09a1eecd834566557a35fa9b20a11e

## Evidence

- [bpf/gotracer/go_nethttp.c](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/bpf/gotracer/go_nethttp.c#L442-L449) reads `len` bytes into `tp_char_buf_mem()` without clearing it.
- [bpf/common/trace_util.h](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/bpf/common/trace_util.h#L83-L112) scans for the traceparent prefix using fixed-prefix checks that can observe stale trailing bytes.
- [bpf/maps/tp_char_buf_mem.h](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/bpf/maps/tp_char_buf_mem.h#L11) defines the shared scratch buffer.

## Steps to reproduce

1. From the repository root on `main`, run this inline Python model of the traceparent scan:
   ```bash
   python3 - <<'PY'
   TRACE_BUF_SIZE = 1024
   TRACE_PARENT_HEADER_LEN = 68
   def is_traceparent(p):
       return (p[0] in (ord('T'), ord('t')) and p[1:5] == b'race' and p[5] in (ord('p'), ord('P')) and p[6:11] == b'arent' and p[11:13] == b': ')
   def scan(buf, buf_len):
       for index in range(buf_len):
           if index >= TRACE_BUF_SIZE - TRACE_PARENT_HEADER_LEN:
               break
           if is_traceparent(buf[index:index+13]):
               return index
       return None
   old = bytearray(b"xtraceparent: 00-0123456789abcdef0123456789abcdef-0123456789abcdef-01\r\n")
   buf = bytearray(b"\x00" * TRACE_BUF_SIZE)
   buf[:len(old)] = old
   new = b"xtrace"
   buf[:len(new)] = new
   pos = scan(buf, len(new))
   print(f"new_len={len(new)} pos={pos}")
   print("is_traceparent_at_1=", is_traceparent(buf[1:14]))
   PY
   ```
3. Let the script simulate a short header overwrite on the shared traceparent scratch buffer.
4. Observe that the output still reports a valid traceparent match from stale bytes beyond the new header length, demonstrating the false-positive reuse.

## Suggested Fix Direction

Clear the scratch buffer before partial reads in the Go HTTP path, or hard-bound the scanner so it never reads beyond the fresh payload length.

## Acceptance Criteria

- The Go `readMimeHeader` path cannot match a traceparent using stale bytes from a previous request.
- Traceparent scanning respects the actual copied buffer length.
- Tests cover short-read and stale-buffer cases.

## Note

MrAlias:
## Summary

- Bound traceparent scanning to the actual number of bytes copied for the current Go `readMimeHeader` invocation.
- Applied the same fresh-length bound to the legacy traceparent scanner path.
- Added stale scratch-buffer tests covering short fresh reads that previously could be completed by bytes from an earlier request.

Fixes #2017.

## Validation

- `make generate`
- `go test -vet=off -run 'TestTraceparentScanBoundsFreshHeaderLength|TestReadMimeHeaderBPFUsesLengthBoundedScan' ./pkg/internal/ebpf/gotracer`

### DIFF
--- a/bpf/common/algorithm.h
+++ b/bpf/common/algorithm.h
@@ -55,11 +55,11 @@ static __always_inline unsigned long min_ulong(unsigned long a, unsigned long b)
         u8: min_u8,                                                                                \
         u16: min_u16,                                                                              \
         u32: min_u32,                                                                              \
-        u64: min_u64,                                                                              \
+        unsigned long long: min_u64,                                                               \
         s8: min_s8,                                                                                \
         s16: min_s16,                                                                              \
         s32: min_s32,                                                                              \
-        s64: min_s64,                                                                              \
+        long long: min_s64,                                                                        \
         long: min_long,                                                                            \
         unsigned long: min_ulong)(a, b)
 
@@ -112,10 +112,10 @@ static __always_inline unsigned long max_ulong(unsigned long a, unsigned long b)
         u8: max_u8,                                                                                \
         u16: max_u16,                                                                              \
         u32: max_u32,                                                                              \
-        u64: max_u64,                                                                              \
+        unsigned long long: max_u64,                                                               \
         s8: max_s8,                                                                                \
         s16: max_s16,                                                                              \
         s32: max_s32,                                                                              \
-        s64: max_s64,                                                                              \
+        long long: max_s64,                                                                        \
         long: max_long,                                                                            \
         unsigned long: max_ulong)(a, b)
--- a/bpf/common/trace_util.h
+++ b/bpf/common/trace_util.h
@@ -6,6 +6,7 @@
 #include <bpfcore/vmlinux.h>
 #include <bpfcore/bpf_helpers.h>
 
+#include <common/algorithm.h>
 #include <common/globals.h>
 #include <common/http_buf_size.h>
 
@@ -106,12 +107,7 @@ static __always_inline u32 traceparent_scan_loop_count(const u16 buf_len) {
         return 0;
     }
 
-    u32 nr_loops = (u32)buf_len - TRACE_PARENT_HEADER_LEN + 1;
-    if (nr_loops > k_tp_max_scan_loops) {
-        return k_tp_max_scan_loops;
-    }
-
-    return nr_loops;
+    return min((u32)buf_len - TRACE_PARENT_HEADER_LEN + 1, k_tp_max_scan_loops);
 }
 
 static __always_inline unsigned char *bpf_strstr_tp_loop(unsigned char *buf, const u16 buf_len) {
@@ -120,6 +116,7 @@ static __always_inline unsigned char *bpf_strstr_tp_loop(unsigned char *buf, con
     }
 
     const u32 nr_loops = traceparent_scan_loop_count(buf_len);
+
     if (nr_loops == 0) {
         return NULL;
     }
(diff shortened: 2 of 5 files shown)


mariomac (approved):
🚀

MrAlias:
Looks like that debug logging (https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2233/commits/97baeee0538b0bf4c0eb049a5baeda6cf932ec02) tripped the verifier complexity into failing.

I'm working on a fix.

rafaelroquetto:
@MrAlias keep me posted as I am also debugging a similar issue in the #2172 PR - and I will do the same.

MrAlias:
Thank you @rafaelroquetto for all the help!

bpf/common/algorithm.h
```diff
@@ -55,11 +55,11 @@ static __always_inline unsigned long min_ulong(unsigned long a, unsigned long b)
         u8: min_u8,                                                                                \
         u16: min_u16,                                                                              \
         u32: min_u32,                                                                              \
-        u64: min_u64,                                                                              \
+        unsigned long long: min_u64,                                                               \
```
mmat11, line 58: question: why these changes?

MrAlias: With the caveat that @rafaelroquetto wrote this so he can authoritatively say, but from my understanding this is for C `_Generic` compatibility. In the BPF vmlinux headers, `u64`/`s64` are `unsigned long long`/`long long`, so the BPF-side dispatch remains the same. In the host test shim, though, `u64` comes from `uint64_t`, which can be `unsigned long` on LP64 hosts. Since this generic list also has explicit `unsigned long`/`long` cases, using `u64`/`s64` directly can create duplicate compatible generic associations. Spelling the 64-bit BPF cases as `unsigned long long`/`long long` avoids that compile-time collision without changing min/max behavior.

rafaelroquetto: That is exactly it!

rafaelroquetto (approved):
I will leave for @mmat11 to give the final stamp as I am also a contributor to this PR

bpf/common/algorithm.h
mmat11: thanks!

mmat11 (approved):
lgtm!

PR #2237, opened 2026-06-03
fix(2029): harden SSL payload guards

### LINKED ISSUE #2029: Failed SSL reads still flow into protocol parsing

MrAlias:
## Summary

- reject non-positive SSL payload lengths in `handle_ssl_buf` before fake connection creation, delayed trace cleanup, or protocol parsing
- keep SSL thread-map cleanup before returning on failed or EOF reads
- use actual SSL write results for `SSL_write` and `SSL_write_ex`, avoiding failed writes and partial writes being parsed as full requested payloads
- replace the mirrored SSL guard test with a focused C test that includes the real `generictracer/ssl_defs.h` helper through test-local stubs

Fixes #2029.

## Validation

- `clang-format --dry-run --Werror bpf/generictracer/libssl.c bpf/generictracer/ssl_defs.h bpf/tests/bpf_ssl_read_guard.c bpf/tests/common/connection_info.h bpf/tests/common/trace_key.h bpf/tests/common/trace_lifecycle.h bpf/tests/generictracer/k_tracer_defs.h bpf/tests/generictracer/protocol_http.h bpf/tests/generictracer/maps/pid_tid_to_conn.h bpf/tests/generictracer/maps/ssl_to_pid_tid.h bpf/tests/maps/ssl_to_conn.h`
- `make -C bpf/tests bpf_ssl_read_guard`
- `./bpf/tests/bpf_ssl_read_guard`
- `make generate`
- `make compile`
- `make license-header-check`

## Notes

- Full `make -C bpf/tests` currently fails later on existing `test_large_buffer_loop.c` compile diagnostics after compiling the SSL guard test.
- `make clang-tidy` currently fails on existing repo-wide diagnostics outside the touched SSL files.


bpf/generictracer/libssl.c
```diff
@@ -226,14 +221,20 @@ int BPF_URETPROBE(obi_uretprobe_ssl_write_ex, int ret) {
...
-        __builtin_memcpy(&saved, args, sizeof(ssl_args_t));
+    if (ret != 1 || !args) {
         bpf_map_delete_elem(&active_ssl_write_args, &id);
-        // must be last in the function, doesn't return
-        handle_ssl_buf(ctx, id, &saved, saved.len_ptr, TCP_SEND);
+        return 0;
     }
 
+    size_t write_len = 0;
+    bpf_probe_read(&write_len, sizeof(write_len), (void *)args->len_ptr);
```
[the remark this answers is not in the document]
MrAlias: Fixed in 3975247e: `SSL_write_ex` now reads the userspace `written` pointer with `bpf_probe_read_user`. I also updated `SSL_read_ex` for the same userspace `size_t *` pattern.

bpf/tests/bpf_ssl_read_guard.c
```diff
@@ -0,0 +1,229 @@
...
+        }
+        return NULL;
+    }
+
+    if (map == &pid_tid_to_conn) {
+        const u64 pid_tid = *(const u64 *)key;
+        if (test_ssl_conn_available && pid_tid == test_mapped_pid_tid) {
+            return &test_ssl_conn;
+        }
+        return NULL;
```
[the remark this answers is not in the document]
MrAlias: Fixed in 3975247e: the `pid_tid_to_conn` lookup now uses `test_mapped_pid_tid_available`, and the test covers reusing a mapped PID/TID connection independently from `ssl_to_conn` availability.

grcevski (approved):
Sweet! LGTM!

bpf/tests/common/connection_info.h
```diff
@@ -0,0 +1,30 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <bpfcore/vmlinux.h>
+
+enum { IP_V6_ADDR_LEN = 16 };
+
+typedef struct connection_info {
```
rafaelroquetto: I am not sure about these, I think the goal is to be able to include the original connection_info header directly here, does that not work?

We don't wanna be duplicating those types, it will become unmaintainable

bpf/tests/common/trace_key.h
```diff
@@ -0,0 +1,15 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <bpfcore/vmlinux.h>
+
+typedef struct trace_key {
```
rafaelroquetto: ditto

rafaelroquetto (changes requested):
Usually we should strive to reuse existing ebpf headers / types

bpf/tests/common/connection_info.h
MrAlias: These were originally added to keep this host-side C test focused on `handle_ssl_buf()` without pulling in the full eBPF include graph. At the time, including the real `common/connection_info.h` also brought in helper/logging/protocol definitions, which collided with the test-local stubs for things like `TCP_RECV`, `TCP_SEND`, `WITH_SSL`, and `pid_from_pid_tgid`.

Rechecking this now, the duplicated type shim is not the right tradeoff. We can use the real shared headers for `connection_info_t`, `pid_connection_info_t`, `ssl_pid_connection_info_t`, and `trace_key_t` if the test stub stops redefining symbols already provided by those headers. I’ll remove these duplicated type headers and keep only the behavior/map stubs that are needed to isolate the test.

MrAlias: Addressed in fb7d80daa. I removed the `bpf/tests/common/connection_info.h` shim entirely and stopped the test `k_tracer_defs.h` stub from redefining `TCP_RECV`, `TCP_SEND`, `WITH_SSL`, and `pid_from_pid_tgid`, so this test build now uses the real shared `common/connection_info.h` types. The remaining test-local headers are behavior/map stubs only.

bpf/tests/common/trace_key.h
MrAlias: Addressed in fb7d80daa as well. I removed the `bpf/tests/common/trace_key.h` shim, so `ssl_defs.h` now sees the real shared `common/trace_key.h` definition during this test build.

bpf/generictracer/libssl.c
```diff
@@ -226,14 +251,49 @@ int BPF_URETPROBE(obi_uretprobe_ssl_write_ex, int ret) {
...
+    ssl_args_t *args = bpf_map_lookup_elem(&active_ssl_write_args, &id);
+
+    bpf_dbg_printk("=== uretprobe SSL_write_ex2 id=%d args %llx ===", id, args);
+
+    if (ret != 1 || !args) {
+        bpf_map_delete_elem(&active_ssl_write_args, &id);
+        return 0;
     }
 
+    size_t write_len = 0;
```
rafaelroquetto, line 301: after you read write_len, you should do a `min()` to check for `INT_MAX`, because handle_ssl_buf takes an integer and this can _theoretically_ truncate/wrap.

rafaelroquetto (approved):
LGTM, just a small nit, approving as it's mostly theoretical

bpf/tests/bpf_ssl_read_guard.c
```diff
@@ -0,0 +1,278 @@
...
+
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <bpfcore/bpf_helpers.h>
+
+enum { BPF_ANY = 0, BPF_NOEXIST = 1 };
+
```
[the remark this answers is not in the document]
rafaelroquetto: Not sure about this one

rafaelroquetto (approved):

bpf/generictracer/libssl.c
MrAlias: Good call. This is already covered on the current head: `ssl_size_to_int()` clamps the `size_t` value to `__INT_MAX__`, and both `SSL_write_ex` and `SSL_write_ex2` pass `ssl_size_to_int(write_len)` into `handle_ssl_buf()`. Resolving as already addressed.

bpf/generictracer/libssl.c
```diff
@@ -125,11 +130,11 @@ int BPF_URETPROBE(obi_uretprobe_ssl_read_ex, int ret) {
     }
```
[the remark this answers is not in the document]
MrAlias: Fixed in 28d388d89: `SSL_read_ex` now handles missing args separately, and on `ret != 1` it deletes both `active_ssl_read_args` and the matching `ssl_to_pid_tid` entry for `args->ssl` before returning.

bpf/tests/bpf_ssl_read_guard.c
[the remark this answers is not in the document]
MrAlias: I rechecked this after Rafael’s note and I do not think this feedback is valid for this host-side test. The test resolves `<bpfcore/bpf_helpers.h>` to the test stub under `bpf/tests/bpfcore/`, so the production helper-header include-order requirement does not apply here. I also verified `pid_from_pid_tgid()` is present in the preprocessed test translation unit, and the focused test still builds and runs with `make -C bpf/tests bpf_ssl_read_guard` and `./bpf/tests/bpf_ssl_read_guard` under `-Wall -Werror`. Resolving without a code change.

PR #2243, opened 2026-06-03
fix(2033): bound legacy traceparent scan to fresh payload

### LINKED ISSUE #2033: Legacy traceparent scan may reuse stale buffer data
## Summary

- The legacy HTTP traceparent scan can reuse bytes left in the per-CPU scratch buffer from a previous request. In the current path, OBI reads only `buf_len` bytes into `tp_char_buf_mem()` and writes a single null terminator, while `bpf_strstr_tp_loop__legacy()` ignores `buf_len` and scans a fixed 350-byte window. A later short request can therefore inherit a previous request's traceparent.

## Impact

- This corrupts trace integrity and can leak trace metadata across requests handled by the same CPU.
- The issue requires `TrackRequestHeaders` or `ContextPropagation` to be enabled, because those settings enable header-buffer capture and traceparent parsing.
- The observed impact is limited to trace metadata propagation; this triage did not find evidence of arbitrary payload disclosure or code execution.

## Environment

- First identified in: 57e7cd462e0300e0acdd3cf76b146e4cd80e225f

## Evidence

- [bpf/generictracer/protocol_http.h](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/bpf/generictracer/protocol_http.h#L458-L483) reads only `buf_len` bytes into `tp_char_buf_mem()` and writes a single terminator instead of clearing the scratch buffer.
- [bpf/generictracer/protocol_http.h](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/bpf/generictracer/protocol_http.h#L530-L546) invokes the legacy traceparent scan on that reused scratch buffer.
- [bpf/common/trace_util.h](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/bpf/common/trace_util.h#L117-L134) shows `bpf_strstr_tp_loop__legacy()` scanning a fixed 350-byte window rather than the current header length.
- [pkg/internal/ebpf/generictracer/generictracer.go](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/pkg/internal/ebpf/generictracer/generictracer.go#L186-L214) shows traceparent parsing is enabled when request-header capture or context propagation is configured.

## Steps to reproduce

1. From the repository root on `main`, create `/tmp/poc_traceparent_stale.c` with these exact contents:
   ```c
   #include <stdio.h>
   #include <string.h>

   static int is_traceparent(const unsigned char *buf) {
       return memcmp(buf, "traceparent:", 12) == 0;
   }

   int main(void) {
       unsigned char scratch[350];
       const char *first = "traceparent: 00-11111111111111111111111111111111-2222222222222222-01\r\n";
       const char *second = "traceparent";
       size_t first_len = strlen(first);
       size_t second_len = strlen(second);
       int first_match = 0;
       int second_match = 0;

       memset(scratch, 0, sizeof(scratch));
       memcpy(scratch, first, first_len);
       for (size_t i = 0; i < first_len; i++) {
           if (is_traceparent(&scratch[i])) {
               first_match = 1;
               break;
           }
       }

       memcpy(scratch, second, second_len);
       for (size_t i = 0; i < second_len; i++) {
           if (is_traceparent(&scratch[i])) {
               second_match = 1;
               break;
           }
       }

       printf("first_match=%d second_match=%d\n", first_match, second_match);
       return 0;
   }
   ```
2. Build it with `gcc -O0 -g -fsanitize=address -o /tmp/poc_traceparent_stale /tmp/poc_traceparent_stale.c`.
3. Run `ASAN_OPTIONS=detect_leaks=0 /tmp/poc_traceparent_stale`.
4. Observe that the program prints `first_match=1 second_match=1`, which shows the second short buffer still matches `traceparent:` by reading the stale trailing `:` left behind from the first request.

## Suggested Fix Direction

- Clear the traceparent scratch buffer before scanning it, or
- bound the legacy scan to `buf_len` instead of scanning a fixed 350-byte window, or
- do both to preserve safety across current and future parser paths.

## Acceptance Criteria

- Short requests without a traceparent cannot match bytes left over from earlier requests in the same scratch buffer.
- The legacy parser does not inspect bytes beyond the valid header length for the current request.
- Regression coverage should include the stale-buffer sequence above so the short-overwrite case stays covered.

## Note

MrAlias:
## Summary

- Bound the legacy HTTP traceparent scan by the fresh header buffer length instead of scanning a fixed window.
- Reject buffers shorter than a complete traceparent header before probing candidate positions.
- Added a functional C regression test for the stale scratch-buffer sequence.

## Root Cause

The legacy traceparent scanner ignored `buf_len` and scanned up to a fixed 350-byte window. Because the HTTP path reuses a per-CPU scratch buffer, a short later request could match bytes left behind by an earlier request and incorrectly reuse stale trace metadata.

## Validation

- `make -C bpf/tests test_trace_util`
- `./bpf/tests/test_trace_util`
- `clang-format --dry-run --Werror bpf/common/trace_util.h bpf/tests/test_trace_util.c`
- `go test -count=1 ./pkg/internal/ebpf/generictracer`

Fixes #2033

### DIFF
--- a/bpf/common/trace_util.h
+++ b/bpf/common/trace_util.h
@@ -127,11 +127,11 @@ static __always_inline unsigned char *bpf_strstr_tp_loop__legacy(unsigned char *
     }
 
     // Limited best-effort search to stay within insns limit
-    const u16 k_besteffort_max_loops = 350;
-    const u16 scan_len = buf_len - TRACE_PARENT_HEADER_LEN + 1;
-    const u16 nr_loops = scan_len < k_besteffort_max_loops ? scan_len : k_besteffort_max_loops;
+    const u32 k_besteffort_max_loops = 350;
+    const u32 scan_len = (u32)buf_len - TRACE_PARENT_HEADER_LEN + 1;
+    const u32 nr_loops = scan_len < k_besteffort_max_loops ? scan_len : k_besteffort_max_loops;
 
-    for (u16 i = 0; i < nr_loops; i++) {
+    for (u32 i = 0; i < nr_loops; i++) {
         if (is_traceparent(&buf[i])) {
             return &buf[i];
         }
--- a/bpf/common/trace_util.h
+++ b/bpf/common/trace_util.h
@@ -118,16 +118,20 @@ static __always_inline unsigned char *bpf_strstr_tp_loop(unsigned char *buf, con
 
 static __always_inline unsigned char *bpf_strstr_tp_loop__legacy(unsigned char *buf,
                                                                  const u16 buf_len) {
-    (void)buf_len;
-
     if (!g_bpf_traceparent_enabled) {
+        return NULL;
+    }
+
+    if (buf_len < TRACE_PARENT_HEADER_LEN) {
         return NULL;
     }
 
     // Limited best-effort search to stay within insns limit
     const u16 k_besteffort_max_loops = 350;
+    const u16 scan_len = buf_len - TRACE_PARENT_HEADER_LEN + 1;
+    const u16 nr_loops = scan_len < k_besteffort_max_loops ? scan_len : k_besteffort_max_loops;
 
-    for (u16 i = 0; i < k_besteffort_max_loops; i++) {
+    for (u16 i = 0; i < nr_loops; i++) {
         if (is_traceparent(&buf[i])) {
             return &buf[i];
         }
--- a/bpf/tests/test_trace_util.c
+++ b/bpf/tests/test_trace_util.c
@@ -1,0 +1,105 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Run me with: make && ./test_trace_util
+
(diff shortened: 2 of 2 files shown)


MrAlias:
I'm looking into CI failures.

rafaelroquetto:
@MrAlias the 5.10 failures will probably benefit from #2233

rafaelroquetto:
as a matter of fact I believe #2233 already encompasses the changes by this PR, maybe it's fine to close this one?

MrAlias:
Yeah, I think I'm chasing my tail here. Going to wait for #2233 to land.

MrAlias:
Superseded by #2233

Issue #2234, opened 2026-06-02
Proposal: Expand DynamicSelector with independent signal controls


damemi:
## Purpose

The goal of this design is to provide developers of higher-level controllers the ability to import OBI as a dependency and dynamically configure application selection for all supported signals in OBI. This design is **not** intended to impact or modify any existing (static config) functionality.

## Background and Motivation

https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1388 introduced the DynamicPIDSelector public API. This created a handle for developers to add and remove AppO11y processes by PID for trace instrumentation.

Expanding on this, we would like to add similar support for dynamic selection in application, network, and stats metrics for full dynamic selector parity with OBI's static config.

## Goals

* Provide a flexible yet concise developer API for process selection for all signals.
* Implement changes in independent, easily reviewable, self-contained phases.

## Non-goals

* Modify existing static configuration use cases, pipelines, or code.
* Break or strictly depend on future phases while implementing intermediate phases.

## Phases

- [x] https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2250
- [x] https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2279
- [x] https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2340
- [x] https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2421
- [x] https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2442

## Design

The outcome of this design will extend DynamicPIDSelector with signal-specific views that allow developers to select applications for instrumentation at both a global level and a targeted level. Such as:

```go
selector.Traces().AddPIDs(1234)
selector.AppMetrics().AddPIDs(1234)
selector.NetworkMetrics().AddPIDs(1234)
selector.StatsMetrics().AddPIDs(1234)
selector.AddPIDs(1234) // global
```

This design will be brought in phases meant to keep changes logical and contained.

### Phase 1: Move DynamicPIDSelector to global context

DynamicPIDSelector currently exists as a [field on the `ContextInfo.AppO11y` struct](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/e2f806535af8505d09b074a78153c52170e40683/pkg/pipe/global/context.go#L77), since it is only used by AppOlly instrumentation. 

[...]

PR #2250, opened 2026-06-04
(DynamicSelector-Phase1) Move DynamicPIDSelector to global context

damemi:
## Summary

Implements Phase 1 of https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2234

This moves the `DynamicPIDSelector` concrete type up one level from `contextInfo.AppO11y.DynamicPIDSelector` to `contextInfo.DynamicPIDSelector` and creates a `PIDSelector` interface that will be used in the later Phase 3 for implementing signal-aware views.

This is technically a breaking change for anyone that was directly accessing `contextInfo.AppO11y.DynamicPIDSelector`, but that was not the documented intended use case for it to be in that location

## Validation

### DIFF
--- a/pkg/appolly/discover/dynamic_pid_selector.go
+++ b/pkg/appolly/discover/dynamic_pid_selector.go
@@ -10,12 +10,15 @@ import (
 	"go.opentelemetry.io/obi/pkg/appolly/app"
 	"go.opentelemetry.io/obi/pkg/appolly/services"
 	"go.opentelemetry.io/obi/pkg/export/otel/perapp"
+	"go.opentelemetry.io/obi/pkg/selection"
 )
 
+var _ selection.PIDSelector = (*DynamicPIDSelector)(nil)
+
 // DynamicPIDSelector holds the runtime set of target PIDs for OBI. It is preloaded from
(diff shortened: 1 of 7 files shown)


pkg/appolly/discover/dynamic_pid_selector.go
```diff
@@ -10,12 +10,15 @@ import (
 	"go.opentelemetry.io/obi/pkg/appolly/app"
 	"go.opentelemetry.io/obi/pkg/appolly/services"
 	"go.opentelemetry.io/obi/pkg/export/otel/perapp"
+	"go.opentelemetry.io/obi/pkg/selection"
 )
 
+var _ selection.PIDSelector = (*DynamicPIDSelector)(nil)
```
mariomac, line 16: This shouldn't be needed if the `interface = implementation` is used somewhere else. E.g. a unit test?

For the rest, LGTM!

mariomac (approved):

pkg/appolly/discover/dynamic_pid_selector.go
damemi: true but it's probably a little more reliable to have it in build code right? not sure if it makes a big difference or not

PR #2254, opened 2026-06-04
Add lint-fix import and collectt fixers

MrAlias:
## Summary

- Route `lint-fix` through explicit fix-specific targets.
- Run `vanity-import-fix` before the vanity import check on the fix path.
- Add suggested fixes to the collectt analyzer so incorrect `EventuallyWithT` assertion targets can be rewritten automatically.
- Add `lint-collectt-fix` and have `lint-fix` use it before `golangci-lint --fix`.

## Why

`lint-fix` previously used the same check-first dependency path as `lint`, so fixable vanity import issues could fail before their fixer ran. The collectt analyzer also reported mechanically fixable issues but did not expose suggested fixes for `singlechecker -fix`.

## Validation

- `go test ./internal/test/analyzer/collectt`
- `make -n lint-collectt-fix`
- `make -n lint-fix`
- `make -n -j8 lint-fix`
- `make -n lint`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,10 +141,12 @@ lint: lint-run
 
 .PHONY: lint-fix
 lint-fix: LINT_EXTRA_ARGS = --fix
-lint-fix: lint-run
+lint-fix: lint-fix-run
 
-.PHONY: lint-run
+.PHONY: lint-run lint-fix-run
 lint-run: vanity-import-check lint-dependency-policy lint-collectt
+lint-fix-run: vanity-import-fix-check lint-dependency-policy lint-collectt
+lint-run lint-fix-run:
 	@echo "### Linting code"
 	go tool $(TOOLS_MODFILE) golangci-lint run ./... --timeout=6m $(LINT_EXTRA_ARGS)
 
@@ -870,8 +872,9 @@ check-ebpf-ver-synced:
 		exit 1; \
 	fi
 
-.PHONY: vanity-import-check
-vanity-import-check:
+.PHONY: vanity-import-check vanity-import-fix-check
+vanity-import-fix-check: vanity-import-fix
+vanity-import-check vanity-import-fix-check:
 	go tool $(TOOLS_MODFILE) porto --include-internal --skip-dirs "^NOTICES$$" -l . || ( echo "(run: make vanity-import-fix)"; exit 1 )
 
 .PHONY: vanity-import-fix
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -141,10 +141,12 @@ lint: lint-run
 
 .PHONY: lint-fix
 lint-fix: LINT_EXTRA_ARGS = --fix
-lint-fix: lint-run
+lint-fix: lint-fix-run
 
-.PHONY: lint-run
+.PHONY: lint-run lint-fix-run
 lint-run: vanity-import-check lint-dependency-policy lint-collectt
+lint-fix-run: vanity-import-fix-check lint-dependency-policy lint-collectt-fix
```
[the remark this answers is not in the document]
MrAlias: Addressed in 50d3234c3: `lint-fix-run` now sets `LINT_EXTRA_ARGS = --fix` itself, so direct `make lint-fix-run` invokes `golangci-lint` in fix mode. I also added `.NOTPARALLEL: lint-fix-run` so the fix path prerequisites are serialized under `make -j lint-fix` and the file-writing fixers cannot overlap.

Verified with:
- `make -n lint-fix-run`
- `make -n lint-fix`
- `make -n -j8 lint-fix`
- `make -n lint`
- `go test ./internal/test/analyzer/collectt`

Makefile
```diff
@@ -169,6 +172,13 @@ lint-collectt:
 	@echo "### Checking EventuallyWithT callbacks use CollectT"
 	go run ./internal/test/analyzer/collectt/cmd/collecttlint ./...
 
+.PHONY: lint-collectt-fix
+lint-collectt-fix:
+	@echo "### Fixing EventuallyWithT callbacks to use CollectT"
+	go run ./internal/test/analyzer/collectt/cmd/collecttlint -fix ./...
+	@echo "### Checking EventuallyWithT callbacks use CollectT"
+	go run ./internal/test/analyzer/collectt/cmd/collecttlint ./...
+
```
mariomac, line 181: Would be amazing to find a way to integrate this inside the golangci-lint file so we don't need a Makefile target only for a concrete lint check.

mariomac (approved):
Looks good to me!

Makefile
MrAlias: Agreed! I can take a look at this as a follow-up.

MrAlias: I took a look at this. We would need to build/run a custom golangci-lint binary via `golangci-lint custom` to get this to work.

I don't think it is impossible, but it comes with its downsides of having to maintain that custom build.

PR #2255, opened 2026-06-05
Add initial support for Go runtime metrics

marctc:
## Summary

- Added initial support for Go runtime metrics. In a follow up PR will add the rest of metrics.
- Added new feature `application_runtime` to enable/disable collection of runtime metrics of instrumented apps.

## Validation


pkg/appolly/instrumenter.go
```diff
@@ -151,52 +153,72 @@ func setupMetricsSubPipeline(
...
+	}
+
+	if jointMetricsConfig.Features.AppRuntime() {
+		swi.Add(otel.ReportRuntimeMetrics(
+			ctxInfo,
+			&config.OTELMetrics,
+			jointMetricsConfig,
+			runtimeMetrics,
+		), swarm.WithID("OTELRuntimeMetricsExport"))
+	}
```
mariomac: I think this could be simplified because the Swarm and Queue types already take care of optional nodes or bypassing channels. But I'm fine keeping as this. We can always simplify after it's merged and working.

pkg/export/otel/metrics_runtime.go
```diff
@@ -0,0 +1,282 @@
...
+	return &metrics, nil
+}
+
+func setupRuntimeMeters(metrics *RuntimeMetrics, meter instrument.Meter) error {
+	return setupGoRuntimeMeters(&metrics.goMetrics, meter)
+}
+
+func setupGoRuntimeMeters(metrics *goRuntimeMetrics, meter instrument.Meter) error {
+	var err error
+	metrics.memoryLimit, err = meter.Int64UpDownCounter("go.memory.limit", instrument.WithUnit("By"))
```
mariomac: Instead of hardcoding metric names here, I'd define them as `Name` entries within the [metric.go file](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/export/attributes/metric.go), together with the rest of entries.

pkg/export/prom/prom_runtime.go
```diff
@@ -0,0 +1,139 @@
...
+	memoryGCCycles *prometheus.CounterVec
+	processorLimit *prometheus.GaugeVec
+	configGOGC     *prometheus.GaugeVec
+	gcCycles       map[string]uint64
+}
+
+func newGoRuntimeMetricsCollector(runtimeLabelNames []string) goRuntimeMetricsCollector {
+	return goRuntimeMetricsCollector{
+		memoryLimit: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "go_memory_limit_bytes",
```
mariomac: Same comment as before, add the entries to [metric.go](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/export/attributes/metric.go) and reference from here

pkg/pipe/global/context.go
```diff
@@ -75,4 +76,6 @@ type AppO11y struct {
 	// (e.g. discover.NewDynamicPIDSelector()), passes it via instrumenter.WithDynamicPIDSelector, and
 	// calls AddPIDs/RemovePIDs/GetPIDs on it directly. The instrumenter does not implement an updater interface.
 	DynamicPIDSelector any
+	// RuntimeMetrics receives application runtime metric snapshots emitted by eBPF tracers.
+	RuntimeMetrics *msg.Queue[[]runtimemetrics.RuntimeMetricSnapshot]
```
mariomac: Do we really need this to be a global, public struct field? In theory should be enough to create locally a variable that is passed as argument to the different stages of the pipeline.

I'd only keep it here if we explicitly want this queue to be available to any software vendoring OBI.

mariomac (commented):
Great job! I'd still need to do a deeper, second view, and let a second pair of expert eyes to take a look.

Here are some comments.

NimrodAvni78 (commented):
Looks great!
Can we wire in weaver to the integration test like we do in other tests?
This will make sure we follow to otel semconv for these metrics

marctc:
> Great job! I'd still need to do a deeper, second view, and let a second pair of expert eyes to take a look.
> 
> Here are some comments.

thanks, good feedback! can you take another look?

pkg/internal/ebpf/gotracer/gotracer.go
```diff
@@ -316,6 +340,83 @@ func (p *Tracer) RegisterOffsets(fileInfo *exec.FileInfo, offsets *goexec.Offset
...
+		return
+	}
+
+	pidInfo, err := runtimeMetricPIDInfo(pid, ns)
+	if err != nil {
+		p.log.Debug("runtime metrics PID key lookup failed", "pid", pid, "ns", ns, "error", err)
+		return
+	}
+
+	symbols, err := goexec.ResolveRuntimeMetricSymbols(fileInfo, pid)
```
grcevski, line 357: I think this approach is good. I was hoping that we might get away without having to recalculate this in userspace, by reading something at eBPF time in the uprobes, but I can't think of an approach now. I don't see mstat or others passed as arguments in any Go functions.

pkg/internal/runtimemetrics/reader.go
```diff
@@ -0,0 +1,135 @@
...
+func runtimeMetricService(
+	currentPIDs map[uint32]map[app.PID]svc.Attrs,
+	key goRuntimeMetricRawKey,
+) (svc.Attrs, bool) {
+	pids, ok := currentPIDs[key.Ns]
+	if !ok {
+		return svc.Attrs{}, false
+	}
+	service, ok := pids[app.PID(key.UserPID)]
+	if !ok || service.SDKLanguage != svc.InstrumentableGolang || !service.Features.AppRuntime() {
```
grcevski: Nit: Do we need to guard for Go here, I'm thinking we'll be adding other runtime metrics in the future?

grcevski (approved):
LGTM! Great work, I just have one small non-blocking nit question, but since this is for Go only for now we can leave it as is.

pkg/internal/ebpf/gotracer/gotracer.go
marctc: Yeah, I was looking also how get around but sadly can't be done without passing the symbols

mariomac (approved):
🚀

pkg/internal/runtimemetrics/reader.go
marctc: probably not, removing!

PR #2258, opened 2026-06-05
PoC: bridge Go Auto SDK manual spans

MrAlias:
## What this is

This is a PoC / steel-thread implementation for making OBI interoperate with manually instrumented Go spans through the same basic mechanism used by `go.opentelemetry.io/auto`.

I am opening this as a draft because I want maintainers to review the shape of the approach before we split it into smaller production-ready pieces. The goal here is to prove that we can support the full Go OTel API path, including the Auto SDK `Span_ended(buf []byte)` JSON payload, while preserving OBI's existing fallback behavior when `bpf_probe_write_user` is not available.

## What changed

This adds an opportunistic privileged path for Go manual spans:

- When OBI has the same permissions required for context propagation / `bpf_probe_write_user`, it attaches additional Go OTel probes.
- `go.opentelemetry.io/otel/internal/global.(*tracer).newSpan` flips the `autoSpan` flag so the global OTel API routes through `go.opentelemetry.io/auto/sdk`.
- `go.opentelemetry.io/auto/sdk.(*tracer).start` writes parent span context, child span context, and sampled state into the Auto SDK output args.
- `go.opentelemetry.io/auto/sdk.(*span).ended` emits the Auto SDK OTLP JSON span payload into a new OBI event.
- Userspace decodes that OTLP JSON with Collector pdata and exports the decoded ScopeSpans under OBI's resource span.

The existing synthetic manual-span path remains the fallback. There is no user-facing config toggle in this PoC: if OBI has the needed permissions, the richer path should work; otherwise OBI keeps doing what it does today.

## Upstream probe references

The BPF code in this PR is intentionally modeled after the `go.opentelemetry.io/auto` probes:

- Auto SDK probe reference: https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/internal/pkg/instrumentation/bpf/go.opentelemetry.io/auto/sdk/bpf/probe.bpf.c
- OTel global tracer flag probe reference: https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/internal/pkg/instrumentation/bpf/go.opentelemetry.io/otel/traceglobal/bpf/probe.bpf.c

## Why this shape

The important bit is the Auto SDK `ended(buf []byte)` path. That payload is already OTLP JSON-shaped, and it includes the manual span data we would otherwise have to reconstruct ourselves: attributes, events, links, status, dropped counts, trace state, IDs, kind, timestamps, and scope data.

Instead of hand-modeling all of that in OBI's fixed synthetic span event, this PR treats the Auto SDK payload as the source of truth and lets pdata decode it. That keeps the BPF event simple and gives us much better parity with the Go OTel API.

## Fallback behavior

The privileged path is guarded by the same context-propagation capability checks OBI already uses. If the kernel or process permissions do not allow `bpf_probe_write_user`, OBI's existing integrity-mode override path should continue to fall back to the current synthetic manual span behavior.

This is intentionally not a new config option in this PoC. The model I want to test is: if the operator granted the privilege needed for this support, use it; if not, degrade gracefully.

## Known gaps

This is not a final production split yet.

- The Auto SDK JSON payload is bounded at 16 KiB in this PoC. Larger manual spans are dropped by the BPF event path.
- OBI does not currently push its userspace sampler decision back into the Auto SDK before recording. We preserve parent sampled state when present and default new roots to sampled.
- TraceState is not written into Go `trace.SpanContext`, matching the upstream Auto `write_span_context` helper behavior, but it is preserved from the Auto SDK JSON payload once emitted.
- This needs integration coverage with a target app using the global OTel API and Auto SDK manual spans before it should be considered merge-ready.
- The change should probably be split further after we agree on the shape: offset plumbing, privileged probes, payload event/parser, and exporter path can be separate PRs.

## Validation

I validated the steel thread locally with:

```sh
make generate
go test ./pkg/ebpf/common ./pkg/export/otel/tracesgen ./pkg/internal/goexec ./pkg/internal/ebpf/gotracer ./cmd/obi
```


grcevski (commented):
Awesome, I like the direction!

MrAlias:
Closing as I have started work to integrate this feature elsewhere.

PR #2260, opened 2026-06-06
fix(2259): share the java_tasks map with the tpinjector program

### LINKED ISSUE #2259: Java thread-pool trace correlation breaks with context propagation enabled: java_tasks map is not shared with the tpinjector BPF object
## What happens

With `OTEL_EBPF_BPF_CONTEXT_PROPAGATION=all`, a Java client span whose egress runs on a different thread than the one that read the server request (the classic `ExecutorService` handoff that the thread-pool support exists for) is emitted as a new root trace, or occasionally stitched to a different request's trace when a kept-alive connection is reused.

This only affects plain-HTTP egress, which is why the existing `TestJavaNestedTraces` async cases pass: they call `https://httpbin.org`, and the TLS path resolves the parent through the generictracer ioctl handler, which sees the populated map.

Measured on current main (3a24937a) with a minimal Spring Boot app (platform threads only, no virtual threads) where the handler submits a plain-HTTP downstream GET to a small `FixedThreadPool` and blocks on the `Future`: out of 400 requests tagged with unique URL path ids,

- correct (client span in the same trace as its own server span): 19 (4.8%)
- orphaned (client span is a new root trace): 317 (79.4%)
- stitched to ANOTHER request's trace: 63 (15.8%)

A same-thread control on the same deployment (downstream call made synchronously in the handler, 200 requests) is 100% correct.

Environment: Linux 6.17, docker compose, app on eclipse-temurin:21, OBI built from main at 3a24937a.

## Why

`find_parent_java_trace` (`bpf/common/trace_parent.h`) resolves the egress thread to the server trace by walking `java_tasks` (the parent-tid edges the Java agent reports on executor handoffs via the `k_ioctl_java_threads` ioctl).

That walk runs in TWO separate BPF objects:

1. generictracer, whose `kprobe/sys_ioctl` handler is also the only WRITER of `java_tasks` (`bpf/generictracer/java_tls.c`);
2. tpinjector, because `bpf/tpinjector/tpinjector.c` includes `common/trace_parent.h`, and with context propagation enabled its `sk_msg` program is the one that decides the outgoing traceparent for a plain-HTTP client request.

`bpf/maps/java_tasks.h` has no `pinning` attribute, so each object gets a private map instance. tpinjector's copy is always empty, its chain walk can never succeed, and the `sk_msg` program injects a fresh trace id, which the rest of the pipeline then reuses for the client span. `server_traces` IS `OBI_PIN_INTERNAL`, which is why same-thread correlation works fine.

This can be confirmed without running anything: `strings pkg/internal/ebpf/tpinjector/bpf_x86_bpfel.o | grep java_tasks` shows the private map definition in the tpinjector object.

## Fix (verified)

Share the map the same way `server_traces` already is:

```diff
 #include <common/map_sizing.h>
+#include <common/pin_internal.h>

 struct {
     __uint(type, BPF_MAP_TYPE_LRU_HASH);
     __type(key, pid_key_t);   // the client thread
     __type(value, pid_key_t); // the server thread
     __uint(max_entries, MAX_CONCURRENT_REQUESTS);
+    __uint(pinning, OBI_PIN_INTERNAL);
 } java_tasks SEC(".maps");
```

Verified on a worktree of main 3a24937a with ONLY this change: the same 400-request pool-handoff load goes from 95.2% broken to 400/400 correct (client span in its own request's trace AND parented to that server span), and the 200-request same-thread control stays 100% correct.

I can open a PR with this change, optionally including an integration test that exercises an executor handoff with a plain-HTTP downstream target.

## Repro

Same triangle as the repro in #2242 (app, nginx downstream, OBI with the text printer), with this handler in the app:

```java
private static final ExecutorService POOL = Executors.newFixedThreadPool(2);

@GetMapping("/pool/{id}")
public Map<String, Object> pool(@PathVariable String id) throws Exception {
    int code = POOL.submit(() -> httpGet("http://downstream/echo/" + id)).get();
    return Map.of("status", code, "id", id);
}
```

Drive it with unique ids (`seq 1 400 | xargs -P 20 -I{} curl -fsS localhost:18080/pool/id{}`), then for each id compare the trace id of the server span `/pool/{id}` against the client span `/echo/{id}` in the OBI output. OBI env: `OTEL_EBPF_OPEN_PORT=8080`, `OTEL_EBPF_TRACE_PRINTER=text`, `OTEL_EBPF_BPF_CONTEXT_PROPAGATION=all`.

lucacavenaghi97:
## What

Adds `OBI_PIN_INTERNAL` to the `java_tasks` BPF map, plus integration coverage for executor handoffs with a plain-HTTP downstream.

Fixes #2259

## Why

`find_parent_java_trace` walks `java_tasks` from two BPF objects: generictracer, which also populates the map from its ioctl handler, and tpinjector via `common/trace_parent.h`. With `OTEL_EBPF_BPF_CONTEXT_PROPAGATION=all` the tpinjector sk_msg program decides the traceparent of plain-HTTP client requests, and without a pinning attribute it resolves a private, always empty copy of the map, so any Java client call running on a different thread than the server read loses its trace context (measured ~95% of pool-handoff requests broken, details in #2259). `server_traces` already uses the same internal pinning.

The existing async cases in `TestJavaNestedTraces` did not catch this because they call httpbin over TLS, and the TLS path resolves the parent through the generictracer ioctl handler, which sees the populated map.

## Test

The new `Nested traces over plain HTTP` cases drive `/api/async-request-c` and `/api/async-request-fj` against a local plain-HTTP target (a second testserver instance on a port outside the instrumented range) and require at least 80% of the returned server traces to contain the nested client span. The ratio assertion matters: the first request on a fresh connection correlates through the connect-time path even without the fix, so a single-request assertion can pass by luck on main.

Verified locally: on main without the map change both new cases fail for the full retry window (nested ratio stays under 10%); with it the whole `TestJavaNestedTraces` suite passes, including the weaver validation.

### DIFF
--- a/bpf/maps/java_tasks.h
+++ b/bpf/maps/java_tasks.h
@@ -8,10 +8,12 @@
 #include <bpfcore/bpf_helpers.h>
 
 #include <common/map_sizing.h>
+#include <common/pin_internal.h>
 
 struct {
     __uint(type, BPF_MAP_TYPE_LRU_HASH);
     __type(key, pid_key_t);   // the client thread
     __type(value, pid_key_t); // the server thread
     __uint(max_entries, MAX_CONCURRENT_REQUESTS);
+    __uint(pinning, OBI_PIN_INTERNAL);
 } java_tasks SEC(".maps");
--- a/internal/test/integration/docker-compose-java-dist.yml
+++ b/internal/test/integration/docker-compose-java-dist.yml
@@ -9,6 +9,15 @@ services:
       LOG_LEVEL: DEBUG
       OTEL_SERVICE_NAME: testserver
 
+  downstream:
+    build:
+      context: ../../..
+      dockerfile: internal/test/integration/components/testserver/Dockerfile
+    image: hatest-testserver
+    environment:
+      STD_PORT: "8086"
+      LOG_LEVEL: DEBUG
+
   obi:
     build:
       context: ../../..
--- a/internal/test/integration/java_dist_test.go
+++ b/internal/test/integration/java_dist_test.go
@@ -51,6 +51,31 @@ func testJavaNestedTraces(t *testing.T, slug string) {
 	}, 2*time.Minute, 5*time.Second)
 }
 
+func testJavaNestedTracesPlainHTTP(t *testing.T, slug string) {
+	t.Log("checking server to client nesting over plain HTTP for [/api/" + slug + "]")
+	require.EventuallyWithT(t, func(ct *assert.CollectT) {
+		for i := 0; i < 10; i++ {
+			ti.DoHTTPGet(ct, "http://localhost:8081/api/"+slug+"?url=http://downstream:8086/rolldice/1", 200)
+		}
+
+		resp, err := http.Get(jaegerQueryURL + "?service=testserver&operation=GET%20%2Fapi%2F" + slug)
+		require.NoError(ct, err)
+		if resp == nil {
+			return
+		}
+		require.Equal(ct, http.StatusOK, resp.StatusCode)
+		var tq jaeger.TracesQuery
+		require.NoError(ct, json.NewDecoder(resp.Body).Decode(&tq))
+		traces := tq.FindBySpan(jaeger.Tag{Key: "url.path", Type: "string", Value: "/api/" + slug})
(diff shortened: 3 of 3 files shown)


grcevski (approved):
LGTM!

PR #2261, opened 2026-06-06
fix(2242): correlate Java traces across virtual thread mounts

### LINKED ISSUE #2242: Java in-process trace context keyed on kernel thread id breaks under JDK 21 virtual threads
## What happens

When a Java service handles a request on a virtual thread (JDK 21+, e.g. Spring Boot with `spring.threads.virtual.enabled=true`) and then makes a synchronous outgoing HTTP call on that same thread, OBI usually fails to attach the client span to the server's trace. The downstream call shows up as its own new root trace, and occasionally under a different request's trace. With platform threads the same code produces correctly nested traces.

## Mechanism

OBI correlates a server request with its outgoing calls by the kernel thread id (gettid). The server context is stored in `server_traces` keyed by the tid that handled the read (`bpf/common/trace_lifecycle.h:115`). On egress OBI looks up the current tid in `server_traces` and walks the `java_tasks` parent-tid chain (`bpf/common/trace_parent.h:198-226`). The connect-time fallback (`cp_support_connect_info`, `trace_parent.h:326`) stores the connect-time tid, so it is tid-anchored as well. When nothing is found for the current tid, OBI generates a fresh trace id with a zero parent (`bpf/generictracer/protocol_http.h:644-645`), which is the orphan we see.

A virtual thread runs on a carrier platform thread and can unmount at a blocking point, then remount on a different carrier later. JEP 444 states that native code on the same virtual thread may observe a different OS thread id across calls. So the tid that handled the read and the tid that runs the egress are frequently different, and every Java correlation path misses.

This affects clients that perform the socket I/O on the calling thread, such as `HttpURLConnection` and the classic Apache HttpClient. Clients that offload I/O to their own threads are not affected, because the egress runs on a stable platform thread regardless of the request's virtual thread (reactor-netty's WebClient uses its event-loop threads, and `java.net.http.HttpClient` uses an internal selector thread).

## Reproduction

A controlled A/B with one variable changed, `spring.threads.virtual.enabled`. A minimal Spring Boot service on JDK 21 (Tomcat) handles `/work`, sleeps briefly to yield the carrier, then makes a synchronous `HttpURLConnection` GET to a local nginx returning 200. The handler also reads its own kernel tid from `/proc/thread-self/stat` at entry and again right before the egress, so carrier migration is measured directly and reported in the response (`"migrated"`).

Driving 400 requests at concurrency 40, with full span capture from OBI's text printer:

| config | carrier migration | broken correlation |
|---|---|---|
| platform threads | 0% (0/400) | 0.2% |
| virtual threads | 91.5% (366/400) | 63.3% |

With virtual threads, the carrier kernel thread changed between handler entry and egress in 91.5% of requests, and 63.3% of downstream client spans were not correlated to their server trace (emitted as new root traces). With platform threads neither happens.

The effect needs carrier contention: running the same load sequentially (concurrency 1) with virtual threads on shows 0% breakage, because the handler remounts on the same carrier when no other work competes for it.

Environment: OBI built from commit 3193b9a0, kernel 6.x with BTF, JDK 21 (Temurin 21.0.10), Spring Boot 3.5.11.

### Repro files

Layout: place this folder at `<obi-repo>/hack/vthread-repro/` (so the default build context `../..` points at the repo root), or set `OBI_REPO=/path/to/opentelemetry-ebpf-instrumentation`. Then:

```
VT=false docker compose up -d --build   # baseline
# or
VT=true  docker compose up -d --build   # virtual threads

seq 1 400 | xargs -P 40 -I{} curl -fsS "localhost:18080/work?url=http://downstream/" -o /dev/null
docker compose logs obi | grep "traceparent=\[00-"
```

`verify.sh` runs both arms and prints the comparison.

<details>
<summary>docker-compose.yml</summary>

```yaml
# Minimal, self-contained reproduction of the Java virtual-thread trace-correlation gap.
# Three services: the app under test, a downstream HTTP target, and OBI.
# OBI prints spans to its own stdout (text printer); no collector/Jaeger needed.
#
# The obi service builds from your OBI checkout. Either place this folder at
# <obi-repo>/hack/vthread-repro/ (so the default ../.. resolves to the repo root),
# or run with OBI_REPO=/path/to/opentelemetry-ebpf-instrumentation.
services:
  app:
    build: ./app
    image: vt-repro-app
    ports:
      - "18080:8080"
    environment:
      # the single variable that triggers the bug
      SPRING_THREADS_VIRTUAL_ENABLED: "${VT:-false}"

  downstream:
    image: nginx:1.27-alpine
    expose:
      - "80"

  obi:
    build:
      context: ${OBI_REPO:-../..}
      dockerfile: internal/test/integration/components/obi/Dockerfile-with-javaagent
    image: hatest-obi-b
    privileged: true
    network_mode: host
    pid: host
    volumes:
      - /sys/fs/cgroup:/sys/fs/cgroup
    environment:
      OTEL_EBPF_OPEN_PORT: "8080"
      OTEL_EBPF_TRACE_PRINTER: "text"
      OTEL_EBPF_BPF_CONTEXT_PROPAGATION: "all"
      OTEL_EBPF_DISCOVERY_POLL_INTERVAL: "500ms"
      OTEL_EBPF_LOG_LEVEL: "info"
    depends_on:
      app:
        condition: service_started
```
</details>

<details>
<summary>app/src/main/java/repro/Application.java</summary>

```java
package repro;

import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;
import org.springframework.web.bind.annotation.GetMapping;
import org.springframework.web.bind.annotation.RequestParam;
import org.springframework.web.bind.annotation.RestController;

import java.io.InputStream;
import java.net.HttpURLConnection;
import java.net.URL;
import java.nio.file.Files;
import java.nio.file.Path;
import java.util.Map;

@SpringBootApplication
@RestController
public class Application {

    public static void main(String[] args) {
        SpringApplication.run(Application.class, args);
    }

    // Confirms whether the request handler runs on a virtual thread.
    @GetMapping("/thread-info")
    public Map<String, Object> threadInfo() {
        Thread t = Thread.currentThread();
        return Map.of("virtual", t.isVirtual(), "thread_name", t.getName());
    }

    // Handles the request, yields the carrier with a short sleep, then makes a
    // synchronous downstream HTTP call on the same thread. Reports the kernel tid
    // at entry vs before the egress so carrier migration is visible.
    @GetMapping("/work")
    public Map<String, Object> work(@RequestParam String url) throws Exception {
        boolean virtual = Thread.currentThread().isVirtual();
        long tidEntry = kernelTid();
        Thread.sleep(50);
        long tidEgress = kernelTid();
        HttpURLConnection conn = (HttpURLConnection) new URL(url).openConnection();
        conn.setRequestMethod("GET");
        conn.setConnectTimeout(10000);
        conn.setReadTimeout(10000);
        int code = conn.getResponseCode();
        try (InputStream is = conn.getInputStream()) {
            is.readAllBytes();
        }
        conn.disconnect();
        return Map.of(
            "virtual", virtual,
            "tid_entry", tidEntry,
            "tid_egress", tidEgress,
            "migrated", tidEntry != tidEgress,
            "status", code
        );
    }

    // Kernel thread id (gettid) of the current OS thread, read from procfs (no JNI).
    private static long kernelTid() {
        try {
            String stat = Files.readString(Path.of("/proc/thread-self/stat"));
            return Long.parseLong(stat.substring(0, stat.indexOf(' ')));
        } catch (Exception e) {
            return -1;
        }
    }
}
```
</details>

<details>
<summary>app/pom.xml</summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <parent>
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-starter-parent</artifactId>
        <version>3.5.11</version>
        <relativePath/>
    </parent>

    <groupId>repro</groupId>
    <artifactId>vt-repro</artifactId>
    <version>1.0.0</version>

    <properties>
        <java.version>21</java.version>
    </properties>

    <dependencies>
        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-web</artifactId>
        </dependency>
    </dependencies>

    <build>
        <finalName>vt-repro</finalName>
        <plugins>
            <plugin>
                <groupId>org.springframework.boot</groupId>
                <artifactId>spring-boot-maven-plugin</artifactId>
            </plugin>
        </plugins>
    </build>
</project>
```
</details>

<details>
<summary>app/Dockerfile</summary>

```dockerfile
FROM maven:3.9-eclipse-temurin-21 AS build
WORKDIR /build
COPY pom.xml .
COPY src ./src
RUN mvn -q -e -DskipTests package

FROM eclipse-temurin:21-jre
COPY --from=build /build/target/vt-repro.jar /app.jar
EXPOSE 8080
ENTRYPOINT ["java", "-jar", "/app.jar"]
```
</details>

<details>
<summary>verify.sh (runs both arms)</summary>

```bash
#!/usr/bin/env bash
set -u
cd "$(dirname "$0")"
DC="docker compose"

run() {
  arm=$1
  echo "================ VT=$arm ================"
  $DC down >/dev/null 2>&1
  VT=$arm $DC up -d --build >/dev/null 2>&1
  for i in $(seq 1 120); do curl -fsS localhost:18080/thread-info >/dev/null 2>&1 && break; sleep 1; done
  echo "thread-info: $(curl -fsS localhost:18080/thread-info)"
  echo "waiting 35s for OBI discovery + java agent injection..."
  sleep 35
  curl -fsS "localhost:18080/work?url=http://downstream/" -o /dev/null 2>/dev/null   # warmup
  seq 1 400 | xargs -P 40 -I{} curl -fsS "localhost:18080/work?url=http://downstream/" -o /tmp/repro_resp_${arm}_{} 2>/dev/null
  sleep 12
  $DC logs obi 2>&1 | grep -E "traceparent=\[00-" | grep -E "/work|HTTPClient" > /tmp/repro_spans_${arm}.log
  echo "carrier migration (from app): $(grep -l '"migrated":true' /tmp/repro_resp_${arm}_* 2>/dev/null | wc -l)/$(ls /tmp/repro_resp_${arm}_* 2>/dev/null | wc -l)"
  python3 analyze.py /tmp/repro_spans_${arm}.log
  $DC down >/dev/null 2>&1
  rm -f /tmp/repro_resp_${arm}_*
}

run false
run true
echo "DONE"
```
</details>

<details>
<summary>analyze.py (counts correlated vs orphaned client spans)</summary>

```python
import re, sys

TP = re.compile(r"traceparent=\[00-([0-9a-f]{32})-([0-9a-f]{16})\[([0-9a-f]{16})\]")
servers = {}      # trace_id -> set(server span ids)
clients = []      # (trace_id, span_id, parent_id)

for line in open(sys.argv[1]):
    m = TP.search(line)
    if not m:
        continue
    tr, sp, pa = m.groups()
    if ") HTTP(subType" in line and "/work" in line:
        servers.setdefault(tr, set()).add(sp)
    elif ") HTTPClient(subType" in line:
        clients.append((tr, sp, pa))

n = len(clients)
correct = sum(1 for tr, sp, pa in clients if tr in servers and pa in servers[tr])
orphan = sum(1 for tr, sp, pa in clients if tr not in servers)
nsrv = sum(len(v) for v in servers.values())
print(f"server spans (/work): {nsrv}")
print(f"client spans:         {n}")
if n:
    print(f"  correctly nested:   {correct}")
    print(f"  orphaned:           {orphan}")
    print(f"  broken correlation: {n - correct}/{n} = {100*(n-correct)/n:.1f}%")
```
</details>

## Impact

On JDK 21+ with virtual threads enabled (a single Spring Boot property), distributed traces for synchronous outgoing calls are broken: downstream spans are detached from their originating request, and sometimes attached to the wrong one. The "Java thread pool" row in `SUPPORT_MATRIX.md` does not list a limitation here.

lucacavenaghi97:
Fixes #2242

Java trace correlation is keyed by kernel tid. Under virtual threads
that key is the carrier tid, and correlation breaks. This PR re-keys
Java correlation by the virtual thread logical id, reported by the
agent on mount/unmount.

## What the investigation found

The breakage has three layers:

1. The reported one: on egress the carrier tid differs from the tid
   that read the request (~92% of requests under contention), so the
   tid-keyed lookups miss and the client span orphans.
2. The thread-pool instrumentation poisons `java_tasks` under virtual
   threads: Loom re-submits the per-VT runContinuation lambda through
   the instrumented Executor surface on every unpark, from platform
   threads, writing stale submitter tids onto carrier tids.
3. The dominant one: tid-keyed `server_traces` self-destructs at read
   time. Two in-flight virtual threads reading on the same carrier tid
   build identical trace keys, and the conflicting-span branch
   invalidates the live entry. This accounts for most of the 63-68%
   baseline breakage and is why agent-only fixes cannot work.

## Fix shape

- Agent: instrument `java.lang.VirtualThread.mount()/unmount()` (the
  seam is unchanged across JDK 21..25 and is the same one
  dd-trace-java uses). On mount exit, report the VT id through the
  existing ioctl channel (new op VT_MOUNT); on unmount, delete the
  entry (VT_UNMOUNT). The emit path is lock-free with preallocated
  buffers: a blocking mount path can stall carriers.
- eBPF: a new internal-pinned LRU map `java_vt_threads[carrier
  pid_key] -> vt id`, and a translate helper that rewrites the tid in
  correlation keys to a synthetic per-VT id everywhere a key is built
  from the current tid (HTTP, HTTP/2, TCP, DNS and the
  connection-fallback captures). A carrier with no mounted VT is never
  translated, so platform-thread behavior is unchanged.
- The thread-pool advices skip Loom-internal tasks and submissions
  made from virtual threads (layer 2).
- `traces_ctx_v1` is not translated; VT-handled requests skip writing
  it (rationale below).

## Validation

Per-request oracle, unique ids, parent span id verified, concurrency
40. Baseline on main: 63-68% of downstream client spans orphaned or
cross-wired under virtual threads. With the fix:

| arm | fix | main |
|---|---|---|
| HTTP, JDK 21, 2000 req | 0 cross-wired, 1 orphan | 63-68% broken |
| HTTP, JDK 25 runtime, 2000 req | 0 cross-wired, 1 orphan | |
| platform threads (regression canary), 2000 req | clean, no change | |
| mixed: 64 churning VTs + 800 platform pool-handoff requests on the same two OS threads | 800/800, map empty post-quiesce | |
| TLS downstream, 1000 req | 998/1000, 0 cross-wired | |
| raw TCP (generic-TCP keying path), 800 req | 800/800, server_traces empty post-quiesce | 34.5% (254 cross-wired, 269 orphaned) |
| DNS spans parented to their request's server span | 24/24 | 7/22 |

Notes: the TLS arm needed connection-close per request because of a
pre-existing, VT-independent bug in the blocking-SSLSocket read path
(read-side advice fails on a classloader visibility issue; I will file
it separately with a fix). The mixed-arm numbers are with #2260 in.

## The log-enricher / traces_ctx_v1 decision

`traces_ctx_v1` is keyed by raw pid_tgid (an OTEP contract, consumers
look up by real thread id), so under virtual threads its entries
cannot be correct. Measured per log line with a JSON-logging workload:
on main with VTs, 0% of lines carry their own trace id and 19% carry
ANOTHER request's; keeping the writes with the fix makes it worse (48%
wrong). The patch therefore suppresses the writes while a VT is
mounted: wrong data is worse than no data on a correlation surface.
Platform-thread enrichment is untouched (100% before and after).
Reverting to fresh-by-recency writes is a two-line change if you
prefer that tradeoff; proper VT-aware enrichment needs its own issue.

## What this PR contains

- the agent and eBPF changes above;
- an integration test asserting a nested-client-span ratio under
  concurrent load (single-request assertions pass by luck on broken
  builds): fails on main at 22%, passes with the patch at 97%;
- SUPPORT_MATRIX and devdocs rows for virtual threads.

The experiment harness and raw results are reproducible:
https://github.com/lucacavenaghi97/obi-java-vthread-lab


bpf/common/common.h
```diff
@@ -153,6 +153,8 @@ typedef struct tcp_req {
     u64 start_monotime_ns;
     u64 end_monotime_ns;
     u64 extra_id;
+    u32 task_tid;
```
mariomac, line 156: Shall we document this new field as java-specific? If so, even rename it to something like: `jvm_tid`?

bpf/generictracer/java_tls.c
```diff
@@ -93,6 +96,38 @@ int BPF_KPROBE(obi_kprobe_sys_ioctl) {
         return 0;
     }
 
+    if (op_cmd == k_ioctl_java_vt_mount) {
```
mariomac: Nit. Should we maybe move the following three `if (op_cmd == ...` to a `switch` clause?

internal/test/integration/components/javatestserver/Dockerfile_jar
```diff
@@ -22,7 +24,7 @@ RUN ./mvnw -version
 RUN ./mvnw clean package
 
 # The App Image
-FROM eclipse-temurin:17.0.18_8-jre@sha256:ff692acf872589e72ca71b416b3e949ec19f35ebe8abb5365d2f31bce298e762
+FROM ${JRE_IMAGE}
```
mariomac, line 27: Might this break auto-updates from Renovate Bot?

pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/Agent.java
```diff
@@ -262,6 +265,34 @@ private static void injectBootstrapClasses(Instrumentation instrumentation) thro
...
+
+      // Force-initialize the BOOTSTRAP copies of every class on the
+      // VirtualThread mount-advice emit path NOW: lazy classloading on the
+      // mount path (JarFile synchronized I/O) can pin carriers and deadlock.
+      for (String name :
+          new String[] {
+            "io.opentelemetry.obi.java.ebpf.ThreadInfo",
+            "io.opentelemetry.obi.java.ebpf.IOCTLPacket",
+            "io.opentelemetry.obi.java.ebpf.OperationType",
+            "io.opentelemetry.obi.java.ebpf.NativeMemory",
```
mariomac: To avoid that any refactor (e.g. moving a class or renaming a package) silently breaks this part of the code, could we replace the hardcoded strings by something like?

```
io.opentelemetry.obi.java.ebpf.ThreadInfo.class.getName(),
io.opentelemetry.obi.java.ebpf.IOCTLPacket.class.getName(),
... etc ...
```

pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/Agent.java
```diff
@@ -262,6 +265,34 @@ private static void injectBootstrapClasses(Instrumentation instrumentation) thro
...
+      // Arm ThreadInfo's preallocated VT emit buffers on the BOOTSTRAP copy
+      // (the one the VirtualThread advices resolve), now that its native
+      // library binding is in place: the mount path must never allocate
+      // direct memory (Bits.reserveMemory can System.gc/sleep).
+      // A failed warmup (e.g. direct-memory pressure throwing an Error) must
+      // not disable the whole agent: emitVtOp falls back to one-shot buffers
+      // while the pool is unarmed, so catch Throwable and continue.
+      try {
+        Class<?> bootstrapThreadInfo =
+            Class.forName("io.opentelemetry.obi.java.ebpf.ThreadInfo", true, null);
```
mariomac: Same as previous comment.

mariomac (commented):
This is amazing! I only have few minor comments.

pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/ebpf/ThreadInfo.java
```diff
@@ -19,4 +20,98 @@ public static void sendParentThreadContext(long parentId) {
...
+          }
+          return;
+        }
+      }
+    }
+    // Pool not armed yet (or transiently exhausted): one-shot buffer.
+    NativeMemory p = new NativeMemory(IOCTLPacket.packetPrefixSize);
+    IOCTLPacket.writePacket(p, 0, op, value);
+    Agent.NativeLib.ioctl(0, Agent.IOCTL_CMD, p.getAddress());
+  }
```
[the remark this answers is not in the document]
lucacavenaghi97: Fixed. The mount/unmount path no longer allocates: it drops the emit if the pool is unarmed or exhausted, and the pool is now sized from `availableProcessors`. A dropped mount falls back to carrier-tid keying; a dropped unmount is reaped by the `sys_exit` backstop + LRU.

bpf/common/common.h
```diff
@@ -153,6 +153,8 @@ typedef struct tcp_req {
...
+    u8 _pad3[4];
     u32 req_len;
```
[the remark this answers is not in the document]
lucacavenaghi97: This aligns the trailing `tp_info_t`, not `task_tid`. Without it the build fails `-Wpadded`: `padding struct 'tcp_req' with 4 bytes to align 'tp'`.

bpf/common/common.h
lucacavenaghi97: I'd keep `task_tid`: it mirrors the existing `http_info_t.task_tid` (same delete-key role). It's also not really a JVM tid, on platform threads it's the raw kernel tid and only gets VT-translated under a mounted virtual thread, so `jvm_tid` would be misleading. Happy to rename both in a follow-up if you'd rather.

bpf/generictracer/java_tls.c
lucacavenaghi97: Done

internal/test/integration/components/javatestserver/Dockerfile_jar
lucacavenaghi97: Renovate expands `ARG` declared before `FROM` and tracks the ARG default. We already do this in `Dockerfile_nginx` and the root `Dockerfile`, and renovate[bot] has bumped both, so it stays updated.

pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/Agent.java
lucacavenaghi97: Done, both use `X.class.getName()` now.

pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/Agent.java
lucacavenaghi97: Same as above, done.

pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/ebpf/ThreadInfo.java
```diff
@@ -19,4 +20,105 @@ public static void sendParentThreadContext(long parentId) {
...
+  // (the one the VirtualThread advices resolve).
+  public static void initVtEmitPool() {
+    for (int i = 0; i < VT_POOL_SIZE; i++) {
+      vtEmitPool.set(i, new NativeMemory(IOCTLPacket.packetPrefixSize));
+    }
+    vtEmitPoolReady = true;
+  }
+
+  private static void emitVtOp(OperationType op, long value) {
+    if (!vtEmitPoolReady) {
```
grcevski, line 90: Since there might be tens of thousands virtual threads, is there anything we can do here to avoid this linear scan? I wonder if we can make a hash and if we fail hashing we simply allocate new NativeMemory?

I have no idea if that's better.

grcevski (approved):
LGTM! Great work @lucacavenaghi97! I just had one question around the efficiency of the native memory pool approach.

bpf/maps/java_vt_threads.h
```diff
@@ -0,0 +1,60 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include "pid/types/pid_key.h"
```
mmat11:
```suggestion
#include <pid/types/pid_key.h>
```

import order should also be fixed

mmat11:
> The log-enricher / traces_ctx_v1 decision
>
> traces_ctx_v1 is keyed by raw pid_tgid (an OTEP contract, consumers
look up by real thread id), so under virtual threads its entries
cannot be correct. Measured per log line with a JSON-logging workload:
on main with VTs, 0% of lines carry their own trace id and 19% carry
ANOTHER request's; keeping the writes with the fix makes it worse (48%
wrong). The patch therefore suppresses the writes while a VT is
mounted: wrong data is worse than no data on a correlation surface.
Platform-thread enrichment is untouched (100% before and after).
Reverting to fresh-by-recency writes is a two-line change if you
prefer that tradeoff; proper VT-aware enrichment needs its own issue.

I'm not familiar with Java, so I have a question: is there no way to refresh the `traces_ctx_v1` map on virtual threads handoffs?

pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/ebpf/ThreadInfo.java
lucacavenaghi97: Thanks! I think it's ok here because the pool is sized to the number of carriers, not the number of virtual threads: a VT only emits while it's mounted on a carrier, so at most about as many emits as carriers can overlap at once. The pool is roughly double that, so it stays about half full and a free buffer is almost always there on the first try. The starting slot is already picked by hashing the thread id, the loop is just the fallback when that one's taken.

On allocating when nothing's free: that's the path I removed earlier, since allocating a NativeMemory there can block in `Bits.reserveMemory` (gc/sleep) and stall the carrier inside `mount()`. So I went with dropping instead, a missed mount just falls back to carrier-tid keying and a missed unmount gets cleaned up by the `sys_exit` backstop. Let me know if you think that carrier bound doesn't hold though.

lucacavenaghi97:
Hi @mmat11 ! It's doable in principle, but not a small change. The agent's mount hook only knows the virtual thread id, not the trace context (that lives in the eBPF maps), so the refresh would have to happen kernel-side on every mount/unmount, copying the VT's context into the carrier-keyed slot and saving/restoring whatever was there before, since virtual threads share carriers. And even then it wouldn't catch trace changes that happen while the VT stays mounted, which is exactly what the normal write path does and what goes wrong under virtual threads.

So for now I went with suppressing rather than writing wrong data (platform threads are untouched). I'll open an issue to track proper VT-aware enrichment and look into it there.

grcevski:
I think adding log enrichment support in a follow-up is a good idea.

grcevski:
Thanks a lot for this work @lucacavenaghi97 !

lucacavenaghi97:
Thanks for reviewing it! Happy to contribute

PR #2263, opened 2026-06-06
fix(2262): inline IOCTLPacket calls in the SSLSocket read advices

### LINKED ISSUE #2262: Java TLS client spans are mostly not captured for blocking SSLSocket clients reusing keep-alive connections
## What happens

A Java app making synchronous HTTPS calls with `HttpsURLConnection` (the
JSSE blocking `SSLSocket` path, also used by classic Apache HttpClient)
over HTTP/1.1 keep-alive connections gets almost no client spans: only the
first request on each TLS connection produces one. Subsequent requests on
the reused connection produce nothing at all.

Measured on current main (d41ba1a5), platform threads only, Spring Boot app
calling an nginx TLS downstream at concurrency 40, 400 requests with unique
URL path ids:

- client spans emitted: 24 (roughly the number of distinct TLS connections)
- requests with NO client span: 374 (94%)

Forcing the downstream to close the connection after every response
(`keepalive_timeout 0`), same app, same OBI build: 300/300 requests get a
client span, all correctly parented to their server span.

A second facet, visible even on fresh connections: the TLS response is
never parsed on this path. Every captured client span has status 499 and
responseLen 0 (the request side is parsed fine, the URL path is correct).

Plain-HTTP client calls from the same app capture 100% of spans with
correct statuses, so this is specific to the agent-fed TLS path. The
existing Java TLS integration tests use netty/WebClient (the `SSLEngine`
path), which is why CI does not see the blocking-`SSLSocket` path.

## Why

Root cause, verified end to end with the agent debug flags
(OTEL_EBPF_JAVAAGENT_DEBUG + DEBUG_INSTRUMENTATION) and a BPF trace-pipe
capture: **the read-side advice of `SSLSocketStreamInst` always throws
`NoClassDefFoundError` and the decrypted response is never delivered to
eBPF.**

Chain of evidence:

1. The weaving itself is correct. `HttpsURLConnection` uses
   `sun.security.ssl.SSLSocketImpl`; `SSLSocketInst` deliberately skips it
   (SSLSocketInst.java:24) and `SSLSocketStreamInst` covers
   `SSLSocketImpl$AppInputStream`/`$AppOutputStream` (verified against JDK
   21 sources; ByteBuddy listener confirms the transform completes).
2. Both advices FIRE. Per request, the agent debug shows
   `OutputStream.write(off,len) intercepted: 110 bytes` AND
   `InputStream.read(off,len) intercepted: 145 bytes` (the response).
3. The read advice then dies:
   `Error in read advice: io/opentelemetry/obi/java/instrumentations/SSLSocketStreamInst`
   - a `NoClassDefFoundError`. The advice bytecode is inlined into
   `AppInputStream`, a BOOTSTRAP-classloader class, but it calls the static
   helper `SSLSocketStreamInst.writeReadPacket(...)`
   (SSLSocketStreamInst.java:84-87,113,149) and `SSLSocketStreamInst`
   lives in the app classloader - invisible from bootstrap. The write
   advices work because they inline `IOCTLPacket`/`NativeMemory`/`Agent`
   calls directly, and those classes ARE bootstrap-injected
   (Agent.injectBootstrapClasses). The `catch (Throwable)` swallows the
   error unless debug is on, so nothing is ever logged.
4. Kernel-side confirmation: a trace-pipe capture over 10 requests shows
   exactly 10 ioctls with `op=1` (TCP_SEND, the request) and ZERO with
   `op=2` (TCP_RECV). The BPF side never has a chance.

Downstream consequences, which produce the observed numbers:

- With the response never seen, `handle_http_response` never runs for
  these requests; on connection close the request is force-finished as
  status 499 / responseLen 0 (`force_finish_http`,
  protocol_http.h:142-156). Hence every captured span is 499/0.
- On keep-alive reuse the never-completed request also wedges the
  per-connection state, so subsequent requests on the same connection do
  not produce spans at all (~1 span per connection, 94% loss at our
  request/connection ratio); closing the connection per request bypasses
  the wedge and capture becomes 100%.
- The Java ioctl path also lacks the SSL-aware delayed-finish that the
  OpenSSL path has (`finish_possible_delayed_tls_http_request` is wired
  only via `ssl_to_conn`, k_tracer.c:381-396), which matters once the read
  side is fixed: responses will be delivered but their submission still
  rides on next-request/close events.

The fix for the dominant facet is small: make the read advices inline the
`IOCTLPacket.writePacketPrefix(...RECEIVE...)` +
`IOCTLPacket.writePacketBuffer(...)` calls exactly like the write advices
do, instead of routing through a non-bootstrap static helper. Validated on
the same arm: capture goes from 24/398 (all status 499) to 388/400 (all
status 200, every span parented to its own request's server span); the
remaining 12 are the last in-flight request per keep-alive connection,
which still waits for the next event on the connection to be submitted
(the delayed-finish facet above, out of scope here).

Why CI never sees this: the upstream Java TLS integration tests use
netty/WebClient (`SSLEngine` path, `NettySSLHandlerInst` - woven into
app-classloader netty classes, where the helper resolution works). Recent
fixes #2218/#2229 touched the SSL receive path (issues #2019/#2025), but
nothing documents this; it appears to be an unrecognized gap, not an
intentional limitation.

## Repro

Same triangle as the repro in #2242 / #2259 (app, nginx downstream, OBI
with the text printer): give nginx a TLS server block with a self-signed
cert, point the app's `HttpsURLConnection` (trust-all for the test cert) at
`https://downstream/echo/{id}`, drive unique ids, and compare ids that have
a client span against ids that do not. OBI env: `OTEL_EBPF_OPEN_PORT=8080`,
`OTEL_EBPF_TRACE_PRINTER=text`, `OTEL_EBPF_BPF_CONTEXT_PROPAGATION=all`.
Toggle `keepalive_timeout 0;` on the nginx TLS block to see capture go from
~6% to 100%.

lucacavenaghi97:
Resolves #2262

## Problem

#2262: a Java app making blocking HTTPS calls (`HttpsURLConnection` or classic Apache HttpClient, the JSSE `SSLSocket` path) over keep-alive connections gets almost no client spans, and the few captured ones have status 499 with an empty response.

The read advices of `SSLSocketStreamInst` are inlined by ByteBuddy into `sun.security.ssl.SSLSocketImpl$AppInputStream`, a bootstrap-classloader class, but they called the static helper `SSLSocketStreamInst.writeReadPacket(...)`. `SSLSocketStreamInst` is not bootstrap-injected (only the classes force-loaded by `Agent.initClassesThatNeedToBeBootstrapped` are), so every read advice threw `NoClassDefFoundError`, silently swallowed by the `Throwable` handler, and the decrypted response never reached eBPF. Full analysis in the issue.

## Fix

Inline the `IOCTLPacket.writePacketPrefix(RECEIVE)` + `writePacketBuffer` calls in both read advices, exactly like the working write advices do, and delete the helper. The inlined code only references bootstrap-injected classes; a comment documents the constraint. The existing unit test (from #2218) keeps guarding the bytes-read semantics through the same call sequence the advices now inline.

## Validation

Per-request oracle (Spring Boot app calling an nginx TLS downstream with keep-alive, unique URL path ids, platform threads, concurrency 40), same main checkout:

| build | client spans | statuses |
|---|---|---|
| main | 24/398 (374 requests with no span) | all 499, responseLen 0 |
| this fix | 388/400, every span parented to its own request's server span | all 200 |

The remaining 12 are the last in-flight request per keep-alive connection, submitted only on the next event on that connection (the delayed-finish facet noted in the issue, out of scope here).

The existing Java TLS integration tests do not cover this path (they use netty/WebClient, the `SSLEngine` path, where helper resolution works); unit tests and spotless pass.

### DIFF
--- a/pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/instrumentations/SSLSocketStreamInst.java
+++ b/pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/instrumentations/SSLSocketStreamInst.java
@@ -81,11 +81,9 @@ public class SSLSocketStreamInst {
                             .and(ElementMatchers.takesArgument(0, byte[].class))));
   }
 
-  static int writeReadPacket(NativeMemory p, Socket socket, byte[] b, int off, int len) {
-    int wOff = IOCTLPacket.writePacketPrefix(p, 0, OperationType.RECEIVE, socket, len);
-    return IOCTLPacket.writePacketBuffer(p, wOff, b, off, len);
-  }
-
+  // The advice bytecode below is inlined into sun.security.ssl classes, which live in the
+  // bootstrap classloader: it must only reference bootstrap-injected classes (see
+  // Agent.injectBootstrapClasses), never SSLSocketStreamInst itself.
   public static final class InputStreamReadAdvice {
     @Advice.OnMethodExit(suppress = Throwable.class)
     public static void read(
@@ -110,7 +108,8 @@ public class SSLSocketStreamInst {
         if (socket != null) {
           try {
             NativeMemory p = new NativeMemory(IOCTLPacket.packetPrefixSize + len);
-            writeReadPacket(p, socket, b, 0, len);
+            int wOff = IOCTLPacket.writePacketPrefix(p, 0, OperationType.RECEIVE, socket, len);
+            IOCTLPacket.writePacketBuffer(p, wOff, b, 0, len);
             Agent.NativeLib.ioctl(0, Agent.IOCTL_CMD, p.getAddress());
           } catch (Throwable t) {
             if (SSLStorage.debugOn) {
@@ -146,7 +145,9 @@ public class SSLSocketStreamInst {
         if (socket != null) {
           try {
             NativeMemory p = new NativeMemory(IOCTLPacket.packetPrefixSize + bytesRead);
-            writeReadPacket(p, socket, b, off, bytesRead);
+            int wOff =
+                IOCTLPacket.writePacketPrefix(p, 0, OperationType.RECEIVE, socket, bytesRead);
+            IOCTLPacket.writePacketBuffer(p, wOff, b, off, bytesRead);
             Agent.NativeLib.ioctl(0, Agent.IOCTL_CMD, p.getAddress());
           } catch (Throwable t) {
             if (SSLStorage.debugOn) {
--- a/pkg/internal/java/agent/src/test/java/io/opentelemetry/obi/java/instrumentations/SSLSocketStreamInstTest.java
+++ b/pkg/internal/java/agent/src/test/java/io/opentelemetry/obi/java/instrumentations/SSLSocketStreamInstTest.java
@@ -9,6 +9,8 @@ import static org.junit.jupiter.api.Assertions.*;
 
 import io.opentelemetry.obi.java.ebpf.IOCTLPacket;
 import io.opentelemetry.obi.java.ebpf.NativeMemory;
+import io.opentelemetry.obi.java.ebpf.OperationType;
+import java.net.Socket;
 import org.junit.jupiter.api.Test;
 
 class SSLSocketStreamInstTest {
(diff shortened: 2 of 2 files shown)


mariomac (approved):
Thanks for your contribution! The explanation and the fix is clear. I approved because I don't think the change can break anything, but before merging I'll anyway let someone else to double-check the changes, as ByteBuddy internals go beyond my knowledge.

MrAlias (approved):
Looks good. I did a compatibility-focused pass on this change and don’t see a blocker.

I verified this advice is inlined into `sun.security.ssl.SSLSocketImpl$AppInputStream`, which is loaded by the bootstrap classloader. That means the inlined method body must not depend on `SSLSocketStreamInst` or any other agent-loader-only class. This patch satisfies that constraint: the read advice now calls only `IOCTLPacket`, `OperationType`, `NativeMemory`, `SSLStorage`, `Agent`, and `Agent.NativeLib`, which are already part of the bootstrap preload/injection path.

I also checked the compiled bytecode and the ByteBuddy-transformed method shape. The executable read-advice path no longer references `SSLSocketStreamInst`, so this addresses the `NoClassDefFoundError` failure mode described in the issue.

The JDK compatibility shape also looks right. On the checked JDKs, `SSLSocketImpl$AppInputStream` exposes `read(byte[], int, int)`, which is the path this fix corrects. The agent still compiles to Java 8 bytecode.

The acknowledged coverage gap for runtime coverage is not ideal, but overall, I’m comfortable with the fix given it's targeted scope.

PR #2278, opened 2026-06-09
Deprecate ip context propagation value and remove undocumented http alias

pinoOgni:
## Summary

This PR deprecates the `ip` value (IP options injection was already removed), the runtime now emits a warning and no-ops instead of rejecting the config.

It also removes "http" from the parser. It is technically a breaking change for anyone who had `context_propagation: http` in their config, but `http` was never documented or surfaced in any schema, so exposure is expected to be zero.

Finally, it also updates JSON schemas (main and v2) and the doc generator so deprecated values are surfaced in the "Deprecated" column ("deprecated values:...") of generated docs rather than silently dropped from the values list.

Note: it fixes also the doc: `disabled` is the default, not `all`.

## Validation

### DIFF
--- a/cmd/config-docs/main.go
+++ b/cmd/config-docs/main.go
@@ -689,22 +689,22 @@ func (g *DocGenerator) valuesString(original, resolved *Schema) string {
 		return ""
 	}
 
-	// Collect enum values from oneOf
+	// Collect enum values from oneOf (skip deprecated branches)
 	if len(s.OneOf) > 0 {
-		var allEnums []string
+		var allEnums, descs []string
 		for _, alt := range s.OneOf {
+			if alt.Deprecated {
+				continue
+			}
 			for _, e := range alt.Enum {
 				allEnums = append(allEnums, fmt.Sprintf("`%v`", e))
+			}
+			if alt.Description != "" {
+				descs = append(descs, alt.Description)
 			}
 		}
 		if len(allEnums) > 0 {
 			return strings.Join(unique(allEnums), ", ")
-		}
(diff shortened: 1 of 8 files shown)


rafaelroquetto (commented):
Non-rhetorical question: currently headers mean both http and http2/grpc injections - this is why we picked 'headers', as it's a more generic name that can be expanded to other protocols should that ever become the case - is there any particular use case why we wanna introduce the `http` alias? Otherwise I'd suggest leaving it out

pinoOgni:
> Non-rhetorical question: currently headers mean both http and http2/grpc injections - this is why we picked 'headers', as it's a more generic name that can be expanded to other protocols should that ever become the case - is there any particular use case why we wanna introduce the `http` alias? Otherwise I'd suggest leaving it out

I saw that `http` was already accepted in `UnmarshalText` function. I just made the docs and schemas consistent with what the code was already doing. Any solution is fine for me.

rafaelroquetto (approved):
@pinoOgni ohh I see. I could be wrong, but I believe it was there only on UnmarshallText for backwards compatibility (so when it's marshalled again, it gets normalised to `headers`) - I think we should remove it, to keep things simpler.

pinoOgni:
> @pinoOgni ohh I see. I could be wrong, but I believe it was there only on UnmarshallText for backwards compatibility (so when it's marshalled again, it gets normalised to `headers`) - I think we should remove it, to keep things simpler.

yep, you are right. I'll remove `http`.

grcevski (approved):
LGTM!

PR #2279, opened 2026-06-09
(DynamicSelector-Phase2) Extend dynamic selection to network and stats

damemi:
## Summary

Phase 2 of https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2234

This adds global dynamic pid selection to network and stats metrics.

The previous PR https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2250 moved the DynamicSelector from an AppO11y field to a global ContextField so it can be accessed by other pipelines. This means that any PID added via `selector.AddPIDs()` will be enabled for traces, network, and stats.

It works by adding a new filter node used by network and stats metrics to check if the data is in the allowed list of PIDs by IP. It adds a new IP-PID mapper helper, which keeps track of which PIDs are associated with which IP. It keeps a count of how many PIDs map to an IP, so that process exit events (such as sidecars) don't necessarily remove the IP from the map.

## Validation

### DIFF
--- a/pkg/filter/dynamic_pid.go
+++ b/pkg/filter/dynamic_pid.go
@@ -1,0 +1,63 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package filter // import "go.opentelemetry.io/obi/pkg/filter"
+
+import (
+	"context"
+
+	"go.opentelemetry.io/obi/pkg/internal/pipe"
+	"go.opentelemetry.io/obi/pkg/kube"
+	"go.opentelemetry.io/obi/pkg/pipe/msg"
+	"go.opentelemetry.io/obi/pkg/pipe/swarm"
+	"go.opentelemetry.io/obi/pkg/pipe/swarm/swarms"
+	"go.opentelemetry.io/obi/pkg/selection"
+)
+
+// ByDynamicPID provides a pipeline node that keeps only records whose source or destination IP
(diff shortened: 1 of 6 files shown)


pkg/filter/dynamic_pid.go
```diff
@@ -0,0 +1,63 @@
...
+// belongs to a dynamically selected application (via DynamicPIDSelector). When the selector is nil,
+// the node is bypassed.
+func ByDynamicPID[T any](
+	ctx context.Context,
+	selector selection.PIDSelector,
+	k8sInformer *kube.MetadataProvider,
+	attrs func(T) *pipe.CommonAttrs,
+	input, output *msg.Queue[[]T],
+) swarm.InstanceFunc {
+	return func(_ context.Context) (swarm.RunFunc, error) {
```
mariomac: I think you should remove the `ctx` as first argument of `ByDynamicPID[T any]` and just give a name to this argument.

There is no guarantee that the context passed to `instancer := ByDynamicPid[T](ctx)` and the context passed to `runFunc, err := instancer(ctx)` will remain being the same (and, for example, the instancer invocation could miss any context cancellation and delay the OBI shutdown after an instantiation error).

mariomac (approved):
LGTM! I just have a minor observation.

pkg/filter/dynamic_pid.go
damemi: yeah that's a good point, thanks!

PR #2269, opened 2026-06-08
Expand internal config v2 capture export parity

MrAlias:
## Summary

Adds the capture-focused export conversion slice from #2251:

- exports discovery rules and per-rule refinements for selector export modes and HTTP routes
- exports application, network, and stats filters
- exports HTTP routes, payload extraction, HTTP enrichment policy/rules, and GenAI payload extraction
- exports SQL MSSQL fields plus advanced network capture/stats CIDRs, enrichment, diagnostics, and `guess_ports`
- exports capture engine map-reader and global map scale-factor fields
- omits empty rule `refine` blocks from schema YAML output
- keeps the conversion internal-only and compatible with `internal/config/schema`

## Scope

This PR intentionally does not add public config v2 APIs, CLI commands, runtime v2 parsing, Collector receiver v2 parsing, generated artifacts, or user-facing config v2 docs.

The top-level resource/provider, capture telemetry, enrich, correlation, and daemon Prometheus telemetry mappings are deferred to the next export-parity PR, as noted in #2251.

## Tests

- `go test ./internal/config/...`
- `go test ./cmd/check-config-v2-parity`
- `make check-config-v2-parity`

Issue #2251, opened 2026-06-04
Config v2 steel-thread landing plan


MrAlias:
Reference PoC: #2080

## Summary

#2080 proves the config v2 steel thread, but it is too large to review and should not be treated as the final implementation shape. The plan is to extract it into serial, reviewable PRs. Each implementation PR is based on `main` after the previous PR merges.

Because this may span multiple releases, no public config v2 behavior should ship until the final gated release group. Early PRs must stay internal: no public Go package, no CLI, no runtime auto-detection, no Collector receiver support, and no user-facing docs that tell users to use config v2.

## Parallel Workstream: Config Drift CI

This can happen independently of the implementation split.

- Move the existing parity checker into `cmd/check-config-v2-parity`.
  - Treat this mostly as a move, not new review burden.
  - Keep it simple and short-lived.
- Add `make check-config-v2-parity`.
- Wire the target into PR checks.
- Fail CI when v1 defaults and merged config v2 definitions drift unintentionally.

Estimated review burden: ~60-150 changed lines, plus moved checker code.

## Serial Implementation PRs

Review-size rule: target ~300 changed logic/test lines; do not exceed ~800. Pure moves and generated artifacts do not count as review burden when they are not mixed with logic changes.

1. **Internal config schema package**
   - Add hidden parser/model under `internal/config/schema`.
   - Include v2 document/extension structs, version detection, standalone vs receiver validation, and tests.
   - No `pkg/config/v2`.
   - Estimate: ~600 lines.

2. **Internal v1-to-v2 conversion foundation**
   - Add `internal/config/convert`.
   - Convert default/runtime config into hidden v2 document shape for core capture fields, protocol enablement, basic network capture, and daemon basics.
   - Estimate: ~500-700 lines.

3. **Export conversion parity**
   - Split the original export parity scope into two review-sized PRs:
     - **Capture export parity** (#2269): discovery rule export, rule refinements, application/network/stats filters, HTTP routes, HTTP enrichment, GenAI payload extraction, SQL MSSQL fields, network CIDRs, network enrichment, stats, diagnostics, and capture engine additions.
     - **Standalone/enrichment/pipeline export parity** (#2288): top-level `resource`, `tracer_provider`, `meter_provider`, capture telemetry reporter cache/TTL fields, Kubernetes/name-resolution/attribute enrichment, correlation/log trace annotation, and daemon Prometheus telemetry fields.
   - Keep changes internal and test-driven.
[...]

PR #2280, opened 2026-06-09
Expand config v2 capture export docs

MrAlias:
## Summary

Follow-up for open-telemetry/opentelemetry-ebpf-instrumentation#2269 and the capture export cleanup notes in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2251#issuecomment-4652952564.

- document the effective config v2 `capture.rules` export shape for v1 discovery selectors, including legacy regex selectors and non-namespace Kubernetes metadata
- expand the config v2 default example and extension schema for payload extraction, MSSQL, network capture cleanup, TCP stats, map reader selection, and map scale factor
- extend the config v2 default parity checker so these documented defaults stay tied to current v1 defaults

## Validation

- `jq empty devdocs/config/version-2.0/obi-extension.schema.json`
- `make lint-markdown`

### DIFF
--- a/cmd/check-config-v2-parity/main.go
+++ b/cmd/check-config-v2-parity/main.go
@@ -316,10 +316,96 @@ func mustMapNetworkFiltersPerSignal(cur map[string]any, ex map[string]any) error
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


grcevski (approved):
LGTM!

PR #2282, opened 2026-06-09
Document directional config v2 HTTP route refinements

MrAlias:
## Summary

This updates the config v2 docs and schema for per-workload HTTP route refinements to preserve route direction explicitly.

The previous draft shape used `refine.http.routes.patterns`, which is directionless. That does not safely represent the v1 per-selector route config because v1 distinguishes routes for requests handled by a workload from routes for requests made by that workload.

The new shape is:

```yaml
refine:
  http:
    routes:
      incoming:
        patterns: []
      outgoing:
        patterns: []
```

`incoming` maps to requests handled by the matched workload, and `outgoing` maps to requests made by the matched workload.

## Details

- Updates the config v2 design doc and default example to use direction-scoped route refinements.
- Updates the v2 extension JSON schema to accept `incoming.patterns` and `outgoing.patterns` under per-workload `http.routes`.
- Removes per-workload `unmatched`, directionless `patterns`, and `ignored_patterns` from the refinement schema for now.
- Leaves global `capture.instrumentation.http.routes` unchanged.

This is intentionally limited to the docs/schema contract. Import/export runtime parity for v1 `routes.incoming` and `routes.outgoing` remains a later config-v2 slice.

## Tests

- `go test ./internal/config/schema`
- `python3 devdocs/config/version-2.0/validate_example.py`
- `make lint-markdown`

### DIFF
--- a/devdocs/config/version-2.0/examples/default-configuration.yaml
+++ b/devdocs/config/version-2.0/examples/default-configuration.yaml
@@ -168,12 +168,15 @@ extensions:
         #       traces: true
         #       metrics: false
         #     http:
-        #       # Per-workload HTTP route patterns (override global patterns for this service).
+        #       # Per-workload HTTP route patterns for requests handled by or made by this workload.
         #       routes:
-        #         unmatched: wildcard
-        #         patterns:
-        #           - /cart/{id}
-        #           - /cart/{id}/items
+        #         incoming:
+        #           patterns:
+        #             - /cart/{id}
+        #             - /cart/{id}/items
+        #         outgoing:
+        #           patterns:
+        #             - /checkout/{id}
         #       # Per-workload HTTP filters (replace global filters for this service).
         #       filters:
         #         traces:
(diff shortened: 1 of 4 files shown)


devdocs/config/version-2.0/examples/default-configuration.yaml
```diff
@@ -168,12 +168,15 @@ extensions:
         #       traces: true
         #       metrics: false
         #     http:
-        #       # Per-workload HTTP route patterns (override global patterns for this service).
+        #       # Per-workload HTTP route patterns for requests handled by or made by this workload.
         #       routes:
-        #         unmatched: wildcard
```
grcevski, line 173: So I guess we are going to have the unmatched strategy global, I think that makes sense.

grcevski (approved):
LGTM!

devdocs/config/version-2.0/examples/default-configuration.yaml
MrAlias: Yep, exactly. I’m keeping `unmatched` global for this slice.

The v1 selector-level route config only has `incoming` and `outgoing` patterns, and the fallback behavior is currently global in the route provider. Keeping `unmatched` under `capture.instrumentation.http.routes` avoids documenting per-workload fallback semantics we do not have yet.

If we decide we need per-workload fallback later, I think we can add it deliberately with explicit precedence semantics.

mariomac (approved):
LGTM

PR #2292, opened 2026-06-10
fix(gotracer): respect context_propagation for Go HTTP/gRPC header in…

### LINKED ISSUE #2276: Disabling context_propagation doesn't disable gRPC traceparent header injection (golang)
### Component

OBI (eBPF instrumentation)

### What happened?

Hello folks, 

Noticed that regardless of `context_propagation` flag my go app injects `traceparent` header in case of grpc protocol.

Notes:
- I've been using `ebpf-instrument:main` and I'm aware that support for http2/grpc was added just recently.
- I'm not sure if this happens with non-go apps.
- I'm not sure if this happens with non-grpc traffic.

Setup:
- amd64 architecture
- kernel 6.12.68+
- go 1.25
- grpc v1.81.1
- image otel/ebpf-instrument:main

Expected that `traceparent` header is not injected when the flag is disabled.

### Steps to reproduce

1. Deploy OBI with context propagation disabled
```
ebpf:
      context_propagation: disabled
      track_request_headers: false
```

2. Then deploy two simple golang apps (client and server), where server logs request metadata, something like this:
```
func (s *server) Ping(ctx context.Context, req *pb.PingRequest) (*pb.PingResponse, error) {
	if md, ok := metadata.FromIncomingContext(ctx); ok {
		log.Printf("ALL METADATA: %v", md)
		if tp := md.Get("traceparent"); len(tp) > 0 {
			fmt.Printf(">>> TRACEPARENT RECEIVED: %s\n", tp[0])
		}
	} else {
		log.Printf("No metadata in request")
	}
	return &pb.PingResponse{Data: []byte("pong")}, nil
}
```

server app logs:
```
2026/06/09 07:04:33 ALL METADATA: map[:authority:[grpc-server:9090] content-type:[application/grpc] traceparent:[00-b39d8d08493afd5eb181267c6c45d8e9-79434a69bf84b013-01] user-agent:[grpc-go/1.81.1]]
```

### OBI configuration

```yaml

```

### Relevant logs

```shell

```

### Additional context

_No response_

### Version

_No response_

### Kernel version

_No response_

### OS / Distribution

_No response_

### Deployment

Kubernetes

mohd-ibadullah:
## Summary

When `context_propagation` is disabled, Go HTTP/gRPC clients still inject `traceparent` headers because the Go-specific propagation path only checks kernel capabilities and ignores the configured propagation mode.

This change aligns `gotracer` behavior with `tpinjector` and `generictracer` by requiring header propagation to be explicitly enabled before registering propagation probes or enabling header injection.

Fixes #2276

## Changes

* Add `headerPropagationEnabled()`
* Gate `g_bpf_header_propagation` on both kernel support and `ContextPropagation.HasHeaders()`
* Align `g_bpf_traceparent_enabled` with `generictracer`
* Register HTTP/gRPC propagation probes only when header propagation is enabled
* Add regression tests covering propagation constants and probe registration

## Validation

* Added unit tests for disabled, tcp, headers, and all propagation modes
* Verified probe registration behavior matches configured propagation mode
* Verified Go gRPC instrumentation remains registered independently of propagation mode


mmat11:
looks like a lot of tests are failing, they probably need explicit config for headers context propagation

mmat11 (approved):
needs linter fix and copilot comment looks valid, lgtm after that

grcevski:
Go tracer test isn't properly formatted, all you need to do is `make fmt` and push the changes.

pkg/internal/ebpf/gotracer/gotracer_test.go
```diff
@@ -1,21 +1,192 @@
...
+func (noopServiceFilter) BlockPID(app.PID, uint32)                                     {}
+func (noopServiceFilter) ValidPID(app.PID, uint32, ebpfcommon.PIDType) bool            { return false }
+func (noopServiceFilter) Filter(spans []request.Span) []request.Span                   { return spans }
+func (noopServiceFilter) CurrentPIDs(ebpfcommon.PIDType) map[uint32]map[app.PID]svc.Attrs {
+	return nil
+}
+
+func testTracer(t *testing.T, cfg *obi.Config) *Tracer {
+	t.Helper()
+	return New(noopServiceFilter{}, cfg, nil, msg.NewQueue[[]runtimemetrics.RuntimeMetricSnapshot]())
```
MrAlias, line 42: This breaks the build. `New` only accepts three arguments: service filter, config, and metrics reporter. Please fix the syntax error.

MrAlias (commented):
@mohd-ibadullah are you still able to work on this?

If so, please address the build failures and the Copilot feedback which looks to be valid.

MrAlias:
Thanks for the contribution. I'm closing this as it looks abandoned. Please re-open if you are able to keep working on this and address the feedback that has been provided.

PR #2305, opened 2026-06-11
Add HotSpot JVM memory metrics via GCTracer uprobe and USDT probes

REASY:
Contributes to grafana/beyla#1608.

  This PR adds HotSpot JVM runtime metrics support to OBI. The remaining Beyla-facing product wiring will be handled in grafana/beyla.

  ## Summary

  Adds JVM runtime metric collection using:

  - a uprobe on HotSpot `GCTracer::report_gc_heap_summary`
  - HotSpot USDT probes:
    - `hotspot:mem__pool__gc__begin`
    - `hotspot:mem__pool__gc__end`

  It also adds the generic USDT attach path needed for this work: `.note.stapsdt` parsing, USDT argument spec parsing, BPF-side argument lookup, semaphore/ref-counter support, and
  cleanup of USDT IP map entries when links are closed.

  ## Details

  JVM runtime events are emitted from eBPF, decoded with generated BPF structs, converted into `RuntimeMetricSnapshot` JVM sections, and routed through the shared runtime metrics path
  used by both the Go tracer and generic tracer.

  Collection is controlled by:

  - `jvm_runtime_metrics.enabled`
  - `jvm_runtime_metrics.sampling_interval`

  Export is controlled by the `application_jvm` metrics feature.

  The BPF side uses `jvm_sampling_interval_ns` as both the sampling interval and feature gate, so disabled JVM runtime metrics do not do USDT argument reads. The BPF side uses `jvm_sampling_interval_ns` as both the sampling interval and feature gate, so disabled JVM runtime metrics do not do USDT argument reads. USDT spec lookups are keyed by PID, PID namespace ID, and instruction pointer. Userspace inserts both the host PID and `NSpid` aliases for the target process, scoped by PID namespace ID, so lookups work whether `valid_pid()` returns the host PID or a namespace PID. Emitted events still carry namespace PID and PID namespace ID for service decoration.

  Metrics added:

  - `jvm.memory.used` / `jvm_memory_used_bytes`
  - `jvm.memory.committed` / `jvm_memory_committed_bytes`
  - `jvm.memory.limit` / `jvm_memory_limit_bytes`
  - `jvm.memory.used_after_last_gc` / `jvm_memory_used_after_last_gc_bytes`
  - `obi.jvm.heap.used` / `obi_jvm_heap_used_bytes`

  Attributes added/used:

  - `jvm.memory.type`
  - `jvm.memory.pool.name`
  - `jvm.gc.phase`

  ## Notes

  `GCTracer::report_gc_heap_summary` and `GCHeapSummary` are HotSpot internals rather than a public JVM ABI. I checked the current OpenJDK LTS lines 8, 11, 17, 21, and 25, and the
  method signature plus the relevant `GCHeapSummary` layout are stable across those versions.

  I also evaluated `github.com/parca-dev/usdt`. This PR keeps the local USDT parser for now because OBI needs target-ELF-driven argument parsing and a BPF map ABI tailored to this
  tracer. A TODO is left in the code to revisit Parca USDT if it exposes reusable pieces that fit this path.

  ## Testing

  - `make docker-generate`
  - `make compile`
  - `make lint`
  - `go test ./pkg/appolly/app/runtime`
  - `go test ./pkg/ebpf/common`
  - `go test ./pkg/runtimemetrics`
  - `go test ./pkg/internal/ebpf/generictracer`
  - `go test -run JVM ./pkg/export/otel ./pkg/export/prom`
  - `go test ./internal/test/integration -run '^TestJVMRuntimeMetrics$' -count=1 -v`
  - `go test ./internal/test/integration -run '^TestJVMRuntimeEventsLive$' -count=1 -v`


grcevski:
Just FYI, this PR started here accidentally https://github.com/grafana/opentelemetry-ebpf-instrumentation/pull/50. I already put some comments in which were already addressed by @REASY. I will review this once more later today.

mmat11:
I'll review this carefully tomorrow, I see the USDT probes parsing/linking implementation in here which should be out of scope -- have you tried using https://github.com/parca-dev/usdt so we have the basic infrastructure already done for us?

REASY:
> I'll review this carefully tomorrow, I see the USDT probes parsing/linking implementation in here which should be out of scope -- have you tried using https://github.com/parca-dev/usdt so we have the basic infrastructure already done for us?

Hi, @mmat11 

I can't comment on the maturity of that library and I do not see the usage of it here. If we accept that risk, I can refactor and use it.

grcevski:
> > I'll review this carefully tomorrow, I see the USDT probes parsing/linking implementation in here which should be out of scope -- have you tried using https://github.com/parca-dev/usdt so we have the basic infrastructure already done for us?
> 
> Hi, @mmat11
> 
> I can't comment on the maturity of that library and I do not see the usage of it here. If we accept that risk, I can refactor and use it.

If it's possible to use parca-dev/usdt that would be much better. Parca is an eBPF profiler from Polar Signals and well maintained project. It would be great if we could use that instead of maintaining our own version, if it's compatible with what you are adding.

REASY:
Hi, @grcevski @mmat11 

I checked parca-dev/usdt. It can likely replace part of our Go-side USDT parsing/linking code, but it is not a full drop-in for this PR. The main concern is that its BPF helper path uses [bpf_get_attach_cookie](https://docs.ebpf.io/linux/helper-function/bpf_get_attach_cookie/), and the project documents Linux 5.15+ as a requirement for that. This repo still needs to support older kernels, including 5.10 in CI, so adopting its BPF-side model directly would change our compatibility assumptions.

A reasonable path may be to use Parca’s ELF/USDT argument parsing on the Go side, then adapt the result into our existing map-based BPF lookup by pid/ns/ip. I’d also want to verify HotSpot probe args don’t need any syntax that our current parser supports but Parca’s parser/helper path does not.

WDYT?

grcevski:
> Hi, @grcevski @mmat11
> 
> I checked parca-dev/usdt. It can likely replace part of our Go-side USDT parsing/linking code, but it is not a full drop-in for this PR. The main concern is that its BPF helper path uses [bpf_get_attach_cookie](https://docs.ebpf.io/linux/helper-function/bpf_get_attach_cookie/), and the project documents Linux 5.15+ as a requirement for that. This repo still needs to support older kernels, including 5.10 in CI, so adopting its BPF-side model directly would change our compatibility assumptions.
> 
> A reasonable path may be to use Parca’s ELF/USDT argument parsing on the Go side, then adapt the result into our existing map-based BPF lookup by pid/ns/ip. I’d also want to verify HotSpot probe args don’t need any syntax that our current parser supports but Parca’s parser/helper path does not.
> 
> WDYT?

Thanks for looking into it. If we can use any of their Go code that would be good, otherwise I'm good with having our own code and then adding a TODO to look into leveraging Parca USDT in the future.

pkg/appolly/instrumenter.go
```diff
@@ -56,7 +58,12 @@ func newGraphBuilder(
 	tracesCh *msg.Queue[[]request.Span],
 	processEventsCh *msg.Queue[exec.ProcessEvent],
 	runtimeMetrics *msg.Queue[[]runtimemetrics.RuntimeMetricSnapshot],
+	jvmRuntimeEventsOpt ...*msg.Queue[[]jvmruntime.JVMRuntimeEvent],
```
grcevski: I'm wondering if we can simplify this now, because at the moment we have these separate arguments that create separate pipelines and if we were to add runtime metrics for NodeJS, Python, Ruby..., it would get unwieldy. 

Can we keep the base type as `runtimemetrics.RuntimeMetricSnapshot` but then inside the struct, refactor it to have `go` section and then `java` section. With this we can hide all the complexity and reduce memory usage, since each prom and otel metrics exporters have memory demands on expirers etc.

Just like we have setupGoRuntimeMeters we can have setup JVM meters etc.

REASY:
@grcevski 

I evaluated [github.com/parca-dev/usdt](https://github.com/parca-dev/usdt), the blockers I found are:

1. The parser is host-architecture driven. At HEAD (aeb59ba), ParseUSDTArguments resolves registers through runtime.GOARCH. In OBI we need to parse USDT args based on the target ELF machine, because the instrumented binary/container may not match the collector build/runtime architecture.

2. The current Parca parser does not model x86 SIB operands, e.g. offset(base,index,scale). This branch currently handles those because some USDT argument expressions can reference both a base register and an index register with scale. Parca’s public arg model has CONST / REG / REG_DEREF, but no equivalent representation for that form.

3. The ELF note parser is narrower than the parser added here. At the checked commit it reads STAPSDT note fields as little-endian ELF64 values. The implementation here is target-ELF driven and handles the ELF class/data encoding explicitly.

Am I missing something?

grcevski:
> @grcevski
> 
> I evaluated [github.com/parca-dev/usdt](https://github.com/parca-dev/usdt), the blockers I found are:
> 
> 1. The parser is host-architecture driven. At HEAD (aeb59ba), ParseUSDTArguments resolves registers through runtime.GOARCH. In OBI we need to parse USDT args based on the target ELF machine, because the instrumented binary/container may not match the collector build/runtime architecture.
> 2. The current Parca parser does not model x86 SIB operands, e.g. offset(base,index,scale). This branch currently handles those because some USDT argument expressions can reference both a base register and an index register with scale. Parca’s public arg model has CONST / REG / REG_DEREF, but no equivalent representation for that form.
> 3. The ELF note parser is narrower than the parser added here. At the checked commit it reads STAPSDT note fields as little-endian ELF64 values. The implementation here is target-ELF driven and handles the ELF class/data encoding explicitly.
> 
> Am I missing something?

That's fine then! Let's keep your code, we have this documented now and if we decide to refactor in the future we can do it. I only had one other comment, moving the handling of the JVM metrics signals under the runtime metrics umbrella, so we don't proliferate even more code in the future. 

You can use the same top object, just put Go as separate pointer and JVM as separate pointer. Then the metrics code can do checks if go != nil, emit Go runtime metrics, if jvm != nil emit JVM runtime metrics etc.

I also really like your design decision to have the last timestamp the metrics were captured, so if you get flurry of GC activity you don't spam OBI now with all of that.

pkg/appolly/app/runtime/jvm.go
```diff
@@ -0,0 +1,220 @@
...
+const JVMRawStringLen = 64
+
+type JVMRuntimeMetricKind string
+
+const (
+	JVMMetricMemoryUsed            JVMRuntimeMetricKind = "jvm.memory.used"
+	JVMMetricMemoryCommitted       JVMRuntimeMetricKind = "jvm.memory.committed"
+	JVMMetricMemoryLimit           JVMRuntimeMetricKind = "jvm.memory.limit"
+	JVMMetricMemoryUsedAfterLastGC JVMRuntimeMetricKind = "jvm.memory.used_after_last_gc"
+	JVMMetricBeylaHeapUsed         JVMRuntimeMetricKind = "beyla.jvm.heap.used"
```
mmat11: this should be `obi`?

bpf/common/maps/obi_usdt_ip_to_spec_id.h
```diff
@@ -0,0 +1,17 @@
...
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_helpers.h>
+#include <common/pin_internal.h>
+#include <common/usdt_types.h>
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __uint(max_entries, k_obi_usdt_max_ip_cnt);
+    __type(key, struct obi_usdt_ip_key);
+    __type(value, u32);
```
mmat11, line 16: can we add a comment with the value type here?

bpf/common/maps/obi_usdt_specs.h
```diff
@@ -0,0 +1,17 @@
...
+
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_helpers.h>
+#include <common/pin_internal.h>
+#include <common/usdt_types.h>
+
+struct {
+    __uint(type, BPF_MAP_TYPE_ARRAY);
+    __uint(max_entries, k_obi_usdt_max_spec_cnt);
+    __type(key, u32);
```
mmat11, line 14: can we add a comment with the value type here?

pkg/obi/config.go
```diff
@@ -624,6 +629,11 @@ type JavaConfig struct {
 	Timeout              time.Duration `yaml:"attach_timeout" env:"OTEL_EBPF_JAVAAGENT_ATTACH_TIMEOUT" validate:"gte=0"`
 }
 
+type JVMRuntimeMetricsConfig struct {
+	Enabled          bool          `yaml:"enabled" env:"BEYLA_JVM_RUNTIME_METRICS_ENABLED"`
```
mmat11: these should also be `OBI_`

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -592,23 +642,59 @@ func (p *Tracer) Run(ctx context.Context, ebpfEventContext *ebpfcommon.EBPFEvent
...
 		ebpfEventContext,
 		cfg,
 		p.bpfObjects.Events,
 		func(record *ringbuf.Record) (request.Span, bool, error) {
-			s, ignore, err := ebpfcommon.ReadBPFTraceAsSpan(parseContext, cfg, record, p.pidsFilter)
-			if !ignore && err == nil && !s.IsValid() {
-				return s, true, nil
-			}
-			return s, ignore, err
+			return p.processSharedRingbufRecord(ctx, parseContext, cfg, record)
```
mmat11, line 656: can't JVM metrics be handled inside `ReadBPFTraceAsSpan`?

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -592,23 +642,59 @@ func (p *Tracer) Run(ctx context.Context, ebpfEventContext *ebpfcommon.EBPFEvent
 	p.log.Info("Launching p.Tracer")
 
 	cfg := &p.cfg.EBPF
+	if p.cfg.JVMRuntimeMetrics.Enabled {
+		if p.jvmRuntimeEventQueue() == nil {
+			p.log.Debug("JVM runtime metrics enabled without event queue")
```
mmat11: should this be a Warn?

pkg/export/attributes/metric.go
```diff
@@ -164,6 +164,31 @@ var (
...
+		Section: "jvm.memory.limit",
+		Prom:    "jvm_memory_limit_bytes",
+		OTEL:    "jvm.memory.limit",
+	}
+	JVMMemoryUsedAfterLastGC = Name{
+		Section: "jvm.memory.used_after_last_gc",
+		Prom:    "jvm_memory_used_after_last_gc_bytes",
+		OTEL:    "jvm.memory.used_after_last_gc",
+	}
+	BeylaJVMHeapUsed = Name{
```
mmat11: should also be `Obi*`

pkg/export/attributes/metric.go
```diff
@@ -164,6 +164,31 @@ var (
...
+		Section: "beyla.jvm.heap.used",
+		Prom:    "beyla_jvm_heap_used_bytes",
+		OTEL:    "beyla.jvm.heap.used",
```
mmat11: same for these

pkg/appolly/instrumenter.go
REASY: @grcevski I've made the changes, let me know whether this is how you envisioned it.

pkg/ebpf/usdt.go
```diff
@@ -0,0 +1,554 @@
...
+	obiUSDTNoteName    = "stapsdt"
+	obiUSDTArgConst    = uint8(0)
+	obiUSDTArgReg      = uint8(1)
+	obiUSDTArgRegDeref = uint8(2)
+	obiUSDTArgSIB      = uint8(3)
+)
+
+// TODO: Reevaluate github.com/parca-dev/usdt if it exposes target-ELF-driven
+// argument parsing for OBI's BPF ABI. v0.0.2 uses runtime.GOARCH for register
+// parsing, lacks x86 SIB args, uses a different spec layout, and its note
```
mmat11: SIB looks to be dead code/unused in this context, right? `docker run --rm --platform linux/amd64 eclipse-temurin:21-jdk bash -c "readelf -n /opt/java/openjdk/lib/server/libjvm.so"`

pkg/ebpf/usdt.go
```diff
@@ -0,0 +1,554 @@
...
+	SemaOff uint64
+	Spec    obiUSDTSpec
+	SpecKey string
+}
+
+var (
+	errUnsupportedUSDTArch = errors.New("unsupported USDT architecture")
+
+	usdtNumberRE     = `[+-]?(?:0x[0-9A-Fa-f]+|\d+)`
+	x86SIBArgRE      = regexp.MustCompile(`^\s*([+-]?\d+)\s*@\s*(` + usdtNumberRE + `)?\s*\(\s*%([A-Za-z0-9]+)\s*,\s*%([A-Za-z0-9]+)\s*(?:,\s*(\d+)\s*)?\)\s*`)
```
mmat11: for simplicity, I would remove any SIB parsing code

pkg/ebpf/instrumenter.go
```diff
@@ -657,39 +872,76 @@ func gatherOffsets(instrPath string, probes map[string][]*ebpfcommon.ProbeDesc,
...
 			if progData == nil {
-				return fmt.Errorf("error reading symbol data for %s (%s)", symbolName, instrPath)
-			}
-
-			returns, err := goexec.FindReturnOffsets(sym.Off, progData)
-			if err != nil {
-				log.Debug("Error finding return offsets", "symbol", sym)
+				log.Debug("error reading symbol data", "symbol", symbolName, "path", instrPath)
 				continue
 			}
```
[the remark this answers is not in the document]
REASY: Fixed

bpf/common/maps/obi_usdt_ip_to_spec_id.h
```diff
@@ -0,0 +1,17 @@
...
+    __uint(pinning, OBI_PIN_INTERNAL);
+} obi_usdt_ip_to_spec_id SEC(".maps");
```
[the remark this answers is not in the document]
REASY: Fixed. Now it track the USDT IP-map keys inserted for each attached target and delete them when the owning uprobe link is closed. This prevents stale IP-to-spec mappings from accumulating across process churn while keeping the existing hash map semantics.

pkg/ebpf/usdt.go
```diff
@@ -0,0 +1,554 @@
...
+	}
+
+	var baseAddr uint64
+	if base := elfFile.Section(".stapsdt.base"); base != nil {
+		baseAddr = base.Addr
+	}
+
+	targets := []usdtTarget{}
+	for offset := 0; offset+12 <= len(data); {
+		namesz := int(elfFile.ByteOrder.Uint32(data[offset:]))
```
mmat11: nit: for clarity, maybe we can define an sdtHeader and `binary.Read()` it, see https://github.com/mmat11/usdt/blob/main/cache.go#L156

bpf/common/maps/obi_usdt_ip_to_spec_id.h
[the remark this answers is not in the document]
grcevski: For safety you can make it LRU hash, we use that extensively for this very purpose.

pkg/appolly/app/runtime/jvm.go
```diff
@@ -0,0 +1,220 @@
...
+
+type RawJVMGCWhenType uint32
+
+const (
+	RawJVMGCWhenBefore RawJVMGCWhenType = iota
+	RawJVMGCWhenAfter
+	RawJVMGCWhenEndSentinel
+)
+
+type RawJVMMemoryPoolEvent struct {
```
mmat11: aren't these events generated by bpf2go?

bpf/common/usdt_types.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include <bpfcore/vmlinux.h>
+
+enum { k_obi_usdt_max_args = 12, k_obi_usdt_max_spec_cnt = 256, k_obi_usdt_max_ip_cnt = 1024 };
+
+enum obi_usdt_arg_type {
+    k_obi_usdt_arg_const = 0,
+    k_obi_usdt_arg_reg = 1,
+    k_obi_usdt_arg_reg_deref = 2,
+    k_obi_usdt_arg_sib = 3,
```
mmat11: this should be unused

bpf/common/usdt.h
```diff
@@ -0,0 +1,188 @@
...
+        err = bpf_probe_read_kernel(&val, sizeof(val), (unsigned char *)ctx + arg->reg_off);
+        if (err) {
+            return err;
+        }
+        err = obi_usdt_read_user_value(val + arg->val_off, arg->arg_bitshift, &val);
+        if (err) {
+            return err;
+        }
+        break;
+    case k_obi_usdt_arg_sib:
```
mmat11: I'd remove this branch for simplicity if it's unused

bpf/common/maps/obi_usdt_ip_to_spec_id.h
```diff
@@ -0,0 +1,17 @@
...
+
+#pragma once
+
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_helpers.h>
+#include <common/pin_internal.h>
+#include <common/usdt_types.h>
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
```
mmat11: we should probably use an HASH_LRU here

mmat11 (commented):
did a first pass, looks good!

pkg/internal/ebpf/generictracer/generictracer.go
REASY: The current reason it is separate is that JVM metric records are not spans and need EBPFEventContext.RuntimeMetrics.

We can add a shared common ring-buffer record handler that first handles JVM metric records, then delegates to ReadBPFTraceAsSpan. That avoids embedding metric side effects into a span decoder while addressing the duplication concern.

bpf/common/maps/obi_usdt_ip_to_spec_id.h
REASY: I already added explicit key cleanup on USDT link close in 9a2494ce. LRU can evict active probe mappings and silently drop metrics. Plain hash plus explicit cleanup gives deterministic behavior.

pkg/appolly/app/runtime/jvm.go
REASY: Addressed. bpf2go now generates both JVM event structs from the BPF declarations, and I added a layout test comparing those generated structs against the runtime package raw structs.

I kept the runtime RawJVM* structs because pkg/appolly/app/runtime owns JVM runtime metric decoding and should not import the internal generictracer generated package. The test makes that mirror explicit and catches ABI drift

bpf/common/maps/obi_usdt_ip_to_spec_id.h
grcevski: Actually it doesn't hurt to change this to LRU hash, it will still be deterministic, one thing that won't happen is  leak if for some reason the delete fails.

grcevski:
Hi @REASY, can you please sync with main, we had some tests break which are now fixed, but they are blocking your PR. If you sync with main I expect we'll get green CI.

pkg/appolly/app/runtime/jvm.go
grcevski: Did you hit an import cycle with importing the generictracer package? I think it makes sense to use that struct instead of duplicating it here.

pkg/appolly/app/runtime/jvm.go
```diff
@@ -0,0 +1,220 @@
...
+			return JVMMemoryTypeNonHeap
+		}
+	}
+	for _, heapName := range []string{"eden", "survivor", "old", "tenured", "young", "zheap", "shenandoah", "epsilon", "humongous"} {
+		if strings.Contains(name, heapName) {
+			return JVMMemoryTypeHeap
+		}
+	}
+	return ""
+}
```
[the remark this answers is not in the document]
grcevski: I like the idea of "unknown" that copilot suggested, I think it's a good suggestion.

pkg/appolly/discover/attacher.go
```diff
@@ -97,6 +97,9 @@ func (ta *traceAttacher) attacherLoop(_ context.Context) (swarm.RunFunc, error)
 	}
 	ta.processInstances = maps.MultiCounter[uint64]{}
 	ta.EbpfEventContext.CommonPIDsFilter = ebpfcommon.NewPIDsFilter(&ta.Cfg.Discovery, slog.With("component", "ebpfCommon.CommonPIDsFilter"), ta.Metrics)
+	if ta.RuntimeMetrics != nil {
```
grcevski, line 100: Just a question, I think this seems reasonable, but I'm wondering why we didn't need this for the Go metrics?

pkg/appolly/instrumenter.go
```diff
@@ -209,7 +212,7 @@ func setupMetricsSubPipeline(
 		), swarm.WithID("PrometheusEndpoint"))
 	}
 
-	if jointMetricsConfig.Features.AppRuntime() {
+	if jointMetricsConfig.Features.AppRuntime() || jointMetricsConfig.Features.AppJVM() {
```
grcevski: Is it possible to combine all runtime metrics under Features.AppRuntime(). We can add as a follow-up option to disable some if needed. That way the check will be much simpler.

pkg/export/otel/metrics_runtime.go
```diff
@@ -109,8 +118,8 @@ func newRuntimeMetricsReporter(
 			llog.Debug("evicting runtime metrics reporter from cache")
 
 			go func() {
-				if err := v.provider.ForceFlush(ctx); err != nil {
-					llog.Warn("error flushing evicted runtime metrics provider", "error", err)
+				if err := v.provider.Shutdown(ctx); err != nil {
```
grcevski, line 124: Nice!

pkg/export/prom/prom.go
```diff
@@ -876,6 +880,9 @@ func newReporter(
 	if jointMetricsConfig.Features.AppRuntime() {
 		registeredMetrics = append(registeredMetrics, mr.goRuntimeMetrics.collectors()...)
 	}
+	if jointMetricsConfig.Features.AppJVM() {
```
grcevski: You see how it's here, I think we need to refactor this slightly. The way the code reads now is:

```
if runtimeMetrics.enabled do something
if jvm.enabled do something
```

it should be like:
```
if runtimeGoMetrics.enabled do something
if runtimeJVMMetrics.enabled do something
```

pkg/internal/ebpf/generictracer/generictracer.go
grcevski: Right, cool, so this is the same pattern as the Go tracer uses for Go runtime metrics.

pkg/internal/ebpf/generictracer/jvm_live_test.go
```diff
@@ -0,0 +1,327 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build linux && jvm_live
+
+package generictracer
+
+import (
```
grcevski, line 8: What is this test for, maybe it's better to just keep an integration test?

pkg/internal/ebpf/gotracer/gotracer.go
```diff
@@ -873,6 +873,9 @@ func (p *Tracer) Run(ctx context.Context, ebpfEventContext *ebpfcommon.EBPFEvent
 		p.cfg,
 		p.bpfObjects.Events,
 		func(record *ringbuf.Record) (request.Span, bool, error) {
+			if handled, err := ebpfcommon.HandleJVMRuntimeMetricRecord(ctx, ebpfEventContext, record, p.pidsFilter, p.log); handled {
```
grcevski: We need to collapse these under one common HandleRuntimeMetricsRecord which will be used by the Go tracer and the generic tracer. That way we don't expand the code here and in the generic tracer part as we add more tracers.

One thing to note, I think the current Go runtime metrics have a bug, they should also be handled by the generic tracer just like you added in this PR. So it's good to have a common function that does if jvm do X, if go do Y.

grcevski (commented):
This is looking pretty good! Thanks for all the hard work @REASY!

I think we just need to refactor and clean-up the way the runtime metrics are checked and used throughout the code. I'd like to have one global helper and common functions, which then internally check for Go and JVM, or don't check for now and we can add that later. But, it's important that all runtime metrics handling to be encapsulated as runtime metrics for the rest of the code.

pkg/appolly/discover/attacher.go
grcevski: Oh I think I can answer my own question, for Go metrics if we just have a Go application they are created in the Go path.

bpf/common/h2_defs.h
```diff
@@ -28,9 +28,9 @@ enum {
     // --- W3C traceparent value layout: "00-<trace_id>-<span_id>-01" ---
     k_tp_val_dash1 = 2,
     k_tp_val_trace_id_start = k_tp_val_dash1 + 1,
-    k_tp_val_dash2 = k_tp_val_trace_id_start + TRACE_ID_SIZE_BYTES * 2,
+    k_tp_val_dash2 = k_tp_val_trace_id_start + (TRACE_ID_SIZE_BYTES * 2),
     k_tp_val_span_id_start = k_tp_val_dash2 + 1,
-    k_tp_val_dash3 = k_tp_val_span_id_start + SPAN_ID_SIZE_BYTES * 2,
+    k_tp_val_dash3 = k_tp_val_span_id_start + (SPAN_ID_SIZE_BYTES * 2),
```
rafaelroquetto: these are unrelated and unnecessary, could you please revert those?

bpf/common/usdt.h
```diff
@@ -0,0 +1,166 @@
...
+    default:
+        return 0;
+    }
+}
+
+static __always_inline u8 obi_usdt_reg_off_ok(s16 reg_off) {
+    if (reg_off < 0) {
+        return 0;
+    }
+
```
rafaelroquetto, line 39: nit: 
```
_Static_assert(sizeof(struct pt_regs) >= sizeof(unsigned long),
               "pt_regs smaller than register width breaks reg_off bound");
```

bpf/common/usdt.h
```diff
@@ -0,0 +1,166 @@
...
+    return (u32)reg_off <= sizeof(struct pt_regs) - sizeof(unsigned long);
```
rafaelroquetto: nit
```suggestion
    return (size_t)reg_off <= sizeof(struct pt_regs) - sizeof(unsigned long);
```

rationale: (u32) reg_off is being implicitly promoted to `(size_t)` after all - might as well make it explicit

bpf/common/usdt.h
```diff
@@ -0,0 +1,166 @@
...
+static __always_inline struct obi_usdt_spec *obi_usdt_spec_for_ctx(struct pt_regs *ctx) {
+    const u64 pid_tgid = bpf_get_current_pid_tgid();
+    const struct task_struct *task = (struct task_struct *)bpf_get_current_task();
+
+    int ns_pid = 0;
+    int ns_ppid = 0;
+    u32 pid_ns_id = 0;
+    ns_pid_ppid(task, &ns_pid, &ns_ppid, &pid_ns_id);
+
+    u32 pid = pid_from_pid_tgid(pid_tgid);
```
rafaelroquetto: I believe you can call `valid_pid` directly here. `valid_pid` already checks for `filter_pids` internally, and will correctly return `(pid >> 32)`.

Also, are you sure you need the entire ns_pid machinery? Why can't `obi_usdt_ip_key` operate on host pids?

bpf/generictracer/jvm.c
```diff
@@ -0,0 +1,313 @@
...
+jvm_read_usdt_string(unsigned char *dst, const unsigned char *src, long src_len) {
+    __builtin_memset(dst, 0, k_jvm_raw_string_len);
+    if (!src || src_len <= 0) {
+        return -1;
+    }
+
+    u32 max_len = k_jvm_raw_string_len - 1;
+    if (src_len < (long)max_len) {
+        max_len = (u32)src_len;
+    }
```
rafaelroquetto:
```suggestion
    const u32 max_len = (u32)min((long)(k_jvm_raw_string_len - 1), src_len);
```

bpf/generictracer/jvm.c
```diff
@@ -0,0 +1,313 @@
...
+
+    for (u32 i = 0; i < k_jvm_raw_string_len - 1; i++) {
+        if (i >= max_len) {
+            continue;
```
rafaelroquetto: I reckon we should `break` here instead of just looping to the end

bpf/generictracer/jvm.c
```diff
@@ -0,0 +1,313 @@
...
+        }
+
+        unsigned char c = 0;
+        if (bpf_probe_read_user(&c, sizeof(c), src + i) != 0) {
+            return -1;
+        }
+        dst[i] = c;
```
rafaelroquetto: So, this is _very_ inneficient. `bpf_probe_read_user` does a lot of checks internally to verify the address range,  it must disable and the re-enable page faults, etc.

We have to do a bulk read here instead. For instance:

```
            bpf_probe_read_user(attr->value,
                                go_attr_value->string.len & (OTEL_ATTRIBUTE_VALUE_MAX_LEN - 1),
                                go_attr_value->string.str);
```

is how we do it in `otel_types.h`

bpf/generictracer/jvm.c
```diff
@@ -0,0 +1,313 @@
...
+                                                   long manager_len,
+                                                   const unsigned char *pool,
+                                                   long pool_len,
+                                                   u64 init_size,
+                                                   u64 used,
+                                                   u64 committed,
+                                                   u64 max_size) {
+    (void)ctx;
+
+    if (!jvm_runtime_metrics_enabled) {
```
rafaelroquetto: I'd move this to the actual toplevel uprobe preambles - this allows you to bail very early before the uprobe doing any work such as `jvm_read_hotspot_usdt_arg`.

bpf/generictracer/jvm.c
```diff
@@ -0,0 +1,313 @@
...
+    }
+    if (when != k_jvm_before_gc && when != k_jvm_after_gc) {
+        return 0;
+    }
+
+    const u64 pid_tgid = bpf_get_current_pid_tgid();
+    const u32 pid = valid_pid(pid_tgid);
+    if (!pid) {
+        return 0;
+    }
```
rafaelroquetto: ditto

bpf/generictracer/protocol_http2.h
```diff
@@ -428,11 +428,11 @@ int obi_protocol_http2_grpc_handle_start_frame_server(void *ctx) {
 
     const u8 flags = h2g_info->data[4];
     const u8 padded = (flags >> 3) & 1;
-    const u32 prefix = padded + ((flags >> 5) & 1) * k_h2_priority_prefix_len;
+    const u32 prefix = padded + (((flags >> 5) & 1) * k_h2_priority_prefix_len);
```
rafaelroquetto: these are unrelated, please revert

bpf/logenricher/types.h
```diff
@@ -54,12 +54,12 @@ enum tty_driver_type___new {
 
 enum tty_driver_subtype___new {
     SYSTEM_TYPE_TTY = 1,
-    SYSTEM_TYPE_CONSOLE,
-    SYSTEM_TYPE_SYSCONS,
-    SYSTEM_TYPE_SYSPTMX,
+    SYSTEM_TYPE_CONSOLE = 2,
+    SYSTEM_TYPE_SYSCONS = 3,
+    SYSTEM_TYPE_SYSPTMX = 4,
```
rafaelroquetto: unrelated

pkg/appolly/app/runtime/jvm.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+	if phase == JVMGCPhaseAfter {
+		events = append(events, withJVMMetric(base, JVMMetricMemoryUsedAfterLastGC, raw.Used))
+	}
+	return events, nil
+}
+
+func DecodeJVMMemoryPoolEvent(payload []byte) ([]JVMRuntimeEvent, error) {
+	var raw RawJVMMemoryPoolEvent
+	if err := readRawPayload(payload, &raw); err != nil {
```
rafaelroquetto: `readRawPayload` should be replaced with `ReinterpretCast` -> this sits on a hotpath, and `binary.*` methos are _very_slow (they use reflection, which is slow in itself) - we used to do this in the past, but went down the `ReinterpretCast` route for performance reasons.

```
raw, err := ebpfcommon.ReinterpretCast[RawJVMMemoryPoolEvent](payload)
if err != nil {
    return nil, err
}
return ParseJVMMemoryPoolEvent(*raw)
````

Also, as the other reviewers suggested, you should not have to declare a mirroring struct to the ebpf counterparts. Let `bpf2go` gerenerate that for you.

File: pkg/appolly/app/runtime/jvm.go [hunk 38]
```diff
@@ -0,0 +1,220 @@
...
+	case RawJVMGCWhenBefore:
+		return JVMGCPhaseBefore, nil
+	case RawJVMGCWhenAfter:
+		return JVMGCPhaseAfter, nil
+	default:
+		return "", fmt.Errorf("unsupported JVM GC phase: %d", raw)
+	}
+}
+
+func readRawPayload[T any](payload []byte, dst *T) error {
```
rafaelroquetto: Please remove and replace call sites with `ReinterpretCast`

pkg/ebpf/common/jvm.go
```diff
@@ -0,0 +1,131 @@
...
+	eventContext *EBPFEventContext,
+	record *ringbuf.Record,
+	filter ServiceFilter,
+	logger *slog.Logger,
+) (bool, error) {
+	if record == nil || len(record.RawSample) == 0 {
+		return false, nil
+	}
+
+	switch record.RawSample[0] {
```
rafaelroquetto: nit: the established pattern across the codebase for clarity is:

```
eventType := record.RawSample[0]

switch eventType { ... }
```

pkg/ebpf/common/jvm.go
```diff
@@ -0,0 +1,131 @@
...
+	record *ringbuf.Record,
+	filter ServiceFilter,
+	logger *slog.Logger,
+) ([]jvmruntime.JVMRuntimeEvent, bool, error) {
+	events, err := jvmruntime.DecodeJVMMemoryPoolEvent(record.RawSample)
+	if err != nil {
+		return nil, false, err
+	}
+
+	decorated := events[:0]
```
rafaelroquetto: it seems that all events here are fanned out from one raw sample and share one PID/namespace, so decoration is all or nothing, I believe you could phrase this as:

```
events, err := jvmruntime.DecodeJVMMemoryPoolEvent(record.RawSample)
if err != nil {
	return nil, false, err
}

if len(events) == 0 {
	return nil, true, nil
}

if !decorateJVMRuntimeEvent(filter, &events[0]) {
	return nil, true, nil
}

for i := 1; i < len(events); i++ {
	events[i].Service = events[0].Service
}
```

which does exactly one filter lookup instead of N

pkg/ebpf/instrumenter.go
```diff
@@ -657,39 +922,99 @@ func gatherOffsets(instrPath string, probes map[string][]*ebpfcommon.ProbeDesc,
 func gatherOffsetsImpl(elfFile *elf.File, probes map[string][]*ebpfcommon.ProbeDesc,
 	instrPath string, log *slog.Logger,
 ) error {
-	syms, err := procs.FindExeSymbols(elfFile, symbolNames(probes))
+	exactSyms, err := procs.FindExeSymbols(elfFile, symbolNames(probes, ebpfcommon.SymbolMatcherExact))
 	if err != nil {
 		return fmt.Errorf("failed to lookup symbols for %s: %w", instrPath, err)
 	}
 
+	substringSyms, err := procs.FindExeSymbolsBySubstring(elfFile, symbolNames(probes, ebpfcommon.SymbolMatcherContains))
```
rafaelroquetto: Now we are iterating the symbol table twice - each call to `fidExeSymbols` needs to:

1. parse all of of `.symtab`
2. collectSymbols()
3. parse all of `.dynsym`
4. collectSymbols() again

You should instead strive to parse once and match both. The current code (prior to your changes) already makes undesirable appearance on the profiler, and this will unfortunately add a lot to it.

pkg/ebpf/instrumenter.go
```diff
@@ -633,11 +888,21 @@ func getCgroupPath() (string, error) {
 	return cgroupPath, err
 }
 
-func symbolNames(m map[string][]*ebpfcommon.ProbeDesc) []string {
+func symbolNames(m map[string][]*ebpfcommon.ProbeDesc, matcher ebpfcommon.SymbolMatcher) []string {
 	keys := make([]string, 0, len(m))
+	seen := map[string]struct{}{}
```
rafaelroquetto: I'm not sure `seen` is necessary - `m` is already keyed by `name`, so the outer `for name, probes := range m` already visits  each `name` exactly once.

If you are worried about several probles within that one `name` slice matching `matcher`, you can just use a break:

```
func symbolNames(m map[string][]*ebpfcommon.ProbeDesc, matcher ebpfcommon.SymbolMatcher) []string {
	keys := make([]string, 0, len(m))
	for name, probes := range m {
		for _, probe := range probes {
			if probe.SymbolMatcher == matcher {
				keys = append(keys, name)
				break
			}
		}
	}
	return keys
}
```

pkg/ebpf/instrumenter.go
```diff
@@ -367,6 +423,205 @@ func (i *instrumenter) uprobes(pid app.PID, p Tracer) error {
 	return nil
 }
 
+func (i *instrumenter) usdtProbes(pid app.PID, ns uint32, p Tracer) error {
+	usdtTracer, ok := p.(USDTTracer)
```
rafaelroquetto: Is there a reason for this? This is clearly the go way but we chose a different path: the `Tracer` interface.

So for consistency, you should make USDT a direct method there - move `USDTProbes()` into `Tracer`, and drop this type assertion from here.

pkg/ebpf/instrumenter.go
```diff
@@ -367,6 +423,205 @@ func (i *instrumenter) uprobes(pid app.PID, p Tracer) error {
...
+				ilog().Debug("error instrumenting optional USDT probe",
+					"pid", pid,
+					"lib", baseLib,
+					"provider", probe.Provider,
+					"name", probe.Name,
+					"error", err,
+				)
+				continue
+			}
+			i.closables = append(i.closables, closers...)
```
rafaelroquetto: these USDT closers are tracked twice: appended to `i.closables` here and to the local `usdtClosers`, but only usdtClosers is passed to AddCloser. Since the instrumenter is persisted and reused across executable instances, a later uprobes call doing AddCloser(i.closables...) will re-register these same closers, so each usdtLinkCloser ends up registered and Close()d multiple times, which is harmless only because its Close() is using sync.Once. Please drop this append and rely solely on AddCloser(usdtClosers...) at line 505 for a single, consistent registration path.

pkg/ebpf/instrumenter.go
```diff
@@ -367,6 +423,205 @@ func (i *instrumenter) uprobes(pid app.PID, p Tracer) error {
...
+	}
+
+	closers := make([]io.Closer, 0, len(targets))
+	for _, target := range targets {
+		specID, err := probe.SpecManager.ID(target.SpecKey, obiUSDTMaxSpecCnt)
+		if err != nil {
+			closeAll(closers)
+			return nil, err
+		}
+		if err := probe.SpecsMap.Put(specID, target.Spec); err != nil {
```
rafaelroquetto, line 547: this spec is never removed - usdtLinkCloser cleans up the IP-map keys and the link on close, but the SpecsMap entry and SpecManager id stick around for the process lifetime (and leak on a later attach failure). I assume that's intentional since the spec is shared by layout across PIDs, but combined with SpecManager never reclaiming ids it's effectively append-only, capped only by obiUSDTMaxSpecCnt. Can we add a one-liner noting specs are never reclaimed so it doesn't read as a leak?

pkg/ebpf/instrumenter.go
```diff
@@ -367,6 +423,205 @@ func (i *instrumenter) uprobes(pid app.PID, p Tracer) error {
...
+
+		elfFile.Close()
+	}
+
+	p.AddCloser(usdtClosers...)
+	return nil
+}
+
+func usdtInstrumentationPath(pid app.PID, lib string, maps []*procfs.ProcMap, exePath string) (string, string) {
+	libMap := procs.LibPath(lib, maps)
```
rafaelroquetto: This duplicates the lib -> instr-path resolution uprobeModules already does inline (LibPath -> map_files -> stat -> fallback to Pathname, lines 223-244). We already reuse resolveExePath/processMaps/matchVersionedUprobeLibrary, so let's pull this out too, e.g. resolveInstrPath(pid, lib, maps, exePath) (instrPath, ino, mappedPath) shared by both. Attach loop is fine to keep separate given the spec/IP-map + semaphore handling.

pkg/ebpf/tracer.go
```diff
@@ -119,6 +119,10 @@ type Tracer interface {
 	Run(context.Context, *ebpfcommon.EBPFEventContext, *msg.Queue[[]request.Span])
 }
 
+type USDTTracer interface {
```
rafaelroquetto: As per my other comment, you should fold this into `Tracer` for consistency.

pkg/ebpf/usdt.go
```diff
@@ -0,0 +1,550 @@
...
+func elfFileOffset(elfFile *elf.File, addr uint64, requireExecutable bool) (uint64, error) {
+	for _, prog := range elfFile.Progs {
+		if prog.Type != elf.PT_LOAD {
+			continue
+		}
+		if requireExecutable && prog.Flags&elf.PF_X == 0 {
+			continue
+		}
+		if addr >= prog.Vaddr && addr < prog.Vaddr+prog.Memsz {
+			if requireExecutable && prog.Flags&elf.PF_X == 0 {
```
rafaelroquetto: this is dead code, see line 295

bpf/common/usdt.h
REASY: For USDT lookup I think host PID is enough because this map is populated by OBI userspace while attaching a uprobe to a host /proc PID, and BPF can read the same host PID from bpf_get_current_pid_tgid(), removed namespace

bpf/common/h2_defs.h
REASY: Reverted, however the reason this was done clang-tidy marked them as readability-parentheses findings

bpf/generictracer/protocol_http2.h
REASY: Reverted, however the reason this was done clang-tidy marked them as readability-parentheses findings

bpf/generictracer/jvm.c
REASY: Fixed using the same bulk-read approach, adapted for JVM strings by bounding the read to k_jvm_raw_string_len - 1 and pre-zeroing the destination so truncated values remain NUL-terminated.

pkg/appolly/app/runtime/jvm.go
REASY: The manual RawJVMMemoryPoolEvent / RawJVMGCHeapSummaryEvent structs are gone from pkg/appolly/app/runtime/jvm.go. Decoding now uses bpf2go-generated structs in generictracer, because importing generictracer into pkg/appolly/app/runtime would create an import cycle.

pkg/appolly/app/runtime/jvm.go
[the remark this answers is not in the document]
REASY: InferJVMMemoryType now returns JVMMemoryTypeUnknown instead of ""

pkg/appolly/instrumenter.go
REASY: I can do that, but want to confirm the intended config semantics first.

Current behavior treats application_runtime and application_jvm as separate runtime-metric feature gates: Go runtime metrics are enabled by AppRuntime(), JVM runtime metrics by AppJVM(). The upside is users can enable JVM runtime metrics independently. The downside is the code has to carry a small Go/JVM runtime-metrics split.

If we move all runtime metrics under Features.AppRuntime(), the code gets cleaner and application_runtime becomes the single umbrella for runtime metrics, which is probably better for future Node/Python/Ruby runtime metrics. The tradeoff is that existing configs with only application_jvm would no longer enable JVM runtime metrics unless users also set application_runtime, and we may still need internal per-runtime flags later to avoid initializing/exporting metrics users do not want.

My preference would be to keep the internal runtimemetrics.Enabled{Go,JVM} helper for clarity, but I’m fine changing the public gate so AppRuntime() is the umbrella if we accept the config behavior change.

pkg/internal/ebpf/generictracer/jvm_live_test.go
REASY: This is not intended to run as a regular unit/package test. It is behind the linux && jvm_live build tag and is invoked by the integration test suite through docker-compose-jvm-runtime-events-live.yml. I put the implementation near generictracer because it exercises the raw HotSpot USDT to shared-ringbuf to runtime-event path without Prometheus/OTel exporter layers. The higher-level TestJVMRuntimeMetrics covers exported Prometheus metrics. If you prefer, I can move the live test implementation fully under internal/test/integration, but functionally it is already only run as an integration test.`

pkg/internal/ebpf/gotracer/gotracer.go
REASY: Done in bdc2cef4. I collapsed the runtime metric dispatch through ebpfcommon.HandleRuntimeMetricsRecord and now generictracer uses the same common path as gotracer.

The helper handles Go runtime metric records directly, so generictracer will now consume Go runtime metrics too if they arrive there. For JVM records, generictracer passes a tracer-local handler into the common helper because the JVM payload structs are generated by bpf2go in the generictracer package, and moving that decoding into common would introduce an import cycle.

So the flow is now: common helper decides "is this a runtime metric record?", handles Go directly, and delegates JVM decoding where the generated types live.

rafaelroquetto:
Hey @REASY, thank you for the changes. Some tests are failing. I will defer approving until the other reviewers get to take another look, but this is looking great. Thanks!

bpf/generictracer/jvm.h
```diff
@@ -0,0 +1,57 @@
...
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_helpers.h>
+#include <generictracer/types/jvm.h>
+
+#include <generictracer/maps/jvm_heap_summary_samples.h>
+#include <generictracer/maps/jvm_mem_pool_samples.h>
+
+// Use https://godbolt.org/z/YcodaPhvY to understand the memory layout of `GCHeapSummary` C++ class
+// https://github.com/openjdk/jdk/blob/jdk-21%2B35/src/hotspot/share/gc/shared/gcHeapSummary.hpp#L76
+struct jvm_gc_heap_summary {
```
mmat11, line 15: is this subject to change between jdk versions?

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -223,6 +226,12 @@ func (p *Tracer) constants() map[string]any {
 
 	m["g_bpf_debug"] = p.cfg.EBPF.BpfDebug
 	m["g_bpf_traceparent_enabled"] = p.cfg.EBPF.TrackRequestHeaders || p.cfg.EBPF.ContextPropagation.IsEnabled()
+	m["jvm_runtime_metrics_enabled"] = uint8(0)
+	m["jvm_sampling_interval_ns"] = uint64(0)
+	if p.cfg.JVMRuntimeMetrics.Enabled {
+		m["jvm_runtime_metrics_enabled"] = uint8(1)
+		m["jvm_sampling_interval_ns"] = uint64(p.cfg.JVMRuntimeMetrics.SamplingInterval.Nanoseconds())
```
mmat11, line 231: ebpf side, can we infer that if the sampling interval is > 0, the runtime metrics are enabled? this way we can have only one variable addition; we should also add permutations for this variable in the verifier tests -> https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/internal/ebpf/verifier/bpf_verifier_test.go#L153

bpf/generictracer/jvm.c
```diff
@@ -0,0 +1,314 @@
...
+    __builtin_memset(e, 0, sizeof(*e));
+    e->type = EVENT_JVM_MEM_POOL_GC;
+    e->timestamp = ts;
+    jvm_fill_mem_pool_pid_fields(pid_tgid, e);
+    e->gc_when_type = when;
+    e->init_size = init_size;
+    e->used = used;
+    e->committed = committed;
+    e->max_size = max_size;
+    __builtin_memcpy(e->manager, key.manager, sizeof(e->manager));
```
mmat11: nit: we should use bpf_memset/bpf_memcpy from bpf_builtins.h instead of these builtins

bpf/generictracer/jvm.h
REASY: I checked the OpenJDK LTS lines 8, 11, 17, 21, and 25. `GCTracer::report_gc_heap_summary(GCWhen::Type, const GCHeapSummary&) const` exists in all of them, and the base `GCHeapSummary` layout is stable: `VirtualSpaceSummary _heap` followed by `size_t _used`.

So yes, this is private HotSpot internals and not a public ABI, but for the LTS versions we care about the method signature and the field layout have stayed consistent. The BPF layout assumes the usual 64-bit C++ object layout: vptr + three `VirtualSpaceSummary` pointers + `used`.

| JDK LTS | `GCHeapSummary` layout source | `GCTracer::report_gc_heap_summary` source |
|---|---|---|
| 8 | [`gcHeapSummary.hpp#L90-L98`](https://github.com/openjdk/jdk8u-dev/blob/master/hotspot/src/share/vm/gc_implementation/shared/gcHeapSummary.hpp#L90-L98) | [`gcTrace.hpp#L126`](https://github.com/openjdk/jdk8u-dev/blob/master/hotspot/src/share/vm/gc_implementation/shared/gcTrace.hpp#L126) |
| 11 | [`gcHeapSummary.hpp#L90-L98`](https://github.com/openjdk/jdk11u-dev/blob/master/src/hotspot/share/gc/shared/gcHeapSummary.hpp#L90-L98) | [`gcTrace.hpp#L121`](https://github.com/openjdk/jdk11u-dev/blob/master/src/hotspot/share/gc/shared/gcTrace.hpp#L121) |
| 17 | [`gcHeapSummary.hpp#L76-L84`](https://github.com/openjdk/jdk17u-dev/blob/master/src/hotspot/share/gc/shared/gcHeapSummary.hpp#L76-L84) | [`gcTrace.hpp#L103`](https://github.com/openjdk/jdk17u-dev/blob/master/src/hotspot/share/gc/shared/gcTrace.hpp#L103) |
| 21 | [`gcHeapSummary.hpp#L76-L84`](https://github.com/openjdk/jdk21u/blob/master/src/hotspot/share/gc/shared/gcHeapSummary.hpp#L76-L84) | [`gcTrace.hpp#L103`](https://github.com/openjdk/jdk21u/blob/master/src/hotspot/share/gc/shared/gcTrace.hpp#L103) |
| 25 | [`gcHeapSummary.hpp#L76-L84`](https://github.com/openjdk/jdk25u/blob/master/src/hotspot/share/gc/shared/gcHeapSummary.hpp#L76-L84) | [`gcTrace.hpp#L103`](https://github.com/openjdk/jdk25u/blob/master/src/hotspot/share/gc/shared/gcTrace.hpp#L103) |

mmat11: sounds good 👍🏻

bpf/logenricher/types.h
```diff
@@ -53,13 +53,8 @@ enum tty_driver_type___new {
 };
 
 enum tty_driver_subtype___new {
-    SYSTEM_TYPE_TTY = 1,
```
mmat11, line 56: why are these deleted?

cmd/obi-schema/main.go
```diff
@@ -798,9 +798,10 @@ func (g *SchemaGenerator) customMapper() func(reflect.Type) *jsonschema.Schema {
 		if t == reflect.TypeFor[time.Duration]() {
 			return &jsonschema.Schema{
 				Type:        "string",
-				Description: "Duration in Go format (e.g., '30s', '5m', '1ms')",
-				Pattern:     "^[0-9]+(ms|s|m)$",
-				Examples:    []any{"30s", "5m", "1ms"},
+				Format:      "duration",
+				Description: "Duration in Go format (e.g., '30s', '5m', '1ms', '500us', '0.5s')",
```
mmat11: is this needed?

bpf/logenricher/types.h
REASY: Good catch, thanks! I accidentely formatted the previously and then removed instead of undoing the changes. The changes are reverted.

cmd/obi-schema/main.go
REASY: Not really, removed `Description` field.

grcevski (approved):
LGTM! Thanks for all this work, this looks great now!

grcevski:
I think there are still some problems with the code, the jvm metrics integration test failed https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/27773051271/job/82208069575?pr=2305 Looks like return code 1, so something didn't run properly.

grcevski:
```
obi-1  |     jvm_live_test.go:127: java stdout: [12.266s][info][gc] GC(80) Pause Young (Allocation Failure) 34M->33M(123M) 2.195ms
obi-1  |     jvm_live_test.go:55: 
obi-1  |         	Error Trace:	/src/pkg/internal/ebpf/generictracer/jvm_live_test.go:286
obi-1  |         	            				/src/pkg/internal/ebpf/generictracer/jvm_live_test.go:55
obi-1  |         	Error:      	timed out waiting for JVM runtime events
obi-1  |         	Test:       	TestJVMRuntimeEventsLiveFromHotSpotProbes
obi-1  |         	Messages:   	
obi-1  | --- FAIL: TestJVMRuntimeEventsLiveFromHotSpotProbes (33.76s)
obi-1  | FAIL
obi-1  | FAIL	go.opentelemetry.io/obi/pkg/internal/ebpf/generictracer	33.860s
obi-1  | FAIL
```

REASY:
The failure in CI was real. I brought namespace awareness back, 0ee9458c9186594ebc320ae8c358567caa7770e1,  because the BPF-side USDT argument decoding lookup happens in the context of the probed task, and the PID identity used there is not always just the host PID.

he USDT attach itself is done from userspace using the host PID, but the BPF program resolves the argument spec at runtime from `obi_usdt_ip_to_spec_id`. That lookup key must match what BPF observes when the USDT probe fires. With PID filtering enabled, `valid_pid(pid_tgid)` can return the namespace PID, and the lookup also needs the PID namespace id to avoid collisions between containers where the same namespace PID can exist, e.g. PID 1 in multiple containers.

The host-PID-only refactor broke this path for HotSpot memory-pool probes: heap-summary events still worked because they are normal uprobes and do not use `obi_usdt_arg()` or the USDT spec map, but memory-pool events disappeared because `obi_usdt_arg()` could not resolve the USDT spec for the firing probe.

CC: @grcevski @rafaelroquetto

cmd/obi-schema/main.go
```diff
@@ -797,10 +797,10 @@ func (g *SchemaGenerator) customMapper() func(reflect.Type) *jsonschema.Schema {
 		// Handle time.Duration as a string (Go duration format)
 		if t == reflect.TypeFor[time.Duration]() {
 			return &jsonschema.Schema{
-				Type:        "string",
-				Description: "Duration in Go format (e.g., '30s', '5m', '1ms')",
-				Pattern:     "^[0-9]+(ms|s|m)$",
-				Examples:    []any{"30s", "5m", "1ms"},
+				Type:     "string",
+				Format:   "duration",
+				Pattern:  "^[-+]?(([0-9]+(\\.[0-9]*)?|\\.[0-9]+)(ns|us|µs|μs|ms|s|m|h))+$",
```
mmat11: copilot says this is more restrictive than Go duration format, which would then be a breaking change. Is this change required or can we use the duration in Go format?

internal/test/integration/components/docker/compose.go
```diff
@@ -69,6 +70,16 @@ func (c *Compose) Up() error {
 	return c.command("up", "--build", "--detach", "--quiet-pull")
 }
 
+func (c *Compose) Run(service string) error {
+	c.skipWait = true
+	args := []string{"up"}
+	if os.Getenv("SKIP_DOCKER_BUILD") == "" {
```
mmat11, line 76: why were this changes added? anything missing in the current testing setup?

internal/test/integration/docker-compose-jvm-runtime-events-live.yml
```diff
@@ -0,0 +1,25 @@
...
+    environment:
+      GOCACHE: /tmp/go-build-cache
+      GOMODCACHE: /tmp/go-mod-cache
+    command:
+      - go
+      - test
+      - -tags=jvm_live
+      - ./pkg/internal/ebpf/generictracer
+      - -run
+      - ^TestJVMRuntimeEventsLiveFromHotSpotProbes$
```
mmat11, line 22: can't this be run like a normal integration or unit test? why is a docker compose file required to run `go test`?

mmat11 (approved):
very nice work! left a couple of comments, but LGTM!

internal/test/integration/docker-compose-jvm-runtime-events-live.yml
REASY: The repo has a few eBPF test modes: normal Linux unit tests for parsing/selection logic, privileged unit tests via `sudo make test-privileged`, verifier tests via `bpf_verifier_tests`, and Docker Compose integration tests for real app/eBPF behavior. This test is in the last category: it attaches real uprobes/USDT probes to a real HotSpot JVM and needs root, tracefs/securityfs, `java`, and `javac`. The compose wrapper provides that reproducible privileged JDK environment while still running a focused Go test internally.

internal/test/integration/components/docker/compose.go
REASY: This was added for the JVM live probe test, which is a one-shot compose service running `go test` inside a privileged Linux/JDK container. The existing `Up()` helper starts long-running stacks detached and assumes the outer integration test will drive traffic/assertions from the host. For this test we need `docker compose up --abort-on-container-exit --exit-code-from obi` so the inner test result becomes the integration test result. Nothing is missing from the existing detached-stack helper; this is just a small helper for command-style compose tests.

cmd/obi-schema/main.go
REASY: Good catch, not really, reverted it.

grcevski:
I think the state is now good and all comments are resolved. I'm OK with merging this PR, any objections @mmat11 and @rafaelroquetto ?

mmat11:
> I think the state is now good and all comments are resolved. I'm OK with merging this PR, any objections @mmat11 and @rafaelroquetto ?

I still see some changes related to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2305#discussion_r3442703350 I think we can merge after all those changes are reverted

grcevski:
> > I think the state is now good and all comments are resolved. I'm OK with merging this PR, any objections @mmat11 and @rafaelroquetto ?
> 
> I still see some changes related to [#2305 (comment)](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2305#discussion_r3442703350) I think we can merge after all those changes are reverted

Ah I guess you are referring to the duration format here: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2305/changes#diff-61bb6c5b79e263b8e5f7bb05d44b847ee841f2721d713a1a60f873c45688e32dR601

@REASY can we revert all those changes related to the duration format? I don't think you need more time precision than a millisecond, which was supported before.

REASY:
@mmat11 @grcevski I've restored duration handling in e6da4bae7822d981a63784f2f2be770dc9796386, please take a look

mmat11 (approved):
thanks!

marctc:
hey @REASY, thanks a lot for working on this, much appreciated! It came in the right time as I recently added the Go part. I couldn't review this PR because i was on leave, but I did a pass now that I'm back. Implementation looks good, but there are something that I would have seen differently. 

- I see that we added `application_jvm`. My intention was that users enable `application_*` if the want to enable any susbset of metrics for their running applications, regardless of the language. In this cases we have a separate `application_runtime` and `application_jvm`, which I'm not very happy about it. Also we wouldn't need  `enable: true` for jvm block if the feature is present.
- Instead of using `obi.jvm.heap.used` we should use the attribute [jvm.memory.type](https://opentelemetry.io/docs/specs/semconv/registry/attributes/jvm/#jvm-memory-type) and use always `jvm.memory.used`
- We need to add documentation for this part in `devdocs/runtimes`.

We can discuss here, in a separate issues or slack and I can work on the issues myself, but I just want to have your input.

grcevski:
Hey @marctc, let's just make the changes as you suggest, combine all runtime metrics under application_runtime and switch to use the semconv as you suggested.

REASY:
Hi, folks, @marctc 

I trust your judgment how this should be done properly. My only issue is that if one does not filter properly `jvm.memory.used` (filter by `jvm.memory.type`), the average metric across different sources that produce that metric might look weird and this would be a breaking change. While if one uses explicit `obi.jvm.heap.used`, they supposedly know what they are doing.

PR #2352, opened 2026-06-12
Restrict `/all-vm-tests` to users with write permission and prevent Docker secret exposure on dispatched refs

MrAlias:
### Motivation

- Prevent a malicious PR from causing repository secrets to be exposed by a slash-command that dispatched a privileged workflow using a PR head `ref` and then checked out and executed PR-controlled code.
- Ensure the slash-command only runs when the commenter has actual repository write/admin/maintain permission rather than relying on `author_association` alone.
- Avoid performing secret-backed Docker login when the VM integration workflow is run via `workflow_dispatch` with an arbitrary `ref` to stop dispatched runs from receiving `DOCKER_TOKEN_EBPF_INSTRUMENTATION` when they will execute untrusted code.

### Description

- In `.github/workflows/slash_command_all_vm_tests.yml` changed job permissions to the minimal needed (`actions: write`, `issues: write`, `pull-requests: read`) and added a live permission check step that calls the GitHub API via `gh api repos/{owner}/{repo}/collaborators/{user}/permission` and requires `admin|maintain|write` before continuing. 
- In `.github/workflows/workflow_integration_tests_vm.yml` added an `ALLOW_DOCKER_TOKEN` guard computed as `github.event_name != 'workflow_dispatch' || inputs.ref == ''` and used it to gate the Docker Hub login step so that `secrets.DOCKER_TOKEN_EBPF_INSTRUMENTATION` is not applied to manually-dispatched runs that reference an arbitrary `ref`.
- Kept the dispatch behavior (the slash command still resolves the PR head SHA and dispatches the VM workflow) but added the above authorization and secret-safety controls to eliminate the secret-exposure attack surface.

### Testing

- Parsed both workflows with Ruby YAML (`ruby -e 'require "yaml"; ARGV.each { |f| YAML.load_file(f) ... }'`) to ensure the modified files are valid YAML and parsing succeeded.

### DIFF
--- a/.github/workflows/slash_command_all_vm_tests.yml
+++ b/.github/workflows/slash_command_all_vm_tests.yml
@@ -1,7 +1,7 @@
 name: "Slash: /all-vm-tests"
 
 # Listens for /all-vm-tests on PR comments. Validates commenter has write
-# access, resolves PR head SHA, then dispatches Integration tests VM with
+# access, verifies actual write permission, resolves PR head SHA, then dispatches Integration tests VM with
 # all_kernels=true. Posts a reply with the run URL.
 on:
   issue_comment:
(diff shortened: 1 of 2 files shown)


rafaelroquetto (approved):
Nice, thank you so much for this!

MrAlias:
Merging early due to sensitive nature.

PR #2353, opened 2026-06-12
Fix top-level JSON fallback extraction

MrAlias:
### Motivation

- A regex-based fallback scanned the captured request/response prefix for fields like `model` and `id` and could match nested, attacker-controlled keys, allowing injected values to surface as GenAI telemetry attributes. 
- The intent is to preserve best-effort parsing for truncated JSON while only promoting genuine top-level fields into telemetry. 
- The change prevents nested or late-occurring keys from being misinterpreted as top-level fields and thus reduces telemetry integrity/high-cardinality risk.

### Description

- Replace the unanchored regex fallback with a streaming JSON top-level extractor using `json.Decoder` that only returns string values for top-level keys and respects a configurable search `window`.
- Add `skipJSONValue(dec *json.Decoder)` to advance the decoder past non-matching values safely, and make `extractModelField` use the top-level extractor.
- Update `rerank` logic to reuse the new `extractJSONStringField` (removing the old `modelFieldRegexp` and `regexp` import) so rerank model fallback only considers top-level model keys in the request prefix.
- Add regression tests that assert nested `id`/`model` keys are ignored (including nested `model` after the previous search window) while preserving existing truncated top-level parsing behavior.

### Testing

- Ran targeted unit tests with `go test ./pkg/ebpf/common/http -run 'TestExtract|TestParse(OpenAI|Vendor|Anthropic|Embedding)'`, which passed. 
- Ran the full package tests with `go test ./pkg/ebpf/common/http`, which passed.

### DIFF
--- a/pkg/ebpf/common/http/partial_json.go
+++ b/pkg/ebpf/common/http/partial_json.go
@@ -5,12 +5,11 @@ package ebpfcommon // import "go.opentelemetry.io/obi/pkg/ebpf/common/http"
 
 import (
 	"bytes"
+	"encoding/json"
 	"io"
 	"log/slog"
 	"net/http"
-	"regexp"
 	"strings"
-	"sync"
 
 	jsoniter "github.com/json-iterator/go"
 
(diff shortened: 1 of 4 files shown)


rafaelroquetto (approved):
LGTM

Issue #2034, opened 2026-05-05
Timeout does not cancel blocked Java extraction


MrAlias:
## Summary

The route harvester wraps Java route extraction in a goroutine and returns on timeout, but it does not cancel the underlying extraction work. A blocked Java attach can therefore continue running after timeout and accumulate leaked goroutines or related resources.

## Impact

Repeated timeouts against problematic Java processes can degrade or exhaust the OBI agent. The impact is local availability only, so `low` severity is appropriate.

## Environment

- First identified in: b523ed4a9074500e0b70be96fb90b4cee9d05cac
- Reviewed against current `main` commit: d51a98d730f950b8d759a8e592669db13d2e4843

## Evidence

- `pkg/internal/transform/route/harvest/harvester.go` starts extraction in a goroutine and returns on `ctx.Done()`.
- The goroutine still calls `javaExtractRoutes(fileInfo.Pid)` with no cancellation path.

## Steps to reproduce

1. From the repository root on `main`, add `pkg/internal/transform/route/harvest/leak_poc_test.go` with this exact test:
   ```go
   package harvest
   import (
       "runtime"
       "testing"
       "time"
       "go.opentelemetry.io/obi/pkg/components/exec"
       "go.opentelemetry.io/obi/pkg/components/svc"
   )
   func TestRouteHarvesterGoroutineLeak(t *testing.T) {
       h := NewRouteHarvester(nil, 50*time.Millisecond)
       blockCh := make(chan struct{})
       h.javaExtractRoutes = func(pid int32) (*RouteHarvesterResult, error) {
           <-blockCh
           return nil, nil
       }
       fi := &exec.FileInfo{Service: svc.Attrs{SDKLanguage: svc.InstrumentableJava}, Pid: 1234}
       before := runtime.NumGoroutine()
       for i := 0; i < 5; i++ {
           _, _ = h.HarvestRoutes(fi)
       }
       time.Sleep(100 * time.Millisecond)
       after := runtime.NumGoroutine()
       if after <= before {
           t.Fatalf("expected goroutines to increase, before=%d after=%d", before, after)
       }
       t.Logf("goroutines before=%d after=%d", before, after)
   }
   ```
2. Run `GOTOOLCHAIN=auto go test ./pkg/internal/transform/route/harvest -run TestRouteHarvesterGoroutineLeak -count=1 -v`.
[...]

PR #2360, opened 2026-06-12
Add context-aware JVM attach

MrAlias:
## Summary

- Add `JAttacher.AttachContext(ctx, pid, argv, ignoreOnJ9)` to the internal `jvmtools` fork while keeping `Attach` as a `context.Background()` wrapper.
- Make the HotSpot attach path honor cancellation during attach-listener startup, Unix socket dialing, and command writes.
- Ensure temporary `.attach_pid*` files are removed on all startup exit paths and close the HotSpot socket when cancellation interrupts a write.
- Keep OpenJ9 behavior unchanged apart from early context checks before entering that path.
- Add Linux package tests for canceled attach contexts, startup cancellation cleanup, and blocked-write cancellation.

## Why

This is step 3 from #2034's fork-cancellation PR stack. It gives the forked attach code the cancellable boundary needed before route harvesting threads its timeout context into Java route extraction.

## Validation

- `go test ./pkg/internal/jvmtools/...`


grcevski (approved):
LGTM!

grcevski:
This may need merge with main to include the java fix.

PR #2364, opened 2026-06-12
Better java route harvester

### LINKED ISSUE #2304: Java route harvesting can freeze target JVM while dumping symbol table
## Summary

Java route harvesting currently uses dynamic attach to run `jcmd VM.symboltable -verbose` against the target JVM:

- `pkg/internal/transform/route/harvest/java.go`

For JVMs with very large symbol tables, this can trigger a long stop-the-world safepoint in the target application. In #2301, the reporter observed application freezes lasting 20-100 seconds while `jcmd <PID> VM.symboltable -verbose` printed hundreds of thousands of lines.

PR #2303 disables Java route harvesting by default as an immediate mitigation, but we should track the underlying production-safety issue separately.

## Impact

This affects the instrumented Java application, not just the OBI agent.

A route discovery attempt can make the target JVM unresponsive for tens of seconds. The current `route_harvester_timeout` only limits how long OBI waits for the extraction result. It does not prevent or cancel JVM-side work once the diagnostic command has triggered a safepoint.

## Related Work

- Discussion: #2301
- Mitigation PR: #2303
- Related issue: #2034

#2034 is related but narrower. It tracks that Java route extraction is not truly cancellable from OBI and can retain blocked workers after timeout. This issue tracks the target-JVM safety problem: the diagnostic command itself can pause the application.

## Evidence

The reported safepoint log includes `DumpHashtable` and a long time at safepoint while dumping the symbol table. The Java route harvester currently extracts routes from the output of:

```text
jcmd VM.symboltable -verbose
```

That command can scale poorly with symbol table size and can pause the JVM during collection.

## Suggested Direction

Short term:

- Keep Java route harvesting disabled by default.
- Document that enabling Java route harvesting can pause the target JVM on large applications.
- Make opt-in behavior explicit in both v1 and v2 configuration docs and schemas.

Long term:

- Investigate a route discovery approach that does not require dumping the full JVM symbol table.
- If dynamic attach remains necessary, add stronger safety controls before enabling by default again.
- Coordinate with #2034 so timeout/cancellation behavior is improved, while recognizing that cancellation alone may not prevent JVM safepoints already in progress.

## Acceptance Criteria

- Java route harvesting is safe by default for production users.
- Documentation clearly describes the risk and opt-in behavior.
- The v1 and v2 config defaults, examples, and schemas agree.
- There is a tested strategy for avoiding, bounding, or isolating target-JVM pauses before Java route harvesting is enabled by default again.

grcevski:
## Summary

Users discovered that our existing Java route harvester could hold up the JVM for unacceptable periods of time. This is due to potentially slow socket communication between OBI and the target JVM, so the JVM waits until OBI has read all the symbols.

To fix this problem I've rewritten the Java route harvester to use the same approach as the node.js harvester. Instead of using the jcmd interface to pull the loaded symbols, the harvester parses the class files found on the java process command line and CLASSPATH environment variables. This also improves our support, since now we support OpenJ9 route harvesting, which we couldn't before because OpenJ9 didn't support VM symbols listing through jcmd.

I've re-enabled the java harvester by default and I shortened the wait period to 5 seconds after Java start. Since we don't really care about what classes are loaded (and instead we parse files), we don't have to wait anymore, at least not 60 seconds. I might follow-up with PR to remove the java delay altogether, I decided against it here because the PR got large.

Closes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2304

## Validation


grcevski:
Ignore the test_files while reviewing, I decided to commit them like we have for nodejs, I needed them in binary and expanded class form that proves the parsing works with jars and directories on the classpath.

pkg/internal/transform/route/harvest/java/test_files/classes/com/example/JaxController.class
MrAlias, line 1: This is checking in a bunch of binary files. Is that intentional?

grcevski: Yes, I need them for the unit tests, since the parser of routes is parsing binaries. Similar to the nodejs checked in test files, except those are text files.

MrAlias: Gotcha 👍

pkg/internal/transform/route/harvest/java/classpath.go
```diff
@@ -0,0 +1,204 @@
...
+		if !ok {
+			return nil, fmt.Errorf("invalid Java jar path %q", launch.jar)
+		}
+		if isRegularFile(root) {
+			return []scanRoot{{path: root}}, nil
+		}
+		return nil, fmt.Errorf("java jar path %q is not a regular file", launch.jar)
+	}
+
+	if launch.classpath == "" {
```
MrAlias, line 59: This misses Java's default user classpath behavior. If the process is launched as `java com.example.Main` without `-cp`, `-classpath`, `--class-path`, or `CLASSPATH`, the JVM uses `.` as the classpath, but this returns `no Java jar or classpath found`.

That means exploded-class deployments launched from their app directory will not be harvested. I think this should fall back to `.` resolved against the process cwd. Empty classpath elements should probably be treated as `.` as well, since Java gives them that meaning.

pkg/internal/transform/route/harvest/java/classpath.go
```diff
@@ -0,0 +1,204 @@
...
+		return javaLaunch{classpath: env[envClasspath]}
+	}
+	return javaLaunch{}
+}
+
+func scanRootsFromClasspath(root, cwd, classpath string) []scanRoot {
+	var dirs []scanRoot
+	var jars []scanRoot
+	for _, entry := range filepath.SplitList(classpath) {
+		if entry == "" || strings.Contains(entry, "*") {
```
MrAlias: Skipping every classpath entry containing `*` drops a common Java launch pattern: `-cp /app/app.jar:/app/lib/* com.example.Main`. Java expands wildcard classpath entries to matching JARs in that directory, but this code ignores them, so routes packaged in those JARs are never harvested.

This should expand wildcard entries to `.jar`/`.war` files under the target directory while still applying the existing root/symlink checks and scan limits.

pkg/internal/transform/route/harvest/java/classpath.go
```diff
@@ -0,0 +1,204 @@
...
+		if info.IsDir() {
+			dirs = append(dirs, scanRoot{path: path, dir: true})
+			continue
+		}
+		if info.Mode().IsRegular() && isJavaArchive(path) {
+			jars = append(jars, scanRoot{path: path})
+		}
+	}
+
+	if len(dirs) > 0 {
```
MrAlias: Returning only directories when any directory is present breaks mixed classpaths. For example, `-cp /app/classes:/app/app.jar:/app/lib/*` will scan only `/app/classes` and skip `app.jar`, even though controllers can live in either location.

Classpath order can legitimately include both directories and archives. This should return all valid scan roots in classpath order instead of preferring directories globally.

pkg/internal/transform/route/harvest/java/extractor.go
```diff
@@ -0,0 +1,132 @@
...
+	out := make([]string, 0, len(m))
+	for k := range m {
+		out = append(out, k)
+	}
+	return out
+}
+
+func sortRoutes(routes []string) []string {
+	sort.Slice(routes, func(i, j int) bool {
+		hasParamsI := strings.Contains(routes[i], "{")
```
MrAlias: Now that this harvester preserves Quarkus-style `:param` routes, this sorter still treats them as non-parameter routes because it only checks for `{`. Java routes are returned as `PartialRoutes`, and `NewPartialRouteMatcher` tries roots in the supplied order, so a harvested `/api/:version` can be tried before an exact `/api/v1` and make `/api/v1` requests match the broader template.

The sorter should classify both `{id}` and `:id` as parameterized routes, or otherwise use the same wildcard semantics as the matcher, so exact/static routes win first.

pkg/internal/transform/route/harvest/java/normalize.go
```diff
@@ -0,0 +1,177 @@
...
+	parts := strings.Split(s, "/")
+	for i, part := range parts {
+		if !strings.Contains(part, ":") {
+			continue
+		}
+		if !strings.HasPrefix(part, ":") {
+			if i <= 1 {
+				return ""
+			}
+			return routePrefix(parts[:i])
```
MrAlias: This still converts malformed colon-containing segments into a broader valid prefix. For example, `/api/foo:bar/details` becomes `/api`, and `/items/:123` becomes `/items`. Because Java results use `PartialRoutes`, adding those shortened prefixes can produce broad route matches for unrelated requests instead of rejecting the malformed annotation.

I think invalid colon segments should cause the whole route to be rejected, not truncated to a prefix. That keeps the harvester from inventing routes that were never declared.

MrAlias (dismissed):
I like the direction of replacing the JVM attach / `jcmd VM.symboltable` route harvester with a classpath/JAR class-file scanner, but I do not think this is ready to merge yet. This changes Java route discovery from runtime symbol-table scraping to filesystem/classpath discovery, so we need to preserve common Java launch behavior pretty carefully.

The remaining issues are all in that new discovery path: default classpath handling, wildcard classpath handling, mixed directory/JAR classpaths, and ordering of colon-parameter routes before partial matching.

Non-blocking, but please also add a short note or script for regenerating the checked-in `.class` and `.jar` fixtures from `test_files/src`, including the exact JAR/WAR layouts. The source fixtures are useful, but the generated archive contents are not obvious from the repository.

pkg/internal/transform/route/harvest/java/classpath.go
grcevski: I was concerned that this would be too expensive. I purposefully limited to the most common way the applications can be deployed, because scanning all the dependencies will be very time consuming on large Java applications. I think the generic case about this being a problem is valid, but it's not how most java web REST services are going be deployed. It's a tradeoff.

grcevski:
> Non-blocking, but please also add a short note or script for regenerating the checked-in `.class` and `.jar` fixtures from `test_files/src`, including the exact JAR/WAR layouts. The source fixtures are useful, but the generated archive contents are not obvious from the repository.

Yeah good point, I'll add that. I manually compiled and zipped the files, but I can make a script that regenerates the binaries and check that in.

pkg/internal/transform/route/harvest/java/classpath.go
grcevski: Fixed.

pkg/internal/transform/route/harvest/java/classpath.go
grcevski: Fixed.

pkg/internal/transform/route/harvest/java/classpath.go
grcevski: I decided to do it, the common approach won't have this issue anyway, so it will be more expensive search for the uncommon case. Fixed.

pkg/internal/transform/route/harvest/java/normalize.go
grcevski: Fixed, I disallow all paths now that have weird colons.

grcevski:
> > Non-blocking, but please also add a short note or script for regenerating the checked-in `.class` and `.jar` fixtures from `test_files/src`, including the exact JAR/WAR layouts. The source fixtures are useful, but the generated archive contents are not obvious from the repository.
> 
> Yeah good point, I'll add that. I manually compiled and zipped the files, but I can make a script that regenerates the binaries and check that in.

I made a script that compiles them all. The dependent packages were all faked to look like what's actually done in the respective dependencies, which is why just javac and jar are needed.

pkg/internal/transform/route/harvest/java/extractor.go
grcevski: Fixed.

MrAlias (commented):
The classpath/JAR discovery fixes look good.

pkg/internal/transform/route/harvest/java/test_files/build-java-route-test-files.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+set -euo pipefail
+
+script_dir="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)"
+repo_root="$(cd -- "${script_dir}/.." && pwd)"
+
+test_dir=.
```
MrAlias: Thanks for adding the fixture regeneration script. Right now `test_dir=.` makes the script depend on the caller's working directory; running it from the repo root would look for `./src` instead of this fixture directory. Since `script_dir` is already computed, please anchor `test_dir` there and drop the unused `repo_root` line.

```suggestion
test_dir="${script_dir}"
```

grcevski:
I committed your suggestion for the compile/package with jar script.

MrAlias (approved):

PR #2365, opened 2026-06-12
Add configurable hostname fetcher to NodeMeta

jereenv:
## Summary

- Adds `VendorHostnameAttr` variable (empty by default): vendors set this to their own attribute key to activate the fetcher
- Adds `hostnameFetcher` that resolves the machine hostname via FQDN lookup and injects it into `NodeMeta.Metadata` under the configured attribute name
- Registers `hostnameFetcher` first in the `NewNodeMeta()` chain (lowest priority, so cloud/kube fetchers take precedence)
- Uses `hostname.CreateResolver` for DNS-based resolution
- Refactors `labelNamesTargetInfo` and `labelValuesTargetInfo` to skip metadata entries that duplicate existing `target_info` label names

## Validation

### DIFF
--- a/pkg/appolly/meta/meta_node.go
+++ b/pkg/appolly/meta/meta_node.go
@@ -26,10 +26,11 @@ func nslog() *slog.Logger {
 
 var connectionTimeout = 2 * time.Second
 
-// FallbackHostIDAttr, when non-empty, enables a hostname-based fallback
+// VendorHostnameAttr, when non-empty, enables a hostname-based fallback
 // for resource attribution. The hostname is resolved via FQDN lookup and stored
-// under this attribute name. Empty by default — vendors activate it by setting their own key
-var FallbackHostIDAttr = ""
+// as a Metadata entry under this attribute name, not in the HostID field.
+// Empty by default — vendors activate it by setting their own key.
+var VendorHostnameAttr = ""
 
 // some attributes from the node need to be filtered out, becausee ither
 // - they are not common to all the services and is going to be specified for each service instance
--- a/pkg/appolly/meta/meta_node.go
+++ b/pkg/appolly/meta/meta_node.go
@@ -26,6 +26,11 @@ func nslog() *slog.Logger {
 
 var connectionTimeout = 2 * time.Second
 
+// FallbackHostIDAttr, when non-empty, enables a hostname-based fallback
+// for resource attribution. The hostname is resolved via FQDN lookup and stored
+// under this attribute name. Empty by default — vendors activate it by setting their own key
+var FallbackHostIDAttr = ""
+
 // some attributes from the node need to be filtered out, becausee ither
 // - they are not common to all the services and is going to be specified for each service instance
 // - they are explicitly added later in the exporters and we don't want duplications
@@ -73,6 +78,7 @@ func NewNodeMeta(
 		// some fetchers will only retrieve the host name while others
 		// will retrieve also host attributes that will be merged
 		// in order of the priority below (the later the highest)
+		hostnameFetcher,
 		linuxLocalFetcher,
 		kubeNodeFetcher(kubeInformer),
 		otelNodeFetcher(azurevm.New()),
--- a/pkg/appolly/meta/meta_node_hostname.go
+++ b/pkg/appolly/meta/meta_node_hostname.go
@@ -12,22 +12,28 @@ import (
 )
 
 func hostnameFetcher(_ context.Context) (NodeMeta, error) {
-	if FallbackHostIDAttr == "" {
+	if VendorHostnameAttr == "" {
 		return NodeMeta{}, nil
 	}
 	resolver := hostname.CreateResolver("", "", true)
 	fullHostName, _, err := resolver.Query()
-	if err != nil || fullHostName == "" {
-		slog.Warn("can't resolve hostname for fallback",
+	if err != nil {
+		slog.Warn("can't resolve hostname for fallback attribute",
 			"component", "meta.hostnameFetcher",
-			"attribute", FallbackHostIDAttr,
+			"attribute", VendorHostnameAttr,
 			"error", err)
+		return NodeMeta{}, nil
+	}
+	if fullHostName == "" {
+		slog.Warn("resolved empty hostname for fallback attribute",
+			"component", "meta.hostnameFetcher",
+			"attribute", VendorHostnameAttr)
 		return NodeMeta{}, nil
 	}
 	return NodeMeta{
 		Metadata: []Entry{
 			{
-				Key:   attr.Name(FallbackHostIDAttr),
+				Key:   attr.Name(VendorHostnameAttr),
 				Value: fullHostName,
 			},
 		},
--- a/pkg/appolly/meta/meta_node_hostname.go
+++ b/pkg/appolly/meta/meta_node_hostname.go
@@ -1,0 +1,35 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package meta // import "go.opentelemetry.io/obi/pkg/appolly/meta"
+
+import (
+	"context"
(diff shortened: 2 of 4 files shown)


pkg/appolly/meta/meta_node.go
```diff
@@ -26,6 +26,11 @@ func nslog() *slog.Logger {
 
 var connectionTimeout = 2 * time.Second
 
+// FallbackHostIDAttr, when non-empty, enables a hostname-based fallback
+// for resource attribution. The hostname is resolved via FQDN lookup and stored
+// under this attribute name. Empty by default — vendors activate it by setting their own key
+var FallbackHostIDAttr = ""
```
mariomac: What's the purpose of this public, global attribute? Shouldn't we invoke `hostnameFetcher` as a fallback solution, no matter if it's defined or not?

pkg/appolly/meta/meta_node_hostname.go
```diff
@@ -0,0 +1,35 @@
...
+	fullHostName, _, err := resolver.Query()
+	if err != nil || fullHostName == "" {
+		slog.Warn("can't resolve hostname for fallback",
+			"component", "meta.hostnameFetcher",
+			"attribute", FallbackHostIDAttr,
+			"error", err)
+		return NodeMeta{}, nil
+	}
+	return NodeMeta{
+		Metadata: []Entry{
```
mariomac, line 34: For consistence with the rest of Host ID name fetchers, shouldn't this be:
```go
return NodeMeta{HostID: fullHostName}
```
?

mariomac (commented):
Thank you for your contributions @jereenv ! I'm missing something w.r.t. `FallbackHostIDATTR`.

pkg/appolly/meta/meta_node_hostname.go
```diff
@@ -0,0 +1,35 @@
...
+	"go.opentelemetry.io/obi/pkg/internal/traces/hostname"
+)
+
+func hostnameFetcher(_ context.Context) (NodeMeta, error) {
+	if FallbackHostIDAttr == "" {
+		return NodeMeta{}, nil
+	}
+	resolver := hostname.CreateResolver("", "", true)
+	fullHostName, _, err := resolver.Query()
+	if err != nil || fullHostName == "" {
```
rafaelroquetto: I think in the case in which `err == nil && fullHostName == ""`, this prints an strange warning:
```
can't resolve hostname for fallback (...) err=nil
```

i.e. a `nil` error but no further information. I'd sugest splitting this in 2, one warning that spells "empty hostname" and the other for the case in which `err != nil`

pkg/appolly/meta/meta_node.go
jereenv: Without the variable the attribute name would need to be hardcoded in OBI, making it vendor-specific. This way vendors define their own key and opt in explicitly. 
Will rename `FallbackHostIDAttr` to `VendorHostnameAttr` to better reflect its purpose.

pkg/appolly/meta/meta_node_hostname.go
jereenv: Since `HostID` already holds stable machine identifiers from the existing fetchers, adding a hostname there could conflict with those values. The hostname goes into Metadata under a separate vendor-defined key instead.

pkg/appolly/meta/meta_node_hostname.go
jereenv: Good catch, will split into two separate warnings.

pkg/appolly/meta/meta_node_hostname.go
mariomac: But if we reach this hostname fetcher, is because the rest of the fetchers in the queue have failed, so there is no data to conflict.

pkg/appolly/meta/meta_node_hostname.go
```diff
@@ -0,0 +1,41 @@
...
+	if fullHostName == "" {
+		slog.Warn("resolved empty hostname for fallback attribute",
+			"component", "meta.hostnameFetcher",
+			"attribute", VendorHostnameAttr)
+		return NodeMeta{}, nil
+	}
+	return NodeMeta{
+		Metadata: []Entry{
+			{
+				Key:   attr.Name(VendorHostnameAttr),
```
MrAlias, line 36: This lets `VendorHostnameAttr` flow into `NodeMeta.Metadata` without checking whether the resulting Prometheus label collides with OBI's fixed labels. In Prometheus, `attr.Name.Prom()` maps dots to underscores, so a vendor value like `host.id`, `service.name`, or `instance` becomes `host_id`, `service_name`, or `instance`, all of which are already present on `target_info`. That produces a duplicate variable label descriptor, and OBI registers those collectors with `MustRegister`, so the Prometheus exporter can panic during startup.

Could we validate/reject colliding keys here, or deduplicate/filter them before building the Prometheus label list?

MrAlias (commented):
I found one exporter-side edge case with the configurable attribute key. The hostname fetcher itself looks small, but because the key is vendor-defined it needs a guard before it flows into Prometheus label construction.

pkg/appolly/meta/meta_node_hostname.go
jereenv: By conflict i meant storing Host name value in same field as machine-ids and cloud instance ids make it impossble for consumer to distinguish what type of value they are reading. Using a separate attribute keeps it clearer

pkg/appolly/meta/meta_node_hostname.go
jereenv: Thanks for highlighting it. To resolve it, I've updated `labelNamesTargetInfo` and `labelValuesTargetInfo` to drop metadata entry whose Prom name already exists in the hardcoded label set.

pkg/appolly/meta/meta_node_hostname.go
mariomac: Sorry, then I totally misunderstood the aim of this hostname Fetcher. Then instead of add a fallback for the `.host.id` fetcher that actually fetches another different field, why not using the HostID and Instance fields that are already decorated in the traces?

The `*svc.Attrs` field of each `request.Span` is already decorated in `read_decorator.go:91`:

```
		s.UID.Instance = hi.ComposeInstance(hostPID)
		s.HostName = hi.HostName
```

This Decorator instance is invoked for each trace in `read_decorator.go:46`.

pkg/appolly/meta/meta_node.go
mariomac: Then in this case I'd, either:

1. Use the `attr.VendorPrefix` global constant in `pkg/export/attributes/names/attrs.go` to define the name into a private variable. Something like: `var fallbackHostIDAttr = attr.VendorPrefix + ".host.id".

2. If you still prefer having a global variable with its unique name, move it to `pkg/export/attributes/names/attrs.go` so we don't spread all these variables throught the codebase.

MrAlias:
Stepping back, I don't think this vendor-specific hostname path belongs in OBI.

OBI already resolves the configured hostname and exports it as `host.name`. This change performs a second lookup that ignores the existing hostname configuration, runs even when OBI has a valid machine or cloud host ID, and injects the result under a process-global, vendor-selected attribute across exporters. That is not fallback behavior, and it couples core metadata collection to a vendor-specific schema.

Please do not continue iterating on this implementation. The vendor integration should copy `host.name` to its proprietary attribute in its Collector/Alloy pipeline. If downstream processing cannot meet the requirement, the next step is to close this PR and open an issue that documents the exact attribute and signals involved, why `host.name` is insufficient, and why the mapping cannot happen downstream. We can evaluate a generic, instance-scoped extension from that requirement.

MrAlias:
@mariomac @rafaelroquetto, do you see a vendor-independent OBI requirement here that cannot be handled by mapping the existing `host.name` downstream?

PR #2387, opened 2026-06-15
Otel col shutdown stuck

florianl:
## Summary

Fix a stuck OTel collector shutdown caused by `bgListenSharedContextCancelation` closing the eBPF ring buffer reader too late, and add a deadline-respecting select in `Controller.Shutdown` so the collector is never blocked if OBI exceeds its internal shutdown timeout.

## Validation

- [ ] ~~If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed.~~


pkg/ebpf/common/ringbuf.go
```diff
@@ -396,8 +396,12 @@ func (rbf *ringBufForwarder[T]) bgListenContextCancelation(ctx context.Context,
...
 	<-ctx.Done()
 	rbf.logger.Debug("context is cancelled. Closing eBPF resources", "len", len(closers))
-	// Often there are hundreds of closers, and don't have time to sequentially close within the
-	// shutdown grace period. Closing them in parallel
+	// Close the events reader before the eBPF objects so the readerLoop unblocks
+	// immediately via ErrClosed, rather than waiting for potentially hundreds of
+	// eBPF closers to finish. This trades a small window of data loss (events
+	// already in the ring buffer but not yet consumed) for a prompt shutdown.
+	rbf.logger.Debug("closing events reader")
+	_ = eventsReader.Close()
```
MrAlias, line 410: This unblocks `sharedReadAndForward` before the eBPF closers below have finished. In the collector shutdown path, that lets the tracer goroutine return, `Instrumenter.stop()` complete, and `Controller.Shutdown` remove the shared controller while the old links/maps/objects may still be closing in this background goroutine.

I think we should preserve the early reader close, but keep the shared tracer path from reporting completion until the closer goroutine is done. Otherwise a fast collector reload can start a new OBI instance while the previous eBPF resources are still attached.

collector/internal/controller_test.go
```diff
@@ -63,6 +63,50 @@ func TestShutdownAfterStartFailureCleansSharedController(t *testing.T) {
...
+	go func() {
+		shutdownDone <- c.Shutdown(ctx)
+	}()
+
+	timer := newShutdownTimer(t)
+	defer stopTestTimer(timer)
+
+	select {
+	case err := <-shutdownDone:
+		if err != context.Canceled {
```
MrAlias: This direct comparison is currently failing Linux lint under `errorlint`; wrapped context errors need to be checked with `errors.Is`.

```suggestion
		if !errors.Is(err, context.Canceled) {
```

This also needs the corresponding `errors` import.

MrAlias (commented):
Thanks for digging into this. The changes look like they are on the right path, but I think there are a couple of details we should address before this is ready.

pkg/ebpf/common/ringbuf.go
florianl: I refactored `sharedReadAndForward()` to address your feedback. It now blocks until the Go routines are finished. Hope this works for you.

mmat11 (approved):

collector/internal/controller.go
```diff
@@ -155,8 +155,15 @@ func (c *Controller) Shutdown(_ context.Context) error {
 		return nil
 	}
 
-	// Wait for OBI to finish
-	<-runDone
+	// Wait for OBI to finish, but respect the caller's shutdown deadline.
+	// If the context expires before OBI is done, return immediately so the
+	// collector is not stuck; OBI will continue cleaning up in the background.
```
rafaelroquetto: nit: I'd do without this comment, I feel the code that follows to be less confusing than the comment itself

florianl: Removed the comment with https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2387/changes/2a52ec65fe90ac046f0e670d55d1ed675e35ea89.

grcevski (approved):
LGTM!

mariomac (approved):
Fine with me too!

PR #2394, opened 2026-06-16
GenAI: support SSE streaming responses and raise HTTP capture limit to 256KB

### LINKED ISSUE #2267: GenAI: support streaming (SSE) responses and large response bodies (>64KB) in HTTP body capture
### Component

OBI (eBPF instrumentation)

### Problem

OBI's GenAI HTTP body capture currently has two structural limitations that prevent reliable observability for real-world LLM and embedding workloads.

Related to #1854 (AI Agent Observability proposal). That issue lists "HTTP body capture buffer limits" as a known risk; this issue tracks the concrete root causes and possible solutions so they can be addressed independently.

#### Limitation 1 — Streaming (SSE) responses are not reliably captured

OpenAI, Gemini, and other LLM providers serve chat/completion responses as Server-Sent Events with `Transfer-Encoding: chunked` and `Content-Type: text/event-stream`. The current pipeline cannot reliably capture and parse these responses for the following reasons:

1. **eBPF capture model is per-syscall.** A single `tcp_recvmsg` invocation only sees the bytes received in that call. SSE events are spread across many `recv` calls, so the eBPF layer never observes the full streamed response.
2. **Userspace parser assumes a complete response.** `httpSafeParseResponse()` in `pkg/ebpf/common/http_transform.go` calls the standard library's `http.ReadResponse()`, which expects a complete HTTP response body. Truncated streams surface as `io.ErrUnexpectedEOF`.
3. **SSE parsing assumes the full event sequence is present.** `parseAnthropicStream()` in `pkg/ebpf/common/http/anthropic.go` already handles SSE for Anthropic, but it assumes all events are available. If eBPF only delivered a partial slice of the stream, parsing fails or yields incomplete results.
4. **OpenAI has no streaming support at all.** `pkg/ebpf/common/http/openai.go` does not detect or parse SSE responses, so streaming chat completions are invisible.

#### Limitation 2 — Response bodies larger than 64 KiB are truncated

The capture pipeline enforces a 64 KiB hard cap that is aligned across three layers:

- eBPF: `k_large_buf_max_http_captured_bytes = 1 << 16` (`bpf/common/large_buffers.h`)
- Config: `MaxCapturedPayloadBytes = 1 << 16` with `validate:"lte=65536"` (`pkg/config/ebpf_tracer.go:181`)
- Userspace: `maxDecompressedResponseBodyBytes = config.MaxCapturedPayloadBytes` (`pkg/ebpf/common/http/responses.go:27`)

Impact on GenAI workloads:

- Long chat/completion responses (typically 10–100 KiB) may be truncated.
- Embeddings responses (often 50–500 KiB for batched inputs or high-dimensional vectors) routinely exceed the cap and cannot be captured at all.
- Aggregated streaming output (10–200 KiB across all SSE events) typically exceeds the cap even when streaming reassembly succeeds.

When the body is truncated, downstream parsers see invalid JSON and fail to extract `model`, `usage.input_tokens`, `usage.output_tokens`, and other key GenAI attributes.

NameHaibinZhang:
## Summary

Fixes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2267

This PR adds SSE (Server-Sent Events) streaming response support for GenAI spans and raises the HTTP payload capture limit to 256KB (accumulated across multiple recv calls).

### Changes

**SSE streaming support:**
- Add `parseOpenAIStream()` to detect and parse OpenAI/Qwen SSE streaming responses, accumulating content deltas, tool calls, and usage statistics across multiple chunks.
- Improve Anthropic JSON vs SSE classification using `looksLikeJSON()` instead of a naive first-byte check, handling responses with leading whitespace.
- Add Qwen support for OpenAI-compatible endpoints (e.g. vLLM deployments) with SSE streaming.

**Large response body support:**
- Raise `MaxCapturedPayloadBytes` from 64KB to 256KB for HTTP. Other protocols (MySQL, Kafka, Postgres, etc.) remain at 64KB.
- The BPF layer per-syscall limit (`k_large_buf_max_http_captured_bytes = 64KB`) is unchanged — the 256KB connection-level limit is achieved by accumulating across multiple recv calls (up to 4 × 64KB chunks).

**Truncated body resilience:**
- Add `extractJSONRawField()` to recover complete JSON fields (e.g. `messages`) from truncated request bodies, improving GenAI attribute extraction when payloads exceed capture limits.

## Validation


pkg/ebpf/common/http/openai_stream.go
```diff
@@ -0,0 +1,191 @@
...
+	Function struct {
+		Name      string `json:"name"`
+		Arguments string `json:"arguments"`
+	} `json:"function"`
+}
+
+// parseOpenAIStream parses the SSE stream from OpenAI-compatible APIs (including Qwen/DashScope)
+// and returns the aggregated response with usage statistics and tool calls.
+func parseOpenAIStream(reader io.Reader) (*request.VendorOpenAI, []request.ToolCall) {
+	scanner := bufio.NewScanner(reader)
```
mariomac, line 55: Not sure if I'm just imagining a theoretical scenario that will never happen or it is possible in the internal HTTP implementation of LLMs.

Usually, HTTP lines are maxed out to 8-16 KB. However, if a given server configuration allows HTTP lines larger than 64KB, later `scanner.Text()` result can be cut down to `bufio.MaxScanTokenSize`, which is 64KB. Returning `ErrTooLarge` on larger buffers.

Not sure if I'm just theorizing or just in case you should add a `scanner.Buffer` invocation to max out the scan buffer to 256KB.

pkg/ebpf/common/http_transform.go
```diff
@@ -318,9 +318,49 @@ func HTTPInfoEventToSpan(parseCtx *EBPFParseContext, event *BPFHTTPInfo) (reques
...
+	headerEnd := bytes.Index(raw, []byte("\r\n\r\n"))
+	if headerEnd < 0 {
+		return nil
+	}
+	bodyStart := headerEnd + 4
+
+	// Scan forward from header end to find a JSON object.
+	for i := bodyStart; i < len(raw); i++ {
+		if raw[i] == '{' {
+			return raw[i:]
```
mariomac: Should we bound this to `raw[i:<ContentLength>]` to avoid going beyond the limits of the request?

pkg/ebpf/common/http_transform.go
```diff
@@ -318,9 +318,49 @@ func HTTPInfoEventToSpan(parseCtx *EBPFParseContext, event *BPFHTTPInfo) (reques
...
+	// Find end of HTTP headers.
+	headerEnd := bytes.Index(raw, []byte("\r\n\r\n"))
+	if headerEnd < 0 {
+		return nil
+	}
+	bodyStart := headerEnd + 4
+
+	// Scan forward from header end to find a JSON object.
+	for i := bodyStart; i < len(raw); i++ {
+		if raw[i] == '{' {
```
mariomac: Would this detect as JSON any text containing a stray `{`?

If this is an acceptable error we can later recover from, should we also detect `[` for JSON arrays?

pkg/ebpf/common/http/openai_stream.go
```diff
@@ -0,0 +1,191 @@
...
+				}
+
+				if tc.ID != "" {
+					accumulators[idx].id = tc.ID
+				}
+				if tc.Function.Name != "" {
+					accumulators[idx].name = tc.Function.Name
+				}
+				if tc.Function.Arguments != "" {
+					accumulators[idx].args.WriteString(tc.Function.Arguments)
```
mariomac: This `args` string Builder is never converted to a string. Did you forget to make use of its value or should we remove it?

mariomac:
In `large_buffer.h`, there is a bunch of definitions like:
```
    k_large_buf_max_http_captured_bytes = 1 << 16,
    k_large_buf_max_mysql_captured_bytes = 1 << 16,
    k_large_buf_max_postgres_captured_bytes = 1 << 16,
    k_large_buf_max_kafka_captured_bytes = 1 << 16,
    k_large_buf_max_mssql_captured_bytes = 1 << 16,
    k_large_buf_max_tcp_captured_bytes = 1 << 16,
```

Do we need to update `k_large_buf_max_http_captured_bytes` at least to fit `256Kb`?

mariomac (commented):
Great addition! Added few suggestions

bpf/common/large_buffers.h
```diff
@@ -24,6 +24,11 @@ enum {
     // Maximum valid value for each protocol's *_max_captured_bytes volatile variable.
     // These must equal the lte= validation values in EBPFBufferSizes (pkg/config/ebpf_tracer.go),
     // which enforces the same ceiling at configuration time.
+    // Per-syscall emit cap for HTTP. Keeps the large_buf_emit_chunks() loop bounded
+    // (64KB / 16KB = 4 iterations) so the verifier stays within the 1M insn budget.
+    // The connection-level total is controlled by the volatile const http_max_captured_bytes,
+    // which may be larger (up to the lte= ceiling in EBPFBufferSizes) and is accumulated
+    // across multiple tcp_recvmsg invocations via bytes_sent.
```
rafaelroquetto: Please remove this comment, it's just a maintenance burden

bpf/common/large_buffers.h
```diff
@@ -24,6 +24,11 @@ enum {
     // Maximum valid value for each protocol's *_max_captured_bytes volatile variable.
     // These must equal the lte= validation values in EBPFBufferSizes (pkg/config/ebpf_tracer.go),
     // which enforces the same ceiling at configuration time.
+    // Per-syscall emit cap for HTTP. Keeps the large_buf_emit_chunks() loop bounded
```
rafaelroquetto: I'd revert this, as this gives the impression that HTTP is special, when it's not - all protocols have the notion of captured bytes as stated by the globals at the top of this file.

These comments (including the original one above your additions) are a maintenance burden. This PR is the prime example of watching then drift. Instead, I'd actually delete them all.

devdocs/config/CONFIG.md
```diff
@@ -181,7 +181,7 @@ EBPFTracer configuration for eBPF programs
 
 ### `ebpf.buffer_sizes`
 
-Per-protocol maximum bytes to capture per request per direction, sent to userspace via large buffer events. Values must stay aligned with MaxCapturedPayloadBytes and the k_large_buf_max_*_captured_bytes constants in bpf/common/large_buffers.h.  Default: 0 (disabled).
+Per-protocol maximum bytes to capture per request per direction, sent to userspace via large buffer events. HTTP supports up to MaxCapturedPayloadBytes (256KB) accumulated across multiple recv calls. Other protocols are limited to their respective k_large_buf_max_*_captured_bytes (64KB).  Default: 0 (disabled).
```
rafaelroquetto: Please revert. The only thing that changes here is the default value for HTTP, but we don't need to be explicit about it.

devdocs/features.md
```diff
@@ -66,14 +66,16 @@ Large payloads are streamed to userspace across multiple ring-buffer events and
...
-| `OTEL_EBPF_BPF_BUFFER_SIZE_HTTP`   | HTTP       | 65535   | 0 (disabled) |
+| `OTEL_EBPF_BPF_BUFFER_SIZE_HTTP`   | HTTP       | 262144[^1] | 0 (disabled) |
 | `OTEL_EBPF_BPF_BUFFER_SIZE_MYSQL`  | MySQL      | 65535   | 0 (disabled) |
 | `OTEL_EBPF_BPF_BUFFER_SIZE_KAFKA`  | Kafka      | 65535   | 0 (disabled) |
 | `OTEL_EBPF_BPF_BUFFER_SIZE_POSTGRES` | PostgreSQL | 65535 | 0 (disabled) |
 | `OTEL_EBPF_BPF_BUFFER_SIZE_MSSQL`  | MSSQL      | 65535   | 0 (disabled) |
 
 Equivalent YAML keys live under `ebpf.buffer_sizes.{http,mysql,kafka,postgres,mssql}`.
 
+[^1]: HTTP payload may be accumulated across multiple `recv` calls (up to 4 × 64KB) before being forwarded to userspace.
```
rafaelroquetto: This is implementation detail, please remove. It is possible for instance that HTTP payloads may be accumulated elsewhere.

The important thing for documentation is to convey the semantics of these variables, namely, that these are  the maximum amount of bytes captures per request and per direction (per protocol).

pkg/config/ebpf_tracer.go
```diff
@@ -175,17 +175,20 @@ func (e *EBPFTracer) CudaInstrumentationEnabled() bool {
...
 // per protocol request direction via large buffer events.
 //
-// It must stay aligned with the k_large_buf_max_*_captured_bytes constants in
-// bpf/common/large_buffers.h and with the validate tags in EBPFBufferSizes.
-const MaxCapturedPayloadBytes = 1 << 16
+// For HTTP, this is the connection-level total limit accumulated across multiple
+// syscalls. The BPF-side per-syscall emission cap (k_large_buf_max_http_captured_bytes)
+// may be smaller to keep verifier complexity bounded.
+// For other protocols, this must stay aligned with k_large_buf_max_*_captured_bytes
+// in bpf/common/large_buffers.h.
```
rafaelroquetto: Same here - please delete this comment - for instance, if someone bumps the Mysql values to be 256K, it will also span multiple system calls (and sometimes, it will span multuple syscall regardless, MySQL already does that because it calls send() twice for the same command, the first send() is the header and the second is the payload.)
```suggestion

```

pkg/config/ebpf_tracer.go
```diff
@@ -175,17 +175,20 @@ func (e *EBPFTracer) CudaInstrumentationEnabled() bool {
...
+const MaxCapturedPayloadBytes = 1 << 18 // 262144
```
```suggestion
const MaxCapturedPayloadBytes = 1 << 18
```

rationale: this is reistating the obvious and maintenance burden / error prone.


To be honest, this kind of highlighted an issue with this constant - there's no requirement to keep it in sync with ebpf at all - the limits governing ebpf are the ones in `large_buffers.h`. All this is one `MaxCapturedPayloadBytes` is doing is setting a decompression budget in responses.go - so we should probably just set it there, with a sane value

pkg/config/ebpf_tracer.go
```diff
@@ -175,17 +175,20 @@ func (e *EBPFTracer) CudaInstrumentationEnabled() bool {
...
 
 // Per-protocol maximum bytes to capture per request per direction, sent to userspace via large buffer events.
-// Values must stay aligned with MaxCapturedPayloadBytes and the
-// k_large_buf_max_*_captured_bytes constants in bpf/common/large_buffers.h.
+// HTTP supports up to MaxCapturedPayloadBytes (256KB) accumulated across multiple recv calls.
```
rafaelroquetto: ditto

pkg/ebpf/common/http/openai.go
```diff
@@ -77,7 +78,18 @@ func OpenAISpan(baseSpan *request.Span, req *http.Request, resp *http.Response)
...
+	var parsedResponse request.VendorOpenAI
+	var toolCalls []request.ToolCall
+
+	if looksLikeJSON(respB) {
+		parsedResponse = parseVendorOpenAI(respB)
+		toolCalls = extractToolCalls(parsedResponse.Choices)
+	} else {
+		reader := bytes.NewReader(respB)
+		streamResponse, tc := parseOpenAIStream(reader)
+		parsedResponse = *streamResponse
```
rafaelroquetto: we shouldn't copy the entire struct here, can't we reuse the pointer?

pkg/ebpf/common/tcp_large_buffer.go
```diff
@@ -56,6 +56,12 @@ func appendTCPLargeBuffer(parseCtx *EBPFParseContext, record *ringbuf.Record) (r
 
 	chunk := record.RawSample[hdrSize : hdrSize+event.Len]
 
+	// Drop TLS ciphertext that leaked from kprobes into HTTP large buffers.
```
rafaelroquetto: This sounds unrelated and smells like a workaround. I'd remove from this PR, and open another one with a proper fix at the ebpf level if feasible

pkg/ebpf/common/http/openai_stream.go
rafaelroquetto, line 55: 🤖 **Claude:** `bufio.NewScanner` defaults to a 64KB max token size (`bufio.MaxScanTokenSize`). A single SSE `data:` line larger than 64KB makes `Scan()` stop with `bufio.ErrTooLong`, and the rest of the stream is silently dropped — and `scanner.Err()` is never checked here. This directly undercuts this PR's own 256KB goal for large streamed responses. Please set a larger buffer with `scanner.Buffer(...)` and check `scanner.Err()` after the loop.

pkg/ebpf/common/http/openai_stream.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+	}
+
+	// Build the final tool calls list.
+	var toolCalls []request.ToolCall
+	for i := range accumulators {
+		if accumulators[i].name == "" {
+			continue
+		}
+		toolCalls = append(toolCalls, request.ToolCall{
```
rafaelroquetto, line 181: 🤖 **Claude:** dead accumulation. `accumulators[i].args` is written on every chunk (line ~142) but never read — the emitted `request.ToolCall` only carries `ID`/`Name` (the struct has no `Arguments` field). So the per-chunk argument concatenation is wasted work. Either drop the `args` builder, or, if capturing arguments is intended, add the field and wire it through — and do the same for the non-stream path for consistency.

pkg/ebpf/common/http_transform.go
rafaelroquetto: 🤖 **Claude:** this returns a slice into `buf.UnsafeView()`, which aliases buffer memory whose lifetime isn't guaranteed past the parse. The returned bytes are retained in `req.Body` (via `bytes.NewBuffer`) and may outlive the buffer, risking reads of recycled/overwritten memory. Copy the bytes before returning if the body must survive. Separately: this and the TLS-drop heuristic in `tcp_large_buffer.go` are both userspace band-aids for SSL-connection capture issues — the principled fix belongs in the BPF layer, in its own PR (see summary #1).

pkg/ebpf/common/http/openai.go
rafaelroquetto: 🤖 **Claude:** this JSON-vs-SSE branch (incl. the `*streamResponse` copy) is duplicated near-verbatim in `qwen.go` (~L70). Please extract a shared helper, e.g. `parseOpenAICompatibleResponse(respB) (request.VendorOpenAI, []request.ToolCall)`, and call it from both. Minor: making `parsedResponse` a `*request.VendorOpenAI` avoids the full-struct copy here since `parseOpenAIStream` already returns a pointer and `GenAI.OpenAI` is a pointer.

rafaelroquetto (commented):
> 🤖 **Review by Claude (Claude Code), posted by @rafaelroquetto on its behalf.** This is an automated, ownership-focused review aligned with `AGENTS.md` and the repo's PR-review guidance — not Rafael's personal review (he has left separate comments inline).

The SSE parsing and the 256KB cross-syscall accumulation are sound at their core. My main concern is **scope and ownership**, per `AGENTS.md` ("Keep changes minimal and scoped to the task. Do not include unrelated edits"). Concrete code-level bugs are left as inline comments; the structural points are below.

## 1. Scope creep — the PR does much more than its title (please split)

Title: *"support SSE streaming responses and raise HTTP capture limit to 256KB"*, but the diff also bundles features the description never mentions:

- **Rerank provider-format support** — `RerankRequest.NestedInput/NestedParams`, `GetQuery/GetDocuments/GetTopN`, `RerankResponse.NestedOutput`, `GetResults/GetID`, `VendorRerank.GetInput/GetOutput`, `gen_ai.rerank.top_n` (`span.go`, `tracesgen.go`).
- **Retrieval top-k** — `RetrievalRequest.TopK/TopKSnake/Limit`, `GetTopK`, `gen_ai.retrieval.top_k`.
- **Embedding dimensions** — `GetEmbeddingDimensions()` + a duplicated emit block in `tracesgen.go`.
- **MCP tool-call argument/result capture** — `mcp.go`, `span.go`, `attrs.go`, `tracesgen.go`.
- **Two SSL-capture band-aids** — `recoverJSONBodyFromBuffer` and the TLS-ciphertext drop, both in the final `merge llm` commit.

None of rerank/retrieval/embedding/SSL is needed for SSE + 256KB. Each is reasonable on its own, but bundled they are hard to review and to revert independently. Please split into focused PRs, each explaining *why* the approach fits this codebase.

## 2. Duplication instead of reuse

The JSON-vs-SSE response branch is copy-pasted near-verbatim in `openai.go` and `qwen.go` (and Anthropic now also uses `looksLikeJSON`). `AGENTS.md`: *"Extend or adapt existing code instead of duplicating functionality."* Please extract one helper, e.g. `parseOpenAICompatibleResponse(respB) (request.VendorOpenAI, []request.ToolCall)`. (Inline note on the openai.go site.)

## 3. Smaller items

- `large_buffers.h`: the pre-existing *"must equal the lte= values"* comment is now false for HTTP (64KB const vs 256KB ceiling), and the appended paragraph frames cross-syscall accumulation as HTTP-specific — every protocol accumulates identically; only HTTP's configured ceiling differs. A one-liner would be clearer.
- `MaxCapturedPayloadBytes` is now HTTP-only in meaning but keeps a generic name and is consumed solely as the decompression budget in `responses.go`; the 64KB→256KB bump silently 4×'s that budget. Consider a dedicated decompression constant.
- `extractJSONRawField` finds the field via a plain `bytes.Index("\"messages\"")`, which can match inside a string value or nested key rather than the documented top-level field.
- Commit messages (`fix test`, `fix attr`, `merge llm`, `fix stream`) don't convey intent; `merge llm` bundles the two unrelated SSL changes.

## Asks
1. Split the unrelated features (rerank/retrieval/embedding/SSL) into separate PRs.
2. Deduplicate the OpenAI-compatible response branch.
3. Fix the inline-flagged bugs (scanner cap, dead `args`, `UnsafeView` lifetime).
4. Confirm local validation (`make verify`, `make test`) on the trimmed change.

pkg/ebpf/common/http/openai_stream.go
NameHaibinZhang: Fixed. Added `scanner.Buffer(make([]byte, 0, 256*1024), 256*1024)` to raise the scan limit to 256KB, and added `scanner.Err()` check after the loop.

pkg/ebpf/common/http_transform.go
NameHaibinZhang: Fixed. The caller now bounds the recovered slice to `req.ContentLength` before using it, and copies the bytes to avoid aliasing `UnsafeView` memory.

pkg/ebpf/common/http_transform.go
NameHaibinZhang: Fixed. Added `'['` detection alongside `'{'` in `recoverJSONBodyFromBuffer` for JSON arrays. A stray `{` or `[` after HTTP headers is unlikely in practice — `json.Unmarshal` downstream will reject truly malformed data, so this is an acceptable heuristic.

pkg/ebpf/common/http/openai_stream.go
NameHaibinZhang: Fixed. Removed the `args` builder — the `ToolCall` struct only carries `ID` and `Name`, so argument accumulation was dead code.

bpf/common/large_buffers.h
NameHaibinZhang: Done, removed.

bpf/common/large_buffers.h
NameHaibinZhang: Agreed. Removed all comments from the enum block — the constant names are self-documenting.

devdocs/config/CONFIG.md
NameHaibinZhang: Reverted.

devdocs/features.md
NameHaibinZhang: Removed the footnote. Kept the updated maximum value (262144) in the table since it documents the user-facing limit without implementation detail.

pkg/config/ebpf_tracer.go
NameHaibinZhang: Done, removed the implementation-detail comment.

pkg/config/ebpf_tracer.go
NameHaibinZhang: Applied. Simplified to just the constant. Also removed the old alignment comment since, as you noted, there's no requirement for this constant to stay in sync with the eBPF-side limits.

pkg/config/ebpf_tracer.go
NameHaibinZhang: Done — reverted struct comment to just the semantic description without protocol-specific details.

pkg/ebpf/common/http/openai.go
NameHaibinZhang: Fixed. Changed `parsedResponse` to `*request.VendorOpenAI` so the stream path reuses the pointer directly without copying.

pkg/ebpf/common/tcp_large_buffer.go
NameHaibinZhang: Agreed — removed the TLS ciphertext filter from this PR. Will open a separate PR with a proper fix at the eBPF level.

pkg/ebpf/common/http/openai_stream.go
NameHaibinZhang: Fixed — see the reply to the same concern raised by @mariomac above. Added `scanner.Buffer(...)` with 256KB limit and `scanner.Err()` check.

pkg/ebpf/common/http/openai_stream.go
NameHaibinZhang: Fixed — removed the dead `args` builder. See the reply to @mariomac's identical finding above.

pkg/ebpf/common/http_transform.go
NameHaibinZhang: Fixed. The caller now copies the recovered bytes with `append([]byte(nil), recovered...)` to avoid retaining aliased `UnsafeView` memory. Also bounded to `ContentLength` per @mariomac's suggestion. The TLS-drop heuristic has been removed from this PR per Rafael's request.

pkg/ebpf/common/http/openai.go
NameHaibinZhang: The pointer copy issue is fixed (see reply above). Regarding the shared helper extraction: the two paths have vendor-specific differences — Qwen has additional `request_id` extraction from the response body, header-based ID fallback, and different operation name logic — making a clean shared helper non-trivial. I'll address this refactoring in a follow-up PR to keep the scope of this one focused.

NameHaibinZhang:
No — `k_large_buf_max_http_captured_bytes` is intentionally kept at 64 KB because it is the **per-syscall emission cap**, not the connection-level limit. The 256 KB budget is reached by accumulating across multiple `tcp_recvmsg` calls: each call emits up to 64 KB (4 × 16 KB ring-buffer events), and the running totals `lb_req_bytes` / `lb_res_bytes` track consumption until `http_max_captured_bytes` (256 KB) is exhausted. Keeping the per-call cap at 64 KB bounds the `large_buf_emit_chunks()` loop to 4 iterations, which is important for verifier complexity. @mariomac

NameHaibinZhang:
## PR Split Summary

Per @rafaelroquetto's review feedback on scope and ownership, this PR has been split into focused, independently reviewable changes:

| # | PR | Scope |
|---|---|---|
| 1 | #2394 (this PR, reset) | **Core**: SSE streaming response parsing + 256KB HTTP capture limit |
| 2 | #2407 | JSON body recovery from truncated TLS capture buffer |
| 3 | #2408 | MCP tool-call arguments and results capture |
| 4 | #2409 | Retrieval top_k extraction (multi-vendor) |
| 5 | #2410 | Rerank nested request/response format support |
| 6 | #2411 | Embedding dimensions derivation from response data |

All PRs relate to #2267. Each is self-contained and can be reviewed/merged independently.

pkg/ebpf/common/http/openai_stream.go
NameHaibinZhang: Thank you for the thorough review. All points addressed:

1. **Scope split** — Done. This PR has been reset to contain only the SSE streaming parser + 256KB buffer limit. The unrelated features are now in separate PRs: #2407 (JSON recovery), #2408 (MCP tool-args), #2409 (retrieval top_k), #2410 (rerank format), #2411 (embedding dimensions).

2. **Deduplication** — Extracted `parseOpenAICompatibleResponse()` as a shared helper in `openai.go`, called by both `openai.go` and `qwen.go`.

3. **Inline bugs** — Fixed in previous commits (scanner buffer cap, dead args builder, UnsafeView lifetime).

4. **Local validation** — Verified on the trimmed change.

mariomac (approved):
Thanks for addressing! Happy with the changes related to this PR (will check the others later). Will wait for Rafael's approval before merging.

bpf/common/large_buffers.h
```diff
@@ -13,17 +13,8 @@ volatile const u32 mssql_max_captured_bytes = 0;
 volatile const u32 tcp_max_captured_bytes = 0;
 
 enum {
-    // Maximum payload size per ring buffer chunk.
```
mmat11, line 16: any reason these comments have been removed?

pkg/config/ebpf_tracer.go
```diff
@@ -174,18 +174,13 @@ func (e *EBPFTracer) CudaInstrumentationEnabled() bool {
 
 // MaxCapturedPayloadBytes is the maximum number of bytes that can be captured
 // per protocol request direction via large buffer events.
-//
-// It must stay aligned with the k_large_buf_max_*_captured_bytes constants in
-// bpf/common/large_buffers.h and with the validate tags in EBPFBufferSizes.
-const MaxCapturedPayloadBytes = 1 << 16
+const MaxCapturedPayloadBytes = 1 << 18
```
rafaelroquetto: I think we missed this one - what I meant was that this constant needs to go - it's only ever used inside `responses.go` - it's private there and where it belongs

rafaelroquetto (changes requested):

rafaelroquetto (changes requested):
This looks much better, many thanks for all of this. I have one comment left regarding MaxCapturedPayloadBytes, once we address that we can merge this

pkg/config/ebpf_tracer.go
NameHaibinZhang: Done! Moved `MaxCapturedPayloadBytes` into `responses.go` as a private `maxCapturedPayloadBytes` constant, since that's its only consumer.

NameHaibinZhang:
@rafaelroquetto done

bpf/common/large_buffers.h
```diff
@@ -13,17 +13,8 @@ volatile const u32 mssql_max_captured_bytes = 0;
...
-    // Rounded up to the next power of 2 above k_large_buf_payload_max_size to account
-    // for the struct overhead.
-    k_large_buf_max_size = 1 << 15, // 32K
-
-    // Maximum valid value for each protocol's *_max_captured_bytes volatile variable.
-    // These must equal the lte= validation values in EBPFBufferSizes (pkg/config/ebpf_tracer.go),
-    // which enforces the same ceiling at configuration time.
+    k_large_buf_payload_max_size = 1 << 14,
+    k_large_buf_max_size = 1 << 15,
     k_large_buf_max_http_captured_bytes = 1 << 16,
```
mmat11: this is gated to 64KB, which means 

```
static __always_inline int http_send_large_buffer(http_info_t *req,
                                                  const void *u_buf,
                                                  u32 bytes_len,
                                                  u8 packet_type,
                                                  u8 direction,
                                                  enum large_buf_action action) {
    if (http_max_captured_bytes > k_large_buf_max_http_captured_bytes) {
        bpf_dbg_printk("BUG: http_max_captured_bytes exceeds maximum allowed value.");
    }
```

will fail if the value is higher than that?

mmat11 (commented):
has this been tested manually? can we also add an integration test which exercise the "super large buffer" path?

NameHaibinZhang:
Yes, this has been tested manually — I verified SSE streaming responses with payloads exceeding 64KB (up to ~200KB) and confirmed the large buffer path correctly accumulates chunks across multiple `tcp_recvmsg` events.

Regarding integration tests: agreed, this would be valuable. However, writing a proper integration test for the "super large buffer" path requires setting up an SSE/streaming server that produces >64KB responses and verifying the assembled payload in userspace — which is a non-trivial addition to the test infra. I'll track this as a follow-up. Would that be acceptable, or would you prefer it gated in this PR? @mmat11

bpf/common/large_buffers.h
NameHaibinZhang: These were removed at @rafaelroquetto's request in his first-round review — he considered them implementation details and a maintenance burden (see his [comment](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2394#discussion_r3424228752): *"Please remove this comment, it's just a maintenance burden"* and [this one](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2394#discussion_r3424286679) noting they give the wrong impression that HTTP is special). Happy to restore them if you two reach a consensus, though.

bpf/common/large_buffers.h
NameHaibinZhang: Good question — it won't cause a functional failure. Here's the layered design:

- **`http_max_captured_bytes`** (set from userspace, up to 256KB) is the **total per-request per-direction budget**. It controls when to stop accumulating data (`bytes_sent >= http_max_captured_bytes`).
- **`k_large_buf_max_http_captured_bytes`** (64KB) is the **per-syscall-event cap**, used in `bpf_clamp_umax(max_available_bytes, k_large_buf_max_http_captured_bytes)` to bound each ring buffer submission. This is critical for the BPF verifier to prove memory safety.
- The 256KB total budget is reached by accumulating multiple 64KB chunks across successive `tcp_recvmsg` events.

The `bpf_dbg_printk("BUG: ...")` check is a debug-only assertion (gated by `g_bpf_debug`, compiled out in production) that predates the HTTP limit raise. It's now stale for HTTP — I can clean it up or update the condition to reflect the new semantic. It does not affect correctness since the code continues past it regardless and the actual safety bound is enforced by `bpf_clamp_umax`.

mmat11: I thought there was a return statement. I think we should clean it up if it's pointless

mmat11:
> Yes, this has been tested manually — I verified SSE streaming responses with payloads exceeding 64KB (up to ~200KB) and confirmed the large buffer path correctly accumulates chunks across multiple `tcp_recvmsg` events.
> 
> Regarding integration tests: agreed, this would be valuable. However, writing a proper integration test for the "super large buffer" path requires setting up an SSE/streaming server that produces >64KB responses and verifying the assembled payload in userspace — which is a non-trivial addition to the test infra. I'll track this as a follow-up. Would that be acceptable, or would you prefer it gated in this PR? @mmat11

follow-up is fine!

bpf/common/large_buffers.h
NameHaibinZhang: Done — removed the stale assertion. It was comparing the per-request budget (256KB) against the per-syscall cap (64KB), which is now expected behavior in the multi-chunk design.

pkg/ebpf/common/http/partial_json.go
```diff
@@ -218,9 +235,125 @@ func parseOpenAIInput(body []byte) request.OpenAIInput {
...
+// extractJSONRawField extracts a top-level JSON array or object value for
+// the given field name using bracket matching. This is a fallback for when
+// json.Unmarshal fails on truncated JSON but the target field's value is
+// complete within the captured bytes.
+//
+// The scan is depth-aware: the key is only matched when we are at the
+// top-level object (depth == 1) and not currently inside a JSON string,
+// which avoids false positives from occurrences nested in strings or
+// inner objects.
+func extractJSONRawField(body []byte, field string) json.RawMessage {
```
mmat11, line 248: this was hard to review manually :D so I let claude review it for me, currently it matches string values as fields too, eg. `extractJSONRawField([]byte(`{"label":"field","field":99}`), "field")`

would this simplified version work too?

```
// extractJSONRawField returns the raw value of a top-level field. It works on
// truncated JSON as long as the target field's value is complete in body.
// Returns nil if the body isn't a JSON object, the field is absent, or its
// value is cut off.
func extractJSONRawField(body []byte, field string) json.RawMessage {
	dec := json.NewDecoder(bytes.NewReader(body))

	// Consume the opening '{'.
	if t, err := dec.Token(); err != nil {
		return nil
	} else if d, ok := t.(json.Delim); !ok || d != '{' {
		return nil
	}

	for dec.More() {
		keyTok, err := dec.Token() // object key
		if err != nil {
			return nil
		}
		var raw json.RawMessage
		if err := dec.Decode(&raw); err != nil { // exactly one value
			return nil
		}
		if key, _ := keyTok.(string); key == field {
			return raw
		}
	}
	return nil
}
```

`extractJSONRawValue` can be removed entirely

bpf/common/large_buffers.h
rafaelroquetto: I don't think this is the right approach. IMHO the statement is correct: it is a bug to set a value that is lager than `k_large_buf_max_http_captured_bytes` because `k_large_buf_max_http_captured_bytes` is used to bound `max_available_bytes` for the verifier. So if `max_available_bytes` is larger than `k_large_buf_max_http_captured_bytes`, we end up truncating the buffer and throwing away the remainder of the payload _silently_.

So IMHO we should at least log that, _or_ try to remove the clamp and see if the verifier is happy.

mmat11: @rafaelroquetto but in this PR we are doing exactly that, we are potentially setting `256KB > 64KB`

rafaelroquetto (commented):
This looks better, thanks for the changes! There are a few failing tests that need to be addressed before approving, and I've reopened one of the discussions regarding k_large_buf_max_http_captured_bytes

pkg/ebpf/common/http/partial_json.go
NameHaibinZhang: Good catch on the false-positive matching! Adopted your `json.NewDecoder` version — it's cleaner and correctly matches only object keys. Also removed `extractJSONRawValue` as suggested.

bpf/common/large_buffers.h
NameHaibinZhang: @mmat11 @rafaelroquetto Thanks for the review! I fix the failing tests.For the k_large_buf_max_http_captured_bytes discussion — you're right that silent truncation isn't ideal. I'm thinking of adding a log/warning when max_available_bytes exceeds the per-syscall cap. Removing the clamp entirely would need verifier testing across kernels, so I'd prefer to tackle that as a follow-up. Will elaborate in the thread.

rafaelroquetto: @mmat11 perhaps I am misunderstanding it but I think we have a problem here: imagine a total buffer size being sent of 256KB, split between 2 syscalls: 128KB + 128KB.

What's going to happen now is that we end up with a final reassembled buffer consisting of the first 64KB of each original syscall, i.e. `|64KB truncated|64KB truncated`.

The previous behaviour ensured that there were no holes possible, the only thing that happened was that we might truncate the final buffer tail, which is fine, as at least we get coherent data in userspace. So I guess we need to solve this before this gets merged, as it's an actual bug.

rafaelroquetto: @NameHaibinZhang I don't think we can ship this as is, this is an actual bug/regression. I agree that dealing with the verifier is not easy, and probably this was one of the reasons why we kept everything bounded at 64KB originally. Either way, there's no point on shipping 256KB to userspace if it's not parseable because it is truncated/corrupted.

mmat11: > @mmat11 perhaps I am misunderstanding it but I think we have a problem here: imagine a total buffer size being sent of 256KB, split between 2 syscalls: 128KB + 128KB.
What's going to happen now is that we end up with a final reassembled buffer consisting of the first 64KB of each original syscall, i.e. |64KB truncated|64KB truncated.

as far as I understood, that's the point of this PR, keep sending as much data as possible and let userspace deal with broken/truncated data; I'm not sure this should be the default behavior though, I also don't like to have another "allow footgun" kind of flag

NameHaibinZhang: @rafaelroquetto You're absolutely right — this was a real bug. If a single `tcp_recvmsg` delivers more than 64KB, BPF truncates the chunk, and blindly appending subsequent truncated chunks would create holes in the reassembled buffer.

Fixed: the userspace reassembly now detects truncation by checking whether the accumulated data for a single emission hits exactly the per-syscall cap (64KB) with a full final chunk (16KB payload). When that pattern is detected, the buffer is "sealed" and subsequent chunks for that direction are discarded — ensuring the assembled data is always a contiguous prefix (tail truncation only, no holes).

Added unit tests covering: truncation detection, non-truncation pass-through, and seal reset on new request.

rafaelroquetto: @NameHaibinZhang thanks for iterating this but I don't think this is the right approach and I'd rather we don't go down the userspace-heuristic path at all.

The userspace simply has no way of knowing whether a buffer was actually truncated. A syscall that got clamped at 64KB and a perfectly healthy one that just happens to land on a 64KB boundary produce exactly the same sequence of chunks, i.e same lengths, same actions, nothing to tell them apart. So len % 64KB == 0 isn't detecting truncation, it's guessing, and it'll guess wrong on legitimate traffic, silently capping perfectly healthy buffers at 64KB, which is the exact regression we're trying to avoid.

The truncation only exists as a fact inside BPF - that's the one place that knows `bytes_len > cap` so that's where it has to be dealt with.

I do think we can do this properly in eBPF with some effort. The verifier makes it fiddly, I know, and that's probably why we kept everything bounded at 64KB in the first place, but I'd much rather we take the time to respect the semantics and get it right than merge a workaround that quietly regresses the common case.

NameHaibinZhang: @rafaelroquetto  You're right — the userspace heuristic can't reliably distinguish truncation from legitimate 64KB-aligned traffic. I'll revert the heuristic and implement this properly in BPF: add a `truncated` flag to the event metadata set when `bytes_len > cap`, so userspace gets an explicit signal to stop accumulation. Will push the BPF-side fix shortly.

rafaelroquetto: @NameHaibinZhang: no, I don't mean it like that - a `truncated` flag is a workaround. We should be able to ship full 256KB from ebpf, the usual semantics. This will require some ebpf work. I'd suggest first trying to simply increasing the constant and see what the verifier says. In the likely event it complains, you will need to chunk the buffers. I'd rather us take some time to get it right than rush it and get it wrong, there's no urgency here.

NameHaibinZhang: @rafaelroquetto  I've already tested raising `k_large_buf_max_http_captured_bytes` to 256KB directly — it fails the BPF verifier on 5.10 kernels. The loop unrolling in `large_buf_emit_chunks` goes from 4 iterations (4×16KB = 64KB) to 16 iterations (16×16KB = 256KB), which exceeds the instruction count limit on older kernels.

So the path forward would be a multi-emission approach (e.g. tail calls or multiple ring buffer submissions across separate BPF program invocations) to stay within verifier bounds while delivering the full 256KB. I'll work on that — it'll take a bit more time to get right.

mmat11: at this point I think we should also update verifier tests to account for these new maximum values

rafaelroquetto: @NameHaibinZhang my gut feeling points doing some sort of tailcall recursion similarly to what we do in obi_packet_extender_find_existing_tp. But I haven't investigated this deeply.

NameHaibinZhang:
@mmat11 Good call — I've updated the verifier tests in `test_large_buffer_loop.c` to account for the new maximum values. The tests now verify that `large_buf_emit_chunks` correctly clamps to `k_large_buf_per_emit_max` (64KB) regardless of input size, including explicit test cases for 256KB and 100KB inputs being clamped to 4 chunks.

bpf/common/large_buffers.h
NameHaibinZhang: ok

NameHaibinZhang: Thanks for the suggestion! I've implemented a tail call recursion approach similar to obi_packet_extender_find_existing_tp. The design keeps the existing 4-iteration loop in large_buf_emit_chunks (clamped to 64KB per emission via new k_large_buf_per_emit_max), and chains up to 4 batches via a self-recursive obi_large_buf_emit_continue tail call program — achieving 256KB total per-syscall while keeping each program within verifier bounds on 5.10 kernels.

Per-CPU scratch memory (large_buf_emit_state_t) passes the emission context (buffer pointer, remaining bytes, batch counter) between tail call invocations. @rafaelroquetto

rafaelroquetto (commented):
@NameHaibinZhang thank you for the changes, I will try to review them soon.

Just one thing I'd like to raise, from `AGENTS.md`:

```
Comments must be minimal:

    Do not add comments that restate the code.
    Prefer clearer code over explanatory comments.
    Add comments only when they provide necessary context, explain non-obvious behavior, or document verifier, kernel, or ABI constraints.
```

There're too many comments here violating that. The reason that policy exist is because these comments, which in theory should help the reader understand the code, are getting in the way, oftentimes poorly restating obvious code, and worse - because they reference other parts of the code/implementation details, they easily drift  in little time into contradictory statements, leaving the reader confused on whether it is the code or the comment that is wrong.

Would you mind having a pass and cleaning it up?

NameHaibinZhang:
Thanks for the feedback @rafaelroquetto! You're absolutely right — I've done a pass and cleaned up the comments that were restating the code or referencing implementation details of other functions. Pushed in 39092f1f.

bpf/generictracer/large_buf_tailcall.h
```diff
@@ -0,0 +1,27 @@
...
+typedef struct large_buf_emit_state {
+    u64 u_buf;                      // user-space buffer address (advanced after each batch)
+    u32 remaining_bytes;            // bytes remaining to emit
+    pid_connection_info_t pid_conn; // key into ongoing_http for updating byte counters
+    u8 batch_iter;                  // current batch index (1-based; batch 0 is inline)
+    u8 packet_type;                 // PACKET_TYPE_REQUEST or PACKET_TYPE_RESPONSE
+    u8 direction;                   // TCP_SEND or TCP_RECV
+    u8 _pad;
+} large_buf_emit_state_t;
+
```
rafaelroquetto:
```suggestion
enum { k_large_buf_max_batches = 4 };

typedef struct large_buf_emit_state {
    u64 u_buf;
    u32 remaining_bytes;
    pid_connection_info_t pid_conn;
    u8 batch_iter;
    u8 packet_type;
    u8 direction;
    u8 _pad;
} large_buf_emit_state_t;

```

bpf/generictracer/protocol_handler.c
```diff
@@ -100,20 +100,15 @@ int obi_handle_buf_with_args(void *ctx) {
                      (info) ? still_reading(info) : 0);
 
         if (args->protocols.http && info && !info->submitted) {
+            if (info->ssl && !args->ssl) {
```
rafaelroquetto, line 103: what's the purpose of this check? This changes the behaviour of the large buffer being sent, no?

bpf/generictracer/protocol_handler.c
```diff
@@ -100,20 +100,15 @@ int obi_handle_buf_with_args(void *ctx) {
...
-                // Packets are split into chunks if OBI injected the Traceparent
-                // Make sure you look for split packets containing the real Traceparent.
-                // Essentially, when a packet is extended by our sock_msg program and
-                // passed down another service, the receiving side may reassemble the
-                // packets into one buffer or not. If they are reassembled, then the
-                // call to bpf_tail_call(ctx, &jump_table, k_tail_protocol_http); will
-                // scan for the incoming 'Traceparent' header. If they are not reassembled
-                // we'll see something like this:
-                // [before the injected header],[70 bytes for 'Traceparent...'],[the rest].
+                // Traceparent may arrive in a separate buffer when sock_msg
```
rafaelroquetto: Was this comment change intended?

bpf/generictracer/protocol_http.h
```diff
@@ -40,9 +41,8 @@ volatile const u32 high_request_volume;
 
 SCRATCH_MEM_SIZED(http_previous_trace_id, TRACE_ID_SIZE_BYTES);
 
-// empty_http_info zeroes and return the unique percpu copy in the map
-// this function assumes that a given thread is not trying to use many
-// instances at the same time
+// This function assumes that a given thread is not trying to use many
```
rafaelroquetto: comment change intended?

bpf/generictracer/protocol_http.h
```diff
@@ -127,8 +127,6 @@ static __always_inline void finish_http(http_info_t *info, pid_connection_info_t
             bpf_dbg_printk("failed to reserve space in the ringbuf");
         }
 
-        // bpf_dbg_printk("Terminating trace for pid=%d", pid_from_pid_tgid(pid_tid));
```
rafaelroquetto, line 130: why?

rafaelroquetto (changes requested):
I did another pass on the ebpf side. This goes in the right direction but  I feel the tailcall should be `http_send_large_buffer` itself. At the moment, we have something strange - 2 functions that send the large buffers, one inline, and one tailcall that can recursice in itself. It's somewhat duplicating the flow, it would be much better if http_send_large_buffer was self-contained.

Also, with regards to comments: there seem to be lots of unrelated changes - my earlier comments about cleaning up overly-verbose comments referred only to the ones introduced by this PR - otherwise it becomes kind of messy and difficult to parse the PR.

bpf/generictracer/protocol_handler.c
NameHaibinZhang: Not intended, reverted.

bpf/generictracer/protocol_http.h
NameHaibinZhang: Not intended, reverted.

bpf/generictracer/protocol_http.h
NameHaibinZhang: Reverted — out of scope for this PR.

bpf/generictracer/large_buf_tailcall.h
NameHaibinZhang: Simplified the comments. Note that `_pad` is now `action` — it carries the caller's action type across the tail call boundary (needed after refactoring `http_send_large_buffer` into a self-contained tail call).

bpf/generictracer/protocol_handler.c
NameHaibinZhang: This guard ensures that when a connection is marked as SSL, only the SSL path
  (SSL_read/SSL_write hooks) processes the payload. Without it, a plain read() syscall on the
  same fd would also enter the HTTP processing flow and emit ciphertext rather than the
  decrypted HTTP body into the large buffer. For SSE streaming, SSL connections accumulate the
   response body across multiple SSL_read calls, so non-SSL events must be excluded.

  For more context: k_tracer.c already has a first-layer guard via is_ssl_connection() that
  skips handle_buf_with_connection() for SSL connections on the syscall path. However, that
  check depends on the active_ssl_connections map being populated during SSL_do_handshake. If
  the handshake uprobe is missed or the connection info doesn't match exactly, non-SSL events
  can still slip through to the protocol handler. This guard uses info->ssl (set by a prior
  successful SSL_read/SSL_write processing), which is a more reliable indicator for an
  already-tracked HTTP connection, serving as a defense-in-depth check.

  Additionally, the return 0 here does not skip any necessary logic — the generic TCP fallback
   (line 170) requires !info, and we're already inside the info && !info->submitted branch, so
   it is unaffected. @rafaelroquetto

NameHaibinZhang:
> I did another pass on the ebpf side. This goes in the right direction but I feel the tailcall should be `http_send_large_buffer` itself. At the moment, we have something strange - 2 functions that send the large buffers, one inline, and one tailcall that can recursice in itself. It's somewhat duplicating the flow, it would be much better if http_send_large_buffer was self-contained.
> 
> Also, with regards to comments: there seem to be lots of unrelated changes - my earlier comments about cleaning up overly-verbose comments referred only to the ones introduced by this PR - otherwise it becomes kind of messy and difficult to parse the PR.

Agreed — refactored so that http_send_large_buffer is now a self-contained tail call program. All large-buffer emission logic lives in a single SEC program; the inline helper only validates, sets up per-CPU state, and triggers the tail call. This eliminates the duplication between the inline path and the continuation program.

Regarding the unrelated comment changes: apologies for the noise, reverted them all. @rafaelroquetto

NameHaibinZhang (commented):
done

bpf/common/large_buffers.h
```diff
@@ -21,10 +21,13 @@ enum {
     // for the struct overhead.
     k_large_buf_max_size = 1 << 15, // 32K
 
+    // Maximum size per single ring buffer emission (verifier safe).
```
mmat11:
```suggestion
    // Maximum size per single ring buffer emission.
```

pkg/ebpf/common/tcp_large_buffer.go
```diff
@@ -76,6 +93,20 @@ func appendTCPLargeBuffer(parseCtx *EBPFParseContext, record *ringbuf.Record) (r
 		return request.Span{}, true, fmt.Errorf("invalid large buffer action: %d", event.Action)
 	}
 
+	// Truncation detection: when the accumulated buffer length is an exact
+	// multiple of the BPF per-syscall cap (256 KB) and the chunk just appended
+	// is full-sized (16 KB), the BPF emission hit its cap — meaning the
+	// underlying syscall had more data that was discarded. Seal the buffer so
+	// that subsequent chunks (from the next syscall) are not appended,
+	// preserving a contiguous prefix with no holes.
```
mmat11: wouldn't it be easier to detect in ebpf if a buffer is being truncated?

pkg/ebpf/common/tcp_large_buffer.go
```diff
@@ -76,6 +93,20 @@ func appendTCPLargeBuffer(parseCtx *EBPFParseContext, record *ringbuf.Record) (r
...
+	if lb, ok := parseCtx.largeBuffers.Get(key); ok {
+		if len(chunk) == largeBufferMaxChunkPayload &&
+			lb.Len() > 0 &&
+			lb.Len()%largeBufferPerSyscallCap == 0 {
+			parseCtx.sealedLargeBuffers.Add(key, struct{}{})
+		}
+	}
```
rafaelroquetto: I believe this needs to be deleted.

pkg/ebpf/common/common.go
```diff
@@ -246,6 +246,7 @@ type EBPFParseContext struct {
 	redisDBCache               *simplelru.LRU[BpfConnectionInfoT, int]
 	couchbaseBucketCache       *simplelru.LRU[BpfConnectionInfoT, CouchbaseBucketInfo]
 	largeBuffers               *expirable.LRU[largeBufferKey, *largebuf.LargeBuffer]
+	sealedLargeBuffers         *expirable.LRU[largeBufferKey, struct{}]
```
rafaelroquetto: Should no longer be required, needs to be removed.

pkg/ebpf/common/tcp_large_buffer.go
```diff
@@ -20,6 +20,15 @@ const (
...
 )
 
+// BPF per-syscall capture limits. These must match the constants in
+// bpf/common/large_buffers.h.
+const (
+	// Maximum bytes emitted by a single large_buf_emit_chunks call.
+	largeBufferPerSyscallCap = 1 << 18 // 256 KB (per-syscall total, matches k_large_buf_max_http_captured_bytes)
+	// Maximum payload carried by a single chunk event.
+	largeBufferMaxChunkPayload = 1 << 14 // 16 KB per chunk
+)
```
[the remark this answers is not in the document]
rafaelroquetto: Copilot is right here.

rafaelroquetto (commented):
Thank you so much for the changes again. It looks much better and I believe we are almost there - I just caught some dead code that needs to be cleaned up.

bpf/common/large_buffers.h
NameHaibinZhang: Done, applied.

pkg/ebpf/common/tcp_large_buffer.go
NameHaibinZhang: Agreed — the BPF tail-call emission now delivers the full 256KB per syscall, so userspace truncation detection is moot. Removed the heuristic and the `sealedLargeBuffers` infrastructure entirely.

pkg/ebpf/common/tcp_large_buffer.go
NameHaibinZhang: Done, removed.

pkg/ebpf/common/common.go
NameHaibinZhang: Done, removed.

pkg/ebpf/common/tcp_large_buffer.go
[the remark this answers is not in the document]
NameHaibinZhang: Fixed — removed `largeBufferPerSyscallCap` entirely since it was only consumed by the (now-deleted) userspace truncation heuristic.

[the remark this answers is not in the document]
NameHaibinZhang: Done — removed the constant entirely.

bpf/generictracer/protocol_http.h
```diff
@@ -776,27 +764,38 @@ __obi_protocol_http(struct pt_regs *ctx, unsigned char *(*tp_loop_fn)(unsigned c
 
         return 0;
     } else if ((args->packet_type == PACKET_TYPE_RESPONSE) && (info->status == 0)) {
-        http_send_large_buffer(info,
+        handle_http_response(
+            args->small_buf, &args->pid_conn, info, args->bytes_len, args->lw_thread);
+        http_send_large_buffer(ctx,
+                               info,
```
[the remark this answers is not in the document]
NameHaibinZhang: After closer inspection, finish_http only deletes the ongoing_http entry when info->delayed == 1 (line 135). In handle_http_response, delayed is set to 1 only in the else branch (line 346), which is mutually exclusive with the finish_http call path (line 338). So when finish_http is called, delayed is always 0, meaning the entry is never deleted

pkg/ebpf/common/http/qwen.go
```diff
@@ -69,16 +78,45 @@ func QwenSpan(baseSpan *request.Span, req *http.Request, resp *http.Response) (r
...
+	}
+	if !isQwenHost(req) {
+		return false
+	}
+	path := requestPath(req)
+	return strings.Contains(path, "/chat/completions") ||
+		strings.Contains(path, "/completions") ||
+		strings.Contains(path, "/embeddings") ||
+		strings.Contains(path, "/generation")
+}
```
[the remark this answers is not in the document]
NameHaibinZhang: Intentional for now — the host restriction avoids false positives on generic OpenAI-compatible endpoints (e.g., a vLLM cluster serving Llama models via `/chat/completions`). The `model` prefix check alone isn't sufficient since model names aren't always present in the request URL/path. We'll revisit if/when self-hosted Qwen routing becomes a priority.

pkg/ebpf/common/http_transform.go
```diff
@@ -366,7 +366,7 @@ func recoverJSONBodyFromBuffer(buf *largebuf.LargeBuffer) []byte {
 	// Scan forward from header end to find a JSON object or array.
 	for i := bodyStart; i < len(raw); i++ {
 		if raw[i] == '{' || raw[i] == '[' {
-			return raw[i:]
+			return append([]byte(nil), raw[i:]...)
```
mmat11, line 369: question: why is this needed?

mmat11 (commented):
looks much better! just one last thing: I added http and tcp large buffer permutations to the verifier test, can you rebase from main and increase the http one to 256k so we are sure the verifier doesn't explode?

pkg/ebpf/common/http_transform.go
NameHaibinZhang: `UnsafeView()` returns a slice that may reference an internal reusable scratch buffer (documented as "MUST NOT be retained across the next UnsafeView or UnsafeViewAt call on the same buffer"). Since this return value is retained beyond the function scope (passed to `bytes.NewBuffer` / `io.NopCloser` for later consumption), we need an independent copy to prevent data corruption from subsequent buffer operations.

NameHaibinZhang:
> looks much better! just one last thing: I added http and tcp large buffer permutations to the verifier test, can you rebase from main and increase the http one to 256k so we are sure the verifier doesn't explode?

Sure! I'll rebase from main and bump the http_max_captured_bytes verifier permutation to 256k (262144).

bpf/generictracer/protocol_http.h
```diff
@@ -813,3 +812,54 @@ SEC("kprobe/http")
...
+    if (!large_buf) {
+        return 0;
+    }
+
+    large_buf->type = EVENT_TCP_LARGE_BUFFER;
+    large_buf->packet_type = state->packet_type;
+    large_buf->direction = state->direction;
+    large_buf->conn_info = info->conn_info;
+    large_buf->action =
+        state->batch_iter == 0 ? (enum large_buf_action)state->action : k_large_buf_action_append;
```
rafaelroquetto: instead of the cast, why don't you make the type of `action` `enum large_buf_action`?

rafaelroquetto (commented):
Just one final nit, otherwise LGTM - thanks for the changes.

bpf/generictracer/protocol_http.h
NameHaibinZhang: Good point! Changed the field type to `enum large_buf_action` directly and removed the cast.

rafaelroquetto (approved):
Thanks!

mmat11 (approved):

grcevski:
Thanks for all your hard work here @NameHaibinZhang!

PR #2399, opened 2026-06-16
Remove unneeded CachedClock

mariomac:
## Summary

TIL about  [Virtual Dynamically linked Shared Object, vDSO](https://blog.cloudflare.com/its-go-time-on-linux/) which maps some common system calls into userspace memory.

In Go, vDSO is used for efficiently querying system time without the cost of a syscall, so our `CachedClock` implementation adds unnecessary complexity.

## Validation


pkg/export/prom/prom_stats.go
```diff
@@ -183,7 +178,6 @@ func (r *statMetricsReporter) collectMetrics(_ context.Context) {
 	for stats := range r.input {
 		// clock needs to be updated to let the expirer
```
mmat11: probably this comemnt needs to be removed; question: don't we need to do the clock.Update() now?

mariomac: Good catch

rafaelroquetto (approved):

Issue #2014, opened 2026-05-05
Cloud node metadata is exported to Prometheus and OTLP without sensitivity filtering


MrAlias:
## Summary

Cloud resource detector output is copied into `NodeMeta.Metadata` and then appended directly to Prometheus `target_info` labels and OTLP resource attributes. Aside from a small duplicate-attribute filter, OBI does not distinguish routine metadata from potentially sensitive cloud identity and topology fields.

## Impact

This can expose cloud account, instance, image, region, or host-type metadata to Prometheus scrapers and OTLP backends. The impact is limited to metadata disclosure rather than secret leakage or host compromise, so `low` severity is appropriate.

## Environment

- First identified in: ce8adf43956e155cc407026ef6040586f12215c9
- Reviewed against current `main` commit: d51a98d730f950b8d759a8e592669db13d2e4843

## Evidence

- [pkg/appolly/meta/meta_node_cloud.go](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/pkg/appolly/meta/meta_node_cloud.go#L67-L72) copies detector attributes into `NodeMeta.Metadata`.
- [pkg/appolly/meta/meta_node.go](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/pkg/appolly/meta/meta_node.go#L76-L80) enables cloud detectors by default.
- [pkg/appolly/meta/meta_node.go](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/pkg/appolly/meta/meta_node.go#L159-L167) filters only a narrow duplicate-attribute set.
- [pkg/export/otel/otelcfg/common.go](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/d51a98d730f950b8d759a8e592669db13d2e4843/pkg/export/otel/otelcfg/common.go#L127-L129) appends `nodeMeta.Metadata` to OTLP resource attributes.
[...]

PR #2403, opened 2026-06-16
Add resource attribute selection for target info

MrAlias:
## Summary

- Add `target_info` and `traces_target_info` sections to `attributes.select` so resource attributes can be filtered explicitly.
- Preserve default export behavior for detected cloud and host resource metadata when no resource selector is configured.
- Apply `target_info` selection to Prometheus target info and OTLP metric resources, and `traces_target_info` selection to trace/span-metric/service-graph resources.
- Update tests and docs with opt-out examples for fields such as `cloud.account.id`, `cloud.resource_id`, `host.image.id`, and `host.type`.

## Context

This follows the direction discussed in #2014: the issue is better treated as a resource-attribute configurability gap than as a default sensitivity-filtering bug, since the relevant cloud and host attributes are semconv-recommended resource attributes.

## Validation

- `go test ./pkg/export/otel/otelcfg ./pkg/export/otel ./pkg/export/prom ./pkg/appolly`
- `go test -run TestCloudResourceMetadata -count=0 ./internal/test/integration`
- `make lint-markdown`
- `go test ./pkg/export/prom`


rafaelroquetto (approved):

mariomac:
Manually solved a merge conflict in `prom.go`

mariomac (changes requested):
Great changes!! However I think some aspects here might lead to some confusion:

1. The documentation says that we are filtering resource attributes but we are actually only filtering node attributes. Some users might want to filter other resource attributes (e.g. pod name, etc...). I think the filtering code should be integrated in two parts of the code:
    - For OTEL exporter, the `otelcfg.GetResourceAttrs` function (and maybe `otelcfg.ResourceAttrsFromEnv` also).
    -  For Prometheus exporter, the `labelNamesTargetInfo` and `labelValuesTargetInfo`.
 
In both cases, you can maybe reuse the code that we already have for the filtering of other metrics: add the attributes to the `spanOTELGetters` switch and then invoke `spanOTELGetters` or `spanPromGetters` from the `GetResourceAttrs` and `label*TargetInfo`

2. Do we really need to provide different selectors for `target_info` and `traces_target_info`? Maybe we can simplify and just provide a `resource` section that is common for both.

MrAlias:
Addressed the requested changes in dc44512f45.

What changed:

- Replaced the separate `target_info` / `traces_target_info` selector sections with one `attributes.select.resource` section.
- Moved filtering from node metadata pruning to resource-attribute filtering after resource attributes are assembled.
- Applied the resource selector to OTLP app/runtime/service-graph metric resources and OTLP trace resources, including env-derived resource attributes.
- Applied the same resource selector to Prometheus `target_info` / `traces_target_info` label names and values, keeping label names and values filtered from the same ordered resource-label list.
- Updated docs/schema examples to document `resource` as the selector key.
- Added tests covering default preservation, cloud/node metadata exclusion, Kubernetes/service metadata exclusion, and env resource attribute exclusion.

Validation run locally:

- `go test ./pkg/export/otel/otelcfg ./pkg/export/otel ./pkg/export/prom ./pkg/appolly`
- `make lint-markdown`

mariomac (approved):
👏

PR #2405, opened 2026-06-16
Bound avoided-services metric cardinality

### LINKED ISSUE #2037: avoided_services has unbounded service-identity cardinality
## Summary

The internal `avoided_services` metric records `service_name`, `service_namespace`, `service_instance_id`, and `telemetry_type` with no cap or eviction. A large number of unique service identities can therefore grow internal metric state indefinitely.

## Impact

This can pressure agent memory and degrade the internal metrics path. The issue is local to observability state and does not cross a stronger security boundary, so `low` severity is appropriate.

## Environment

- First identified in: c5cd68cf4d5fe696bcbb25ea97809fa4deb3147d

## Evidence

- `pkg/ebpf/common/pids.go` extracts service identity and calls the avoided-services reporters.
- `pkg/export/imetrics/iprom.go` records `avoided_services` with full service-identity labels.
- `pkg/export/otel/metrics_internal.go` records the same attributes in OTEL internal metrics.

## Steps to reproduce

1. From the repository root on `main`, create `/tmp/poc_avoid_services.go` with this exact program:
   ```go
   package main
   import (
       "fmt"
       "os"
       "strconv"
       "runtime"
       "github.com/prometheus/client_golang/prometheus"
       "go.opentelemetry.io/obi/pkg/components/imetrics"
   )
   func main() {
       n := 10000
       if v := os.Getenv("POC_N"); v != "" {
           if parsed, err := strconv.Atoi(v); err == nil {
               n = parsed
           }
       }
       reg := prometheus.NewRegistry()
       pr := imetrics.NewPrometheusReporter(nil, nil, reg)
       for i := 0; i < n; i++ {
           name := fmt.Sprintf("svc-%d", i)
           ns := fmt.Sprintf("ns-%d", i)
           inst := fmt.Sprintf("inst-%d", i)
           pr.AvoidInstrumentationMetrics(name, ns, inst)
       }
       mfs, err := reg.Gather()
       if err != nil {
           fmt.Printf("gather error: %v\n", err)
           return
       }
       var avoidedCount int
       for _, mf := range mfs {
           if mf.GetName() == "obi_avoided_services" {
               avoidedCount = len(mf.GetMetric())
           }
       }
       var ms runtime.MemStats
       runtime.ReadMemStats(&ms)
       fmt.Printf("registered avoided_services series: %d\n", avoidedCount)
       fmt.Printf("heap alloc: %d bytes\n", ms.HeapAlloc)
   }
   ```
2. Run it with `POC_N=200000 go run -mod=vendor -gcflags=all='-N -l' /tmp/poc_avoid_services.go`.
3. Let the program create many unique `service_name` and `service_namespace` combinations and print the registered series count.
4. Observe that the reported `avoided_services` series count grows linearly with the number of unique service tuples, proving unbounded label cardinality.

## Suggested Fix Direction

Bound or aggregate avoided-services metrics, or avoid including full service identity by default.

## Acceptance Criteria

- Avoided-services metrics do not grow without bound as service identities vary.
- High-cardinality service identifiers are either aggregated, capped, or omitted.
- Tests cover repeated unique service identities.

## Note

MrAlias:
## Summary

This bounds the cardinality of the internal `avoided_services` metric so repeated detection of already-instrumented services cannot create unbounded backend-visible series.

The change adds a small internal limiter shared by the Prometheus and OTEL internal metrics reporters. The limiter keeps normal labels for the logical service (`service.name`, `service.namespace`) and avoided telemetry type until the configured series limit is reached, then records additional detections through the OpenTelemetry overflow attribute (`otel.metric.overflow=true`). The Prometheus exporter uses the equivalent `otel_metric_overflow` label.

The metric intentionally does not report `service.instance.id` / `service_instance_id`; that value is unique per service instance and would turn service-instance churn into time-series churn in Prometheus-compatible backends.

The new configuration lives under `internal_metrics.avoided_services`:

- `disabled`: disables the avoided-services internal metric
- `limit`: bounds avoided-services metric series, including the overflow series; defaults to the OTel metric cardinality default of `2000`

Docs and generated config schema have been updated alongside focused Prometheus and OTEL internal-metrics tests.

Fixes #2037.

## Validation

- `go test ./pkg/internal/avoidedsvc ./pkg/export/imetrics ./pkg/export/otel`
- `go test ./pkg/obi`
- `make check-config-schema`
- `make lint-markdown`

### DIFF
--- a/devdocs/config/CONFIG.md
+++ b/devdocs/config/CONFIG.md
@@ -350,6 +350,15 @@ InternalMetricsConfig options for the different metrics exporters
 | `internal_metrics.bpf_metric_scrape_interval` | `duration` | `OTEL_EBPF_BPF_METRIC_SCRAPE_INTERVAL` | `15s` | `30s`, `5m`, `1ms`, etc |  |  |
 | `internal_metrics.exporter` | `string` | `OTEL_EBPF_INTERNAL_METRICS_EXPORTER` | `disabled` | `disabled`, `otel`, `prometheus` |  |  |
 
+### `internal_metrics.avoided_services`
+
+AvoidedServicesConfig controls the avoided-services internal metric.
+
+| YAML Path | Type | Env Var | Default | Values | Deprecated | Description |
+|---|---|---|---|---|---|---|
+| `internal_metrics.avoided_services.disabled` | `boolean` | `OTEL_EBPF_INTERNAL_METRICS_AVOIDED_SERVICES_DISABLED` | `false` |  |  | Disables the avoided-services internal metric. |
+| `internal_metrics.avoided_services.limit` | `integer` | `OTEL_EBPF_INTERNAL_METRICS_AVOIDED_SERVICES_LIMIT` | `2000` |  |  | Bounds the number of avoided-services metric series, including the overflow series. 0 uses the OpenTelemetry default metric cardinality limit. |
+
 ### `internal_metrics.prometheus`
 
 TODO: TLS
--- a/devdocs/config/config-schema.json
+++ b/devdocs/config/config-schema.json
@@ -109,6 +109,24 @@
       "type": "object",
       "description": "AttributesConfig stores the user-provided section for filtering either Application or Network records by attribute values"
     },
+    "AvoidedServicesConfig": {
+      "properties": {
+        "disabled": {
+          "type": "boolean",
+          "description": "Disables the avoided-services internal metric.",
+          "default": false,
+          "x-env-var": "OTEL_EBPF_INTERNAL_METRICS_AVOIDED_SERVICES_DISABLED"
+        },
+        "limit": {
+          "type": "integer",
+          "description": "Bounds the number of avoided-services metric series, including the overflow series. 0 uses the OpenTelemetry default metric cardinality limit.",
+          "default": 2000,
+          "x-env-var": "OTEL_EBPF_INTERNAL_METRICS_AVOIDED_SERVICES_LIMIT"
+        }
+      },
+      "type": "object",
+      "description": "AvoidedServicesConfig controls the avoided-services internal metric."
+    },
     "BedrockConfig": {
       "properties": {
         "enabled": {
@@ -1370,6 +1388,9 @@
     },
     "InternalMetricsConfig": {
       "properties": {
+        "avoided_services": {
+          "$ref": "#/$defs/AvoidedServicesConfig"
+        },
         "bpf_metric_scrape_interval": {
           "type": "string",
           "pattern": "^[0-9]+(ms|s|m)$",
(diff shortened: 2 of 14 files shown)


mariomac (commented):
Looks great! However, do we really need to add some high cardinality attributes like the service instance ID? Maybe instead of a limiter, we can just:
- Remove unnecessary attributes in the metric. Internal metrics are usually _atributeless_ counters.
- Wrap the metric inside an `Expirer` ([pkg/export/otel/expirer.go](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/export/otel/expirer.go)), as we already do for the rest of metric. This would prevent this metric to grow indefinitely over time.

MrAlias:
Thanks @mariomac. I agree with removing `service.instance.id` here.

After checking the OTel service semantic conventions again, `service.instance.id` is the wrong dimension for this metric. It is expected to be unique per service instance within a `service.namespace` / `service.name` pair, so using it on an always-on internal metric makes the backend-facing cardinality problem much worse.

I do not think `Expirer` should replace the limiter, though. `Expirer` helps local exporter state by removing stale labelsets, but it does not protect Prometheus-compatible backends from series they have already scraped. Once a backend has seen a metric name plus labelset, local expiration only stops future samples for that series; it does not undo the historical cardinality until backend retention/compaction catches up.

So I think the right shape for this PR is:

- remove `service.instance.id` / `service_instance_id` from the metric
- keep `service.name`, `service.namespace`, and `telemetry.type` so the metric still identifies the logical service/signal being avoided
- keep the pre-export limiter/overflow path so new identities beyond the cap are collapsed before they become backend-visible series
- treat `Expirer` as a possible follow-up cleanup improvement, not the cardinality-control mechanism

That keeps the metric useful for diagnosing which logical services OBI skipped, while avoiding the per-instance churn and still bounding what we expose to backends.

MrAlias:
Pushed the follow-up in `6174f39af`.

Changes made:

- removed `service.instance.id` from the OTEL `obi.avoided.services` attributes
- removed `service_instance_id` from the Prometheus `obi_avoided_services` labels
- updated the limiter identity so cardinality is scoped to logical service namespace/name plus telemetry type
- added a limiter regression test proving instance ID changes do not consume additional cardinality slots
- updated docs and the internal-metrics integration assertion to match the lower-cardinality metric shape
- updated the PR description to call out that instance IDs are intentionally omitted because they would create backend-visible time-series churn

Validation:

- `go test ./pkg/internal/avoidedsvc ./pkg/export/imetrics ./pkg/export/otel`
- `go test ./pkg/obi`
- `make check-config-schema`
- `make lint-markdown`

mariomac (approved):
🎬

rafaelroquetto (approved):

PR #2407, opened 2026-06-17
GenAI: recover JSON body from truncated TLS capture buffer

NameHaibinZhang:
## Changes

Add utilities for recovering JSON body content from truncated capture buffers:

- `extractJSONRawField(body, key)` - Extract a specific field's raw JSON value using bracket matching
- `extractJSONRawValue(data, startIdx)` - Parse a complete JSON value from a given offset using bracket/quote tracking
- `recoverJSONBodyFromBuffer()` - Scan buffer for JSON start after HTTP frame boundaries, recovering body that may span TLS record splits

### Safety measures:
- Returned bytes are copied to avoid aliasing UnsafeView memory
- Bounded to ContentLength to prevent over-reads
- Falls back gracefully (returns nil) on malformed input

Relates to #2267


mariomac (approved):
LGTM!

pkg/ebpf/common/http_transform.go
```diff
@@ -318,9 +318,52 @@ func HTTPInfoEventToSpan(parseCtx *EBPFParseContext, event *BPFHTTPInfo) (reques
 		return httpRequestToSpan(event, requestBuffer), false, nil
 	}
 
+	// When the body is empty but Content-Length indicates data should be
+	// present, the body bytes may be at a different offset in the raw
+	// buffer (e.g. SSL connections where headers and body arrive in
+	// separate writes that get interleaved). Scan the raw buffer for a
+	// JSON body and replace req.Body so downstream detectors can parse it.
+	if req.ContentLength > 0 {
+		body, readErr := io.ReadAll(req.Body)
```
mmat11: how rare is for this to happen? this `ReadAll` is happening for every buffer, just to test emptiness, is there any way we can refactor or gate this to not happen every time in the happy path?

NameHaibinZhang: Good point! Replaced the `io.ReadAll` with a single-byte probe read: we `Read(probe[:1])` and only enter the recovery path when `n == 0`. On the happy path (body already parsed correctly by `http.ReadRequest`), we just prepend the consumed byte back via `io.MultiReader` — no full-body allocation or copy.

grcevski (approved):
LGTM!

mmat11 (approved):

PR #2410, opened 2026-06-17
GenAI: support nested rerank request/response formats

NameHaibinZhang:
## Changes

Add support for nested rerank request/response formats used by providers like DashScope/Qwen:

- `RerankRequest`: Add `NestedInput` (input.query, input.documents) and `NestedParams` (parameters.top_n) with getter methods that prefer nested values over top-level
- `RerankResponse`: Add `NestedOutput` (output.results) with getter methods
- `VendorRerank`: Add `GetInput()`/`GetOutput()` for unified access to rerank I/O
- Generate `gen_ai.rerank.top_n` attribute from the unified getter

Relates to #2267


mariomac (approved):
👏

Issue #2073, opened 2026-05-08
Implement spec-compliant selective query parameter redaction for `url.full` and `url.query`


NameHaibinZhang:
### Component

OBI (eBPF instrumentation)

## Background

Currently, we strip the entire query string from `url.full` by default to prevent leaking sensitive data (tokens, PII). When `url.query` is opted in, the full query string is preserved as-is. This provides a safe baseline but doesn't fully align with the [OTel HTTP semantic conventions](https://opentelemetry.io/docs/specs/semconv/http/http-spans/).

## What the spec says

Per the semantic conventions:

> Sensitive content provided in `url.full` SHOULD be scrubbed when instrumentations can identify it.

The spec defines a default list of sensitive query parameter keys whose **values** should be redacted:

- `AWSAccessKeyId`
- `Signature`
- `sig`
- `X-Goog-Signature`

Additionally:

- Matching of query parameter keys against the sensitive list SHOULD be case-sensitive
- When a query string value is redacted, the query string key SHOULD still be preserved (e.g. `sig=REDACTED`)
- Instrumentation MAY provide a way to override this list via declarative configuration
- Users may signal additional sensitive parameters

## Proposed Changes

1. **`url.full`**: Instead of stripping the entire query string, selectively redact values of known sensitive keys while preserving the rest of the query string
2. **`url.query`**: When opted in, populate it as a separate attribute with the same redaction rules applied
3. **Configuration**: Provide a way for users to extend the default sensitive key list with additional keys

## Example

Input URL: `https://api.example.com/path?q=OpenTelemetry&sig=abc123&page=1`

| Scenario | Output |
|----------|--------|
| Current (no opt-in) | `https://api.example.com/path` |
| Expected (no opt-in) | `https://api.example.com/path?q=OpenTelemetry&sig=REDACTED&page=1` |
| Expected (opt-in `url.query`) | `url.full`: `https://api.example.com/path?q=OpenTelemetry&sig=REDACTED&page=1`<br>`url.query`: `q=OpenTelemetry

### Describe the solution you'd like

#2069

PR #2415, opened 2026-06-17
Emit `url.query` with automatic redaction for non-Go HTTP spans

pinoOgni:
## Summary

partially addresses #2073 

This PR emits `url.query` in all http spans by default, aligning with otel semconv:
- generic tracer (non-Go): `url.query` populated from the raw request-target
- sensitive query parameters values (from https://opentelemetry.io/docs/specs/semconv/http/http-spans/) are replaced with `REDACTED`.  A user can extend the list using `attributes.extra_redact_query_params`.
- `url.full` on http client spans now includes the query string with the sensitive values redacted.

Moreover, for fastcgi spans, now the full URL is visible in `url.path` and `url.query`

**Breaking changes**: `url.query` is now on by default and that `url.full` now carries query parameters. We should document it in the release notes or somewhere else. 

## Validation

### DIFF
--- a/pkg/export/attributes/attr_selector.go
+++ b/pkg/export/attributes/attr_selector.go
@@ -81,10 +81,23 @@ func parseExtraAttrGroup(group string) (AttrGroups, error) {
 	}
 }
 
+// DefaultRedactQueryParams is the built-in set of query-parameter keys whose
+// values are replaced with REDACTED when url.query or url.full is emitted.
+// Users can override this list via the redact_query_params config field;
+// set to an empty slice to disable all redaction.
+var DefaultRedactQueryParams = []string{
+	"sig", "Signature", "AWSAccessKeyId", "token", "key", "password",
+	"secret", "api_key", "apikey", "access_token", "auth",
+}
+
 // SelectorConfig defines settings for filtering attributes and adding additional attributes
 type SelectorConfig struct {
 	SelectionCfg            Selection
 	ExtraGroupAttributesCfg map[string][]attr.Name
+	// RedactQueryParams lists query-parameter keys whose values are replaced
+	// with REDACTED in url.full and url.query. Nil means use DefaultRedactQueryParams.
+	// Set to []string{} to disable all redaction.
+	RedactQueryParams []string
 }
 
 // AttrSelector returns, for each metric, the attributes that have to be reported
--- a/pkg/export/attributes/attr_selector.go
+++ b/pkg/export/attributes/attr_selector.go
@@ -81,29 +81,54 @@ func parseExtraAttrGroup(group string) (AttrGroups, error) {
 	}
 }
 
-// DefaultRedactQueryParams is the built-in set of query-parameter keys whose
-// values are replaced with REDACTED when url.query or url.full is emitted.
-// All five entries are mandated by the OTel HTTP semconv (matching is case-sensitive):
-// https://opentelemetry.io/docs/specs/semconv/http/http-spans/
-// This list is always applied. Users can extend it via the
-// extra_redact_query_params config field (OTEL_EBPF_EXTRA_REDACT_QUERY_PARAMS);
-// the built-in set cannot be disabled or overridden.
-var DefaultRedactQueryParams = []string{
-	// OTel semconv mandated — do not remove.
-	"X-Amz-Signature",
-	"X-Amz-Credential",
-	"X-Amz-Security-Token",
-	"X-Goog-Signature",
-	"sig", // Microsoft Azure SAS token
+// DefaultSensitiveQueryParams is the built-in list of query-parameter keys whose values
+// are redacted by default. Users can extend or narrow it via SensitiveQueryParamsConfig.
+var DefaultSensitiveQueryParams = []string{
+	// OTel semconv-recommended — https://opentelemetry.io/docs/specs/semconv/http/http-spans/
+	"X-Amz-Signature", "X-Amz-Credential", "X-Amz-Security-Token",
+	"X-Goog-Signature", "sig",
+	// Common sensitive parameters
+	"token", "access_token", "refresh_token", "id_token", "jwt",
+	"session", "sid", "signature",
+	"api_key", "apikey", "client_secret", "secret",
+	"password", "pass", "pwd",
+	"reset_token", "invite_token", "verify_token",
+	"otp", "totp", "mfa_code", "verification_code",
+	"SAMLResponse", "assertion",
+	"card", "cc", "pan", "cvv",
+	"ssn", "tax_id",
+}
+
+// SensitiveQueryParamsConfig controls which query-parameter keys are redacted.
+// The effective list is DefaultSensitiveQueryParams + Add - Remove.
+// When both Add and Remove are empty, DefaultSensitiveQueryParams is used unchanged.
+type SensitiveQueryParamsConfig struct {
+	Add    []string `yaml:"add" env:"OTEL_EBPF_SENSITIVE_QUERY_PARAMS_ADD" envSeparator:","`
+	Remove []string `yaml:"remove" env:"OTEL_EBPF_SENSITIVE_QUERY_PARAMS_REMOVE" envSeparator:","`
+}
+
+func (c SensitiveQueryParamsConfig) Effective() []string {
+	if len(c.Add) == 0 && len(c.Remove) == 0 {
+		return DefaultSensitiveQueryParams
+	}
+	removeSet := make(map[string]struct{}, len(c.Remove))
+	for _, k := range c.Remove {
+		removeSet[k] = struct{}{}
+	}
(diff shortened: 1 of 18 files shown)


pkg/export/attributes/attr_selector.go
```diff
@@ -81,10 +81,29 @@ func parseExtraAttrGroup(group string) (AttrGroups, error) {
...
 }
 
+// DefaultRedactQueryParams is the built-in set of query-parameter keys whose
+// values are replaced with REDACTED when url.query or url.full is emitted.
+// All five entries are mandated by the OTel HTTP semconv (matching is case-sensitive):
+// https://opentelemetry.io/docs/specs/semconv/http/http-spans/
+// This list is always applied. Users can extend it via the
+// extra_redact_query_params config field (OTEL_EBPF_EXTRA_REDACT_QUERY_PARAMS);
+// the built-in set cannot be disabled or overridden.
+var DefaultRedactQueryParams = []string{
```
grcevski: I asked Codex about this and it came up with a list of others that seem reasonable to be added:

Common URL/query parameters worth treating as sensitive or redactable:
- Auth/session: token, access_token, refresh_token, id_token, jwt, session, sid, auth, authorization
OAuth/SAML: code, client_secret, assertion, SAMLResponse, state when app-specific
- API keys/signatures: api_key, apikey, key, secret, signature, sig, X-Amz-Signature, X-Amz-Credential, X-Amz-Security-Token
- Passwords/reset flows: password, pass, pwd, reset_token, invite_token, verify_token
- MFA: otp, totp, mfa_code, verification_code
- PII: email, phone, address, name, dob, ssn, tax_id
- Payment: card, cc, pan, cvv, iban, account_number, routing_number

Usually redacted by broad pattern around token|secret|key|password|session|signature|credential|code|ssn|card|email|phone|account.

grcevski (approved):
This looks good to me, I'd just expand the initial set of values we scrub, better to be safer and have customer relax their rules if they want to.

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -549,13 +560,18 @@ func TraceAttributesSelector(span *request.Span, optionalAttrs map[attr.Name]str
...
-		// sensitive data (tokens, PII). Include them only when explicitly opted in.
-		var queryString string
-		if idx := strings.IndexByte(urlPath, '?'); idx > 0 {
-			queryString = urlPath[idx+1:]
+		// Strip or scrub query parameters from url.full to avoid leaking sensitive
+		// data (tokens, PII). When url.query is opted in, known sensitive keys are
+		// redacted (REDACTED) rather than the entire string being dropped.
+		var scrubbedQS string
+		if idx := strings.IndexByte(urlPath, '?'); idx >= 0 {
 			if _, ok := optionalAttrs[attr.HTTPUrlQuery]; !ok {
```
MrAlias: This still couples `url.full` query preservation to whether `url.query` is selected. If an operator excludes `url.query` to avoid emitting the separate high-cardinality attribute, client spans go back to dropping the query from `url.full` entirely, which undercuts the main redaction fix in this PR.

I think `url.full` should always keep the scrubbed query when it is available, and the selector should only decide whether to append the separate `url.query` attribute.

pkg/export/attributes/attr_selector.go
```diff
@@ -81,10 +81,29 @@ func parseExtraAttrGroup(group string) (AttrGroups, error) {
 	}
 }
 
+// DefaultRedactQueryParams is the built-in set of query-parameter keys whose
+// values are replaced with REDACTED when url.query or url.full is emitted.
+// All five entries are mandated by the OTel HTTP semconv (matching is case-sensitive):
+// https://opentelemetry.io/docs/specs/semconv/http/http-spans/
+// This list is always applied. Users can extend it via the
```
MrAlias: This configuration shape does not match the OTel semconv guidance for sensitive query parameter configuration. Semconv says instrumentation may provide declarative config, but if it does, it should use `sensitive_query_parameters` as a full override of the default list, not an additive list on top of non-disableable defaults.

If OBI intentionally wants additive-only behavior, can we avoid describing this as semconv-mandated and document it as an OBI policy choice? Otherwise, this should probably be modeled as an override so operators can also relax false positives.

MrAlias (changes requested):
Thanks for tightening the query-string handling here. I found one behavior issue around `url.full` and one semconv/configuration mismatch that I think we should address before merging.

pkg/export/otel/tracesgen/tracesgen.go
pinoOgni: Oh yes, agree!

pkg/export/attributes/attr_selector.go
pinoOgni: > This configuration shape does not match the OTel semconv guidance for sensitive query parameter 
configuration. Semconv says instrumentation may provide declarative config..

Thanks!

Yes, we need to change it because the current design (`DefaultRedactQueryParams` as a non-overridable floor) also conflicted with the goal of letting users relax false positives from a broader list, @grcevski suggestion.

We can rename the field to `sensitive_query_params` to match the semconv guidance, with full-override semantics and when set by the user it **completely** replaces the default. 

The default value will include the 5 semconv-recommended keys plus Nikola's broader list, so users get broad protection out of the box and can relax it if needed.

Note: a user who explicitly `sets sensitive_query_params: []` would emit raw X-Amz-Signature and so on. We can just document it.

Update: ok, I need to think about it more carefully. I'll write a list of ideas asap.

pinoOgni: So, these are some ideas:

1.  built-in list, additive only (`extra_redact_query_params`): the user can only extend the built-in list, never remove from it. Safe by default, no risk of accidentally disabling protection. This was the original approach and matches the `ExtraGroupAttributes` pattern already in OBI. The issue is that users can't remove false positives from the built-in list.
2.  single list, full-override semantics (`sensitive_query_params`): one flat list; when set by the user it replaces the built-in list entirely. Gives full control, but if a user sets one custom key accidentally wipes the entire default protection. Matches the `sensitive_query_parameters` guidance in OTel semconv.
3.  built-in list + add/remove:  mirrors the `InclusionLists` (include/exclude) pattern already used here for attribute selection. The built-in list (OTel semconv keys + common auth/credential params from Nikola) is always the baseline. Users can extend or fix false positives without rewriting the whole list:
```
attributes:
  sensitive_query_params:
    add: [my_internal_token]
    remove: [sig]
```
Implemented in this last [commit](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2415/commits/06c6981c76cff36531f2e2514141867097c2453b)
4. flat list + mode flag (`sensitive_query_params` + `sensitive_query_params_mode: addition|override`): a flat list paired with an enum that controls whether the list extends or replaces the built-in set.

MrAlias (commented):
This looks mergeable to me if we narrow the PR text around "all HTTP spans" and track the Go net/http query-capture gap separately.

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -500,6 +503,13 @@ func TraceAttributesSelector(span *request.Span, optionalAttrs map[attr.Name]str
 			attrs = append(attrs, semconv.GraphQLOperationName(span.GraphQL.OperationName))
 			attrs = append(attrs, request.GraphqlOperationType(span.GraphQL.OperationType))
 		}
+		if _, ok := optionalAttrs[attr.HTTPUrlQuery]; ok {
+			if idx := strings.IndexByte(span.FullPath, '?'); idx >= 0 {
```
MrAlias, line 535: This still leaves Go `net/http` spans out of the new `url.query` behavior because the Go tracer reads only `Request.URL.Path` and `HTTPRequestTraceToSpan` copies that value into `FullPath`. In a Go service handling `/search?sig=secret`, `FullPath` remains `/search`, so the selector has no query string to emit or scrub.

I’m okay tracking this as a follow-up if we narrow this PR’s claim from “all HTTP spans” to the HTTP paths that preserve a full request target today, and open an issue for Go `net/http` query capture. Otherwise, I think this should be fixed here before closing #2073 broadly.

pinoOgni: Oh, good catch! Yes, I prefer to merge this one and then open a follow-up issue to add to the Go net/http spans. 
Thanks Tyler!

MrAlias: Sounds good. @pinoOgni can you link the issue tracking the follow-up work in this tread once you have it created so we don't lose track?

MrAlias (approved):

pkg/export/otel/tracesgen/tracesgen.go
pinoOgni: Follow up issue for Go `net/http` here https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2475

PR #2423, opened 2026-06-18
Propagate route harvest cancellation

### LINKED ISSUE #2034: Timeout does not cancel blocked Java extraction

MrAlias:
## Summary

- Thread the route harvest timeout context into Java route extraction.
- Use the context-aware JVM attach path for Java route harvesting.
- Close the route output reader when the harvest context is canceled so blocked reads can unwind.
- Add tests for repeated Java harvest timeouts and canceled blocked route output reads.

## Why

This completes the route-harvester integration step for #2034 after the internal JVM attach path gained context support. Previously, the harvester could return on timeout while the Java extraction worker continued blocked in the background. Passing the timeout context through attach and output reading gives the worker a cancellation path instead of accumulating retained workers across repeated timeouts.

Fixes #2034.

## Validation

- `make generate`
- `go test ./pkg/internal/transform/route/harvest`
- `go test ./pkg/internal/jvmtools/...`
- `GOOS=darwin go test -c -o /tmp/harvest-darwin.test ./pkg/internal/transform/route/harvest`
- `go test ./pkg/internal/transform/route/...`

### DIFF
--- a/pkg/internal/transform/route/harvest/harvester.go
+++ b/pkg/internal/transform/route/harvest/harvester.go
@@ -5,6 +5,7 @@ package harvest // import "go.opentelemetry.io/obi/pkg/internal/transform/route/
 
 import (
 	"context"
+	"errors"
 	"log/slog"
 	"os"
 	"path/filepath"
@@ -28,7 +29,7 @@ type RouteHarvester struct {
 	mux      *sync.Mutex
 
 	// testing related
-	javaExtractRoutes func(fileInfo *exec.FileInfo) (*RouteHarvesterResult, error)
+	javaExtractRoutes func(ctx context.Context, fileInfo *exec.FileInfo) (*RouteHarvesterResult, error)
 	nodeExtractRoutes func(pid app.PID) (*RouteHarvesterResult, error)
 }
 
@@ -108,7 +109,7 @@ func (h *RouteHarvester) HarvestRoutes(fileInfo *exec.FileInfo) (*RouteHarvester
 		switch fileInfo.SDKLanguage() {
 		case svc.InstrumentableJava:
 			if _, ok := h.disabled[svc.InstrumentableJava]; !ok {
-				r, err := h.javaExtractRoutes(fileInfo)
+				r, err := h.javaExtractRoutes(ctx, fileInfo)
 				if err != nil {
 					resultChan <- result{err: err}
 					return
@@ -138,6 +139,10 @@ func (h *RouteHarvester) HarvestRoutes(fileInfo *exec.FileInfo) (*RouteHarvester
 	// Wait for either completion or timeout
 	select {
 	case result := <-resultChan:
+		if errors.Is(result.err, context.DeadlineExceeded) {
+			h.log.Warn("route harvesting timed out", "timeout", h.timeout, "pid", fileInfo.Pid())
+			return nil, &HarvestError{Message: "route harvesting timed out"}
+		}
 		return result.r, result.err
 	case <-ctx.Done():
 		h.log.Warn("route harvesting timed out", "timeout", h.timeout, "pid", fileInfo.Pid())
--- a/pkg/internal/transform/route/harvest/java.go
+++ b/pkg/internal/transform/route/harvest/java.go
@@ -6,6 +6,7 @@
 package harvest // import "go.opentelemetry.io/obi/pkg/internal/transform/route/harvest"
 
 import (
+	"context"
 	"fmt"
 	"log/slog"
 
@@ -23,8 +24,8 @@ func NewJavaRoutesHarvester() *JavaRoutes {
 	}
 }
 
-func (h *JavaRoutes) ExtractRoutes(fileInfo *exec.FileInfo) (*RouteHarvesterResult, error) {
-	routes, err := javaharvest.ExtractRoutes(fileInfo)
+func (h *JavaRoutes) ExtractRoutes(ctx context.Context, fileInfo *exec.FileInfo) (*RouteHarvesterResult, error) {
+	routes, err := javaharvest.ExtractRoutes(ctx, fileInfo)
 	if err != nil {
 		return nil, fmt.Errorf("extracting Java routes from class files: %w", err)
 	}
--- a/pkg/internal/transform/route/harvest/harvester_test.go
+++ b/pkg/internal/transform/route/harvest/harvester_test.go
@@ -4,6 +4,7 @@
 package harvest
 
 import (
+	"context"
 	"errors"
 	"os"
 	"path/filepath"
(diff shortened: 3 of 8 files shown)


pkg/internal/transform/route/harvest/harvester.go
```diff
@@ -97,14 +97,6 @@ func (h *RouteHarvester) HarvestRoutes(fileInfo *exec.FileInfo) (*RouteHarvester
...
 	resultChan := make(chan result, 1)
 
-	// We need to fix this in the downstream library and then we can remove this code
-	if fileInfo.SDKLanguage() == svc.InstrumentableJava {
-		runtime.LockOSThread()
-		defer runtime.UnlockOSThread()
-		h.java.Attacher.Init()
-		defer h.java.Attacher.Cleanup()
-	}
-
```
mariomac: Did we fix it in the downstream library?

pkg/internal/transform/route/harvest/java.go
```diff
@@ -28,7 +30,7 @@ type JavaRoutes struct {
 type JavaAttacher interface {
 	Init()
 	Cleanup()
-	Attach(pid app.PID, argv []string, ignoreOnJ9 bool) (io.ReadCloser, error)
+	AttachContext(ctx context.Context, pid app.PID, argv []string, ignoreOnJ9 bool) (io.ReadCloser, error)
```
mariomac: Nit: given that contextless `Attach(` method is unused (or limited to few tests), there is no need to add the `Context` suffix to the function name.  I'd just keep `Attach(context, pid, argv, ignoreJ9)`.

pkg/internal/transform/route/harvest/java.go
```diff
@@ -203,3 +223,7 @@ func (j RealJavaAttacher) Cleanup() {
 func (j RealJavaAttacher) Attach(pid app.PID, argv []string, ignoreOnJ9 bool) (io.ReadCloser, error) {
 	return j.Attacher.Attach(int(pid), argv, ignoreOnJ9)
 }
+
+func (j RealJavaAttacher) AttachContext(ctx context.Context, pid app.PID, argv []string, ignoreOnJ9 bool) (io.ReadCloser, error) {
```
mariomac: As mentioned in an earlier comment, I'd remove the old `Attach` and rename this method to `Attach`.

pkg/internal/transform/route/harvest/java.go
```diff
@@ -153,8 +155,19 @@ func (h *JavaRoutes) processSymbolLine(lineBytes []byte, routes []string) []stri
 }
 
 func (h *JavaRoutes) ExtractRoutes(pid app.PID) (*RouteHarvesterResult, error) {
+	return h.ExtractRoutesContext(context.Background(), pid)
+}
+
+func (h *JavaRoutes) ExtractRoutesContext(ctx context.Context, pid app.PID) (*RouteHarvesterResult, error) {
```
mariomac: Similar to my previous comment, given how limited is this method invoked, I'd just remove the old `ExtractRoutes` and rename `ExtractRoutesContext` to `ExtractRoutes`.

pkg/internal/transform/route/harvest/java.go
```diff
@@ -165,6 +178,10 @@ func (h *JavaRoutes) ExtractRoutes(pid app.PID) (*RouteHarvesterResult, error) {
 	}
 
 	defer out.Close()
+	stopClose := context.AfterFunc(ctx, func() {
```
mariomac: TIL

mariomac (commented):
Pretty good!!! I just have few cosmetic suggestions.

pkg/internal/transform/route/harvest/harvester.go
MrAlias: Yes. The cancellable boundary was added in the internal `pkg/internal/jvmtools` fork in #2360 via `AttachContext`, and this PR consumes that from route harvesting. The old harvester-side workaround is no longer needed because the attach path itself now has a context-aware API.

pkg/internal/transform/route/harvest/java.go
MrAlias: Agreed. I removed the context-free route-local wrapper and changed the route-harvesting `JavaAttacher` interface to expose `Attach(ctx, pid, argv, ignoreOnJ9)` directly. The lower-level `pkg/internal/jvmtools` API still keeps `AttachContext`, but the route package now has only the context-bearing shape.

pkg/internal/transform/route/harvest/java.go
MrAlias: Done. I removed the old route-local `Attach` wrapper and renamed the context-bearing route-level method to `Attach(ctx, pid, argv, ignoreOnJ9)`.

pkg/internal/transform/route/harvest/java.go
MrAlias: Done. I removed the context-free `ExtractRoutes` wrapper and made `ExtractRoutes(ctx, pid)` the single route-harvesting entry point.

grcevski (approved):
LGTM!

PR #2417, opened 2026-06-17
Add direct Go channel handoff probes

MrAlias:
## Summary

This adds the next #2238 feature slice for direct unbuffered Go channel handoffs.

The Go tracer now registers `runtime.chansend1`, `runtime.chanrecv1`, and `runtime.chanrecv2` probes when `runtime.hchan` offsets are available for a binary. The new runtime probes capture active OBI span contexts on channel send/receive and emit the existing `EVENT_GO_CHANNEL_LINK` record for direct unbuffered channel handoffs.

## Scope

This intentionally does not add buffered channel slot correlation, `runtime.selectgo` support, docs/examples, or feature configuration. The SIG decided no feature flag or configuration is needed for now, so this slice gates runtime channel probe attachment on offset availability.

Included here:

- add direct channel sender/receiver state maps and invocation tracking
- emit receiver-side channel-link events for unbuffered channel handoffs
- ignore nil channels, buffered channels, invalid span contexts, and self-links
- mark multiple waiters on the same channel pointer as ambiguous rather than guessing
- record per-binary channel offset availability before Go probe registration
- cover probe-registration gating with focused unit tests

## Validation

- `make generate`
- `clang-format -i bpf/gotracer/go_runtime.c bpf/gotracer/maps/runtime.h`
- `go test -count=1 ./pkg/ebpf/common ./pkg/runtimemetrics ./pkg/internal/ebpf/gotracer`
- `make compile`

Issue #2238, opened 2026-06-03
Support receiver-side span links for Go channel handoffs


MrAlias:
### Component

OBI (eBPF instrumentation)

### Problem

Go applications commonly hand work between goroutines through channels. OBI can emit spans for the sending and receiving work, but today it does not express the async causality between those spans when the work crosses a Go `chan` boundary.

PR #1727 demonstrates a working PoC, but it is intentionally too large for review: it includes generic span-link export support, Go runtime channel probes, offset tracking, pending-link reconstruction, config/docs, and an example stack in one draft PR.

### Describe the solution you would like

Implement an opt-in feature for receiver-side span links across supported Go channel handoffs.

### Initial production scope:

- Add links only to OBI-generated spans.
- Emit receiver-side links only: the receiver span links to the sender span.
- Do not emit reciprocal links.
- Do not rewrite trace IDs or parent-child relationships.
- Support `runtime.chansend1`, `runtime.chanrecv1`, and `runtime.chanrecv2` first.
- Support direct unbuffered channel handoffs and buffered channel slot correlation.
- Gate all runtime channel probes behind `ebpf.go_channel_span_links`, defaulting to disabled.

### Proposed PR breakdown:

1. Add generic span-link export plumbing: `request.SpanLink`, `Span.Links`, OTLP link export, and tests, including link placement when OBI creates queue/processing subspans.
2. Add config and `runtime.hchan` offset discovery: `ebpf.go_channel_span_links`, env/schema/docs, and offset availability tracking.
3. Add channel-link event parsing and pending-link reconstruction in user space: event type, event struct, bounded pending-link cache, dedupe, invalid/self-link filtering, and unit tests.
4. Add direct unbuffered Go channel handoff probes behind the feature flag.
5. Add buffered channel handoff correlation using `runtime.hchan` `sendx`/`recvx` offsets.
6. Add docs and a runnable `examples/go-channel-links` stack after behavior is settled.

### Completion criteria:

- With the feature disabled, OBI behavior and probe attachment are unchanged.
- With the feature enabled, supported Go channel handoffs produce a span link on the receiver span pointing to the sender span.
- Sender spans do not receive reciprocal links.
- The linked spans remain separate traces unless they already shared a trace for other reasons.
- Missing `runtime.hchan` offsets disable channel probes for that binary instead of failing instrumentation.
- Pending links are bounded, deduplicated, expire safely, and ignore invalid/self links.
[...]

PR #2432, opened 2026-06-18
Track buffered Go channel handoffs

MrAlias:
## Summary

This adds the next #2238 feature slice after the direct unbuffered channel probes merged in #2417.

The Go runtime probes now correlate buffered channel handoffs by recording sender span context for the runtime `hchan` buffer slot used by a completed send, then consuming that sender context when a receiver takes the corresponding buffered slot. Matching handoffs emit the existing `EVENT_GO_CHANNEL_LINK` event so the receiver span receives a link to the sender span through the existing user-space pending-link path.

Included here:

- add a buffered channel sender map keyed by channel identity and buffer slot
- capture sender and receiver span context in channel invocation state
- use `runtime.hchan.qcount` with `sendx` and `recvx` to distinguish buffered queue operations from direct handoffs
- add `qcount` offset discovery, BPF/Go offset enum alignment, and probe gating coverage
- regenerate `pkg/internal/goexec/offsets.json` with `make update-offsets`

## Scope

This intentionally keeps the existing direct unbuffered behavior from #2417 and does not add `runtime.selectgo` support, docs/examples, or feature configuration.

The offset database also includes the version freshness updates produced by the full `make update-offsets` run.

## Validation

- `make update-offsets`
- `make generate/all`
- `go test -count=1 ./pkg/internal/goexec ./pkg/ebpf/common ./pkg/runtimemetrics ./pkg/internal/ebpf/gotracer`
- `make compile`


bpf/gotracer/go_runtime.c
```diff
@@ -291,18 +291,43 @@ static __always_inline void emit_channel_handoff(chan_handoff_t *sender, chan_ha
...
+read_channel_u64(const void *chan_ptr, go_offset_const field, u64 *value) {
+    if (!chan_ptr || !value) {
         return false;
     }
 
     off_table_t *ot = get_offsets_table();
-    const u64 dataqsiz_off = go_offset_of(ot, (go_offset){.v = _hchan_dataqsiz_pos});
-    if (dataqsiz_off == (u64)-1) {
+    const u64 offset = go_offset_of(ot, (go_offset){.v = field});
+    if (offset == (u64)-1) {
```
grcevski, line 302: We don't set the values that are not found to -1 in the RegisterOffsets, I think we probably should. They just stay as 0s.

bpf/gotracer/go_runtime.c
```diff
@@ -392,10 +457,23 @@ static __always_inline int channel_send_return(struct pt_regs *ctx) {
...
+
+    if (dataqsiz == 0) {
         emit_direct_channel_handoff(&chan_key);
+        goto done;
     }
 
+    u64 qcount = 0;
+    u64 sendx = 0;
+    if (invocation->has_handoff && read_channel_qcount((void *)invocation->chan_ptr, &qcount) &&
+        qcount > 0 && read_channel_sendx((void *)invocation->chan_ptr, &sendx)) {
```
grcevski: I think based on the comment in the Go source here: https://github.com/golang/go/blob/master/src/runtime/chan.go#L229

the channel will bypass the buffer if there's a receiver waiting. In that case qcount = 0.

bpf/gotracer/go_runtime.c
```diff
@@ -427,11 +506,25 @@ static __always_inline int channel_recv_start(struct pt_regs *ctx) {
...
-            record_direct_channel_receiver(&chan_key, &receiver);
+        if (invocation.has_handoff) {
+            record_direct_channel_receiver(&chan_key, &invocation.handoff);
         } else {
             record_direct_channel_receiver(&chan_key, NULL);
         }
+    } else {
+        u64 qcount = 0;
+        u64 recvx = 0;
+        if (read_channel_qcount((void *)chan_ptr, &qcount) && qcount > 0 &&
```
grcevski: I think we should find a different place where to put this probe, or at least where to read these values. What happens is that many goroutines can technically end up here and we might read the wrong span, since the lock on the channel happens only later here: https://github.com/golang/go/blob/master/src/runtime/chan.go#L586

This probe is on the function start and many goroutines will read the same value.

grcevski: I think we probably want to handle this in the return probe.

bpf/gotracer/go_runtime.c
MrAlias: Good call. I initialized the hchan offset entries with the existing `-1` sentinel before copying discovered offsets into the BPF table. That keeps `qcount`'s legitimate offset `0` distinct from a missing field, and the new unit test covers that behavior.

bpf/gotracer/go_runtime.c
MrAlias: Yes, thanks. Buffered sends with `qcount == 0` now go through the direct handoff path at send return, so the waiting-receiver path that bypasses the buffer can be linked instead of being ignored.

bpf/gotracer/go_runtime.c
MrAlias: Agreed. I moved buffered receive slot consumption out of the start probe and into the return probe, using the post-operation `recvx` to consume the previous buffer slot. The start probe now only records direct-waiting receivers, so it no longer consumes buffered sender state from pre-lock queue values.

grcevski (approved):
LGTM!

PR #2436, opened 2026-06-19
Fix connection and goroutine leak in waitUntilReadyToServe

### LINKED ISSUE #1917: waitUntilReadyToServe doesn't close the response body
Address https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1916#discussion_r3137134307

mariomac:
## Summary

Closes #1917.

Fixes an Connection/FD leak in an integration test helper. The current implementation was also leaking a goroutine that would keep polling the URL indefinitely if the endpoint is unavailable or does not respond OK.

Not an important issue but it will prevent future complaints from code linters/verificators or automated reviews.

## Validation


internal/test/integration/cloud_test.go
```diff
@@ -181,23 +180,18 @@ func setupMockGCPIMDS(t *testing.T, net dockertest.Network) {
...
-		for {
-			resp, err := http.Get(metaURL)
-			if err != nil || resp.StatusCode != http.StatusOK {
-				time.Sleep(500 * time.Millisecond)
-				continue
-			}
-			close(done)
-			return
+	for time.Since(startTime) < 30*time.Second {
+		resp, err := http.Get(metaURL)
```
pinoOgni: question: can the mock server accepts the TCP connection but never writes a response? Because   `http.Get` uses the default client which has no timeout, so the call blocks indefinitely and the `time.Since` loop guard never gets a chance to evaluate it again. If yes, we can add something like: 
```
client := &http.Client{Timeout: 500 * time.Millisecond}
resp, err := client.Get(metaURL)
```

That way each attempt is bounded and the 30-second loop timeout is reliable.

pinoOgni (approved):

internal/test/integration/cloud_test.go
mariomac: Good point! Will implement it just in case

grcevski (approved):
LGTM!

PR #2437, opened 2026-06-19
fix(postgres): detect connections with multi-message TCP segments

### LINKED ISSUE #1464: PostgreSQL protocol detection fails on multi-message TCP segments
## Description

PostgreSQL protocol detection (`is_postgres` in `bpf/generictracer/protocol_postgres.h`) fails to classify many real-world PostgreSQL connections because the `message_size == data_len` strict equality check rejects TCP segments containing multiple concatenated PostgreSQL wire protocol messages.

## Observed Behavior

When running OBI v0.5.0 (standalone binary) against a PostgreSQL 16 server with simple queries (`SELECT 1`, `CREATE TABLE`, `INSERT`, `SELECT * FROM t1 LIMIT 5`), **no PostgreSQL traces are emitted**. MySQL traces (on the same node, same OBI instance) work correctly.

BPF debug logs show:
```
is_postgres: message length mismatch: message_size=274 data_len=467
is_postgres: no known command found
is_postgres: postgres! message_type=81  <-- sometimes succeeds for simple Q messages
```

The `message_type=81` (`'Q'` = SimpleQuery) detection occasionally succeeds, but most segments fail the `message_size == data_len` check.

## Root Cause Analysis

Looking at `protocol_postgres.h`, the `is_postgres` function:

1. Iterates through up to `k_pg_messages_in_packet_max` (10) messages in the TCP segment
2. Accumulates `message_size` from each message header
3. **Requires `message_size == data_len`** — if the sum of parsed message lengths doesn't exactly match the TCP payload size, detection fails

This fails for several common scenarios:

### Scenario 1: Multi-message response segments
PostgreSQL server responses often concatenate multiple backend messages in a single TCP segment (e.g., `T` (RowDescription) + `D` (DataRow) + `C` (CommandComplete) + `Z` (ReadyForQuery)). Since the parser only checks for frontend message types (`Q`, `P`, `B`, `E`), `includes_known_command` is always `false` for response segments — but the parser still runs and produces the "mismatch" or "no known command found" logs.

### Scenario 2: Startup messages
PostgreSQL's `StartupMessage` and `SSLRequest` messages use a different wire format (no type byte prefix), causing the 5-byte header parser to misinterpret data and accumulate incorrect sizes.

### Scenario 3: Messages exceeding loop limit
TCP segments with more than 10 concatenated messages hit the `k_pg_messages_in_packet_max` loop limit, leaving `message_size < data_len`.

## Expected Behavior

PostgreSQL connections should be reliably detected when standard SQL queries are being executed, similar to how MySQL connections are detected via `protocol_mysql.h`.

## Environment

- OBI version: v0.5.0 (standalone binary from GitHub releases)
- PostgreSQL: 16-alpine (official Docker image)
- Kernel: 6.12 (GKE)
- Client: `psql` from postgres:16-alpine image
- Traffic: Simple queries (`SELECT 1`, `CREATE TABLE IF NOT EXISTS`, `INSERT`, `SELECT * FROM t1 LIMIT 5`)

## Reproduction Steps

1. Deploy PostgreSQL 16 and a simple traffic generator:
   ```bash
   # PostgreSQL server
   docker run -d --name pg -e POSTGRES_PASSWORD=test -e POSTGRES_DB=testdb postgres:16-alpine
   
   # Traffic generator (in a loop)
   while true; do
     PGPASSWORD=test psql -h <pg_ip> -U postgres testdb -c "SELECT 1"
     PGPASSWORD=test psql -h <pg_ip> -U postgres testdb -c "CREATE TABLE IF NOT EXISTS t1 (id INT PRIMARY KEY, val VARCHAR(100))"
     PGPASSWORD=test psql -h <pg_ip> -U postgres testdb -c "INSERT INTO t1 VALUES (floor(random()*10000), 'test') ON CONFLICT (id) DO UPDATE SET val='test'"
     PGPASSWORD=test psql -h <pg_ip> -U postgres testdb -c "SELECT * FROM t1 LIMIT 5"
     sleep 2
   done
   ```

2. Run OBI with `trace_printer: text` and `bpf_debug: true`
3. Observe: MySQL traces appear as `SQLServer`/`SQLClient`, but no PostgreSQL traces

## Suggested Fix

Consider relaxing the strict equality check to `message_size > data_len` (reject only if parsed sizes exceed the data length), or allow classification when at least one known frontend command was found even if `message_size < data_len` (which can happen due to the loop limit or startup message artifacts).

Alternatively, could the connection be classified as postgres when `message_type=81` (Query) is detected, even if the full segment doesn't parse cleanly?

mariomac:
## Summary

  Fixes #1464.

  PostgreSQL connections were frequently not detected, so no traces were emitted
  for them. The kernel-side protocol classifier `is_postgres`
  (`bpf/generictracer/protocol_postgres.h`) required the parsed messages to fill
  the captured TCP segment **exactly** (`message_size == data_len`) before
  accepting a connection as Postgres. That exact-match requirement rejected
  several common, valid cases:

  - **Multiple concatenated messages** whose total doesn't exactly fill the segment.
  - **More than `k_pg_messages_in_packet_max` (10) messages** in one segment: the
    walk stops at the loop limit, so `message_size < data_len`.
    totals never match.

  ### The fix

  Classify a connection as Postgres as soon as we parse **at least one
  well-formed frontend command message** (`Q`/`P`/`B`/`E`), instead of requiring
  an exact size match. While walking the segment we now:

  - Only count messages that **fit entirely** within the captured segment.
  - Validate each message's declared length is `>= 4` (the length field counts
    itself, so anything smaller can't be Postgres).

  ### Why this doesn't introduce false positives

  The "must fit entirely within the segment" rule doubles as the guard against
  misclassifying other protocols. Text-based protocols are rejected naturally:
  their bytes are printable ASCII, so when interpreted as a big-endian length they
  yield huge values that cannot fit in the segment. For example, an HTTP `POST`
  request starts with `P` (which looks like a `Parse` command), but the following
  four bytes `"OST "` decode to a length of `0x4F535420` (~1.3 billion); that
  message can't fit, so it's not counted, no frontend command is recorded, and the
  connection is correctly rejected. Only binary data with small, self-consistent,
  chained lengths — i.e. genuine Postgres — passes the filter.

  ### Tests

  - **Unit test** (`bpf/tests/bpf_postgres_detection.c`): a host-compilable test
    for the classifier covering the three regression scenarios above plus the
    false-positive guards (HTTP, startup message, response-only segment,
    too-short buffer, invalid length).
  - **Integration test**: a new `/pipeline` endpoint on the Python Postgres test
    server uses psycopg pipeline mode to batch several extended-protocol
    statements into one TCP segment (more than 10 Postgres messages), reproducing
    the issue. `TestSuite_PythonPostgres` asserts a `SELECT accounting.invoices`
    trace appears — `accounting.invoices` is queried only by this endpoint, so its
    trace proves the multi-message connection was classified.

  ### Docs

  Adds `devdocs/protocols/tcp/postgres.md` documenting the (previously
  undocumented) Postgres parser and the detection heuristic.

  ## Validation

  - [ ] If this enhances / fixes / changes a core feature, I have updated the [features
  documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support
  matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed.

### DIFF
--- a/bpf/tests/bpf_postgres_detection.c
+++ b/bpf/tests/bpf_postgres_detection.c
@@ -50,7 +50,10 @@ enum protocol_type {
 };
 
 #define BPF_ANY 0
+
+#ifndef __always_inline
 #define __always_inline inline
+#endif
 
 static void bpf_probe_read(void *dst, u32 size, const void *src) {
     memcpy(dst, src, size);
--- a/bpf/tests/bpf_postgres_detection.c
+++ b/bpf/tests/bpf_postgres_detection.c
@@ -1,0 +1,326 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * The following code is copied from bpf/generictracer/protocol_postgres.h and
+ * adapted to run as a host unit test. The function under test is:
+ *
+ *   static __always_inline u8 is_postgres(connection_info_t *conn_info,
+ *                                         const unsigned char *data,
+ *                                         u32 data_len,
+ *                                         enum protocol_type *protocol_type);
+ *
+ * Together with its helper postgres_parse_hdr(). The BPF-only helpers
+ * (bpf_probe_read, bpf_ntohl, bpf_dbg_printk, bpf_map_update_elem and the
+ * protocol_cache map) are mocked below. The real header cannot be #included
+ * directly on a non-BPF target because its map definitions use SEC(".maps").
+ *
+ * These tests reproduce https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1464
+ * where request segments that concatenate several Postgres messages, exceed the
+ * per-segment message limit, or end with a message split across TCP segments
+ * were rejected because the parser required message_size == data_len.
+ */
+
+#include <arpa/inet.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+
+typedef uint8_t u8;
+typedef uint16_t u16;
+typedef uint32_t u32;
+typedef uint64_t u64;
+
+// ---------------------------------------------------------------------------
+// Mocks for the BPF runtime helpers used by is_postgres / postgres_parse_hdr.
+// ---------------------------------------------------------------------------
+
+typedef struct connection_info {
+    u32 src_ip;
+    u32 dst_ip;
+    u16 src_port;
+    u16 dst_port;
+} connection_info_t;
+
+enum protocol_type {
+    k_protocol_type_unknown = 0,
+    k_protocol_type_http,
+    k_protocol_type_postgres,
+};
+
+#define BPF_ANY 0
+#define __always_inline inline
+
+static void bpf_probe_read(void *dst, u32 size, const void *src) {
+    memcpy(dst, src, size);
+}
+
+#define bpf_ntohl(x) ntohl(x)
+
+// Discard the formatted debug output, mirroring bpf_dbg_printk when BPF debug
+// is disabled.
+#define bpf_dbg_printk(...) ((void)0)
+
+// protocol_cache and bpf_map_update_elem are irrelevant to the classification
+// decision, so they are reduced to no-ops here.
+static int protocol_cache;
+static long bpf_map_update_elem(void *map, const void *key, const void *value, u64 flags) {
+    (void)map;
+    (void)key;
+    (void)value;
+    (void)flags;
+    return 0;
+}
+
+// ---------------------------------------------------------------------------
+// Code under test (copied verbatim from protocol_postgres.h).
+// ---------------------------------------------------------------------------
+
+struct postgres_hdr {
+    u32 message_len;
+    u8 message_type;
+    u8 _pad[3];
+};
+
+enum {
+    k_pg_hdr_size = 5,
+    k_pg_messages_in_packet_max = 10,
+
+    k_pg_msg_bind = 'B',
+    k_pg_msg_execute = 'E',
+    k_pg_msg_parse = 'P',
+    k_pg_msg_query = 'Q',
+};
+
+static __always_inline struct postgres_hdr postgres_parse_hdr(const unsigned char *data) {
+    struct postgres_hdr hdr = {};
+
+    u8 header[k_pg_hdr_size] = {};
+    bpf_probe_read(header, k_pg_hdr_size, data);
+
+    u32 message_len_le;
+    __builtin_memcpy(&message_len_le, header + 1, sizeof(message_len_le));
+
+    hdr.message_type = header[0];
+    hdr.message_len = bpf_ntohl(message_len_le);
+
+    return hdr;
+}
+
+static __always_inline u8 is_postgres(connection_info_t *conn_info,
+                                      const unsigned char *data,
+                                      u32 data_len,
+                                      enum protocol_type *protocol_type) {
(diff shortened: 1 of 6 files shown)


bpf/tests/bpf_postgres_detection.c
```diff
@@ -0,0 +1,326 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * The following code is copied from bpf/generictracer/protocol_postgres.h and
```
mariomac, line 5: Test structure copied from https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/bpf/tests/bpf_kafka_large_buffer.c

Not a super fan of this formula but I guess we don't have yet a better way to  C tests without having to duplicate code (that will become eventually forgotten and differ from the original :-) )

bpf/tests/bpf_postgres_detection.c
```diff
@@ -0,0 +1,326 @@
...
+} connection_info_t;
+
+enum protocol_type {
+    k_protocol_type_unknown = 0,
+    k_protocol_type_http,
+    k_protocol_type_postgres,
+};
+
+#define BPF_ANY 0
+#define __always_inline inline
```
grcevski, line 55: This causes the test to fail to compile for me. I tried it like this:

```
make -C bpf/tests bpf_postgres_detection
```

output:
```
make: Entering directory '/home/nino/work/obi/bpf/tests'
clang -Wall -Werror -g -I. -I..  -o bpf_postgres_detection bpf_postgres_detection.c 
bpf_postgres_detection.c:53:9: error: '__always_inline' macro redefined [-Werror,-Wmacro-redefined]
   53 | #define __always_inline inline
      |         ^
/usr/include/x86_64-linux-gnu/sys/cdefs.h:441:10: note: previous definition is here
  441 | # define __always_inline __inline __attribute__ ((__always_inline__))
      |          ^
1 error generated.
make: *** [Makefile:11: bpf_postgres_detection] Error 1

```

grcevski (approved):

grcevski (approved):
LGTM! The change looks good to me, but I think we need to fix the testcase compile problem.

mmat11:
I'm a bit worried about false positives

> Only binary data with small, self-consistent,
chained lengths — i.e. genuine Postgres — passes the filter.

this can be true for other protocols as well, right?

bpf/tests/bpf_postgres_detection.c
mariomac: Fixed!

mariomac:
@mmat11 there is a chance that you fall within a `P` followed by an BigEndian int32 representing a small enough number. If I understand your concern, the previous implementation was safer against false positives because this `int32` after the operation byte had to coincide with the total message size, right?

Maybe we can check for extra data to be 99.99% safe against a false positive.

mmat11:
> @mmat11 there is a chance that you fall within a `P` followed by an BigEndian int32 representing a small enough number. If I understand your concern, the previous implementation was safer against false positives because this `int32` after the operation byte had to coincide with the total message size, right?
> 
> Maybe we can check for extra data to be 99.99% safe against a false positive.

I remember some cases of misclassification when we had an implementation that didn't check for total size. I guess we can try to go ahead with this and eventually add extra checks if we have real false positives

mmat11 (approved):

PR #2421, opened 2026-06-17
(DynamicSelector-Phase4) Propagate signal ownership through app discovery and lifecycle consumers

damemi:
This implements phase 3 of https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2234

Previous PRs:
1. https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2250
2. https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2279
3. https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2340

This PR wires up the actual per-signal ownership for AppO11y, Network, and Stats Metrics. AppO11y still uses a union view for app traces+metrics (will be split in the next PR). Dynamic/Parent pid tracking is added for `appolly/app/svc/svc.go` and `discover/matcher.go` with the `DynamicSelectorPID` field.

## Summary

Describe the change briefly.

## Validation

- [x] I have read and followed the [contributing guidelines](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/CONTRIBUTING.md)
- [x] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed.

PR #2340, opened 2026-06-12
(DynamicSelector-Phase3) Refactor DynamicPIDSelector into signal-aware views

damemi:
## Summary

This implements phase 3 of https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2234

Previous PRs:
1. https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2250
2. https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2279

This refactors the DynamicSelector backend implementation into signal specific views, introducing hierarchical interfaces that support the following uses:

```go
sel := discover.NewDynamicPIDSelector()
sel.Traces().AddPIDs(1234)           // traces only
sel.NetworkMetrics().AddPIDs(5678)   // network metrics only
sel.AddPIDs(9999)                    // all signals (legacy root behavior)
```

These signal specific paths are not enforced yet (everything is still global), they will be in the next phase

## Validation

- [ ] I have read and followed the [contributing guidelines](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/CONTRIBUTING.md)
- [ ] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed.

PR #2442, opened 2026-06-19
(DynamicSelector-Phase5) Gate AppO11y Traces and Metrics Separately

damemi:
## Summary

Implements phase 5 of https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2234

Previous PRs:
1. https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2250
2. https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2279
3. https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2340
4. https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2421

This adds the final signal split for app signals (metrics+traces) with 2 new filter nodes that are inserted in the trace and metrics pipelines

## Validation

### DIFF
--- a/pkg/appolly/dynamic_signal_gate.go
+++ b/pkg/appolly/dynamic_signal_gate.go
@@ -116,12 +116,13 @@ func (g *dynamicSignalProcessEventGate) handleProcessEvent(pe exec.ProcessEvent)
 	switch pe.Type {
 	case exec.ProcessEventCreated:
 		g.current[pid] = pe.File
+		if g.forwarded[pid] {
+			return
+		}
 		if g.selector.IncludesPID(processEventSignalPID(pe.File)) {
 			g.forwarded[pid] = true
 			g.output.Send(pe)
-			return
 		}
-		g.forwarded[pid] = false
 	case exec.ProcessEventTerminated:
 		if g.forwarded[pid] {
 			g.output.Send(pe)
--- a/pkg/appolly/dynamic_signal_gate.go
+++ b/pkg/appolly/dynamic_signal_gate.go
@@ -1,0 +1,162 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package appolly // import "go.opentelemetry.io/obi/pkg/appolly"
+
+import (
+	"context"
+
+	"go.opentelemetry.io/obi/pkg/appolly/app"
+	"go.opentelemetry.io/obi/pkg/appolly/app/request"
+	"go.opentelemetry.io/obi/pkg/appolly/discover/exec"
+	"go.opentelemetry.io/obi/pkg/pipe/msg"
+	"go.opentelemetry.io/obi/pkg/pipe/swarm"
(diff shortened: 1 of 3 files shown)


pkg/appolly/dynamic_signal_gate.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+	}
+}
+
+func (g *dynamicSignalProcessEventGate) handleProcessEvent(pe exec.ProcessEvent) {
+	pid := pe.File.Pid()
+	switch pe.Type {
+	case exec.ProcessEventCreated:
+		g.current[pid] = pe.File
+		if g.selector.IncludesPID(processEventSignalPID(pe.File)) {
```
grcevski, line 122: I think there might be slight timing hole here related to how Kubernetes handles event notifications. Sometimes they send multiple events for a pod, but maybe it's just theoretical problem.

So the scenario I'm thinking of is: a process was instrumented and we have metrics tracking for it. Then you remove it from selected pids and you send it async to RemovedNotify. Before you get to process RemovedNotify, something updates the pod and we get new ProcessCreatedEvent. What that would do now is make g.forwarded[pid] = false.

Since handleSelectorRemove does this check:

```
			if !g.forwarded[pid] || processEventSignalPID(file) != signalPID {
				continue
			}
```
I think it will never send the event.

pkg/appolly/instrumenter.go
```diff
@@ -115,13 +115,16 @@ func newGraphBuilder(
 	if exportableSpans == nil {
 		exportableSpans = msg2.QueueFromConfig[[]request.Span](config, "exportableSpans")
 	}
+	attrFilteredSpans := msg2.QueueFromConfig[[]request.Span](config, "attrFilteredSpans")
```
grcevski, line 118: I had a question about the new runtime metrics that were added. We have go for now, but JVM is coming in an approved PR. Do we plan to gate those in the future or is it omission in this PR.

damemi: that's a good question, tbh I haven't been following the runtime metrics as I've been so heads down in this design and I think I pulled this design together before the runtime metrics pipeline was in. But looking at it now that support could be added pretty easily imo just like we're doing with the other signal splits here. I would do it as a follow up though so we can wrap up this design

pkg/appolly/dynamic_signal_gate.go
damemi: nice catch, thank you. Updated this so that duplicate events shouldn't re-set the forwardedPid entry

pkg/appolly/instrumenter.go
grcevski: Yeah it's cool to be a follow-up, they are kind of new. Thanks!

grcevski (approved):
LGTM!

pkg/appolly/instrumenter.go
damemi: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2461 thanks for the suggestion!

Issue #1969, opened 2026-04-30
BPF application-level metrics: Add internal metrics for ringbuf writes


mariomac:
Analogous to Net O11y packet stats: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/bpf/netolly/flows.c#L173C1-L174C1

We could add a set of two metrics: total ringbuffer attempts and failed ringbuf writes. This would allow monitoring whether OBI ringbuf reader is not able to read/parse at the pace spans are produced.

cne27:
Hi @mariomac, I'd like to take this one.

I'll follow the existing NetO11y BPFPacketStats path: a per-CPU counter map alongside the events ringbuf, bumped at the write sites, then surfaced through imetrics.Reporter and both exporters. Naming would follow the general obi_bpf_* internal-metric convention, something like obi_bpf_ringbuf_writes_total and obi_bpf_ringbuf_write_failures_total.

One question before I start: should these land directly in otel dot format per #1965, or match the current obi_bpf_* Prometheus naming for now?

mariomac:
Amazing @cne27 ! Thank you for volunteering to contribute :)

Actually it should follow both formats, depending on the exporter. 

In Prometheus, we define them as `snake_case` here https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/export/imetrics/iprom.go#L72

In the OTEL internal metrics exporter, they are defined in `dot.format` here: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/export/otel/metrics_internal.go#L72

So the process is usually:
1- Add the new metric method to the `imetrics.Exporter` interface (and NoopReporter in the same file): https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/export/imetrics/imetrics.go#L62
2- Integrate the new method wherever is needed.
3- Add the new metric in both exporters
   - Prom: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/export/imetrics/iprom.go#L72
   - Otel: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/export/otel/metrics_internal.go#L72

cne27:
[...]

PR #2447, opened 2026-06-20
Add internal metrics for events ringbuffer writes

cne27:
## Summary

Part of #1969.

Under load, OBI's eBPF probes can produce span events faster than the userspace reader drains the `events` ringbuffer. When that happens the buffer fills and writes are silently dropped, with no signal that telemetry is being lost.

This adds two internal metrics that count writes to the `events` ringbuffer and how many of them failed because the buffer was full, so operators can detect when the reader is falling behind:

- Prometheus: `obi_bpf_ringbuf_writes_total`, `obi_bpf_ringbuf_write_failures_total`
- OTEL: `obi.bpf.ringbuf.writes.total`, `obi.bpf.ringbuf.write.failures.total`

## Scope

Only the generic tracer's `events` write sites are instrumented here; the metric descriptions say "from the generic tracer" so the counters aren't read as covering every write. The Go tracers write to the same shared map and can be wired through the same helper in a follow-up.

## Details

- A per-CPU map (`rb_write_stats`) holds attempted and failed write counts. It sits next to the `events` map in `bpf/common/ringbuf.h` and is `OBI_PIN_INTERNAL`, so all tracers share one instance.
- A helper, `account_ringbuf_write(failed)`, is called at each generic-tracer write site: `failed` is `bpf_ringbuf_reserve` returning `NULL`, or `bpf_ringbuf_output` returning non-zero.
- The internal-metrics scrape loop reads the map (folded into the existing per-map scan), sums the per-CPU counters, and forwards them through a new `BPFRingbufWriteStats` method on `imetrics.Reporter`.
- Both exporters implement it, converting the absolute BPF counters to deltas — the same approach as the existing `BPFPacketStats`.

Follows the process from #1969: new method on the `imetrics` interface (and `NoopReporter`), integrate it, then add the metric to both exporters in their respective formats (`snake_case` for Prometheus, `dot.format` for OTEL).

## Validation

- [x] N/A — internal self-metrics, not a core user-facing feature; the existing internal metrics aren't documented in features.md / SUPPORT_MATRIX, so nothing to update there

Built and tested locally (Fedora, podman):
- `make docker-generate`, `make compile`
- `make test` (full `-short -race ./...` — all packages green, incl. the new ringbuf reporter tests)
- `make lint`
- `clang-format` / `clang-tidy` on the changed `bpf/` files

### DIFF
--- a/bpf/common/ringbuf.h
+++ b/bpf/common/ringbuf.h
@@ -27,7 +27,6 @@ typedef struct ringbuf_write_count_t {
     u64 failed;
 } ringbuf_write_count;
 
-// Per-CPU counters of attempted and failed writes to the events ringbuffer.
 struct {
     __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
     __type(key, u32);
--- a/bpf/common/ringbuf.h
+++ b/bpf/common/ringbuf.h
@@ -28,7 +28,6 @@ typedef struct ringbuf_write_count_t {
 } ringbuf_write_count;
 
 // Per-CPU counters of attempted and failed writes to the events ringbuffer.
-// OBI_PIN_INTERNAL, like events itself, so all tracers share one instance.
 struct {
     __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
     __type(key, u32);
--- a/bpf/common/ringbuf.h
+++ b/bpf/common/ringbuf.h
@@ -22,6 +22,21 @@ struct {
     __uint(pinning, OBI_PIN_INTERNAL);
 } events SEC(".maps");
 
+typedef struct ringbuf_write_count_t {
+    u64 total;
+    u64 failed;
+} ringbuf_write_count;
+
+// Per-CPU counters of attempted and failed writes to the events ringbuffer.
+// OBI_PIN_INTERNAL, like events itself, so all tracers share one instance.
+struct {
+    __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+    __type(key, u32);
+    __type(value, ringbuf_write_count);
+    __uint(max_entries, 1);
+    __uint(pinning, OBI_PIN_INTERNAL);
+} rb_write_stats SEC(".maps");
+
 // To be Injected from the user space during the eBPF program load & initialization
 volatile const u32 wakeup_data_bytes;
 
@@ -36,4 +51,19 @@ static __always_inline long get_flags() {
 
     const u64 sz = bpf_ringbuf_query(&events, BPF_RB_AVAIL_DATA);
     return sz >= wakeup_data_bytes ? BPF_RB_FORCE_WAKEUP : BPF_RB_NO_WAKEUP;
+}
+
+// Records one events-ringbuffer write attempt, flagged failed when the
+// reserve/output did not reach userspace. Call after each write on &events.
+static __always_inline void account_ringbuf_write(bool failed) {
+    const u32 key = 0;
+    ringbuf_write_count *stats = bpf_map_lookup_elem(&rb_write_stats, &key);
+    if (!stats) {
+        return;
+    }
+
+    stats->total++;
+    if (failed) {
+        stats->failed++;
+    }
 }
--- a/bpf/generictracer/protocol_common.h
+++ b/bpf/generictracer/protocol_common.h
@@ -54,7 +54,9 @@ static __always_inline u32 large_buf_emit_chunks(tcp_large_buffer_t *large_buf,
         u32 total_size = sizeof(tcp_large_buffer_t) + payload_size;
         bpf_clamp_umax(total_size, k_large_buf_max_size);
 
-        if (bpf_ringbuf_output(&events, large_buf, total_size, get_flags()) != 0) {
+        const long rb_err = bpf_ringbuf_output(&events, large_buf, total_size, get_flags());
+        account_ringbuf_write(rb_err != 0);
+        if (rb_err != 0) {
             break;
         }
 
--- a/pkg/export/imetrics/iprom.go
+++ b/pkg/export/imetrics/iprom.go
@@ -54,6 +54,11 @@ type PrometheusReporter struct {
 	bpfIgnoredPacketCount prometheus.Counter
 
 	queueCapacityRatio *prometheus.GaugeVec
+
+	totalRingbufWrites        uint64
+	totalRingbufWriteFailures uint64
+	bpfRingbufWriteCount      prometheus.Counter
+	bpfRingbufWriteFailures   prometheus.Counter
 }
 
 func NewPrometheusReporter(cfg *InternalMetricsConfig, manager *connector.PrometheusManager, registry *prometheus.Registry) *PrometheusReporter {
@@ -145,6 +150,14 @@ func NewPrometheusReporter(cfg *InternalMetricsConfig, manager *connector.Promet
 			Name: attr.VendorPrefix + "_queue_capacity_ratio",
 			Help: "Ratio [0-1] between the unread messages of an internal Go channel and its total capacity",
 		}, []string{"subscriber"}),
+		bpfRingbufWriteCount: prometheus.NewCounter(prometheus.CounterOpts{
+			Name: attr.VendorPrefix + "_bpf_ringbuf_writes_total",
+			Help: "How many writes to the events ringbuffer from the generic tracer have been attempted",
+		}),
+		bpfRingbufWriteFailures: prometheus.NewCounter(prometheus.CounterOpts{
+			Name: attr.VendorPrefix + "_bpf_ringbuf_write_failures_total",
+			Help: "How many writes to the events ringbuffer from the generic tracer failed because the buffer was full",
+		}),
 	}
 	if !cfg.AvoidedServices.Disabled {
 		pr.avoidedServicesLimiter = avoidedsvc.NewLimiter(cfg.AvoidedServices.Limit)
@@ -176,6 +189,8 @@ func NewPrometheusReporter(cfg *InternalMetricsConfig, manager *connector.Promet
 		pr.bpfPacketCount,
 		pr.bpfIgnoredPacketCount,
 		pr.queueCapacityRatio,
+		pr.bpfRingbufWriteCount,
+		pr.bpfRingbufWriteFailures,
 	}
 	if pr.avoidedServices != nil {
 		metrics = append(metrics, pr.avoidedServices)
@@ -278,4 +293,10 @@ func (p *PrometheusReporter) BPFPacketStats(count, ignored uint64) {
 
 func (p *PrometheusReporter) QueueBufferUtilization(subscriber string, ratio float64) {
 	p.queueCapacityRatio.WithLabelValues(subscriber).Set(ratio)
+}
+
+func (p *PrometheusReporter) BPFRingbufWriteStats(total, failed uint64) {
+	p.bpfRingbufWriteCount.Add(float64(total - p.totalRingbufWrites))
+	p.bpfRingbufWriteFailures.Add(float64(failed - p.totalRingbufWriteFailures))
+	p.totalRingbufWrites, p.totalRingbufWriteFailures = total, failed
 }
--- a/pkg/export/imetrics/iprom.go
+++ b/pkg/export/imetrics/iprom.go
@@ -152,7 +152,7 @@ func NewPrometheusReporter(cfg *InternalMetricsConfig, manager *connector.Promet
 		}, []string{"subscriber"}),
 		bpfRingbufWriteCount: prometheus.NewCounterVec(prometheus.CounterOpts{
 			Name: attr.VendorPrefix + "_bpf_ringbuf_writes_total",
-			Help: "How many writes to the ringbuffer from the generic tracer have been attempted",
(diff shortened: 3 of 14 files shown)


bpf/common/ringbuf.h
```diff
@@ -22,6 +22,21 @@ struct {
     __uint(pinning, OBI_PIN_INTERNAL);
 } events SEC(".maps");
 
+typedef struct ringbuf_write_count_t {
+    u64 total;
+    u64 failed;
+} ringbuf_write_count;
+
+// Per-CPU counters of attempted and failed writes to the events ringbuffer.
+// OBI_PIN_INTERNAL, like events itself, so all tracers share one instance.
```
pinoOgni: I think the second line of the comment is not needed.

pkg/export/prom/prom_bpf.go
```diff
@@ -196,10 +200,41 @@ func (bc *BPFCollector) collectInternalMetrics(ctx context.Context) {
...
 			}
+
+			if ringbufWritesOK {
+				bc.ctxInfo.Metrics.BPFRingbufWriteStats(ringbufWrites.Total, ringbufWrites.Failed)
+			}
 		}
 	}
 }
 
+// BPF map names are truncated to 15 chars in the kernel; keep within that.
```
pinoOgni: Same here.

pkg/export/imetrics/iprom.go
```diff
@@ -145,6 +150,14 @@ func NewPrometheusReporter(cfg *InternalMetricsConfig, manager *connector.Promet
 			Name: attr.VendorPrefix + "_queue_capacity_ratio",
 			Help: "Ratio [0-1] between the unread messages of an internal Go channel and its total capacity",
 		}, []string{"subscriber"}),
+		bpfRingbufWriteCount: prometheus.NewCounter(prometheus.CounterOpts{
```
pinoOgni: The metric names `obi_bpf_ringbuf_writes_total` and `obi_bpf_ringbuf_write_failures_total` are scoped to the `events` ringbuffer today, but OBI has other ringbufs (`stats_events` for example) that could be instrumented the same way later. The current names read as covering all ringbufs (even if in the description it says "generic tracer"), which is misleading.

What about adding a `ringbuf` label to both metrics? 
For example `obi_bpf_ringbuf_writes_total{ringbuf="events"}`? 

I think it is easy to add it.

pinoOgni (changes requested):

cne27:
Thanks @pinoOgni, good call on the label. Added `ringbuf="events"` to both metrics (Prometheus and OTEL) and dropped the two comments, in 9f6542e3.

pkg/export/imetrics/iprom.go
```diff
@@ -145,6 +150,16 @@ func NewPrometheusReporter(cfg *InternalMetricsConfig, manager *connector.Promet
 			Name: attr.VendorPrefix + "_queue_capacity_ratio",
 			Help: "Ratio [0-1] between the unread messages of an internal Go channel and its total capacity",
 		}, []string{"subscriber"}),
+		bpfRingbufWriteCount: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: attr.VendorPrefix + "_bpf_ringbuf_writes_total",
+			Help: "How many writes to the ringbuffer from the generic tracer have been attempted",
```
pinoOgni:
```suggestion
			Help: "How many writes to the named ringbuffer have been attempted",
```

pkg/export/imetrics/iprom.go
```diff
@@ -145,6 +150,16 @@ func NewPrometheusReporter(cfg *InternalMetricsConfig, manager *connector.Promet
...
+		}, []string{"ringbuf"}),
+		bpfRingbufWriteFailures: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: attr.VendorPrefix + "_bpf_ringbuf_write_failures_total",
+			Help: "How many writes to the ringbuffer from the generic tracer failed because the buffer was full",
```
pinoOgni:
```suggestion
			Help: "How many writes to the named ringbuffer failed because the buffer was full",
```

pkg/export/otel/metrics_internal.go
```diff
@@ -217,6 +222,24 @@ func NewInternalMetricsReporter(ctx context.Context, ctxInfo *global.ContextInfo
 		return nil, err
 	}
 
+	bpfRingbufWriteCount, err := meter.Int64Counter(
+		attr.VendorPrefix+".bpf.ringbuf.writes.total",
+		instrument.WithDescription("How many writes to the ringbuffer from the generic tracer have been attempted"),
```
pinoOgni:
```suggestion
		instrument.WithDescription("How many writes to the named ringbuffer have been attempted"),
```

pkg/export/otel/metrics_internal.go
```diff
@@ -217,6 +222,24 @@ func NewInternalMetricsReporter(ctx context.Context, ctxInfo *global.ContextInfo
...
+		instrument.WithUnit("{write}"),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	bpfRingbufWriteFailures, err := meter.Int64Counter(
+		attr.VendorPrefix+".bpf.ringbuf.write.failures.total",
+		instrument.WithDescription("How many writes to the ringbuffer from the generic tracer failed because the buffer was full"),
```
pinoOgni:
```suggestion
		instrument.WithDescription("How many writes to the named ringbuffer failed because the buffer was full"),
```

pinoOgni (approved):
Thanks! 
Just some nits.

cne27:
Thanks for the review @pinoOgni, much appreciated. The CI workflows seem to be waiting on a maintainer to approve them for the latest commit. Could you or @mariomac approve the run when you get a chance? Happy to address anything they surface.

bpf/common/ringbuf.h
```diff
@@ -22,6 +22,20 @@ struct {
     __uint(pinning, OBI_PIN_INTERNAL);
 } events SEC(".maps");
 
+typedef struct ringbuf_write_count_t {
+    u64 total;
+    u64 failed;
+} ringbuf_write_count;
+
+// Per-CPU counters of attempted and failed writes to the events ringbuffer.
```
rafaelroquetto: nit: remove (it states the obvious)

non-nit: even better - use `SCRATCH_MEM` (see `scratch_mem.h`)

pkg/export/prom/prom_bpf.go
```diff
@@ -36,6 +36,10 @@ type BPFCollector struct {
 	probeMetrics     func() []ProbeMetrics
 	mapMetrics       func() []BpfMapMetrics
 	mu               sync.Mutex
+
+	// populated by getMapMetrics under mu, as a side effect of the map scan
```
rafaelroquetto: please remove - these kind of comments add little value and are easy to drift, as they reference code

pkg/export/prom/prom_bpf.go
```diff
@@ -196,10 +200,40 @@ func (bc *BPFCollector) collectInternalMetrics(ctx context.Context) {
...
+			if ringbufWritesOK {
+				bc.ctxInfo.Metrics.BPFRingbufWriteStats("events", ringbufWrites.Total, ringbufWrites.Failed)
+			}
 		}
 	}
 }
 
+const ringbufWriteStatsMapName = "rb_write_stats"
+
+type ringbufWriteCount struct {
```
rafaelroquetto, line 212: this is duplicating `ringbuf_write_count_t` - we should use that instead

rafaelroquetto (changes requested):

cne27:
Thanks @rafaelroquetto. I removed both comments. I tried the other two suggestions and each hit a blocker here, so I wanted to check direction with you.

On the generated type: `prom` cannot import `generictracer` (where `BpfRingbufWriteCount` lives) without creating an import cycle. `generictracer` imports `pkg/obi`, and `pkg/obi` imports `pkg/export/prom` for `PrometheusConfig`, so `prom` importing `generictracer` closes the loop and does not compile. That cycle is why the value struct is currently defined locally in the collector. To actually use the generated type we would need to break the `obi` to `prom` dependency, for example by moving the Prometheus config types out of `pkg/export/prom` into a config package that `obi` imports instead. Doable, but a wider refactor than this PR.

On `SCRATCH_MEM`: it declares the map with `value_size` rather than a typed value, so `bpf2go` stops emitting `BpfRingbufWriteCount` (a `-type` flag to recover it fails with `collect C types: not found`, since the struct is no longer in BTF). So it works against the generated type suggestion above. It also drops the `OBI_PIN_INTERNAL` pin that keeps this map as a single shared instance across the tracers that load one copy per executable, the same way the `events` map it shadows does. Without the pin it becomes one map per executable that the collector has to sum.

Given that, I kept the dedicated pinned map and the local struct for now. If you would prefer one of the alternatives, the config refactor to enable the generated type, or `SCRATCH_MEM` with a local struct and summing across the per executable maps, I am happy to take it on. What works best for you?

Separately, the one red check is `Kernel bpf-next`, which hung inside `ProgLoad` and hit the test timeout rather than failing verification. All stable kernels from 5.10 through 6.18 pass the same verifier test, and the subtests running at the time were `netolly` programs unrelated to this change, so it looks like a `bpf-next` flake. Could you re-run that job when you get a chance?

bpf/common/ringbuf.h
```diff
@@ -37,3 +50,18 @@ static __always_inline long get_flags() {
     const u64 sz = bpf_ringbuf_query(&events, BPF_RB_AVAIL_DATA);
     return sz >= wakeup_data_bytes ? BPF_RB_FORCE_WAKEUP : BPF_RB_NO_WAKEUP;
 }
+
+// Records one events-ringbuffer write attempt, flagged failed when the
+// reserve/output did not reach userspace. Call after each write on &events.
+static __always_inline void account_ringbuf_write(bool failed) {
```
grcevski, line 56: This shouldn't be called by default. It needs to be guarded by an option to enable this extra tracking only if internal metrics are enabled.

bpf/generictracer/protocol_common.h
```diff
@@ -54,7 +54,9 @@ static __always_inline u32 large_buf_emit_chunks(tcp_large_buffer_t *large_buf,
         u32 total_size = sizeof(tcp_large_buffer_t) + payload_size;
         bpf_clamp_umax(total_size, k_large_buf_max_size);
 
-        if (bpf_ringbuf_output(&events, large_buf, total_size, get_flags()) != 0) {
+        const long rb_err = bpf_ringbuf_output(&events, large_buf, total_size, get_flags());
```
grcevski, line 57: This needs to be combined in a function or a macro that will be used everywhere to track the failure to write to the ring buffer along with the write. Having it separate means that we'll likely miss some next time someone adds a new reserve on the ring buffer.

grcevski (changes requested):

rafaelroquetto:
@cne27 

I believe the fact you are running into a dependency cycle is a side-effect is an alert that the actual design is not correct. You are trying to access an ebpf object (a map in this case) from a type that should not be interfacing with ebpf at first place. Only tracers are allowed to interface with ebpf object _values_ - actually, this is further evidenced by the fact that you need to compare a map by name - so half of the functionality is in generictracer and half is in BPFCollector - breaking orthogonality and yielding the cyclic dependency.

`BPFCollector` does _not_ interface with ebpf maps or programs in that sense - look closely, it simply iterates the maps from userspace to gather the entry count and other *metadata*, but it does not try to access those _entries_ nor does it need to know what the types are.

You will need to refactor this to go thru generictracer - or as a matter of fact, via any tracer - I assume that we want to collect stats even when generictracer is not loaded, right?

The problem is that we don't have such an infrastructure yet, so basically, to get this right, you will need to come up with a design that:

 - lives at the same layer of the tracers and pushes data to upper layers
 - allows for using auto generated ebpf structs - having to redefine ebpf types in userspace manually is a no-go
 
To give you an example, this is more or less what `BPFPacketStats` in `MapTracer` does - the tracer pulls the map values, and pushes it up to imetrics. The tracer has access to the map types, etc. It's indeed a larger refactor, but it is what it is. We unfortunately can't merge the current design as is.

re the `SCRATCH_MEM` case, you can always change the macro to include OBI_PIN_INTERNAL.

I'd like to take this opportunity to highlight our [AI Policy](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/AI-POLICY.md).

cne27:
Thanks @rafaelroquetto, that clarified a lot. I'll dig into `BPFPacketStats` in `MapTracer` and redesign around that pattern. Heads up I'm new to the repo, so it may take me a little extra time. I'll follow up with an update.

marctc:
> `BPFCollector` does _not_ interface with ebpf maps or programs in that sense - look closely, it simply iterates the maps from userspace to gather the entry count and other _metadata_, but it does not try to access those _entries_ nor does it need to know what the types are.

Yeah, we can't add those metrics there. This collector is meant to be used to periodically fetch some stats exposed via `cillium/ebpf` library to gather information from maps and probes. I checked their code and they don't expose anything related to ringbufs. I suggest to look at their code/project and perhaps there's something there you can figure out how to solve this problem in deeper way. 

As Rafael mentioned, the design is not correct, so to avoid further reviews here (like me right now) I'm closing the PR. Please reopen when you come up with a better solution. Thanks!

PR #2458, opened 2026-06-22
(DynamicSelector) Support dynamic resource attributes

damemi:
## Summary

This adds support for custom, dynamic attributes through the DynamicSelector API. For example:

```go
sel.Traces().AddPID(1234, selection.DynamicPIDOptions{
    ServiceName:      "checkout-api",
    ServiceNamespace:   "production",
    ResourceAttributes: map[string]string{
        "deployment.environment": "prod",
        "service.version":        "1.2.3",
        "team":                   "payments",
    },
})
```

New dynamic selector methods `GetPID()` and `SetPID()` are intended to be used to post-instrumentation checks and updates.

Resource attributes are *per-pid*, not per-signal, meaning that while you can use the individual signal views to set and modify a process's attributes (like above with `sel.Traces().AddPID()`) those attributes are intentionally shared among all signals that the PID is enabled for. This is for 2 reasons:

1. Technical: it fits will into the global signal bitmask approach implemented by https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2234, meaning that individual signals don't need to share unique copies of attribute maps
2. Ideological: The point of OTel is to make cross-signal correlations based on resource attributes. There isn't a good reason to support unique attributes (that can't be handled by a downstream collector)

This works by essentially fleshing out the existing DynamicSelector adapter that was used in the AppO11y pipeline to retrieve any set values as stored in the selector.

## Validation

### DIFF
--- a/pkg/appolly/discover/dynamic_pid_selector.go
+++ b/pkg/appolly/discover/dynamic_pid_selector.go
@@ -5,11 +5,13 @@ package discover // import "go.opentelemetry.io/obi/pkg/appolly/discover"
 
 import (
 	"iter"
+	"maps"
 	"slices"
 	"sync"
 
 	"go.opentelemetry.io/obi/pkg/appolly/app"
 	"go.opentelemetry.io/obi/pkg/appolly/services"
+	attr "go.opentelemetry.io/obi/pkg/export/attributes/names"
 	"go.opentelemetry.io/obi/pkg/export/otel/perapp"
 	"go.opentelemetry.io/obi/pkg/selection"
 )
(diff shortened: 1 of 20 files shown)


grcevski:
The kernel next bpf verifier tests can be fixed with merge with main now.

pkg/appolly/discover/dynamic_pid_selector.go
```diff
@@ -268,13 +309,53 @@ func (d *DynamicPIDSelector) getPIDs(mask dynamicPIDSignal) ([]app.PID, bool) {
...
+	defer d.mu.RUnlock()
+	rec, ok := d.byPID[app.PID(pid)]
+	if !ok {
+		return selection.DynamicPIDEntry{}, false
+	}
+	return entryFromRecord(app.PID(pid), rec.attrs), true
+}
+
+// SetPID updates shared attributes for a PID that is already tracked by the selector.
+func (d *DynamicPIDSelector) SetPID(entry selection.DynamicPIDEntry) bool {
```
grcevski, line 396: If I understand this correctly, this is only done once when the PID is instrumented. There's the possibility on Kubernetes to get a metadata update sometimes, e.g. the Kubernetes will give us one pod info, then we'll instrument the process, and then they'll ship us an update. In this scenario I think the service attributes will not be updated to reflect the updated state.

damemi: Thanks @grcevski, I didn't understand the role that the FileInfo played, if I understand it that is basically the source of truth post-instrumentation in the pipeline now right? So SetPID not updating that defeats the point of the function. I updated this to have SetPID update the fileinfo

I also made some changes to wire similar decoration into the net and stats metrics too

grcevski (approved):
LGTM! Nice!

PR #2462, opened 2026-06-22
Import config v2 capture rules

MrAlias:
## Summary

Implements the first review-sized slice of step 6, "Expand import conversion parity", from #2251.

This PR is intentionally scoped to capture rule import parity:

- import advanced capture rule selectors into runtime discovery selectors
- support glob and regex process selectors, Kubernetes namespace/metadata/label/annotation selectors, target PIDs, ports, and containers-only matching
- import OTLP-export exclusion rules back to `ExcludeOTelInstrumentedServices` and `DefaultOtlpGRPCPort`
- import per-rule export refinements and direction-aware HTTP route refinements

This stays internal-only and does not add CLI support, runtime v2 loading, receiver support, generated docs, or public config v2 APIs.

## Non-goals

To keep this under the review-size limit in #2251, the following step-6 areas are left for follow-up PRs:

- document-level resource/tracer_provider/meter_provider import
- global HTTP route, payload extraction, and enrichment import
- enrichment attribute and Kubernetes import
- network capture/stats import parity
- mechanical round-trip/audit coverage for the remaining import/export gaps

## Tests

### DIFF
--- a/internal/config/convert/import.go
+++ b/internal/config/convert/import.go
@@ -4,6 +4,7 @@
 package convert // import "go.opentelemetry.io/obi/internal/config/convert"
 
 import (
+	"fmt"
 	"reflect"
 	"slices"
 	"strings"
@@ -33,6 +34,9 @@ var v2StatsMetricsFeatureMask = export.FeatureStats
 // runtime loading.
 func V2ToRuntime(src *schema.Extension) (*obi.Config, error) {
 	if err := schema.ValidateStandalone(src); err != nil {
+		return nil, err
+	}
+	if err := validateV2RulePatterns(src.Capture.Rules); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


rafaelroquetto (approved):

internal/config/convert/import.go
```diff
@@ -58,6 +60,7 @@ func runtimeConfigDefaults() obi.Config {
 
 func applyV2Capture(cfg *obi.Config, src *schema.Extension) {
 	applyV2Policy(cfg, src.Capture.Policy, completePolicy(src.Capture.Policy))
+	applyV2Rules(cfg, src.Capture.Rules)
```
grcevski, line 67: So if I understand correctly, in v2 we have the capture policy, default action to be include, so if I wrote a policy like this:

```
capture:
  policy:
    default_action: include
    match_order: first_match_wins
  rules:
    - action: exclude
      match:
        process:
          exe_path_glob: ["*/obi", "obi"]
```

I think the default action include is not translated, so we will end up with:

```
cfg.Discovery.Instrument = nil
cfg.Discovery.Services = nil
cfg.Discovery.ExcludeInstrument = []...{obi exclude}
```

internal/config/convert/import.go
```diff
@@ -88,6 +91,270 @@ func applyV2Policy(cfg *obi.Config, policy schema.CapturePolicy, complete bool)
...
+	applyRuntimeDiscoveryRules(cfg, runtimeDiscoveryRulesFromV2(rules))
+}
+
+func runtimeDiscoveryRulesFromV2(rules []schema.Rule) runtimeDiscoveryRules {
+	var converted runtimeDiscoveryRules
+	for _, rule := range rules {
+		if collectV2ExportsOTLPExclusionRule(&converted, rule) {
+			continue
+		}
+		globSelector, regexSelector, ok := selectorFromRule(rule)
```
grcevski, line 121: I wonder if we should bother with regex for the conversion, or just always do glob. The reason I'm thinking we should just do glob is because in the matcher at the moment, you can either go with regex or glob, you can't mix and match them. I think the code doesn't work if you do if I remember correctly. But the converter treats them as they are both possible to be defined.

internal/config/convert/import.go
```diff
@@ -88,6 +91,270 @@ func applyV2Policy(cfg *obi.Config, policy schema.CapturePolicy, complete bool)
...
+	}
+	return *in
+}
+
+func globAttr(values []string) services.GlobAttr {
+	switch len(values) {
+	case 0:
+		return services.GlobAttr{}
+	case 1:
+		return services.NewGlob(values[0])
```
grcevski: So newGlob calls MustCompile, which will panic if the value was wrong. I think we should maybe check for malformed pattern and return an error? I don't know if it's possible at this level.

grcevski (commented):
This looks great, I had some comments/questions, but it's more that perhaps my understanding is incomplete. I think I'm certain about the crash on NewGlob, but I want to see if I understood the implication of the others.

internal/config/convert/import.go
MrAlias: Yes, that's the intended runtime representation. The v1 runtime config does not have a separate "default include" field; include-by-default is represented by empty positive selectors (`Discovery.Instrument` / `Discovery.Services`) plus the exclusion selectors. So this imports to "include everything except the `obi` patterns".

I added `TestV2ToRuntimeDefaultIncludeUsesOnlyExclusions` in `aa5cf80` to make that expectation explicit.

internal/config/convert/import.go
MrAlias: I think we should keep regex support for import parity. `RuntimeToV2` can emit regex-backed rules from the runtime regex selector surface (notably the legacy `discovery.services` / `exclude_services` path), and converting those to glob would not be equivalent.

The importer does avoid the unsupported mixed-selector case: if a single v2 rule contains both glob and regex fields, `selectorFromRule` skips it instead of building a runtime selector that cannot work correctly. `TestV2ToRuntimeSkipsMixedGlobRegexRules` covers that.

internal/config/convert/import.go
MrAlias: Agreed, this was actionable. Fixed in `aa5cf80`: `V2ToRuntime` now validates rule glob and regex patterns through error-returning parsing before applying them, so malformed patterns return a conversion error instead of reaching the `MustCompile` constructors.

I added `TestV2ToRuntimeRejectsMalformedRulePatterns` for both glob and regex cases.

grcevski (approved):
LGTM!

PR #2465, opened 2026-06-22
Import config v2 document telemetry sections

MrAlias:
Depends on #2462.

## What

This adds an internal `DocumentToRuntime` entry point for standalone config v2 documents. It starts from the existing OBI extension import path, then applies the document-level OpenTelemetry sections emitted by `RuntimeToV2`.

The document import covers resource host overrides, tracer sampler and batch OTLP gRPC settings, meter periodic OTLP gRPC settings, default histogram aggregation, and the Prometheus development pull port.

Unsupported provider shapes are intentionally skipped instead of being partially interpreted, keeping the importer scoped to the exported config shape.

## Validation

- `make generate`

### DIFF
--- a/internal/config/convert/import_document.go
+++ b/internal/config/convert/import_document.go
@@ -72,7 +72,7 @@ func stringResourceAttribute(attr otelconfx.AttributeNameValue) (string, bool) {
 }
 
 func applyV2TracerProvider(cfg *obi.Config, provider *otelconfx.TracerProvider) {
-	if provider == nil {
+	if provider == nil || !exportedTracerProvider(provider) {
 		return
 	}
 
@@ -98,6 +98,11 @@ func applyV2TracerProvider(cfg *obi.Config, provider *otelconfx.TracerProvider) 
 		cfg.Traces.TracesEndpoint = *batch.Exporter.OTLPGrpc.Endpoint
 	}
 	cfg.Traces.TracesProtocol = otelcfg.ProtocolGRPC
+}
+
+func exportedTracerProvider(provider *otelconfx.TracerProvider) bool {
+	return provider.Limits == nil &&
+		provider.TracerConfiguratorDevelopment == nil
 }
 
 func exportedBatchSpanProcessor(provider *otelconfx.TracerProvider) (*otelconfx.BatchSpanProcessor, bool) {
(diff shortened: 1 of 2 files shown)


internal/config/convert/import_document.go
```diff
@@ -0,0 +1,323 @@
...
+		exporter.AdditionalProperties != nil ||
+		!exportedOTLPGrpcExporter(exporter.OTLPGrpc) {
+		return nil, false
+	}
+
+	return processor.Batch, true
+}
+
+func exportedOTLPGrpcExporter(exporter *otelconfx.OTLPGrpcExporter) bool {
+	return exporter.Compression == nil &&
```
grcevski, line 133: should we also check for TLS? Same on line 268...

internal/config/convert/import_document.go
```diff
@@ -0,0 +1,323 @@
...
+func applyV2TracerProvider(cfg *obi.Config, provider *otelconfx.TracerProvider) {
+	if provider == nil {
+		return
+	}
+
+	if sampler, ok := samplerConfigFromV2(provider.Sampler); ok {
+		cfg.Traces.SamplerConfig = sampler
+	}
+
+	batch, ok := exportedBatchSpanProcessor(provider)
```
grcevski, line 83: Maybe we should make the partial import of tracer_provider.limits, tracer_configurator/development, meter_provider.views, exemplar_filter etc. explicit in tests?

internal/config/convert/import_document.go
MrAlias: Done in a83c48707d: the importer now treats OTLP gRPC TLS as part of the supported exporter shape. It accepts the `tls.insecure` form emitted by `RuntimeToV2`, and rejects unsupported CA/cert/key TLS fields for both trace and metric exporters so we do not partially import those exporter shapes.

internal/config/convert/import_document.go
MrAlias: Done in a83c48707d: I added explicit tests for unsupported top-level provider fields. `tracer_provider.limits`, `tracer_configurator/development`, `meter_provider.exemplar_filter`, `meter_configurator/development`, and `meter_provider.views` now keep the runtime defaults instead of being partially imported.

grcevski (approved):
LGTM!

PR #2467, opened 2026-06-22
Import config v2 HTTP settings

MrAlias:
Depends on #2462

## What changed
- Import v2 global HTTP route configuration and route harvester discovery settings into the runtime config.
- Import HTTP payload extraction memberships, SQL++ endpoint patterns, and HTTP enrichment policy/rules from the v2 shape.
- Import application HTTP filters when the exported trace and metric filter maps match.

## Validation
- `make generate`

### DIFF
--- a/internal/config/convert/import.go
+++ b/internal/config/convert/import.go
@@ -623,9 +623,7 @@ func applyV2Instrumentation(cfg *obi.Config, instrumentation schema.Instrumentat
 }
 
 func applyFullV2Instrumentation(cfg *obi.Config, instrumentation schema.Instrumentation) {
-	cfg.EBPF.TrackRequestHeaders = instrumentation.HTTP.TrackRequestHeaders
-	cfg.EBPF.HTTPRequestTimeout = instrumentation.HTTP.RequestTimeout.TimeDuration()
-	cfg.EBPF.BufferSizes.HTTP = instrumentation.HTTP.BufferSize
+	applyFullV2HTTPInstrumentation(cfg, instrumentation.HTTP)
 
 	cfg.EBPF.HeuristicSQLDetect = instrumentation.SQL.HeuristicDetect
 	cfg.EBPF.BufferSizes.MySQL = instrumentation.SQL.MySQL.BufferSize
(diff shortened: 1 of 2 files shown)


internal/config/convert/import.go
```diff
@@ -713,6 +703,210 @@ func applyPartialV2Instrumentation(cfg *obi.Config, instrumentation schema.Instr
...
+		case payloadExtractorRerank:
+			http.GenAI.Rerank.Enabled = true
+		case payloadExtractorRetrieval:
+			http.GenAI.Retrieval.Enabled = true
+		case payloadExtractorJSONRPC:
+			http.JSONRPC.Enabled = true
+		case payloadExtractorEnrichment:
+			http.Enrichment.Enabled = true
+		}
+	}
```
grcevski, line 942: Should we fail here on unknown supplied string in enabled?

internal/config/convert/import.go
```diff
@@ -713,6 +703,210 @@ func applyPartialV2Instrumentation(cfg *obi.Config, instrumentation schema.Instr
...
+	applyV2HTTPFilters(cfg, http.Filters, false)
+	applyPartialV2HTTPRoutes(cfg, http.Routes)
+	applyPartialV2HTTPPayloadExtraction(cfg, http.PayloadExtraction)
+}
+
+func applyV2HTTPFilters(cfg *obi.Config, filters schema.SignalFilters, complete bool) {
+	if zeroValue(filters) && !complete {
+		return
+	}
+	if !reflect.DeepEqual(filters.Traces, filters.Metrics) {
```
grcevski: If I understand correctly, the config will allow you to just define `http.filters.traces` without the metrics, but I think this will not apply anything unless both are defined, which I think it's a silent failure.

MrAlias: Handled in f27771ea3f: one-sided HTTP filters now import the supplied trace or metric map, and conflicting trace/metric maps fail validation instead of silently no-oping.

internal/config/convert/import.go
MrAlias: Handled in f27771ea3f: v2 runtime import now validates `payload_extraction.enabled` and returns a precise config-path error for unknown extractor names.

grcevski (approved):
LGTM!

PR #2468, opened 2026-06-22
Import v2 enrichment config

MrAlias:
Depends on #2462

## Summary

- Import standalone `enrich.attributes` into runtime attribute settings.
- Import `enrich.enrichers.kubernetes` fields exported by `RuntimeToV2`.
- Add focused round-trip and default-preservation tests for enrichment import parity.

## Validation

- `make generate`

### DIFF
--- a/internal/config/convert/import.go
+++ b/internal/config/convert/import.go
@@ -13,6 +13,7 @@ import (
 	"go.opentelemetry.io/obi/pkg/appolly/services"
 	obiconfig "go.opentelemetry.io/obi/pkg/config"
 	"go.opentelemetry.io/obi/pkg/export"
+	"go.opentelemetry.io/obi/pkg/export/attributes"
 	"go.opentelemetry.io/obi/pkg/export/debug"
 	"go.opentelemetry.io/obi/pkg/export/instrumentations"
 	"go.opentelemetry.io/obi/pkg/filter"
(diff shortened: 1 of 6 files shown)


internal/config/convert/import.go
```diff
@@ -1026,11 +1027,129 @@ func applyPartialV2CaptureTelemetry(cfg *obi.Config, telemetry schema.CaptureTel
...
+	if completeKubernetesEnricher(kubernetes) {
+		applyFullV2KubernetesEnricher(cfg, kubernetes)
+		return
+	}
+
+	applyPartialV2KubernetesEnricher(cfg, kubernetes)
+}
+
+func applyFullV2KubernetesEnricher(cfg *obi.Config, kubernetes schema.KubernetesEnricher) {
+	cfg.Attributes.Kubernetes.Enable = kubernetes.Mode
```
grcevski: I think this is a bug here, v2 schema uses: autodetect, enabled, disabled, but validation expects: autodetect, true, false

internal/config/convert/import.go
```diff
@@ -1301,6 +1420,56 @@ func completeCaptureTelemetry(telemetry schema.CaptureTelemetry) bool {
 		!zeroValue(telemetry.Metrics)
 }
 
+func completeEnrichmentAttributes(attrs schema.EnrichmentAttributes) bool {
+	return !zeroValue(attrs.MetadataRetry.Timeout) &&
```
grcevski: I'm not sure what is the right call here, the RetryConfig allows 0, so exporting a v2 config with `metadata_retry.timeout: 0s` will just import the default 30s.

internal/config/convert/import.go
```diff
@@ -1301,6 +1420,56 @@ func completeCaptureTelemetry(telemetry schema.CaptureTelemetry) bool {
...
+		!zeroValue(attrs.MetadataRetry.StartInterval) &&
+		!zeroValue(attrs.MetadataRetry.MaxInterval)
+}
+
+func completeKubernetesEnricher(kubernetes schema.KubernetesEnricher) bool {
+	return kubernetes.Mode != "" &&
+		!zeroValue(kubernetes.Informers.InitialSyncTimeout) &&
+		!zeroValue(kubernetes.Informers.ReconnectInitialInterval) &&
+		!zeroValue(kubernetes.Informers.ResyncPeriod) &&
```
grcevski: Same issue here with InformersResyncPeriod allowing 0, but the v2 config doesn't so we import the default.

internal/config/convert/import.go
MrAlias: Fixed in d943649cd5. The v2 schema now has its own Kubernetes mode enum, export emits `enabled`/`disabled`/`autodetect`, and import maps those back to the runtime `true`/`false`/`autodetect` values.

internal/config/convert/import.go
MrAlias: Fixed in d943649cd5. `metadata_retry.timeout: 0s` is now preserved for exported full v2 configs, and the enrichment round-trip test covers that case.

internal/config/convert/import.go
MrAlias: Fixed in d943649cd5. `informers.resync_period: 0s` is now preserved for exported full v2 configs, with round-trip coverage alongside the metadata retry zero case.

grcevski (approved):
LGTM!

PR #2478, opened 2026-06-24
Add path-based conditional triggers to CI workflows

pinoOgni:
## Summary

Several workflows had no path filters and fired on every PR, including documentation-only changes. 

This PR adds `paths:` guards to some workflows so CI is skipped when only irrelevant files change.

Note: path lists may be incomplete or overly restrictive.

## Validation

### DIFF
--- a/.github/workflows/bpf_verifier.yml
+++ b/.github/workflows/bpf_verifier.yml
@@ -7,12 +7,16 @@ on:
     branches: [main, release-*]
     paths:
       - ".github/workflows/bpf_verifier.yml"
+      - ".github/actions/generate-bpf/**"
+      - ".github/actions/free-disk/**"
+      - ".github/actions/setup-lvh-kernel/**"
       - "bpf/**"
       - "**.go"
       - "go.mod"
       - "go.sum"
       - "internal/test/vm/**"
       - "scripts/kernel-matrix-json.sh"
+      - "Makefile"
   workflow_call:
     inputs:
       ref:
--- a/.github/workflows/bpf_verifier.yml
+++ b/.github/workflows/bpf_verifier.yml
@@ -5,6 +5,14 @@ on:
     branches: [main, release-*]
   pull_request:
     branches: [main, release-*]
+    paths:
+      - ".github/workflows/bpf_verifier.yml"
+      - "bpf/**"
+      - "**.go"
+      - "go.mod"
+      - "go.sum"
+      - "internal/test/vm/**"
+      - "scripts/kernel-matrix-json.sh"
   workflow_call:
     inputs:
       ref:
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -5,6 +5,22 @@ on:
     branches: [main, release-*]
   pull_request:
     branches: [main, release-*]
+    paths:
+      - ".github/workflows/pull_request.yml"
+      - "**.go"
+      - "go.mod"
+      - "go.sum"
+      - "bpf/**"
+      - "cmd/**"
+      - "collector/**"
+      - "configs/**"
+      - "internal/**"
+      - "pkg/**"
+      - "schemas/**"
+      - "scripts/**"
+      - "Makefile"
+      - "Dockerfile"
+      - "**/*Dockerfile*"
   workflow_call:
     inputs:
       ref:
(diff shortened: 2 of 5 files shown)


skl (approved):
The important thing is to always trigger everything relevant on main and release branches - as this only affects PRs, then lgtm!

.github/workflows/bpf_verifier.yml
```diff
@@ -5,6 +5,14 @@ on:
     branches: [main, release-*]
   pull_request:
     branches: [main, release-*]
+    paths:
```
MrAlias, line 8: This filter misses several inputs that the verifier workflow executes directly. The jobs use `./.github/actions/generate-bpf`, `./.github/actions/free-disk`, and `./.github/actions/setup-lvh-kernel`, and they also invoke Make targets via the generate action and `make -C internal/test/vm`.

In a PR that only changes one of those local actions or verifier-related Make targets, this workflow would now be skipped, so we could merge a broken verifier path without running it. Please include the relevant `.github/actions/.../**` paths and `Makefile` here.

.github/workflows/unit_tests.yml
```diff
@@ -5,6 +5,17 @@ on:
     branches: [main, release-*]
   pull_request:
     branches: [main, release-*]
+    paths:
```
MrAlias, line 8: This filter skips unit tests for changes to the `Makefile`, even though this workflow builds the shard matrix and runs the tests through Make targets (`unit-test-matrix-json`, `prereqs`, `run-unit-test-shard`, and `cov-exclude-generated`). It also uses `./.github/actions/generate-bpf`.

A Makefile-only or generate-action-only PR can therefore bypass the unit-test workflow that would validate it. Please add `Makefile` and `.github/actions/generate-bpf/**` to this path list.

.github/workflows/pull_request.yml
```diff
@@ -5,6 +5,22 @@ on:
     branches: [main, release-*]
   pull_request:
     branches: [main, release-*]
+    paths:
```
MrAlias, line 8: This workflow now triggers on changes to the workflow file itself, but not on changes to the local `generate-bpf` action it calls. A PR that only updates `./.github/actions/generate-bpf` could break `make docker-generate` or artifact publishing without running the main PR checks.

Please include `.github/actions/generate-bpf/**` in this filter.

.github/workflows/pull_request_privileged.yml
```diff
@@ -5,6 +5,17 @@ on:
     branches: [main, release-*]
   pull_request:
     branches: [main, release-*]
+    paths:
```
MrAlias, line 8: This filter omits `Makefile`, but the privileged test workflow runs `make prereqs docker-generate`, `sudo make test-privileged`, and `make cov-exclude-generated`.

If a PR changes only one of those Make targets, the privileged tests would be skipped even though their behavior changed. Please include `Makefile` in this path list.

.github/workflows/pull_request_docker_build_test.yml
```diff
@@ -3,6 +3,15 @@ name: Build test
 on:
   pull_request:
     branches: [main, release-*]
+    paths:
```
MrAlias, line 6: The Docker build workflow still depends on files outside this filter. Both Dockerfiles copy and execute `Makefile`, the root image copies `dependencies.Dockerfile`, `LICENSE`, `NOTICE`, and `NOTICES`, and `.dockerignore` can change what reaches the build context.

With the current filter, a Makefile-only or Docker-context-only change can skip image build validation. Please include the build inputs this workflow relies on, at minimum `Makefile`, `.dockerignore`, and `dependencies.Dockerfile`.

MrAlias (changes requested):
Thanks for tightening the PR CI surface. I think the new filters need one dependency pass before this lands: a few workflows can now skip when one of their own local actions, Make targets, or Docker build inputs changes. I left the concrete cases inline.

pinoOgni:
> Thanks for tightening the PR CI surface. I think the new filters need one dependency pass before this lands: a few workflows can now skip when one of their own local actions, Make targets, or Docker build inputs changes. I left the concrete cases inline.

Thanks Tyler! Applied all the suggestions!

MrAlias (approved):

PR #2490, opened 2026-06-24
Forward host processes through kube discovery

### LINKED ISSUE #2486: Kubernetes enricher drops explicitly targeted host processes such as kubelet
### Summary

When Kubernetes metadata enrichment is enabled, OBI appears to drop host processes that are not running in a container, even when those processes are explicitly targeted for instrumentation.

This prevents instrumenting node processes such as `kubelet` while also keeping Kubernetes metadata enrichment enabled.

### Environment

- OBI app version: `v0.9.0`
- OBI revision observed in logs: `e92b0c2`
- Helm chart: `open-telemetry/opentelemetry-ebpf-instrumentation` `0.9.4`
- Kubernetes: kind, Kubernetes `v1.30.0`
- DaemonSet settings included `hostPID: true`, `hostNetwork: true`, and privileged mode

### Reproduction

Target kubelet explicitly, for example with:

```yaml
attributes:
  kubernetes:
    enable: true

discovery:
  instrument:
    - name: kubelet
      exe_path: "*kubelet*"
      target_pids:
        - 718
```

or with:

```yaml
env:
  OTEL_EBPF_TARGET_PID: "718"
```

In the test cluster, kubelet was visible as:

```text
pid: 718
exe: /usr/bin/kubelet
cgroup: 0::/kubelet.slice/kubelet.service
```

### Actual Behavior

With Kubernetes metadata enabled, OBI's process watcher sees kubelet:

```text
Executable ready path=/usr/bin/kubelet pid=718 port=10248
process added pid=718 port=10248
```

Then the Kubernetes enricher drops it:

```text
can't get container info for PID pid=718 error="/proc/718/cgroup: couldn't find any docker entry for process with PID 718"
```

The process never reaches language detection, criteria matching, or trace attachment.

This looks consistent with the current code path:

- `ProcessWatcher` sends events into `WatcherKubeEnricher`
- `watcherKubeEnricher.onNewProcess` returns `false` when container info cannot be found
- non-container host processes are therefore not forwarded to later stages

Relevant source:

- `finder.go` wires `ProcessWatcher` through `WatcherKubeEnricher` before language detection and criteria matching:
  https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/e92b0c2dd4d88b90d28b7b706eacbac2fabd2623/pkg/appolly/discover/finder.go#L90-L112

- `watcher_kube.go` drops processes without container info: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/e92b0c2dd4d88b90d28b7b706eacbac2fabd2623/pkg/appolly/discover/watcher_kube.go#L171-L180

- top-level `target_pids` does create PID-only selection criteria: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/e92b0c2dd4d88b90d28b7b706eacbac2fabd2623/pkg/appolly/discover/matcher.go#L389-L401

### Expected Behavior

A non-container host process should be forwarded undecorated when Kubernetes metadata cannot be found, rather than dropped.

That would allow explicit PID/path selectors to instrument host processes such as `kubelet`, while still preserving Kubernetes metadata decoration for normal pod/container workloads.

### Validation

When Kubernetes metadata was disabled:

```yaml
attributes:
  kubernetes:
    enable: false
```

the same kubelet process was successfully instrumented:

```text
detected type pid=718 type=go
found process pid=718 comm=/usr/bin/kubelet
instrumenting process cmd=/usr/bin/kubelet pid=718 type=go
```

OBI then resolved pod gateway traffic as kubelet:

```text
peer_ip=10.244.0.1 peer_name=kubelet
```

### Why This Matters

Kubelet and other node-level processes can generate traffic that appears in service graphs as inferred services or raw node/pod-network IPs. Today, users must choose between:

1. Kubernetes metadata enrichment for pod workloads, or
2. host process instrumentation for node processes such as kubelet.

It would be useful for these to coexist.

### Possible Fix

In `watcherKubeEnricher.onNewProcess`, when `getContainerInfo` fails because the process is not containerized, return the original `ProcessAttrs` with `true` instead of dropping it.

That would preserve current metadata decoration for containerized processes while allowing non-container host processes to continue through language detection, criteria matching, and attachment.

MrAlias:
## What changed

This keeps Kubernetes metadata enrichment from dropping host processes when the process has no supported container cgroup entry. Those processes now continue through discovery undecorated, so explicit PID or path selectors can still instrument node-level processes such as kubelet while Kubernetes metadata enrichment remains enabled.

The container helper now exposes a sentinel `ErrContainerNotFound` error. The Kubernetes watcher forwards only that expected host-process case and keeps the previous drop behavior for other container lookup failures such as `/proc` read or namespace errors.

Fixes #2486.

## Validation

- `make generate`
- `go test ./pkg/appolly/discover`
- `go test ./pkg/internal/helpers/container`

### DIFF
--- a/pkg/internal/helpers/container/container.go
+++ b/pkg/internal/helpers/container/container.go
@@ -6,6 +6,7 @@ package container // import "go.opentelemetry.io/obi/pkg/internal/helpers/contai
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"os"
 	"regexp"
@@ -20,6 +21,9 @@ var (
 	procRoot        = "/proc/"
 	namespaceFinder = procs.FindNamespace
 )
+
+// ErrContainerNotFound is returned when a process has no supported container cgroup entry.
+var ErrContainerNotFound = errors.New("container not found")
 
 // Info that we need to keep from a container: its ContainerID in Kubernetes and
 // the PIDNamespace of its processes.
@@ -69,7 +73,7 @@ func InfoForPID(pid app.PID) (Info, error) {
 			return Info{PIDNamespace: ns, ContainerID: cgroupID}, nil
 		}
 	}
-	return Info{}, fmt.Errorf("%s: couldn't find any docker entry for process with PID %d", cgroupFile, pid)
+	return Info{}, fmt.Errorf("%s: couldn't find any docker entry for process with PID %d: %w", cgroupFile, pid, ErrContainerNotFound)
 }
 
 // look for a cgroup ID on all the possible formats
--- a/pkg/internal/helpers/container/container.go
+++ b/pkg/internal/helpers/container/container.go
@@ -73,7 +73,7 @@ func InfoForPID(pid app.PID) (Info, error) {
 			return Info{PIDNamespace: ns, ContainerID: cgroupID}, nil
 		}
 	}
-	return Info{}, fmt.Errorf("%s: couldn't find any docker entry for process with PID %d: %w", cgroupFile, pid, ErrContainerNotFound)
+	return Info{}, ErrContainerNotFound
 }
 
 // look for a cgroup ID on all the possible formats
--- a/pkg/appolly/discover/watcher_kube.go
+++ b/pkg/appolly/discover/watcher_kube.go
@@ -5,6 +5,7 @@ package discover // import "go.opentelemetry.io/obi/pkg/appolly/discover"
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"sync"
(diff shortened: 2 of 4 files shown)


grcevski (approved):
LGTM! 🥇

pkg/internal/helpers/container/container.go
```diff
@@ -69,7 +73,7 @@ func InfoForPID(pid app.PID) (Info, error) {
 			return Info{PIDNamespace: ns, ContainerID: cgroupID}, nil
 		}
 	}
-	return Info{}, fmt.Errorf("%s: couldn't find any docker entry for process with PID %d", cgroupFile, pid)
+	return Info{}, fmt.Errorf("%s: couldn't find any docker entry for process with PID %d: %w", cgroupFile, pid, ErrContainerNotFound)
```
NimrodAvni78: if we ignore this error anyway can we just return `Info{}, ErrContainerNotFound` to not allocate a new error instance every time?

MrAlias: Good call. Updated `InfoForPID` to return `ErrContainerNotFound` directly for that expected host-process path, while keeping wrapped errors for the other lookup failures.

mmat11 (approved):
lgtm!

PR #2497, opened 2026-06-25
Add store-demo architecture doc with CI sync check

pinoOgni:
## Summary

In the last SIG, we talked about improving the store-demo.

This PR adds `ARCHITECTURE.md`, which shows the store-demo topology as a Mermaid graph (services colored by their implementation language) plus connection and service-language tables. The graph and tables are generated from the k8s manifests (and the services' source trees) by `fix_architecture.py` so they're not hand-maintained.

It also adds a CI check that fails if the committed doc doesn't match what the manifests would generate (services, connections, ports, protocols, or languages), so the diagram can't silently go stale. After changing a manifest, regenerate the doc with `python3 examples/store-demo/fix_architecture.py`; CI runs the same script with `--check`.

image

## Validation


examples/store-demo/ARCHITECTURE.md
```diff
@@ -0,0 +1,122 @@
+# Store Demo Architecture
+
+This describes the service topology of the vendored Online Boutique app (`GoogleCloudPlatform/microservices-demo` v0.10.5, see [PROVENANCE.md](./PROVENANCE.md)) as wired by the manifests in [`k8s`](./k8s).
```
mmat11: should the version be removed from here? is it checked as well or could it go stale?

examples/store-demo/ARCHITECTURE.md
```diff
@@ -0,0 +1,122 @@
...
+
+> Keep this file in sync with the manifests. When you change a `*_SERVICE_ADDR`
+> env var, add or remove a service manifest, or change a listen port, update the
+> diagram and the table below in the same change.
+>
+> CI enforces the connection table against the manifests via
+> [check_architecture.py](./check_architecture.py). Run it locally with
```
mmat11: should we rename it to `fix_architecture.py` or similar to automatically fix it and call it with `--check` in CI?

.github/workflows/store-demo-architecture.yml
```diff
@@ -0,0 +1,32 @@
...
+  check:
+    name: ARCHITECTURE.md matches manifests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          persist-credentials: false
+      - name: Verify store-demo architecture
+        run: python3 examples/store-demo/check_architecture.py
```
mmat11: can we merge this into the PR checks workflow?

examples/store-demo/ARCHITECTURE.md
pinoOgni: good question, ~we can leave the original demo version (v0.10.5) as a reference and then add a store-demo OBI version that is checked. wdyt?~ 
better idea: point to `PROVENANCE.md` file for the version.

examples/store-demo/ARCHITECTURE.md
pinoOgni: mmmmh, let me try

.github/workflows/store-demo-architecture.yml
pinoOgni: IMO this is fine like that and it's like the `markdown-fail-fast` job. Is there any reasons to merge it? Maybe I'm missing something

examples/store-demo/ARCHITECTURE.md
pinoOgni: Done! :)

.github/workflows/store-demo-architecture.yml
mmat11: I don't have a strong opinion, having these fast checks separate adds UI bloat, but let's leave it as is for now, we'll eventually refactor in the future if it gets problematic

NimrodAvni78 (commented):
Can we manage it with something more standard like a make file?
also ideally we could rewrite fix_arcitecture in go to not add python dependecy to the project
wdyt?

pinoOgni:
> Can we manage it with something more standard like a make file? 

Oh yes, nice suggestion!

> also ideally we could rewrite fix_arcitecture in go to not add python dependecy to the project wdyt?

I'm not sure I understand the "python dependency" issue. The CI already runs python tools and tests. Could this new job cause problems?

NimrodAvni78:
oh i didnt know we already run python tooling in CI, if so i guess thats fine

.github/workflows/store-demo-architecture.yml
```diff
@@ -0,0 +1,34 @@
+name: Store Demo Architecture
+
+on:
+  push:
+    paths:
+      - "examples/store-demo/k8s/**"
```
mmat11: why not just `- "examples/store-demo/**"`?

examples/store-demo/ARCHITECTURE.md
```diff
@@ -0,0 +1,130 @@
...
+| frontend | Go | `go.mod` |
+| loadgenerator | Python | `requirements.txt` |
+| paymentservice | Node.js | `package.json` |
+| productcatalogservice | Go | `go.mod` |
+| recommendationservice | Python | `requirements.txt` |
+| redis-cart | Redis (datastore) | upstream `redis:alpine` image |
+| shippingservice | Go | `go.mod` |
+<!-- /generated:languages -->
+
+## Notes For OBI
```
mmat11: ideally the demo should be "balanced" with the use of different protocols and so on, it's little to no use noting that majority of calls are grpc based. If we plan on enriching the demo with more services/languages/protocols/products, I think we should just omit these notes and gather in a GH umbrella issue what's missing here

examples/store-demo/fix_architecture.py
```diff
@@ -0,0 +1,234 @@
...
+
+The only fact not in the manifests is the wire protocol, which is a function of
+the callee: the frontend is reached over HTTP, redis-cart over Redis, and every
+other service over gRPC (see PROTOCOL_BY_CALLEE).
+
+Usage:
+  python3 fix_architecture.py            # rewrite the generated regions in place
+  python3 fix_architecture.py --check    # fail (exit 1) if the doc is stale; CI mode
+
+stdlib only — no PyYAML — so it cannot fail on a missing dependency.
```
mmat11:
```suggestion
```

examples/store-demo/fix_architecture.py
```diff
@@ -0,0 +1,234 @@
...
+# name: SOMETHING_ADDR  then  value: "callee:port" on the following line.
+ADDR_RE = re.compile(
+    r'name:\s*([A-Za-z0-9_]+_ADDR)\s*[\r\n]+\s*value:\s*"?([^"\r\n]+?)"?\s*$',
+    re.MULTILINE,
+)
+
+# Protocol is the one thing not encoded in the manifests. In this demo it is a
+# property of the callee: everything is gRPC except the HTTP frontend and the
+# Redis datastore. New non-gRPC callees must be added here.
+PROTOCOL_BY_CALLEE = {"frontend": "HTTP", "redis-cart": "Redis"}
```
mmat11: should these be manually maintained?

examples/store-demo/fix_architecture.py
```diff
@@ -0,0 +1,234 @@
...
+DEFAULT_PROTOCOL = "gRPC"
```
mmat11: why is this the default protocol?

examples/store-demo/fix_architecture.py
```diff
@@ -0,0 +1,234 @@
...
+    ("Java", "java", "darkorange", "chocolate", "black"),
+    ("Redis (datastore)", "datastore", "firebrick", "darkred", "white"),
+]
+
+
+def app_manifests() -> list[Path]:
+    """Per-service manifests only: skip numbered infra files and kustomization."""
+    return [
+        y for y in sorted(K8S.glob("*.yaml"))
+        if not y.name[0].isdigit() and y.name != "kustomization.yaml"
```
mmat11, line 71: I think we should document next services shouldn't start with a number so they will be filtered here. I would also validate the option to use an hardcoded blacklist

examples/store-demo/fix_architecture.py
```diff
@@ -0,0 +1,234 @@
...
+    ]
+
+
+def manifest_services() -> set[str]:
+    """Names of every Deployment declared in the per-service manifests."""
+    services = set()
+    for yaml in app_manifests():
+        expect_name = False
+        for line in yaml.read_text().splitlines():
+            if re.match(r"\s*kind:\s*Deployment\s*$", line):
```
mmat11, line 86: we should also document to only use `Deployment` as kind for future additions, as this would filter out daemonsets, pods, etc.

examples/store-demo/fix_architecture.py
```diff
@@ -0,0 +1,234 @@
...
+    edges = set()
+    for yaml in app_manifests():
+        caller = yaml.stem  # one service per manifest; filename == caller
+        for _name, value in ADDR_RE.findall(yaml.read_text()):
+            callee, _, port = value.strip().rpartition(":")
+            edges.add((caller, callee, port))
+    return edges
+
+
+def detect_language(service: str) -> tuple[str, str]:
```
mmat11, line 129: this heuristic also needs to be documented

pinoOgni: Sure

examples/store-demo/ARCHITECTURE.md
pinoOgni: Yep, so the idea is to remove the notes and I'll open an umbrella issue to track the demo enrichment. Is it ok?

examples/store-demo/fix_architecture.py
pinoOgni: Protocols can't be auto-detected reliably so yes, it's manually maintained in the python script.
We can add a `appProtocol` in the k8s manifest and the script will read it.

examples/store-demo/fix_architecture.py
pinoOgni: Just because all the demo is grpc. It conflicts with the "balanced" comment you said so I need to remove it. Moreover if we go ahed with `appProcol` idea this will disappears entirely.

examples/store-demo/ARCHITECTURE.md
mmat11: I think so, yes

examples/store-demo/fix_architecture.py
mmat11: can we add an "how to add a new service" section to the docs with all the places that should be manually maintained?

grcevski (approved):
LGTM!

mmat11 (approved):

PR #2499, opened 2026-06-25
Reject symlinked Java classpath paths under proc roots

MrAlias:
### Motivation

- Prevent a monitored Java process from causing the agent to resolve and scan host filesystem paths by embedding absolute symlinked entries in its classpath under `/proc/<pid>/root`.
- Address a logic gap where proc-root resolution performed only `os.Stat` on the lexical host path while later scan operations followed symlinks and could traverse outside the intended process root.

### Description

- Update `resolveProcessPath` to reject proc-root resolutions that contain symlink components by adding a `pathHasSymlink(root, containerPath)` check before `os.Stat` for `/proc/<pid>/root` paths.
- Introduce `procRootPath` indirection (defaulting to `isProcRoot`) to allow deterministic test coverage of the proc-root branch.
- Add unit testing for rejection of symlink components for proc roots in `classpath_test.go` that exercises the proc-root symlink-rejection behavior.

### Testing

- Ran `make generate BPF_GEN_ALL=pkg/ebpf/common/bpf_bpfel.go` successfully to ensure generated artifacts were available.
- Ran `go test ./pkg/internal/transform/route/harvest/java` and the package tests (including the new regression test) passed (`ok`).

### DIFF
--- a/pkg/internal/transform/route/harvest/java/archive.go
+++ b/pkg/internal/transform/route/harvest/java/archive.go
@@ -48,9 +48,12 @@ func (e *Extractor) scanClassFile(ctx context.Context, path string) error {
 	if err := ctx.Err(); err != nil {
 		return err
 	}
-	info, err := os.Stat(path)
+	info, err := os.Lstat(path)
 	if err != nil {
 		e.log.Debug("error stating Java class file", "path", path, "error", err)
+		return nil
+	}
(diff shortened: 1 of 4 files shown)


grcevski (approved):
LGTM! Nice catch!

mmat11 (approved):
lgtm!

PR #2503, opened 2026-06-25
Avoid GenAI error text in span status

MrAlias:
### Motivation
- Prevent exporting raw GenAI provider error messages via `status.message`. Unlike DB error status messages, which are derived from the optional `db.response.error` attribute only when selected, GenAI provider error text was copied directly into the OTLP status message and bypassed `attributes.select.traces`.

### Description
- Remove the provider-specific GenAI fallback from `SpanStatusMessage` so `genAIErrorMessage` is no longer returned into span status for HTTP spans.  
- Remove the now-unused `genAIErrorMessage` helper from `pkg/appolly/app/request/span.go`.  
- Update tests in `pkg/export/otel/traces_test.go` to assert `topSpan.Status().Message()` is empty for GenAI error cases while keeping `error.type` and other non-sensitive attributes intact.  
- Preserve existing behavior for JSON-RPC, MCP (`MCP.ErrorMessage`), and DNS status paths.

### Testing
- `go test ./pkg/appolly/app/request -run TestSpanStatusMessage` 
- `go test ./pkg/export/otel -run TestGenerateTracesAttributes`

### DIFF
--- a/pkg/appolly/app/request/span.go
+++ b/pkg/appolly/app/request/span.go
@@ -1586,38 +1586,10 @@ func SpanStatusMessage(span *Span) string {
 		if span.SubType == HTTPSubtypeMCP && span.GenAI != nil && span.GenAI.MCP != nil && span.GenAI.MCP.ErrorMessage != "" {
 			return span.GenAI.MCP.ErrorMessage
 		}
-		if msg := genAIErrorMessage(span); msg != "" {
-			return msg
-		}
 	case EventTypeDNS:
 		if span.Status != 0 {
 			return dnsparser.RCode(span.Status).String()
 		}
-	}
-	return ""
-}
-
-func genAIErrorMessage(span *Span) string {
-	if span.GenAI == nil {
-		return ""
-	}
-	if span.GenAI.OpenAI != nil && span.GenAI.OpenAI.Error.Message != "" {
-		return span.GenAI.OpenAI.Error.Message
-	}
-	if span.GenAI.Anthropic != nil && span.GenAI.Anthropic.Output.Error != nil && span.GenAI.Anthropic.Output.Error.Message != "" {
-		return span.GenAI.Anthropic.Output.Error.Message
-	}
-	if span.GenAI.Gemini != nil && span.GenAI.Gemini.Output.Error != nil && span.GenAI.Gemini.Output.Error.Message != "" {
-		return span.GenAI.Gemini.Output.Error.Message
-	}
-	if span.GenAI.Qwen != nil && span.GenAI.Qwen.Error.Message != "" {
-		return span.GenAI.Qwen.Error.Message
-	}
-	if span.GenAI.Bedrock != nil && span.GenAI.Bedrock.Output.ErrorMessage != "" {
-		return span.GenAI.Bedrock.Output.ErrorMessage
-	}
-	if span.GenAI.Rerank != nil && span.GenAI.Rerank.Output.Error != nil && span.GenAI.Rerank.Output.Error.Message != "" {
-		return span.GenAI.Rerank.Output.Error.Message
 	}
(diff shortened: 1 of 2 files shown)


pkg/appolly/app/request/span.go
```diff
@@ -1586,9 +1586,6 @@ func SpanStatusMessage(span *Span) string {
 		if span.SubType == HTTPSubtypeMCP && span.GenAI != nil && span.GenAI.MCP != nil && span.GenAI.MCP.ErrorMessage != "" {
 			return span.GenAI.MCP.ErrorMessage
 		}
-		if msg := genAIErrorMessage(span); msg != "" {
```
grcevski, line 1589: Are we concerned this will have sensitive information, where should we put the error message? I think it's useful to have it, as it often reports things such as: you have reached your token quota limit, or bad API key.

MrAlias: I agree the raw message is useful, but I do not think `status.message` is the right default place for it.

The concern is not that every provider error message is sensitive. It is that this field is provider-supplied free text, and OBI cannot reliably know whether it contains safe text like "quota exceeded" or more sensitive material such as API key fragments, organization/project/account IDs, request details, prompt excerpts, tenant metadata, or customer usage/billing state. Once it is placed in OTLP `status.message`, it bypasses OBI's trace attribute selection controls.

The OpenTelemetry trace spec allows a span status description/message only with `Error`, but it also says instrumentation-set descriptions should be documented and predictable.[^1] The general error semantic conventions recommend setting `error.type`, and only setting the status description when the extra information is not expected to contain sensitive details.[^2] GenAI semantic conventions also model `error.type` as the low-cardinality provider/client error identifier, which fits errors like `insufficient_quota`, `authentication_error`, `invalid_api_key`, `rate_limit_error`, etc.[^3]

So I think the safe default is:

- keep `status.code = Error`
- keep/report `error.type` for the actionable classification
- do not put raw provider error text in `status.message` by default

If we want to expose the raw provider message, I would prefer doing it behind an explicit opt-in control, similar in spirit to how GenAI input/output content is opt-in and how DB status messages are currently derived from the optional `db.response.error` path only when selected. That makes the tradeoff explicit for users who want diagnostic detail and have an appropriate telemetry storage/access model.

[^1]: OpenTelemetry Trace API: `Status` has `StatusCode` and optional `Description`; OTLP calls this field `message`. Instrumentation-set error descriptions should be documented and predictable. https://opentelemetry.io/docs/specs/otel/trace/api/#set-status
[^2]: OpenTelemetry Recording Errors semantic conventions: for errored spans, instrumentation should set span status code to `Error`, should set `error.type`, and should set status description only when it adds information that is not expected to contain sensitive details. https://opentelemetry.io/docs/specs/semconv/general/recording-errors/#recording-errors-on-spans
[^3]: OpenTelemetry GenAI semantic conventions: GenAI spans include conditionally required `error.type` for failed operations; it should match the provider/client error code, exception class, or another low-cardinality error identifier. https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-spans.md

MrAlias: Tracking follow-up with https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2507.

grcevski: OK, cool, yeah optional message is great solution.

grcevski (approved):
LGTM!

NameHaibinZhang:
LGTM! Good catch — raw provider error messages shouldn't bypass attributes.select.traces. The opt-in approach tracked in #2507 makes sense for cases where users explicitly need the error detail for debugging.

mmat11 (approved):

PR #2505, opened 2026-06-25
Preserve truncated request body read errors

MrAlias:
### Motivation

- A one-byte probe read from `req.Body` could consume an EOF/error and be ignored, which left downstream readers observing an empty body with a `nil` error instead of the original truncation error. 
- This changed downstream GenAI detectors' behavior when a request had a `Content-Length > 0` but the captured request body was truncated or missing.

### Description

- Move the probe/recovery logic into a new helper `recoverRequestBody(req, requestBuffer)` to centralize behavior and reduce accidental error swallowing. 
- Introduce `readErrorCloser` that implements `io.ReadCloser` to preserve and re-expose the original probe `Read` error to downstream readers when JSON recovery fails. 
- Keep the happy-path behavior of prepending a consumed probe byte via `io.MultiReader` when the probe read returned bytes. 
- Add focused unit tests `TestRecoverRequestBodyPreservesProbeErrorWhenRecoveryFails` and `TestRecoverRequestBodyPrependsProbeByte` validating error preservation and happy-path byte prepending.

### Testing

- Ran `go test ./pkg/ebpf/common` which completed successfully.

### DIFF
--- a/pkg/ebpf/common/http_transform.go
+++ b/pkg/ebpf/common/http_transform.go
@@ -327,24 +327,46 @@ func HTTPInfoEventToSpan(parseCtx *EBPFParseContext, event *BPFHTTPInfo) (reques
 	// We probe a single byte instead of ReadAll to avoid allocating and
 	// copying the entire body on the happy path.
 	if req.ContentLength > 0 {
-		var probe [1]byte
-		n, _ := req.Body.Read(probe[:])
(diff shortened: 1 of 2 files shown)


grcevski (approved):
LGTM!

rafaelroquetto (approved):

PR #2506, opened 2026-06-25
Fix Go channel probe offset discovery

MrAlias:
### Motivation

- Discovery builds the list of Go symbols to inspect before per-executable `hchan` offsets are recorded, which caused `runtime.chansend1`, `runtime.chanrecv1`, and `runtime.chanrecv2` to be omitted from the initial offset inspection and therefore never have return offsets discovered. 
- The intent of the change is to ensure those runtime channel-handoff symbols are always included in symbol-discovery while preserving the existing per-binary conditional attachment behavior that only enables the runtime probes when offsets are available.

### Description

- Introduce a shared symbol list `goChannelLinkProbeSymbols` and exported accessor `GoChannelLinkProbeSymbols()` in `pkg/internal/ebpf/gotracer` and use it where appropriate instead of duplicating literal symbol strings. 
- Change `gotracer.GoProbes()` to index the conditional probe map by the shared symbol list, keeping actual probe attachment conditional on recorded offsets via `goChannelLinkProbesEnabled()`. 
- Update discovery so `typer.loadAllGoFunctionNames()` always adds the channel-link symbols to `t.allGoFunctions` (using a new helper `addGoFunctionName`) so `goexec.InspectOffsets` is asked for their offsets during discovery. 
- Add a unit test in `pkg/appolly/discover/typer_test.go` to assert discovery includes the channel-link symbols and update existing gotracer tests to reference the shared symbol accessor.

### Testing
- `go test ./pkg/internal/ebpf/gotracer ./pkg/appolly/discover`
- `go test ./pkg/internal/ebpf/gotracer -run 'TestGoChannel|TestProcessBinary'`

### DIFF
--- a/pkg/appolly/discover/typer.go
+++ b/pkg/appolly/discover/typer.go
@@ -22,6 +22,7 @@ import (
 	"go.opentelemetry.io/obi/pkg/ebpf"
 	attr "go.opentelemetry.io/obi/pkg/export/attributes/names"
 	"go.opentelemetry.io/obi/pkg/export/imetrics"
+	"go.opentelemetry.io/obi/pkg/internal/ebpf/gotracer"
 	"go.opentelemetry.io/obi/pkg/internal/goexec"
 	"go.opentelemetry.io/obi/pkg/internal/procs"
 	"go.opentelemetry.io/obi/pkg/internal/transform/route/clusterurl"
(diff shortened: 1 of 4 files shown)


grcevski (approved):
Nice catch! Are these the only functions that suffer from this?

MrAlias:
> Are these the only functions that suffer from this?

Yes, I believe these are the only ones with this specific circular dependency. Discovery builds `allGoFunctions` from `GoProbes()` before `RegisterOffsets()` has recorded per-binary state. The only `GoProbes()` entries gated on that per-binary offset state are `runtime.chansend1`, `runtime.chanrecv1`, and `runtime.chanrecv2` via `goChannelLinkProbesEnabled()`, as far as I can tell.

The other conditional probe entries are gated by config or kernel capability, like BPF loop support, route harvesting, or context propagation, so leaving those out of offset inspection is intentional because they would not be attached under that configuration anyway. The goal here is to keep channel probe attachment conditional, but makes sure these three symbols are always included in the initial offset inspection.

grcevski:
> > Are these the only functions that suffer from this?
> 
> Yes, I believe these are the only ones with this specific circular dependency. Discovery builds `allGoFunctions` from `GoProbes()` before `RegisterOffsets()` has recorded per-binary state. The only `GoProbes()` entries gated on that per-binary offset state are `runtime.chansend1`, `runtime.chanrecv1`, and `runtime.chanrecv2` via `goChannelLinkProbesEnabled()`, as far as I can tell.
> 
> The other conditional probe entries are gated by config or kernel capability, like BPF loop support, route harvesting, or context propagation, so leaving those out of offset inspection is intentional because they would not be attached under that configuration anyway. The goal here is to keep channel probe attachment conditional, but makes sure these three symbols are always included in the initial offset inspection.

Makes sense, thanks!

mmat11 (approved):

PR #2509, opened 2026-06-26
Redact legacy AWS signed URL query keys

MrAlias:
### Motivation

- The default query-parameter redaction list omitted legacy AWS signed-URL keys such as `AWSAccessKeyId` and the case-sensitive `Signature`, allowing those values to be exported unredacted when `url.query`/`url.full` are emitted by default.
- The scrubber performs exact, case-sensitive lookups on decoded keys, so adding the legacy keys is necessary to prevent credential leakage for signed URLs.

### Description

- Add `"AWSAccessKeyId"` and `"Signature"` to `DefaultSensitiveQueryParams` in `pkg/export/attributes/attr_selector.go` so exact-match redaction covers legacy AWS signed-URL parameters.
- Add `TestDefaultSensitiveQueryParamsIncludesLegacyAWSSignedURLKeys` in `pkg/export/attributes/attr_selector_test.go` to assert the default list contains the new entries.
- Add a span-level test in `pkg/export/otel/tracesgen/tracesgen_test.go` that verifies `url.query` redacts `AWSAccessKeyId` and `Signature` when the default sensitive list is applied.

### Testing
- `go test ./pkg/export/attributes ./pkg/export/otel/tracesgen -run 'TestDefaultSensitiveQueryParamsIncludesLegacyAWSSignedURLKeys|TestHTTPServerSpanURLQuery' -count=1`
- `go test ./pkg/export/attributes ./pkg/export/otel/tracesgen -count=1`

### DIFF
--- a/pkg/export/attributes/attr_selector.go
+++ b/pkg/export/attributes/attr_selector.go
@@ -86,6 +86,7 @@ func parseExtraAttrGroup(group string) (AttrGroups, error) {
 var DefaultSensitiveQueryParams = []string{
 	// OTel semconv-recommended — https://opentelemetry.io/docs/specs/semconv/http/http-spans/
 	"X-Amz-Signature", "X-Amz-Credential", "X-Amz-Security-Token",
+	"AWSAccessKeyId", "Signature", "SecurityToken",
 	"X-Goog-Signature", "sig",
 	// Common sensitive parameters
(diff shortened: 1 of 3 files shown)


mariomac (approved):
Good point!

mmat11 (approved):

PR #2510, opened 2026-06-26
Preserve text glob patterns during export

MrAlias:
### Motivation

- Runtime v2 export serialized glob selectors via `GlobAttr.MarshalYAML()` but `GlobAttr.UnmarshalText()` (used for env/text-backed values) compiled the pattern without preserving its original text, causing exported rules to lose their glob strings.
- The change ensures env/text-provided glob selectors round-trip through marshal/unmarshal so exported v2 configs remain equivalent to the original runtime selection criteria.

### Description

- Populate the internal `str` field in `GlobAttr.UnmarshalText` and clear it for empty input so that `MarshalYAML` returns the original pattern for env/text-loaded globs (`pkg/appolly/services/attr_glob.go`).
- Add a unit test verifying that `UnmarshalText` preserves the pattern for subsequent `MarshalYAML` and that matching still works (`pkg/appolly/services/attr_glob_test.go`).
- Add an integration-style unit case to `internal/config/convert/export_test.go` asserting that env-backed top-level `AutoTargetExe` and `AutoTargetLanguage` selectors are exported with `exe_path_glob` and `language_glob` set correctly.

### Testing
- `go test ./pkg/appolly/services -run TestGlobAttrUnmarshalTextPreservesPatternForMarshalYAML -count=1 -v`
- `go test ./pkg/appolly/services -count=1`
- `go test ./internal/config/convert -run TestRuntimeToV2EffectiveDiscoveryCriteria -count=1 -v`

### DIFF
--- a/internal/config/convert/export_test.go
+++ b/internal/config/convert/export_test.go
@@ -649,6 +649,22 @@ func TestRuntimeToV2EffectiveDiscoveryCriteria(t *testing.T) {
 		require.Equal(t, []string{"go", "java"}, value(t, ext.Capture.Rules[0].Match, "process", "language_glob"))
 	})
 
+	t.Run("env-backed top-level glob selectors", func(t *testing.T) {
+		t.Parallel()
+
+		cfg := minimalSelectionConfig()
+		require.NoError(t, cfg.AutoTargetExe.UnmarshalText([]byte("/srv/*")))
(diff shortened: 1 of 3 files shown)


mariomac (approved):
Good catch!

mmat11 (approved):

PR #2512, opened 2026-06-26
Add AST-based route harvesting for Node.js variable paths

### LINKED ISSUE #929: Extend nodejs route harvesting with variable matching
As pointed out during the [review](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/924#discussion_r2560087152), the current nodejs route harvester is mostly looking at static strings. This issue is to follow-up on the review and implement a smarter scanning using the mentioned library.

Vedanshu7:
#### Summary

The existing regex-based route harvester only matched hardcoded string paths like `app.get('/users', handler)`. It missed routes where the path came from a variable, a template literal, or string concatenation.

This PR adds an AST-based fallback pass using Goja's parser that handles those cases:

- variable: `const p = '/users'; app.get(p, handler)`
- template literal: `app.get(/api/${version}/users, handler)`
- string concatenation: `app.get('/api' + '/users', handler)`
- object property access and destructuring
- loop variables from `for...of`
- cross-file paths via `require()`

#### Link to tracking issue
Fixes #929

#### Testing
Added unit tests in `pkg/internal/transform/route/harvest/js_ast_test.go` covering variable paths, template literals, string concatenation, object property access, destructuring, loop variables, and cross-file require(). All tests pass with go test `./pkg/internal/transform/route/harvest/....`

#### Authorship

- [x] I, a human, wrote this pull request description myself.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/caarlos0/env/v11 v11.4.1
 	github.com/cilium/ebpf v0.20.0
 	github.com/containers/common v0.64.2
+	github.com/dop251/goja v0.0.0-20260618133527-c9b2ea77db59
 	github.com/gin-gonic/gin v1.12.0
 	github.com/go-logr/logr v1.4.3
 	github.com/go-logr/stdr v1.2.2
@@ -159,6 +160,7 @@ require (
 	github.com/go-openapi/swag/yamlutils v0.25.5 // indirect
 	github.com/go-playground/locales v0.14.1 // indirect
 	github.com/go-playground/universal-translator v0.18.1 // indirect
+	github.com/go-sourcemap/sourcemap v2.1.3+incompatible // indirect
 	github.com/goccy/go-yaml v1.19.2 // indirect
 	github.com/golang/snappy v1.0.0 // indirect
 	github.com/google/btree v1.1.3 // indirect
--- a/pkg/internal/transform/route/harvest/js.go
+++ b/pkg/internal/transform/route/harvest/js.go
@@ -473,6 +473,16 @@ func (e *RouteExtractor) scanFile(filePath string) error {
 		return err
 	}
 
+	// Fallback pass: resolve routes whose path comes from a variable or a
+	// template literal, which the line-based regex harvesters cannot match.
+	src, ok, err := readJSFileForScan(filePath)
+	if err != nil {
+		return err
+	}
+	if ok {
+		e.routes = append(e.routes, e.resolveASTRoutes(filePath, src)...)
+	}
+
 	return nil
 }
 
--- a/pkg/internal/transform/route/harvest/js_ast.go
+++ b/pkg/internal/transform/route/harvest/js_ast.go
@@ -1,0 +1,495 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package harvest // import "go.opentelemetry.io/obi/pkg/internal/transform/route/harvest"
+
+import (
+	"io"
+	"path/filepath"
+	"strings"
+
+	"github.com/dop251/goja/ast"
+	"github.com/dop251/goja/file"
+	"github.com/dop251/goja/parser"
+	"github.com/dop251/goja/token"
+)
+
+// readJSFileForScan returns the bounded contents of a JS/TS file using the same
+// safety guards as the line scanner. ok is false for non-regular or oversized
+// files that should be skipped.
+func readJSFileForScan(path string) (string, bool, error) {
+	f, ok, err := openJSFileForScan(path)
+	if err != nil || !ok {
+		return "", ok, err
+	}
+	defer f.Close()
+
+	b, err := io.ReadAll(io.LimitReader(f, MaxJSFileScanBytes))
+	if err != nil {
+		return "", false, err
+	}
+	return string(b), true, nil
+}
+
+// astRouteMethods maps the JS route-registration method name to the HTTP method
+// it represents. It mirrors the verbs handled by the regex harvesters.
+var astRouteMethods = map[string]string{
+	"get":     "GET",
+	"post":    "POST",
+	"put":     "PUT",
+	"patch":   "PATCH",
+	"delete":  "DELETE",
+	"del":     "DELETE",
+	"head":    "HEAD",
+	"options": "OPTIONS",
+	"all":     "ALL",
+}
+
+// resolveASTRoutes parses a JS source file and extracts routes whose path is
+// supplied through a string variable or a template literal. The line-based
+// regex harvesters only match quoted string literals, so this acts as a
+// fallback for the cases highlighted in
+// https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/929.
+//
+// It returns nil when the source cannot be parsed (for example TypeScript),
+// leaving the regex-based results untouched.
+func (e *RouteExtractor) resolveASTRoutes(filePath, src string) []RoutePattern {
+	return e.resolveASTRoutesInner(filePath, src, map[string]bool{filePath: true})
+}
+
+func (e *RouteExtractor) resolveASTRoutesInner(filePath, src string, visited map[string]bool) []RoutePattern {
+	prog, err := parser.ParseFile(nil, filePath, src, 0)
+	if err != nil {
+		return nil
+	}
+
+	dir := filepath.Dir(filePath)
+
+	// First pass: collect all possible string values for each variable.
+	// Repeated until stable to resolve chained dependencies.
+	// strVars stores both simple names ("x" → values) and flattened object
+	// property access ("obj.key" → values) for dot-notation and destructuring.
+	strVars := map[string][]string{}
+	for {
+		added := 0
+		walk(prog, func(n ast.Node) {
+			switch v := n.(type) {
+			case *ast.Binding:
+				added += e.firstPassBinding(v, strVars, dir, visited)
+			case *ast.AssignExpression:
+				id, ok := v.Left.(*ast.Identifier)
+				if !ok {
+					return
+				}
+				name := id.Name.String()
+				if _, exists := strVars[name]; exists {
+					return
+				}
+				if vals := stringValues(v.Right, strVars); len(vals) > 0 {
+					strVars[name] = vals
+					added++
+				}
+			case *ast.ForOfStatement:
+				name := loopVarName(v.Into)
+				if name == "" || strVars[name] != nil {
+					return
+				}
+				vals := stringValues(v.Source, strVars)
+				if len(vals) > 0 {
+					strVars[name] = vals
+					added++
+				} else if e.log != nil {
+					e.log.Debug("unresolvable loop source", "file", filePath)
+				}
+			}
+		})
+		if added == 0 {
+			break
+		}
+	}
+
+	// Second pass: find route registration calls whose first argument is a
+	// variable, template literal, or expression — resolve to all possible paths.
+	var routes []RoutePattern
+	walk(prog, func(n ast.Node) {
+		call, ok := n.(*ast.CallExpression)
+		if !ok || len(call.ArgumentList) == 0 {
+			return
+		}
+		method, ok := routeMethod(call.Callee)
+		if !ok {
+			return
+		}
+		first := call.ArgumentList[0]
+		// Plain string literals and no-interpolation template literals are already
+		// handled by the regex pass.
+		if _, isLit := first.(*ast.StringLiteral); isLit {
+			return
+		}
+		if tmpl, isTmpl := first.(*ast.TemplateLiteral); isTmpl && len(tmpl.Expressions) == 0 {
+			return
+		}
+		paths := stringValues(first, strVars)
+		if len(paths) == 0 && e.log != nil {
+			e.log.Debug("unresolvable route path", "file", filePath, "line", lineOf(prog, call.Idx0()))
+		}
+		for _, path := range paths {
+			if !strings.HasPrefix(path, "/") {
+				continue
+			}
+			routes = append(routes, RoutePattern{
+				Method: method,
+				Path:   path,
+				File:   filePath,
+				Line:   lineOf(prog, call.Idx0()),
+			})
+		}
+	})
+	return routes
+}
+
+// firstPassBinding processes a single *ast.Binding node during the first pass,
+// updating strVars with newly resolvable string values. Returns the number of
+// new entries added so the caller can detect fixpoint.
+func (e *RouteExtractor) firstPassBinding(v *ast.Binding, strVars map[string][]string, dir string, visited map[string]bool) int {
+	added := 0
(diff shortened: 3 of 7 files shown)


mariomac (commented):
## Review

Solid feature and good test coverage. I left detailed inline comments anchored to the specific lines — a quick map of what they cover:

**Correctness**
- Interpolated template literals → duplicate/conflicting routes (the regex pass already matches backtick strings). *Highest priority, and currently untested.*
- Possible slice-aliasing mutation in the `ConditionalExpression` branch of `stringValues`.
- Flat global symbol table (no scoping) → false positives / dropped routes.
- `routeMethod` does not constrain the receiver object.

**Dead / over-engineered**
- `visited`/`inner` recursion plumbing in `requireExports` is never used (cross-file is one level deep).
- Hand-rolled 177-line AST walker — check whether goja exposes `ast.Walk`.

**Performance**
- Each file is read and parsed twice.
- `.ts`/`.tsx` files are fully read and parse-attempted for nothing.
- Unbounded cartesian-product expansion with no cap.

**Dependency**
- New heavyweight dep (`dop251/goja`) pinned to an untagged pseudo-version; unrelated `Masterminds/semver` bump in `go.sum`.

The single most impactful fix is the interpolated-template duplication.

---
🤖 Posted by Claude (via Claude Code) on behalf of @mariomac.

pkg/internal/transform/route/harvest/js_ast.go
```diff
@@ -0,0 +1,495 @@
...
+		if !ok {
+			return
+		}
+		first := call.ArgumentList[0]
+		// Plain string literals and no-interpolation template literals are already
+		// handled by the regex pass.
+		if _, isLit := first.(*ast.StringLiteral); isLit {
+			return
+		}
+		if tmpl, isTmpl := first.(*ast.TemplateLiteral); isTmpl && len(tmpl.Expressions) == 0 {
```
mariomac, line 129: **Interpolated template literals produce duplicate/conflicting routes (highest priority).** The existing `Typical` regex (`js.go:116`) includes backtick (`\x60`) in its char class, so it already matches `app.get(`/api/${ver}/users`)` and emits the raw `/api/${ver}/users`, which `CleanupRegexPath` turns into `/api/:id/users`. This AST pass emits `/api/v1/users` for the *same* call — both survive dedup, so one registration yields two routes. Same with `${req.params.id}` → regex `:id` vs AST `:param`. This guard skips plain literals and non-interpolated templates, but not the interpolated case, which is exactly the new one. The tests don't catch it because `assertHasRoute` only checks presence; consider adding a "no unexpected routes" assertion.

pkg/internal/transform/route/harvest/js_ast.go
```diff
@@ -0,0 +1,495 @@
...
+			return vars[id.Name.String()+"."+v.Identifier.Name.String()]
+		}
+	case *ast.TemplateLiteral:
+		return templateValues(v, vars)
+	case *ast.BinaryExpression:
+		if v.Operator == token.PLUS {
+			return concatProduct(stringValues(v.Left, vars), stringValues(v.Right, vars))
+		}
+	case *ast.ConditionalExpression:
+		return append(stringValues(v.Consequent, vars), stringValues(v.Alternate, vars)...)
```
mariomac, line 378: **Possible slice-aliasing mutation.** When the consequent is an identifier, `stringValues` returns `vars[name]` directly (see the `*ast.Identifier` case), so `append(...)` here can mutate the map's backing array when there's spare capacity, corrupting that variable's stored values for later lookups. Safer to allocate a fresh slice before appending both branches.

pkg/internal/transform/route/harvest/js_ast.go
```diff
@@ -0,0 +1,495 @@
...
+		return nil
+	}
+
+	dir := filepath.Dir(filePath)
+
+	// First pass: collect all possible string values for each variable.
+	// Repeated until stable to resolve chained dependencies.
+	// strVars stores both simple names ("x" → values) and flattened object
+	// property access ("obj.key" → values) for dot-notation and destructuring.
+	strVars := map[string][]string{}
```
mariomac, line 72: **No variable scoping — flat global symbol table.** `strVars` is keyed by bare name across the whole file with first-wins semantics. Two functions each declaring `const p = '/a'` / `'/b'` silently drop one route, and an unrelated `p` can resolve to the wrong path. Probably acceptable for a best-effort harvester, but worth acknowledging as a known false-positive/missed-route source.

pkg/internal/transform/route/harvest/js_ast.go
```diff
@@ -0,0 +1,495 @@
...
+				exports[prop] = vals
+			}
+		}
+	})
+	return exports
+}
+
+// routeMethod reports whether callee is a method call like `app.get` /
+// `router.post` and, if so, returns the corresponding HTTP method.
+func routeMethod(callee ast.Expression) (string, bool) {
```
mariomac, line 341: **`routeMethod` doesn't constrain the receiver object.** Any `X.get/.post/.delete/.head/.options/.all(stringVar, …)` whose variable resolves to a string starting with `/` becomes a route — e.g. `cache.get(...)`, `map.get(...)`. The regex pass shares the shape but is bounded to literals physically present; variable resolution broadens the false-positive surface. The `HasPrefix(path, "/")` check is the only filter.

pkg/internal/transform/route/harvest/js_ast.go
```diff
@@ -0,0 +1,495 @@
...
+		src = s
+		resolved = candidate
+		break
+	}
+	if resolved == "" {
+		return nil
+	}
+
+	// Extend visited to prevent circular requires in deeper recursion.
+	inner := make(map[string]bool, len(visited)+1)
```
mariomac, line 282: **Dead recursion machinery.** This `inner` copy of `visited` is never used — `requireExports` only walks for `module.exports`/`exports.x` and does no nested `require` resolution, so cross-file is strictly one level deep and the circular-require protection is non-functional. Either complete the recursion or drop the `inner`/`visited` plumbing.

pkg/internal/transform/route/harvest/js_ast.go
```diff
@@ -0,0 +1,495 @@
...
+		} else {
+			results = concatProduct(results, vals)
+		}
+	}
+	return results
+}
+
+// concatProduct returns the pairwise concatenation of every element in lefts
+// with every element in rights — the cartesian product under string addition.
+func concatProduct(lefts, rights []string) []string {
```
mariomac, line 421: **Unbounded cartesian-product expansion.** `concatProduct` × `templateValues` × conditionals/arrays/chained vars can blow up combinatorially with no cap on result count or `strVars` slice sizes. A large (≤10MB) file could explode memory; a defensive cap would help.

pkg/internal/transform/route/harvest/js_ast.go
```diff
@@ -0,0 +1,495 @@
...
+			results[j] += static
+		}
+		if i >= len(tmpl.Expressions) {
+			continue
+		}
+		expr := tmpl.Expressions[i]
+		vals := stringValues(expr, vars)
+		if len(vals) == 0 {
+			placeholder := ":param"
+			if id, ok := expr.(*ast.Identifier); ok {
```
mariomac, line 406: **Placeholder naming inconsistency.** Unresolved interpolation becomes `:<identname>` for identifiers but `:param` otherwise, diverging from the regex pass's `:id`. This feeds the duplicate-route issue flagged on the literal/template guard above.

pkg/internal/transform/route/harvest/js_ast.go
```diff
@@ -0,0 +1,495 @@
...
+}
+
+// requireExports reads and parses a required JS file, returning its exported
+// string values keyed by export name. Only handles relative paths and the
+// common module.exports = {...} / exports.key = '...' patterns.
+func (e *RouteExtractor) requireExports(dir, reqPath string, visited map[string]bool) map[string][]string {
+	if !strings.HasPrefix(reqPath, "./") && !strings.HasPrefix(reqPath, "../") {
+		return nil
+	}
+	base := filepath.Join(dir, reqPath)
```
mariomac, line 261: **Minor:** `requireExports` joins arbitrary `./`/`../` paths. `openJSFileForScan` bounds it (regular file, ≤10MB, `O_NOFOLLOW`), so traversal risk is low, but the bare extension-less `base` candidate can read non-JS files before the parse fails.

pkg/internal/transform/route/harvest/js_ast_walk.go
```diff
@@ -0,0 +1,177 @@
...
+import (
+	"reflect"
+
+	"github.com/dop251/goja/ast"
+)
+
+// walk performs a depth-first traversal of the AST rooted at n, invoking fn for
+// every node visited (including n itself).
+func walk(n ast.Node, fn func(ast.Node)) {
+	if n == nil {
```
mariomac, line 15: **Hand-rolled AST walker (177 lines).** Does goja's `ast` package already expose `ast.Walk`/`Visitor`? If so this is redundant and a maintenance hazard — any node type the manual `children` switch misses silently drops a subtree (class bodies, spread, optional chaining, etc. are not handled).

pkg/internal/transform/route/harvest/js.go
```diff
@@ -473,6 +473,16 @@ func (e *RouteExtractor) scanFile(filePath string) error {
 		return err
 	}
 
+	// Fallback pass: resolve routes whose path comes from a variable or a
+	// template literal, which the line-based regex harvesters cannot match.
+	src, ok, err := readJSFileForScan(filePath)
```
mariomac, line 478: **Every file is read and parsed twice.** The regex pass above streams line-by-line; this AST pass then reads the whole file into a string (up to `MaxJSFileScanBytes` = 10MB) and full-parses it, without reusing the content already read.

pkg/internal/transform/route/harvest/js.go
```diff
@@ -473,6 +473,16 @@ func (e *RouteExtractor) scanFile(filePath string) error {
...
+	if err != nil {
+		return err
+	}
+	if ok {
+		e.routes = append(e.routes, e.resolveASTRoutes(filePath, src)...)
```
mariomac, line 483: **TypeScript files pay full cost for nothing.** `WalkJSFiles` includes `.ts`/`.tsx`, goja can't parse TS, so every TS file is read fully + parse-attempted + discarded (`resolveASTRoutes` returns nil on parse error). Consider gating the AST pass by extension or syntax.

mariomac (commented):
Inline observations below. The most impactful is the interpolated-template duplication on `js_ast.go` — it's a user-visible correctness regression and currently untested.

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/caarlos0/env/v11 v11.4.1
 	github.com/cilium/ebpf v0.20.0
 	github.com/containers/common v0.64.2
+	github.com/dop251/goja v0.0.0-20260618133527-c9b2ea77db59
```
mariomac, line 11: **New heavyweight dependency.** `dop251/goja` (a full JS interpreter, plus transitive `go-sourcemap/sourcemap`) is pulled in only for its parser/ast, and pinned to an untagged pseudo-version. Worth a sentence justifying the weight vs. a lighter JS parser. Separately, `go.sum` bumps `Masterminds/semver/v3` 3.4.0 → 3.5.0 — looks like an unrelated `go mod tidy` side effect; confirm it's intended here.

mariomac:
Thank you a lot for your contribution @Vedanshu7 ! I've asked Claude for a first review and it placed some inline comments.

As to be expected with AI, many observations might lack context and be incorrect. Feel free to not address all of them but please then reply with a justification to each item you won't address.

rafaelroquetto:
@mariomac review covered the technical issues, I want to step back to the approach itself.

Before going deeper on the details: is this the best approach, and what alternatives did you consider? IMHO #929 is really variables + interpolation, e.g. a `${x}` -> `:x` rewrite on the backtick paths the scanner already matches, and variables a small symbol table on top of the current line scan. Neither needs an AST or a new dependency.

- why goja + a hand-rolled walker over extending the existing scanner?
- were lighter options weighed, and what ruled them out?

I ask because whatever lands here becomes ours to maintain, and a generic AST visitor with reflect-based nil checks plus a fixpoint resolver is a lot to keep correct over time for what it adds to a best-effort heuristic. If it genuinely needs to be like this, than so be it, but I'd appreciate if you could expand on the reasins, otherwise perhaps a smaller, scanner-based version could suit us better here.

go.mod
grcevski: Yeah this is a good point, if implementing this means we have to bring in a full JS interpreter, perhaps we should wait until a customer reports an issue with our existing route parsing. So far there hasn't been any requests.

grcevski:
Given the complexity of this approach, I think we need to close this PR and figure out another approach. I'm not only concerned that we now have full blown JS interpreter as a dependency, but also about how much memory and CPU will it use while parsing some sort of minified JavaScript file.

Vedanshu7:
@grcevski @mariomac fair point, so I measured it instead of guessing.

Built `obi` for linux/amd64 with and without this PR, same flags. 122,161,792 bytes vs 121,741,525 bytes, about 410 KiB, 0.35% bigger. That's because we only import goja/ast, goja/file, goja/parser and goja/token, never the root goja package where the VM and eval live (43.7k of goja's 52.6k lines). Nothing here can execute JavaScript, it's a static parser producing an AST, similar in kind to pulling in a YAML parser. What we actually compile in is about 6,600 lines, no unsafe, no cgo.

Also, goja wasn't picked at random, it's the library suggested in the PR #924 review that this tracking issue came out of, used exactly as scoped there: a fallback only when the regex pass can't resolve a literal path.

Given the real footprint is closer to "add a parser" than "embed an interpreter," I'd lean toward keeping it, but happy to defer to you. I'll fix the correctness issues mariomac flagged either way. If the untagged dependency is a blocker on principle, that's understandable too.

Vedanshu7:
> @mariomac review covered the technical issues; I want to step back to the approach itself.
> 
Separate from the size/dependency question above, on the approach itself: fair, and honestly no, I didn't seriously weigh a scanner-based version first. I went straight to AST because I wanted full coverage, and goja was the library that the library already suggested in the #924 thread.

Looking at it again with your suggestion in mind, I think you're right. The template literal case is already caught by the existing Typical regex, so a `${expr}` -> `:name` rewrite on that captured string handles it directly and also fixes the duplicate-route bug in one move. Simple variable assignment is a first-pass symbol table over the line scan. Same-line string concatenation and array/ternary sources are regex-reachable too.

The harder cases (destructuring, cross-file requires, non-trivial loop sources) are where I reached for AST. Still, the current implementation doesn't actually handle those robustly either, so it's not clearly buying more real coverage than a scanner would, just more code and a new dependency to get there.

I'll rework this as a scanner + small symbol table on top of the existing line-based pass and drop the goja dependency. That also makes the earlier size/maintenance question moot. Will close the PR and implement it again.

PR #2515, opened 2026-06-26
Avoid eager MQTT response buffer access

MrAlias:
### Motivation

- A previous change eagerly dereferenced the response buffer via `rpkt.UnsafeView()` at function entry which can panic if `rpkt` is `nil` even when the request buffer (`pkt`) contains a valid MQTT packet. 
- The goal is to make `ProcessPossibleMQTTEvent` robust for callers that may pass a nil response buffer while preserving existing behavior when the response buffer is present.

### Description

- In `pkg/ebpf/common/mqtt_detect_transform.go` defer the response-buffer dereference and replace the eager `rpkt.UnsafeView()` with a lazy call inside the error path: `ProcessMQTTEvent(rpkt.UnsafeView())`.
- In `pkg/ebpf/common/mqtt_detect_transform_test.go` add a regression test case `PUBLISH in request buffer without response buffer` and adjust the test harness to pass a `nil` `*largebuf.LargeBuffer` when `response` is absent so the function is exercised with a `nil` response buffer.
- Applied formatting (`gofmt`) to the modified files.

### Testing

- `go test ./pkg/internal/ebpf/mqttparser`
- `go test ./pkg/internal/largebuf`
- `go test ./pkg/ebpf/common -run TestProcessPossibleMQTTEvent`


rafaelroquetto (approved):

grcevski (approved):
LGTM!

PR #2516, opened 2026-06-26
Limit SunRPC record fragments

MrAlias:
### Motivation

- Prevent a CPU and allocation amplification vector in the SunRPC-over-TCP parser where many tiny fragments could force unbounded loop iterations and slice growth.
- Bound parser work per record while preserving valid fragmented SunRPC records, including zero-byte record-marking fragments allowed by RFC 5531.

### Description

- Add a per-record fragment cap `maxRecordFragments = 64` and wire it into the SunRPC reader to reject pathological records early in `readRecord` in `pkg/internal/sunrpcparser/parser.go` by returning `ErrNotSunRPC` when the cap is exceeded.
- Accept empty non-final fragments within the fragment cap so legal RM records are still parsed while parser work remains bounded.
- Add unit tests in `pkg/internal/sunrpcparser/parser_test.go` to cover a valid fragmented `CALL`, rejection of too many fragments, and acceptance of empty non-final fragments.
- Strengthen the fragment-limit regression test so it uses zero-byte fragments followed by a valid final record, ensuring the failure is caused by the cap rather than an invalid reassembled payload.
- Add the `wrapTCPRecordFragments` test helper for constructing multi-fragment records.

### Testing

- `go test ./pkg/internal/sunrpcparser`

### DIFF
--- a/pkg/internal/sunrpcparser/parser.go
+++ b/pkg/internal/sunrpcparser/parser.go
@@ -38,8 +38,9 @@ const (
 	rmLastFrag = 0x80000000
 	rmFragLen  = 0x7fffffff
 
-	maxRecordSize   = 1 << 20
-	maxMessagesRead = 32
+	maxRecordSize      = 1 << 20
+	maxRecordFragments = 64
+	maxMessagesRead    = 32
 )
 
 var (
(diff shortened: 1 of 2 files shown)


grcevski (approved):
LGTM! This looks like a good check to add. I don't know much about the SunRPC protocol to judge if 64 fragments is too many or too few, I think it's reasonable.

grcevski:
I requested @damemi as a reviewer.

MrAlias:
> I don't know much about the SunRPC protocol to judge if 64 fragments is too many or too few, I think it's reasonable.

Yeah, not sure either. It seemed reasonable for OBI's current use, but it is heuristic, not a SunRPC-standard value. [RFC 5531](https://datatracker.ietf.org/doc/html/rfc5531#section-11) does not cap RM fragment count. For OBI’s current captured-buffer use, 64 is intentionally generous: generic TCP large buffers max out at 64 KiB, so this still allows roughly 1 KiB average fragments while bounding zero/tiny-fragment amplification. If we ever reuse this as a full-stream SunRPC parser, we may want a different limit or a configurable one. But seems reasonable for now.

damemi (approved):
this seems reasonable to me. I don't think we have enough anecdata to say for sure if this will hit any issue (someone sending tons of small fragments in a record?) but anything is better than unbounded

thanks for the ping!

PR #2517, opened 2026-06-26
Fix dynamic PID notification fan-out

MrAlias:
### Motivation

- The `DynamicPIDSelector` previously returned the same single-consumer channels from `AddedPIDsNotify` and `RemovedNotify`, making notification deliveries a work-queue instead of a broadcast and allowing competing consumers to steal add/remove batches.
- New network/stat filters and `DynamicAppIPs` trackers introduced additional subscribers which created races where the application matcher could miss remove notifications, leaving deselected PIDs instrumented and leaking telemetry.

### Description

- Replaced single shared notify channels with per-subscriber channels tracked on the notifier and implemented `addedNotify` and `removedNotify` to create a new buffered channel for each caller.
- Changed the drain loops to clone the subscriber list and fan out each PID batch to every subscriber, and to wait until at least one subscriber exists before sending.
- Updated `AddedPIDsNotify` and `RemovedNotify` to return per-subscriber channels via the new notifier methods.
- Updated `DynamicSignalProcessEventGate` to subscribe once before its select loop so repeated process events do not register orphaned notification subscribers.
- Added regression tests for notification fan-out and stable gate subscriptions.

### Testing

- `go test ./pkg/appolly/discover -run TestDynamicPIDSelector`

### DIFF
--- a/pkg/appolly/discover/dynamic_pid_selector.go
+++ b/pkg/appolly/discover/dynamic_pid_selector.go
@@ -93,6 +93,7 @@ func (s *dynamicPIDSubscriber) notify(batch []app.PID) {
 	default:
 	}
 
+	// Queue on the subscriber so a full subscriber channel cannot block notifier fan-out.
 	s.mu.Lock()
 	for _, pid := range batch {
 		if s.maxPending > 0 && len(s.pending) == s.maxPending {
@@ -139,6 +140,7 @@ func (s *dynamicPIDSubscriber) takePending() []app.PID {
 		return nil
 	}
 
+	// Send a stable batch while later notify calls append to a fresh pending queue.
 	batch := slices.Clone(s.pending)
 	s.pending = nil
 	return batch
--- a/pkg/appolly/discover/dynamic_pid_selector.go
+++ b/pkg/appolly/discover/dynamic_pid_selector.go
@@ -45,8 +45,8 @@ type dynamicPIDRecord struct {
 }
 
 type dynamicPIDNotifier struct {
-	addedCh   chan []app.PID
-	removedCh chan []app.PID
+	addedSubscribers   []chan []app.PID
+	removedSubscribers []chan []app.PID
 
 	addedPending []app.PID
 	addedMu      sync.Mutex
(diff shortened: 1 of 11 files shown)


MrAlias:
Each call to `AddedPIDsNotify` or `RemovedNotify` now allocates a new buffered channel and appends it to a notifier-owned subscriber slice. The notifier later broadcasts each PID batch by sending synchronously to every retained channel. Because there is no unsubscribe path and the per-subscriber buffer is only one element, any abandoned or slow subscriber can fill its channel. On the next notification batch, the drain goroutine blocks forever on that subscriber before reaching later active subscribers. While blocked, notifyAdded/notifyRemoved can continue appending to the pending slices, causing dynamic PID notifications to stop and pending PID batches to grow in memory. In-tree code mostly subscribes once, but this is still a regression in the public selector notification API and can also affect reused selectors after pipeline shutdown/restart because old subscriber channels remain registered.

Looking into a fix.

MrAlias:
cc @damemi @grcevski

pkg/appolly/discover/dynamic_pid_selector.go
```diff
@@ -57,11 +61,91 @@ type dynamicPIDNotifier struct {
...
+	for _, pid := range batch {
+		if s.maxPending > 0 && len(s.pending) == s.maxPending {
+			break
+		}
+		s.pending = append(s.pending, pid)
+	}
+	s.mu.Unlock()
+
+	select {
+	case s.wake <- struct{}{}:
```
mariomac, line 107: What's the advantage of waking up subscribers then cloning the pending slice in `takePending`and then forwarding the batch via `s.ch`?

Wouldn't it be simpler to just send a local copy of  the`pending` slice directly to the `s.ch` channel?

MrAlias: Good question. The extra hop is intentional as the backpressure boundary between notifier fan-out and each individual subscriber.

`subscriber.notify` is called from the notifier drain loop for every subscriber. If it sends directly to `s.ch`, a stale/full subscriber can block that drain loop and prevent later subscribers from receiving updates. With the wake path, `notify` only appends to that subscriber’s private pending queue and returns; the subscriber’s own goroutine can block on its `s.ch` without blocking fan-out to other subscribers.

`takePending` clones so the receiver owns a stable batch while new notifications can keep appending to `pending` under the lock. For legacy subscribers the pending queue is bounded; context subscribers can be canceled and cleaned up.

I agree this is not obvious from the code. I'll look into adding a short comment around `notify`/`run` to make the backpressure behavior clearer.

MrAlias: Done in 8c13d830ab. I added short comments around `notify` and `takePending` to make the backpressure boundary explicit: notifier fan-out only queues on each subscriber, while the subscriber goroutine owns the potentially blocking send and sends a stable cloned batch.

mariomac (approved):
Thanks!

PR #2518, opened 2026-06-26
Fix MCP tool call attribute selection

MrAlias:
### Motivation

- A recent change captured MCP tool-call arguments and results but these sensitive payloads were gated by broad GenAI input/output selection, preventing operators from independently including or excluding `gen_ai.tool.call.arguments` and `gen_ai.tool.call.result`.
- The goal is to restore per-attribute control so tool-call payloads cannot be exported implicitly when operators opt into other GenAI data.

### Description

- Add `attr.GenAIToolCallArguments` and `attr.GenAIToolCallResult` to the trace attribute selector definitions so they appear in `attributes.select` configuration (`pkg/export/attributes/attr_defs.go`).
- Change MCP span emission to gate tool-call arguments/results on their own optional attribute names instead of `GenAIInput`/`GenAIOutput` (`pkg/export/otel/tracesgen/tracesgen.go`).
- Add selector tests to verify inclusion/exclusion behavior for `gen_ai.tool.call.*` (`pkg/export/attributes/attr_selector_test.go`).
- Add trace tests to ensure `gen_ai.tool.call.arguments`/`gen_ai.tool.call.result` are not emitted via `GenAIInput`/`GenAIOutput` and are emitted when their own selectors are enabled (`pkg/export/otel/tracesgen/tracesgen_test.go`).

### Testing

- `go test ./pkg/export/attributes ./pkg/export/otel/tracesgen`

### DIFF
--- a/pkg/export/attributes/attr_defs.go
+++ b/pkg/export/attributes/attr_defs.go
@@ -408,13 +408,15 @@ func getDefinitions(
 				attr.GraphQLDocument: false,
 				// url.query is Conditionally Required by OTel semconv (emitted when a query string is present).
 				// You can opt out via attributes.select.traces.exclude: [url.query].
-				attr.HTTPUrlQuery:      true,
-				attr.GenAIInput:        false,
-				attr.GenAIOutput:       false,
-				attr.GenAIInstructions: false,
-				attr.GenAIMetadata:     false,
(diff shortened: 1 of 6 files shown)


grcevski (approved):
LGTM! Looks great

PR #2519, opened 2026-06-26
Resume HTTP/2 traceparent scan after invalid candidate

MrAlias:
### Motivation

- The HTTP/2 HPACK detector previously stopped at the first name/value-length fingerprint match and validated only that candidate.
- When that first candidate looked like `traceparent` but had an invalid value, OBI created and injected a new traceparent instead of continuing to a later valid one.
- Continuing after malformed candidates prevents duplicate traceparent injection while still adopting an existing valid context when one is present later in the HPACK block.

### Description

- Initialize the per-frame HPACK candidate scan offset when starting detection for a `HEADERS` frame.
- Retry malformed HPACK traceparent candidates by advancing past the full literal entry before scanning again:
  - `k_h2_tp_hpack_size` for plain `traceparent` names.
  - `k_h2_tp_hpack_huffman_size` for Huffman-encoded `traceparent` names.
- Count malformed-candidate retries against the existing HTTP/2 per-packet tail-call budget, preserving room for create/write tail calls.
- Keep the retry path verifier-friendly by shifting the pulled HPACK window for each retry instead of making the 192-byte scan loop skip to a saved scalar offset.
- Validate a found candidate from its pulled candidate window directly, so packet access stays at fixed offsets and avoids verifier state explosion on arm64.

### Testing

- `clang-format --dry-run --Werror bpf/tpinjector/tpinjector.c`
- `make generate`

### DIFF
--- a/bpf/tpinjector/tpinjector.c
+++ b/bpf/tpinjector/tpinjector.c
@@ -150,7 +150,7 @@ typedef struct tailcall_ctx {
     u32 h2_hpack_offset;          // start of HPACK bytes (after PADDED/PRIORITY prefix)
     u32 h2_hpack_len;             // HPACK length (frame payload minus prefix and trailing pad)
     u32 h2_scan_pos;              // resume offset for detect_h2 across tail calls
-    u32 h2_tp_candidate_pos;      // HPACK candidate scan offset (>= k_h2_max_hpack_scan = none)
+    u32 h2_tp_candidate_pos;      // HPACK scan start or found candidate offset (>= max = none)
     u8 niter;                     // HTTP/1 find-existing scan iteration counter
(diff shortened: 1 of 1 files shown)


MrAlias:
Looking into verifier failures 👀

MrAlias:
@rafaelroquetto please take a look at this one. I think this should address the issue, but not sure if my workaround the verifier limitations is optimal.

grcevski (approved):
LGTM! This looks great!

I originally thought that finding invalid TP when we see 'traceparent' would be very rare and questionable on what we should do, I realised though that this infrastructure you added will help us what we want to do next, find traceparents when they are hpack encoded. In those situations, we will likely need to loop further with failed validation. So I think this is pretty cool.

PR #2520, opened 2026-06-26
Fix chunked response size overflow

MrAlias:
### Motivation

- The manual chunked HTTP decoder parsed the chunk size as a signed `int64` and added it to a slice index without checking for overflow, allowing an attacker-controlled large chunk size to cause a runtime panic and crash the agent.

### Description

- Parse the chunk-size line with `strconv.ParseUint` and remove the signed-negative check to avoid incorrect casting.
- Compute `available := len(data) - chunkStart` and treat any `chunkSize` greater than `available` as a truncated chunk before converting to `int` and slicing.
- Add a regression test in `TestDechunkBody` that asserts an oversized advertised chunk (`"7fffffffffffffff\r\nA"`) is treated as truncated and returns the available bytes.

### Testing

- `go test ./pkg/ebpf/common -run 'TestDechunkBody|TestHttpSafeParseResponseChunked'`

### DIFF
--- a/pkg/ebpf/common/http_transform.go
+++ b/pkg/ebpf/common/http_transform.go
@@ -466,8 +466,8 @@ func dechunkBody(data []byte) []byte {
 			break
 		}
 
-		chunkSize, err := strconv.ParseInt(sizeLine, 16, 64)
-		if err != nil || chunkSize < 0 {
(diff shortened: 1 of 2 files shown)


rafaelroquetto (approved):

grcevski (approved):
LGTM!

PR #2526, opened 2026-06-28
Weaver: document undocumented produced attributes + violate on extends_namespace

NimrodAvni78:
## Summary

OBI produces a few attributes which are not part of the semconv. weaver does not take it as a violation since its under a known namespace. this both fixes this and adds `extends_namespace` advisories to produce weaver errors. and adds the undocumented attributes to OBIs schema registry

## Validation

### DIFF
--- a/internal/test/integration/weaver.go
+++ b/internal/test/integration/weaver.go
@@ -65,7 +65,7 @@ var weaverIgnoredAdviceMessages = map[string]struct{}{
 	"Namespace 'iface' collides with existing attribute 'iface.direction'": {},
 }
 
-// actionableAdviceTypes lists the weaver `advice_type` values OBI treats as
+// actionableAdviceTypes lists the weaver finding-type values OBI treats as
 // failures in addition to `violation`-level advice. Hoisted here (rather than
 // matched as an inline string literal) so the coupling to weaver's advice-type
 // vocabulary lives in one documented place and is easy to extend.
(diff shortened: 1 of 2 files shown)


mariomac (approved):
👌🏻

internal/test/integration/weaver.go
```diff
@@ -96,6 +114,7 @@ type weaverStatistics struct {
 type weaverAdvice struct {
 	Message    string `json:"message"`
 	Level      string `json:"level"`
+	AdviceType string `json:"advice_type"`
```
MrAlias: This currently decodes the legacy policy input field rather than the field weaver writes in the final live-check report. In weaver v0.24.1, `live_check_result.all_advice` is serialized from `PolicyFinding`, whose machine-readable finding type is `id`; `advice_type` only exists in the Rego output before the custom deserializer maps it into `id`.

Because of that, `AdviceType` stays empty here, so the `actionableAdviceTypes` check never catches `extends_namespace` information advisories. I think this should decode `json:"id"` instead, or rename the Go field to something like `FindingID` and compare that. A small unit test around `validateWeaverReport` with an `all_advice` entry containing `id: "extends_namespace"` would catch this regression directly; the current live checks only prove the fixed registry has no such advisory left.

```suggestion
	AdviceType string `json:"id"`
```

MrAlias (changes requested):
Thanks for tightening the weaver validation path and adding the missing OBI registry attributes. I found one issue in the new enforcement logic that keeps the `extends_namespace` advisory from becoming actionable.

internal/test/integration/weaver.go
NimrodAvni78: thanks, fixed!

mmat11 (approved):

MrAlias (approved):

PR #2530, opened 2026-06-29
feat: add Gemini SSE streaming response parsing

NameHaibinZhang:
#2394 

#1854 

Implement parseGeminiStream() to handle Server-Sent Events responses from Gemini's streamGenerateContent endpoint. This adds streaming support comparable to what exists for OpenAI and Anthropic providers.

Changes:
- New gemini_stream.go with SSE parsing and text/functionCall aggregation
- Modified GeminiSpan() to dispatch between JSON and SSE responses
- Comprehensive test coverage for streaming scenarios

## Summary

Describe the change briefly.

## Validation

### DIFF
--- a/pkg/ebpf/common/http/gemini_stream.go
+++ b/pkg/ebpf/common/http/gemini_stream.go
@@ -36,7 +36,9 @@ type geminiStreamContent struct {
 }
 
 type geminiTextPart struct {
-	Text string `json:"text"`
+	Text             string `json:"text"`
+	Thought          bool   `json:"thought,omitempty"`
+	ThoughtSignature string `json:"thoughtSignature,omitempty"`
 }
 
 type geminiFunctionCallPart struct {
@@ -44,15 +46,33 @@ type geminiFunctionCallPart struct {
 }
 
 type geminiFunctionCallData struct {
-	Name string          `json:"name"`
-	Args json.RawMessage `json:"args,omitempty"`
+	Name         string          `json:"name"`
+	Args         json.RawMessage `json:"args,omitempty"`
+	PartialArgs  string          `json:"partialArgs,omitempty"`
+	WillContinue *bool           `json:"willContinue,omitempty"`
+}
+
+// geminiStreamError represents a bare error envelope that Gemini may send on
+// an HTTP 200 stream (as observed by the official Go client).
+type geminiStreamError struct {
+	Error *request.GeminiError `json:"error"`
 }
 
 // candidatePart is a single ordered part in a candidate's content.
-// Either text is non-empty (text part) or fcRaw is non-nil (function-call part).
+// Either textBuilder is non-nil (text part) or fcRaw is non-nil (function-call part).
 type candidatePart struct {
-	text  string
-	fcRaw json.RawMessage
+	textBuilder *strings.Builder
+	thought     bool
+	fcRaw       json.RawMessage
+}
+
+// fcAggregator accumulates streaming function call arguments for Vertex AI's
+// streamFunctionCallArguments feature where args arrive across multiple chunks.
+type fcAggregator struct {
+	name       string
+	argsAccum  strings.Builder
+	hasFullArg bool            // true if args came as a complete JSON object
+	fullArg    json.RawMessage // stored when args is a complete JSON object
 }
 
 // candidateAggregator accumulates streamed parts for a single candidate
@@ -60,6 +80,56 @@ type candidatePart struct {
 type candidateAggregator struct {
 	parts        []candidatePart
 	finishReason string
+	// activeFC tracks a function call being built across multiple stream chunks
+	// (Vertex AI streamFunctionCallArguments).
+	activeFC *fcAggregator
+}
+
+// flushActiveFC finalizes any in-progress function call aggregation and
+// appends the result to the candidate's parts list. Returns the function
+// call name for toolCalls tracking (empty if nothing was flushed).
+func (ca *candidateAggregator) flushActiveFC() string {
+	if ca.activeFC == nil {
+		return ""
+	}
+	fc := ca.activeFC
+	ca.activeFC = nil
+
+	if fc.name == "" {
+		return ""
+	}
+
+	var raw json.RawMessage
+	if fc.hasFullArg {
+		// Complete args arrived as a JSON object.
+		raw = buildFunctionCallRaw(fc.name, fc.fullArg)
+	} else if fc.argsAccum.Len() > 0 {
+		// Partial args were accumulated as string fragments.
+		raw = buildFunctionCallRaw(fc.name, json.RawMessage(fc.argsAccum.String()))
+	} else {
+		// Name-only function call with no args.
+		raw = buildFunctionCallRaw(fc.name, nil)
+	}
+
+	ca.parts = append(ca.parts, candidatePart{fcRaw: raw})
+	return fc.name
+}
+
+// buildFunctionCallRaw constructs the raw JSON for a function call part.
+func buildFunctionCallRaw(name string, args json.RawMessage) json.RawMessage {
+	type fcData struct {
+		Name string          `json:"name"`
+		Args json.RawMessage `json:"args,omitempty"`
+	}
+	type fcWrapper struct {
+		FunctionCall fcData `json:"functionCall"`
+	}
+	w := fcWrapper{FunctionCall: fcData{Name: name, Args: args}}
+	raw, err := json.Marshal(w)
+	if err != nil {
+		return nil
+	}
+	return raw
 }
 
 // parseGeminiStream parses the SSE stream from Gemini APIs and returns
@@ -69,6 +139,10 @@ type candidateAggregator struct {
 // intentionally only support this observed single-line framing
 // (consistent with the OpenAI SSE parser) rather than the full
 // multi-line SSE spec.
+//
+// Error handling: bare {"error": ...} records on an HTTP 200 stream
+// are treated as API errors (consistent with the official Google Gen AI
+// Go client behavior).
 func parseGeminiStream(reader io.Reader) (*request.GeminiResponse, []request.ToolCall) {
 	scanner := bufio.NewScanner(reader)
 	scanner.Buffer(make([]byte, 0, 256*1024), 256*1024)
@@ -78,12 +152,17 @@ func parseGeminiStream(reader io.Reader) (*request.GeminiResponse, []request.Too
 	var modelVersion string
 	var responseID string
 	var usage *request.GeminiUsage
+	var streamError *request.GeminiError
 
 	for scanner.Scan() {
 		line := scanner.Text()
 
 		data, ok := extractSSEData(line)
 		if !ok {
+			// Check for bare error envelope (not wrapped in "data:" prefix).
+			if err := tryParseErrorLine(line); err != nil {
+				streamError = err
+			}
 			continue
 		}
 
@@ -91,6 +170,14 @@ func parseGeminiStream(reader io.Reader) (*request.GeminiResponse, []request.Too
 		if err := json.Unmarshal([]byte(data), &chunk); err != nil {
 			slog.Debug("parseGeminiStream: failed to parse chunk", "error", err)
 			continue
+		}
+
+		// Check if this data line is itself an error envelope.
+		if chunk.Candidates == nil && chunk.UsageMetadata == nil {
+			if err := tryParseErrorLine(data); err != nil {
+				streamError = err
+				continue
+			}
 		}
 
 		if chunk.ModelVersion != "" {
@@ -124,22 +211,26 @@ func parseGeminiStream(reader io.Reader) (*request.GeminiResponse, []request.Too
 			for _, rawPart := range c.Content.Parts {
 				var textPart geminiTextPart
 				if err := json.Unmarshal(rawPart, &textPart); err == nil && textPart.Text != "" {
-					if n := len(agg.parts); n > 0 && agg.parts[n-1].fcRaw == nil {
-						agg.parts[n-1].text += textPart.Text
-					} else {
-						agg.parts = append(agg.parts, candidatePart{text: textPart.Text})
+					// Flush any active function call before appending text.
+					if name := agg.flushActiveFC(); name != "" {
+						toolCalls = append(toolCalls, request.ToolCall{Name: name})
 					}
+					appendTextPart(agg, textPart)
 					continue
 				}
 
 				var fcPart geminiFunctionCallPart
-				if err := json.Unmarshal(rawPart, &fcPart); err == nil && fcPart.FunctionCall != nil && fcPart.FunctionCall.Name != "" {
-					toolCalls = append(toolCalls, request.ToolCall{
-						Name: fcPart.FunctionCall.Name,
-					})
-					agg.parts = append(agg.parts, candidatePart{fcRaw: rawPart})
+				if err := json.Unmarshal(rawPart, &fcPart); err == nil && fcPart.FunctionCall != nil {
+					processFunctionCallPart(agg, fcPart.FunctionCall, &toolCalls)
 				}
 			}
+		}
+	}
+
+	// Flush any remaining active function calls.
+	for _, agg := range candidates {
+		if name := agg.flushActiveFC(); name != "" {
+			toolCalls = append(toolCalls, request.ToolCall{Name: name})
 		}
 	}
 
@@ -155,10 +246,108 @@ func parseGeminiStream(reader io.Reader) (*request.GeminiResponse, []request.Too
 	if usage != nil {
 		resp.UsageMetadata = *usage
 	}
+	if streamError != nil {
+		resp.Error = streamError
+	}
 
 	resp.Candidates = buildGeminiCandidates(candidates)
 
 	return resp, toolCalls
+}
+
+// appendTextPart adds a text fragment to the candidate aggregator, coalescing
+// only with the previous part when both are text parts with equivalent metadata
+// (thought flag). Uses strings.Builder for efficient concatenation.
+func appendTextPart(agg *candidateAggregator, tp geminiTextPart) {
+	n := len(agg.parts)
+	// Coalesce with previous text part only if metadata (thought) matches.
+	if n > 0 && agg.parts[n-1].textBuilder != nil && agg.parts[n-1].thought == tp.Thought {
+		agg.parts[n-1].textBuilder.WriteString(tp.Text)
+		return
+	}
+	b := &strings.Builder{}
+	b.WriteString(tp.Text)
+	agg.parts = append(agg.parts, candidatePart{
+		textBuilder: b,
+		thought:     tp.Thought,
+	})
+}
+
+// processFunctionCallPart handles a function call part, supporting both
+// complete single-chunk calls and Vertex AI's streaming function call
+// arguments where the name arrives first and args follow in subsequent chunks.
+func processFunctionCallPart(agg *candidateAggregator, fc *geminiFunctionCallData, toolCalls *[]request.ToolCall) {
+	if fc.Name != "" {
+		// New named function call: flush any previous active FC.
+		if name := agg.flushActiveFC(); name != "" {
+			*toolCalls = append(*toolCalls, request.ToolCall{Name: name})
+		}
+
+		// If the call has complete args and no continuation, store directly.
+		if len(fc.Args) > 0 && fc.PartialArgs == "" {
+			agg.activeFC = &fcAggregator{
+				name:       fc.Name,
+				hasFullArg: true,
+				fullArg:    fc.Args,
+			}
+			// If willContinue is nil or false, this is a complete call.
+			if fc.WillContinue == nil || !*fc.WillContinue {
+				if name := agg.flushActiveFC(); name != "" {
+					*toolCalls = append(*toolCalls, request.ToolCall{Name: name})
+				}
+			}
+			return
+		}
+
+		// Start aggregation (name only, or name with partial args).
+		agg.activeFC = &fcAggregator{name: fc.Name}
+		if fc.PartialArgs != "" {
+			agg.activeFC.argsAccum.WriteString(fc.PartialArgs)
+		}
+
+		// If no continuation expected and no partial args, flush immediately.
+		if fc.WillContinue == nil && fc.PartialArgs == "" && len(fc.Args) == 0 {
+			if name := agg.flushActiveFC(); name != "" {
+				*toolCalls = append(*toolCalls, request.ToolCall{Name: name})
+			}
+		}
+		return
+	}
+
+	// Continuation fragment (no name): append to active function call.
+	if agg.activeFC == nil {
+		return
+	}
+	if fc.PartialArgs != "" {
+		agg.activeFC.argsAccum.WriteString(fc.PartialArgs)
+	} else if len(fc.Args) > 0 {
+		// Args as JSON in continuation.
+		agg.activeFC.argsAccum.Write(fc.Args)
+	}
+
+	// If willContinue is explicitly false or absent, the call is complete.
+	if fc.WillContinue == nil || !*fc.WillContinue {
+		if name := agg.flushActiveFC(); name != "" {
+			*toolCalls = append(*toolCalls, request.ToolCall{Name: name})
+		}
+	}
+}
+
+// tryParseErrorLine attempts to parse a line as a bare Gemini error envelope.
+// Returns the error if found, nil otherwise.
+func tryParseErrorLine(line string) *request.GeminiError {
+	line = strings.TrimSpace(line)
+	if line == "" || line[0] != '{' {
+		return nil
+	}
+	var envelope geminiStreamError
+	if err := json.Unmarshal([]byte(line), &envelope); err != nil {
+		return nil
+	}
+	if envelope.Error != nil && (envelope.Error.Code != 0 || envelope.Error.Status != "" || envelope.Error.Message != "") {
+		return envelope.Error
+	}
+	return nil
 }
 
 // extractSSEData extracts the JSON payload from an SSE data line.
@@ -209,14 +398,10 @@ func buildGeminiCandidates(aggs map[int]*candidateAggregator) []request.GeminiCa
 
 // buildGeminiStreamParts constructs the parts JSON from ordered candidate
 // parts, preserving the original text/function-call ordering and coalescing
-// only adjacent text fragments.
+// only adjacent text fragments with equivalent metadata (thought flag).
 func buildGeminiStreamParts(parts []candidatePart) json.RawMessage {
 	if len(parts) == 0 {
 		return nil
-	}
-
-	type textPartJSON struct {
-		Text string `json:"text"`
 	}
 
 	var rawParts []json.RawMessage
@@ -225,9 +410,9 @@ func buildGeminiStreamParts(parts []candidatePart) json.RawMessage {
 			rawParts = append(rawParts, p.fcRaw)
 			continue
 		}
-		if p.text != "" {
-			raw, err := json.Marshal(textPartJSON{Text: p.text})
-			if err == nil {
+		if p.textBuilder != nil && p.textBuilder.Len() > 0 {
+			raw := marshalTextPart(p.textBuilder.String(), p.thought)
+			if raw != nil {
 				rawParts = append(rawParts, raw)
 			}
 		}
@@ -238,6 +423,29 @@ func buildGeminiStreamParts(parts []candidatePart) json.RawMessage {
 	}
 
 	raw, err := json.Marshal(rawParts)
+	if err != nil {
+		return nil
+	}
+	return raw
+}
+
+// marshalTextPart marshals a text part, including thought metadata if present.
+func marshalTextPart(text string, thought bool) json.RawMessage {
+	if thought {
+		type thoughtPartJSON struct {
+			Text    string `json:"text"`
+			Thought bool   `json:"thought"`
+		}
+		raw, err := json.Marshal(thoughtPartJSON{Text: text, Thought: true})
+		if err != nil {
+			return nil
+		}
+		return raw
+	}
+	type textPartJSON struct {
+		Text string `json:"text"`
+	}
+	raw, err := json.Marshal(textPartJSON{Text: text})
 	if err != nil {
 		return nil
 	}
--- a/pkg/ebpf/common/http/gemini_stream.go
+++ b/pkg/ebpf/common/http/gemini_stream.go
@@ -25,9 +25,10 @@ type geminiStreamChunk struct {
 }
 
 type geminiStreamCandidate struct {
-	Index        int                  `json:"index"`
-	Content      *geminiStreamContent `json:"content"`
-	FinishReason string               `json:"finishReason"`
+	Index         int                  `json:"index"`
+	Content       *geminiStreamContent `json:"content"`
+	FinishReason  string               `json:"finishReason"`
+	SafetyRatings json.RawMessage      `json:"safetyRatings,omitempty"`
 }
 
 type geminiStreamContent struct {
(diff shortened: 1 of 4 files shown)


mariomac (approved):
Looks good! Thanks for your contribution

pkg/ebpf/common/http/gemini_stream.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+		if chunk.UsageMetadata != nil && chunk.UsageMetadata.TotalTokenCount > 0 {
+			usage = chunk.UsageMetadata
+		}
+
+		if len(chunk.Candidates) == 0 {
+			continue
+		}
+
+		candidate := &chunk.Candidates[0]
```
MrAlias: Gemini streaming returns `GenerateContentResponse` chunks, and those chunks can contain more than one candidate. This only processes `Candidates[0]`, then the parser later reconstructs a single output candidate.

If a request uses `generationConfig.candidateCount > 1`, we will drop the other candidates' text, finish reasons, and tool calls from the exported GenAI attributes. Can we keep per-candidate aggregation here, ideally keyed by the candidate `index`, and rebuild all observed candidates in the final `GeminiResponse`?

pkg/ebpf/common/http/gemini_stream.go
```diff
@@ -0,0 +1,168 @@
...
+type geminiStreamContent struct {
+	Parts []json.RawMessage `json:"parts"`
+	Role  string            `json:"role"`
+}
+
+type geminiTextPart struct {
+	Text string `json:"text"`
+}
+
+type geminiFunctionCallPart struct {
```
MrAlias: This reduces streamed `functionCall` parts to only the function name. Gemini function calls can also include arguments, and our non-streaming Gemini path preserves those arguments through `Content.Parts` so they show up in the normalized `gen_ai.output.messages` value.

With the current streaming path, a tool call like `get_weather({"location":"NYC"})` becomes only `get_weather`, which makes streamed Gemini spans less informative than non-streamed ones. Can we preserve the raw function-call part, or at least carry the function-call arguments through when rebuilding the response parts?

pkg/ebpf/common/http/gemini_stream.go
```diff
@@ -0,0 +1,168 @@
...
+			continue
+		}
+
+		if chunk.ModelVersion != "" {
+			modelVersion = chunk.ModelVersion
+		}
+		if chunk.ResponseID != "" {
+			responseID = chunk.ResponseID
+		}
+		if chunk.UsageMetadata != nil && chunk.UsageMetadata.TotalTokenCount > 0 {
```
MrAlias: This only keeps `usageMetadata` when `totalTokenCount > 0`. For export we mainly use `promptTokenCount` and `candidatesTokenCount`, so dropping the whole usage block because `totalTokenCount` is absent or zero can still lose useful input/output token counts.

Can we accept usage metadata when any of the token fields we export are present, rather than gating it only on `totalTokenCount`?

pkg/ebpf/common/http/gemini_stream.go
```diff
@@ -0,0 +1,168 @@
...
+	var toolCalls []request.ToolCall
+	var finishReason string
+	var modelVersion string
+	var responseID string
+	var usage *request.GeminiUsage
+
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		if !strings.HasPrefix(line, "data: ") {
```
MrAlias: This parser only accepts SSE lines with the exact `data: ` prefix. Gemini REST streaming uses SSE, where the space after `data:` is optional, and an event can contain multiple `data:` lines that should be joined before parsing the event payload.

If Google always emits a single JSON object as `data: {...}` this works in practice, but this is narrower than the SSE format we are claiming to parse. Can we make this parse the SSE `data` field more directly, or add a comment/test documenting that we intentionally support only Gemini's observed one-line framing?

pkg/ebpf/common/http/gemini_stream_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	var parts []struct {
+		Text string `json:"text"`
+	}
+	require.NoError(t, json.Unmarshal(resp.Candidates[0].Content.Parts, &parts))
+	require.Len(t, parts, 1)
+	assert.Equal(t, "Hello world", parts[0].Text)
+}
+
+func TestParseGeminiStream_ToolCalls(t *testing.T) {
```
MrAlias, line 64: The current tests cover the basic text-stream path and a function-call name, but they do not cover the Gemini response-shape cases above: multiple candidates, function-call arguments, partial usage metadata, or SSE framing variants.

Can we add tests for at least the multi-candidate and function-call-arguments cases? Those are the cases where the current parser loses valid Gemini response data.

MrAlias (changes requested):
Thanks for adding Gemini streaming support. The overall integration direction matches what we do for the other GenAI providers, but the stream parser is currently a bit too lossy for Gemini's response shape. I think we should tighten the parser and add coverage for those cases before merging.

pkg/ebpf/common/http/gemini_stream.go
```diff
@@ -0,0 +1,168 @@
...
+
+		if chunk.ModelVersion != "" {
+			modelVersion = chunk.ModelVersion
+		}
+		if chunk.ResponseID != "" {
+			responseID = chunk.ResponseID
+		}
+		if chunk.UsageMetadata != nil && chunk.UsageMetadata.TotalTokenCount > 0 {
+			usage = chunk.UsageMetadata
+		}
```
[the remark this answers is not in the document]
NameHaibinZhang: Fixed. Usage metadata is now accepted when any of the exported token fields (`promptTokenCount`, `candidatesTokenCount`, or `totalTokenCount`) are non-zero, via a new `geminiUsageHasTokens()` helper.

pkg/ebpf/common/http/gemini_stream.go
```diff
@@ -0,0 +1,168 @@
...
+
+		if len(chunk.Candidates) == 0 {
+			continue
+		}
+
+		candidate := &chunk.Candidates[0]
+		if candidate.FinishReason != "" {
```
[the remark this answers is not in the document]
NameHaibinZhang: Fixed. The parser now aggregates per-candidate by index using a `map[int]*candidateAggregator`, and rebuilds all observed candidates in the final response.

pkg/ebpf/common/http/gemini_stream.go
NameHaibinZhang: Fixed. Added an `Index` field to `geminiStreamCandidate` and switched to per-index aggregation using `map[int]*candidateAggregator`. The final response now rebuilds all observed candidates ordered by index. Added `TestParseGeminiStream_MultipleCandidates` covering the two-candidate case.

pkg/ebpf/common/http/gemini_stream.go
NameHaibinZhang: Fixed. The streaming parser now preserves the entire raw `functionCall` JSON part (including `args`) when rebuilding the response. `buildGeminiStreamParts` appends the original `json.RawMessage` rather than reconstructing with only the name. Added `TestParseGeminiStream_FunctionCallArguments` that verifies args are preserved in the output.

pkg/ebpf/common/http/gemini_stream.go
NameHaibinZhang: Fixed. Introduced `geminiUsageHasTokens()` that returns true when any of `PromptTokenCount`, `CandidatesTokenCount`, or `TotalTokenCount` is non-zero. Added `TestParseGeminiStream_PartialUsageMetadata` covering the prompt-only case.

pkg/ebpf/common/http/gemini_stream.go
NameHaibinZhang: Fixed both: extracted an `extractSSEData()` helper that handles both `"data: "` and `"data:"` (without space) prefixes. Added a doc comment on `parseGeminiStream` documenting that we intentionally support only the observed single-line framing (consistent with the OpenAI SSE parser). Added `TestParseGeminiStream_DataPrefixWithoutSpace` to cover the no-space case.

pkg/ebpf/common/http/gemini_stream_test.go
NameHaibinZhang: Done. Added four new tests: `TestParseGeminiStream_MultipleCandidates` (two candidates aggregated by index), `TestParseGeminiStream_FunctionCallArguments` (args preserved in output), `TestParseGeminiStream_PartialUsageMetadata` (prompt-only usage), and `TestParseGeminiStream_DataPrefixWithoutSpace` (SSE framing variant).

NameHaibinZhang (commented):
done

pkg/ebpf/common/http/gemini_stream.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	maxIdx := 0
+	for idx := range aggs {
+		if idx > maxIdx {
+			maxIdx = idx
+		}
+	}
+
+	result := make([]request.GeminiCandidate, maxIdx+1)
```
MrAlias, line 585: `geminiStreamCandidate.Index` is decoded directly from the Gemini HTTP response body, so it is external input rather than a value we control. We then use it both to size `result` and as an index into that slice. A malformed response with a negative index will panic at `result[idx]`, while a very large index can force an excessive allocation. We already guard equivalent response-provided indices in `openai_stream.go`; can we apply the same defensive bound here and add regression cases for negative and oversized values?

```suggestion
// maxGeminiStreamCandidates bounds allocations derived from response indices.
const maxGeminiStreamCandidates = 256

// buildGeminiCandidates constructs the final candidate list from the
// per-index aggregators, ordered by candidate index.
func buildGeminiCandidates(aggs map[int]*candidateAggregator) []request.GeminiCandidate {
	maxIdx := 0
	for idx := range aggs {
		if idx < 0 || idx >= maxGeminiStreamCandidates {
			delete(aggs, idx)
			continue
		}
		if idx > maxIdx {
			maxIdx = idx
		}
	}
	if len(aggs) == 0 {
		return nil
	}

	result := make([]request.GeminiCandidate, maxIdx+1)
```

pkg/ebpf/common/http/gemini_stream.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	var parts []json.RawMessage
+	if text != "" {
+		raw, err := json.Marshal(textPartJSON{Text: text})
+		if err == nil {
+			parts = append(parts, raw)
+		}
+	}
+	parts = append(parts, fcParts...)
```
MrAlias: One more consequence of splitting text and function calls into separate accumulators is that we lose the original `Content.Parts` ordering. `buildGeminiStreamParts` always emits all text first and then appends the calls, so `functionCall → text` or `text → functionCall → text` is exported differently from the non-streaming response. Can we keep the per-candidate parts ordered and only coalesce adjacent text fragments? An interleaved regression test would help pin that behavior down.

MrAlias (commented):
Thanks for the updates. The earlier issues around multiple candidates, function-call arguments, partial usage metadata, and the `data:` framing variants are addressed. I took another pass through the revised aggregation logic and left two follow-up comments inline.

pkg/ebpf/common/http/gemini_stream.go
NameHaibinZhang: Fixed. Added `maxGeminiStreamCandidates = 256` constant and a bounds check in `parseGeminiStream` that skips candidates with `index < 0 || index >= maxGeminiStreamCandidates`, consistent with the `openai_stream.go` guard. Added `TestParseGeminiStream_NegativeCandidateIndex` and `TestParseGeminiStream_OversizedCandidateIndex` regression tests covering both cases.

pkg/ebpf/common/http/gemini_stream.go
NameHaibinZhang: Fixed. Refactored `candidateAggregator` to use an ordered `[]candidatePart` list instead of separate text/function-call accumulators. `buildGeminiStreamParts` now serializes parts in their original order, coalescing only adjacent text fragments. Added `TestParseGeminiStream_InterleavedTextAndFunctionCall` covering the `text → functionCall → text` case.

pkg/ebpf/common/http/gemini_stream.go
```diff
@@ -0,0 +1,245 @@
...
+			}
+			if c.Content == nil {
+				continue
+			}
+
+			for _, rawPart := range c.Content.Parts {
+				var textPart geminiTextPart
+				if err := json.Unmarshal(rawPart, &textPart); err == nil && textPart.Text != "" {
+					if n := len(agg.parts); n > 0 && agg.parts[n-1].fcRaw == nil {
+						agg.parts[n-1].text += textPart.Text
```
MrAlias: This uses repeated string concatenation for every chunk, which copies all text accumulated so far. At the 256 KiB capture ceiling this becomes quadratic allocator and CPU work; a roughly 250 KB valid SSE fixture allocated about 143 MB in the parser.

`openai_stream.go` already uses `strings.Builder` for streamed content. Can we use a builder or buffer per adjacent text run here as well, while preserving the part ordering?

pkg/ebpf/common/http/gemini_stream.go
```diff
@@ -0,0 +1,245 @@
...
+	var toolCalls []request.ToolCall
+	var modelVersion string
+	var responseID string
+	var usage *request.GeminiUsage
+
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		data, ok := extractSSEData(line)
+		if !ok {
```
MrAlias, line 347: This skips every non-`data:` line. The official Google Gen AI Go client also treats a bare `{"error": ...}` record on an HTTP 200 stream as an API error ([reader](https://github.com/googleapis/go-genai/blob/2935d983a2845cde26b2e26cc59e0d3f0597c535/api_client.go#L444-L515), [regression case](https://github.com/googleapis/go-genai/blob/2935d983a2845cde26b2e26cc59e0d3f0597c535/api_client_test.go#L429-L438)). If that record arrives after normal data chunks, `GeminiSpan` has already selected the SSE path and we silently drop it. `Output.Error` remains nil, so `HTTPSpanStatusCode` reports the failed call as unset instead of error.

Can we parse the error envelope here and add a regression case with a valid data chunk followed by the HTTP-200 error record?

pkg/ebpf/common/http/gemini_stream.go
```diff
@@ -0,0 +1,245 @@
...
+					if n := len(agg.parts); n > 0 && agg.parts[n-1].fcRaw == nil {
+						agg.parts[n-1].text += textPart.Text
+					} else {
+						agg.parts = append(agg.parts, candidatePart{text: textPart.Text})
+					}
+					continue
+				}
+
+				var fcPart geminiFunctionCallPart
+				if err := json.Unmarshal(rawPart, &fcPart); err == nil && fcPart.FunctionCall != nil && fcPart.FunctionCall.Name != "" {
```
MrAlias: This still assumes every streamed `functionCall` fragment carries the function name. With Vertex AI's `streamFunctionCallArguments`, only the initial fragment is named; later fragments carry `partialArgs` and `willContinue`, and the terminator can be empty ([documented sequence](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling#streaming_function_call_arguments)). The `Name != ""` guard drops all of those continuation fragments, so the exported call keeps its name but loses the generated arguments.

Can we aggregate these fragments per candidate and call until the call terminates? A fixture matching that documented sequence would distinguish this from the complete one-chunk `args` case fixed earlier.

pkg/ebpf/common/http/gemini_stream.go
```diff
@@ -0,0 +1,245 @@
...
+			if c.FinishReason != "" {
+				agg.finishReason = c.FinishReason
+			}
+			if c.Content == nil {
+				continue
+			}
+
+			for _, rawPart := range c.Content.Parts {
+				var textPart geminiTextPart
+				if err := json.Unmarshal(rawPart, &textPart); err == nil && textPart.Text != "" {
```
MrAlias: This reduces every text part to `Text` before aggregation. Gemini thinking streams mark thought summaries with `thought: true` and may attach `thoughtSignature` to the part, then follow them with normal answer text. The current merge turns that into one `<thought><answer>` part and also drops empty signature-only parts, so the reconstructed response no longer matches the provider's part boundaries ([documented behavior](https://ai.google.dev/gemini-api/docs/generate-content/thinking#thought-summaries)).

Can we retain the text-part metadata and only coalesce adjacent fragments with equivalent metadata? A thought-to-answer regression test would help pin this down.

MrAlias (changes requested):
Thanks for the updates. The earlier comments are addressed. I took another pass through the current aggregation logic against Gemini's current streaming response shapes and left four follow-up comments inline. The repeated string concatenation is the main blocker; the other cases still drop or merge valid response data. I think we should address these before merging.

pkg/ebpf/common/http/gemini_stream.go
NameHaibinZhang: Fixed. Replaced repeated string concatenation with `*strings.Builder` per adjacent text run in `candidatePart`, consistent with `openai_stream.go`. The `appendTextPart` helper now uses `WriteString` for O(1) amortized appends, and `buildGeminiStreamParts` calls `.String()` only at serialization time. Added `TestParseGeminiStream_StringsBuilderPerformance` exercising a 1000-chunk text accumulation.

pkg/ebpf/common/http/gemini_stream.go
NameHaibinZhang: Fixed. Added `tryParseErrorLine` that detects bare `{"error":{...}}` envelopes on both non-`data:` lines and inside `data:` lines (when no candidates/usage are present). The parsed error populates `resp.Error` so `HTTPSpanStatusCode` reports the failed call as error. Added `TestParseGeminiStream_ErrorEnvelopeBare` and `TestParseGeminiStream_ErrorEnvelopeInDataLine` regression tests covering both delivery paths.

pkg/ebpf/common/http/gemini_stream.go
NameHaibinZhang: Fixed. Introduced `fcAggregator` and per-candidate `activeFC` tracking in `candidateAggregator`. `processFunctionCallPart` now handles:
- Complete single-chunk calls (name + args, immediate flush)
- Streaming function call arguments (name first, `partialArgs`/`willContinue` in continuations)
- Name-only calls (no args)

Continuation fragments without a name append to the active function call and flush when `willContinue` is nil or false. Added `TestParseGeminiStream_StreamingFunctionCallArguments` (documented Vertex AI sequence) and `TestParseGeminiStream_StreamingFCNameOnly` regression tests.

pkg/ebpf/common/http/gemini_stream.go
NameHaibinZhang: Fixed. Extended `geminiTextPart` to parse `thought` and `thoughtSignature` fields. `appendTextPart` now only coalesces adjacent text fragments when their `thought` flag matches. `marshalTextPart` emits `{"text":"...","thought":true}` for thought parts and plain `{"text":"..."}` for answer parts, preserving the provider's part boundaries. Added `TestParseGeminiStream_ThoughtToAnswerBoundary` regression test covering the thought→answer transition.

pkg/ebpf/common/http/gemini_stream.go
```diff
@@ -0,0 +1,454 @@
...
+}
+
+type geminiFunctionCallPart struct {
+	FunctionCall *geminiFunctionCallData `json:"functionCall"`
+}
+
+type geminiFunctionCallData struct {
+	Name         string          `json:"name"`
+	Args         json.RawMessage `json:"args,omitempty"`
+	PartialArgs  string          `json:"partialArgs,omitempty"`
```
MrAlias: Thanks for addressing the unnamed continuation chunks. There is still a wire-format mismatch here: Vertex sends `partialArgs` as an array of `PartialArg` objects (`jsonPath` plus a typed value), not as string fragments ([documentation](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling#streaming_function_call_arguments)). With the real shape, unmarshalling into this struct fails, so `processFunctionCallPart` is never reached and the call loses its arguments—or the whole named part when it contains `partialArgs`. The new test uses a string shape the API does not emit.

Can we model the documented array and reconstruct `args` from its paths and typed values, then update the test fixture to match the provider sequence?

pkg/ebpf/common/http/gemini_stream.go
```diff
@@ -0,0 +1,454 @@
...
+
+type geminiStreamContent struct {
+	Parts []json.RawMessage `json:"parts"`
+	Role  string            `json:"role"`
+}
+
+type geminiTextPart struct {
+	Text             string `json:"text"`
+	Thought          bool   `json:"thought,omitempty"`
+	ThoughtSignature string `json:"thoughtSignature,omitempty"`
```
MrAlias: Keeping the `thought` flag fixes the thought-to-answer boundary, but `thoughtSignature` is still discarded. It is decoded here but never carried into `candidatePart` or `marshalTextPart`; signature-only parts are skipped by the `Text != ""` guard, and `buildFunctionCallRaw` also rebuilds function-call parts without the outer signature. Gemini requires signatures to remain on their exact parts and says signed parts must not be merged ([documentation](https://ai.google.dev/gemini-api/docs/generate-content/thought-signatures)).

Can we preserve the signature through both text and function-call aggregation, use it when deciding whether parts may coalesce, and add coverage for distinct and empty-text signatures?

pkg/ebpf/common/http/gemini_stream_test.go
```diff
@@ -0,0 +1,385 @@
...
+
+	require.NotNil(t, resp)
+	require.Len(t, resp.Candidates, 1)
+	var parts []struct {
+		Text string `json:"text"`
+	}
+	require.NoError(t, json.Unmarshal(resp.Candidates[0].Content.Parts, &parts))
+	require.Len(t, parts, 1)
+	assert.Greater(t, len(parts[0].Text), 5000) // 1000 * "chunk " = 6000 chars
+}
```
MrAlias: This test still puts 1,000 repetitions inside one text field in one SSE record, so it never exercises repeated aggregation. This replacement sends 1,000 records through the parser and names the test after the behavior it asserts. If we want to enforce allocation behavior as well, I think that should be a benchmark.

```suggestion
func TestParseGeminiStream_ManyTextChunks(t *testing.T) {
	const chunk = `data: {"candidates":[{"content":{"parts":[{"text":"chunk "}],"role":"model"}}]}` + "\n\n"

	var stream strings.Builder
	for range 1000 {
		stream.WriteString(chunk)
	}

	resp, _ := parseGeminiStream(strings.NewReader(stream.String()))

	require.NotNil(t, resp)
	require.Len(t, resp.Candidates, 1)

	var parts []struct {
		Text string `json:"text"`
	}
	require.NoError(t, json.Unmarshal(resp.Candidates[0].Content.Parts, &parts))
	require.Len(t, parts, 1)
	assert.Equal(t, strings.Repeat("chunk ", 1000), parts[0].Text)
}
```

MrAlias (commented):
Thanks for the updates. The builder and error-envelope fixes look good. I took another pass through the current code against Gemini's documented stream shapes and left three follow-ups inline. The first two still lose valid provider data, so I think they should be addressed before merging.

pkg/ebpf/common/http/gemini_stream.go
```diff
@@ -0,0 +1,454 @@
...
+	UsageMetadata *request.GeminiUsage    `json:"usageMetadata"`
+	ModelVersion  string                  `json:"modelVersion"`
+	ResponseID    string                  `json:"responseId"`
+}
+
+type geminiStreamCandidate struct {
+	Index        int                  `json:"index"`
+	Content      *geminiStreamContent `json:"content"`
+	FinishReason string               `json:"finishReason"`
+}
```
grcevski, line 38: I noticed that the original GeminiCandidate has safetyRatings, does the streaming not have them, or it's just an omission?

pkg/ebpf/common/http/gemini_stream.go
NameHaibinZhang: Fixed. `partialArgs` is now modeled as the documented Vertex AI array of `PartialArg` objects. `PartialArgs` was changed from `string` to `json.RawMessage`, and a new `addPartialArgs` handles both shapes: for the array shape it reads each element's `jsonPath` and typed value and reconstructs the `args` object via `setByJSONPath` (supporting simple dotted `$.a.b` paths); the legacy string-fragment shape is still concatenated for backward compatibility, and unknown shapes are ignored rather than failing the whole part. The test fixture now sends the array sequence (`[{"jsonPath":"$.location","stringValue":"NYC"}]`, then `$.unit`, then `$.days`) and asserts the reconstructed `args`.

One assumption I'd like you to confirm: since the project doesn't depend on the genai package, I couldn't pin down the exact typed-value field name. I take the value from the first non-`jsonPath` field and keep its raw JSON verbatim, so `stringValue`/`numberValue`/`value`/etc. all work. Let me know if you'd prefer I match a specific field set.

pkg/ebpf/common/http/gemini_stream.go
NameHaibinZhang: Fixed. `thoughtSignature` is now preserved through both text and function-call aggregation. `candidatePart` and `fcAggregator` carry a `signature`, `marshalTextPart` and `buildFunctionCallRaw` re-emit it (with `omitempty`), and `appendTextPart` only coalesces adjacent parts when neither the previous nor the current part is signed, so signed parts stay distinct and are never merged. Signature-only parts (empty text with a `thoughtSignature`) are no longer dropped by the `Text != ""` guard. Added coverage for distinct signatures (`TestParseGeminiStream_ThoughtSignaturePreserved`), empty-text signature-only parts (`TestParseGeminiStream_SignatureOnlyPart`), and function-call signatures (`TestParseGeminiStream_FunctionCallSignaturePreserved`).

pkg/ebpf/common/http/gemini_stream_test.go
NameHaibinZhang: Done. Adopted the suggestion: the test is renamed to `TestParseGeminiStream_ManyTextChunks` and now sends 1,000 separate SSE records through the parser (exercising repeated aggregation) and asserts the coalesced result equals `strings.Repeat("chunk ", 1000)`. Agreed that enforcing allocation behavior belongs in a benchmark rather than this test.

pkg/ebpf/common/http/gemini_stream.go
NameHaibinZhang: Good catch — it was an omission. The non-streaming path preserves `safetyRatings` automatically because `request.GeminiResponse` keeps it as a `json.RawMessage`, but the streaming candidate struct didn't have the field, so it was dropped. Added `SafetyRatings json.RawMessage` to `geminiStreamCandidate`, captured it in the aggregation loop, and carried it through `buildGeminiCandidates`. Added `TestParseGeminiStream_SafetyRatingsPreserved` to cover it.

pkg/ebpf/common/http/gemini_stream.go
```diff
@@ -0,0 +1,570 @@
...
+			// value, ...) since we keep the raw JSON value verbatim.
+			var value any
+			for k, v := range elem {
+				if k == "jsonPath" {
+					continue
+				}
+				_ = json.Unmarshal(v, &value)
+				break
+			}
+			setByJSONPath(fc.argsObj, path, value)
```
MrAlias: This still does not implement the documented continuation semantics inside each `PartialArg`. The type has an explicit value union plus a separate `willContinue` flag; taking the first non-`jsonPath` map entry can therefore select `willContinue` instead of the value. It also replaces the value at a repeated path, so Google's recorded `"USA"` fragment followed by the empty terminator becomes `country: ""` in the rebuilt call ([schema](https://github.com/googleapis/js-genai/blob/5fb5354de34e8fe394c92ad279bd07e756328005/api-report/genai-node.api.md#L3810-L3817), [recorded sequence](https://github.com/googleapis/js-genai/blob/5fb5354de34e8fe394c92ad279bd07e756328005/test/system/recordings/Client_Tests_generateContentStream_partial_function_calling_Vertex_AI_should_stream_generate_content_enabled_streamFunctionCallArguments.json#L128-L168)).

Can we decode the explicit `stringValue` / `numberValue` / `boolValue` / `nullValue` fields and accumulate string fragments per path until that argument's `willContinue` is false? The regression fixture should include the same path more than once and set per-argument `willContinue`, so it covers the provider sequence.

pkg/ebpf/common/http/gemini_stream.go
```diff
@@ -0,0 +1,570 @@
...
+// setByJSONPath assigns value at a simple dotted JSON path (e.g. "$.a.b") into
+// the target map, creating intermediate objects as needed. Array indices and
+// complex expressions are not supported and are skipped.
+func setByJSONPath(m map[string]any, path string, value any) {
+	path = strings.TrimPrefix(path, "$")
+	path = strings.TrimPrefix(path, ".")
+	if path == "" {
+		return
+	}
+	keys := strings.Split(path, ".")
```
MrAlias: This still treats the JSONPath as dot-separated object keys. For `$.items[0].id`, the rebuilt argument is `{"items[0]":{"id":7}}` instead of `{"items":[{"id":7}]}`. Array segments are part of the contract—the API schema uses `$.foo.bar[0].data` as its example—so valid tool arguments are changed here ([schema](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/reference/rpc/google.cloud.aiplatform.v1#partialarg)).

I don't think we should maintain a partial JSONPath implementation for this. The repository already uses `github.com/ohler55/ojg/jp`, whose `Expr.Set` creates missing map and array elements. Can we reuse that path parser/setter and add an array-path regression case?

MrAlias (commented):
The signature, many-chunk, and safety-ratings fixes look good. I found two remaining `partialArgs` cases that still corrupt valid function-call arguments. I think these should be addressed before merging.

pkg/ebpf/common/http/gemini_stream.go
NameHaibinZhang: Fixed. `partialArgs` elements are now decoded into an explicit `geminiPartialArg` struct with the typed value union (`stringValue` / `numberValue` / `boolValue` / `nullValue`) and a per-argument `willContinue`, instead of picking the first non-`jsonPath` field (which could land on `willContinue`). String values are now accumulated per `jsonPath` across fragments until `willContinue` is false/absent, so the recorded `"USA"` fragment followed by the empty terminator now yields `country: "USA"` instead of `""`. Added `TestParseGeminiStream_StreamingFunctionCallStringFragments`, which repeats the same path with per-argument `willContinue` and a trailing empty terminator.

pkg/ebpf/common/http/gemini_stream.go
NameHaibinZhang: Fixed. Dropped the hand-rolled dotted-key setter and now reuse the existing `github.com/ohler55/ojg/jp` parser (`jp.ParseString`) to interpret the JSONPath, then walk the parsed fragments to create the intermediate objects and arrays. `$.items[0].id` now reconstructs `{"items":[{"id":7}]}` instead of `{"items[0]":{"id":7}}`. Added `TestParseGeminiStream_StreamingFunctionCallArrayPath` as an array-path regression. One note: I tried `jp.Expr.Set` directly, but on a `map[string]any` root it doesn't vivify a nested object inside a freshly-created array slot (it returns `can not follow a <nil> at '$.items[0]'`), so I kept ojg's parser for the path grammar and do the map/array creation in a small recursive setter. Happy to switch to `Expr.Set` if you know a form that vivifies array-of-object paths.

pkg/ebpf/common/http/gemini_stream.go
```diff
@@ -0,0 +1,644 @@
...
+		key := string(f)
+		m[key] = setAtFrags(m[key], frags[1:], value)
+		return m
+	case jsonpath.Nth:
+		idx := int(f)
+		if idx < 0 {
+			return container
+		}
+		s, _ := container.([]any)
+		for len(s) <= idx {
```
MrAlias, line 241: `setAtFrags` still grows slices directly from `partialArgs[].jsonPath`, which is response-provided data. In a malformed stream, a path like `$.items[1000000000].id` drives this loop until the slice reaches that index, so a single captured response can allocate unbounded memory during Gemini parsing.

Can we bound supported array indices here and skip paths outside the limit? The candidate index path already uses a defensive `maxGeminiStreamCandidates` guard; this setter needs the same kind of cap plus an oversized-array-index regression test.

MrAlias (commented):
Thanks for the updates. I found one remaining allocation guard issue in the JSONPath setter that should be addressed before merging.

grcevski (approved):
LGTM!

pkg/ebpf/common/http/gemini_stream.go
NameHaibinZhang: Good catch — `setAtFrags` indeed needs a defensive cap on array indices to prevent unbounded allocation from malformed paths. Will add a `maxPartialArgArrayIndex` bound (consistent with the existing `maxGeminiStreamCandidates` pattern), skip paths that exceed it, and include an oversized-array-index regression test.

NameHaibinZhang:
@MrAlias Could you please re-run the failed tests for me? It doesn't look like the issue was caused by the code this time, and all comments have been addressed.

MrAlias (approved):
🚀

PR #2542, opened 2026-06-30
Scope context propagation socket capture to monitored processes

mariomac:
## Summary

Fixes grafana/beyla#2691 ("sock_ops/sk_msg programs
conflicting with Cilium L7 LB generating stalled connections").

When context propagation is enabled, `tpinjector` claims TCP sockets into the
`sock_dir` sockhash and runs an `sk_msg` verdict (`obi_packet_extender`) over
them to inject `traceparent`. Both capture paths claimed **every** socket in the
cgroup, regardless of whether it belonged to a process OBI monitors:

- `obi_sockmap_tracker` (sockops) on `BPF_SOCK_OPS_ACTIVE_ESTABLISHED_CB` called
  `bpf_sock_hash_update` unconditionally. The `valid_pid()` gate only runs later,
  in the `sk_msg` program — by then the socket is already in the sockhash.
- `obi_sk_iter_tcp` (startup iter/tcp) added every existing socket too.

On hosts where another component manages the same sockets via sockmap — e.g.
Cilium's socket-LB / L7 proxy — both run an `sk_msg`/`sk_skb` verdict over the
same sockets and OBI additionally rewrites bytes with `bpf_msg_push_data`,
corrupting/stalling those connections. Reporter confirmed:
`context_propagation=disabled` → no stall; `=headers` → stall.

This is the sockmap-layer analogue of the existing TC-layer Cilium handling
(`EnsureCiliumCompatibility`), which had no counterpart here.

### Fix

Only claim sockets owned by a monitored process. The owning PID isn't reliable
in the sockops callback (ACTIVE_ESTABLISHED runs in softirq for non-loopback
sockets), so we consult `sock_pids` — keyed by the connection tuple and
populated in process context, gated by `valid_pid()`, from `kprobe/tcp_connect`
(`store_sock_pid`), which runs before establishment.

- `tpinjector.c`: `bpf_sock_ops_active_est_cb` now gates on
  `sock_belongs_to_monitored_pid()`.
- `sock_iter.c`: `obi_sk_iter_tcp` now gates on `iter_sock_is_monitored()`;
  `iterConstants()` propagates `filter_pids` to the iter collection.
- `filter_pids == 0` (PID filter off) preserves the legacy "track everything"
  behavior.

### Behavior change

Connections open **before** OBI starts aren't in `sock_pids`, so the startup
iterator no longer tracks them. This is intentional: such sockets can't be
attributed to a monitored process, and tracking them is exactly what triggers
the Cilium conflict.

### Tests

New privileged tests (`sockmap_coexistence_privileged_test.go`) load the real
BPF objects, attach them, and drive loopback traffic:

- `TestSockmap_DoesNotClaimBystanderSockets` — fails pre-fix, passes post-fix.
- `TestSockmapIter_DoesNotTrackBystanderSockets` — iterator counterpart.
- `*_CapturesWhenPidFilterOff` / `*_TracksWhenPidFilterOff` — positive controls
  ensuring the fix didn't disable capture entirely.

Run: `make test-privileged`. Verified on kernel 7.0. The monitored-IS-captured
path remains covered by the end-to-end traceparent integration suites.

## Validation

### DIFF
--- a/bpf/tpinjector/sock_iter.c
+++ b/bpf/tpinjector/sock_iter.c
@@ -6,13 +6,44 @@
 #include <bpfcore/bpf_helpers.h>
 #include <bpfcore/bpf_endian.h>
 
+#include <common/connection_info.h>
 #include <common/protocol_defs.h>
+#include <common/sockaddr.h>
 
 #include <logger/bpf_dbg.h>
 
 #include <maps/sock_dir.h>
+#include <maps/sock_pids.h>
 
 char __license[] SEC("license") = "Dual MIT/GPL";
+
+// Mirrors pid.h's filter_pids. Set for this (separate) collection via
+// tpinjector.iterConstants(). When PID filtering is off we keep the legacy
+// behaviour of tracking every existing socket.
+volatile const s32 filter_pids = 0;
+
+// Whether a pre-existing socket belongs to a monitored process. See
+// sock_belongs_to_monitored_pid() in tpinjector.c for the full rationale:
+// claiming unrelated sockets collides with other sockmap users such as Cilium
+// (https://github.com/grafana/beyla/issues/2691). sock_pids is populated in process context by kprobe/tcp_connect,
+// so at startup it only holds connections OBI has already seen connect; truly
+// pre-existing connections are intentionally left untracked rather than risk
+// capturing foreign sockets.
+static __always_inline bool iter_sock_is_monitored(struct sock_common *skc) {
+    if (!filter_pids) {
+        return true;
+    }
+
+    connection_info_t conn = {};
+    // struct sock starts with struct sock_common (__sk_common at offset 0), so
+    // skc can be read as a struct sock * by parse_sock_info.
+    if (!parse_sock_info((struct sock *)skc, &conn)) {
+        return false;
+    }
+    sort_connection_info(&conn);
+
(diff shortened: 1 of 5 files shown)


mariomac:
Privileged tests failed. Submitting the fix and see if they now pass

mariomac:
This PR currently is not going to completely fix the issue, but reducing the blast radius as we are only taking care of sockets from monitored programs. However they might still be handled both by OBI and Cilium.

The complete fix is going to be to actually close the intersection, OBI needs to stop coexisting on the same sockets when a sockmap peer is present — the sockmap-layer analogue of the existing EnsureCiliumCompatibility (TC layer). Concretely:

- Detect a Cilium sockmap/socket-LB peer (enumerate cgroup-attached sockops/sk_msg programs and match Cilium's cil_ name prefix — same technique hasCiliumTCX() uses for TCX), and/or detect socket-LB being active.
- When detected, don't run black-box sockmap CP. Options, roughly in order of safety: disable black-box context propagation; or fall back to a non-sockmap injection path (TC-based, like the network path that already has Cilium handling); or restrict to TCP-options injection (no byte rewriting) if the disambiguation shows mechanism (1) isn't the culprit.
- Surface it as a clear log/warning (mirroring the TC path) so it's observable.

That detection-and-disable is what makes it robust rather than probabilistic.

Issue #2531, opened 2026-06-29
Instrument Aerospike


NimrodAvni78:
Instrument the [Aerospike](https://aerospike.com/) database.
Aerospike clients currently have no otel semantic conventions, or any existing otel instrumentations, but we can follow the general [DB semantic convention](https://opentelemetry.io/docs/specs/semconv/db/)

MrAlias:
Aerospike instrumentation landed in #2545 and is now available on main, including configuration, telemetry export, documentation, and test coverage. This satisfies the scope of this issue for v0.11.0.

#2532 is demo follow-up work rather than a requirement for shipping the instrumentation, so I'm going to keep it open and make it independent of this issue.

The current user-space detection limitations are documented. Any future kernel-side detection work should be tracked separately.

Closing as completed.

PR #2545, opened 2026-06-30
Feat: Aerospike instrumentation

NimrodAvni78:
## Summary

First step in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2531
implement instrumentation for [Aerospike Database](https://aerospike.com/) calls.

Important things to mention
- Aerospike currently does not have official or non-official otel instrumentation. and we relied on the general [DB semantic conventions](https://opentelemetry.io/docs/specs/semconv/db/). with a new non-existing db.system.name: aerospike
- Protocol detection is currently purely in user-space. only limitations i found currently is that some clients (for example java)  receive responses split. first the header and only then the data. in that case we cannot read the status code correctly and assume a failed response is actually a success, we still try to parse the status code for best effort. and kernel side detction and implementation will be implemented in next steps
- `db.query.text` is populated using the key used for the query. the key is not present in the request if the `sendKey` WriteOption is not enabled by the client.
 
All of this is also described in `devdocs/protocols/tcp/aerospike.md`

## Validation


pkg/export/attributes/names/attrs.go
```diff
@@ -222,6 +222,7 @@ const (
 	DBResponseStatusCode = Name(semconv.DBResponseStatusCodeKey)
 	DBNamespace          = Name(semconv.DBNamespaceKey)
 	DBResponseError      = Name("db.response.error")
+	DBOperationBatchSize = Name("db.operation.batch.size")
```
pinoOgni:
```suggestion
	DBOperationBatchSize = Name(semconv.DBOperationBatchSizeKey)
```

mmat11 (approved):
lgtm

pinoOgni (approved):
lgtm! Just on comment.

marctc (commented):
LGTM but i have a question: is there any reason to not use OATS tests here like most of other protocols? I also try sometimes to add for a second language (typically Go and Java/Python), i had some problems with implementation of memcached but perhaps in this case is not necessary.

NimrodAvni78:
> LGTM but i have a question: is there any reason to not use OATS tests here like most of other protocols? I also try sometimes to add for a second language (typically Go and Java/Python), i had some problems with implementation of memcached but perhaps in this case is not necessary.

TBH I'm not sure what is the standard we are using for testing protocols e2e
For most feature I implemented I ended up writing integration tests and not oats tests. Also currently the weaver schema validation only runs as part of integration tests (I'm working on wiring it to oats and k8s tests now)
We should probably agree on a best practice on what we should use to test, hand crafted integration tests? Oats tests? Or some other test framework entirely

marctc:
> TBH I'm not sure what is the standard we are using for testing protocols e2e For most feature I implemented I ended up writing integration tests and not oats tests. 

Sadly there's no standard for this, as we have a mix of both, but the majority of protocols use OATS as it as simple as write yaml with the queries for tempo and prometheus.

> Also currently the weaver schema validation only runs as part of integration tests (I'm working on wiring it to oats and k8s tests now) We should probably agree on a best practice on what we should use to test, hand crafted integration tests? Oats tests? Or some other test framework entirely

Agree. Perhaps to discuss on the sig?

marctc (approved):

NimrodAvni78:
> Agree. Perhaps to discuss on the sig?

Yes! will add it to the agenda

grcevski (approved):
LGTM!

PR #2554, opened 2026-07-01
Fix empty `telemetry.distro.version` in published Docker images

pinoOgni:
## Summary

The current obi docker image has an empty `telemetry.distro.version` resource attribute and blank `Version=` in startup logs.

This PR fixes it and passes `RELEASE_VERSION` and `RELEASE_REVISION` as Docker build ARGs from CI, where the correct tag or commit SHA is already known, rather than computing them with git inside the container.

Release tag builds (e.g. v0.10.1) embed the full semver tag; snapshot builds (push to main) embed the 7-char commit SHA rather than the literal string "main"; local make image-build derives the value from git on the host as before.

Notes: 
- binary release tarballs produced by `make release` were never affected because that job checks out with full history on the host where git describe works correctly.
- In the Makefile we now have  `cut -d/ -f2-` so branch names containing slashes (e.g. pgn/...) are not silently truncated when building locally.

## Validation

### DIFF
--- a/.github/workflows/publish_dockerhub_k8s_cache_main.yml
+++ b/.github/workflows/publish_dockerhub_k8s_cache_main.yml
@@ -48,6 +48,10 @@ jobs:
           ref: ${{ inputs.ref || github.sha }}
           persist-credentials: false
 
+      - name: Compute short SHA
+        id: short_sha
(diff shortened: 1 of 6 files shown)


MrAlias (approved):
Awesome 👍

NimrodAvni78 (approved):
🔥

PR #2559, opened 2026-07-01
Test Weaver namespace advisory enforcement

MrAlias:
## Summary

I added focused regression coverage for Weaver namespace-extension advisories:

- extract actionable-advisory counting into a helper used by `validateWeaverReport`
- construct a synthetic Weaver report containing an information-level `extends_namespace` finding
- assert that the report produces one actionable advisory

## Motivation

#2526 corrected the serialized finding field from `advice_type` to `id`, but the live integration checks no longer emit an `extends_namespace` advisory after the missing attributes were registered. That meant the enforcement path could regress while CI remained green.

The synthetic report locks down both Weaver's serialized `id` field and OBI's decision to treat `extends_namespace` findings as actionable despite their information level.

## Validation

- `go test ./internal/test/integration -run TestExtendsNamespaceAdviceIsActionable -count=1`
- `make lint`

### DIFF
--- a/internal/test/integration/weaver.go
+++ b/internal/test/integration/weaver.go
@@ -251,11 +251,7 @@ func validateWeaverReport(t *testing.T, report *weaverReport) {
 	// Build message → {level, signals} lookup from the sample data.
 	adviceByMsg := collectAdviceInfo(report.Samples)
 
-	// Log all advisory messages grouped by level, and count actionable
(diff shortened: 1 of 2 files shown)


NimrodAvni78 (approved):

mariomac (approved):
LGTM!

mmat11 (approved):

PR #2575, opened 2026-07-02
Fix buffered channel receive handoff matching

MrAlias:
## Motivation

- The receive path could misclassify buffered receives as direct handoffs by inspecting `hchan.qcount` before the Go runtime acquired the channel lock. This could leave stale sender entries in the buffered sender LRU map and produce incorrect span links.
- Buffered sender entries should be consumed when a receive takes from the buffer, and receives should be classified as direct handoffs only when it is safe to do so.

## Description

- Record the channel receive index when a receive starts and compare it with the index after the receive completes.
- Consume the matching buffered sender entry when the receive index advances, then consider direct-handoff matching only when no buffered slot was consumed.

## Testing

- `go test ./pkg/internal/ebpf/gotracer`

### DIFF
--- a/bpf/gotracer/go_runtime.c
+++ b/bpf/gotracer/go_runtime.c
@@ -565,7 +565,8 @@ static __always_inline int channel_recv_return(struct pt_regs *ctx) {
     bool consumed_buffered_sender = false;
     const bool recvx_read = read_channel_recvx((void *)invocation->chan_ptr, &recvx);
     const bool recvx_changed = recvx_read && recvx != invocation->recvx;
-    if (recvx_changed) {
+    const bool single_slot_buffered_receive = dataqsiz == 1 && !invocation->direct_handoff;
+    if (recvx_changed || single_slot_buffered_receive) {
         const u64 slot = previous_channel_slot(recvx, dataqsiz);
         if (invocation->has_handoff) {
             consumed_buffered_sender =
--- a/bpf/gotracer/go_runtime.c
+++ b/bpf/gotracer/go_runtime.c
@@ -509,7 +509,9 @@ static __always_inline int channel_recv_start(struct pt_regs *ctx) {
     }
 
     if (dataqsiz != 0) {
-        read_channel_recvx((void *)chan_ptr, &invocation.recvx);
+        if (!read_channel_recvx((void *)chan_ptr, &invocation.recvx)) {
+            return 0;
+        }
 
         u64 qcount = 0;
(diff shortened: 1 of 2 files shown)


bpf/gotracer/go_runtime.c
```diff
@@ -553,18 +556,36 @@ static __always_inline int channel_recv_return(struct pt_regs *ctx) {
...
-            consume_buffered_channel_sender(&chan_key, slot, &invocation->handoff);
+            consumed_buffered_sender =
+                consume_buffered_channel_sender(&chan_key, slot, &invocation->handoff);
         } else {
-            consume_buffered_channel_sender(&chan_key, slot, NULL);
+            consumed_buffered_sender = consume_buffered_channel_sender(&chan_key, slot, NULL);
+        }
+    }
+
+    if (recvx_read && !consumed_buffered_sender && !recvx_changed && dataqsiz != 1) {
```
grcevski, line 581: I wonder if this regresses the `make(chan T, 1)` case? I think for `make(chan T, 1)` the recvx will always be 0 before and after a buffered receive, which I think a buffered sender at slot 0 is never consumed.

I wonder if we can track whether the receive started with some buffered data, not just track if recvx has changed? For capacity - 1 channels, maybe something like !invocation.direct_handoff && qcount > 0, or maybe !invocation.direct_handoff && dataqsiz == 1?

MrAlias: Fixed in aceb634a96. Capacity-one buffered receives now use the entry-time `!direct_handoff` classification even though `recvx` remains zero, so the slot-zero buffered sender is consumed.

bpf/gotracer/go_runtime.c
```diff
@@ -507,11 +508,13 @@ static __always_inline int channel_recv_start(struct pt_regs *ctx) {
         invocation.has_handoff = true;
     }
 
-    if (dataqsiz == 0) {
-        invocation.direct_handoff = true;
-    } else {
+    if (dataqsiz != 0) {
+        read_channel_recvx((void *)chan_ptr, &invocation.recvx);
```
skl: nit: should the return value of `read_channel_recvx` be checked here?

skl (approved):

grcevski (approved):
LGTM!

bpf/gotracer/go_runtime.c
MrAlias: Fixed in ded51cdae2. The receive entry probe now returns without recording the invocation when `read_channel_recvx` fails, avoiding correlation with a zero-initialized index.

PR #2580, opened 2026-07-03
fix: workaround ioctl(FIONREAD) wrongly returning zero when context propagation is enabled

mmat11:
## Summary

On certain kernels, ioctl(FIONREAD) returns zero for sockets with sockmaps when no verdict program is attached. This PR works around it by replacing the return value with the existing byte count so applications relying on ioctl(FIONREAD) won't hang anymore.

Fixes: https://github.com/grafana/beyla/issues/2941 , https://github.com/grafana/alloy/issues/6085

## Validation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -237,7 +237,7 @@ update-offsets:
 BPF_ROOT = pkg/
 
 # Find all generated Go and object files (used as Make targets)
-BPF_GEN_GO := $(shell find $(BPF_ROOT) -type f \( -name 'bpf_*_bpfe[lb].go' -o -name 'net_*_bpfe[lb].go' -o -name 'netsk_*_bpfe[lb].go' -o -name 'stats_*_bpfe[lb].go' \))
+BPF_GEN_GO := $(shell find $(BPF_ROOT) -type f \( -name 'bpf*_bpfe[lb].go' -o -name 'net_*_bpfe[lb].go' -o -name 'netsk_*_bpfe[lb].go' -o -name 'stats_*_bpfe[lb].go' \))
 BPF_GEN_OBJ := $(BPF_GEN_GO:.go=.o)
 BPF_GEN_ALL := $(if $(BPF_GEN_GO),$(BPF_GEN_GO) $(BPF_GEN_OBJ))
 
--- a/pkg/internal/ebpf/tpinjector/fionread_check.go
+++ b/pkg/internal/ebpf/tpinjector/fionread_check.go
@@ -1,0 +1,160 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build linux
+
+package tpinjector // import "go.opentelemetry.io/obi/pkg/internal/ebpf/tpinjector"
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"time"
+
+	"github.com/cilium/ebpf"
+	"golang.org/x/sys/unix"
+
+	ebpfconvenience "go.opentelemetry.io/obi/pkg/internal/ebpf/convenience"
+)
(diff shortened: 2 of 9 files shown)


pkg/internal/ebpf/tpinjector/fionread_check.go
```diff
@@ -0,0 +1,160 @@
...
+// test-load on a throwaway copy so LoadSpecs can skip the bundle on kernels
+// that reject bpf_probe_write_user (lockdown)
+func loadableFIONREADFixup() (*ebpf.CollectionSpec, error) {
+	spec, err := LoadBpfFionreadFixup()
+	if err != nil {
+		return nil, err
+	}
+	test := spec.Copy()
+	for _, m := range test.Maps {
+		if m.Pinning == ebpfconvenience.PinInternal {
```
grcevski, line 133: I'm assuming we had to clear the pin internal here for the throw away map so we don't get these throwaways shared?

Makefile
```diff
@@ -237,7 +237,7 @@ update-offsets:
 BPF_ROOT = pkg/
 
 # Find all generated Go and object files (used as Make targets)
-BPF_GEN_GO := $(shell find $(BPF_ROOT) -type f \( -name 'bpf_*_bpfe[lb].go' -o -name 'net_*_bpfe[lb].go' -o -name 'netsk_*_bpfe[lb].go' -o -name 'stats_*_bpfe[lb].go' \))
+BPF_GEN_GO := $(shell find $(BPF_ROOT) -type f \( -name 'bpf*_bpfe[lb].go' -o -name 'net_*_bpfe[lb].go' -o -name 'netsk_*_bpfe[lb].go' -o -name 'stats_*_bpfe[lb].go' \))
```
grcevski, line 240: what prompted this change? does the newly generated bpf file not follow the bpf_ pattern?

grcevski (commented):
This is awesome! 

I did think of something else. If I understand correctly, the problem is that we have sockets in the hashmap, but there's no verdict program for the ingress path. What if we attached a dummy ingress verdict?

grcevski (approved):
LGTM! It seems that attaching to ingress may expose us to other bugs already fixed in prior kernel versions.

pkg/internal/ebpf/tpinjector/fionread_check.go
mmat11: yes!

Makefile
mmat11: yeah it was being generated without the underscore, I didn't follow up on "why"

skl (approved):

neilkuan:
Thanks for this fix @grcevski @mmat11 🙏 — heads-up from field testing on **Beyla v3.27.0** (which ships this PR):

On **kernel 6.12.79 (aarch64)** the workaround **detects** the bug but the compensation **fails to attach**, so nginx large-response transfers still stall/truncate:

```
WARN  tpinjector msg="kernel misreports ioctl(FIONREAD) ... enabling BPF compensation for tracked sockets"
INFO  msg="Attaching sock msgs"
INFO  msg="Attaching sock ops"
ERROR tpinjector msg="FIONREAD compensation is ineffective (attach failed or blocked?); ... set context_propagation: disabled to avoid impact"
```

Measured: 6.12.79 aarch64 → 87.4s + `curl` exit 18 (truncated at 1.5/3.28 MB); a 6.12.73 node with the same image → 5.9s full transfer. Only the 6.12.75+ node trips the detection.

Full write-up + cross-node comparison in grafana/beyla#2941 (reopened). Would appreciate a pointer on why the `sock_msg`/`sock_ops` attach ends up ineffective on this kernel — glad to test a debug build.

PR #2583, opened 2026-07-05
OATs weaver validation tests

### LINKED ISSUE #2570: Weaver validation should cover OATS tests

NimrodAvni78:
## Summary

Add weaver validation tests to OATs tests.
Resolves https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2570

## Review Guidelines
This PR spans many files, most of them are duplicates, this is a guideline of where the important changes are and where its duplicate code.

Product code actual changes emitted telemetry where we were not following the semantic conventions:
1. pkg/export/otel/metrics.go — 1 line: token-usage histogram unit "1" → "{token}" (semconv-mandated).
2. pkg/appolly/app/request/metric_attributes.go — messaging.destination.partition.id emitted as string instead of int (semconv type). Anything consuming the int attribute sees a type change.

The validation core:

3. internal/test/weavercheck/weavercheck.go — Validate enforcement semantics, isActionableAdvice/countActionableAdvisories, FetchReport with the stale-report mtime guard, moving the weaver validation to a shared package instead of under the integration tests package. Also 2 new ignores and the ignore lists (dns.question.name and gen_ai.tool.name string[] in open questions)
4. internal/test/weavercheck/weavercheck_test.go — ported #2559 regression test.
5. internal/test/integration/weaver.go — slimmed to a Docker-transport driver; check nothing was lost vs the old inline logic.
6. internal/test/integration/weaver_test.go — extends_namespace test removed (moved to weavercheck), manifest-version test kept.

OATS integration:

7. internal/test/oats/harness/weaver.go — the ginkgo hook; the subtle bit is the ECONNREFUSED-means-not-wired skip.
8. internal/test/oats/harness/harness.go — teardown defer moved before Start() (leak fix; small but load-bearing).
9. internal/test/oats/weaver/docker-compose-weaver.yml + otelcol-weaver.yaml — the shared fragment; the NOTE about --project-directory path resolution is the thing to sanity-check.
10. internal/test/oats/amqp/docker-compose-obi-{python,java}-amqp.yml + configs/otelcol-weaver-host.yaml — the one genuinely different wiring (host networking: collector on 4319, weaver OTLP on 4327, self-metrics disabled).

Registry + build:

11. schemas/obi/groups/spanmetrics.yaml — legacy-name metric declarations; the unit: "" rationale comment is the reviewable claim.
12. schemas/obi/groups/traces.yaml — gen_ai.metadata declaration.
13. Makefile — oats-prereq now depends on fetch-upstream-semconv (what makes CI work).

Repeating mechanical changes — review one instance, skim the rest

- 45 testcase yamls (*/yaml/oats_*.yaml): identical single line appended — - ../../weaver/docker-compose-weaver.yml. 
- ~40 compose files: two one-line env edits each — OTEL_EXPORTER_OTLP_ENDPOINT: http://lgtm:4318 → http://collector:4318, and OTEL_EBPF_METRICS_INTERVAL: "10ms" → "1s" (matching #2008). Review one bridge compose; amqp's two are the exception already in the deep list.
- 20 go.mod/go.sum files (harness + 9 groups): identical require go.opentelemetry.io/obi v0.0.0 + replace => ../../../.. pair per module; go.sum churn is tidy fallout (testify-weight, no eBPF deps). Review harness/go.mod (has the explanatory comment) and one group's (e.g. nats/go.mod).
- ai/configs/instrumenter-config-traces-genai.yml — trivial residue (whitespace) after the dns opt-in experiment was reverted.

If you still think this PR is too large to effectively validate let me know and ill try to split it

### Open Questions
As part of this effort i discovered 2 schema violations, currently they are ignored until we decide to settle
1. ~`Required attribute 'dns.question.name' is not present.` - according to the semantic conventions the dns.question.name attribute is required on dns metrics, but we intentionally state it as optional to avoid cardinality. that is a semconv violations and we need to decide what we do (keep it this way and ignore or conform to the semantic convention)~
resolved in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2583/changes/9b2b47154b459b1f33fdac2a9f1c243fbfa8bf93
2. ~`Attribute 'gen_ai.tool.name' has type 'string[]'. Type should be 'string'.` - on GenAI spans, we append all tool calls on any span instead of splitting them to separate spans. this does not follow to otel semconv, and we need to decide what to do there, the options are:~
     ~-  declare a new OBI attribute `gen_ai.tools.name` with a list of tools`
     ~-  concatenate all the tools into 1 string with a separator `,` or something.~
     ~-  see if we can split to 1 span per tool call~
     ~-  keep ignoring this violation~
resolved in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2588
     

## Validation


skl (approved):

pkg/appolly/app/request/metric_attributes.go
```diff
@@ -121,7 +122,7 @@ func DBResponseStatusCode(val string) attribute.KeyValue {
 }
 
 func MessagingPartition(val int) attribute.KeyValue {
-	return attribute.Key(attr.MessagingPartition).Int(val)
+	return attribute.Key(attr.MessagingPartition).String(strconv.Itoa(val))
 }
```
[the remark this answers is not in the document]
NimrodAvni78: this and all copilot comments are misleading. `messaging.destination.partition.id` should be a string and not an int
resolving them

marctc (approved):
Nice stuff! If someone adds a new protocol and forgets to wire weaver, would it fail?

NimrodAvni78:
follow up issue on weaver for overriding attributes / groups:
https://github.com/open-telemetry/weaver/issues/1578

NimrodAvni78:
> Nice stuff! If someone adds a new protocol and forgets to wire weaver, would it fail?

https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2583/changes/bbeaf08f5b578be3fc0d0815a7f7266a8ad37a1c makes it so, with optional env var to allow skipping

scripts/lint-schema.sh
```diff
@@ -25,22 +43,39 @@ REGISTRY_PATH="$3"
...
   exit 1
 fi
 
-if [ -n "$out" ] && [ "$out" != "[]" ]; then
+remaining=$(printf '%s' "${out:-[]}" | python3 -c '
+import json
+import sys
+
+def allowlisted(d):
+    return "DuplicateMetricName" in d.get("error", {})
```
MrAlias: This allowlists every `DuplicateMetricName`, even though the only intentional duplicate is `dns.lookup.duration`. If a later schema change accidentally reuses another metric name, this filter removes that diagnostic too, so an otherwise clean `make lint-schema` run succeeds and live-check can silently select the wrong declaration.

The multiline Python is also an awkward fit in this Bash helper; `jq` is already the repository's established JSON tool. Can we parse the diagnostics once with `jq`, remove only the exact DNS duplicate (including its expected two provenances), and add a small fixture proving an unrelated duplicate still fails?

internal/test/oats/harness/go.mod
```diff
@@ -1,49 +1,60 @@
 module go.opentelemetry.io/obi/internal/test/oats/harness
 
-go 1.25.7
+go 1.25.11
 
 require (
 	github.com/grafana/oats v0.6.1
 	github.com/onsi/ginkgo/v2 v2.28.1
 	github.com/onsi/gomega v1.40.0
+	go.opentelemetry.io/obi v0.0.0
```
MrAlias, line 9: This makes the OATS harness depend on the entire root module just to import `weavercheck`. That dependency participates in MVS: this diff already moves the collector pdata, OTel, and gRPC versions used by OATS, and because `replace` directives are not transitive every leaf module now repeats the root `require`/`replace`.

As a result, an unrelated root dependency update can change or break the OATS harness, and each new OATS group has another local replacement to remember. Can we keep `weavercheck` behind a lightweight boundary—for example, place it under the existing harness module and have the root integration tests consume that package—so OATS does not depend back on the full OBI module?

schemas/obi/groups/spanmetrics.yaml
```diff
@@ -97,3 +97,37 @@ groups:
       - ref: span.kind
       - ref: status.code
       - ref: source
+
+  # Legacy Grafana-style names for the same two metrics, matching the older
+  # spanmetrics processor output. This is what OBI emits by DEFAULT — the
+  # `traces_span_metrics_*` names above are the opt-in OTel naming
+  # (`OTEL_EBPF_OTEL_METRICS_FEATURES=application_span_otel`).
```
MrAlias, line 104: These declarations make the registry description stale: `schemas/obi/manifest.yaml` still says the legacy `application_span` underscored variants are not declared. Since that manifest describes the registry's complete emitted contract, can we update it to say both the canonical and legacy names are covered?

MrAlias (commented):
I found two issues that should be addressed before merging and one documentation mismatch.

scripts/lint-schema.sh
NimrodAvni78: fixed in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2583/changes/6dba60f0bc93f739877994998335ae6215b161ce
switched to jq, created `scripts/lint-schema-filter.jq` with rules for allowed schema linting violations and tested with `scripts/lint_schema_filter_test.go`

schemas/obi/groups/spanmetrics.yaml
NimrodAvni78: Done

internal/test/oats/harness/go.mod
NimrodAvni78: The catch with moving weavercheck under the harness module is that the integration tests live in the root module, so root go.mod would need a require/replace on an unpublished nested module — which breaks downstream consumers of go.opentelemetry.io/obi (Beyla/Alloy) unless we tag that module on every release. The options I see: (a) keep the current direction and accept the bounded MVS coupling (weavercheck is stdlib+testify only, verified no heavy deps move), (b) standalone internal/test/weavercheck module + root requires it + add it to release tagging, or (c) move the integration tests out of the root module entirely — structurally right but a much bigger refactor.
WDYT?

MrAlias: Fair point—the unpublished nested-module dependency makes the suggested inversion unsuitable, option (b) is a bad idea as it will couple indirectly internal modules with other releases, and option (c) is disproportionate for this PR. I’m okay with (a). The coupling is still broader than the comment suggests, though: MVS upgrades pdata/OTel/gRPC and requires the leaf replacements even if `weavercheck` itself only imports stdlib + testify. Could we update that comment to document the intentional MVS coupling and capture the larger module-layout cleanup as follow-up work, perhaps as an issue?

NimrodAvni78: comment updated, will open a follow up issue soon

MrAlias (approved):

mmat11 (approved):

PR #2588, opened 2026-07-06
fix: create individual execute_tool spans per tool call for semconv compliance

NameHaibinZhang:
## Summary

Align `gen_ai.tool.name` attribute with OTel GenAI semantic conventions.

## Problem

The OTel semconv defines `gen_ai.tool.name` as type `string` (used on individual `execute_tool` spans), but OBI was aggregating all tool call names into a `string[]` attribute on the parent inference span. This was flagged as a schema validation violation in PR #2583.

## Solution

- Remove the aggregated `string[]` approach (`genAIToolCallAttributes`)
- Create individual `execute_tool` INTERNAL child spans per tool call
- Each child span carries:
  - `gen_ai.operation.name`: `execute_tool`
  - `gen_ai.tool.name`: tool name (single string)
  - `gen_ai.tool.call.id`: tool call ID (when available)
- Child spans inherit the parent's trace ID and use the parent span as their parent
- Timing: child spans share the parent span's start/end timestamps (eBPF cannot determine per-tool-call timing)

## Changes

- `pkg/export/otel/tracesgen/tracesgen.go`:
  - Remove `genAIToolCallAttributes()` function and its 4 call sites (OpenAI, Anthropic, Gemini, Qwen)
  - Add `createToolCallSpans()` to generate individual INTERNAL child spans
  - Add `getSpanToolCalls()` helper to extract tool calls from any vendor
  - Call `createToolCallSpans` in `generateTracesWithAttributes` loop
- `pkg/export/otel/tracesgen/tracesgen_test.go`:
  - Remove `TestGenAIToolCallAttributes`
  - Add `TestCreateToolCallSpans` covering nil/empty/single/multiple/skip-empty-names/no-ID cases

## Testing

- All unit tests pass locally
- CI on fork repo passed (Lint, Unit tests, etc.)

## Related

- Resolves schema validation violation found in #2583
- Reference: [OTel GenAI semconv registry](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/model/gen-ai/registry.yaml) defines `gen_ai.tool.name` as `type: string`

### DIFF
--- a/pkg/export/otel/tracesgen/tracesgen.go
+++ b/pkg/export/otel/tracesgen/tracesgen.go
@@ -775,7 +775,7 @@ func traceAttributesSelectorInternal(span *request.Span, optionalAttrs map[attr.
 				}
 			}
 		}
-		
+
 		if span.SubType == request.HTTPSubtypeAnthropic && span.GenAI != nil && span.GenAI.Anthropic != nil {
 			ai := span.GenAI.Anthropic
 			attrs = append(attrs, semconv.GenAIProviderNameAnthropic)
@@ -845,7 +845,7 @@ func traceAttributesSelectorInternal(span *request.Span, optionalAttrs map[attr.
 				attrs = append(attrs, semconv.ErrorTypeKey.String(ai.Output.Error.Type))
 			}
 		}
-		
+
 		if span.SubType == request.HTTPSubtypeGemini && span.GenAI != nil && span.GenAI.Gemini != nil {
 			ai := span.GenAI.Gemini
 			attrs = append(attrs, semconv.GenAIProviderNameGCPGemini)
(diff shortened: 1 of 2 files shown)


NimrodAvni78 (commented):
overall LGTM
but i would expect some integration tests to break because of this?
or we don't have any integration tests asserting tool calls?

NameHaibinZhang:
@NimrodAvni78  Good question! We don't currently have integration tests that assert tool call attributes on LLM inference spans.

The existing test coverage for `gen_ai.tool.name`:
- **MCP integration test** (`red_test_python_mcp.go`): This already expects `gen_ai.tool.name` as a single `string` on individual `execute_tool` spans — it's the MCP path which was already correct and unaffected by this change.
- **OATS tests** (`internal/test/oats/ai/yaml/`): These test OpenAI/Anthropic/Gemini/Qwen/Bedrock but assert on other attributes (model, operation name, tokens, etc.) — none of them assert on `gen_ai.tool.name` for tool calls in the response.

We could add an integration test that sends a request triggering a tool call response and verifies the child `execute_tool` spans are created correctly. Happy to add that as a follow-up if you think it's needed.

NimrodAvni78 (approved):
LGTM

PR #2594, opened 2026-07-06
Limit sock_iter to only add active socket connections

grcevski:
## Summary

While investigating the issue fixed in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2580, I noticed that part of the problem was that the socket iterator was adding all existing sockets to the sock hash map, including passive (server) sockets we currently don't handle.

This PR filters out the not established (listen) and server sockets before we add them to the sock hash map. The tpinjector.c code already filters on active only. The fix is needed on the socket iterator only.

## Validation

### DIFF
--- a/bpf/tpinjector/sock_iter.c
+++ b/bpf/tpinjector/sock_iter.c
@@ -7,10 +7,13 @@
 #include <bpfcore/bpf_endian.h>
 
 #include <common/protocol_defs.h>
+#include <common/sock_port_ns.h>
 
 #include <logger/bpf_dbg.h>
 
 #include <maps/sock_dir.h>
+
+#include <tpinjector/maps/iter_listening_ports.h>
 
 char __license[] SEC("license") = "Dual MIT/GPL";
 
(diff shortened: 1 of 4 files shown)


pkg/internal/ebpf/tpinjector/tpinjector.go
```diff
@@ -147,6 +147,11 @@ func (p *Tracer) SockMsgs() []ebpfcommon.SockMsg {
 			MapFD:    p.bpfObjects.SockDir.FD(),
 			AttachAs: ebpf.AttachSkMsgVerdict,
 		},
+		// {
+		// 	Program:  p.bpfObjects.ObiSkSkbVerdict,
+		// 	MapFD:    p.bpfObjects.SockDir.FD(),
+		// 	AttachAs: ebpf.AttachSkSKBVerdict,
```
skl: Does this need a kernel >= 5.13 check?

- https://docs.ebpf.io/linux/program-type/BPF_PROG_TYPE_SK_SKB/#as-bpf_sk_skb_verdict-program

grcevski: I've removed that code now. Bugs on older kernel versions cause problems when the ingress verdict is attached.

skl (approved):
lgtm

bpf/tpinjector/sock_iter.c
mmat11: out of context for this PR, I wonder why we double print here?

mmat11 (approved):
lgtm!

bpf/tpinjector/sock_iter.c
grcevski: I forgot to remove it :)

bpf/tpinjector/sock_iter.c
```diff
@@ -90,6 +118,19 @@ int obi_sk_iter_tcp(struct bpf_iter__tcp *ctx) {
...
+    // don't look at sockets that aren't established
+    u8 skc_state = BPF_CORE_READ(skc, skc_state);
+    if (skc_state != TCP_ESTABLISHED) {
+        return 0;
+    }
+
+    // skip passive sockets, their local port is a listening port we
+    // found in the prior pass
+    struct sock_port_ns pn = sock_port_ns_from_skc(skc);
+    if (bpf_map_lookup_elem(&iter_listening_ports, &pn) != 0) {
```
MrAlias, line 130: This still infers a passive accepted socket from the presence of a current listener on the same local port. If OBI starts while a server is draining established connections after the listener has already been closed, those sockets are still server-side but this lookup misses them, so the iterator adds them to `sock_dir` and the FIONREAD/sockhash issue can come back for the server sockets this PR is trying to keep out.

Can we key this off the socket itself instead of the listener snapshot? The BTF for `sock_common` exposes `skc_listener`, and `bpf_get_listener_sock` may also be usable here; if neither works in `iter/tcp`, I think we should cover and document the listener-closed case explicitly.

grcevski: We no longer rely on this code for the FIONREAD, having the sockets in the map after Mattia's fix is harmless. This is just an improvement to avoid tracking sockets that we don't need to track. Missing some that are draining is fine IMO.

MrAlias: Yep, that makes sense. I missed that. 👍

MrAlias (approved):

Issue #2453, opened 2026-06-22
Proposal: Enhanced Network Quality Metrics for OBI


chengshuyi:
### Component

OBI (eBPF instrumentation)

### Problem

OBI's `statsolly` pipeline currently provides four TCP metrics: smoothed RTT (at connection close), retransmit event count, failed connection count, and TCP I/O bytes. While these cover basic visibility, they fall short of the network quality observability required for production troubleshooting:

- **RTT is sampled once at `tcp_close`** — long-lived connections get a single data point; no visibility into RTT variance during the connection lifetime
- **Retransmits lack direction and cause** — the `tcp_retransmit_skb` raw tracepoint fires only for local retransmissions with no differentiation between timeout-triggered, fast-retransmit, or spurious retransmissions
- **No zero-window or OOO tracking** — two of the most common TCP performance issues are completely invisible
- **No packet loss rate** — `statsolly` and `netolly` are isolated subsystems; retransmit counts cannot be correlated with total packet counts
- **No connection lifecycle metrics** — no concurrent connection gauge, no handshake duration, no SYN retransmission tracking

Production environments routinely require these signals for root-cause analysis. In Kubernetes deployments, distinguishing pod-to-pod network degradation (retransmissions, zero-window stalls) from application-level slowness is a top troubleshooting pain point — without per-connection network quality metrics, operators resort to `ss -ti` or `tcpdump` on individual nodes, which doesn't scale. The Linux kernel already tracks most of these counters internally in `struct tcp_sock`; OBI simply needs to expose them via the kprobe/tracepoint approach that already fits its `statsolly` architecture.

### Architecture

All new metrics extend the existing `statsolly` pipeline. No new pipelines or subsystems are introduced.

```
                    Kernel Space                          User Space
               ┌─────────────────────┐            ┌──────────────────────┐
               │  NEW kprobes/TPs    │            │  statsolly Go code   │
               │                     │            │                      │
  tcp_ack ─────┤  periodic SRTT      │            │  Stat struct         │
               │  sample + jitter    ├──ringbuf──►│  (extended types)    │
               │                     │            │                      │
  tcp_send_    │  zero window        │            │  OTel metric         │
  probe0 ──────┤  event              │            │  definitions         │
               │                     │            │  (stats.yaml)        │
  tcp_retrans  │  retransmit with    │            │                      │
  _synack ─────┤  direction + type   │            │  Prometheus +        │
[...]

PR #2598, opened 2026-07-07
feat(statsolly): add TCP connection summary metrics on tcp_close

chengshuyi:
## Summary

Implements the **TCPConnectionSummary** (StatType=7) event from the enhanced network quality metrics proposal (#2453).

At `tcp_close`, a new kprobe reads lifetime TCP statistics from `struct tcp_sock` and emits them as a single ring buffer event. The event carries 6 high-value fields that are exported as OpenTelemetry/Prometheus histograms for aggregation (P50/P95/P99 distributions).

### New Metrics

| Metric | Type | Unit | Source |
|--------|------|------|--------|
| `obi_stat_tcp_connection_summary.srtt` | Float64Histogram | seconds | `tp->srtt_us >> 3` |
| `obi_stat_tcp_connection_summary.mdev` | Float64Histogram | seconds | `tp->mdev_us >> 2` |
| `obi_stat_tcp_connection_summary.retransmits` | Int64Histogram | count | `tp->total_retrans` |
| `obi_stat_tcp_connection_summary.ooo_packets` | Int64Histogram | count | `tp->rcv_ooopack` |
| `obi_stat_tcp_connection_summary.segs_out` | Int64Histogram | count | `tp->segs_out` |
| `obi_stat_tcp_connection_summary.segs_in` | Int64Histogram | count | `tp->segs_in` |

### Design Decisions

- **srtt scaling**: `>> 3` to undo kernel `<< 3` (consistent with existing `obi_stats_kprobe_tcp_close_srtt`)
- **mdev scaling**: `>> 2` to undo kernel `<< 2` (kernel stores mdev with `<< 2` factor)
- **srtt filtering**: `srtt_us == 0` (no RTT sample) and `> k_max_srtt_allowed` are dropped, consistent with the existing srtt probe
- **Role attribute**: `client`/`server` label via existing `sock_role` map lookup
- **Feature flag**: `stats_tcp_connection_summary` gated under `FeatureStats` composite, controlled via `OTEL_EBPF_METRICS_FEATURES`
- **4 low-value fields removed** from initial design (`snd_cwnd`, `snd_wnd`, `rcv_wnd`, `rtt_min_us`) — these are instantaneous snapshots at connection close with no meaningful aggregation semantics

### Files Changed (17 files, +670/-22)

**BPF (eBPF C):**
- `bpf/statsolly/k_tcp.c` — new `tcp_connection_summary_t` struct + `obi_stats_kprobe_tcp_close_summary` kprobe
- `bpf/statsolly/types.h` — `k_event_stat_tcp_connection_summary = 7` enum

**Go user-space:**
- `pkg/internal/statsolly/ebpf/stat.go` — `TCPConnectionSummary` + `StatsTCPConnectionSummary` (BPF mirror) structs, `StatType` constant
- `pkg/internal/statsolly/ebpf/stat_getters.go` — role getter for connection summary
- `pkg/internal/statsolly/ebpf/stats_tracer.go` — kprobe registration, bpf2go type, feature gating, fixup spec
- `pkg/internal/statsolly/ebpf/stats_tracer_test.go` — updated test cases
- `pkg/internal/statsolly/stats/tracer_ringbuf.go` — ring buffer event parser
- `pkg/internal/statsolly/stats/tracer_ringbuf_test.go` — end-to-end parsing + scaling verification tests
- `pkg/internal/statsolly/export/printer.go` — debug print support

**Export:**
- `pkg/export/feature.go` — `FeatureStatsTCPConnectionSummary` flag + mapper
- `pkg/export/attributes/metric.go` — metric name definition
- `pkg/export/attributes/attr_defs.go` — attribute selection definitions
- `pkg/export/otel/metrics_stats.go` — 6 OTel histogram exporters via `Expirer` pattern
- `pkg/export/prom/prom_stats.go` — 6 Prometheus histogram exporters

**Documentation & Schema:**
- `SUPPORT_MATRIX.md` — added TCP Connection Summary entry to Statistical Metrics table
- `devdocs/metrics.md` — updated known limitations and performance considerations
- `schemas/obi/groups/stats.yaml` — registered 6 new metrics in the schema registry

## AI Usage Disclosure

An AI coding assistant was used as a tool during this contribution. Specifically:
- AI was used to explore the codebase, understand existing patterns, and scaffold initial code structure
- All generated code was manually reviewed, corrected, and tested by the contributor before submission
- Key corrections made during human review: fixed mdev_us kernel scaling (`>> 2` was missing), added srtt_us filtering for consistency with the existing probe, removed 4 low-value instantaneous snapshot fields, and fixed gofmt alignment issues identified by a reviewer
- The contributor has verified every change, understands the rationale, and takes full responsibility for the submission

## Validation

- [x] `go build ./pkg/internal/statsolly/... ./pkg/export/...` — passes
- [x] `go vet ./pkg/internal/statsolly/... ./pkg/export/...` — passes
- [x] `go test -v ./pkg/internal/statsolly/...` — all 13 tests pass (7 FixupSpec subtests + 5 printer tests + 3 new ring buffer parsing/scaling tests)
- [x] `bpf2go` code generation verified — generated `StatsTcpConnectionSummaryT` matches hand-written Go struct (64 bytes)

Relates to #2453


pkg/export/otel/metrics_stats.go
```diff
@@ -210,6 +216,66 @@ func newStatMetricsExporter(
...
+		srttLog := log.With("metricFamily", "StatsTCPConnectionSummarySrtt")
+		tcpConnSrtt, err := ebpfEvents.Float64Histogram(
+			attributes.StatTCPConnectionSummary.OTEL+".srtt",
+			metric2.WithUnit("s"),
+		)
+		if err != nil {
+			srttLog.Error("creating stats tcp connection summary srtt histogram", "error", err)
+			return nil, err
+		}
+		nme.tcpConnSummarySrtt = NewExpirer[*ebpf.Stat, metric2.Float64Histogram, float64](ctx, tcpConnSrtt, connSummaryAttrs, timeNow, cfg.Metrics.TTL)
```
pinoOgni, line 246: These six histograms need explicit views, or they'll fall back to OTel's default explicit-bucket boundaries. You can add view registrations mirroring the Prometheus buckets you already defined in `prom_stats.go`, e.g. next to the existing `metric.WithView(statHistogramView(attributes.StatTCPRtt.OTEL, cfg.Buckets.StatTCPRttHistogram, ...)` call

bpf/statsolly/types.h
```diff
@@ -11,6 +11,8 @@ enum {
     k_event_stat_tcp_failed_connection = 2, // StatTypeTCPFailedConnection
     k_event_stat_tcp_retransmit = 3,        // StatTypeTCPRetransmit
     k_event_stat_tcp_io = 4,                // StatTypeTCPIo
+    // 5, 6 reserved for future stat types
```
pinoOgni: what about adding the issue and the metric names? Something like:
`// 5, 6 reserved for TCPZeroWindow / TCPOoo (see #2453)`

pkg/export/prom/prom_stats.go
```diff
@@ -159,6 +166,62 @@ func newStatsReporter(
...
+			NativeHistogramBucketFactor:     cfg.Config.NativeHistogram.BucketFactor,
+			NativeHistogramMaxBucketNumber:  cfg.Config.NativeHistogram.MaxBucketNumber,
+			NativeHistogramMinResetDuration: cfg.Config.NativeHistogram.MinResetDuration,
+		}, labelNames(mr.tcpConnSummaryAttrs)).MetricVec, timeNow, cfg.Config.TTL)
+		register = append(register, mr.tcpConnSummarySrtt)
+
+		mr.tcpConnSummaryMdev = NewExpirer[prometheus.Histogram](prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:                            attributes.StatTCPConnectionSummary.Prom + "_mdev_seconds",
+			Help:                            "RTT mean deviation at connection close, in seconds",
+			Buckets:                         cfg.Config.Buckets.StatTCPRttHistogram,
```
pinoOgni: `mr.tcpConnSummaryMdev` reuses `StatTCPRttHistogram` like ` tcpConnSummarySrtt. srtt` which makes sense (it's literally  `obi_stat_tcp_rtt_seconds` metric), but `mdev` is a different statistic. 

Did you consider a separate bucket config for `mdev` (or a new shared one scoped to the connection-summary family, decoupled from the legacy standalone `tcp_rtt` histogram)?

pkg/internal/statsolly/ebpf/stat.go
```diff
@@ -106,6 +110,17 @@ type TCPIo struct {
 	Bytes     uint32 `json:"bytes"`
 }
 
+// TCPConnectionSummary holds lifetime stats for a closed TCP connection.
```
pinoOgni: nit: comment not needed here.

pkg/internal/statsolly/ebpf/stat.go
```diff
@@ -152,3 +167,18 @@ type StatsTCPIo struct {
 
 // TCPIoBatchSize mirrors k_tcp_io_batch_size in bpf/statsolly/types.h.
 const TCPIoBatchSize = 10
+
+// StatsTCPConnectionSummary mirrors tcp_connection_summary_t from bpf/statsolly/k_tcp.c.
```
pinoOgni: Same this one.

pkg/internal/statsolly/ebpf/stat_getters.go
```diff
@@ -58,6 +58,10 @@ func StatGetters(name attr.Name) (attributes.Getter[*Stat, attribute.KeyValue],
 				role = s.TCPFailedConnection.Role
 			case StatTypeTCPRtt:
 				role = s.TCPRtt.Role
+			case StatTypeTCPConnectionSummary:
+				if s.TCPConnectionSummary != nil {
```
pinoOgni: nit: not needed. `handleStatEvent` guarantees the field is non-nil whenever `Type` matches (same for `readTCPConnectionSummaryIntoStat`, which always sets both together).

pkg/internal/statsolly/ebpf/stat.go
```diff
@@ -16,6 +16,9 @@ const (
 	StatTypeTCPFailedConnection
 	StatTypeTCPRetransmit
 	StatTypeTCPIo
+	_ // 5 reserved
```
pinoOgni: Same here, a comment about the related issue.

pinoOgni (changes requested):
Hi @chengshuyi  thanks so much for this contribution! 

Left a few comments below.

pkg/export/feature.go
```diff
@@ -55,6 +56,7 @@ var FeatureMapper = map[string]Features{
 	"stats_tcp_failed_connections": FeatureStatsTCPFailedConnections,
 	"stats_tcp_retransmits":        FeatureStatsTCPRetransmits,
 	"stats_tcp_io":                 FeatureStatsTCPIo,
+	"stats_tcp_connection_summary": FeatureStatsTCPConnectionSummary,
```
MrAlias, line 59: This only registers the new feature for the legacy `metrics.features` path. The v2 path translates `capture.network.stats.features` through `statsFeatureMask` and `statsFeatures`, and both still know only the four existing families.

As a result, `features: [tcp_connection_summary]` silently enables nothing, and exporting a config that contains `FeatureStatsTCPConnectionSummary` drops it. Can we add the v2 feature name to both conversion directions, cover the round trip, and update the v2 config docs and example?

pkg/export/prom/prom_stats.go
```diff
@@ -159,6 +166,62 @@ func newStatsReporter(
...
+			NativeHistogramBucketFactor:     cfg.Config.NativeHistogram.BucketFactor,
+			NativeHistogramMaxBucketNumber:  cfg.Config.NativeHistogram.MaxBucketNumber,
+			NativeHistogramMinResetDuration: cfg.Config.NativeHistogram.MinResetDuration,
+		}, labelNames(mr.tcpConnSummaryAttrs)).MetricVec, timeNow, cfg.Config.TTL)
+		register = append(register, mr.tcpConnSummaryMdev)
+
+		mr.tcpConnSummaryRetrans = NewExpirer[prometheus.Histogram](prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:    attributes.StatTCPConnectionSummary.Prom + "_retransmits",
+			Help:    "total retransmissions at connection close",
+			Buckets: []float64{0, 1, 2, 5, 10, 25, 50, 100, 250, 500, 1000},
```
MrAlias: These four count histograms omit the `NativeHistogram...` options that the new SRTT/MDEV histograms and existing histogram exporters apply. With the default native-histogram config enabled, retransmits, OOO, `segs_out`, and `segs_in` therefore expose only classic buckets while the rest of the metric family exposes native histograms.

Can we pass `BucketFactor`, `MaxBucketNumber`, and `MinResetDuration` to all four `HistogramOpts` blocks and add coverage that gathers one of these metrics under a non-default native schema?

MrAlias (changes requested):
I found two additional user-visible integration gaps that should be addressed before merging.

MrAlias:
Thanks for addressing the implementation feedback so far. Taking a step back, I don't think #2453 was ready to move into an implementation phase yet. It is still an open, unprioritized design/epic issue, and the public telemetry contract has not been solidified through broader maintainer and community discussion.

The main concern is the shape of the feature rather than the mechanics of this patch. The connection fields are exported as six independent histograms, which loses the per-connection correlation needed for derived quality signals such as retransmit or reordering ratios. The lifetime counts also scale with connection size, so high percentiles of retransmits, out-of-order packets, or segments do not by themselves distinguish a degraded connection from a large or long-lived one. SRTT duplicates the existing close-time RTT signal, all observations are delayed until `tcp_close`, and adding six histogram families to the aggregate `stats` feature has a non-trivial storage and default-behavior impact.

The related semantic-conventions proposal, open-telemetry/semantic-conventions#3682, is also still open and its RTT discussion is moving toward `rtt_min`, explicit observation-point/path semantics, and a generic latency/jitter model. Publishing another OBI-specific metric family before that direction is settled risks locking in semantics and names that we will need to migrate later.

Before this is ready to merge, I think we need broader agreement on the operator questions this feature should answer, which normalized signals are actually useful, whether the correlated data should be represented as aggregate metrics or events, the close-time sampling bias, expected cardinality and storage cost, and whether an experiment should be standalone and opt-in rather than included in `stats`.

I think this PR should be closed or converted to draft, but I would like other maintainers and community members to weigh in. I will also follow up on #2453 so the durable design discussion and decisions live there. Until that design is agreed, I don't think we should ask for more implementation polish or merge the current metric contract.

MrAlias:
Closing as this looks abandoned. Please re-open if this is in error.

PR #2607, opened 2026-07-08
netolly: fix RingBufTracer "queue is closed" panic on shutdown

### LINKED ISSUE #2606: netolly RingBufTracer panics with "queue is closed" on shutdown due to ClosingAttempts mismatch
### Component

OBI (eBPF instrumentation)

### What happened?

On shutdown, the netolly (network observability) pipeline can crash with a
`panic: ebpfFlows: queue is closed` instead of exiting gracefully.

The `ebpfFlows` queue has **two** senders that feed the same queue:

- `MapTracer` (`pkg/internal/netolly/flow/tracer_map.go`)
- `RingBufTracer` (`pkg/internal/netolly/flow/tracer_ringbuf.go`)

Both goroutines call `out.MarkCloseable()` (via `defer`) when their trace loops
return. However, the `ebpfFlows` queue is created in
`pkg/netolly/agent/pipeline.go` (line 55) via
`msgh.QueueFromConfig[[]*ebpf.Record](f.cfg, "ebpfFlows")` **without** overriding
`ClosingAttempts`, so it uses the default value of `1`
(see `pkg/pipe/msg/queue.go:43`).

Because `closingAttempts` defaults to `1`, the queue closes after the **first**
sender marks it closeable, even though a second sender is still alive:

1. On context cancellation, `MapTracer` (a blocking `select` loop) returns
   immediately and its deferred `MarkCloseable()` decrements the closer counter
   `1 -> 0`, which **closes the queue**.
2. Meanwhile `RingBufTracer` is still blocked inside `ReadRingBuf()`. When it
   returns with a record, it calls `Send(...)` at
   `pkg/internal/netolly/flow/tracer_ringbuf.go:123`, which triggers
   `assertNotClosed()` and **panics**.

A contributing factor: `listenAndForwardRingBuffer` uses the deprecated
`Send(...)` (no context) instead of `SendCtx(ctx, ...)`. As a result it has no
opportunity to bail out when the context is already cancelled — unlike
`MapTracer.evictFlows`, which correctly uses `SendCtx(ctx, ...)`
(`pkg/internal/netolly/flow/tracer_map.go:135`).

This only manifests on the shutdown path. Steady-state operation is unaffected,
but the process exits via a crash rather than a graceful shutdown.

The netolly pipeline should shut down gracefully on context cancellation. The
`ebpfFlows` queue should only be closed after **both** of its senders
(`MapTracer` and `RingBufTracer`) have marked it closeable, and in-flight sends
occurring during shutdown should not panic the process.

### Steps to reproduce

1. Run OBI with network flow collection enabled (netolly pipeline active), on a
   host generating enough network traffic that the eBPF flow map occasionally
   overflows and emits records through the ring buffer.
2. While the `RingBufTracer` is blocked in `ReadRingBuf()` and network traffic
   is still flowing, cancel the root context (e.g. send `SIGTERM`/`SIGINT` to
   stop the agent).
3. Observe that, when the timing lines up (`MapTracer` closes the queue before
   `RingBufTracer`'s pending record is sent), the process panics with
   `panic: ebpfFlows: queue is closed` instead of shutting down cleanly.

Note: this is a shutdown-time race, so reproduction is timing-dependent.

### Relevant logs

```shell
panic: ebpfFlows: queue is closed

goroutine 260 [running]:
go.opentelemetry.io/obi/pkg/pipe/msg.(*Queue[...]).assertNotClosed(...)
        /src/pkg/pipe/msg/queue.go:419
go.opentelemetry.io/obi/pkg/pipe/msg.(*Queue[...]).Send(...)
        /src/pkg/pipe/msg/queue.go:196
go.opentelemetry.io/obi/pkg/internal/netolly/flow.(*RingBufTracer).listenAndForwardRingBuffer(...)
        /src/pkg/internal/netolly/flow/tracer_ringbuf.go:123
go.opentelemetry.io/obi/pkg/netolly/agent.init.func2.(*RingBufTracer).TraceLoop.1(...)
        /src/pkg/internal/netolly/flow/tracer_ringbuf.go:90
go.opentelemetry.io/obi/pkg/pipe/swarm.(*Runner).Start.func2()
        /src/pkg/pipe/swarm/runner.go:82
created by go.opentelemetry.io/obi/pkg/pipe/swarm.(*Runner).Start in goroutine 62
```

### Additional context

Proposed fix (two complementary changes):

1. Create the `ebpfFlows` queue with `msg.ClosingAttempts(2)` so it is only
   closed after **both** senders (`MapTracer` and `RingBufTracer`) have called
   `MarkCloseable()`. `msgh.QueueFromConfig` already accepts override
   `QueueOpts`, so this can be passed at the call site in
   `pkg/netolly/agent/pipeline.go:55`.
2. Defensive hardening: change
   `pkg/internal/netolly/flow/tracer_ringbuf.go:123` from `Send(...)` to
   `SendCtx(ctx, ...)`, so that sends are safely dropped once the context is
   cancelled (mirroring `MapTracer.evictFlows`).

### Version

v0.10.0

### Kernel version

5.10

### OS / Distribution

Linux (Kubernetes node)

### Deployment

Kubernetes

NameHaibinZhang:
## Summary

The netolly flow pipeline creates a single `ebpfFlows` queue that has **two senders**: `MapTracer` and `RingBufTracer`. Both mark the queue closeable on shutdown, but the queue was created with the default `ClosingAttempts` of `1`. On context cancellation, `MapTracer` could close the queue first, so a subsequent `Send()` from `RingBufTracer` would panic with `queue is closed`.

This PR applies two fixes:

1. **`pkg/netolly/agent/pipeline.go`** — create the `ebpfFlows` queue with `msg.ClosingAttempts(2)`, so the queue is only closed after *both* senders have marked it closeable.
2. **`pkg/internal/netolly/flow/tracer_ringbuf.go`** — replace the deprecated `Send(...)` with `SendCtx(ctx, ...)` (with `ctx` threaded through the function signature), so in-flight sends are dropped once the context is cancelled, mirroring `MapTracer.evictFlows`.

Fixes #2606

## Validation

### DIFF
--- a/pkg/internal/netolly/flow/tracer_ringbuf.go
+++ b/pkg/internal/netolly/flow/tracer_ringbuf.go
@@ -87,7 +87,7 @@ func (m *RingBufTracer) TraceLoop(out *msg.Queue[[]*ebpf.Record]) swarm.RunFunc 
 				rtlog.Debug("exiting trace loop due to context cancellation")
 				return
 			default:
-				if err := m.listenAndForwardRingBuffer(debugging, out); err != nil {
+				if err := m.listenAndForwardRingBuffer(ctx, debugging, out); err != nil {
 					if errors.Is(err, ringbuf.ErrClosed) {
 						rtlog.Debug("Received signal, exiting..")
 						return
@@ -100,7 +100,7 @@ func (m *RingBufTracer) TraceLoop(out *msg.Queue[[]*ebpf.Record]) swarm.RunFunc 
 	}
 }
 
-func (m *RingBufTracer) listenAndForwardRingBuffer(debugging bool, forwardCh *msg.Queue[[]*ebpf.Record]) error {
+func (m *RingBufTracer) listenAndForwardRingBuffer(ctx context.Context, debugging bool, forwardCh *msg.Queue[[]*ebpf.Record]) error {
 	event, err := m.ringBuffer.ReadRingBuf()
 	if err != nil {
 		return fmt.Errorf("reading from ring buffer: %w", err)
@@ -120,7 +120,7 @@ func (m *RingBufTracer) listenAndForwardRingBuffer(debugging bool, forwardCh *ms
 		m.mapFlusher.Flush()
 	}
 
-	forwardCh.Send([]*ebpf.Record{ebpf.NewRecord(readFlow.Id, readFlow.Metrics)})
+	forwardCh.SendCtx(ctx, []*ebpf.Record{ebpf.NewRecord(readFlow.Id, readFlow.Metrics)})
 
 	return nil
 }
--- a/pkg/netolly/agent/pipeline.go
+++ b/pkg/netolly/agent/pipeline.go
@@ -51,8 +51,10 @@ func (f *Flows) buildPipeline(ctx context.Context) (*swarm.Runner, error) {
 	}
 
 	swi := &swarm.Instancer{}
-	// Start nodes: those generating flow records (reading them from eBPF)
-	ebpfFlows := msgh.QueueFromConfig[[]*ebpf.Record](f.cfg, "ebpfFlows")
+	// Start nodes: those generating flow records (reading them from eBPF).
+	// ebpfFlows has two senders (MapTracer and RingBufTracer), so it must only be
+	// closed after both of them have marked it closeable, hence ClosingAttempts(2).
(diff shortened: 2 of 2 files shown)


mariomac (approved):
Nice!

PR #2608, opened 2026-07-08
fix(ebpf): share puma correlation maps across BPF collections

### LINKED ISSUE #2582: [BUG] Puma trace correlation always fails: puma_task_connections / puma_worker_tasks are not OBI_PIN_INTERNAL, so the injection path reads an empty map instance
### Component

OBI (eBPF instrumentation)

### What happened?

Distributed traces through a Ruby on Rails / Puma service are always broken for requests where Puma's reactor thread reads the request and a worker thread performs the outbound HTTP call (i.e. every reused keep-alive connection — the exact scenario the Puma support in #968 was written to fix). The outbound request gets a freshly minted trace id instead of the incoming one.

The Puma-specific uprobes from #968 (`rb_obj_call_init_kw`, `rb_ary_shift`) fire and store their correlation entries correctly, but `find_puma_parent_trace()` never finds them: the two maps in `bpf/maps/puma_tasks.h` are declared **without** `__uint(pinning, OBI_PIN_INTERNAL)`. Map sharing across BPF collections only happens for `PinInternal` maps (`ResolveMaps()` in `pkg/internal/ebpf/convenience/convenience.go` routes them through `MapReplacements`), so the writer (ruby uprobes, generictracer instance) and the reader (`tpinjector.c` includes `common/trace_parent.h` and calls `find_parent_trace_for_client_request` from `obi_packet_extender`; generictracer is also instantiated per traced executable) operate on **different, unshared map instances**. The lookup misses 100% of the time.

Every other map consumed by the `find_parent_trace` chain follows the pinning convention (`server_traces`, `server_traces_aux`, `java_vt_threads` from #2261) — the two puma maps appear to be an omission.

Applying the pinning to the two maps (diff below) takes our reproduction from **0/10 to 10/10** correct propagation, with `find_puma_parent_trace` hitting on every request.

The traceparent of the incoming request is propagated to the outbound HTTP call made by the Puma worker thread, as designed in #968: `rb_ary_shift` binds the worker tid to the Client object, `find_puma_parent_trace()` resolves worker tid → object → connection → `server_traces_aux` and reuses the incoming trace context.

### Steps to reproduce

Topology: `HTTP client → Puma (Rails, makes an outbound Net::HTTP call) → downstream Go service that echoes back the traceparent it received`. (We run it on Kubernetes with an Istio sidecar on the Puma pod, but the mechanism is not mesh-specific — any client that reuses a keep-alive connection to Puma works.)

1. Puma with `workers 2`, `threads 5`, `queue_requests true` (reactor enabled — defaults).
2. Instrument the Puma process with OBI, `context_propagation: all`.
3. Open **one** keep-alive connection to Puma and send a request with a distinct, recognizable `traceparent` header; the Rails handler calls the downstream service; compare the trace id the downstream service received against the one sent.
4. Repeat on the same connection: (a) with 20s idle gaps between requests (reactor reads the socket, hands off to a worker → cross-thread case), and (b) rapid-fire ~0.2s apart.

Observed (OBI vendored in Beyla 3.25.0, stock): idle **0/10** match, rapid **1/10** (only the very first request on a fresh connection matches — that one resolves through the pinned `server_traces` legacy path, not the puma path).

Also tested `context_propagation: headers`: still 0/10 — consistent with per-executable generictracer instances, ruling out "only the sk_msg path is affected".

### OBI configuration

```yaml
ebpf:
  wakeup_len: 1
  context_propagation: all
  track_request_headers: true
  high_request_volume: true
otel_traces_export:
  endpoint: http://alloy.monitoring.svc.cluster.local:4317
  protocol: grpc
discovery:
  instrument:
    - k8s_namespace: "default"
    - k8s_namespace: "monitoring"
routes:
  unmatched: heuristic
attributes:
  kubernetes:
    enable: autodetect


(Deployed via the Beyla helm chart; `OTEL_EBPF_BPF_DEBUG=true`, `BEYLA_TRACE_PRINTER=text` for the evidence below.)
```

### Relevant logs

```shell
Store succeeds (uprobe side, worker thread tid 10878):


03:30:28.569 msg="rb_ary_shift ret ==> slice ffff8c032d20, item ffff8c1fcbd8, flags a007 [obi_rb_ary_shift]" pid=10878 comm="puma srv tp 001"
03:30:28.569 msg="stored item to id correlation, id = 2a7700002a7e, item ffff8c1fcbd8 [obi_rb_ary_shift]" pid=10878 comm="puma srv tp 001"


6 ms later, the injection path on the **same thread** misses and mints a new trace id:


03:30:28.575 msg="puma lookup: task_id=0 [find_puma_parent_trace]" pid=10878 comm="puma srv tp 001"
03:30:28.575 msg="written TP to headers: 00-0c466374bd1a4f2f74ca099dda7fea75-cdf708ab101d7834-01" pid=10878 comm="puma srv tp 001"


A same-key store followed by a same-key miss on the same thread rules out key-derivation and LRU-eviction explanations and points at two different map instances. Aggregate over a 20-request run: `stored item to id correlation` ×26, `puma lookup: task_id=0` ×613, `found item` ×**0**.

With the pinning patch applied (same protocol): idle **10/10**, rapid **10/10**, `find_puma_parent_trace` non-zero on every request, `found item` present throughout.

Proposed fix (tested):


--- a/bpf/maps/puma_tasks.h
+++ b/bpf/maps/puma_tasks.h
@@
 #include <common/connection_info.h>
 #include <common/map_sizing.h>
 #include <common/puma_task_id.h>

 struct {
     __type(key, puma_task_id_t); // the array and item pair
     __type(value, connection_info_part_t);
     __uint(max_entries, MAX_CONCURRENT_REQUESTS);
 } puma_task_connections SEC(".maps");

 struct {
     __type(key, u64);              // the pid:tgid
     __type(value, puma_task_id_t); // the array and item pair
     __uint(max_entries, MAX_CONCURRENT_REQUESTS);
 } puma_worker_tasks SEC(".maps");


Happy to submit this as a PR.
```

### Additional context

- Why existing tests likely pass: a single-request-per-connection test matches through the pinned `server_traces` path (same-thread read/write on a fresh connection) without ever needing the puma maps. Catching this class of bug needs a keep-alive client connection with either idle gaps or rapid reuse, so the reactor performs the read.
- Suggested follow-up: audit other non-pinned maps consumed from `common/*.h` headers included by more than one collection (`nginx_fd` / `fd_info` looked like candidates from a quick scan).
- The bug is invisible at the uprobe level — `rb_ary_shift` debug output looks perfectly healthy — which makes it easy to misattribute to the Puma threading model itself.

### Version

OBI as vendored in grafana/beyla 3.25.0 (submodule commit `4bf91071`, ≈ v0.9.0). The relevant code is unchanged on current `main`. Also reproduced with beyla 3.7.0.

### Kernel version

6.10.14 (aarch64; Docker Desktop linuxkit VM hosting a kind node)

### OS / Distribution

Node: Debian GNU/Linux 12 (bookworm) (kind, Kubernetes v1.34.0); Puma workload: Rails 7.2 / Puma 6.6 on the official ruby slim image; containerd 2.1.3.

### Deployment

None

KarasWinds:
## Summary

Fixes #2582.

The Puma/Ruby trace-correlation support from #968 stores its correlation entries in `puma_task_connections` and `puma_worker_tasks` (`bpf/maps/puma_tasks.h`), but both maps were declared without `__uint(pinning, OBI_PIN_INTERNAL)`. Maps are only shared across BPF collections when they are `PinInternal` (`ResolveMaps()` in `pkg/internal/ebpf/convenience/convenience.go` routes them through `MapReplacements`), so:

- the **writers** (ruby uprobes `obi_rb_obj_call_init_kw` / `obi_rb_ary_shift`, in a per-executable generictracer instance) populate one map instance, while
- the **reader** (`find_puma_parent_trace()` via `common/trace_parent.h`, reached from the tpinjector sk_msg path with `context_propagation: all`, and from other generictracer instances) looks up a different, empty instance.

The lookup therefore misses 100% of the time and a fresh trace id is minted for every outgoing request from Puma — exactly the scenario #968 was written to fix. All other maps consumed by the `find_parent_trace` chain already follow the pinning convention (`server_traces`, `server_traces_aux`, `java_vt_threads`); the two puma maps appear to be an omission.

This PR pins both maps internally (2 lines + include). Generated BPF bindings are gitignored in this repo, so the diff is the 3-line header change only; verified locally with `make docker-generate` (idempotent on second run) and `clang-format --dry-run --Werror`.

## Validation

Reproduction from #2582 (Rails 7.2 / Puma 6.6, `workers 2` / `threads 5` / `queue_requests true`, single keep-alive client connection, distinct recognizable traceparent per request, downstream Go service echoes the traceparent it received; `context_propagation: all`):

| | before | after (this patch) |
|---|---|---|
| idle-20s rounds (reactor reads, worker writes) | 0/10 | **10/10** |
| rapid keep-alive reuse (~0.2s apart) | 1/10 (first request on the fresh connection only) | **10/10** |
| `find_puma_parent_trace` debug | `task_id=0` on all lookups, `found item` ×0 | non-zero on every request, `found item` throughout |

The "before" failure signature that pinpoints the unshared maps — the same thread stores and then misses the same key 6 ms later:

```
03:30:28.569 "stored item to id correlation, id = 2a7700002a7e, item ffff8c1fcbd8" pid=10878 comm="puma srv tp 001"
03:30:28.575 "puma lookup: task_id=0 [find_puma_parent_trace]"                    pid=10878 comm="puma srv tp 001"
```

Note on test coverage: single-request-per-connection tests pass without the puma maps (fresh connection → worker reads directly → pinned `server_traces` legacy path), which is likely why this wasn't caught. A regression test needs a reused keep-alive connection so the reactor performs the read — happy to work on one as a follow-up if useful.

**AI disclosure** (per `AI-POLICY.md`): root-cause analysis, patch, and validation were AI-assisted (Claude Code); all changes and test results were personally reviewed and reproduced by the submitter.

## Validation checklist

- [x] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed (bugfix — no feature surface or support-matrix change)

### DIFF
--- a/bpf/maps/puma_tasks.h
+++ b/bpf/maps/puma_tasks.h
@@ -8,6 +8,7 @@
 
 #include <common/connection_info.h>
 #include <common/map_sizing.h>
+#include <common/pin_internal.h>
 #include <common/puma_task_id.h>
 
 struct {
@@ -15,6 +16,7 @@ struct {
     __type(key, puma_task_id_t); // the array and item pair
     __type(value, connection_info_part_t);
     __uint(max_entries, MAX_CONCURRENT_REQUESTS);
+    __uint(pinning, OBI_PIN_INTERNAL);
 } puma_task_connections SEC(".maps");
 
 struct {
@@ -22,4 +24,5 @@ struct {
     __type(key, u64);              // the pid:tgid
     __type(value, puma_task_id_t); // the array and item pair
     __uint(max_entries, MAX_CONCURRENT_REQUESTS);
+    __uint(pinning, OBI_PIN_INTERNAL);
 } puma_worker_tasks SEC(".maps");


mmat11 (approved):
lgtm!

MrAlias (approved):

grcevski (approved):
LGTM!

grcevski:
Thanks so much for your contribution @KarasWinds !

PR #2574, opened 2026-07-02
Tighten release artifact Cosign verification

MrAlias:
### Motivation

- Fix a supply-chain verification weakness in `RELEASING.md` where `cosign` verification used a repository-prefix `--certificate-identity-regexp` and could accept OIDC signatures from non-release workflows in the same repository.

### Description

- Update `RELEASING.md` to construct `certificate_identity="https://github.com/${repository}/.github/workflows/release.yml@refs/tags/${release_tag}"` and use `--certificate-identity "${certificate_identity}"` in `cosign verify-blob` so verification is bound to the release workflow and tag.

### Testing

- `make lint-markdown`

PR #2609, opened 2026-07-08
Add OBI nightly releases + sha based image tags on main

### LINKED ISSUE #2553: Add OBI nightly releases
With the current pace of feature releases in OBI and with our emphasis on stable versioning, it makes it hard to let people test out features until we decide to cut an official release.
there is always an option to take the `main` image but its a mutable tag and can release between pull versions.

My suggestion is to have a similar workflow to the [otel collector nightly release](https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/.github/workflows/nightly-release.yaml). running every night and releasing 2 docker images.

Versioned: `ebpf-instrument:0.10.0-nightly.<shortcommit>` - immutable
Ephemeral: `ebpf-instrument:nightly` - mutable

This can run for both `obi` and the `k8s-cache` components.

NimrodAvni78:
## Summary

Resolves #2553 

## Changes
- created a new `Nightly Release` pipeline which runs at `"0 2 * * *"` (every day at 2AM UTC) and publishes 2 tags, daily `nightly-<YYYYMMDD>` and a mutable `nightly`
- added a `type=sha,prefix=,format=short` tag to the metadata step, which on every master push will release an exact `ebpf-instrument:aaaaaaa` style image tag (short commit sha)

## Validation

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -14,8 +14,7 @@ on:
     - cron: "0 2 * * *"
   workflow_dispatch:
 
-permissions:
-  contents: read
+permissions: {}
 
 jobs:
   # Derive the dated nightly version string (nightly-<YYYYMMDD>).
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -1,0 +1,61 @@
+name: Nightly Release
+
+# Publishes nightly multi-arch images for both the OBI and k8s-cache components,
+# modelled on the OpenTelemetry Collector nightly release
+# (https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/.github/workflows/nightly-release.yaml).
+#
+# Tag scheme agreed in the SIG (issue #2553):
+#   - Versioned (immutable):  nightly-<YYYYMMDD>  e.g. nightly-20260708
+#   - Ephemeral (mutable):    nightly             (always points at the latest nightly)
+
+on:
+  schedule:
+    # Every day at 02:00 UTC.
+    - cron: "0 2 * * *"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  # Derive the dated nightly version string (nightly-<YYYYMMDD>).
+  compute-version:
+    name: Compute nightly version
+    if: github.repository == 'open-telemetry/opentelemetry-ebpf-instrumentation'
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.version.outputs.version }}
+    steps:
+      - name: Compute nightly version
+        id: version
+        run: |
+          set -euo pipefail
+          VERSION="nightly-$(date -u +'%Y%m%d')"
+          echo "Nightly version: ${VERSION}"
+          echo "version=${VERSION}" >> "${GITHUB_OUTPUT}"
+
+  nightly-obi:
+    name: Publish nightly OBI image
+    needs: compute-version
+    uses: ./.github/workflows/publish_dockerhub_main.yml
+    permissions:
+      contents: read
+      id-token: write
+      packages: write
+    with:
+      nightly_version: ${{ needs.compute-version.outputs.version }}
+    secrets:
+      DOCKER_TOKEN_EBPF_INSTRUMENTATION: ${{ secrets.DOCKER_TOKEN_EBPF_INSTRUMENTATION }}
+
+  nightly-k8s-cache:
+    name: Publish nightly k8s-cache image
+    needs: compute-version
+    uses: ./.github/workflows/publish_dockerhub_k8s_cache_main.yml
+    permissions:
+      contents: read
+      id-token: write
+      packages: write
+    with:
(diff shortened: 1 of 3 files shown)


.github/workflows/nightly-release.yml
```diff
@@ -0,0 +1,61 @@
...
+#   - Ephemeral (mutable):    nightly             (always points at the latest nightly)
+
+on:
+  schedule:
+    # Every day at 02:00 UTC.
+    - cron: "0 2 * * *"
+  workflow_dispatch:
+
+permissions:
+  contents: read
```
skl: Workflow-level permissions might be able to be lower here, as there's no checkout and the nightly jobs define their own permissions

```suggestion
  contents: {}
```

skl (commented):
Worth a note in the README about the new images? Bear in mind incoming changes from #2574

.github/workflows/publish_dockerhub_main.yml
```diff
@@ -78,10 +83,13 @@ jobs:
...
             latest=false
           tags: |
-            type=semver,pattern=v{{version}},value=${{ inputs.ref }}
+            type=semver,pattern=v{{version}},value=${{ inputs.ref }},enable=${{ inputs.ref != '' }}
             type=raw,value=latest,enable=${{ inputs.latest == true }}
-            type=ref,event=branch
+            type=raw,value=${{ inputs.nightly_version }},enable=${{ inputs.nightly_version != '' }}
+            type=raw,value=nightly,enable=${{ inputs.nightly_version != '' }}
+            type=ref,event=branch,enable=${{ inputs.nightly_version == '' }}
             type=ref,event=tag
```
MrAlias: This still leaves tag refs enabled in the nightly path. If `nightly-release.yml` is manually dispatched from a tag ref, `nightly_version` is set but `docker/metadata-action` can still emit that selected tag, so the nightly OBI build can publish a release tag alongside `nightly`.

Can we gate tag refs the same way as branch refs so nightly runs only publish the agreed `nightly-YYYYMMDD` and `nightly` tags?

```suggestion
            type=ref,event=tag,enable=${{ inputs.nightly_version == '' }}
```

.github/workflows/publish_dockerhub_main.yml
```diff
@@ -157,10 +165,13 @@ jobs:
...
```
MrAlias: Same issue in the manifest/tagging step. Even if the per-arch build is gated correctly, this metadata output drives `imagetools create` and signing, so a tag-ref manual nightly run can still create and sign the release tag.

```suggestion
            type=ref,event=tag,enable=${{ inputs.nightly_version == '' }}
```

.github/workflows/publish_dockerhub_k8s_cache_main.yml
```diff
@@ -78,10 +83,13 @@ jobs:
...
             latest=false
           tags: |
-            type=semver,pattern=v{{version}},value=${{ inputs.ref }}
+            type=semver,pattern=v{{version}},value=${{ inputs.ref }},enable=${{ inputs.ref != '' }}
             type=raw,value=latest,enable=${{ inputs.latest == true }}
-            type=ref,event=branch
+            type=raw,value=${{ inputs.nightly_version }},enable=${{ inputs.nightly_version != '' }}
+            type=raw,value=nightly,enable=${{ inputs.nightly_version != '' }}
+            type=ref,event=branch,enable=${{ inputs.nightly_version == '' }}
             type=ref,event=tag
```
MrAlias: This leaves the same tag-ref path open for the k8s-cache nightly build. Manual dispatch from a tag ref can still add that release tag to the nightly image.

```suggestion
            type=ref,event=tag,enable=${{ inputs.nightly_version == '' }}
```

.github/workflows/publish_dockerhub_k8s_cache_main.yml
```diff
@@ -157,10 +165,13 @@ jobs:
...
```
MrAlias: Same issue in the k8s-cache manifest/tagging step; this metadata output controls the manifest tags and the set of signed references.

```suggestion
            type=ref,event=tag,enable=${{ inputs.nightly_version == '' }}
```

NimrodAvni78:
@MrAlias great catches, fixed them and validated each trigger produces only its relevant tags

MrAlias (approved):
No blocking issues from me. The nightly tag gating and short-SHA tags now look consistent.

One non-blocking follow-up: the PR description still calls `nightly-YYYYMMDD` immutable, but the final date-based scheme does not enforce immutability across another run on the same UTC date. Please update the description to match that behavior.

damemi (commented):
don't remember if it was discussed but does this also publish the generated artifacts with a nightly, or is that possible?

MrAlias:
I thought we had scoped this to Docker images(?)

NimrodAvni78:
@MrAlias @damemi i think we also scoped this to only docker images, we can extend in a follow up PR?

MrAlias:
SGTM

mmat11 (approved):
lgtm

PR #2611, opened 2026-07-08
gate JVM runtime metrics on the shared application_runtime feature

marctc:
## Summary

- Remove the `application_jvm` metrics feature; JVM probes and export now
  gate on `application_runtime`.
- Remove the `jvm_runtime_metrics.enabled` config flag
- Drop the OBI-specific `obi.jvm.heap.used` metric and use  `jvm.memory.used` with `jvm.memory.type=heap`.
- Add `devdocs/runtimes/jvm.md`.

## Validation


pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -663,6 +663,10 @@ func (p *Tracer) Run(
 	)(ctx, append(p.closers, &p.bpfObjects), eventsChan)
 }
 
+func (p *Tracer) jvmRuntimeMetricsEnabled() bool {
+	return p.cfg != nil && p.cfg.Metrics.Features.AppRuntime()
```
skl: Suggest finding a way to reuse `appolly.JoinMetricsConfig()` here (maybe by moving it to `obi.Config`?) so that per-service overrides work

pkg/appolly/app/runtime/jvm.go
```diff
@@ -123,7 +122,8 @@ func ParseJVMGCHeapSummaryEvent(
 		PID:            app.PID(nsPID),
 		PIDNamespaceID: pidNamespaceID,
 		Time:           jvmKernelTime(timestamp),
-		Kind:           JVMMetricObiHeapUsed,
+		Kind:           JVMMetricMemoryUsed,
```
skl: Does this collide with `ParseJVMMemoryPoolEvent`?

`ParseJVMMemoryPoolEvent` already emits `jvm.memory.used` per pool. The heap summary would now emit into the same metric. So `sum(jvm_memory_used_bytes{jvm_memory_type="heap"})` would count the heap twice.

Suggest removing heap summary or using another metric name for it?

pkg/export/feature.go
```diff
@@ -65,7 +64,6 @@ var FeatureMapper = map[string]Features{
 	"application_service_graph":    FeatureGraph,
 	"application_host":             FeatureApplicationHost,
 	"application_runtime":          FeatureApplicationRuntime,
-	"application_jvm":              FeatureApplicationJVM,
```
MrAlias, line 68: `application_jvm` and `jvm_runtime_metrics.enabled` were released in v0.10.0. Removing both here is silent: `LoadFeatures` looks up names in `FeatureMapper` without validating them, and YAML ignores the removed `enabled` field. An existing v0.10 config with both settings still starts, but its JVM probes and metrics are disabled.

Can we keep `application_jvm` as a deprecated alias to `FeatureApplicationRuntime` for the migration window, or fail or warn explicitly on the removed configuration, and add an upgrade regression test? We should not turn a previously valid telemetry configuration into silent data loss.

pkg/internal/ebpf/generictracer/jvm_live_test.go
```diff
@@ -276,9 +276,9 @@ func waitForJVMRuntimeEvents(
 				assert.NotZero(t, snapshot.PID)
 				assert.NotZero(t, snapshot.Time)
 
-				if snapshot.JVM.Kind == jvmruntime.JVMMetricObiHeapUsed {
+				if snapshot.JVM.Kind == jvmruntime.JVMMetricMemoryUsed && snapshot.JVM.PoolName == "" {
```
MrAlias: This condition can no longer be true. The PR removes the aggregate heap-summary event, and the remaining memory-used events come from the memory-pool path with a pool name. `seenHeap` therefore remains false, so this helper times out after receiving valid pool metrics; `TestJVMRuntimeEventsLive` is currently failing in shard 6 on every retry.

Can we update the live test to assert a heap memory-pool event instead of waiting for the removed empty-pool aggregate, and remove the obsolete `seenHeap` requirement?

MrAlias (approved):
No blocking issues from me. The implementation now consistently gates JVM runtime metrics on application_runtime, removes the heap-summary metric path, and updates the tests and docs around the memory-pool source.

One non-blocking cleanup: since this keeps application_jvm as a deprecated alias, I think the generated schema should either continue accepting that alias for the migration window, or we should drop the alias/comment/test and make the pre-1.0 break explicit. I would not block this PR on that.

grcevski (approved):
LGTM!

PR #2612, opened 2026-07-09
Simplify hostname resolver

mariomac:
## Summary

The hostname resolver was overengineered for a use case that is never happening, because we currently invoke the `Query()` method only once, during startup, and we keep using the full hostname for the whole lifespan of the OBI instance, so this PR:

- Removes the logic of "notifying when a hostname dynamically changes" (never happens because we don't invoke Query() more than once)
- Removes the subscribers/notifications logic, because no entity was subscribed to it.
- Removes the `short` hostname return value, as it was never used outside of the resolver.
- Removes the logic to override `short` hostname, as it is not possible to override it from the user side.

We removed the "update" logic from the unit tests but kept untouched the logic refering to the first query (try full hostname then backup to internal hostname then backup to logic, and avoid localhosts).

## Validation

### DIFF
--- a/pkg/internal/traces/hostname/hostname.go
+++ b/pkg/internal/traces/hostname/hostname.go
@@ -79,11 +79,9 @@ type fallbackResolver struct {
 
 // Query returns the full and the short host name, or error if none of both can't be returned.
 // This implementation assumes the full host name may fail since it can depend on an external
-// service (e.g. a DNS server). If the full name resolution fails, it considers the following
-// fallback actions (in priority):
-// 1 - return the previous successful full name resolution
-// 2 - ask for the full hostname to the OS (and consider the returned value as successful)
-// 3 - The short host name if it has never been successfully resolved.
+// Query returns the full hostname, or an error if no hostname can be resolved.
+// If full hostname resolution fails or returns a localhost name, it falls back
+// to the internal hostname and then the short hostname.
 func (r *fallbackResolver) Query() (string, error) {
 	if r.overriddenFull != "" {
 		return r.overriddenFull, nil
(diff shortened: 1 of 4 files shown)


MrAlias (commented):
I found one failure-path regression that should be addressed before merging. I also left a documentation mismatch inline.

pkg/internal/traces/hostname/hostname.go
```diff
@@ -131,124 +84,40 @@ type fallbackResolver struct {
...
+		if isLocalhost(full) {
+			if isLocalhost(short) {
 				full = ""
+			} else {
+				full = short
 			}
 		}
 	}
-	return r.updateAndGet(full, short, err)
+	return full, nil
```
MrAlias, line 115: This returns success even when every hostname source fails. In that case, `NewHostInstance` takes the success path with an empty hostname, so spans get an empty `host.name` and instance IDs such as `:<pid>` instead of the existing warning.

The old `updateAndGet` path returned an error when neither name was available. Please preserve that contract and keep a regression test for the all-resolvers-fail case.

```suggestion
	if full == "" {
		return "", errors.New("can't resolve either full or short hostname")
	}
	return full, nil
```

pkg/internal/traces/hostname/hostname.go
```diff
@@ -131,124 +84,40 @@ type fallbackResolver struct {
 // 1 - return the previous successful full name resolution
 // 2 - ask for the full hostname to the OS (and consider the returned value as successful)
 // 3 - The short host name if it has never been successfully resolved.
```
MrAlias: These comments still describe the removed cache and the old two-value return. `Query` no longer stores a previous successful result or returns the short hostname separately, so the contract now contradicts the implementation and the PR's single-query premise.

```suggestion
// Query returns the full hostname, or an error if no hostname can be resolved.
// If full hostname resolution fails or returns a localhost name, it falls back
// to the internal hostname and then the short hostname.
```

MrAlias (approved):

grcevski (approved):
LGTM!

PR #2613, opened 2026-07-09
tracers: don't add/remove pids for tracers that failed loading

mmat11:
## Summary

don't add/remove pids for tracers that failed loading

## Validation

### DIFF
--- a/pkg/appolly/discover/attacher_linux_test.go
+++ b/pkg/appolly/discover/attacher_linux_test.go
@@ -1,0 +1,78 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build linux
+
+package discover
+
+import (
+	"errors"
+	"log/slog"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"go.opentelemetry.io/obi/pkg/appolly/app/svc"
+	execpkg "go.opentelemetry.io/obi/pkg/appolly/discover/exec"
+	"go.opentelemetry.io/obi/pkg/ebpf"
+	ebpfcommon "go.opentelemetry.io/obi/pkg/ebpf/common"
+	"go.opentelemetry.io/obi/pkg/export/imetrics"
+	"go.opentelemetry.io/obi/pkg/internal/helpers/maps"
+	"go.opentelemetry.io/obi/pkg/obi"
+	"go.opentelemetry.io/obi/pkg/pipe/msg"
+)
+
+type failingLoadTracer struct {
+	recordingTracer
+}
+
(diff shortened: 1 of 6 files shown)


pkg/internal/ebpf/logenricher/logenricher_test.go
```diff
@@ -91,6 +91,13 @@ func TestBlockPIDClearsNamespacedPIDCache(t *testing.T) {
 	require.ErrorIs(t, m.Lookup(nsPk, &value), ebpf.ErrKeyNotExist)
 }
 
+func TestPIDOpsWithoutLoadedObjectsDoNotPanic(t *testing.T) {
+	tr := newTestTracer(t, false)
+
+	require.Error(t, tr.addPID(tr.pidKey(1, 42)))
+	require.Error(t, tr.removePID(tr.pidKey(1, 42)))
```
MrAlias, line 98: This test only covers the log-enricher guard directly. It does not assert the attacher behavior this change relies on: after an optional common tracer fails during `ProcessTracer.Init`, that tracer must be removed from `ta.commonTracers` while successfully loaded common tracers remain.

Without that assertion, the filtering can regress while this test still passes, and other failed common tracers can again receive PID notifications without loaded BPF objects. Can we add an attacher-level regression test with one failed and one successful common tracer, then verify only the successful tracer receives `AllowPID` and `BlockPID`?

mmat11: done

pkg/appolly/discover/attacher_linux_test.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// After an optional common tracer fails during ProcessTracer.Init, it must be
+// pruned from ta.commonTracers so that only successfully loaded common tracers
+// receive AllowPID and BlockPID notifications.
+func TestCommonTracersPrunedAfterLoadFailure(t *testing.T) {
+	okTracer := &recordingTracer{}
+	failedTracer := &failingLoadTracer{}
+
+	cfg := &obi.Config{}
```
MrAlias, line 41: Small test-hygiene nit: `ProcessTracer.Init` uses `cfg.EBPF.BPFFSPath` when it builds the `otel` bpffs path. With an empty config, this test can create an `otel` directory in the package working directory when run locally.

Using a temp dir keeps the test isolated:

```suggestion
	cfg := &obi.Config{}
	cfg.EBPF.BPFFSPath = t.TempDir()
```

MrAlias (approved):

grcevski (approved):
LGTM!

pkg/appolly/discover/attacher_linux_test.go
mmat11: done

PR #2620, opened 2026-07-10
bpf: add negative caching to pid_cache for untracked PIDs

NameHaibinZhang:
## Summary

During TCP connection storms from untracked processes, each kprobe invocation of valid_pid() incurs expensive ns_pid_ppid() + pid_matches() calls (~150ns) on every cache miss. Since untracked PIDs were never cached, the same PID would repeatedly take the slow path on subsequent send/recv/close calls.

This change caches negative results (value=0) in pid_cache, reducing repeated lookups for the same untracked PID from ~150ns to ~10ns.

To maintain correctness:
- AllowPID() now writes positive entries to pid_cache, overriding any stale negative cache entries when a PID starts being tracked.
- BlockPID() deletes the cache entry so the PID gets re-evaluated.

## Validation

### DIFF
--- a/bpf/pid/pid.h
+++ b/bpf/pid/pid.h
@@ -81,7 +81,7 @@ static __always_inline u32 valid_pid(u64 id) {
 
         // Cache negative result: PID not in tracked set
         u32 neg = 0;
-        bpf_map_update_elem(&pid_cache, &a_pid, &neg, BPF_ANY);
+        bpf_map_update_elem(&pid_cache, &a_pid, &neg, BPF_NOEXIST);
     }
 
     return 0;
--- a/bpf/pid/pid.h
+++ b/bpf/pid/pid.h
@@ -78,6 +78,10 @@ static __always_inline u32 valid_pid(u64 id) {
                 return a_pid;
             }
         }
+
+        // Cache negative result: PID not in tracked set
+        u32 neg = 0;
+        bpf_map_update_elem(&pid_cache, &a_pid, &neg, BPF_ANY);
     }
 
     return 0;
--- a/pkg/internal/ebpf/generictracer/generictracer.go
+++ b/pkg/internal/ebpf/generictracer/generictracer.go
@@ -127,11 +127,21 @@ func (p *Tracer) rebuildValidPids() {
 func (p *Tracer) AllowPID(pid app.PID, ns uint32, fi *exec.FileInfo) {
 	p.pidsFilter.AllowPID(pid, ns, fi, ebpfcommon.PIDTypeKProbes)
 	p.rebuildValidPids()
+	// Override potential negative cache entry for this PID
+	if p.bpfObjects.PidCache != nil {
+		pidU32 := uint32(pid)
+		_ = p.bpfObjects.PidCache.Put(pidU32, pidU32)
+	}
 }
 
 func (p *Tracer) BlockPID(pid app.PID, ns uint32) {
 	p.pidsFilter.BlockPID(pid, ns)
 	p.rebuildValidPids()
+	// Remove from cache so next access re-evaluates
+	if p.bpfObjects.PidCache != nil {
+		pidU32 := uint32(pid)
+		_ = p.bpfObjects.PidCache.Delete(pidU32)
+	}
 }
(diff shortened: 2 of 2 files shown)


pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -127,11 +127,21 @@ func (p *Tracer) rebuildValidPids() {
 func (p *Tracer) AllowPID(pid app.PID, ns uint32, fi *exec.FileInfo) {
 	p.pidsFilter.AllowPID(pid, ns, fi, ebpfcommon.PIDTypeKProbes)
 	p.rebuildValidPids()
+	// Override potential negative cache entry for this PID
+	if p.bpfObjects.PidCache != nil {
+		pidU32 := uint32(pid)
+		_ = p.bpfObjects.PidCache.Put(pidU32, pidU32)
```
MrAlias, line 133: This only repairs the cache entry for the PID passed to `AllowPID()`, but `valid_pid()` can also accept a process when its namespaced parent PID matches `valid_pids`. If a child PID hits the negative cache before its parent is discovered, then later allowing the parent rebuilds `valid_pids` but leaves `pid_cache[child] = 0`, so subsequent probes return before the `ns_ppid` check and the child stays filtered until the LRU evicts the entry or the child is explicitly allowed.

Can we invalidate `pid_cache` when rebuilding `valid_pids`, or otherwise key/version negative entries so they cannot survive changes to the parent allow-list?

grcevski: I believe we call AllowPID for all children processes when a parent is instrumented, but I think your suggestion is safer. Maybe we can just remove all entries from the pid_cache map when we call AllowPID. In steady state this will not matter. Also I think there's a clear api on the maps, something that drops all entries.

grcevski: From attacher:
```
	for _, ct := range ta.commonTracers {
		ct.AllowPID(ie.FileInfo.Pid(), ie.FileInfo.Ns(), ie.FileInfo)
		for _, pid := range ie.ChildPids {
			ct.AllowPID(pid, ie.FileInfo.Ns(), ie.FileInfo)
		}
	}

```

I think it's safe to have it as-is.

grcevski (approved):
LGTM!

bpf/pid/pid.h
```diff
@@ -78,6 +78,10 @@ static __always_inline u32 valid_pid(u64 id) {
                 return a_pid;
             }
         }
+
+        // Cache negative result: PID not in tracked set
+        u32 neg = 0;
+        bpf_map_update_elem(&pid_cache, &a_pid, &neg, BPF_ANY);
```
mmat11: should this be `BPF_NOEXIST`?

pkg/internal/ebpf/generictracer/generictracer.go
NameHaibinZhang: Thanks for the thoughtful analysis. As @grcevski pointed out, the attacher already calls `AllowPID` for all child processes when a parent is instrumented (iterating `ie.ChildPids`), and `AllowPID` explicitly overwrites any negative cache entry with a positive one (`pid_cache[pid] = pid`), so a stale negative entry for a child is repaired as soon as the parent (and its children) are allowed.

Additionally, per @mmat11's suggestion below, I've changed the negative cache write to use `BPF_NOEXIST`, which prevents a TOCTOU race where an in-flight probe could overwrite a fresh positive entry inserted by a concurrent `AllowPID` call.

bpf/pid/pid.h
NameHaibinZhang: Good catch — changed to `BPF_NOEXIST`. This prevents a TOCTOU race: between the cache-miss lookup (line 48) and this negative write, a concurrent `AllowPID` could insert a positive entry. With `BPF_NOEXIST`, the negative write safely fails if the key was inserted in the meantime, preserving the positive cache entry.

mariomac (approved):
👌🏻

MrAlias (approved):

PR #2623, opened 2026-07-10
postgres: cache db namespace on client startup

mmat11:
## Summary

postgres: cache db namespace on client startup

## Validation

### DIFF
--- a/pkg/ebpf/common/sql_detect_postgres.go
+++ b/pkg/ebpf/common/sql_detect_postgres.go
@@ -36,6 +36,14 @@ const (
 	// pgHeaderLen is the size of the Postgres message header:
 	// 1 byte type + 4 bytes length field.
 	pgHeaderLen = 5
+
+	// StartupMessage major protocol version (3.x: 3.0 since PG 7.4, 3.2 since PG 18)
+	pgStartupProtocolMajor = 3
+	// pgStartupMinLen: 4 bytes length + 4 bytes protocol + 1 byte terminator
+	pgStartupMinLen     = 9
+	pgStartupMaxLen     = 10000
+	pgSSLRequestCode    = 80877103
+	pgGSSENCRequestCode = 80877104
 )
 
 var (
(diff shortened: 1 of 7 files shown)


pkg/ebpf/common/tcp_detect_transform.go
```diff
@@ -177,7 +177,11 @@ func dispatchMSSQL(parseCtx *EBPFParseContext, event *TCPRequestInfo, requestBuf
 // detectGenericProtocol runs deterministic protocol detection for unclassified events:
 // SQL, FastCGI, MongoDB, Couchbase, and Memcached noreply.
 func detectGenericProtocol(parseCtx *EBPFParseContext, cfg *config.EBPFTracer, event *TCPRequestInfo, requestBuffer, responseBuffer *largebuf.LargeBuffer) (request.Span, bool, bool, error) {
-	if span, ignore, matched, err := matchSQL(cfg, event, requestBuffer, responseBuffer); matched {
+	if span, ignore, matched, err := matchPostgresStartup(parseCtx, event, requestBuffer, responseBuffer); matched {
```
NimrodAvni78: i think this should be under matchSql no?

NimrodAvni78 (approved):

grcevski (approved):
LGTM! Nice!

pkg/ebpf/common/tcp_detect_transform.go
mmat11: I think it's not important, these are ignored spans anyway and _shouldn't_ have false positives that match both handlers (unless I misunderstood and you meant inside matchSQL, for clarity, in that case I can move it in there)

pkg/ebpf/common/sql_detect_postgres.go
```diff
@@ -215,6 +223,67 @@ func postgresPreparedStatements(b *largebuf.LargeBuffer) (string, string, string
...
+		code, err := r.ReadI32BE()
+		if err != nil {
+			return "", false
+		}
+		if (code == pgSSLRequestCode || code == pgGSSENCRequestCode) && msgLen == 8 {
+			continue
+		}
+		// trailing bytes past the startup's own length = some other length-prefixed protocol
+		if code>>16 != pgStartupProtocolMajor || msgLen < pgStartupMinLen ||
+			int(msgLen)-8 < r.Remaining() {
```
MrAlias, line 244: This can turn a truncated StartupMessage into a cached `db.namespace=user` value even though the later, uncaptured params may contain an explicit different database. Would it be better to only apply the `database=user` fallback after we know we’ve captured the complete StartupMessage, and otherwise leave the namespace unset?

pkg/ebpf/common/tcp_detect_transform.go
NimrodAvni78: i mean clarity/semantically but its just a nit

pkg/ebpf/common/sql_detect_postgres.go
mmat11: done

pkg/ebpf/common/tcp_detect_transform.go
mmat11: done

PR #2625, opened 2026-07-10
fix: never log structs carrying the process environment

mmat11:
## Summary

Spans and serviceAttrs embed EnvVars which, when logged, dumps the whole environ of the process.
Log uids and counts instead, also give Attrs a value-receiver String() so nested values can't dump everything either

## Validation

### DIFF
--- a/pkg/appolly/app/svc/svc.go
+++ b/pkg/appolly/app/svc/svc.go
@@ -142,7 +142,8 @@ func (i *Attrs) GetUID() UID {
 	return i.UID
 }
 
-func (i *Attrs) String() string {
+// value receiver so nested Attrs values (e.g. Span.Service) never reflection-dump EnvVars
+func (i Attrs) String() string {
 	return i.Job()
 }
 
--- a/pkg/appolly/app/svc/svc.go
+++ b/pkg/appolly/app/svc/svc.go
@@ -4,6 +4,8 @@
 package svc // import "go.opentelemetry.io/obi/pkg/appolly/app/svc"
 
 import (
+	"log/slog"
+
 	"go.opentelemetry.io/otel/sdk/trace"
 	semconv "go.opentelemetry.io/otel/semconv/v1.41.0"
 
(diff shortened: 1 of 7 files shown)


pkg/export/otel/metrics.go
```diff
@@ -1289,7 +1289,7 @@ func (mr *MetricsReporter) onSpan(spans []request.Span) {
 		reporter, err := mr.reporters.For(&s.Service)
 		if err != nil {
 			mlog().Error("unexpected error creating OTEL resource. Ignoring metric",
-				"error", err, "service", s.Service)
+				"error", err, "service", s.Service.UID)
```
MrAlias, line 1292: This fixes this one error path, but the PR still leaves `svc.Attrs` flowing into slog elsewhere. With `OTEL_EBPF_LOG_FORMAT=json`, `slog.NewJSONHandler` does not use `String()`; it marshals exported fields, so `EnvVars` is still emitted when `newMetricsInstance`/`newMetricSet` log `service` directly, and the service graph reporter has the same pattern.

Please stop passing `svc.Attrs` to slog entirely, or give it a safe slog/JSON representation and cover the JSON handler case in the test.

pkg/appolly/app/svc/svc.go
```diff
@@ -142,7 +142,8 @@ func (i *Attrs) GetUID() UID {
 	return i.UID
 }
 
-func (i *Attrs) String() string {
+// value receiver so nested Attrs values (e.g. Span.Service) never reflection-dump EnvVars
```
MrAlias: This comment currently fails `staticcheck` (`ST1020`) because it documents the exported `String` method without starting with `String`. Both lint jobs are red on this line.

```suggestion
// String returns the service job name. The value receiver keeps nested Attrs values
// from reflection-dumping EnvVars in text formatting.
```

pkg/export/otel/metrics.go
mmat11: good catch, fixed

pkg/appolly/app/svc/svc.go
mmat11: done

NimrodAvni78 (approved):
nice!

PR #2630, opened 2026-07-10
Add CPU, memory, and goroutine Go runtime metrics

marctc:
## Summary  

This PR adds another batch of missing Go runtime metrics (scalar values only).
The following metrics are added:

  - `go.memory.used`
  - `go.memory.allocated`
  - `go.memory.allocations`
  - `go.cpu.time`
  - `go.goroutine.count`

## Validation


pkg/internal/ebpf/gotracer/gotracer.go
```diff
@@ -388,13 +430,49 @@ func (p *Tracer) recordGoChannelOffsetAvailability(fileInfo *exec.FileInfo, offs
...
+	hasOffsets := hasGoRuntimeMetricOffsets(offsets)
+	p.goRuntimeOffsetsByIno[ino] = hasOffsets
+
+	if !hasOffsets && p.log != nil {
+		p.log.Debug("skipping Go runtime metrics for binary with missing runtime offsets",
+			"pid", fileInfo.Pid(), "ino", ino, "cmd", fileInfo.CmdExePath())
+	}
+}
+
+func hasGoRuntimeMetricOffsets(offsets *goexec.Offsets) bool {
```
pinoOgni, line 452: I run some tests with regular go build binaries on Go 1.17 and 1.20 and `hasGoRuntimeMetricOffsets` requires all the new offsets before registering the runtime-metric target, so one missing offset drops every runtime metric, including the pre-existing gc.cycles / memory.limit / processor.limit / gogc.

I think we can gate registration on just the base offsets (numgc/gcPercent, present since 1.17) and let each collector's `valid_mask` bit decide per-group. In this way old Go keeps the legacy metrics and skips only the newer ones it can't compute. 
WDYT?

devdocs/runtimes/go.md
```diff
@@ -9,6 +9,11 @@ instrumented Go services and exports the following metric set.
 | --- | --- | --- | --- |
 | `go.memory.limit` | `go_memory_limit_bytes` | `runtime.gcController.memoryLimit` | Emits positive runtime memory limits; treats `math.MaxInt64` as the runtime's unset sentinel. |
 | `go.memory.gc.cycles` | `go_memory_gc_cycles_total` | `runtime.memstats.numgc` | Emits the total completed GC cycle count. |
+| `go.memory.used` | `go_memory_used_bytes` | `runtime.memstats.heapStats` and runtime sys stats | Emits `go.memory.type=stack` and `go.memory.type=other` values. Passive heap-stat reads can drift slightly from `runtime/metrics` because OBI does not rotate Go's internal heap-stat ring. |
```
pinoOgni, line 12: not related to this PR: we could add a comment like this "Go runtime metrics require the binary's symbol table; builds using `-ldflags=-s` (or strip) are not supported."

configs/offsets/tracker_input.json
```diff
@@ -28,9 +28,43 @@
...
+        "[1.23.0]GCIdleTime",
+        "[1.23.0]GCPauseTime",
+        "[1.23.0]ScavengeAssistTime",
+        "[1.23.0]ScavengeBgTime",
+        "[1.23.0]IdleTime",
+        "[1.23.0]UserTime"
+      ],
+      "runtime.schedt": ["[1.23.0]ngsys", "[1.23.0]gFree"],
+      "runtime.p": ["[1.23.0]gFree"],
+      "runtime.gList": ["[1.23.0]size"],
```
MrAlias, line 66: `runtime.gList.size` does not exist in Go 1.23 or 1.24. In those releases, `gList` contains only the head pointer, and the counts used by `gcount` are stored in `sched.gFree.n` and `pp.gFree.n` ([Go 1.23 source](https://github.com/golang/go/blob/go1.23.0/src/runtime/proc.go#L5357-L5361)). The generated offsets database therefore records `size` as offset 0 for Go 1.23/1.24, and the reads in `go_runtime_collect_goroutine_count` interpret the low 32 bits of the list head pointer as a count. This usually underflows and clamps `go.goroutine.count` to 1.

Can we mark `gList.size` as Go 1.25+ and resolve the pre-1.25 `n` fields separately, with regression coverage for Go 1.23/1.24 and 1.25+? This affects both the DWARF and prefetched-offset paths.

bpf/gotracer/go_runtime.c
```diff
@@ -45,6 +45,366 @@ typedef struct new_func_invocation {
...
+    const u64 stats_addr = target->memstats_addr + heap_stats_pos + stats_pos;
+    // heapStatsDelta ends with smallFreeCount[gc.NumSizeClasses].
+    const u64 heap_stats_delta_size =
+        small_free_count_pos + (go_runtime_size_classes * sizeof(u64));
+    s64 committed = 0;
+    s64 in_stacks = 0;
+    u64 allocated = 0;
+    u64 allocations = 0;
+
+    for (u32 slot = 0; slot < go_runtime_heap_stats_slots; slot++) {
```
MrAlias, line 215: `consistentHeapStats.stats` is a rotating ring, and summing all three slots is only stable while no runtime reader is rotating it. `runtime/metrics.Read` merges the previous slot into the current slot and then clears the previous slot ([Go runtime source](https://github.com/golang/go/blob/go1.26.5/src/runtime/mstats.go#L840-L897)). If that overlaps this loop, BPF can read the old previous slot plus the merged current slot and double-count history, or read the pre-merge current slot plus the cleared previous slot and lose history. The allocation counters can then jump or regress, and both exporters interpret a regression as a process reset.

Can we collect this at a stable runtime point or use a rotation-aware snapshot that cannot observe the merge and clear halfway through? We should also stress this path by calling `runtime/metrics.Read` concurrently with forced GC.

bpf/gotracer/go_runtime.c
```diff
@@ -45,6 +45,366 @@ typedef struct new_func_invocation {
...
+                                sched_gfree_no_stack_pos + glist_size_pos)) {
+        return;
+    }
+    if (!go_runtime_read(&allp_data, sizeof(allp_data), target->allp_addr)) {
+        return;
+    }
+    if (!go_runtime_read(&allp_len, sizeof(allp_len), target->allp_addr + k_go_slice_len_offset)) {
+        return;
+    }
+    if (allp_len > go_runtime_max_processors) {
```
MrAlias, line 375: `len(allp)` tracks `GOMAXPROCS`, which is not capped at 256. A service configured with `GOMAXPROCS=257`, or running with that effective processor count on a large host, takes this return and silently loses only `go.goroutine.count` while the other runtime metrics continue. This creates an undocumented data-loss boundary.

Can we use a collection strategy that supports the runtime's valid range, or explicitly configure and report this limitation, and add a `GOMAXPROCS=257` regression test?

pkg/internal/goexec/runtime_metrics.go
```diff
@@ -7,18 +7,42 @@ import (
...
 	"errors"
 	"fmt"
+	"regexp"
+	"strings"
 
 	"go.opentelemetry.io/obi/pkg/appolly/app"
 	"go.opentelemetry.io/obi/pkg/appolly/discover/exec"
 	"go.opentelemetry.io/obi/pkg/internal/procs"
+
+	"golang.org/x/mod/semver"
```
MrAlias, line 17: Both lint jobs are failing `gci` because this third-party import follows the project import group. Please run the formatter so `golang.org/x/mod/semver` is grouped before the `go.opentelemetry.io/obi` imports.

MrAlias (changes requested):
This PR is quite large and spans Go runtime layout and version handling, eBPF collection, event decoding, and both exporter paths. Based on this size and breadth, I'm uncertain I have effectively reviewed all of the changes here. Can this be split into smaller or stacked PRs—for example, offset and symbol resolution, BPF collection and decoding, and exporter and test changes—to enable more focused and effective review?

The inline correctness issues are blocking in the current version.

pkg/internal/ebpf/gotracer/gotracer.go
marctc: Yes, I agree. I was already digging on how to implement that but involved make this PR bigger. Would be acceptable in a follow up PR?

pinoOgni: If it is done in another PR it means that until it is merged we have runtime metrics only for new versions of go

marctc:
I opened a new PR only for cpu part and with offset validation per version. Superseed by #2665

PR #2631, opened 2026-07-10
Improve Nestjs route harvesting

NimrodAvni78:
## Summary

The Node.js route harvester could not resolve `http.route` for NestJS applications — the dominant framework shape among Node services (including NestJS running on `@nestjs/platform-fastify`). This PR fixes that in four self-contained commits.

### Problem

NestJS declares routes as decorated controller classes:

```ts
@Controller('invoice')
export class InvoiceController {
  @Get(':id')      getInvoice() {}   // GET /invoice/:id
  @Post()          ingest() {}       // POST /invoice
}
```

The harvester only captured the method decorator path (`@Get(':id')` → `:id`), so:

- the `@Controller('invoice')` mount prefix was never applied, and the harvested template could not match the observed URL path (`/invoice/123`); relative paths like `:id` were even dropped by the path cleanup, which requires a leading slash;
- bare decorators (`@Post()`), the common shape for webhook controllers, were not matched at all;
- apps shipping only compiled output (the norm for production NestJS images — tsc emits into `dist/`) harvested nothing: the walker skips `dist//build/`, skipped the scan root itself when the entrypoint is an absolute path like `/app/dist/main.js` (aborting the whole walk), and the `@`-decorator patterns cannot match compilers' lowered decorator calls (`(0, common_1.Get)(':id')` inside `__decorate` blocks);
- app-level route shaping (`setGlobalPrefix`, URI versioning, `@Controller` object/array forms) was unsupported.

As a result, requests to these services fell through to the heuristic path classifier instead of getting their real declared route templates.

### Changes (one commit per step)

1. **Join `@Controller()` prefixes with method decorator routes.** The scanner tracks the prefix of the most recently seen `@Controller()` while scanning a file and joins it with each method decorator path into a complete absolute route. TypeScript decorators always live in the same file as the class they decorate, with `@Controller()` preceding the method decorators, so per-file tracking is sufficient (multiple controllers per file are handled).

2. **Harvest bare method decorators.** `@Post()` with no argument now resolves to the controller prefix itself (`@Controller('callbacks/acme')` + `@Post()` → `/callbacks/acme`).

3. **Scan compiled output (`dist/`) into partial routes.** When the source scan finds no framework-declared routes, the application directory is rescanned in a compiled mode that descends into `dist//build/` (and never skips the scan root), recognizes lowered decorator calls (tsc `(0, common_1.Get)('store')`, swc `(0, _common.Get)(':id')`, direct `common_1.Get('store')`), and disables the fallback pattern (compiled code is full of path-like string literals). Compiled output scatters the controller/method association across `__decorate` blocks, so compiled matches are harvested as standalone route fragments and returned as `PartialRoutes`, reusing the partial matcher introduced for the Java harvester. Fragments are ordered literals-first / most-specific-first so partial matching is deterministic and specific fragments win over catch-alls.

4. **NestJS global prefix, URI versioning, and `@Controller` object/array forms.** `app.setGlobalPrefix('api')` is prepended to every Nest route (a standalone fragment in compiled mode); `app.enableVersioning()` URI versioning (NestJS's default type) inserts `/v<version>/` from `@Version()` on the method, the `version` option of `@Controller({...})`, or the configured `defaultVersion`, with `@Version(VERSION_NEUTRAL)` suppressing it; `@Controller({ path: 'catalog', version: '2' })` and `@Controller(['ledger', 'books'])` are supported, plus their compiled equivalents.

### Example

For a NestJS-on-Fastify app, harvested routes go from effectively nothing to:

| Declaration | Harvested route |
|---|---|
| `@Controller('invoice')` + `@Get(':id')` | `/invoice/:id` |
| `@Controller('callbacks/acme')` + `@Post()` | `/callbacks/acme` |
| dist-only image, compiled `(0, common_1.Get)(':id')` | fragments matching `/invoice/123` → `/invoice/:id` (partial) |
| `setGlobalPrefix('api')` + `@Controller({path:'catalog', version:'2'})` + `@Get('featured')` | `/api/v2/catalog/featured` |

### Known limitations (deliberately out of scope)

- `RouterModule.register()` per-module prefixes (requires statically resolving the module graph).
- The `exclude` option of `setGlobalPrefix()`.
- Minified/bundled single-file output (webpack/esbuild) — fragment recognition depends on the decorator name surviving next to a string literal.

## Validation

- Unit tests cover every step with fixture apps mirroring real NestJS-on-Fastify layouts: source controllers (prefixed, prefix-less, bare-decorator webhook), a tsc-compiled `dist/`-only app (realistic `__decorate` output), and a versioned app (`setGlobalPrefix` + `enableVersioning` + object/array controller forms). End-to-end `ExtractNodejsRoutes` tests exercise both entrypoint shapes (`node dist/main.js` and absolute `/dist/main.js`) and assert partial matching resolves `/invoice/8f31ac/receipt` → `/invoice/:id/receipt`.
- `go build ./pkg/...`, `go vet` and the test suites of the route packages and all dependents (`pkg/transform`, `pkg/appolly/...`, `pkg/internal/nodejs`) pass.
- A `.gitignore` carve-out is included for the test fixture that must literally live in a directory named `dist/` (the scan behavior under test keys on that name).


pkg/internal/transform/route/harvest/js.go
```diff
@@ -233,23 +363,323 @@ func (e *RouteExtractor) handleRestify(filePath, line string, lineNum int) bool
...
+		e.setNestController(nil, nil)
+		return true
+	}
+	return false
+}
+
+// handleNestJSVersion tracks a @Version() decorator; it applies to the next
+// method decorator in the file. Non-string arguments (VERSION_NEUTRAL) declare
+// the route version-neutral.
+func (e *RouteExtractor) handleNestJSVersion(line string) bool {
```
grcevski, line 464: Can there be other non-string versions like `@Version(API_V2)`?

grcevski (approved):
Looks great! I left one comment, but I'm not sure we can do anything about it.

mmat11 (approved):

pkg/internal/transform/route/harvest/js.go
NimrodAvni78: this might be possible but we need to keep track of variable references and resolve them at extraction time
i have some POC of this but i preferer to implement it in a next phase

PR #2633, opened 2026-07-12
Fix / Document invalid enum attribute values and enforce weaver `undefined_enum_variant`

### LINKED ISSUE #2572: Weaver validation should alert violations on `undefined_enum_variant`

NimrodAvni78:
## Summary

Fixes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2572

Weaver `registry live-check` reported `undefined_enum_variant` findings caused by real
emitter bugs: enumerated semconv attributes were emitted with values outside their
declared members — an HTTP status leaking into the gRPC status attribute, and empty
strings for `connection_type`, `error.type`, and `messaging.operation.type`.

This PR fixes each emitter (omitting the attribute when there is no valid value,
never inventing one) and then flips enforcement: `undefined_enum_variant` is now an
actionable advice type in `weavercheck`, so any regression fails the suites.

### 1. gRPC status code carrying an HTTP status (`rpc.response.status_code`)

Before the semconv v1.41 migration OBI emitted `rpc.grpc.status_code` as a raw
`span.Status` int, so a leaked HTTP status surfaced as `rpc.grpc.status_code=200`.
After the migration the same leak produced fabricated `rpc.response.status_code`
values: `GRPCStatusCodeString` returned `"CODE(200)"` (or `"CODE(65535)"` from the
eBPF client-side `u16(-1)` status-read fallback) for anything outside the gRPC
enum (0–16).

- `pkg/appolly/app/request/grpc_status.go` — `GRPCStatusCodeString` now returns `""`
  for statuses outside 0–16 instead of `"CODE(n)"`.
- `pkg/appolly/app/request/span_getters.go` (`attr.RPCResponseStatusCode` getter) and
  `pkg/export/otel/tracesgen/tracesgen.go` (gRPC server + client span attributes) —
  the attribute is omitted when `GRPCStatusCodeString` returns `""`.

### 2. `connection_type` emitted as an empty string on service-graph metrics

`ConnectionTypeForSpan` returns `""` for direct HTTP/gRPC requests, but the
service-graph recorder attached the attribute unconditionally, emitting an empty
(invalid) enum variant on every direct-request metric.

- `pkg/export/otel/metrics_svc_graph.go` — the attribute is only attached when the
  connection type is non-empty.
- `schemas/obi/groups/service_graph.yaml` — brief updated: the attribute is omitted
  for direct requests (previously documented as "empty string represents a direct
  request").

### 3. `error.type` emitted as an empty string on successful requests

- `pkg/appolly/app/request/span_getters.go` (`attr.ErrorType` getter) — returned
  `error.type=""` for every successful request on metrics; now omits the attribute.
- `pkg/export/otel/tracesgen/tracesgen.go` — Elasticsearch spans attached
  `error.type` unguarded (empty on success), and SQL error spans attached it even
  when the SQLSTATE was not captured; both are now guarded.

### 4. `messaging.operation.type` emitted as an empty string

- `pkg/appolly/app/request/span_getters.go` (`attr.MessagingOpType` getter) —
  defaulted to `""` for non-messaging spans / unknown operations; now omits.
- `pkg/export/otel/tracesgen/tracesgen.go` — Kafka/MQTT/NATS/AMQP spans attached the
  attribute even when `span.Method` was empty, and AWS SQS spans when the operation
  type was unrecognized; all sites now omit instead.

### 5. `gen_ai.operation.name` emitted as an empty string (found by CI after the flip)

The first weaver-enforced CI run flagged `gen_ai.operation.name=""` on OpenAI
error spans (`span:POST /v1/responses`) and on the gen_ai metrics.

- `pkg/appolly/app/request/span_getters.go` (`attr.GenAIOperationName`,
  `attr.GenAIProviderName` getters) — omit when the span carries no value.
- `pkg/export/otel/tracesgen/tracesgen.go` — OpenAI, Anthropic, and Qwen span
  attribute sites now omit `gen_ai.operation.name` when the operation was not
  classified.

### 6. HTTP/2 connection preface reported as a request span (`PRI *`)

Go's h2c upgrade path lets `net/http` parse the HTTP/2 client connection
preface (`PRI * HTTP/2.0`, RFC 9113 §3.4) as a literal request before
hijacking the connection, so the Go tracer uprobes surfaced a bogus
`http.request.method="PRI"` span (`span:PRI *`) on every h2c connection
(flagged by weaver in the gRPC-mux suite).

- `pkg/ebpf/common/common.go` — `isH2CPrefacePseudoRequest` drops these
  pseudo-requests at span-conversion time; the real HTTP/2 exchanges on the
  connection are traced separately by the http2 tracer.

### 7. `gen_ai.operation.name` missing on failed OpenAI calls

The omission fix (5) traded an empty enum value for a missing-required-
attribute violation: OpenAI error responses carry no `object` field, so
spans/metrics for failed `/v1/responses` calls had no operation name at all —
but `gen_ai.operation.name` is `required` on the gen_ai client metrics.

- `pkg/ebpf/common/http/openai.go` — the operation name is now derived from
  the request path (authoritative even on errors): `/v1/responses` →
  `response`, `/v1/conversations` → `conversation` (matching the existing
  `/v1/chat/completions` / `/v1/embeddings` handling). Failed calls now get
  the semconv span name (`response gpt-5-mini`) and fully-attributed metrics.
- `internal/test/oats/ai/yaml/oats_open_ai_test.yaml` — error-span assertion
  updated accordingly.

### Registry overrides (values OBI emits on purpose)

CI flagged values OBI emits intentionally but that upstream semconv v1.41
cannot validate. These are declared as attribute overrides in
`schemas/obi/groups/`, in two styles (documented once in
`schemas/obi/README.md`):

**Closed enums, extended** (upstream members duplicated verbatim + OBI
extensions — weaver has no enum-member merge, a local re-declaration replaces
the upstream definition in live-check's lookup):

- `messaging.yaml` — `messaging.system` + `amqp`, `mqtt`, `nats` (OBI detects
  the wire protocol, not the broker product).
- `gen_ai.yaml` — `gen_ai.provider.name` + `qwen`, `voyage`, `jina`,
  `pinecone`, `qdrant`, `milvus`, `zilliz`, `chroma`, `weaviate`, `generic`
  (mirrors the provider-detection tables in `pkg/ebpf/common/http/`).
- `openai.yaml` — `openai.api.type` + `embeddings`.
- `telemetry.yaml` — `telemetry.sdk.language` + `generic` (processes with no
  recognized language runtime; the `unknown` bug-marker fallback is
  deliberately NOT declared so weaver keeps flagging it).
- `db.yaml` (new) — `db.system.name` + `aerospike` (OBI's Aerospike client
  instrumentation; `unknown` again deliberately not declared).
- `rpc.yaml` (new) — `rpc.system.name` + `aws-api` (AWS SDK calls detected on
  the wire) and `onc_rpc` (SunRPC instrumentation).

**Open-ended value spaces, re-typed as string** (for values which are declared as enums by the otel semconv but are explicitly not enums):

- `error.yaml` (new) — `error.type`: upstream declares only `_OTHER` and its
  examples are arbitrary domain strings; observed CI values include
  `insufficient_quota`, `NXDomain`, `ValidationException`, `error`.
- `gen_ai.yaml` — `gen_ai.operation.name`: OBI emits provider-specific
  operation vocabulary (`invoke_model`, `message`, `conversation`,
  `response`, `rerank`, ...) and passes MCP JSON-RPC method names through
  verbatim.

For these two, the "never emit an empty value" guarantee moves entirely to
the emitter unit tests (weaver cannot flag an empty string on a string-typed
attribute).

**Test-config fix (no override)**: the Rails suites set
`deployment.environment.name=to` in their own `OTEL_RESOURCE_ATTRIBUTES`
(passthrough verification with an arbitrary value), which weaver flagged
against upstream's enum (production/staging/test/development). OBI is a pure
passthrough here — verified `ParseOTELResourceVariable` does not truncate —
so the tests now use the valid member `staging` instead of declaring a
registry override for a user-configured attribute
(`internal/test/integration/docker-compose-ruby*.yml`,
`red_test_ruby.go`).

Group ids use a `x.obi.<ns>` prefix: weaver v0.24.1 resolves duplicate
attribute names last-insert-wins over id-sorted groups, and upstream `span.*`
groups that `ref` these attributes sort after `registry.*` (verified
empirically with `weaver registry live-check`: OBI extensions, re-typed
values, and all upstream members accepted; bogus values and `unknown` still
flagged; also verified the bug is unfixed on weaver `main` as of 2026-07 —
see weaver#1578). The resulting `DuplicateAttributeId` lint findings are
allowlisted (tightly, per attribute + group pair) in
`scripts/lint-schema-filter.jq`, with cases added to
`scripts/lint_schema_filter_test.go`. The full override mechanics are
documented once in `schemas/obi/README.md`; the per-file comments only state
WHAT is overridden and why.

### Omission plumbing

Getters signal omission by returning an invalid (zero) `attribute.KeyValue`, which
the OTLP metrics `Expirer` already skips. `spanPromGetters` now maps omission to
`""` explicitly — Prometheus label sets are fixed, and `Value.Emit()` on an invalid
value would otherwise have produced the string `"unknown"`.

### Enforcement flip

- `internal/test/weavercheck/weavercheck.go` — `undefined_enum_variant` added
  to `actionableAdviceTypes` (as a named const, per review) with a doc comment
  in the existing style. Weaver classifies these findings as
  `information`-level (enums are open), but every occurrence seen so far has
  been an emitter bug. There is no in-code exemption list: values OBI emits on
  purpose are declared in the registry overrides, so anything weaver still
  flags is a bug by definition.

## Testing

- New/updated unit tests:
  - `grpc_status_test.go`: out-of-range statuses (99, 200, 0xFFFF, −1) map to `""`.
  - `span_getters_test.go`: omission cases for the RPC status getter (leaked HTTP
    200, failed status read, non-RPC spans), plus new tests for `error.type` and
    `messaging.operation.type` omission vs. genuine error/operation values.
  - `traces_test.go`: gRPC spans with valid status keep `rpc.response.status_code`;
    leaked/invalid statuses omit it; Kafka spans with unknown operation omit
    `messaging.operation.type`; successful Elasticsearch spans omit `error.type`.
  - `metrics_svc_graph_test.go`: direct-HTTP service-graph metrics must not carry a
    `connection_type` attribute at all.
  - `weavercheck_test.go`: `undefined_enum_variant` counts as actionable
    (including empty enum values); other information-level advice types still
    do not.
  - `common_test.go` (`pkg/ebpf/common`): the h2c preface pseudo-request is
    recognized and dropped; real requests are not.
  - `openai_test.go`: failed `/v1/responses` calls carry the path-derived
    `response` operation name and `responses` API type.
- `CGO_ENABLED=0 GOOS=linux go build ./...` passes.
- `go test` (linux, docker `golang:1.25`) passes for `internal/test/weavercheck`,
  `pkg/appolly/app/request`, `pkg/ebpf/common/http`, `pkg/export/otel`,
  `pkg/export/otel/tracesgen`, `pkg/export/prom`, `pkg/export/attributes`,
  `scripts`; the full `./pkg/...` sweep has only pre-existing
  environment-dependent failures (kernel BTF, lockdown, kubebuilder etcd).
- `make lint-schema` passes with the extended allowlist.
- `go vet -tags=integration ./internal/test/integration/` passes (covers the
  `red_test_ruby.go` assertion change; the Rails suites themselves run in CI).
- `weaver registry live-check` (v0.24.1, file input) against the updated
  registry accepts every intentional value (`nats`, `qwen`, `aerospike`,
  `aws-api`, `onc_rpc`, `deployment.environment.name=staging`,
  `error.type=NXDomain/insufficient_quota`, `gen_ai.operation.name=tools/call`)
  while still flagging `db.system.name=unknown`, `telemetry.sdk.language=unknown`,
  `deployment.environment.name=to`, and bogus enum values.

## Notes / follow-ups for reviewers

- **Detection layer left unchanged**: the kprobes HTTP/2 path still passes through
  `span.Status` values 17–99, and the eBPF gRPC client fallback still records
  `u16(-1)`; these are now omitted at emission rather than clamped at detection.
  Clamping (e.g. to `UNKNOWN`) would be a product decision.
- **`publish` vs `send`**: OBI emits `messaging.operation.type="publish"` for
  producer spans. Verified empirically: semconv v1.41 still declares `publish`
  (and `deliver`) as deprecated enum members and weaver accepts deprecated
  members, so no action is needed. Renaming to `send` remains a
  compat-sensitive decision out of scope here.
- **Semconv bump duty**: the enum-override files under `schemas/obi/groups/`
  duplicate upstream member lists; when bumping the semconv dependency they
  must be re-synced (drift resurfaces as `undefined_enum_variant` failures in
  the weaver-validated suites — caught, not silent). Documented in
  `schemas/obi/README.md`.
- **`error.type="error"` generic fallback**: the metrics `attr.ErrorType`
  getter reports the literal `error` for failed requests with no specific
  code; semconv's designated fallback member is `_OTHER`. Changing the
  emitted value is a compat-sensitive product decision, left as-is (validates
  cleanly now that `error.type` is re-typed as a string in the registry).
- **Prometheus parity**: prom exporters keep the fixed `connection_type` /
  `rpc_response_status_code` labels with `""` when the OTLP attribute is omitted
  (empty label ≡ absent in the Prometheus data model; weaver validates OTLP only).

## Validation

### DIFF
--- a/pkg/appolly/app/request/span_getters.go
+++ b/pkg/appolly/app/request/span_getters.go
@@ -221,7 +221,9 @@ func spanOTELGetters(name attr.Name) (attributes.Getter[*Span, attribute.KeyValu
 				}
 				return ErrorType("error")
 			}
-			return ErrorType("")
+			// error.type only applies to failed requests: return an invalid
+			// KeyValue so the attribute is omitted instead of emitted empty.
+			return attribute.KeyValue{}
 		}
 	case attr.MessagingSystem:
 		getter = func(span *Span) attribute.KeyValue {
@@ -272,17 +274,23 @@ func spanOTELGetters(name attr.Name) (attributes.Getter[*Span, attribute.KeyValu
 		}
 	case attr.MessagingOpType:
 		getter = func(span *Span) attribute.KeyValue {
+			opType := ""
 			switch span.Type {
 			case EventTypeKafkaClient, EventTypeKafkaServer,
 				EventTypeMQTTClient, EventTypeMQTTServer,
 				EventTypeNATSClient, EventTypeNATSServer,
 				EventTypeAMQPClient:
-				return MessagingOperationType(span.Method)
+				opType = span.Method
 			}
 			if span.Type == EventTypeHTTPClient && span.SubType == HTTPSubtypeAWSSQS && span.AWS != nil {
-				return MessagingOperationType(span.AWS.SQS.OperationType)
+				opType = span.AWS.SQS.OperationType
 			}
-			return MessagingOperationType("")
+			if opType == "" {
+				// messaging.operation.type is a semconv enum: omit the
+				// attribute rather than emitting an empty (invalid) variant.
+				return attribute.KeyValue{}
+			}
+			return MessagingOperationType(opType)
 		}
 	case attr.MessagingMessageID:
 		getter = func(span *Span) attribute.KeyValue {
@@ -487,11 +495,22 @@ func spanOTELGetters(name attr.Name) (attributes.Getter[*Span, attribute.KeyValu
 		}
 	case attr.GenAIOperationName:
 		getter = func(s *Span) attribute.KeyValue {
-			return semconv.GenAIOperationNameKey.String(s.GenAIOperationName())
+			if op := s.GenAIOperationName(); op != "" {
+				return semconv.GenAIOperationNameKey.String(op)
+			}
+			// Omit gen_ai.operation.name rather than emitting an empty value
+			// (required on the gen_ai client metrics when present, and an
+			// empty string carries no information).
+			return attribute.KeyValue{}
 		}
 	case attr.GenAIProviderName:
 		getter = func(s *Span) attribute.KeyValue {
-			return semconv.GenAIProviderNameKey.String(s.GenAIProviderName())
+			if provider := s.GenAIProviderName(); provider != "" {
+				return semconv.GenAIProviderNameKey.String(provider)
+			}
+			// gen_ai.provider.name is a semconv enum: omit the attribute
+			// rather than emitting an empty (invalid) variant.
+			return attribute.KeyValue{}
 		}
 	case attr.GenAITokenTypeInput:
 		getter = func(_ *Span) attribute.KeyValue {
@@ -529,12 +548,19 @@ func spanOTELGetters(name attr.Name) (attributes.Getter[*Span, attribute.KeyValu
 				return semconv.RPCResponseStatusCode(SunRPCResponseStatusCode(s.Status))
 			}
 			if s.Type == EventTypeGRPC || s.Type == EventTypeGRPCClient {
-				return semconv.RPCResponseStatusCode(GRPCStatusCodeString(s.Status))
+				// GRPCStatusCodeString returns "" for statuses outside the
+				// gRPC enum (e.g. an HTTP code that leaked through protocol
+				// detection): omit the attribute instead of inventing a value.
+				if code := GRPCStatusCodeString(s.Status); code != "" {
+					return semconv.RPCResponseStatusCode(code)
+				}
+				return attribute.KeyValue{}
 			}
 			if s.SubType == HTTPSubtypeJSONRPC && s.JSONRPC != nil && s.JSONRPC.ErrorCode != 0 {
 				return semconv.RPCResponseStatusCode(strconv.Itoa(s.JSONRPC.ErrorCode))
 			}
-			return semconv.RPCResponseStatusCode("")
+			// No status to report: omit the attribute instead of emitting "".
+			return attribute.KeyValue{}
 		}
 	}
 	// default: unlike the Prometheus getters, we don't check here for service name nor k8s metadata
@@ -551,7 +577,16 @@ func spanPromGetters(attrName attr.Name) attributes.Getter[*Span, string] {
 		return func(s *Span) string { return s.SunRPCProcedureRouteForExport() }
 	}
 	if otelGetter, ok := spanOTELGetters(attrName); ok {
-		return func(span *Span) string { return otelGetter(span).Value.Emit() }
+		return func(span *Span) string {
+			// OTEL getters return an invalid (zero) KeyValue to signal
+			// "omit this attribute". Prometheus label sets are fixed, so
+			// map omission to the empty string rather than Value.Emit()'s
+			// "unknown" placeholder for invalid values.
+			if kv := otelGetter(span); kv.Valid() {
+				return kv.Value.Emit()
+			}
+			return ""
+		}
 	}
 	// unlike the OTEL getters, when the attribute is not found, we need to look for it
 	// in the metadata section
--- a/pkg/export/otel/tracesgen/tracesgen.go
+++ b/pkg/export/otel/tracesgen/tracesgen.go
@@ -564,10 +564,15 @@ func traceAttributesSelectorInternal(span *request.Span, optionalAttrs map[attr.
 		attrs = []attribute.KeyValue{
 			semconv.RPCMethod(span.Path),
 			semconv.RPCSystemNameGRPC,
-			semconv.RPCResponseStatusCode(request.GRPCStatusCodeString(span.Status)),
 			request.ClientAddr(request.PeerAsClient(span)),
 			request.ServerAddr(request.SpanHost(span)),
 			request.ServerPort(span.HostPort),
+		}
+		// GRPCStatusCodeString returns "" for statuses outside the gRPC enum
+		// (e.g. an HTTP code that leaked through protocol detection): omit
+		// the attribute instead of inventing a value.
+		if code := request.GRPCStatusCodeString(span.Status); code != "" {
+			attrs = append(attrs, semconv.RPCResponseStatusCode(code))
 		}
 	case request.EventTypeHTTPClient:
 		// SQL++ spans should only have DB attributes, not HTTP attributes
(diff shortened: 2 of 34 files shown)


pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -666,7 +675,11 @@ func traceAttributesSelectorInternal(span *request.Span, optionalAttrs map[attr.
 		if span.SubType == request.HTTPSubtypeAWSSQS && span.AWS != nil {
 			sqs := span.AWS.SQS
 			attrs = append(attrs, request.MessagingOperationName(sqs.OperationName))
-			attrs = append(attrs, request.MessagingOperationType(sqs.OperationType))
+			// messaging.operation.type is a semconv enum: omit it instead of
+			// emitting an empty (invalid) variant when the type is unknown.
+			if sqs.OperationType != "" {
```
MrAlias, line 680: The integration test still expects this tag to be present when `expectedOperationType` is empty, so this change leaves shard 9 red. Can we update `assertSQSOperation` to check that the tag is absent in that case?

pkg/appolly/app/request/span_getters.go
```diff
@@ -487,11 +495,22 @@ func spanOTELGetters(name attr.Name) (attributes.Getter[*Span, attribute.KeyValu
...
+			}
+			// Omit gen_ai.operation.name rather than emitting an empty value
+			// (required on the gen_ai client metrics when present, and an
+			// empty string carries no information).
+			return attribute.KeyValue{}
 		}
 	case attr.GenAIProviderName:
 		getter = func(s *Span) attribute.KeyValue {
-			return semconv.GenAIProviderNameKey.String(s.GenAIProviderName())
+			if provider := s.GenAIProviderName(); provider != "" {
```
MrAlias, line 508: `RerankSpan` still returns `"unknown"` for custom `/rerank` endpoints. That value is not in the new provider enum, and the trace exporter writes it directly too, so live-check will still fail for those spans.

Can we either use a declared fallback or omit the provider from both metrics and traces, and add a test for this case?

MrAlias (changes requested):

NimrodAvni78:
@MrAlias thanks for the review
fixed!

MrAlias (approved):

Issue #2638, opened 2026-07-13
Retained ELF string table causes 100+ MiB heap growth per instrumented binary


bourbonkk:
> **⚠️ UPDATE 2026-07-13 — root cause re-diagnosed, see [correction comment below](#issuecomment-4955203264)**
>
> The original report below framed the issue as retention on Rust/C++/Envoy binaries, and attributed it to `.strtab` substring semantics. Both framings are now known to be **inaccurate**:
>
> - **Substring theory is wrong**: Go stdlib `debug/elf.getString()` returns via `string([]byte)` which always allocates a new backing array. The original `.strtab` slice is GC-eligible; substrings don't hold it.
> - **The heap the profiles captured is a mix of two independent causes**:
>   1. **Retained** heap on **Go** binaries — due to a missing `probe.Skip = true` in `pkg/ebpf/instrumenter.go:gatherGoOffsets`, which causes `link.Executable.Uprobe` to be called with `Address=0`, forcing cilium's `lazyLoadSymbols()` to parse the full ELF and retain it on `Executable.cachedSymbols` for the tracer's lifetime.
>   2. **Transient** parse cost when generic tracer's `resolveInstrPath` falls back to the executable itself (Rust with debug info, Envoy static binary). This is GC-eligible after `gatherOffsets` returns, but pprof snapshots taken before the next GC show it in inuse-space.
>
> The two clusters in the original report both had Go services co-located (`flagd-build`, `product-catalog` in Cluster A; `flagd` sidecars in Cluster B), so the retained portion of the observed heap was very likely from those, not from the Rust/Envoy processes named in the headline.
>
> **PR #2642 fixes cause (1)** — retention on Go binaries. Cause (2) (transient parse cost on native-path exe fallback) remains open; it may warrant a separate issue.
>
> The rest of this body is preserved for history but its root-cause section should be read alongside the correction comment.

---

## Summary

When instrumenting large binaries (Rust, C++/Envoy, Erlang), OBI holds **100+ MiB of parsed ELF string table per unique instrumented binary** in its resident heap. The retention persists after uprobe attach completes, and is not reclaimed by GC until all processes referencing that binary exit.

[...]

PR #2642, opened 2026-07-13
fix(ebpf): mark unresolved Go probes as Skip to avoid full ELF symbol parse

bourbonkk:
## Summary

Fixes a heap retention where OBI holds 100+ MiB of parsed ELF symbols per instrumented Go binary. Refs #2638.

`gatherGoOffsets` did not set `probe.Skip = true` when a Go probe symbol was missing from the resolved offsets. `instrumentProbes` then attached the probe with `probe.StartOffset == 0`, which reaches cilium/ebpf as `UprobeOptions{Address: 0}`. cilium's `Executable.address()` only short-circuits when `Address > 0`, so `Address == 0` forces `lazyLoadSymbols()` to parse the full ELF symbol table into `Executable.cachedSymbols`. That map is retained for the lifetime of the `*Executable`, which is held by `instrumenter.exe` on the tracer stored in `ProcessTracer.Instrumentables[ino]` — i.e. as long as any process for the binary is alive.

The bug reproduces reliably because gotracer's `GoProbes()` registers ~60 probes across `net/http`, `net/rpc/jsonrpc`, `github.com/lib/pq`, `golang.org/x/net/http2`, `crypto/tls`, `database/sql`, etc. Most Go services don't use every one of these, so at least one symbol lookup fails on every Go binary.

The native uprobe path (`gatherOffsetsImpl` at line 900-916) already sets `probe.Skip = true` on unresolved symbols; this change brings the Go path into symmetry.

**Retention path** (before this fix):

```
attacherLoop
 └─ traceAttacher.getTracer
    └─ ProcessTracer.NewExecutable
       └─ i.goprobes(p)
          └─ i.gatherGoOffsets(goProbes)          // probe.Skip stays false on missed lookup
          └─ i.instrumentProbes(exe, goProbes)
             └─ i.uprobe(exe, probe)               // probe.Skip == false → attach attempted
                └─ exe.Uprobe("", prog, &UprobeOptions{Address: 0})   // ← the bug
                   └─ cilium address() → lazyLoadSymbols()
                      └─ debug/elf (*File).Symbols + DynamicSymbols
                         └─ Executable.cachedSymbols map[string]symbol   // retained
```

**Fix:** mark probes as `Skip` when their offset is not present in `i.offsets.Funcs`, matching the native path's behavior. Also set `Skip = false` on the resolved branch to guard against stale state across reuse (`NewExecutableInstance`).

## Validation

- [x] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed. (N/A — no user-facing feature change)

## Related

- Refs #2638 (detailed root cause analysis in a follow-up comment on the issue)
- The `gatherOffsetsImpl` symmetry it aligns to was introduced at line 900-916 of `pkg/ebpf/instrumenter.go`

### DIFF
--- a/pkg/ebpf/instrumenter.go
+++ b/pkg/ebpf/instrumenter.go
@@ -991,10 +991,16 @@ func (i *instrumenter) gatherGoOffsets(goProbes map[string][]*ebpfcommon.ProbeDe
 			// probes as Skip so instrumentProbes does not attempt to attach
 			// them with Address=0, which would force cilium/ebpf to parse the
 			// full ELF symbol table via lazyLoadSymbols and retain it on
-			// Executable.cachedSymbols for the tracer's lifetime.
+			// Executable.cachedSymbols for the tracer's lifetime. Emit an
+			// InstrumentationError with a symbol_not_found label to preserve
+			// observability of missing symbols where they might've been
+			// expected.
 			log.Debug("ignoring function", "function", symbolName)
 			for _, probe := range descs {
 				probe.Skip = true
+			}
+			if i.metrics != nil {
+				i.metrics.InstrumentationError(i.processName, imetrics.InstrumentationErrorSymbolNotFound)
 			}
 			continue
 		}
--- a/pkg/ebpf/instrumenter.go
+++ b/pkg/ebpf/instrumenter.go
@@ -987,12 +987,20 @@ func (i *instrumenter) gatherGoOffsets(goProbes map[string][]*ebpfcommon.ProbeDe
 		offs, ok := i.offsets.Funcs[symbolName]
 
 		if !ok {
-			// the program function is not in the detected offsets. Ignoring
+			// The program function is not in the detected offsets. Mark the
+			// probes as Skip so instrumentProbes does not attempt to attach
+			// them with Address=0, which would force cilium/ebpf to parse the
+			// full ELF symbol table via lazyLoadSymbols and retain it on
+			// Executable.cachedSymbols for the tracer's lifetime.
 			log.Debug("ignoring function", "function", symbolName)
+			for _, probe := range descs {
+				probe.Skip = true
+			}
 			continue
 		}
 
 		for _, probe := range descs {
+			probe.Skip = false
 			probe.StartOffset = offs.Start
 			probe.ReturnOffsets = offs.Returns
 		}
--- a/pkg/ebpf/instrumenter_test.go
+++ b/pkg/ebpf/instrumenter_test.go
@@ -196,6 +196,54 @@ func TestGatherOffsetsFailsMissingRequiredSymbol(t *testing.T) {
 	assert.Empty(t, desc.ReturnOffsets)
 }
 
+func TestGatherGoOffsetsMarksMissingSymbolAsSkip(t *testing.T) {
+	// Regression for the retention bug fixed alongside this test: when
+	// gatherGoOffsets does not find an offset for a probe symbol it must
+	// mark the probe as Skip. Otherwise instrumentProbes attaches with
+	// probe.StartOffset == 0, which reaches cilium/ebpf as
+	// UprobeOptions{Address: 0} and forces a full ELF symbol table parse
+	// retained on Executable.cachedSymbols for the tracer's lifetime.
+	i := &instrumenter{
+		offsets: &goexec.Offsets{
+			Funcs: map[string]goexec.FuncOffsets{},
+		},
+	}
+	probes := probeDescMap{
+		"net/rpc/jsonrpc.(*serverCodec).ReadRequestHeader": {{}},
+	}
+
+	i.gatherGoOffsets(probes)
+
+	desc := probes["net/rpc/jsonrpc.(*serverCodec).ReadRequestHeader"][0]
+	assert.True(t, desc.Skip)
+	assert.Zero(t, desc.StartOffset)
+	assert.Empty(t, desc.ReturnOffsets)
+}
+
+func TestGatherGoOffsetsAppliesResolvedOffsetsAndClearsSkip(t *testing.T) {
+	i := &instrumenter{
+		offsets: &goexec.Offsets{
+			Funcs: map[string]goexec.FuncOffsets{
+				"net/http.serverHandler.ServeHTTP": {
+					Start:   0x1234,
+					Returns: []uint64{0x1250, 0x1260},
+				},
+			},
+		},
+	}
+	// Seed Skip = true to ensure the resolved branch clears stale state on reuse.
+	probes := probeDescMap{
+		"net/http.serverHandler.ServeHTTP": {{Skip: true}},
+	}
+
+	i.gatherGoOffsets(probes)
+
+	desc := probes["net/http.serverHandler.ServeHTTP"][0]
+	assert.False(t, desc.Skip)
+	assert.Equal(t, uint64(0x1234), desc.StartOffset)
+	assert.Equal(t, []uint64{0x1250, 0x1260}, desc.ReturnOffsets)
+}
+
 func TestInstrumentProbesSkipsMarkedOptionalProbe(t *testing.T) {
 	i := &instrumenter{}
 	probes := probeDescMap{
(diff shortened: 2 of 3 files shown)


bourbonkk:
# Field validation of this fix's impact

Just ran a validation experiment on our production-like otel-demo cluster (Kubernetes DaemonSet, 7 nodes). Results support this PR having significant real-world impact.

Full write-up in [#2638 (comment)](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2638#issuecomment-4955786952), summary here:

| Node | Instrumented mix | Before | After Go excluded via `discovery.exclude_instrument` |
|---|---|---|---|
| `worker-bvnh` | Rust + Erlang + **Go×3** | 350 MiB | **218 MiB** (−132) |
| `worker-x644` | Go×2 + JVM×2 + Node + Python | ~300+ MiB | **219 MiB** (~−80) |

Nodes with **Rust `shipping` but no Go apps** were already at baseline (~250 MiB), confirming the retained heap is a function of instrumented Go binaries — not Rust/C++/Envoy as originally framed in #2638.

The experiment used `exclude_instrument` as a proxy for the fix (since the goal was to identify the source, not to test the diff). Applying this PR's actual change should produce equivalent savings without needing to exclude the Go apps from instrumentation. Ready to re-run with the patched image if a preview build is available.

mariomac (approved):
Clever!! Thank you for your contribution

grcevski (approved):
LGTM!

MrAlias (approved):

bourbonkk:
Thanks for the approvals @mariomac @grcevski @MrAlias 🙏

Heads-up: the Integration tests job is failing consistently on this PR — but only `TestInstrumentationErrors/Instrumentation_error_metrics` (shard-3). All other CI jobs (Unit, K8S, ARM, OATS, Lint, BPF verifier, PR checks, Build) pass. The same test passed on main just before this PR (5s runtime on main → 60s timeout here).

## What's happening

The test expects `obi_instrumentation_errors_total >= 1`, driven by a docker-compose scenario that runs OBI against a Go testserver with restricted capabilities (`cap_drop: ALL` + `SYS_ADMIN, SYS_PTRACE, DAC_READ_SEARCH, PERFMON`). With these capabilities, eBPF program load and kprobes succeed, so the only error path that consistently populated the counter on main was:

1. `gatherGoOffsets` couldn't find an offset for a Go probe symbol
2. `probe.Skip` was left false, so `instrumentProbes` called `Uprobe(Address=0)`
3. cilium's `lazyLoadSymbols` triggered → attach eventually failed → `InstrumentationError(attaching_uprobe)` was recorded

This PR eliminates step 2 (that was the whole point — it's what made cilium retain 100+ MiB of parsed symbols per Go tracer). But it also silently removes the counter increment the test relied on. In practice on main, this counter was going up on **every** Go binary OBI attached to that didn't use every one of gotracer's ~60 registered symbols — which is essentially always. So the test was passing by observing a symptom of the very bug this PR fixes.

## Options I see (would appreciate your input)

**(A)** Still emit `InstrumentationError` on symbol miss in `gatherGoOffsets`, with a new `error_type` (e.g. `symbol_not_found` or reusing `attaching_uprobe`). Preserves observability of missing symbols and keeps this test's assertion valid. Minimal test churn.

**(B)** Treat symbol miss as a normal skip (no error metric) and update `TestInstrumentationErrors` to exercise a different failure — e.g. reduce capabilities further to force a real kprobe/sockmsg attach failure, or point discovery at a genuinely uninstrumentable target.

**(C)** Add a separate `obi_uprobe_symbol_not_found_total` metric distinct from `InstrumentationError`, and update the test to check that instead. Bigger surface change.

I'd lean toward (A) as the smallest, most compatible fix — but the framing question is really "does the project consider 'requested symbol not present in binary' an instrumentation error or a normal outcome?". Happy to push whichever direction you'd prefer.

bourbonkk:
Follow-up with the exact code showing why the counter went to zero — this makes the framing question above more concrete.

The path from `gatherGoOffsets` to the counter increment on main:

**Step 1 — `gatherGoOffsets` misses the symbol** (before this PR, `pkg/ebpf/instrumenter.go`):

```go
if !ok {
    log.Debug("ignoring function", "function", symbolName)
    continue   // probe.Skip stays false; probe.StartOffset stays 0
}
```

**Step 2 — `instrumentProbes` proceeds because `Skip == false`** (`pkg/ebpf/instrumenter.go:114-127`):

```go
for symbolName, probeArray := range probes {
    for _, probe := range probeArray {
        if probe.Skip {
            if probe.Required { ... }
            continue                    // ← what this PR now takes when Skip=true
        }
        cls, err := i.uprobe(exe, probe)  // ← reached on main
        ...
    }
}
```

**Step 3 — `i.uprobe` calls cilium with `Address: 0`, fails, and increments the counter** (`pkg/ebpf/instrumenter.go:619-651`):

```go
func (i *instrumenter) uprobe(exe *link.Executable, probe *ebpfcommon.ProbeDesc) ([]io.Closer, error) {
    if probe.Start != nil {
        up, err := exe.Uprobe("", probe.Start, &link.UprobeOptions{
            Address: probe.StartOffset,   // ← 0 when gatherGoOffsets missed
        })
        if err != nil {
            if i.metrics != nil {
                i.metrics.InstrumentationError(i.processName,
                    imetrics.InstrumentationErrorAttachingUprobe)  // ← counter++ regardless of Required
            }
            return closers, fmt.Errorf("setting uprobe (offset): %w", err)
        }
        ...
    }
    if probe.End != nil {
        if len(probe.ReturnOffsets) == 0 {
            if i.metrics != nil {
                i.metrics.InstrumentationError(i.processName,
                    imetrics.InstrumentationErrorAttachingUprobe)  // ← same
            }
            ...
        }
        for _, offset := range probe.ReturnOffsets {
            up, err := exe.Uprobe("", probe.End, &link.UprobeOptions{Address: offset})
            if err != nil {
                if i.metrics != nil {
                    i.metrics.InstrumentationError(i.processName,
                        imetrics.InstrumentationErrorAttachingUprobe)  // ← same
                }
                ...
            }
        }
    }
}
```

None of gotracer's `GoProbes()` are marked `Required: true`, so on main the `Required==false` branch in `instrumentProbes` (line 132) is what applied — no counter there — but every failed `i.uprobe` call **inside** that branch still bumped `obi_instrumentation_errors_total{error_type=attaching_uprobe}` at lines 628 / 639 / 650. That's what the test observed.

This PR short-circuits at line 118 (`if probe.Skip { continue }`) so `i.uprobe` is never called for missing symbols — and the counter stays at 0.

So the three options really break down as:

- **(A)** Keep the fix, but re-emit `InstrumentationError(..., "attaching_uprobe")` (or a new `symbol_not_found` label) from the new skip branch in `gatherGoOffsets`. One-line addition, test keeps working, observability parity.
- **(B)** Keep the fix as-is, treat this as "not an error anymore", and rewrite `TestInstrumentationErrors` to force a different failure the counter still sees (e.g. drop `PERFMON` too, or use a config that causes tracepoint / sockmsg attach to fail).
- **(C)** New dedicated metric (`obi_uprobe_symbol_not_found_total`) + update the test to check it.

Personally still leaning (A) for minimal churn, but happy to switch. Let me know which direction fits the project's semantics best.

grcevski:
I think we should go with option (A), saying symbol not found in the metric if possible. It would be good to know when we didn't discover certain symbols where they might've been expected.

bourbonkk:
Thanks @grcevski — went with option (A). Pushed `4c89fce0`:

- New `imetrics.InstrumentationErrorSymbolNotFound = "symbol_not_found"` label
- `gatherGoOffsets` now calls `InstrumentationError(processName, symbol_not_found)` on the skip branch
- Added unit-test assertions on the reporter for both the miss and resolve branches

Integration test `TestInstrumentationErrors` should pass again since it only checks that `obi_instrumentation_errors_total` has ≥1 entry with `process_name` + `error_type` labels — the new `symbol_not_found` value satisfies that.

bourbonkk:
Quick note on cardinality — the new `error_type` value is a single constant (`symbol_not_found`), so the label set grows by one value. `process_name` reuses existing values already emitted by other error types. Per-Go-app volume is bounded by the number of `GoProbes` gotracer registers (~60), incremented once at tracer creation time (not a hot path). Net effect vs main is the same total count under a differently-labelled bucket. Let me know if you'd prefer a per-process aggregate instead.

grcevski:
Thanks for your contribution @bourbonkk !

PR #2644, opened 2026-07-13
Accept application_jvm alias in config schema

MrAlias:
## Summary

- Expose the deprecated `application_jvm` feature alias in the generated configuration schema.
- Add a regression test and regenerate the configuration reference artifacts.

## Rationale

#2611 retained `application_jvm` as an alias for `application_runtime` so v0.10 configurations continue to work, but excluded the alias from `Features.JSONSchema()`. That left schema-aware editors and validation tooling rejecting a configuration OBI itself accepts.

Keeping the alias in the schema aligns validation with runtime behavior for the migration window while `application_runtime` remains the canonical feature name.

## Validation

- `make verify`
- `go test ./pkg/export`
- `make check-config-schema`
- `make lint-markdown`


MrAlias:
Not sure this is worth it.

PR #2647, opened 2026-07-13
Add per rule obfuscation string option

grcevski:
## Summary

In HTTP payload extraction we have the option to obfuscate sensitive information with an obfuscation string. I'm adding an option to define a per rule obfuscation string, such that, traces can be searched with the kind of sensitive information they contained. For example, we can obfuscate PII data with [PII] and PHI data with [PHI]. This information can be used to build sensitive data flow knowledge graph between services.

## Validation

### DIFF
--- a/internal/config/schema/instrumentation.go
+++ b/internal/config/schema/instrumentation.go
@@ -180,8 +180,8 @@ type HTTPEnrichment struct {
 
 // HTTPEnrichmentPolicy describes default HTTP payload enrichment actions.
 type HTTPEnrichmentPolicy struct {
-	DefaultAction     HTTPEnrichmentDefaultAction `yaml:"default_action"`
-	ObfuscationString string                      `yaml:"obfuscation_string"`
+	DefaultAction            HTTPEnrichmentDefaultAction `yaml:"default_action"`
+	DefaultObfuscationString string                      `yaml:"obfuscation_string"`
 }
 
 // HTTPEnrichmentDefaultAction describes default enrichment actions for headers
--- a/pkg/config/payload_extraction.go
+++ b/pkg/config/payload_extraction.go
@@ -171,6 +171,10 @@ type EnrichmentConfig struct {
 // Required fields (action, type, scope) are enforced by validate:"required" tags.
 func (c EnrichmentConfig) Validate() error {
 	for i, rule := range c.Rules {
+		if rule.ObfuscationString != nil && rule.Action != HTTPParsingActionObfuscate {
+			return fmt.Errorf("rule %d: obfuscation_string can only be used with action \"obfuscate\"", i)
+		}
+
 		switch rule.Type {
 		case HTTPParsingRuleTypeHeaders:
 			if err := validateHeaderRule(i, rule); err != nil {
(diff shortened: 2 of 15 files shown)


MrAlias (commented):
Nice addition. Being able to distinguish the kind of data that was obfuscated should make this much more useful in practice.

pkg/ebpf/common/http/http_enrichment.go
```diff
@@ -149,11 +159,11 @@ func (e *HTTPEnricher) resolveHeaderAction(
 		}
 		for i := range rule.Match.Patterns {
 			if rule.Match.Patterns[i].MatchString(matchName) {
-				return rule.Action
+				return rule.Action, &rule
```
MrAlias: `resolveHeaderAction` runs once for every request and response header. Returning `&rule` makes the range-variable copy escape to the heap, so this adds one allocation per header on the enrichment hot path. The existing eight-header benchmark moves from 34 allocs / 2,064 B to 42 allocs / 3,088 B per operation.

Can we iterate `e.headerRules` by index and return `&e.headerRules[i]`, or return only the resolved obfuscation string? I verified the index-based change restores the base allocation count without changing behavior.

grcevski: Good catch. I switched the whole loop and another one to indexing. We were copying also the rule in the loop iterator.

MrAlias (approved):

internal/config/schema/instrumentation.go
```diff
@@ -180,8 +180,8 @@ type HTTPEnrichment struct {
 
 // HTTPEnrichmentPolicy describes default HTTP payload enrichment actions.
 type HTTPEnrichmentPolicy struct {
-	DefaultAction     HTTPEnrichmentDefaultAction `yaml:"default_action"`
-	ObfuscationString string                      `yaml:"obfuscation_string"`
+	DefaultAction            HTTPEnrichmentDefaultAction `yaml:"default_action"`
+	DefaultObfuscationString string                      `yaml:"obfuscation_string"`
```
mmat11, line 184: should the yaml attribute be `default_obfuscation_string`, for consistency?

pkg/config/payload_extraction.go
```diff
@@ -229,8 +233,9 @@ func validateBodyRule(i int, rule HTTPParsingRule) error {
 type HTTPParsingPolicy struct {
 	// DefaultAction specifies what to do when no rule matches, per type.
 	DefaultAction HTTPParsingDefaultAction `yaml:"default_action"`
-	// ObfuscationString is the replacement string used when a rule's action is "obfuscate"
-	ObfuscationString string `yaml:"obfuscation_string" env:"OTEL_EBPF_HTTP_ENRICHMENT_OBFUSCATION_STRING"`
+	// DefaultObfuscationString is the replacement string used when a rule's action is "obfuscate" and
+	// the rule doesn't define it's own obfuscation_string.
+	DefaultObfuscationString string `yaml:"obfuscation_string" env:"OTEL_EBPF_HTTP_ENRICHMENT_OBFUSCATION_STRING"`
```
mmat11, line 238: same here, should yaml and env vars named `*default_obfuscation_string*`?

internal/config/schema/instrumentation.go
grcevski: I think leaving it as obfuscation_string is fine, we use similar naming for example for sampler and then per selection criteria sampler. I think docs on this should be sufficient to explain IMO.

mmat11 (approved):

PR #2652, opened 2026-07-13
Optionally add db collection name to the db.client.metrics

grcevski:
## Summary

DB collection name was missing from the db.client.operations metrics and from the span names for SQL databases. I made it optional and false by default to avoid cardinality increases.

https://opentelemetry.io/docs/specs/semconv/db/database-metrics/
https://opentelemetry.io/docs/specs/semconv/db/sql/

## Validation


mmat11 (approved):

pkg/appolly/app/request/span_getters.go
```diff
@@ -327,10 +327,14 @@ func spanOTELGetters(name attr.Name) (attributes.Getter[*Span, attribute.KeyValu
...
-			}
-			if s.Type == EventTypeAerospikeClient {
+			switch s.Type {
+			case EventTypeHTTPClient:
+				if s.SubType == HTTPSubtypeElasticsearch && s.Elasticsearch != nil {
+					return DBCollectionName(s.Elasticsearch.DBCollectionName)
+				} else if s.SubType == HTTPSubtypeSQLPP {
+					return DBCollectionName(s.Route)
+				}
+			case EventTypeSQLClient, EventTypeSQLServer, EventTypeAerospikeClient:
```
MrAlias: This leaves `db.collection.name` empty for MongoDB and Couchbase metrics. `Metrics.record` sends both `EventTypeMongoClient` and `EventTypeCouchbaseClient` through the same `db.client.operation.duration` attribute set, and both already keep the collection in `Span.Path` for trace export. When a user opts into this attribute, those records therefore lose the available collection dimension.

Can we include both event types in the path-backed case and extend the metric test to cover them?

```suggestion
			case EventTypeSQLClient, EventTypeSQLServer, EventTypeMongoClient,
				EventTypeCouchbaseClient, EventTypeAerospikeClient:
```

pkg/appolly/app/request/span_getters.go
```diff
@@ -327,10 +327,14 @@ func spanOTELGetters(name attr.Name) (attributes.Getter[*Span, attribute.KeyValu
...
 				return DBCollectionName(s.Path)
```
MrAlias, line 338: `Span.Path` already contains comma-joined tables for some SQL operations, but this PR newly exposes that value on `db.client.operation.duration`. `SQLParseOperationAndTable` returns lists for joins, subqueries, and multiple statements—for example, `ACCESS_TOKENS,SECURITY_POLICIES`—so this line publishes that entire list as one `db.collection.name` and creates a distinct series for every table combination.

The [database metric convention](https://opentelemetry.io/docs/specs/semconv/db/database-metrics/#metric-dbclientoperationduration) only defines this attribute when a call targets one collection. Can we keep the new metric attribute unset unless the instrumentation knows the operation targets exactly one collection, and add a join or multi-statement regression case?

pkg/appolly/app/request/span_getters.go
grcevski: Good catch!

pkg/appolly/app/request/span_getters.go
grcevski: Ah good catch, I didn't read that rule carefully. I think it's useful to have multiple names in there, but if the spec says otherwise we shouldn't. I'll check if we have the presence of comma and if so I'll put the name as empty.

Issue #2660, opened 2026-07-14
Support manual instrumentation for Javascript


NimrodAvni78:

NimrodAvni78:
closing this as complete even though there are still some follow ups.

PR #2661, opened 2026-07-14
Node.js manual span capture

NimrodAvni78:
Part of https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2660

### Summary

Captures spans that Node.js applications create through `@opentelemetry/api`
(`tracer.startSpan` / `startActiveSpan` / `span.end`) when the app has **no
OpenTelemetry SDK registered**. Without an SDK those API calls return
non-recording no-ops and the telemetry is silently lost — this feature makes
them real and exports them through OBI's normal traces pipeline, correlated
with OBI's existing traces.

It's the Node.js analog of OBI's existing Go manual-span support
(`bpf/gotracer/go_sdk.c`), which uprobes the no-op global-API tracer and backs
off when a real SDK is present. Since JIT-compiled JavaScript can't be uprobed
(anonymous executable mappings, no inode to attach to), the capture happens
**in-process** via a tiny injected JS bridge; only the transport back to OBI is
eBPF.

**Opt-in:** `nodejs.manual_spans: true` (`OTEL_EBPF_NODEJS_MANUAL_SPANS`,
default `false`), on top of the existing `nodejs.enabled`.

### How it works

```
customer app (@opentelemetry/api, no SDK)
   │  tracer.startSpan() / span.end()
   ▼
spanbridge.js ──── injected over the V8 inspector alongside fdextractor.js
   │               (SIGUSR1 → CDP Runtime.evaluate). Registers a minimal,
   │               dependency-free TracerProvider + AsyncLocalStorage context
   │               manager into globalThis[Symbol.for('opentelemetry.js.api.1')].
   ▼
fs.accessSync('/dev/null/obi-span/<json>')   ── sentinel uv_fs_access() path
   ▼
obi_uv_fs_access uprobe (bpf/generictracer/nodejs.c)
   │  new '-span/' branch → node_span_event_t; stamps bpf_ktime_get_ns() + pid
   │  + the current request's trace context from traces_ctx_v1 (kept fresh by
   │  fdextractor.js's async-context sentinels)
   ▼
events ringbuf → EVENT_NODE_SPAN (21)
   ▼
ReadNodeSpanEventIntoSpan → request.Span{Type: EventTypeManualSpan}
   ▼
existing traces exporter (unchanged)
```

- **Zero customer action / restart:** delivered through OBI's existing Node
  inspector injector, so it works on already-running processes (tracers
  acquired before injection go live — the bridge sets the delegate on every
  already-loaded `@opentelemetry/api` copy via `require.cache`, since
  pre-acquired `ProxyTracer`s resolve through that, not the global registry).
- **No native artifacts:** the bridge is one dependency-free JS file; transport
  reuses the sentinel-syscall channel `fdextractor.js` already uses — no addon
  to build/ship, no new eBPF attach machinery.
- **Correlation with automatic spans:** BPF stamps the in-flight request
  context at span-end, so bridge-root spans parent under OBI's automatic server
  span (specifically its "processing" sub-span), and nested manual spans keep
  their own parent chain — all on one trace. Orphan manual spans (no
  surrounding request) export with a bridge-generated trace ID.
- **Never fights an app SDK:** the bridge stays completely inert (doesn't even
  create the API registry) if a provider/context manager is already registered
  or any `@opentelemetry/sdk-*` is loaded; the app's own registration and
  exporters are untouched. The existing OTLP-export overlap detection applies
  on top.
- **Non-intrusive by default:** runs in the customer's process and is silent —
  no stdout/stderr. `OTEL_EBPF_NODEJS_SPAN_BRIDGE_DEBUG=1` opts into stderr
  diagnostics (why it stayed inert, when it activated, unexpected errors) for
  troubleshooting an injection.

### Changes

- **`pkg/internal/nodejs/spanbridge.js`** — the injected bridge: minimal
  TracerProvider/Tracer/Span + AsyncLocalStorage context manager, span→JSON
  serialization with bounded fields, sentinel transport, opt-in diagnostics.
- **`pkg/internal/nodejs/{nodejs.go,injector.go}`** — embed the bridge and
  evaluate it after `fdextractor.js` when the flag is set. Fixes a pre-existing
  injector bug: gorilla's default 4 KB websocket write buffer fragmented larger
  scripts, which the Node inspector rejects (`close 1006`) — enlarged the
  dialer buffers so each script is a single frame.
- **`bpf/generictracer/nodejs.c`** — new `-span/` sentinel branch →
  `node_span_event_t` on the ringbuf.
- **`bpf/common/{common.h,event_defs.h,common.c}`** — `node_span_event_t`,
  `EVENT_NODE_SPAN` (21), BTF anchor.
- **`pkg/ebpf/common/{common.go,node_otel_transform.go}`** — ringbuf dispatch +
  decoder producing the same `EventTypeManualSpan` shape as the Go path
  (attributes encoded identically to `tracesgen.SpanAttr`; OTel-JS
  `SpanStatusCode` mapped to Go `codes.Code`, which are inverted).
- **`pkg/ebpf/instrumenter.go`** — fixes a pre-existing double-attach: `"node"`
  and `"libuv.so"` both fall back to the executable, so `uv_fs_access` was
  uprobed twice and every sentinel fired twice; added per-module
  (symbol, program) dedup.
- **`pkg/obi/config.go`** + regenerated **`devdocs/config/{config-schema.json,CONFIG.md}`** —
  `NodeJSConfig.ManualSpans`.
- **`devdocs/nodejs-manual-spans.md`** — full architecture, constraints, and
  diagnostics doc.

### Testing

- `make docker-generate` (BPF compiles amd64+arm64), `go vet`, `golangci-lint`,
  clang-format, `make check-config-schema` — all clean.
- Decoder unit tests: `pkg/ebpf/common/node_otel_transform_test.go`
  (re-anchoring, attribute encoding, status mapping, timing, error paths).
- **Integration test** `internal/test/integration/traces_node_manual_test.go`
  (wired into `TestSuite_NodeJS`): a `/manual` route on the Node test server
  creates nested manual spans via `@opentelemetry/api` with no SDK; the test
  asserts via Jaeger that they're captured, carry their attributes/status, nest
  correctly, and re-anchor onto OBI's automatic trace. Full suite passes on a
  Linux VM including weaver schema validation.

### Known limitations

- Attributes: ≤16 per span, key ≤64 / string value ≤256 chars; array/object
  values are stringified (matches the Go path's `otel_attribute_t` budgets); on
  payload overflow (~1900 B) attributes/events are dropped so the core span
  still ships.
- Span events carry names only; links and `traceState` aren't forwarded.
- eBPF **client** spans made inside a manual span are currently siblings, not
  children, of the manual span (they parent via the fd-correlation map, not
  `traces_ctx_v1`). Go-parity nesting is implemented on a follow-up branch.
- An app that loads an OTel SDK *after* injection isn't covered by the
  inertness guard.

## Validation


bpf/generictracer/nodejs.c
```diff
@@ -51,6 +58,54 @@ static __always_inline int handle_async_switch(char *buf, const u64 pid_tgid) {
...
+// traces_ctx_v1 — maintained by the async-context sentinels above — so the
+// span can be parented under OBI's automatic server span.
+static __always_inline int handle_node_span(const char *path, const u64 pid_tgid) {
+    node_span_event_t *ev = bpf_ringbuf_reserve(&events, sizeof(node_span_event_t), 0);
+    if (!ev) {
+        return 0;
+    }
+
+    ev->type = EVENT_NODE_SPAN;
+    ev->_pad[0] = 0;
```
grcevski: Any reason we are setting these _pad bytes to 0? I don't think anything uses them.

bpf/generictracer/nodejs.c
```diff
@@ -107,10 +168,17 @@ int BPF_KPROBE(obi_uv_fs_access, void *loop, void *req, const char *path) {
 
     const u64 pid_tgid = bpf_get_current_pid_tgid();
 
-    // Async context switch: /dev/null/obi-ctx/XXXX
-    // Fires from the async_hooks 'before' callback in fdextractor.js to refresh
-    // traces_ctx_v1 before each JS callback.
     if (buf[k_delim_offset] == '-') {
+        // Manual span: /dev/null/obi-span/<json>
+        if (buf[k_variant_offset] == 's') {
+            if (!valid_pid(pid_tgid)) {
```
grcevski: Actually looking at this code I wonder why we don't actually check for valid_pid right at the start of the function, like elsewhere. I suppose we wouldn't be injecting the agent into a process we are not instrumenting, but it's a maybe similar security concern as we had for Java. Someone can add something to a node app and mimic what our agent does and we'd happily read the data. I think so far it didn't matter it's only adding internal thread correlation, so it's good you added the check, but I'd say let's pull it up at the front.

pkg/ebpf/common/node_otel_transform.go
```diff
@@ -0,0 +1,216 @@
...
+// re-anchored onto that trace: it inherits the request's trace ID, and spans
+// without an in-bridge parent are parented under OBI's automatic server span.
+func ReadNodeSpanEventIntoSpan(record *ringbuf.Record) (request.Span, bool, error) {
+	event, err := ReinterpretCast[NodeSpanEvent](record.RawSample)
+	if err != nil {
+		return request.Span{}, true, err
+	}
+
+	plen := int(event.PayloadLen)
+	if plen <= 0 || plen > len(event.Payload) {
```
grcevski, line 72: Should this be <=1, I think the BPF code breaks on <=1, although it will discard the buffer in that case.

grcevski: Is there another minimum payload size we could potentially check for?

pkg/ebpf/instrumenter.go
```diff
@@ -232,7 +232,9 @@ func (i *instrumenter) uprobeModules(p Tracer, pid app.PID, maps []*procfs.ProcM
 
 		mod, ok := modules[instrumentedIno]
 		if ok {
-			mod.probes = append(mod.probes, pMap)
+			if filtered := dedupModuleProbes(mod.probes, pMap); len(filtered) > 0 {
```
grcevski, line 419: Interesting, nice catch. So same symbol exists in two different modules from the same process? What an edge case.

pkg/ebpf/instrumenter.go
```diff
@@ -241,6 +243,51 @@ func (i *instrumenter) uprobeModules(p Tracer, pid app.PID, maps []*procfs.ProcM
 	return modules
 }
 
+// dedupModuleProbes filters out probe descriptors that would attach the same
+// eBPF programs to the same symbol as an already-collected probe map for the
+// module. This happens when several library entries (e.g. "node" and
+// "libuv.so") resolve to the same file — typically through the
+// instrument-the-executable fallback — and would otherwise attach the same
+// program twice at the same offset, duplicating every event it emits.
+func dedupModuleProbes(
```
grcevski, line 436: Do you mind adding a unit test for this?

pkg/internal/nodejs/spanbridge.js
```diff
@@ -0,0 +1,371 @@
...
+
+  // IMPORTANT: do not create or modify the registry until every guard has
+  // passed — a registry object we create carries a version string, and any
+  // app api copy with a different exact version would then fail its own
+  // setGlobalTracerProvider (registerGlobal requires an exact version match
+  // between registrants).
+  const existing = g[API_KEY];
+  // An SDK (or another agent) already registered a provider or context
+  // manager — stay inert rather than fight it.
+  if (existing && (existing.trace || existing.context)) {
```
grcevski, line 113: Do you know if the OpenTelemetry SDK currently uses similar check for double loading? I wonder the loading of the SDK happens delayed, us setting registry.trace and registry.context prevents the app from actually using OTel SDK? I'm thinking of OTel instrumented services.

If this is a real situation, I think maybe we need to write some code in the nodejs.go that will inspect the depdendencies file and see the application requires the SDK, not the API. If it has the SDK maybe it's unsafe to do this injection?

pkg/internal/nodejs/spanbridge.js
```diff
@@ -0,0 +1,371 @@
...
+        startNs: this._startWallNs.toString(),
+        durNs: durNs.toString(),
+        status: this.status.code,
+        statusMsg: this.status.message ? truncate(String(this.status.message), MAX_STATUS_MSG_LEN) : undefined,
+        attrs: this._serializeAttributes(),
+        events: this._events.length ? this._events : undefined,
+        scope: this._scope,
+      };
+      let payload = JSON.stringify(rec);
+      if (payload.length > MAX_PAYLOAD) {
```
grcevski: I think Javascript strings can be UTF-16, therefore this lenght size lies about the true byte size here. Web search tells me the byte size can be found with this `if (Buffer.byteLength(payload, 'utf8') > MAX_PAYLOAD)`. Can we use Buffer.byteLength or that needs a dependency?

Alternatively, maybe we handle partial json in the userspace code?

pkg/ebpf/common/node_otel_transform.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// nodeSpanAttr mirrors tracesgen.SpanAttr (itself the Go mirror of the BPF
+// otel_attribute_t): the JSON marshaling of a slice of these is the exact
+// Statement encoding manualSpanAttributes expects for EventTypeManualSpan.
+type nodeSpanAttr struct {
+	ValLength uint16
+	Vtype     uint8
+	Reserved  uint8
+	Key       [32]uint8
```
grcevski, line 48: Are these supposed to be the ATTR_KEY and ATTR_VALUE from the spanbridge.js? If so, I think your sizes are double in the JS code. 
```js
const MAX_ATTR_KEY_LEN = 64;
const MAX_ATTR_VALUE_LEN = 256;
```

grcevski (commented):
What a  great idea! This is really cool. I left bunch of comments, I think the really interesting issue is the registration of the SDK, I don't know what the OTel SDK does and if they do check not to overwrite the registry. If they overwrite than we are good, if not then we should try to figure out if we add more safety checks.

bpf/generictracer/nodejs.c
NimrodAvni78: yes you are right, moving it up front

pkg/internal/nodejs/spanbridge.js
NimrodAvni78: it does the [exact same thing](https://github.com/open-telemetry/opentelemetry-js/blob/0bfe818b76e5f976b8299d6875cd20b051d0f6c1/api/src/internal/global-utils.ts#L50)
```javascript
  if (!allowOverride && api[type]) {
    // already registered an API of this type
    const err = new Error(
      `@opentelemetry/api: Attempted duplicate registration of API: ${type}`
    );
    diag.error(err.stack || err.message);
    return false;
  }
```

the requires cache is stronger then the depdendencies file, depdendencies file says "is it declared as a dependency?", requires cache says "is the imported by the application?", but after looking deeper both seem not correct enough.
Applications can depend and import the otel sdk but end up not starting the instrumentations (lets say via ` OTEL_SDK_DISABLED=true` or some other runtime condition), so this will provide a false negetive.

What ill implement instead is a step-aside yield, which will basicly give up our instrumentation in the case another instrumentation tries to register. I think this is the least intrusive solution

pkg/ebpf/common/node_otel_transform.go
NimrodAvni78: These are actually already consistent — they're just in different units. bpf_probe_read_user_str returns the length including the NUL terminator, so the BPF side discards len <= 1 and then stores payload_len = len - 1. On the Go side plen is that payload_len (NUL excluded), so i think plen <= 0 is right here.
regarding minimum payload it needs to be a valid json, and i guess there are some mandatory fields we can sum up to get a minimum value. something like
```
{"v":1,"name":"x","tid":"<32 hex>","sid":"<16 hex>","kind":0,
 "startNs":"0","durNs":"0","status":0,"attrs":{},"scope":"x"}
``` 
But that might be fragile and we need to keep it up to date, so i think we can keep it this way, worst case json marshaling will fail. WDYT?

pkg/internal/nodejs/spanbridge.js
grcevski: Oh I like this approach. I think I know what you mean. Great idea!

pkg/ebpf/common/node_otel_transform.go
grcevski: Ah right, I forgot you are adjusting the size before you send it out. Yeah that sounds good, reading from Json failure should be sufficient for the weird payloads.

grcevski (approved):
LGTM! Very cool!

Makefile
```diff
@@ -319,6 +319,11 @@ test:
 	@echo "### Testing code"
 	KUBEBUILDER_ASSETS="$(shell go tool $(TOOLS_MODFILE) setup-envtest use $(ENVTEST_K8S_VERSION) -p path)" go test -short -race -a ./... -coverpkg=./... -coverprofile $(TEST_OUTPUT)/cover.all.txt
 
+.PHONY: test-nodejs
+test-nodejs:
```
mmat11, line 330: should this be tested in CI?

mmat11 (approved):
lgtm! left one comment about CI testing

Makefile
NimrodAvni78: done in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2661/changes/e9c8c65d4360d32f1809ec2c0ad50b30098bad56

mmat11:
Markdown linter is failing, can be merged after it's been fixed

pkg/internal/nodejs/spanbridge.js
```diff
@@ -0,0 +1,444 @@
...
+    const orig = apiObj[method].bind(apiObj);
+    const wrapped = function (...args) {
+      yieldToApp(why);
+      return orig(...args);
+    };
+    wrapped.__obiWrapped = true;
+    apiObj[method] = wrapped;
+  };
+
+  for (const key of Object.keys(require.cache ?? {})) {
```
MrAlias, line 482: Only API copies already present in `require.cache` have their setters wrapped. If OBI injects before an app lazily loads `@opentelemetry/api` and its SDK, that later copy calls an unwrapped `setGlobalTracerProvider`; it sees the bridge's global `trace` registration and rejects the app provider as a duplicate. The app's exporter never takes ownership, and spans continue through OBI.

The handoff needs to cover API copies loaded after injection. One viable direction is to extract this wiring into a helper and invoke it from a module-load hook before the SDK receives the API exports; because that hook is process-wide, it also needs to compose safely with existing loader patches. Please add an injection-before-first-API-load regression case on Node 18, 20, and 22.

bpf/generictracer/nodejs.c
```diff
@@ -51,6 +58,56 @@ static __always_inline int handle_async_switch(char *buf, const u64 pid_tgid) {
...
+    // event: a memset over the ~2KB struct compiles to an out-of-line memset
+    // call the BPF loader rejects ("unknown func"). The payload bytes past
+    // payload_len are never read by user space, and the _pad bytes are never
+    // read at all, so both are fine left uninitialized. has_parent_ctx MUST be
+    // set though — user space reads the parent ids only when it is non-zero.
+    ev->type = EVENT_NODE_SPAN;
+    ev->end_ktime = bpf_ktime_get_ns();
+    task_pid(&ev->pid);
+
+    const obi_ctx_info_t *octx = obi_ctx__get(pid_tgid);
```
MrAlias, line 106: `obi_ctx__get` is not guaranteed to describe the callback ending this span. `fdextractor.js` emits a `-ctx/` sentinel only when its `AsyncLocalStorage` store has an incoming FD; callbacks created outside a request emit nothing. If a background timer fires while a request is still in flight, this lookup therefore returns the last request context and incorrectly parents the background span into that trace.

Can we have the no-store async-hook path send an explicit context-clear signal and delete this map entry before node-span events rely on it? A regression test with a background span overlapping a slow request would cover this.

pkg/internal/nodejs/spanbridge.js
```diff
@@ -0,0 +1,444 @@
...
+      if (Buffer.byteLength(payload, 'utf8') > MAX_PAYLOAD) {
+        // The BPF payload buffer is fixed-size. If a span with many/large
+        // attributes overflows it, drop the variable-length parts (attributes
+        // and events) so the core span still reaches OBI rather than being
+        // dropped entirely.
+        rec.attrs = {};
+        rec.events = undefined;
+        payload = JSON.stringify(rec);
+      }
+      emit(payload);
```
MrAlias, line 345: `events` and `scope` consume transport budget here, but `nodeSpanRecord` has no events field and `ReadNodeSpanEventIntoSpan` never reads `Scope`, so neither reaches the exported span. `scope` is also unbounded and survives this fallback, allowing BPF to truncate the final JSON and userspace to drop the span.

For v1, I think we should omit both unused fields, recheck the post-fallback UTF-8 length, and return before `emit` if the bounded core still cannot fit. Actual event and scope export can follow once it has downstream model support; please update the payload docs and remove the unused decoder field with this change.

```suggestion
        attrs: this._serializeAttributes(),
      };
      let payload = JSON.stringify(rec);
      // Measure UTF-8 bytes, not String#length (UTF-16 code units): the BPF
      // side reads the sentinel path as bytes into a fixed buffer, so a
      // multi-byte payload that looks short by .length could still overflow.
      if (Buffer.byteLength(payload, 'utf8') > MAX_PAYLOAD) {
        rec.attrs = {};
        payload = JSON.stringify(rec);
      }
      if (Buffer.byteLength(payload, 'utf8') > MAX_PAYLOAD) {
        debug('dropping span: core payload exceeds transport limit');
        return;
      }
      emit(payload);
```

MrAlias (changes requested):
The approach is promising, but the SDK takeover and correlation regressions should block merging.

pkg/internal/nodejs/spanbridge.js
NimrodAvni78: Thats a great edge case, will do!

MrAlias (commented):
The CommonJS handoff fixes the prior late-load issue, but native ESM still lets the bridge block a customer's SDK from registering.

pkg/internal/nodejs/spanbridge.js
```diff
@@ -0,0 +1,483 @@
...
+  // call an unwrapped setGlobalTracerProvider: it would see our registry.trace,
+  // be refused as a duplicate registration, and the app's exporter would never
+  // take over — spans would keep flowing through OBI. Hook the CommonJS module
+  // loader and wire each api copy as it loads. We call the original loader
+  // first and guard everything, so this composes with other loader patches
+  // (e.g. import-in-the-middle) and can never break a require. NOTE: this
+  // covers CommonJS require() only; an api pulled in purely through the native
+  // ESM loader is not intercepted here.
+  try {
+    const Module = require('module');
```
MrAlias, line 509: This still only covers CommonJS. Native ESM imports never reach `Module._load`; if OBI injects before an ESM app imports `@opentelemetry/api`, that API copy never gets its setter wrapped. Its later `setGlobalTracerProvider` sees the bridge's `registry.trace`, rejects the app provider as a duplicate, and the app's exporter never takes over.

The handoff needs an ESM-safe path, or the bridge needs to remain inert where that path cannot be made safe. Please add a native-ESM injection-before-first-API-load regression on Node 18, 20, and 22.

devdocs/nodejs-manual-spans.md
```diff
@@ -0,0 +1,272 @@
...
+- **Opt-in only.** Existing Node.js support (fd extraction, context
+  propagation) is unaffected when `nodejs.manual_spans` is off; the BPF
+  `-span/` branch simply never fires because nothing emits the sentinel.
+- **Injection prerequisites are inherited** from the existing Node injector:
+  the process must not have a custom SIGUSR1 handler (checked before
+  sending the signal), and the inspector must be reachable. Injection
+  happens once per process; apps started after OBI are picked up by
+  discovery as usual.
+- **Late-registering SDKs** are handled by the step-aside (see Guards): the
+  bridge yields and the app's SDK takes over. The residual gap is an
```
MrAlias: This no longer describes the implementation: `Module._load` now wires API copies loaded by CommonJS `require()`, while native ESM remains the unhandled case. Please document those two cases explicitly, and remove the later claim that no require hook or module wrapping is used.

pkg/internal/nodejs/spanbridge.js
NimrodAvni78: @MrAlias pushed and updated:

We stopped occupying the global registry entirely. The bridge now captures only by setting itself as the delegate on each reachable @opentelemetry/api copy's ProxyTracerProvider (the require.cache scan + the Module._load hook) and never writes to the global registry. Since it no longer holds the global trace slot, there is no path left where it can block a customer's own SDK registration — ESM or otherwise. That's the "remain inert where the path can't be made safe".

Native ESM is actually covered, and there's a regression proving it. @opentelemetry/api ships no import/node export condition (only default → ./build/src/index.js, a CommonJS file); Node's runtime resolver ignores the module field, so import '@opentelemetry/api' resolves to the CommonJS entry and therefore still flows through Module._load. So a pure-ESM app that first imports the api after injection is wired like any other: it captures, and a late SDK registration hands off cleanly. Added scenario_esm.mjs (inject-before-first-API-import) to the suite, run on Node 18/20/22.

The genuinely unreachable case is an api the CommonJS loader never sees — one inlined into a bundle (webpack/esbuild), or a hypothetical future native-ESM build of the api. For those the bridge is simply inert: not captured, but not blocked (it never took the global slot). Documented as a known limitation.

In a follow up PR I prototyped a way to capture the unreachable (bundled/native-ESM) copies too, without blocking — a non-blocking accessor on the global slot that serves the bridge to capture reads but steps aside on a registration read, plus a version stamp guarded by a surface-parity tripwire against the latest api. It's validated on 18/20/22, but it reintroduces a narrow block risk (an app mangling setGlobalTracerProvider's name while keeping OTel working — which in practice breaks OTel itself) and adds a version contract to maintain. I'd rather propose it on its own so it can be weighed separately rather than expand this PR.

MrAlias (commented):
The registry redesign avoids the earlier global-slot blocker, but the mixed-copy handoff can still leave application-owned and OBI-owned telemetry active in the same process. The current head also needs to be brought back to a buildable state.

pkg/internal/nodejs/injector.go
```diff
@@ -185,6 +186,12 @@ func upgradeConnWithTimeout(conn net.Conn, wsURL string, writeBufferSize int, ti
 		NetDial: func(_, _ string) (net.Conn, error) {
 			return conn, nil
 		},
+		// Messages larger than the write buffer are sent as fragmented
+		// websocket frames, which the Node.js inspector does not support
+		// (it abruptly closes the connection). Keep the buffer larger than
+		// any script we evaluate so every message is a single frame.
+		WriteBufferSize: 256 * 1024,
```
MrAlias: `WriteBufferSize` is already set from the `writeBufferSize` argument above, so this second key makes the package fail to compile with `duplicate field name WriteBufferSize`. The caller passes `len(payload)` specifically to keep inspector messages unfragmented; can we remove this hard-coded duplicate and retain the payload-sized value?

pkg/internal/nodejs/spanbridge.js
```diff
@@ -0,0 +1,481 @@
...
+  }
+
+  // --- tracer / provider ----------------------------------------------------
+
+  // After we have yielded, the application's own provider owns the global.
+  // Tracers the app acquired-and-used before injection cached OUR tracer
+  // (OTel ProxyTracer caches the first real delegate), so route them through
+  // to the app's current tracer instead of producing dead bridge spans.
+  const activeAppTracer = (scope, version) => {
+    if (!yielded) return null;
```
MrAlias: Yielding currently depends on registration going through a setter we wrapped. If an API copy hidden in a bundle registers the SDK while a reachable copy has already cached our `Tracer`, that setter never calls `yieldToApp`; `yielded` remains false here, so the cached tracer keeps exporting through OBI instead of the application provider. This reproduces with two same-version API copies and leaves old and new tracers on different providers.

Could we treat an application provider appearing in the global registry as a handoff signal too, and add a mixed reachable/unreachable-copy regression?

pkg/internal/nodejs/injector.go
NimrodAvni78: fixed!

pkg/ebpf/common/node_otel_transform.go
```diff
@@ -0,0 +1,215 @@
...
+	var parentSpanID trace.SpanID
+	if event.HasParentCtx != 0 {
+		traceID = trace.TraceID(event.ParentTraceId)
+		parentSpanID = trace.SpanID(event.ParentSpanId)
+	} else if tid, err := parseTraceID(rec.TraceID); err == nil {
+		traceID = tid
+	}
+	// An in-bridge parent (nested manual span) wins over the request context:
+	// its own root is the one anchored to the automatic span.
+	if rec.ParentID != "" {
```
MrAlias: When `startSpan` receives an explicit external parent, `psid` belongs to that parent's trace. If BPF also supplies an OBI request context, this code replaces the trace ID but keeps the external parent ID, exporting a cross-trace parent relationship.

For the initial version, flattening manual spans under the OBI request parent is fine if distinguishing bridge-owned parents would expand the change. Could we make that fallback safe and add a regression?

pkg/internal/nodejs/spanbridge.js
```diff
@@ -0,0 +1,505 @@
...
+    // Serialize at most MAX_ATTRS attributes into a plain object, truncating
+    // keys/values and coercing unsupported value types to strings, to match
+    // what the BPF-side fixed-size attribute layout can carry.
+    _serializeAttributes() {
+      const out = {};
+      let count = 0;
+      for (const [rawKey, value] of Object.entries(this.attributes)) {
+        if (count >= MAX_ATTRS) break;
+        count++;
+        const key = truncate(rawKey, MAX_ATTR_KEY_LEN);
```
MrAlias: The whole-payload check now uses UTF-8 bytes, but the per-attribute limits still truncate JavaScript UTF-16 code units before Go copies them into fixed byte arrays. Sixteen `é` characters pass the 31-unit key limit but encode to 32 bytes, so the 31-byte copy splits the final code point and exports invalid UTF-8.

Could we truncate keys and string values on a valid UTF-8 byte boundary and add a multibyte boundary test?

MrAlias (commented):
The remaining concerns are narrow data-integrity issues.

grcevski:
Hey @NimrodAvni78, do you think you can address the latest comments from @MrAlias? It would be very cool to get this in.

NimrodAvni78:
@grcevski yeah will try fixing it today

NimrodAvni78:
@MrAlias sorry for the long delay, i did a round of fixes, can you have a look?

pkg/internal/nodejs/spanbridge.js
```diff
@@ -0,0 +1,532 @@
...
+      const parentCtx = context ?? contextManager.active();
+      const span = this.startSpan(name, options, parentCtx);
+      const ctx = parentCtx.setValue(SPAN_KEY, span);
+      return contextManager.with(ctx, fn, undefined, span);
+    }
+  }
+
+  const tracerProvider = {
+    getTracer(name, _version, _options) {
+      return new Tracer(name || 'unknown');
```
MrAlias, line 434: Non-blocking follow-up: a versioned tracer used before late SDK registration caches this bridge `Tracer`, which retains only `name`; after handoff, the app provider receives neither the original version nor options. Provider selection and span delivery remain correct, so I don't think this should hold the initial implementation. Could we track preserving all three `getTracer` arguments—and a versioned pre-acquired-tracer regression—separately?

devdocs/nodejs-manual-spans.md
```diff
@@ -0,0 +1,315 @@
...
+   ▼
+fs.accessSync('/dev/null/obi-span/<json>')      ── sentinel uv_fs_access path
+   ▼
+obi_uv_fs_access uprobe (bpf/generictracer/nodejs.c)
+   │  '-span/' branch: copies the JSON payload into a node_span_event_t,
+   │  stamps bpf_ktime_get_ns() + pid, and attaches the current request
+   │  trace context from traces_ctx_v1 (kept fresh per async context by the
+   │  fdextractor.js '-ctx/' sentinels)
+   ▼
+events ringbuf → EVENT_NODE_SPAN (20)
```
MrAlias: `EVENT_NODE_SPAN` is now `24` after the main-branch merges (`20` is `EVENT_GO_AUTO_SPAN`).

```suggestion
events ringbuf → EVENT_NODE_SPAN (24)
```

MrAlias (approved):
The latest fixes address my remaining blockers, and the manual-span capture is ready to iterate on. Please refresh the PR description to match the current registry handoff and limits; the scope-metadata edge case can follow separately.

pkg/internal/nodejs/spanbridge.js
NimrodAvni78: will take as a follow up

PR #2663, opened 2026-07-14
redis: better handling of existing connections, better pipeline parsing

mmat11:
## Summary

- don't misread old connections: data arriving first on a non-listening port is a late reply, not a request -- ignore it instead of inverting the connection
- if buffers arrive swapped, recover the command from the right side instead of dropping the event
- never name spans after reply data
- pipelines: one span per command, each with its own error
- handle binary values and truncated buffers safely
- debug printer now shows attributes for Redis client spans (was server-only)
- add BullMQ integration test

## Validation

### DIFF
--- a/pkg/ebpf/common/redis_detect_transform.go
+++ b/pkg/ebpf/common/redis_detect_transform.go
@@ -40,6 +40,50 @@ var redisErrors = [...]struct {
 	{[]byte("READONLY "), "READONLY"},
 }
 
+// core command words per https://github.com/redis/redis/tree/unstable/src/commands;
+// only a tie-break for reversal detection, so completeness is not required
+var redisKnownOps = func() map[string]struct{} {
+	ops := map[string]struct{}{}
+	for _, op := range []string{
+		"ACL", "APPEND", "AUTH", "BGREWRITEAOF", "BGSAVE", "BITCOUNT", "BITFIELD", "BITOP", "BITPOS",
+		"BLMOVE", "BLMPOP", "BLPOP", "BRPOP", "BRPOPLPUSH", "BZMPOP", "BZPOPMAX", "BZPOPMIN",
+		"CLIENT", "CLUSTER", "COMMAND", "CONFIG", "COPY", "DBSIZE", "DEBUG", "DECR", "DECRBY",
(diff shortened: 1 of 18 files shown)


pkg/ebpf/common/redis_detect_transform.go
```diff
@@ -272,12 +456,34 @@ func TCPToRedisToSpan(trace *TCPRequestInfo, op, text string, status, db int, db
...
 }
 
-func ReadGoRedisRequestIntoSpan(record *ringbuf.Record) (request.Span, bool, error) {
+func ReadGoRedisRequestIntoSpan(parseCtx *EBPFParseContext, record *ringbuf.Record) (request.Span, bool, error) {
 	event, err := ReinterpretCast[GoRedisClientInfo](record.RawSample)
 	if err != nil {
 		return request.Span{}, true, err
 	}
 
+	cmds := parseRedisCommands(event.Buf[:])
```
grcevski: I think we need to communicate the actual length of the buffer from ebpf side, now that we parse properly.

pkg/ebpf/common/tcp_detect_transform.go
grcevski, line 377: I think these checks here skip all RESP3 and we won't process any.

pkg/ebpf/common/redis_detect_transform.go
```diff
@@ -111,96 +155,236 @@ func crlfTerminatedMatch(buf []uint8, matches func(c uint8) bool) bool {
...
-	status := 0
-	if isError {
-		status = 1
+	t := buf[pos]
+	line, next, ok := readRESPLine(buf, pos)
+	if !ok {
+		return 0, false
 	}
-	return dbError, status
+	switch t {
```
grcevski, line 333: I looked into RESP3, should we handle also `!` (bulk error) and `|` (attributes)?

grcevski (commented):
Really great stuff, I left a few comments/questions. I'm not 100% certain if they are problems.

pkg/ebpf/common/tcp_detect_transform.go
mmat11: good catch, increased test coverage as well

grcevski (approved):
LGTM!

NimrodAvni78 (approved):

PR #2665, opened 2026-07-14
Add CPU go runtime metrics

marctc:
## Summary

This PR adds CPU go runtime metrics (`go.cpu.time`).

## Validation


internal/test/integration/suites_runtime_test.go
```diff
@@ -21,3 +21,17 @@ func TestRuntimeMetrics(t *testing.T) {
 	runWeaverValidation(t)
 	require.NoError(t, compose.Close())
 }
+
+func TestRuntimeMetricsGo117(t *testing.T) {
```
grcevski: Can we please add a version with Prometheus export?

pkg/export/otel/metrics_runtime.go
```diff
@@ -318,6 +325,97 @@ func recordRuntimeCounter(
...
+) {
+	if cpu == nil {
+		removeGoRuntimeCPUTime(ctx, metrics)
+		return
+	}
+
+	recordFloatRuntimeCounterWithAttributes(ctx, metrics.cpuTime, &metrics.cpuTimeValues,
+		"user", cpu.UserTime, semconv.GoCPUStateUser)
+	recordFloatRuntimeCounterWithAttributes(ctx, metrics.cpuTime, &metrics.cpuTimeValues,
+		"gc/mark/assist", cpu.GCAssistTime,
```
grcevski: Can we turn this into a shared table that is looked up between metrics.go and prom.go? I think it's easy to mistype the "gc/.." constants or miss one.

pkg/export/otel/metrics_runtime.go
```diff
@@ -318,6 +325,97 @@ func recordRuntimeCounter(
...
+	if *previous == nil {
+		*previous = map[string]int64{}
+	}
+	options := []instrument.AddOption{instrument.WithAttributes(attrs...)}
+	removeOptions := []instrument.RemoveOption{instrument.WithAttributes(attrs...)}
+
+	prev, ok := (*previous)[key]
+	currentSeconds := float64(current) / float64(time.Second)
+	if !ok || current < prev {
+		metric.Remove(ctx, removeOptions...)
```
grcevski, line 369: Nit: maybe move currentSeconds under the if !ok...

grcevski:
This looks great! The code comments I put are just nits, you can ignore them, but I think we should add a Prometheus based test.

internal/test/integration/suites_runtime_test.go
marctc: This is already using Prometheus

grcevski: I see, do you have a test for OTel export elsewhere then?

marctc: We have unit tests for the OTel path, but no dedicated integration tests for it. This integration suite validates that we expose the correct runtime metric values.

grcevski: Is it possible we add one? There's a good example how we do this in the go tracer. We have the exact same integration test, we just choose which config file to use and we have a new Test funciton in the suite. You don't have to add a new test, just another combination in the suite with an otel exporter config file and control through an environment variable which one the docker-compose is using. checkout `TestSuite_Go` and `TestSuite_PrometheusScrape`.

I want to make sure we don't accidentally mess up and produce invalid metrics in the future, e.g. someone forgets to add the new metric in otel and adds it in prometheus.

marctc: makes sense, addig it now!

mmat11 (approved):
lgtm!

grcevski (approved):
LGTM!

PR #2676, opened 2026-07-15
Fix Go process instrumentation ~regression

paweljw:
## Summary

Fix two process-discovery issues that can combine to terminate OBI while inspecting short-lived child processes:

1. A child process that did not directly match discovery criteria could inherit its tracked parent’s match
2. (Actual root cause) One such deleted executable reported `unknown` as its Go build version.

(2) caused OBI to `panic`, and my attempts to exclude the offending executable to avoid that uncovered (1). Hence the two separate safeguards. In my eyes this is regression-like, the panic didn't occur under OBI v0.7.1 - not adding support for any new versions.

Added tests cover:

 - An excluded child of a matched parent is not emitted for instrumentation.
 - Valid Go versions still load module-data offsets.
 - An unknown Go version returns an error without panicking.

Let me know if more tests, splitting this into two PRs, or providing a full repro is desired. 🙏 

## Validation

### DIFF
--- a/pkg/appolly/discover/matcher.go
+++ b/pkg/appolly/discover/matcher.go
@@ -183,7 +183,7 @@ func (m *Matcher) filterCreated(obj ProcessAttrs) (Event[ProcessMatch], bool) {
 	}
 
 	// We didn't match the process, but let's see if the parent PID is tracked, it might be the child hasn't opened the port yet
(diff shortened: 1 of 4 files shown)


mmat11 (approved):
lgtm!

grcevski (approved):
LGTM! Thanks for making this fix.

PR #2677, opened 2026-07-15
sql: fix span names to better follow semconv

mmat11:
## Summary

Span names now follow the [semconv database span naming rules](https://opentelemetry.io/docs/specs/semconv/db/database-spans/#name):
`db.query.summary` (`{op} {target1} {target2} ...`, space-separated, 255-char cap) when the query has targets,
otherwise `{db.operation.name} {target}` with target = collection, then namespace.

| Query | Span name (before) | Span name (after) |
|---|---|---|
| `SELECT * FROM users` | `SELECT users` | `SELECT users` |
| `SELECT ... FROM "public"."Users"` (quoted identifiers: Prisma, EF, Django) | `SELECT public` ❌ table name lost | `SELECT public.Users` |
| `SELECT ... FROM "public"."User" u JOIN "public"."Order" o JOIN ...` | `SELECT public,public,public,public` ❌ | `SELECT public.User public.Order ...` |
| `SELECT t.id FROM ACCESS_TOKENS t, SECURITY_POLICIES sp` | `SELECT ACCESS_TOKENS,SECURITY_POLICIES` | `SELECT ACCESS_TOKENS SECURITY_POLICIES` |
| `SELECT ... FROM users a JOIN users b` (self-join) | `SELECT users,users` | `SELECT users` |
| `INSERT INTO t1 (...) SELECT ... FROM t2` | `INSERT t1` | `INSERT t1 t2` |
| `SELECT 1` (Postgres, database known from startup message) | `SELECT` | `SELECT mydb` |
| `BEGIN` | `BEGIN` | `BEGIN` |

## `db.collection.name`: before / after

Per semconv, `db.collection.name` is only captured "if a database call is performed on a single collection".

| Query | `db.collection.name` (before) | `db.collection.name` (after) |
|---|---|---|
| `SELECT * FROM users` | `users` | `users` |
| `SELECT ... FROM "public"."Users"` | `public` ❌ | `public.Users` |
| `SELECT ... FROM t1 JOIN t2` | `t1,t2` ❌ not a real collection name | *(not set)* |
| self-join `users a JOIN users b` | `users,users` ❌ | `users` |

Multi-table information is not lost: it moves to the span name (query summary) and remains in `db.query.text`.

## Validation


grcevski (approved):
LGTM! Sweet!

PR #2679, opened 2026-07-15
Expose Config v2 validation and migration CLI

### LINKED ISSUE #2534: Add Config v2 validation and migration CLI commands
Config v2 needs user-facing tooling for validating new configuration and migrating existing OBI configuration before Config v2 becomes part of the public runtime surface.

This issue implements the release-gated CLI portion of #2251. It must ship in the same release as standalone Config v2 loading, Collector receiver Config v2 support, and the published Config v2 documentation and artifacts.

## Scope

- Add `obi config validate` for Config v2 documents.
- Add `obi config migrate` for converting supported v1 configuration to Config v2.
- Reuse the internal Config v2 schema and conversion packages rather than introducing a second parser or conversion path.
- Produce validation and migration errors that identify the relevant field or section when possible.
- Define stable exit-code behavior suitable for scripts and CI.
- Ensure migrated output passes `obi config validate`.

### Validation

- Accept valid standalone Config v2 documents.
- Reject malformed Config v2 documents.
- Reject unsupported versions.
- Report standalone-versus-receiver section errors clearly when the selected validation mode requires it.
- Do not silently reinterpret malformed Config v2 as v1 configuration.

### Migration

- Convert the supported v1 file-based configuration surface to Config v2.
- Preserve supported runtime behavior and defaults.
- Report fields that cannot be migrated without user input or that are outside the supported v1-to-v2 migration contract.
- Produce deterministic output suitable for review and source control.

## Tests

- Command tests for successful validation and migration.
- Invalid YAML and invalid Config v2 schema tests.
- Unsupported-version tests.
- Migration tests covering representative discovery, capture, network, exporter, enrichment, and daemon configuration.
- A test proving migrated output passes validation.
- Exit-code tests.

## Acceptance criteria

- [ ] `obi config validate` validates supported Config v2 documents without starting OBI.
- [ ] `obi config migrate` produces valid Config v2 for the supported v1 migration surface.
- [ ] Invalid and unsupported input returns actionable errors and non-zero exit codes.
- [ ] Migrated output is deterministic.
- [ ] The commands use the same internal parser and conversion implementation as the runtime.
- [ ] User documentation covers command usage and limitations.

## Related

- #2251
- #1758
- #594

MrAlias:
## Summary

- expose `obi config validate` for standalone and Collector receiver Config v2 files
- expose deterministic `obi config migrate` support for the v1-to-v2 contract
- reject malformed, unsupported, unknown, or lossy input with actionable errors and stable exit codes
- validate generated migrations through the shared schema and runtime conversion path before emitting YAML
- document command usage, stream behavior, environment substitution, and exit codes

## Migration behavior

Migration starts from the v1 runtime configuration model, applies the existing v1 normalization semantics, and reuses the internal Config v2 converter. It emits canonical YAML on stdout and a deterministic mapping report on stderr. A source-field round trip prevents known-but-unrepresented v1 settings from being silently dropped and reports the exact v1 paths that require manual migration.

Validation is explicitly v2-only and never falls back to v1 interpretation. Receiver mode uses the receiver schema boundary and rejects standalone-only sections before runtime validation.

## Validation

- `make generate`
- `go test ./cmd/obi/internal/configcmd ./pkg/obi ./internal/config/convert ./internal/config/schema`
- project `golangci-lint` configuration on the changed Go packages
- `make compile`
- `make lint-markdown`
- `make license-header-check`
- built-binary smoke tests for standalone validation, receiver validation, and representative v1 migration

Closes #2534

Part of #2251


cmd/obi/internal/configcmd/configcmd.go
```diff
@@ -0,0 +1,449 @@
...
+	"flag"
+	"fmt"
+	"io"
+	"os"
+	"reflect"
+	"sort"
+	"strings"
+
+	"go.yaml.in/yaml/v3"
+	legacyyaml "gopkg.in/yaml.v3"
```
mariomac, line 18: What's the reason of having two external dependencies for YAML? Can't both structures be encoded/decoded with the same library?

mariomac (commented):
This is looking great!

When the functionality is finally available I wonder if, as an extra testing, we can pick up few existing integration tests (e.g. 2 docker integration tests and 2 k8s tests), then convert v1 config to v2, and make OBI run with V2 and see that the required configuration is still active.

cmd/obi/internal/configcmd/configcmd.go
MrAlias: Yeah, good question. `legacyyaml` ensures v1 configs are decoded exactly like OBI’s existing runtime loader.

The v1 configuration types implement custom `UnmarshalYAML` methods using `gopkg.in/yaml.v3.Node`, including feature sets, selectors, regexes, and port/PID lists. Because `gopkg.in/yaml.v3.Node` and `go.yaml.in/yaml/v3.Node` are distinct Go types, using the newer package would bypass those custom unmarshallers or produce incompatible behavior.

Meaning:

- `legacyyaml` parses v1 using the same package as `obi.LoadConfig`.
- `go.yaml.in/yaml/v3` parses and emits the newer v2 schema.
- The alias also disambiguates the two packages in the same file.

Removing it would require migrating all legacy runtime YAML types to the newer module path, which is broader than this CLI change.

MrAlias:
> When the functionality is finally available I wonder if, as an extra testing, we can pick up few existing integration tests (e.g. 2 docker integration tests and 2 k8s tests), then convert v1 config to v2, and make OBI run with V2 and see that the required configuration is still active.

Agreed 👍. I'll look into adding this when the standalone OBI runtime loading for Config v2 is added in #2682.

MrAlias:
This is quite large. I'm going to split off the internal config v2 hardening that I've identified in the course of creating this into its own PR as a pre-requisite for this PR.

mariomac (approved):
LGTM!

PR #2680, opened 2026-07-15
Unify directional HTTP route policies in config v2

### LINKED ISSUE #747: Cleaner per-service and routing config
Right now there are two places where the route matching config exists:

One is on a [`per-service` level](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/e1a412f8b605fd0fe2173288a8b77923582c278e/pkg/services/custom_routes.go#L6-L9): 

```yaml
discovery:
  services:
    - namespace: multi-k
      name: java-serv
      open_ports: 8085
      routes:
        incoming:
          - /greeting123/{something}
        outgoing:
          - /*
```

And another is on a [global level](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/e1a412f8b605fd0fe2173288a8b77923582c278e/pkg/transform/routes.go#L51-L63):

```yaml
routes:
  patterns:
    - /basic/:rnd
  unmatched: path
  ignored_patterns:
    - /metrics
  ignore_mode: traces
```

We need to reconcile the two. For example, I think `ignore_patterns` and `unmatched` are useful in the `per-service` level as well. And that `incoming / outgoing` differentiation is useful on a global level as well.

Also the two being in completely different places is weird as well.

MrAlias:
Closes #747.

## Summary

- define one direction-scoped HTTP route policy for global and per-service config v2 settings
- make ignored patterns, ignore mode, unmatched routes, wildcard selection, and cardinality available at both scopes
- preserve the v1 configuration surface while mapping flat global settings into both v2 directions
- update the v2 schema, default example, migration guidance, and parity checks

## Configuration semantics

Per-service route policies override global policies field by field. Omitted fields inherit the global value, while explicitly configured arrays replace the inherited array, including an empty array that clears it. Route evaluation applies ignored patterns first, then configured or discovered routes, then unmatched-route behavior.

## Validation

- `make generate`
- `make check-config-schema`
- `make lint-markdown`
- `python3 devdocs/config/version-2.0/validate_example.py`
- `go test -race ./...`
- focused tests for config conversion, discovery, and route transformation
- `golangci-lint` across every modified Go package


mariomac (commented):
Looks great! I wonder if there is a way to simplify the code:
* Remove the code from old, non-directional route decorator (`provideRoutes`) and keep only the directional route decorator `provideDirectionalRoutes`
* If the user defines the non-directional routing patterns, copy them to both `Incoming` and `Outgoing` configuration so the directional routing code will behave as the old, global, non-directional route decorator.

MrAlias:
Thanks, Mario. I agree that a single directional path would be a cleaner end state.

I’m hesitant to take it in this PR. The non-directional decorator exists solely for config v1 compatibility, and the plan is to retire that surface once v2 has proven stable. Converting legacy configuration into the directional representation now would add a transient compatibility layer and broaden this into a v1 behavioral-parity exercise—particularly around service-level routing precedence—only to remove it with the v1 path.

I’d keep this PR focused on making v2 correct and stable, then delete the non-directional implementation as part of the v1 sunset. If you think it is worth retaining the idea, I’m happy to open a follow-up issue so we can revisit it if the v1 support horizon changes or the duplicate logic causes a concrete maintenance issue. Let me know what you think.

mariomac:
Ok @MrAlias your strategy sounds reasonable to me.

pkg/appolly/services/custom_routes.go
```diff
@@ -3,7 +3,119 @@
...
+}
+
+func (p DirectionalRoutePolicies) Clone() DirectionalRoutePolicies {
+	return DirectionalRoutePolicies{
+		Incoming: p.Incoming.Clone(),
+		Outgoing: p.Outgoing.Clone(),
+	}
+}
+
+func (o *RoutePolicyOverride) Apply(base RoutePolicy) RoutePolicy {
```
grcevski, line 78: I think this logic causes the per service routes to override the global patterns, while in v1 they were additive.

For example, users may have global rule for /health, but then define per service rule for certain problematic routes.

internal/config/convert/import.go
```diff
@@ -986,34 +1088,27 @@ func applyV2HTTPRouteDiscovery(cfg *obi.Config, discovery schema.HTTPRouteDiscov
...
-	}
-	if routes.IgnoreMode != nil {
-		dst.IgnoredEvents = *routes.IgnoreMode
-	}
-	if routes.WildcardChar != nil {
-		dst.WildcardChar = *routes.WildcardChar
-	}
-	if routes.MaxPathSegmentCardinality != nil {
-		dst.MaxPathSegmentCardinality = *routes.MaxPathSegmentCardinality
+func activateV2RuleRoutePolicies(cfg *obi.Config) {
```
grcevski, line 1095: When incoming is globally configured but outgoing is absent, an outgoing rule containing only patterns inherits an empty Unmatch, which I think is interpreted as a wildcard. The same rule with both globals absent inherits unset.

pkg/appolly/services/custom_routes.go
MrAlias: I kept replacement semantics for v2. Omitting `patterns` inherits the global array, while specifying it replaces the array; an explicit empty array can therefore clear inherited routes. This is an intentional difference from v1, documented in `migration.md` and covered by `TestDirectionalRoutesPerServiceOverrides`. Making arrays additive would remove the ability to replace or clear global patterns and conflict with the field-by-field override contract.

internal/config/convert/import.go
MrAlias: Fixed in `97194b2c`. `activateV2RuleRoutePolicies` now assigns `UnmatchUnset` independently to each direction without a global policy. An outgoing pattern-only rule no longer inherits legacy wildcard behavior when only the incoming global policy exists. `TestV2ToRuntimePreservesAbsentGlobalHTTPDirectionWithRule` covers this case. `make test` and `make compile` pass.

pkg/appolly/services/custom_routes.go
grcevski: OK, that's fair. I think that might be better for certain scenarios where there are conflicts.

grcevski (approved):
LGTM!

MrAlias:
> Ok @MrAlias your strategy sounds reasonable to me.

#2780

PR #2682, opened 2026-07-15
Enable standalone Config v2 runtime loading

### LINKED ISSUE #2535: Enable Config v2 standalone runtime loading
The Config v2 parser and conversion implementation currently remain internal and are not used by normal OBI startup. Config v2 cannot become public until the standalone `obi` command can identify, validate, and load it safely.

This issue implements the release-gated standalone runtime portion of #2251. It must ship in the same release as the Config v2 CLI, Collector receiver Config v2 support, and the published Config v2 documentation and artifacts.

## Scope

- Detect Config v2 during normal standalone OBI startup.
- Parse and validate the Config v2 document using the internal Config v2 schema package.
- Convert the Config v2 document into the runtime configuration using the internal conversion package.
- Fall back to v1 loading only when version detection returns an explicit not-Config-v2 result.
- Preserve existing v1 configuration behavior.
- Surface malformed or unsupported Config v2 as a startup error rather than retrying it as v1.
- Ensure standalone-only Config v2 sections are accepted in standalone mode.

## Required behavior

- A valid Config v2 document starts OBI with the expected runtime configuration.
- Existing v1 files continue to load through the existing path.
- A malformed Config v2 document cannot silently fall back to v1.
- An unsupported Config v2 version produces an actionable error.
- Defaulting is applied exactly once and omitted fields preserve the documented defaults.
- Configuration logging and diagnostics identify the selected configuration version without exposing sensitive values.

## Tests

- End-to-end startup test with a representative Config v2 file.
- Existing v1 configuration regression test.
- Explicit not-v2 fallback test.
- Malformed-v2 no-fallback test.
- Unsupported-version test.
- Tests for standalone-only sections.
- Default-preservation tests.

## Acceptance criteria

- [ ] Standalone OBI starts successfully from a supported Config v2 document.
- [ ] Existing v1 configuration remains supported.
- [ ] Fallback to v1 occurs only for an explicit not-v2 result.
- [ ] Malformed and unsupported Config v2 fails with an actionable startup error.
- [ ] Runtime behavior is covered by end-to-end tests.
- [ ] The feature is enabled only as part of the complete Config v2 public release gate.

## Related

- #2251
- #1758

MrAlias:
## Summary

- detect Config v2 during standalone OBI startup and convert the full declarative document through the internal schema and runtime conversion packages
- fall back to the legacy v1 loader only when v2 detection returns an explicit `NotV2Error`, while preserving the original input for v1 environment handling and defaults
- reject malformed and unsupported v2 documents without attempting v1 parsing, and emit a version-only startup diagnostic after logging is configured
- cover representative standalone sections, top-level document conversion, v1 compatibility, environment substitution, defaults, and v2 failure modes

## Why

Standalone OBI currently routes every configuration document through the legacy v1 loader. That prevents valid v2 documents from reaching the existing schema and conversion path, and an unconditional fallback could hide invalid v2 configuration by interpreting it as v1.

This change makes version detection authoritative: valid v2 documents use the complete v2 conversion path, explicit non-v2 input retains the established v1 behavior, and recognized but invalid v2 input fails startup with its v2 diagnostic.

## Runtime impact

Existing v1 configuration remains supported, including legacy environment-variable precedence and escaping. V2 defaulting and environment substitution run once, standalone-only sections are accepted, and startup logs identify only the selected configuration version without printing configuration values.

## Verification

- `go test -race ./cmd/obi ./internal/config/schema ./internal/config/convert`
- `go vet ./cmd/obi`
- `golangci-lint run ./cmd/obi --timeout=6m`
- `make license-header-check`
- `make compile`

This is part of the Config v2 release gate and should ship together with the Collector receiver work, migration documentation, and published v2 artifacts tracked from #2251.

Closes #2535


MrAlias:
Blocked by https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2700

internal/config/convert/import_document.go
```diff
@@ -238,142 +354,342 @@ func parentBasedSamplerConfig(sampler *otelconfx.ParentBasedSampler) (services.S
...
+}
+
+func httpProtocol(
+	encoding *otelconfx.OTLPHttpEncoding,
+	path string,
+) (otelcfg.Protocol, error) {
+	if encoding == nil || *encoding == otelconfx.OTLPHttpEncodingProtobuf {
+		return otelcfg.ProtocolHTTPProtobuf, nil
+	}
+	if *encoding == otelconfx.OTLPHttpEncodingJson {
```
grcevski, line 644: I think this may not work, since we rely on the env variable to be picked up by the OTel exporters, we don't explicitly set the protocol in traces.go and exporter.go.

internal/config/convert/import_document.go
```diff
@@ -238,142 +354,342 @@ func parentBasedSamplerConfig(sampler *otelconfx.ParentBasedSampler) (services.S
...
+}
+
+func headerInjector[T ~*string](
+	headers []otelconfx.NameStringValuePair,
+	headersList T,
+	path string,
+) (func(map[string]string), error) {
+	values := map[string]string{}
+	if headersList != nil {
+		if err := parseKeyValueList(*headersList, values); err != nil {
```
grcevski: I think we should use the code from OTel conf, our parseKeyValueList can't handle the headers properly I think.

```
		// Parsing follows https://github.com/open-telemetry/opentelemetry-configuration/blob/568e5080816d40d75792eb754fc96bde09654159/schema/type_descriptions.yaml#L584.
		headerslist, err := baggage.Parse(*headersList)
```

grcevski (commented):
Looks great generally, I left a few comments/questions.

internal/config/convert/import_document.go
MrAlias: The conversion assigns the runtime protocol directly for both signal paths: `TracesProtocol` is set for OTLP/gRPC and OTLP/HTTP, and `MetricsProtocol` is set the same way. `TestDocumentToRuntimeImportsOTLPHTTPExporters` covers JSON traces and protobuf metrics, while the default exporter test covers gRPC, so this path does not rely on the exporter environment lookup.

internal/config/convert/import_document.go
MrAlias: Done in 724ed1d3. Declarative `headers_list` now uses `baggage.Parse`; the exporter test covers percent-decoding (`trace%20list`) and explicit-header precedence, and `TestDocumentToRuntimeRejectsInvalidOTLPHeadersList` covers invalid syntax.

grcevski (commented):
LGTM!

This looks good to me now. When I did a secondary pass over the code I noticed stuff I hadn't from prior changes, up to you if you want to maybe fix them here or follow-up.

- `import.go` around line 2007, does it make sense to check for nullness of the things we dereference like the TraceID?
- `import.go` line 1040 should we check for channels.BufferLen >= 0?

grcevski (approved):

PR #2683, opened 2026-07-15
Fix Go Kafka Sarama nil promise issue

grcevski:
## Summary

While testing the OpenTelemetry Demo checkout service, I noticed that the Kafka publish messages were missing. Essentially, they use Go Kafka Sarama with async, but they don't provide a callback. Therefore the promise that is supposed to get called when the publish event is successfully sent is NULL and we never see the callback.

I also fixed another issue. The correlation_id that was used as a key before is not sufficient, the id is unique per kafka stream, but there could be multiple streams. Now I use the promise address which will be unique, and in the case where the client doesn't care about the event being sent we complete the event without waiting for the result to come back in.

I added an integration test based on the OTel demo checkout service. I don't think this scenario is common, but nonetheless we should handle it.

## Validation


bpf/gotracer/go_sarama.c
```diff
@@ -15,6 +15,7 @@
 
 //go:build obi_bpf_ignore
 
+#include "maps/incoming_trace_map.h"
```
mmat11:
```suggestion
#include <maps/incoming_trace_map.h>
```

bpf/gotracer/go_sarama.c
```diff
@@ -115,12 +122,28 @@ int obi_uprobe_sarama_broker_write(struct pt_regs *ctx) {
...
+            // if there's no callback we report the event right away, there's no
+            // easy way to correlate with the response
+            if (!invocation->promise) {
+                req.end_monotime_ns = bpf_ktime_get_ns();
+                kafka_client_req_t *trace =
+                    bpf_ringbuf_reserve(&events, sizeof(kafka_client_req_t), 0);
+                if (trace) {
+                    bpf_dbg_printk("Sending kafka client go trace");
+
+                    __builtin_memcpy(trace, &req, sizeof(kafka_client_req_t));
```
mmat11: nit: use `bpf_memcpy`

bpf/gotracer/go_sarama.c
mmat11: also this seems to be unused?

mmat11 (approved):
left a couple nits, lgtm

bpf/gotracer/go_sarama.c
```diff
@@ -32,24 +33,30 @@ int obi_uprobe_sarama_sendInternal(struct pt_regs *ctx) {
...
     if (b_ptr) {
-        bpf_probe_read(&correlation_id,
+        bpf_probe_read(&event.correlation_id,
                        sizeof(u32),
                        b_ptr + go_offset_of(ot, (go_offset){.v = _sarama_broker_corr_id_pos}));
     }
 
-    if (correlation_id) {
-        bpf_dbg_printk("correlation_id=%d", correlation_id);
+    if (event.correlation_id) {
```
MrAlias: non-blocking: Now that `kafka_requests` is keyed by the promise address, do we still need to gate tracking on `correlation_id`? Sarama can legitimately assign zero to the first request when `ApiVersionsRequest` is disabled, so this preserves the existing first-request drop even though the ID is no longer used for matching.

I think we can remove the correlation-ID read and both guards, relying on the promise pointer for response correlation and `if (p)` for nil protection. This can also be a follow-up if we want to keep this change scoped.

MrAlias (approved):
Looks good.

mariomac (approved):
Looks great!

PR #2684, opened 2026-07-15
Produce mqtt events only if they are valid

grcevski:
## Summary

SSL traffic can accidentally create false positives for MQTT events making ghost MQTT packets. We now consider MQTT only when we have parsed something more than the prefix. e.g. clientID or topic must be valid.

## Validation


pkg/ebpf/common/mqtt_detect_transform_test.go
```diff
@@ -6,455 +6,55 @@ package ebpfcommon
...
 
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-
-	"go.opentelemetry.io/obi/pkg/appolly/app/request"
 	"go.opentelemetry.io/obi/pkg/internal/ebpf/mqttparser"
-	"go.opentelemetry.io/obi/pkg/internal/largebuf"
 )
 
-func TestProcessMQTTEvent(t *testing.T) {
```
mariomac, line 17: Are all the old tests removed by error?

MrAlias (commented):
Thanks for working on this. I think the parser-boundary and multi-packet correctness gaps should be addressed before merging.

pkg/ebpf/common/mqtt_detect_transform.go
```diff
@@ -103,6 +103,9 @@ func ProcessMQTTEvent(pkt []byte) (*MQTTInfo, bool, error) {
 			continue
 		}
 		if !ignore {
+			if !isValidMQTTPacket(info) {
+				return nil, true, errNoSpanWorthyMQTTPackets
+			}
```
MrAlias, line 109: `ParseMQTTPackets` intentionally returns every packet in the TCP segment, and this loop is supposed to find the first one that can produce span information. Returning here means an empty-topic PUBLISH at the front of the segment prevents a later valid PUBLISH from being processed, even though the existing parse-error path advances `offset` and continues.

Can we treat this validation failure the same way and add a multi-packet regression case?

```suggestion
			if !isValidMQTTPacket(info) {
				offset += packet.Length()
				continue
			}
```

pkg/ebpf/common/mqtt_detect_transform.go
```diff
@@ -111,6 +114,13 @@ func ProcessMQTTEvent(pkt []byte) (*MQTTInfo, bool, error) {
 	return nil, true, errNoSpanWorthyMQTTPackets
 }
 
+func isValidMQTTPacket(info *MQTTInfo) bool {
+	if info == nil {
+		return false
+	}
+	return info.ClientID != "" || info.Topic != ""
```
MrAlias: This only proves the field is non-empty. `mqttparser.PacketReader.ReadString` bounds-checks the length but does not validate UTF-8 or MQTT topic rules, so arbitrary encrypted bytes can still pass as a topic. For example, `30 03 00 01 ff` parses as a PUBLISH with `Topic == "\xff"`, and this predicate accepts it, producing the ghost MQTT span this change is intended to prevent.

Can we validate MQTT UTF-8 and topic constraints in the parser before considering the packet span-worthy, and add a regression case with a non-empty invalid topic?

pkg/ebpf/common/mqtt_detect_transform.go
grcevski: good suggestion

pkg/ebpf/common/mqtt_detect_transform.go
grcevski: Good idea, I'll add more validation.

pkg/ebpf/common/mqtt_detect_transform_test.go
grcevski: Oh, man, I copy pasted a new test and I removed all the old onces.

grcevski: Fixed.

pkg/ebpf/common/mqtt_detect_transform.go
grcevski: Implemented.

pkg/internal/ebpf/mqttparser/validate.go
```diff
@@ -0,0 +1,21 @@
...
+	return utf8.ValidString(s) && !strings.ContainsRune(s, 0)
+}
+
+func ValidTopicName(s string) bool {
+	return s != "" && ValidUTF8String(s) && !strings.ContainsAny(s, "+#")
+}
+
+func ValidTopicFilter(s string) bool {
+	return s != "" && ValidUTF8String(s)
+}
```
MrAlias, line 40: This only checks that the filter is non-empty valid UTF-8. MQTT topic filters also constrain wildcard placement: `+` must occupy a whole level, and `#` must occupy the final whole level. As written, filters such as `a#b` and `a/+b` pass this check, so bytes that parse as a SUBSCRIBE with either filter can still produce the ghost MQTT span this change is meant to prevent.

Can we validate that grammar here and add the corresponding invalid-SUBSCRIBE cases to `TestIsValidMQTTPacket` and `TestValidTopicFilter`?

```suggestion
func ValidTopicFilter(s string) bool {
	if s == "" || !ValidUTF8String(s) {
		return false
	}

	levels := strings.Split(s, "/")
	for i, level := range levels {
		switch level {
		case "#":
			if i != len(levels)-1 {
				return false
			}
		case "+":
		default:
			if strings.ContainsAny(level, "+#") {
				return false
			}
		}
	}

	return true
}
```

MrAlias (commented):
The false-positive guard still permits malformed SUBSCRIBE filters to emit MQTT spans.

pkg/internal/ebpf/mqttparser/validate.go
grcevski: Sure!

MrAlias (approved):

PR #2686, opened 2026-07-15
Detect sql type in Go SQL spans

grcevski:
## Summary

Go SQL spans typically use the Go generic package and we always say the DB type is Generic, while in fact we know the type in certain cases. I added the type explicitly when we know we are dealing with Postgres or MySQL, but I also added heuristic code that determines the DB type based on the server port, which are well known for Postgres, MySQL and MSSQL.

## Validation

### DIFF
--- a/bpf/gotracer/go_sql.c
+++ b/bpf/gotracer/go_sql.c
@@ -32,13 +32,13 @@
 
 // Sets the driver type based on the Go enum:
 // const (
-//	DBGeneric SQLKind = iota + 1
+//	DBGeneric SQLKind = iota
 //	DBPostgres
 //	DBMySQL
 //	DBMSSQL
 // )
 
-enum db_sub_type { k_db_generic = 1, k_db_postgres = 2, k_db_mysql = 3, k_db_mssql = 4 };
+enum db_sub_type { k_db_generic = 0, k_db_postgres = 1, k_db_mysql = 2, k_db_mssql = 3 };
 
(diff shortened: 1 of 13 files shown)


marctc (approved):
LGTM!

bpf/gotracer/go_sql.c
```diff
@@ -30,6 +30,16 @@
...
 
+// Sets the driver type based on the Go enum:
+// const (
+//	DBGeneric SQLKind = iota + 1
+//	DBPostgres
+//	DBMySQL
+//	DBMSSQL
+// )
+
+enum db_sub_type { k_db_generic = 1, k_db_postgres = 2, k_db_mysql = 3, k_db_mssql = 4 };
```
mmat11: question: should we start from 0 here so generic also means "unknown"/unset and the userspace heuristics will also work for future additions if not set explicitly?

mmat11 (commented):
should we add a couple assertions in the integration tests on `db.system.type`, if we have any?

bpf/gotracer/go_sql.c
grcevski: I can change that, I'll make the change in the go struct.

grcevski: Changed

mmat11 (approved):
lgtm!

rafaelroquetto (approved):

PR #2692, opened 2026-07-16
store-demo: fix arm64 build failures for cartservice, emailservice, recommendationservice, loadgenerator

mtanda:
## Summary

Following the store-demo README on an Apple Silicon (arm64) Mac, 4 of the 11
service images failed to build.

- `cartservice`: `dotnet publish` fails because the linux_arm64 `protoc`
  binary bundled with `Grpc.Tools` 2.76.0 segfaults (exit code 139), a known
  upstream issue (grpc/grpc#26032, grpc/grpc#38538) already worked around in
  this repo (`internal/test/integration/components/grpc_relay/dotnet/Dockerfile`).
  Applied the same fix: install the distro `protobuf-compiler` and point
  `Protobuf_ProtocFullPath` at it instead of the NuGet-bundled binary.

- `emailservice` / `recommendationservice`: `pip install --require-hashes`
  fails because `google-auth==2.56.0` now requires `cryptography`, which
  isn't pinned/hashed in the committed `requirements.txt`. The lock had also
  drifted from `requirements.in`: `requirements.in` still pinned
  `google-api-core==2.28.1` / `requests==2.32.5`, while `requirements.txt`
  was already at `google-api-core==2.32.0` from a prior Renovate update that
  only touched the lock file. Since `google-api-core==2.32.0` requires
  `requests>=2.33.0`, the old `requests` pin also had to move. Updated
  `requirements.in` to `google-api-core==2.32.0` / `requests==2.34.2`
  (matching `main`) and regenerated `requirements.txt` with
  `uv pip compile --generate-hashes`.

- `loadgenerator`: same kind of staleness — the committed lock pinned
  `gevent==26.5.0` with hashes that no longer match what PyPI serves, and
  that version conflicts with `locust`'s `gevent<26.0.0` constraint.
  Regenerated `requirements.txt` from the existing `requirements.in`
  (`locust==2.43.0`, already consistent) with `uv pip compile --generate-hashes`.

## Test plan

- [x] Built all 4 images locally on Apple Silicon (arm64) via
      `docker build --no-cache`, with `--require-hashes` intact for all
      three Python services
- [x] Loaded images into a local `kind` cluster and confirmed `/`,
      `/product/:id`, and `/cart` on the frontend return 200 with valid HTML


mariomac (approved):
Tested locally and works!

You should first manually fix some merge conflicts.

Thanks for your contribution!

mtanda:
@mariomac 

Thanks for testing it and for the review!

I've already rebased onto the latest main, so it should be conflict-free now.

While working on this, I noticed the root cause of the drift was that Renovate wasn't configured to fully regenerate these requirements.txt lock files, it was only patching individual pinned lines instead of recompiling from requirements.in. I opened a follow-up PR to fix that: #2701. Could you also take a look at that one?

PR #2693, opened 2026-07-16
Emit one span per topic for multi-topic Kafka `Produce/Fetch`

pinoOgni:
## Summary

This PR allows OBI to emit a distinct span per topic for Kafka `Produce/Fetch` requests that reference multiple topics while previously only `Topics[0]` produced a span and the rest were dropped.

In brief:
- we now parse every topic in the `Metadata response`, not just the first, so the UUID->name cache is complete; before only first topic cached and  others resolved to `*`. (note: cache only matters for v13+, which identify topics by UUID, not name.)
- we now consume each partition entry completely so the reader stays aligned to the next topic; before under-read -> every topic after the first mis-parsed (transform used only `Topics[0]`, so the parser never needed to read past the first topic)
- emit one span per topic where the primary span from `topic[0]` and emit the rest as `extras`: same `TraceID`, unique `SpanID`, name `publish <topic> / process <topic>`.

Note: Go Sarama uprobe path unchanged in behavior because it captures one operation at a time, so it keeps using the first parsed topic.

## Validation


mariomac (approved):
LGTM!

grcevski (approved):
LGTM!

PR #2694, opened 2026-07-16
Add Go runtime memory metrics

marctc:
## Summary

  Add eBPF collection and export for the following Go runtime metrics:

  - `go.memory.used`
  - `go.memory.allocated`
  - `go.memory.allocations`

  Go 1.23+ heap statistics are collected during a stable stop-the-world GC snapshot. Older Go versions continue exporting the existing scalar runtime metrics.

## Validation


bpf/gotracer/go_runtime.c
```diff
@@ -188,22 +194,266 @@ static __always_inline void go_runtime_collect_cpu_time(const go_runtime_metric_
...
+        go_offset_of(ot, (go_offset){.v = _runtime_heap_stats_delta_large_alloc_count_pos});
+    const u64 small_alloc_count_pos =
+        go_offset_of(ot, (go_offset){.v = _runtime_heap_stats_delta_small_alloc_count_pos});
+    const u64 small_free_count_pos =
+        go_offset_of(ot, (go_offset){.v = _runtime_heap_stats_delta_small_free_count_pos});
+    const u64 stats_addr = target->memstats_addr + heap_stats_pos + stats_pos;
+
+    if (small_alloc_count_pos % sizeof(u64) || small_free_count_pos % sizeof(u64) ||
+        small_free_count_pos <= small_alloc_count_pos) {
+        bpf_dbg_printk("invalid Go runtime size-class layout");
```
mariomac: Shall we provide some extra context? e.g. the PID+NS for the process showing this behavior or some other relevant data that could point us to the source of the issue if we are debugging this feature.

Same for the rest of `bpf_dbg_printk` below.

bpf/gotracer/go_runtime.c
```diff
@@ -188,22 +194,266 @@ static __always_inline void go_runtime_collect_cpu_time(const go_runtime_metric_
...
+    u64 buckhash_sys = 0;
+    u64 gc_misc_sys = 0;
+    u64 other_sys = 0;
+    if (!go_runtime_read_offset(
+            &stacks_sys,
+            sizeof(stacks_sys),
+            target->memstats_addr,
+            go_offset_of(ot, (go_offset){.v = _runtime_memstats_stacks_sys_pos}))) {
+        bpf_dbg_printk("can't read Go runtime stacks_sys");
+        return;
```
mariomac, line 397: If there any chance that after the first `go_runtime_read_offset` invocation fails, the rest of invocations might work (and return correct data)?

In that case, I'd just zero the destination value and remove this `return`.

mariomac (commented):
Looks great! I'll wait for a second pair of eyes before merging :-D

bpf/gotracer/go_runtime.c
marctc: They might, but if one read fails the metric is incomplete. Treating it as zero would report an incorrect value, so I’d keep the early return.

MrAlias (approved):
🚀

mmat11 (approved):
lgtm! running one full VM suite just to make sure it doesn't break old verifiers

mmat11:
> lgtm! running one full VM suite just to make sure it doesn't break old verifiers

all good!

PR #2697, opened 2026-07-17
Add trace context to plain-text logs

### LINKED ISSUE #2158: Implement trace log enrichment for plain-text logs
### Component

OBI (eBPF instrumentation)

### Problem

OBI is often used for source-unavailable or legacy applications, where it's not possible to change the log format to JSON to enable OBI's trace-log correlation. Therefore it would be nice to support trace log correlation on plain text logs. We can simply prefix the log lines with the TraceID-SpanID.

I understand all the challenges with adding text to unstructured logs, it can lead to "corruption" of the logs or break existing log appenders. Therefore, I'm proposing that the support can be optionally enabled and that the user should be able to override the default TraceID-SpanID format. For example, the default behaviour could be to prepend to the logs `[TraceID-SpanID]` and we can express this like format `[%traceId-%spanId]`. The %traceId and %spanId variables are placeholders where to write the actual values, so a user can specify any format they like or remove for example %spanId altogether and get only the TraceID in the logs.

An alternative would be for OBI to write the logs through OTLP, but this would require customers to change their log scraping/appending practices. 

Another alternative would be to automatically turn plain-text logs to JSON, where the original text is in one blob field. This could prove problematic with multi line logs, like Java exception traces.

### Additional context

_No response_

MrAlias:
## Summary

- enrich non-JSON captured writes with space-separated trace and span `key=value` fields
- add shared field-name, placement, and physical-line selection settings to both configuration versions
- preserve existing correlation fields, valid non-object JSON, line endings, empty lines, and unterminated writes
- document the format, upgrade behavior, capture limits, and version 2 mappings

## Behavior

Implements [this design](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2158#issuecomment-4995690325).

Plain-text annotation is enabled by default for every service selected by the log enricher. This changes the previous treatment of non-JSON writes, which were re-emitted unchanged. Setting `plain_text.enabled: false` restores that behavior without disabling JSON enrichment.

The initial fixed `key=value` representation targets unstructured and free-form text. It is not a native encoding for every structured non-JSON protocol. Prefix or suffix placement and first, last, or every nonempty physical line can be selected explicitly.

JSON objects continue to receive structured fields. Existing configured fields are never overwritten, valid JSON arrays and scalars remain byte-identical, and services exporting their own OpenTelemetry traces receive trace-only annotation.

## Validation

- `make generate`
- race-enabled tests for all changed Go packages
- targeted `golangci-lint` for all changed Go packages
- `make check-config-schema`
- `make lint-markdown`
- `make license-header-check`
- version 2 default-example validation against the OBI and OpenTelemetry schemas
- 8 KiB single-line and many-line formatter benchmarks

Closes #2158


pkg/config/log_enricher.go
```diff
@@ -34,6 +43,80 @@ func (p LogEnricherConfig) Enabled() bool {
...
+	default:
+		return fmt.Errorf("invalid log_enricher plain text multiline mode %q", p.PlainText.Multiline)
+	}
+
+	return nil
+}
+
+func validateLogEnricherFieldName(field, value string) error {
+	if value == "" {
+		return fmt.Errorf("log_enricher field name %s must not be empty", field)
```
atoulme:
```suggestion
		return fmt.Errorf("log_enricher field name %q must not be empty", field)
```

mariomac (approved):
LGTM! Asking copilot for a second pair of "eyes"

devdocs/trace-log-correlation.md
```diff
@@ -26,14 +27,41 @@ The logenricher hooks into write paths (`tty_write`, `pipe_write`, `ksys_write`,
...
 
 Both `ITER_UBUF` (kernel ≥ 6.0, used by `write()`) and `ITER_IOVEC` (all kernel versions, used by `writev()`) iterator types are supported. The `do_writev` kprobe captures the fd for `writev()` calls so `pipe_write` can resolve the file descriptor (registered as non-required — if the symbol isn't available, `write()`-based enrichment still works).
 
+## Plain-text formatting
+
+Plain-text annotation is enabled by default for services selected by the log enricher. It emits lowercase, fixed-width IDs as space-separated `key=value` fields:
+
+> [!IMPORTANT]
+> This changes upgrade behavior for every selected service: non-JSON writes that were previously re-emitted unchanged now receive trace context by default. Set `plain_text.enabled: false` before upgrading to retain the earlier non-JSON behavior. The initial fixed `key=value` representation targets unstructured or free-form text; it is not a native encoding for every structured non-JSON protocol, whose parsers may require a format-specific representation.
```
mmat11, line 43: I think this is a bit risky to enable by default for every logenriched service. Stuff might break in v0.11 if logenricher is enabled on eg. a k8s namespace. Should we make it opt-in?

internal/config/convert/export.go
```diff
@@ -663,9 +663,24 @@ func serviceNameEnrichment(cfg *obi.Config) schema.ServiceName {
...
+	spanIDFieldName := cfg.EBPF.LogEnricher.FieldNames.SpanID
+	plainTextEnabled := cfg.EBPF.LogEnricher.PlainText.Enabled
+	plainTextPlacement := cfg.EBPF.LogEnricher.PlainText.Placement
+	plainTextMultiline := cfg.EBPF.LogEnricher.PlainText.Multiline
+
 	return &schema.Correlation{
 		LogTraceAnnotation: schema.LogTraceAnnotation{
 			Enabled: cfg.EBPF.LogEnricher.Enabled(),
+			FieldNames: schema.FieldNames{
+				TraceID: &traceIDFieldName,
```
mmat11: nit: can't these be inlined like the lines below?

edit: why are these pointers?

pkg/config/log_enricher.go
```diff
@@ -34,6 +43,80 @@ func (p LogEnricherConfig) Enabled() bool {
...
+		if r == '=' || unicode.IsSpace(r) || unicode.IsControl(r) {
+			return fmt.Errorf("log_enricher field name %s contains an invalid character", field)
+		}
+	}
+
+	return nil
+}
+
+type LogEnricherFieldNames struct {
+	// TraceID is the literal trace ID field name
```
mmat11: I think we should mention that these are the both the field names that will be checked for existence in logs already enriched by SDKs and field names that will be injected if those are missing (not sure it's worth to have 2 different options for these two scopes)

mmat11 (commented):
we need to add integration test coverage for plaintext logs (especially on the multiline part)

we also need to add integration test coverage for stuff like https://ndjson.com/ which I think would fail the JSON unmarshal and be broken by the plaintext injection

grcevski:
> we also need to add integration test coverage for stuff like https://ndjson.com/ which I think would fail the JSON unmarshal and be broken by the plaintext injection

Do you think it would break for multi line JSONs? I wonder if we can detect on the fly that log is in json format and mark the file descriptor as banned for plain text injection. Once we see a line starting with { or [ or something like that?

pkg/internal/ebpf/logenricher/formatter.go
```diff
@@ -0,0 +1,237 @@
...
+
+func newLogFormatter(cfg config.LogEnricherConfig) logFormatter {
+	return logFormatter{
+		fieldNames:       cfg.FieldNames,
+		plainText:        cfg.PlainText,
+		traceFieldPrefix: []byte(cfg.FieldNames.TraceID + "="),
+		spanFieldPrefix:  []byte(cfg.FieldNames.SpanID + "="),
+	}
+}
+
```
grcevski, line 29: Maybe something like this:
```
func isNDJSON(logLine []byte) bool {
	records := 0
	valid := true

	visitPhysicalLines(logLine, func(start, end int) bool {
		line := bytes.TrimSpace(logLine[start:end])
		if len(line) == 0 {
			return true
		}
		if !json.Valid(line) {
			valid = false
			return false
		}

		records++
		return true
	})

	return valid && records > 1
}
```

pkg/internal/ebpf/logenricher/formatter.go
```diff
@@ -0,0 +1,237 @@
...
+		if err != nil {
+			return logLine, err
+		}
+		return append(out, '\n'), nil
+	} else {
+		var unmarshalTypeError *json.UnmarshalTypeError
+		if errors.As(err, &unmarshalTypeError) {
+			return logLine, nil
+		}
+	}
```
grcevski, line 49: And then here:
```
	if isNDJSON(logLine) {
		return logLine, nil
	}
```

grcevski (commented):
Looks great, but +1 on Mattia's catch for NDJSON. I made a suggestion on how we might detect it.

mmat11:
> > we also need to add integration test coverage for stuff like https://ndjson.com/ which I think would fail the JSON unmarshal and be broken by the plaintext injection
> 
> Do you think it would break for multi line JSONs?

I think `json.Unmarshal()` would fail for multi line JSONs, making it be injected as plaintext, which would break downstream parsers

> I wonder if we can detect on the fly that log is in json format and mark the file descriptor as banned for plain text injection. Once we see a line starting with { or [ or something like that?

this would be an option, I guess we can always find edge cases here... would it make sense to start injecting a subset of widely known text formats/standards? eg. `logfmt`/`syslog`/`CLF` and language specific formats (log4j, klog, etc.)

grcevski:
> this would be an option, I guess we can always find edge cases here... would it make sense to start injecting a subset of widely known text formats/standards? eg. `logfmt`/`syslog`/`CLF` and language specific formats (log4j, klog, etc.)

Yeah maybe go more conservative and ensure we support the most common formats for Python, NodeJS, Java, .NET, Go? I added a suggestion for catching NDJSON, let's try it and have this enabled optionally by default as you suggsted and see how it behaves.

devdocs/trace-log-correlation.md
MrAlias: Yeah, I agree that changing emitted log bytes deserves care. But I am hesitant to disable the feature by default.

From a user’s perspective, selecting a service for log enrichment expresses the intent: “correlate this service’s logs with its traces.” Whether that service happens to emit JSON or plain text is usually an implementation detail, and requiring users to discover and enable a second nested option for the latter produces an inconsistent result: JSON logs work, while plain-text logs silently do not.

I think disabling one and not the other is going to be the more surprising outcome. A user can reasonably see trace correlation working for one log line or one application, then assume the enricher is broken when another selected service writes plain text. Making plain-text support available by default means the feature fulfills its primary promise across the common log formats without configuration archaeology.

The compatibility tradeoff is real, but it is scoped: log enrichment is only active for explicitly selected services, not for every workload after an upgrade. For users with a downstream consumer that requires byte-for-byte plain-text output, `plain_text.enabled: false` provides a simple, explicit opt-out while preserving the zero-config path for everyone else.

Additionally, keeping this enabled while the feature is unstable also gives us the fastest and broadest feedback on real log formats, compatibility problems, and edge cases—while the behavior is still explicitly subject to change. We should document the behavior and opt-out clearly, and revisit the default before stabilization if that feedback shows the compatibility cost is higher than expected. With those boundaries in place, I think we should retain the default of `true`, but I'm interested to hear what the rest of you all think.

internal/config/convert/export.go
MrAlias: `nil` is used to distiguish unset from empty. v2 needs to distinguish omitted settings from explicit zero values, especially `enabled: false`. The locals are only there because these exported fields need addresses. I agree the preamble is noisy; I’ll look into inline this with an appropriate pointer helper.

MrAlias: Done in f1e6499. The export now takes the configuration field addresses directly. The pointers remain necessary to preserve the distinction between omitted values and explicit zero values, including enabled: false.

pkg/config/log_enricher.go
MrAlias: Done in f1e6499. The configuration comments and docs now state that these names both preserve existing SDK-injected context and identify the fields OBI injects when they are missing. I kept one shared setting for both behaviors.

pkg/internal/ebpf/logenricher/formatter.go
MrAlias: Implemented in c646a7a with formatter unit coverage and Go gRPC integration coverage. Valid multi-record NDJSON is detected before plaintext fallback; object records receive structured trace context, while scalar records and original LF/CRLF delimiters are preserved. I chose per-record JSON enrichment rather than returning NDJSON unchanged, so it stays parseable while still providing correlation. A pretty-printed single JSON document remains handled by the existing whole-document JSON path.

pkg/internal/ebpf/logenricher/formatter.go
MrAlias: Following the detection in the preceding thread, the implementation does not return valid NDJSON unchanged: it enriches each object record as JSON. This avoids plaintext annotation corrupting an NDJSON stream while preserving trace correlation. Could you please confirm this meets the intent of the suggestion?

MrAlias:
@mmat11 @grcevski Follow-up after the rebase in 43e799c: the Go gRPC suite now covers default first_line plaintext multiline writes and NDJSON end to end. NDJSON is detected before plaintext fallback and object records are enriched as structured JSON; the associated unit tests also cover scalar preservation and LF/CRLF handling.

devdocs/trace-log-correlation.md
mmat11: I agree this would be the most sensible option if we were designing the logenricher from scratch. But the feature is already in use, so this amounts to a silent opt-in for existing users -- and the risk of breaking stuff injecting into plain-text is arguably higher than it was for JSON.

> Additionally, keeping this enabled while the feature is unstable also gives us the fastest and broadest feedback on real log formats, compatibility problems, and edge cases

WDYT about the second part of this message? https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2697#issuecomment-5023936693 
This would keep it enabled by default with reduced blast radius

MrAlias: There is an upgrade cost for people already running log enrichment on a service that writes text. I still think plaintext belongs on by default, though. Selecting a service for log enrichment is already the opt-in; the user has asked us to correlate that service’s logs. Whether a particular write came from a JSON logger or a text logger should not decide whether that happens. Given this is an unstable feature, I see this as acceptable.

I would not use format detection to soften that change. The initial proposal deliberately stops at the current write buffer. It does not try to identify the logger’s format or join separate writes into an event, because we do not have enough information to do that reliably.

An fd is not a format boundary. The same stdout fd can carry a JSON startup message, a stack trace, and normal application logs. If a JSON-looking write bans the fd, later text logs lose correlation and there is no configuration or visible signal explaining why.

We can add real support for logfmt, syslog, or runtime-specific formats later, but that should be separate work with a parser and defined behavior for each format. For now, JSON and NDJSON have structured handling; other output gets the documented key=value form. plain_text.enabled: false is the explicit opt-out for users who cannot change their text output.

If the upgrade risk turns out to be too high, defaulting plaintext to false can be a temporary rollout measure. I do not think we should start there, and I would not want it to become the long-term behavior. Leaving plaintext off permanently would make log correlation depend on a logger’s encoding even after the service has been selected for enrichment.

grcevski (approved):
LGTM!

mmat11 (approved):
lgtm -- I think we should make it really clear in the next changelog that log enrichment will automatically work on plain text from now on

PR #2700, opened 2026-07-17
Strengthen Config v2 conversion boundaries

MrAlias:
## Summary

Internal Config v2 hardening work split from the CLI surface (#2679 and #2682) so the conversion contract can be reviewed independently.

- Reject selectors and declarative fields that the legacy runtime cannot represent.
- Preserve selector semantics while applying Config v2 defaults.
- Separate static configuration validation from host-dependent checks for standalone and receiver callers.
- Harden schema parsing and generated-document handling.

## Verification

- `make generate/all`
- `go test ./internal/config/convert ./internal/config/schema ./pkg/obi`

### DIFF
--- a/devdocs/config/version-2.0/examples/default-configuration.yaml
+++ b/devdocs/config/version-2.0/examples/default-configuration.yaml
@@ -600,28 +600,12 @@ extensions:
             source_labels:
               service_name: ""
               service_namespace: ""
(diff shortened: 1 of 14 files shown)


mariomac (approved):
👍

grcevski (approved):
LGTM!

PR #2702, opened 2026-07-18
chore: Docker image housekeeping

thompson-tomo:
## Summary

Explicitly pin docker version to improve info/context available to renovate.

The jaegar tracing image has been switched as existing image is eol jaegertracing/jaeger#6321

Note kafka has not been updated as that is a very invasive change.

## Validation


internal/test/integration/docker-compose-go-mqtt.yml
```diff
@@ -2,7 +2,7 @@ services:
   vernemq:
     restart: always
     container_name: vernemq-mqtt-broker
-    image: vernemq/vernemq:latest@sha256:ad60e4c04f4fa6cadeea43ecf37351b8a76d7c343ecb5e86eda6053aed0cf524
+    image: vernemq/vernemq:2.1.2@:sha256:2c25737dde809085e28f32f3d6a0fd76dc6bbcf739ade188eccb59ea32862171
```
MrAlias: `2.1.2@:sha256` is not a valid Docker reference, so Compose aborts before starting VerneMQ with `invalid reference format`. The same typo is present in `docker-compose-python-mqtt.yml`; both references need the extra colon removed.

```suggestion
    image: vernemq/vernemq:2.1.2@sha256:2c25737dde809085e28f32f3d6a0fd76dc6bbcf739ade188eccb59ea32862171
```

internal/test/integration/docker-compose-node-rdkafka.yml
```diff
@@ -122,13 +122,12 @@ services:
       - "9090:9090"
 
   jaeger:
-    image: jaegertracing/all-in-one:1.60@sha256:4fd2d70fa347d6a47e79fcb06b1c177e6079f92cba88b083153d56263082135e
+    image: jaegertracing/jaeger:2.19.0@sha256:ede4864215be4cd85bd8c3129a2fea6c5713c5653c7282c429dba123014bc68b
```
MrAlias, line 125: Jaeger v2.19 does not accept the `--query.ui-config` flag retained below. It exits immediately with `unknown flag: --query.ui-config`, which is causing the NodeRdkafka and Java Kafka integration suites to fail. The same v1 flag remains in `docker-compose-java-kafka-400.yml`, `docker-compose-java-kafka-400-tls.yml`, and `docker-compose-java-kafka-400-lb.yml`; migrate or remove it in all four definitions before switching images.

internal/test/integration/docker-compose-php-fpm.yml
```diff
@@ -13,7 +13,7 @@ services:
       - php-fpm
 
   php-fpm:
-    image: php:fpm@sha256:3d98d6bc0e3928478209db6ccc56fd4d5e796dab9d6a7ab56055c9304bf48003
+    image: php:8.5.8fpm@sha256:0dc450d0a0e81ba501973b8e303f5d45af2ed989e08730f597d8fc07fb289efd
```
MrAlias: The intended compatibility tag is `8.5.8-fpm`. [Renovate's Docker versioning](https://docs.renovatebot.com/modules/versioning/docker/) treats the hyphenated suffix as a compatibility constraint; without the hyphen, `fpm` is parsed as a prerelease segment instead. The digest keeps this pull working, but Renovate will not track the intended FPM tag family.

```suggestion
    image: php:8.5.8-fpm@sha256:0dc450d0a0e81ba501973b8e303f5d45af2ed989e08730f597d8fc07fb289efd
```

MrAlias (commented):
Pinning these images gives Renovate the version context it needs. A few references and Jaeger configurations need adjustment so the affected integration environments continue to start cleanly.

internal/test/integration/docker-compose-node-rdkafka.yml
```diff
@@ -122,14 +122,12 @@ services:
...
     ports:
       - "16686:16686"
       - "4317"
       - "4318"
     environment:
-      - COLLECTOR_OTLP_ENABLED=true
       - LOG_LEVEL=debug
     volumes:
       - ./configs/jaeger-ui-config.json:/etc/jaeger/ui-config.json
```
MrAlias: `internal/test/integration/configs/jaeger-ui-config.json` is deleted in this patch, but this short bind mount remains. When its source is missing, Compose recreates that path as a directory in the checkout. Could we remove this now-unused `volumes` block?

internal/test/integration/docker-compose-go-mqtt.yml
```diff
@@ -116,11 +116,10 @@ services:
...
   jaeger:
-    image: jaegertracing/all-in-one:1.60@sha256:4fd2d70fa347d6a47e79fcb06b1c177e6079f92cba88b083153d56263082135e
+    image: jaegertracing/jaeger:2.19.0@sha256:ede4864215be4cd85bd8c3129a2fea6c5713c5653c7282c429dba123014bc68b
     ports:
       - "16686:16686" # Query frontend
       - "4317" # OTEL GRPC traces collector
       - "4318" # OTEL HTTP traces collector
     environment:
-      - COLLECTOR_OTLP_ENABLED=true
       - LOG_LEVEL=warn
```
MrAlias: `LOG_LEVEL=warn` is a Jaeger v1 setting. [Jaeger v2 only reads environment variables referenced by its config](https://www.jaegertracing.io/docs/2.19/deployment/configuration/#environment-variables), and the bundled all-in-one config fixes `service.telemetry.logs.level` at `info`, so this is ignored. The same stale setting remains in the other migrated compose files (`debug` in most of them). Could we pass `--set=service.telemetry.logs.level=...` or remove these variables if `info` is intentional?

internal/test/integration/docker-compose-python-graphql.yml
```diff
@@ -94,13 +94,12 @@ services:
       - "9090:9090"
 
   jaeger:
-    image: jaegertracing/all-in-one:1.60@sha256:4fd2d70fa347d6a47e79fcb06b1c177e6079f92cba88b083153d56263082135e
+    image: jaegertracing/all-in-one:1.60.0@sha256:4fd2d70fa347d6a47e79fcb06b1c177e6079f92cba88b083153d56263082135e
```
MrAlias, line 97: This still resolves to the v1.60 image and digest; only the tag changed from `1.60` to `1.60.0`, while the PR description says Jaeger was switched off the EOL image. Is this a deliberate exception after the v2 startup failure? If so, could we document it alongside the Kafka exception and track the remaining migration?

MrAlias (commented):
The image pins look sound overall. The remaining comments are non-blocking Jaeger migration cleanup and an exception worth documenting.

internal/test/integration/docker-compose-node-rdkafka.yml
thompson-tomo: Done

internal/test/integration/docker-compose-go-mqtt.yml
thompson-tomo: Have added the command option.

internal/test/integration/docker-compose-python-graphql.yml
thompson-tomo: Apologies, i got sidetracked when creating an issue to track this. Has now been raised as https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2891, in effect issues arose when upgrading.

internal/test/integration/docker-compose-python-jsonrpc.yml
```diff
@@ -94,11 +94,10 @@ services:
...
+    image: jaegertracing/jaeger:2.19.0@sha256:ede4864215be4cd85bd8c3129a2fea6c5713c5653c7282c429dba123014bc68b
     ports:
       - "16686:16686" # Query frontend
       - "4317" # OTEL GRPC traces collector
       - "4318" # OTEL HTTP traces collector
-    environment:
-      - COLLECTOR_OTLP_ENABLED=true
-      - LOG_LEVEL=debug
+    command:
+  - "--set=service.telemetry.logs.level=debug"
```
MrAlias: This line starts with a non-breaking space, which is not YAML indentation, so Compose rejects the file with `could not find expected ':'`. The same character is present in `docker-compose-python-kafka.yml`, which is why both PythonJsonRPC and PythonKafka fail before starting any containers.

```suggestion
      - "--set=service.telemetry.logs.level=debug"
```

MrAlias (commented):
The Jaeger log-level fix leaves two Compose definitions unparsable, so the affected integration suites cannot start.

MrAlias:
Not sure what is going on with CI, but can you address the merge conflicts and hopefully that will kick off another attempt?

thompson-tomo:
All done @MrAlias

MrAlias (approved):

mmat11 (approved):

PR #2712, opened 2026-07-20
Fix OTLP export detection: compare server port (HostPort) not client ephemeral port (PeerPort)

### LINKED ISSUE #2711: exclude_otel_instrumented_services: gRPC port fallback can't distinguish traces vs metrics (surfaces as a PeerPort/HostPort bug)
> **Update:** originally filed as a one-field bug (`PeerPort` vs `HostPort`). On closer analysis the port comparison is only the surface symptom of a deeper design flaw in `exclude_otel_instrumented_services`: it uses the destination port to decide traces-vs-metrics, but both signals share the same OTLP port, so it cannot attribute an opaque-`:path` export to a signal type. Fixing the field alone just trades under-detection (double collection) for over-suppression (RED loss). Retitled and expanded accordingly. The draft PR #2712 documents the field bug but is intentionally not proposed for merge as-is.

## Summary

`exclude_otel_instrumented_services` (default `true`) suppresses OBI telemetry for a service once it detects that the service exports that signal via OTLP, by inspecting OBI's own client spans. For gRPC, when the `:path` (`.../MetricsService/Export` / `.../TraceService/Export`) can't be decoded (HPACK dynamic-table index → `Path: "*"`), it falls back to comparing the span's port against the process's OTLP endpoint port. This fallback is unsound in two layers.

## Layer 1 — the port fallback compares the wrong field (dead code)

The fallback compares `s.PeerPort`, which for a client span is the client's **ephemeral source port**, not the destination port (`s.HostPort`). So it can never match in practice, and gRPC export detection effectively works only when the `:path` decodes. Exporters whose `:path` is opaque are never flagged → OBI keeps emitting RED metrics alongside the SDK → silent double collection.

Code (pinned to `8076887`, current `main`):

```go
func (s *Span) sendsOnDefaultGrpcOtelPort(defaultOtlpGRPCPort int) bool {
	otlpPort, ok := s.portFromEndpointEnvVar(envOTLPEndpoint)
	if ok {
		return otlpPort == s.PeerPort   // ephemeral source port — never matches
	}
	return s.PeerPort == defaultOtlpGRPCPort
}
```
- [`sendsOnDefaultGrpcOtelPort`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/8076887b0ee95eee145c82e7938f8e8212b20554/pkg/appolly/app/request/span.go#L2052-L2058), same pattern in [`sendsTracesOnGrpcOtelPort`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/8076887b0ee95eee145c82e7938f8e8212b20554/pkg/appolly/app/request/span.go#L2071) and [`sendsMetricsOnGrpcOtelPort`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/8076887b0ee95eee145c82e7938f8e8212b20554/pkg/appolly/app/request/span.go#L2105).
- Client spans set `PeerPort = S_port` (source), `HostPort = D_port` (destination) — [http2grpc_transform.go#L297-L299](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/8076887b0ee95eee145c82e7938f8e8212b20554/pkg/ebpf/common/http2grpc_transform.go#L297-L299); [`connection_info.h#L187`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/8076887b0ee95eee145c82e7938f8e8212b20554/bpf/common/connection_info.h#L187) ("The destination port is the server port in userspace"). The rest of `span.go` already treats `HostPort` as the server port (e.g. the `serverPort` attribute at [span.go#L1243](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/8076887b0ee95eee145c82e7938f8e8212b20554/pkg/appolly/app/request/span.go#L1243)).
- The unit tests encode the same inverted assumption (`PeerPort` set to the destination port), so they mask the bug. Introduced in #596, unchanged v0.10.0 → `main`.

## Layer 2 — the port fallback can't distinguish traces from metrics (only the last-resort branch)

The layered design itself is reasonable — path is per-signal precise, and the port fallback even prefers a signal-specific endpoint (`OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`) before the generic default. The unsound part is narrow: the **generic-default** port fallback, when the `:path` can't be decoded and a service uses a single shared endpoint for both signals. **Traces and metrics go to the same OTLP port**, so such a span carries no signal-type information. And detection is:

```go
func (s *Span) IsExportMetricsSpan(...) bool {
	...
	return s.isMetricsExportURL() || s.sendsMetricsOnOtelPort(defaultOtlpGRPCPort)
}
```
A **traces**-export span (path `TraceService/Export`, not a metrics URL) fails `isMetricsExportURL()` and falls through to the shared `sendsOnDefaultGrpcOtelPort` fallback. Note the path is only used as a *positive* metrics signal — a decoded traces path does **not** exclude the span from metrics classification. So if the port comparison were fixed, **a traces-export span (even one whose path decodes cleanly to `TraceService`) would be misclassified as a metrics export**, and the service would have its OBI RED metrics suppressed — *even if it exports no metrics at all*. (This decoded-path case is fixable — see the fix design below; only the opaque-`:path` + shared-endpoint case is genuinely ambiguous.)

Net effect of the two layers:
- **As-is (`PeerPort`, dead fallback):** under-detection → double collection for services whose metrics `:path` is opaque (e.g. Python exporters).
- **Field fixed (`HostPort`, live fallback):** over-suppression → RED loss for any traces-over-gRPC service. In opentelemetry-demo this regresses `currency` (C++, no SDK RED — OBI is the sole RED source) and `recommendation` (Python, SDK emits only traces + system metrics, no server RED).

Neither is correct. A narrow guard (require a signal-specific `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT/PROTOCOL` before using the metrics port fallback) stops the `currency` regression but also stops detecting the services this was meant to fix — `product-reviews`/`load-generator` only set the generic `OTEL_EXPORTER_OTLP_ENDPOINT`. With a generic endpoint + opaque `:path`, `product-reviews`' metrics export and `currency`' traces export are indistinguishable at the port level.

## Evidence (opentelemetry-demo 2.2.0, OBI v0.10.0)

`max by (service_name, telemetry_type) (obi_avoided_services{service_namespace="otel-demo"})`, gRPC/4317 exporters:

| service_name | runtime | traces flagged | metrics flagged |
|---|---|:--:|:--:|
| product-catalog | Go | ✅ | ✅ |
| payment | Node.js | ✅ | ✅ |
| cart | .NET | ✅ | ✅ |
| flagd | Go | ✅ | ✅ |
| frontend | Node.js | ✅ | ✅ |
| **product-reviews** | **Python** | ✅ | ❌ |
| **load-generator** | **Python** | ✅ | ❌ |
| **recommendation** | **Python** | ✅ | ❌ |

Metrics detection splits by runtime, not by endpoint/port: non-Python services are caught via the decoded `:path`; Python services (whose `:path` is frequently indexed → `Path: "*"`) fall through to the dead port fallback and are missed. `product-reviews`/`load-generator` therefore double-collect RED (`http_client_request_duration_seconds` from OBI + `http.client.duration` from the SDK). `recommendation` is a missed-detection case only (its SDK metrics don't duplicate OBI's server RED).

## Suggested directions (for discussion)

The port-based signal discrimination should be replaced, not just re-typed. Options:
1. **Reliable `:path` decoding** so detection doesn't depend on the port fallback at all (root cause of the opaque-path case).
2. **Granular suppression** — suppress only the OBI metrics whose names actually overlap the SDK's exported metrics, instead of blanket per-service suppression. This also fixes the pre-existing coarseness where a service exporting non-RED OTLP metrics (system/custom) loses its OBI server RED.
3. If a port fallback is kept, at minimum it must not classify a span whose `:path` clearly indicates the *other* signal (a `TraceService` path must never count as a metrics export), and the ambiguous `Path: "*"` case should bias toward not suppressing (avoid RED loss) rather than over-suppressing.

➡️ **A concrete, code-level fix design (Fix 1 field correction + Fix 2 decoded-path exclusion guard + Fix 3 policy for the opaque case) is in [this comment](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2711#issuecomment-5030828769).**

Happy to help with whichever direction maintainers prefer.

bourbonkk:
## Summary

Fixes #2711.

The gRPC port fallback in the OTLP export detection compares the span's `PeerPort` against the process's OTLP endpoint port (from `OTEL_EXPORTER_OTLP_*ENDPOINT`) or `default_otlp_grpc_port`. For a **client** span, `PeerPort` is the client's ephemeral source port; the destination (server) port is `HostPort`. So the fallback could never match in practice, and gRPC export detection effectively relied solely on decoding the `:path` (`.../MetricsService/Export` etc.).

Consequently `exclude_otel_instrumented_services` (default on) silently missed exporters whose `:path` was not decodable — e.g. Python gRPC exporters where `:path` is frequently only seen as an HPACK dynamic-table index reference and OBI reports `Path: "*"`. Those services kept getting OBI RED metrics emitted alongside the app SDK's own metrics (double collection).

### Change

Use `s.HostPort` instead of `s.PeerPort` in `sendsOnDefaultGrpcOtelPort`, `sendsTracesOnGrpcOtelPort`, and `sendsMetricsOnGrpcOtelPort`. Also fix the unit tests in `span_test.go`, which encoded the same inverted assumption (they set `PeerPort` to the destination port); they now set `HostPort`, matching how client spans are actually populated (`PeerPort = S_port`, `HostPort = D_port` in `http2grpc_transform.go`).

Field semantics reference: client spans set `PeerPort: int(info.ConnInfo.S_port)` / `HostPort: int(info.ConnInfo.D_port)`, and the rest of `span.go` already treats `HostPort` as the server port (e.g. the `serverPort` attribute).

### Evidence

On an opentelemetry-demo deployment, `obi_avoided_services` showed gRPC/4317 Python services (`product-reviews`, `load-generator`, `recommendation`) flagged for `traces` but not `metrics`, while gRPC/4317 non-Python services (Go/Node.js/.NET) were flagged for both. With this fix, the port fallback matches on the endpoint port even when `:path` is `"*"`, so the Python services are detected consistently.

## Validation

- [x] `go test ./pkg/appolly/app/request/` passes.
- [ ] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed. (n/a — bugfix, no behavior/support-matrix surface change)

---

_AI disclosure (per [AI-POLICY.md](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/AI-POLICY.md)): root-cause analysis and this patch were prepared with the assistance of an AI coding tool. The author has reviewed, tested, and takes full responsibility for the change._

### DIFF
--- a/pkg/appolly/app/request/span.go
+++ b/pkg/appolly/app/request/span.go
@@ -2052,9 +2052,9 @@ func (s *Span) isTracesExportURL() bool {
 func (s *Span) sendsOnDefaultGrpcOtelPort(defaultOtlpGRPCPort int) bool {
 	otlpPort, ok := s.portFromEndpointEnvVar(envOTLPEndpoint)
 	if ok {
-		return otlpPort == s.PeerPort
+		return otlpPort == s.HostPort
 	}
-	return s.PeerPort == defaultOtlpGRPCPort
+	return s.HostPort == defaultOtlpGRPCPort
 }
 
 func (s *Span) sendsTracesOnGrpcOtelPort(defaultOtlpGRPCPort int) bool {
@@ -2068,7 +2068,7 @@ func (s *Span) sendsTracesOnGrpcOtelPort(defaultOtlpGRPCPort int) bool {
 	}
 	otlpTracesPort, ok := s.portFromEndpointEnvVar(envOTLPTracesEndpoint)
 	if ok {
-		return otlpTracesPort == s.PeerPort
+		return otlpTracesPort == s.HostPort
 	}
 	return s.sendsOnDefaultGrpcOtelPort(defaultOtlpGRPCPort)
 }
@@ -2102,7 +2102,7 @@ func (s *Span) sendsMetricsOnGrpcOtelPort(defaultOtlpGRPCPort int) bool {
 	}
 	otlpMetricsPort, ok := s.portFromEndpointEnvVar(envOTLPMetricsEndpoint)
 	if ok {
-		return otlpMetricsPort == s.PeerPort
+		return otlpMetricsPort == s.HostPort
 	}
 	return s.sendsOnDefaultGrpcOtelPort(defaultOtlpGRPCPort)
 }
--- a/pkg/appolly/app/request/span_test.go
+++ b/pkg/appolly/app/request/span_test.go
@@ -810,33 +810,33 @@ func TestDetectsOTelExport(t *testing.T) {
 			exports: false,
 		},
 		{
-			name: "OTEL_EXPORTER_OTLP_METRICS_ENDPOINT != span.PeerPort doesn't export",
+			name: "OTEL_EXPORTER_OTLP_METRICS_ENDPOINT != span.HostPort doesn't export",
 			span: Span{
-				Type: EventTypeGRPCClient, PeerPort: 8080, Method: "GET", Path: "*", RequestStart: 100, End: 200, Status: 0,
+				Type: EventTypeGRPCClient, HostPort: 8080, Method: "GET", Path: "*", RequestStart: 100, End: 200, Status: 0,
 				Service: svc.Attrs{EnvVars: map[string]string{"OTEL_EXPORTER_OTLP_METRICS_ENDPOINT": "http://localhost:4317"}},
 			},
 			exports: false,
 		},
 		{
-			name: "OTEL_EXPORTER_OTLP_ENDPOINT != span.PeerPort doesn't export",
+			name: "OTEL_EXPORTER_OTLP_ENDPOINT != span.HostPort doesn't export",
 			span: Span{
-				Type: EventTypeGRPCClient, PeerPort: 8080, Method: "GET", Path: "*", RequestStart: 100, End: 200, Status: 0,
+				Type: EventTypeGRPCClient, HostPort: 8080, Method: "GET", Path: "*", RequestStart: 100, End: 200, Status: 0,
 				Service: svc.Attrs{EnvVars: map[string]string{"OTEL_EXPORTER_OTLP_ENDPOINT": "http://localhost:4317"}},
 			},
 			exports: false,
 		},
 		{
-			name: "OTEL_EXPORTER_OTLP_METRICS_ENDPOINT == span.PeerPort export",
+			name: "OTEL_EXPORTER_OTLP_METRICS_ENDPOINT == span.HostPort export",
 			span: Span{
-				Type: EventTypeGRPCClient, PeerPort: 9090, Method: "GET", Path: "*", RequestStart: 100, End: 200, Status: 0,
+				Type: EventTypeGRPCClient, HostPort: 9090, Method: "GET", Path: "*", RequestStart: 100, End: 200, Status: 0,
 				Service: svc.Attrs{EnvVars: map[string]string{"OTEL_EXPORTER_OTLP_METRICS_ENDPOINT": "http://localhost:9090"}},
 			},
 			exports: true,
 		},
 		{
-			name: "OTEL_EXPORTER_OTLP_ENDPOINT == span.PeerPort export",
+			name: "OTEL_EXPORTER_OTLP_ENDPOINT == span.HostPort export",
 			span: Span{
-				Type: EventTypeGRPCClient, PeerPort: 9090, Method: "GET", Path: "*", RequestStart: 100, End: 200, Status: 0,
+				Type: EventTypeGRPCClient, HostPort: 9090, Method: "GET", Path: "*", RequestStart: 100, End: 200, Status: 0,
 				Service: svc.Attrs{EnvVars: map[string]string{"OTEL_EXPORTER_OTLP_ENDPOINT": "http://localhost:9090", "OTEL_EXPORTER_OTLP_TRACES_PROTOCOL": "http/protobuf"}},
 			},
 			exports: true,
@@ -844,14 +844,14 @@ func TestDetectsOTelExport(t *testing.T) {
 		{
 			name: fmt.Sprintf("no otel metrics environment sends to %x export", defaultOtlpGRPCPort),
 			span: Span{
-				Type: EventTypeGRPCClient, PeerPort: 4317, Method: "GET", Path: "*", RequestStart: 100, End: 200, Status: 0,
+				Type: EventTypeGRPCClient, HostPort: 4317, Method: "GET", Path: "*", RequestStart: 100, End: 200, Status: 0,
 				Service: svc.Attrs{EnvVars: map[string]string{"OTEL_EXPORTER_OTLP_TRACES_PROTOCOL": "http/protobuf"}},
 			},
 			exports: true,
 		},
 		{
 			name:    fmt.Sprintf("no otel environment sends to anything other the %d doesn't export", defaultOtlpGRPCPort),
-			span:    Span{Type: EventTypeGRPCClient, PeerPort: 8080, Method: "GET", Path: "*", RequestStart: 100, End: 200, Status: 0},
+			span:    Span{Type: EventTypeGRPCClient, HostPort: 8080, Method: "GET", Path: "*", RequestStart: 100, End: 200, Status: 0},
 			exports: false,
 		},
 	}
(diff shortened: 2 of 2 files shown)


bourbonkk:
### Observed detection asymmetry (opentelemetry-demo, live `obi_avoided_services`)

`max by (service_name, telemetry_type) (obi_avoided_services{service_namespace="otel-demo"})`, filtered to the gRPC/4317 exporters:



| service_name | runtime | traces flagged | metrics flagged |
|---|---|:--:|:--:|
| product-catalog | Go | ✅ | ✅ |
| payment | Node.js | ✅ | ✅ |
| cart | .NET | ✅ | ✅ |
| flagd | Go | ✅ | ✅ |
| frontend | Node.js | ✅ | ✅ |
| **product-reviews** | **Python** | ✅ | ❌ |
| **load-generator** | **Python** | ✅ | ❌ |
| **recommendation** | **Python** | ✅ | ❌ |

All eight services export OTLP over gRPC to the same collector port (`OTEL_EXPORTER_OTLP_ENDPOINT=http://<collector>:4317`). Metrics export detection succeeds for the Go/Node.js/.NET services but consistently fails for the Python ones, so `exclude_otel_instrumented_services` keeps emitting OBI RED metrics for `product-reviews`/`load-generator` alongside their SDK metrics (double collection). `recommendation` is included as detection evidence — it exports OTLP metrics too, but its SDK doesn't emit duplicate server RED, so it isn't a harmful-duplication case, only a missed-detection one.

To rule out the "these services just don't export metrics" reading: they are configured to export metrics over gRPC/4317 (the same endpoint used for traces, which *is* detected), and their SDK metrics do arrive in the backend.

This is symptom/correlation evidence (metrics detection splits by runtime, not by endpoint or port). The root cause is the `PeerPort` vs `HostPort` comparison in the gRPC port fallback described above: non-Python services are still caught via the decoded `:path`, while Python services — whose `:path` is frequently only observed as an HPACK dynamic-table index (`Path: "*"`) — fall through to the fallback, which never matches because it compares the client's ephemeral port. This PR fixes that comparison.

bourbonkk:
### Converting to draft — the one-line fix trades under-detection for over-suppression

Digging further, swapping `PeerPort` → `HostPort` corrects the dead comparison but exposes a deeper design problem, so I don't think it should land as-is.

**Root of the deeper issue:** traces and metrics are both exported to the same OTLP port, so the port fallback cannot distinguish signal type for a span whose gRPC `:path` did not decode (`Path: "*"`). And `IsExportMetricsSpan` is `isMetricsExportURL() || sendsMetricsOnOtelPort(...)` — a **traces**-export span (path `TraceService/Export`, not a metrics URL) always falls through to the shared `sendsOnDefaultGrpcOtelPort` fallback.

**Consequence of the "fixed" comparison:** any service that exports **traces** over gRPC to the OTLP endpoint port gets its trace-export span misclassified as a metrics export, so it is flagged as a metrics exporter and its OBI RED metrics are suppressed — *even if it exports no metrics at all*. In the opentelemetry-demo this regresses services where OBI is the sole/valuable RED source, e.g. `currency` (C++, no SDK RED) and `recommendation` (Python, SDK emits only traces + system metrics, no server RED). They would lose their OBI RED after this change.

**Why a narrower guard doesn't work:** requiring a signal-specific env var (`OTEL_EXPORTER_OTLP_METRICS_ENDPOINT/PROTOCOL`) before using the metrics port fallback would stop the `currency` regression, but the services this was meant to fix (`product-reviews`, `load-generator`) only set the generic `OTEL_EXPORTER_OTLP_ENDPOINT`, so they'd stop being detected too. With a generic endpoint + opaque `:path`, `product-reviews`' metrics export and `currency`' traces export are indistinguishable at the port level.

So the port-based signal discrimination is fundamentally unsound, not just mis-typed. I'll keep this PR as a draft to document the dead-code bug, and move the discussion to a redefined #2711 focused on the design flaw. Likely correct directions: (a) reliable `:path` decoding so detection doesn't depend on the port fallback, or (b) make suppression granular — suppress only the OBI metrics whose names actually overlap the SDK's exported metrics — instead of blanket per-service suppression.

grcevski (approved):
LGTM! Nice catch, yeah our hostport is always the server port for both servers and clients.

mariomac (approved):
Thanks for your contribution!

PR #2716, opened 2026-07-20
Fix OTLP trace-export internal metrics to reflect real send outcomes

NimrodAvni78:
## Summary

`obi.otel.trace.exports` and `obi.otel.trace.export.errors` did not reflect what
actually happened on the wire. In `getTracesExporter`, the OTLP exporter was
created with its own asynchronous sending queue, and the instrumentation wrapper
(`instrumentedTracesExporter`) sat *above* that queue. `ConsumeTraces` therefore
returned as soon as the batch was **enqueued** — always success — while the real
HTTP/gRPC send (and any failure) happened later on the exporter's internal
goroutine, invisible to the wrapper.

Observable effect, against an unreachable collector:

| | `trace.exports` | `trace.export.errors` |
|---|---|---|
| before | counts up (false — these are enqueues) | never increments |
| after | `0` | counts real connection failures |

### Change

`getTracesExporter` now keeps the underlying otlphttp / otlpgrpc exporter
**synchronous** (its own queue and retry disabled) and stacks the sending queue
and retry on an outer `exporterhelper` whose base consume function is the
instrumentation wrapper (`queueInstrumentedTraces`). The wrapper is now invoked
by the queue consumer with the real send result, so:

- `obi.otel.trace.exports` counts genuinely delivered batches, not enqueues.
- `obi.otel.trace.export.errors` counts real send failures (per attempt).

Applied to both the HTTP and gRPC paths (gRPC previously had no outer helper at
all, so it now matches HTTP). This also removes the redundant double-queue on the
HTTP path — the exporter's internal queue plus the outer `exporterhelper` queue
were both configured from the same settings.

### Metrics exporter — no change needed

`obi.otel.metric.exports` / `obi.otel.metric.export.errors` were already correct:
the metrics path uses the OTel **SDK** exporter (`otlpmetrichttp` / `grpc`) driven
by a synchronous `PeriodicReader`, which has no async sending queue — the wrapper
already observes the real `Export()` outcome. A regression test is added to lock
this in.

## Validation

New unit tests in `pkg/export/otel`:

- `TestTracesExportInternalMetrics` — a reachable collector records a success on
  `trace.exports`; an unreachable endpoint records `trace.export.errors` (the
  case that was impossible before). The tests **enable the sending queue** so
  they exercise the async path; the failure case was confirmed to fail on the
  pre-fix code and pass on the fix.
- `TestMetricsExportInternalMetrics` — same success/failure coverage for the
  metrics exporter, locking in that it already observes real outcomes.

End-to-end on a live OBI instance:

- Traces to a dead endpoint → `obi_otel_trace_export_errors_total` increments
  (`connect: connection refused`) with `trace_exports` at `0`; healthy endpoint →
  `trace_exports` climbs, errors stay flat.
- Metrics to a dead endpoint (internal metrics observed via the Prometheus
  exporter) → `obi_otel_metric_export_errors_total` increments, confirming the
  metrics path already surfaces real failures.

### DIFF
--- a/pkg/export/otel/metrics_export_internal_test.go
+++ b/pkg/export/otel/metrics_export_internal_test.go
@@ -20,16 +20,14 @@ import (
 	"go.opentelemetry.io/obi/pkg/export/imetrics"
 )
 
-// countingMetricsReporter records how many metric exports the instrumentation
-// wrapper observed as successful vs failed.
 type countingMetricsReporter struct {
 	imetrics.NoopReporter
 	exports    atomic.Int64
 	exportErrs atomic.Int64
 }
 
-func (c *countingMetricsReporter) OTELMetricExport(int)        { c.exports.Add(1) }
-func (c *countingMetricsReporter) OTELMetricExportError(error) { c.exportErrs.Add(1) }
+func (c *countingMetricsReporter) OTELMetricExport(metrics int) { c.exports.Add(int64(metrics)) }
+func (c *countingMetricsReporter) OTELMetricExportError(error)  { c.exportErrs.Add(1) }
 
 func oneResourceMetric() *metricdata.ResourceMetrics {
 	return &metricdata.ResourceMetrics{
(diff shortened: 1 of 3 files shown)


grcevski (approved):
LGTM!

mariomac (approved):
Good catch!

PR #2721, opened 2026-07-20
Add host metadata to OBI internal metrics

NimrodAvni78:
## Summary

OBI's own internal metrics (`obi.*`) were shipping **without host metadata** —
no `host.id`, no cloud/node resource attributes. Every internal metric carried
only the service/SDK attributes, so they couldn't be attributed to the host they
came from.

Root cause is an ordering defect in `BuildCommonContextInfo`
(`pkg/instrumenter/instrumenter.go`). The internal-metrics reporter builds its
OTLP resource from `ctxInfo.NodeMeta` at construction time, but the reporter was
created **before** `NodeMeta` was populated:

```
internalMetrics()      → reads ctxInfo.NodeMeta   // still zero-valued here
K8sInformer            → needs ctxInfo.Metrics
NodeMeta = NewNodeMeta → needs ctxInfo.K8sInformer // populated only now
```

The three form a cycle (`Metrics → K8sInformer → NodeMeta → Metrics`), so the
reporter simply couldn't be moved after `NodeMeta` without breaking a link.

### Change

The cycle is breakable: `NewMetadataProvider` only *stores* the internal-metrics
reporter (it's read later, when the store is built in `Get()`), and `NewNodeMeta`
uses the informer but not its metrics reporter. So the order is now:

1. Create the kube informer with a `NoopReporter`.
2. Resolve `NodeMeta`.
3. Create the internal-metrics reporter — now reading a populated `NodeMeta`.
4. Wire the real reporter back into the informer via the new
   `MetadataProvider.SetInternalMetrics(...)`.

## Validation


grcevski (approved):
LGTM!

grcevski:
needs merge with main

PR #2722, opened 2026-07-20
Parse all topics in Kafka Produce requests

MrAlias:
## Summary

This follow-up completes multi-topic Kafka Produce request instrumentation by consuming every topic and partition entry while retaining the first partition for existing span attributes. It skips nullable and compact record payloads so the parser remains aligned for subsequent topics.

Coverage now exercises flexible and non-flexible Produce requests, including the transformed Kafka attributes emitted for multiple topics.

## Testing

- `make docker-generate`
- `make lint`
- `make test`


grcevski (approved):
LGTM! Nice catch!

grcevski:
needs merge with main

mmat11 (approved):

Issue #2571, opened 2026-07-02
Weaver validation should cover k8s tests


NimrodAvni78:

MrAlias:
@NimrodAvni78 did https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2723 resolve this?

MrAlias:
Waiting on https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2785

PR #2723, opened 2026-07-20
Weaver validation for OBI's internal telemetry

NimrodAvni78:
## Summary

Declares OBI's own meta-telemetry (the `obi.*` internal metrics) in the weaver
semantic-convention registry and adds integration suites that emit each metric
over OTLP and live-check it.

The goal is a *live* test per metric: the metric is actually emitted by OBI and
weaver-validated in a real run — not just declared.

## What the change contains

| File | Kind | Purpose |
|------|------|---------|
| `schemas/obi/groups/obi_internal.yaml` | modified | registry declarations for the `obi.*` internal surface, beside the `target.info` family |
| `pkg/export/otel/metrics_internal.go` | modified | attribute-key namespacing (see below) |
| `internal/test/weavercheck/weavercheck.go` | new | parse a weaver live-check report and `Validate()` it (gate on actionable advice types) |
| `internal/test/integration/weaver.go` | modified | `runWeaverValidation()` — run enforce-mode weaver validation over the report emitted by the weaver sidecar |
| `internal/test/integration/internal_metrics_otel_test.go` + `docker-compose-internal-metrics.yml` + `configs/obi-config-internal-metrics.yml` + `configs/prometheus-config-internal-metrics.yml` | new | dedicated internal-metrics suite (metrics + eBPF families) |
| `internal/test/integration/internal_metrics_traces_otel_test.go` + `docker-compose-internal-metrics-traces.yml` + `configs/obi-config-internal-metrics-traces.yml` | new | traces-enabled internal-metrics suite (trace-export + ring-buffer-flush families) |
| `internal/test/integration/configs/obi-config-netolly.yml` + `suites_network_test.go` | modified | netolly suite validates the network packet counters |
| `internal/test/integration/avoided_services_otel_test.go` + `docker-compose-avoided-services.yml` + `configs/obi-config-avoided-services.yml` | new | avoided-services suite |

Suites are discovered automatically by the CI integration-test matrix
(`scripts/generate-integration-matrix.sh` greps `func Test*`).

## Registry declarations

`obi_internal.yaml` declares the full `obi.*` internal surface emitted by
`pkg/export/otel/metrics_internal.go` (**16 metrics**) so the registry is a
complete contract. Metric instrument/unit mirror what OBI emits over OTLP.

## Attribute namespacing (user-facing)

Acting on weaver's "must include a namespace" advice, the previously bare
internal-metric attribute keys are namespaced. Emitter and registry are changed
together so the emitted keys match the declarations. **This renames attribute
keys on OBI's internal OTLP metrics — a user-facing change for anyone consuming
them.** The Prometheus internal exporter (`iprom.go`) is intentionally left
unchanged: Prometheus labels cannot carry dots and follow the underscore
convention, and weaver only validates the OTLP path.

| Old (bare) key | New key |
|---|---|
| `process_name` | semconv **`process.executable.name`** |
| `error_type` | semconv **`error.type`** |
| `error` | `obi.error` |
| `goos` / `goarch` / `goversion` | `obi.goos` / `obi.goarch` / `obi.goversion` |
| `version` / `revision` | `obi.version` / `obi.revision` (reuse the existing resource attributes) |

`bpf.probe.{id,type,name}`, `bpf.map.{id,type,name}`, `telemetry.type` and
`otel.metric.overflow` were already namespaced and are unchanged.

Note: mapping `error_type` → the semconv **`error.type`** attribute is safe —
`schemas/obi/groups/error.yaml` re-declares `error.type` as an open-ended string,
so OBI's domain-specific values (e.g. `attaching_uprobe`) validate cleanly rather
than raising an `undefined_enum_variant` advisory.

## Live coverage — 14 / 16 metrics

Each metric below is emitted in a real run and enforce-validated by weaver, with
its presence asserted from the collector's Prometheus export.

| Metric | Suite |
|---|---|
| `obi.internal.build.info` | dedicated (`TestInternalOTelMetrics`) |
| `obi.instrumented.processes` | dedicated |
| `obi.otel.metric.exports` | dedicated |
| `obi.bpf.map.entries_total` | dedicated |
| `obi.bpf.map.max_entries_total` | dedicated |
| `obi.bpf.probe.executions` | dedicated (needs kernel BPF run-stats + sustained traffic) |
| `obi.bpf.probe.latency_seconds_total` | dedicated |
| `obi.instrumentation.errors` | dedicated (validated when present, not presence-gated) |
| `obi.otel.trace.exports` | traces (`TestInternalOTelMetricsTraces`) |
| `obi.ebpf.tracer.flushes` | traces |
| `obi.bpf.network.packets.total` | netolly (`TestNetwork_Deduplication`) |
| `obi.bpf.network.ignored.packets.total` | netolly |
| `obi.avoided.services` | avoided-services (`TestAvoidedServicesOTelMetrics`) |
| `obi.otel.trace.export.errors` | dedicated (`TestInternalOTelMetricsTraceErrors`) |

## Not live-validated — 2 / 16, and why

All three are **declared** in the registry (the contract is complete; declaring
unobserved metrics is safe — weaver only raises advisories on samples it sees),
but no suite exercises them live yet.

| Metric | Why omitted |
|---|---|
| `obi.otel.metric.export.errors` | Architecturally hard: internal metrics ride the **same** OTLP metrics endpoint that would have to be dead for the error to fire — so the error metric could not itself reach weaver. Needs a two-collector arrangement or the Prometheus internal exporter as the observation channel. |
| `obi.kube.cache.forward.lag` | Blocked by https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2571 |

## Removed: `obi.queue.capacity.ratio` (never emitted)

Investigation showed this metric is emitted by **nothing**: the queue-utilization
gauge is only wired when a queue is built with `msg.InternalMetrics()`, which has
**zero production callers** (agent or kube-cache). It was therefore dropped from
the registry and the emitter left as on `main` — declaring a never-emitted metric
would violate the observe-first rule. Re-add it if the pipeline queues are ever
instrumented.

## Verification

- `make lint-schema` green (weaver `registry check --future`, which promotes
  missing-examples to errors).
- Each suite above passes with `0 actionable` weaver advisories and its expected
  internal metrics present in the report.

## Validation


MrAlias (commented):
The schema coverage needs tests that prove the relevant metrics are emitted; Weaver only validates the data that reaches it.

internal/test/integration/avoided_services_otel_test.go
```diff
@@ -0,0 +1,67 @@
...
+					_ = pokeAvoidedServer()
+					time.Sleep(50 * time.Millisecond)
+				}
+			}
+		}()
+		defer close(stop)
+
+		// Give OBI time to discover the process, observe an OTLP export, and
+		// record the avoided-services metric before weaver is stopped.
+		time.Sleep(20 * time.Second)
```
MrAlias: This only proves that the testserver accepts requests. `runWeaverValidation` validates samples it receives, but it does not require `obi.avoided.services` to appear. A regression that stops OBI from classifying or exporting avoided services therefore leaves the report empty and this test still passes.

Could we fan this collector out to Prometheus and `Eventually` assert `obi_avoided_services`, as `TestInternalOTelMetrics` does for the other internal metrics?

internal/test/integration/internal_metrics_otel_test.go
```diff
@@ -0,0 +1,113 @@
...
+// underscores, monotonic sums get a _total suffix). The eBPF probe families
+// require kernel BPF run-stats plus sustained traffic (driven below).
+var internalMetricsExpected = []string{
+	"obi_internal_build_info",             // gauge, emitted once at startup
+	"obi_instrumented_processes",          // updowncounter, +1 per instrumented process
+	"obi_otel_metric_exports_total",       // counter, every metric export
+	"obi_bpf_map_entries_total",           // gauge, per eBPF map
+	"obi_bpf_map_max_entries_total",       // gauge, per eBPF map
+	"obi_bpf_probe_executions_total",      // counter, needs BPF run-stats + traffic
+	"obi_bpf_probe_latency_seconds_total", // counter, needs BPF run-stats + traffic
```
MrAlias, line 35: `obi.instrumentation.errors` is declared as part of this internal surface, but this suite neither emits an instrumentation error nor requires the series in `internalMetricsExpected`. Weaver only validates received datapoints, so a broken `process.executable.name` or `error.type` mapping is invisible here.

Please either add a deterministic failure and assert this series, or remove it from the live-validated coverage claim.

internal/test/integration/internal_metrics_otel_test.go
```diff
@@ -0,0 +1,113 @@
...
+		// between two consecutive scrapes for obi.bpf.probe.* to report) and the
+		// trace/metric exporters keep firing.
+		stop := make(chan struct{})
+		go func() {
+			for {
+				select {
+				case <-stop:
+					return
+				default:
+					_ = pokeInternalMetricsServer()
```
MrAlias, line 72: Could we rate-limit this loop? `pokeInternalMetricsServer` sends `FORCE_GC`, so it triggers a full GC as fast as a TCP round trip completes. On slower runners that can monopolize a core until the metric appears. The other traffic loops already use 10ms or 50ms pacing.

```suggestion
					_ = pokeInternalMetricsServer()
					time.Sleep(10 * time.Millisecond)
```

grcevski:
needs merge with main.

NimrodAvni78:
@MrAlias I am planning a more robust solution to actually enforce coverage of the whole schema registry we declare.
my plan is to narrow down what we inherit from the official semconv to include only the signals that we declare to produce
and in the end of the full integration + oats + k8s suite tests aggregate all the results and make sure we fully cover our our telemetry schema with live checks

All this to say that I can add validation here that these metrics are present but this is still an wide issue (actually prove we validate all of the telemetry schema) and i would prefer to handle it all together in a next PR
WDYT?

Edit:
looking at these again I think asserting their presense here is good either way to validate they exist so i'll be resolving those soon
the plan to validate coverage still remains

mariomac (approved):
LGTM!

mmat11 (approved):

internal/test/integration/suites_network_test.go
```diff
@@ -33,6 +33,8 @@ func TestNetwork_Deduplication(t *testing.T) {
 		require.Contains(t, f.Metric, "iface_direction")
 	}
 
+	// The network feature drives OBI's internal eBPF packet counters
+	// (obi.bpf.network.*), which are semconv-validated by weaver here.
```
MrAlias: Both packet metrics appear in the current Weaver report, but this test does not require them. `getNetFlows` supplies `obi.network.flow.bytes` samples, so either counter could disappear while `runWeaverValidation` still passes.

Could we assert both series through Prometheus before stopping Weaver, as the dedicated internal-metrics suites do?

MrAlias (commented):
The local presence checks now cover the dedicated suites; the two network counters are the remaining exception.

internal/test/integration/suites_network_test.go
NimrodAvni78: done

PR #2725, opened 2026-07-20
Fix java 8 and unit test failure

grcevski:
## Summary

Updating java agent jar and fixing unit test

## Validation


collector/config_linux_test.go
```diff
@@ -43,8 +43,20 @@ func TestReceiverConfigUnmarshalV2(t *testing.T) {
 	require.NoError(t, component.Unmarshal(cfg))
 	require.NoError(t, cfg.Validate())
 	require.Equal(t, 123, cfg.runtime.ChannelBufferLen)
-	require.Len(t, cfg.runtime.Discovery.Instrument, 1)
-	require.Equal(t, []int{8080}, cfg.runtime.Discovery.Instrument[0].OpenPorts.AllValues())
+	require.Len(t, cfg.runtime.Discovery.Instrument, 2)
```
grcevski, line 46: @MrAlias so I made the test pass here, but now I'm wondering if this is the right thing to do with the catch all case for config v2. What should happens when we have just include with port 8080, do we expect to have a default selector that matches all executables?

MrAlias: Yes. `default_action` is `include` when omitted, so an `include` rule for port 8080 is not an allow-list: unmatched executables must still be selected by the catch-all selector.

To instrument only port 8080, set `capture.policy.default_action: exclude`. Keep the assertion for both selectors; optionally make `default_action: include` explicit in this test.

MrAlias (approved):

collector/config_linux_test.go
grcevski: yes ok makes sense.

Issue #2577, opened 2026-07-03
Unbounded trace merging with `context_propagation: headers` on pure-Go services — what is the recommended way to bound this?


nishant-sharechat:
## Summary

With header-based context propagation enabled, OBI produces unbounded "mega-traces"
for us: thousands of unrelated server requests stitched under one trace id, growing
for minutes to hours, which makes distributed tracing unusable during those windows.

We previously reported a Node.js variant (#2551), which was closed as
working-as-expected with the guidance to re-enable Node.js support. This report shows
the same unbounded merging with **pure-Go originators with fully-attached Go
uprobes** — no Node.js service anywhere in the offending path — so per-request
runtime context does not appear to resolve it.

We're aware this is not the only production-grade report against the trace-context
injection machinery — see grafana/beyla#2941 (intermittent 20–80s send-stalls on
nginx from the sk_msg/sockmap injection path). Between that and what we describe
below, we'd like to understand what the supported/recommended configuration is,
and **what we can do** to keep distributed tracing usable at this scale.

## Context — our setup and architecture

We run OBI specifically **for distributed tracing** over a large multi-runtime
mesh (Go, Java, Node.js services behind pooled keep-alive connections,
gRPC/HTTP2, ingress proxies and mesh waypoints). Metrics and per-service spans
work well for us — this issue is only about the stitching behavior when
propagation is enabled.

**Zero-code instrumentation is the point of our adoption**: hundreds of services
owned by many teams, and we deliberately chose eBPF-level instrumentation so
that application code does not need SDK integration. "Instrument the
applications instead" is therefore not an outcome that works for us — we want
OBI to be the instrumentation layer, which is why we're motivated to help make
its propagation robust rather than fall back to app-level tracing.

Telemetry architecture (per production cluster):

```
 [apps on every node]
        │  (eBPF instrumentation)
 OBI/Beyla DaemonSet
        │  OTLP gRPC (node-local)
 OTel Collector agent DaemonSet
        │  produce (otlp_proto)
 Kafka-compatible queue (topic per signal)
        │  consume (single consumer group)
 OTel Collector "traces consumer" deployment
        │  tail_sampling → attribute-propagation → batch
 ClickHouse (traces store)
```

Notes relevant to the numbers below:

- We verified the post-OBI pipeline does not create the merging: exact-duplicate
  rate end-to-end is ~0.1–0.2% (span ids are unique), and synthetic tests with
[...]

PR #2726, opened 2026-07-20
Break forever go transactions

grcevski:
## Summary

The generic tracer has logic that prevents forever transactions where one server operation can post many client operations in a loop. I'm implementing the same support in the Go tracer.

Relates to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2577

The above issue still needs investigation because the user claims that it was a single client only loop that caused the problem, no server request.

## Validation


bpf/gotracer/go_common.h
```diff
@@ -338,8 +341,15 @@ static __always_inline u8 client_trace_parent(void *goroutine_addr, tp_info_t *t
         tp_info_t *tp = tp_info_from_parent_go(&g_key, 0);
 
         if (tp) {
-            tp_from_parent(tp_i, tp);
-        } else {
+            if (should_be_in_same_transaction(tp, tp_i)) {
```
MrAlias, line 344: Manual Go SDK spans stored in `go_trace_map` have `ts == 0`: `go_sdk.c` builds them with `tp_from_parent`, which does not copy the timestamp. A client call inside an active SDK span will therefore fail this comparison and begin a new trace.

Set the timestamp when the SDK span is stored, or handle that parent form before applying the transaction-age check.

bpf/gotracer/go_common.h
```diff
@@ -338,8 +341,15 @@ static __always_inline u8 client_trace_parent(void *goroutine_addr, tp_info_t *t
...
+                tp_from_parent(tp_i, tp);
+                found_trace_id = 1;
```
MrAlias, line 346: Setting `found_trace_id` here skips the span-ID generation at line 356. `tp_from_parent` only fills trace ID, parent ID, and flags, so a normal Go client call under a valid parent retains its zero-initialized span ID and injects an invalid traceparent.

Keep parent selection separate from the SQL reuse path, or generate the child span ID after selecting the Go parent.

internal/test/integration/traces_test.go
```diff
@@ -1856,3 +1856,113 @@ func testGoGenericHTTPSTraces(t *testing.T) {
...
+		if ok {
+			if p.TraceID == server.TraceID && p.SpanID == server.SpanID {
+				processing = r
+				break
+			}
+		}
+	}
+
+	require.NotNil(t, processing)
+	sd = queue.Diff(
```
MrAlias: This looks up `processing`, but validates `queue` again, so the processing-span assertion is never exercised.

```suggestion
	sd = processing.Diff(
```

mariomac (approved):
LGTM! I wonder if at some point is worth providing an alternative `OTEL_EBPF_BPF_MAX_TRANSACTION_LEN` that would break traces by number of sub-spans instead of by duration. Thinking on long-living jobs.

bpf/gotracer/go_common.h
grcevski: Good catch, the one place that's not using this helper.

bpf/gotracer/go_common.h
grcevski: I think the code is correct, the only thing I skip is  `urand_bytes(tp_i->trace_id, TRACE_ID_SIZE_BYTES);`, which initialising a new traceId, the `urand_bytes(tp_i->span_id, SPAN_ID_SIZE_BYTES);` generation is unguarded.

internal/test/integration/traces_test.go
grcevski: Nice catch.

grcevski:
> LGTM! I wonder if at some point is worth providing an alternative `OTEL_EBPF_BPF_MAX_TRANSACTION_LEN` that would break traces by number of sub-spans instead of by duration. Thinking on long-living jobs.

Yeah that would be nice! I need to think if it would be possible, maybe in userspace and then somehow tell the bpf side to avoid correlating.

rafaelroquetto (approved):

mmat11 (approved):
lgtm

Issue #2706, opened 2026-07-19
v0.10.0 sk_msg HPACK context propagation corrupts multiplexed (yamux) gRPC connections - breaks GitLab Gitaly


christensenjairus:
### Component

OBI (eBPF instrumentation)

### What happened?

After upgrading OBI **v0.9.x → v0.10.0**, all GitLab **Gitaly** traffic began failing
intermittently. `git pull` mostly worked (short streams, succeed on retry) but `git push`
failed consistently with:

    remote: rpc error: code = Unavailable desc = The git server, Gitaly, is not
    available at this time. Please contact your administrator.

The onset correlated exactly with the OBI upgrade, and reverting OBI to the pre-v0.10.0
image restored service immediately.

## Root cause

v0.10.0 added the feature:

> "Added gRPC/HTTP2 context propagation through `sk_msg` HPACK injection."

Gitaly's gRPC connections (port 8075) are **not plain HTTP/2**. Gitaly multiplexes a **yamux backchannel** (used by Praefect↔Gitaly for reference-transaction voting on writes) and **sidechannels** (bulk packfile transfer) over the same sockets. OBI's protocol detector classifies this traffic as gRPC/HTTP2 and the `sk_msg` program injects HPACK-encoded `traceparent` bytes into the stream. That shifts the yamux framing, and Gitaly reads a garbage protocol-version byte.

The corruption is visible directly in Gitaly's logs (byte `47` = ASCII `/`, i.e. foreign bytes where the yamux version byte `0` is expected):

    {"component":"backchannel.YamuxServer","level":"info","msg":"[ERR] yamux: Invalid protocol version: 47"}
    {"level":"warning","msg":"[core] grpc: Server.processUnaryRPC failed to write status:connection error: desc = \"transport is closing\""}

Client side (gitlab-workhorse → Praefect/Gitaly), the resets surface as:

    rpc error: code = Unavailable desc = error reading from server: EOF handleUploadPack: ... PostUploadPackWithSidechannel: max retries exhausted: failed after 4 attempts

Writes fail hard because a push is one long backchannel-coordinated stream — any mid-stream reset aborts it. Reads sometimes survive on retry.

### What did you expect to happen?

OBI should not inject HPACK/`traceparent` bytes into connections that are not genuine, unframed HTTP/2 — or should detect yamux/connection-multiplexed streams and skip injection — so that instrumenting a service never corrupts its wire protocol.

### Steps to reproduce

[...]

PR #2734, opened 2026-07-21
fix(tpinjector): only inject HPACK traceparent into genuine HTTP/2

mariomac:
## Summary

Fixes issue #2706.

With `context_propagation: all`, OBI's `sk_msg` HPACK traceparent injection corrupts connections that tunnel a non-HTTP/2 protocol over what OBI classifies as HTTP/2 — notably GitLab Gitaly's yamux-multiplexed backchannel, which fails with `yamux: Invalid protocol version` and breaks `git push`.

To fix it, we now require the mandatory post-preface SETTINGS frame before injecting. Every [HTTP/2-compliant stack MUST send it](https://datatracker.ietf.org/doc/html/rfc7540#section-3.5):

> The server connection preface consists of a potentially empty SETTINGS frame ([Section 6.5](https://datatracker.ietf.org/doc/html/rfc7540#section-6.5)) that MUST be the first frame the server sends in the HTTP/2 connection.

The per-socket flag (`sk_h2_conn_flag`) becomes a small state machine: `none → preface → confirmed | rejected`. Only `confirmed` sockets are injected into. Rejection happens only on a cleanly-parsed non-SETTINGS first frame; on any uncertainty (short packet, unparseable frame) the socket stays `preface` and waits, so genuine connections are never rejected by a transient hiccup. This also overrides the generic tracer's preface-only signal (`already_tracked_plain_http2`), which had the same weakness.

Injection only ever runs on active (client) egress, where preface+SETTINGS is guaranteed — server sockets (bare SETTINGS, no `PRI`) are unaffected.

The validity of the solution has been tested on a synthetic service/client using Hashicorp's YAMUX: `TestSuite_YamuxGRPC`.

The first commit of this PR reproduces the issue, so `TestSuite_YamuxGRPC` initially fails. The second commit fixes it and makes the test pass.

## Validation

### DIFF
--- a/internal/test/integration/yamux_grpc_test.go
+++ b/internal/test/integration/yamux_grpc_test.go
@@ -18,11 +18,11 @@ import (
 	"go.opentelemetry.io/obi/internal/test/integration/components/jaeger"
 )
 
-// yamuxStats mirrors the /stats payload exposed by the yamux_h2_corrupt component.
+// yamuxStats mirrors the /stats payload exposed by the yamux component.
 type yamuxStats struct {
-	Successes  int64 `json:"successes"`
-	Failures   int64 `json:"failures"`
-	Corruption int64 `json:"corruption"`
+	Successes  int `json:"successes"`
+	Failures   int `json:"failures"`
+	Corruption int `json:"corruption"`
 }
 
 func getYamuxStats(t require.TestingT, url string) yamuxStats {
@@ -86,9 +86,7 @@ func TestSuite_YamuxGRPC(t *testing.T) {
 
 	const (
 		clientStats = "http://localhost:8081/stats"
-		clientReset = "http://localhost:8081/reset"
 		serverStats = "http://localhost:8080/stats"
-		serverReset = "http://localhost:8080/reset"
 	)
 
 	// Both roles must be up: /health returns 200 once each HTTP status server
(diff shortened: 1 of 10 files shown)


internal/test/integration/yamux_grpc_test.go
```diff
@@ -0,0 +1,157 @@
...
+		var tq jaeger.TracesQuery
+		require.NoError(ct, json.NewDecoder(r.Body).Decode(&tq))
+		require.NotEmpty(ct, tq.Data, "yamux-client not yet instrumented by OBI")
+	}, 3*time.Minute, time.Second)
+	t.Log("yamux-client instrumented; sk_msg propagation path is now active")
+
+	// Give OBI a moment to also attach to the server and settle, then zero both
+	// components' counters so the assertion window only measures traffic that
+	// flows while instrumentation is fully live.
+	// TODO: replace by eventually
```
mariomac: Before sending the PR. I forgot to find async alternatives for the `time.Sleep` below as well as the `for time.Now().Before(deadline)`. I'll amend this PR to fix this before merging, to avoid that a single test case takes 25 seconds.

rafaelroquetto (approved):
LGTM

grcevski (approved):
LGTM!

PR #2735, opened 2026-07-21
Omit empty http method from http metrics

NimrodAvni78:
## Summary

Fixes the OATS weaver semantic-convention check that was failing on
`run-oats_java_tls_sync_async` (the reason the OATS weaver validation was
temporarily disabled in #2727), and re-enables that validation.

OBI was emitting the enumerated attribute `http.request.method` with an empty
value (`""`) on client-side HTTP metrics (`http.client.request.body.size`,
`http.client.request.duration`, `http.client.response.body.size`). Weaver
classifies an empty enum value as `undefined_enum_variant`, and the OATS weaver
gate treats it as actionable — the check reported `4 actionable` advisories and
failed with "Should be zero, but was 4".

### Root cause

For TLS / async client traffic, connection info is often unavailable, so span
construction falls back to `httpRequestToSpan`, which derives the method from a
raw buffer via `httpMethodFromBuf`. When the buffer holds no
`METHOD SP URL SP VERSION` request line, that function returns `""`. Nothing
downstream dropped the attribute: the metrics getter passed the value through,
and the expirer's only omit-gate (`attribute.KeyValue.Valid()`) considers an
empty string valid. Per the weaver-check contract, when OBI has no valid value
for an enumerated attribute it must **omit** the attribute rather than emit an
empty one.

### Changes

- **Omit `http.request.method` when empty (metrics):** the OTEL method getter
  now returns an invalid `attribute.KeyValue{}` for an empty method, so the
  expirer's existing `Valid()` gate drops it — the established skip-idiom in
  `span_getters.go`.
- **Omit `http.request.method` when empty (traces):** the HTTP server and
  client span builders now append the method only when non-empty, matching the
  existing `if span.Method != ""` guard already used for DB / messaging spans.
- **Surface failing advisories first:** the weaver report now prints an
  "actionable advisories (these fail the check, most frequent first)" block
  before the full advisory dump, so the cause of a failure is visible at a
  glance instead of buried among hundreds of ignored/informational lines.
  Ordering is deterministic (by occurrence count, then message).
- **Re-enable the OATS weaver validation** by reverting the temporary
  `|| true` short-circuit added in #2727, now that the root cause is fixed.
- Regression tests: `http.request.method` is present for a known method and
  absent for an empty one, on both client and server spans, and at the
  metrics-getter level.

### Example: new weaver output ordering

```
  actionable advisories:
    [information/undefined_enum_variant] [4x] Enum attribute 'http.request.method' has value '' which is not documented. (signals: metric:http.client.request.body.size, metric:http.client.request.duration, metric:http.client.response.body.size)
  advisory details:
    ... (full per-level list unchanged, for context)
```

## Validation

### DIFF
--- a/internal/test/oats/harness/weaver.go
+++ b/internal/test/oats/harness/weaver.go
@@ -38,7 +38,7 @@ const (
 // shared weaver compose fragment), unless the run explicitly opts out via
 // TESTCASE_SKIP_WEAVER=true.
 func validateWeaver() {
-	if os.Getenv(skipWeaverEnv) == "true" || true {
+	if os.Getenv(skipWeaverEnv) == "true" {
 		ginkgo.GinkgoWriter.Printf("%s=true — skipping weaver validation\n", skipWeaverEnv)
 		return
 	}
--- a/internal/test/weavercheck/weavercheck.go
+++ b/internal/test/weavercheck/weavercheck.go
@@ -255,7 +255,7 @@ func Validate(t TestingT, report *Report) {
 		actionableAdvisories += e.Occurrences
 	}
 	if len(actionable) > 0 {
-		t.Logf("  actionable advisories (these fail the check, most frequent first):")
+		t.Logf("  actionable advisories:")
 		for _, e := range actionable {
 			signals := "unknown"
 			if len(e.Signals) > 0 {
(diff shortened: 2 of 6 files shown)


mariomac (approved):
Good catch!

grcevski (approved):
LGTM!

grcevski:
I need to look into those empty methods, this was a good catch.

PR #2736, opened 2026-07-21
Fix double-traceparent injection in programs using the Go OTEL SDK

### LINKED ISSUE #2732: Go uprobe appends a duplicate traceparent to SDK-instrumented HEADERS, breaking downstream traces
### Component

OBI (eBPF instrumentation)

### What happened?

When OBI is attached to a Go service that is **already instrumented with the OpenTelemetry Go SDK** (`otelgrpc`) — i.e. the app itself writes a valid `traceparent` into its outgoing HTTP/2 HEADERS (HPACK) — OBI's Go-specific uprobe **appends its own `traceparent` to the same frame without checking whether one already exists**, and rewrites the HTTP/2 frame length field. The frame ends up carrying two `traceparent` headers and a length that no longer matches what the application wrote.

Downstream, this breaks the trace in one of two ways:

- **New root** — the receiver fails to extract the incoming `traceparent`, so `ParentSpanId` is empty and it starts a new root.
- **Phantom orphan** — the receiver adopts a parent span-id that does not exist anywhere (the span-id OBI generated for itself).

This happens with `context_propagation` **enabled** and is distinct from #2276 (see *Additional context*).

It occurs on the **sending** side, for an SDK-instrumented Go service. The representative case is a grpc-go client instrumented with `otelgrpc` (the same unguarded-write pattern also exists in the Go `net/http` injection path).

**Observed on the standard OpenTelemetry Demo** (sender = the demo's `checkout` service, Go + grpc-go + `otelgrpc`; steps below):

- **Before OBI (baseline):** across 3 `user_checkout_multi` traces, 0 orphans — every downstream server span of `checkout` connects to its correct parent.
Image

- **With OBI deployed (`checkout` instrumented):** immediately after attach, in all 4 traces inspected, `checkout`-originated downstream breaks:
  - **cart / payment / email server spans disappear from the trace and start new roots** — aggregated over the window: cart `EmptyCart` **17/17 = 100% as root**, payment `Charge` **17/17 as root**, email **17/18 as root** (all checkout-originated calls). 
  - **shipping server spans become phantom orphans** — the parent span-id of `/get-quote` / `/ship-order` does not exist anywhere in the trace, while `checkout`'s actual client span for the same call is present under a different id — the `traceparent` was already corrupted on the wire at send time.
  - Control within the same deployment: the same `cart` service stays correctly connected for **frontend (Node.js)-originated `AddItem` calls (53/53)** — the only variable that determines breakage is whether the sender is an OBI-attached Go service.
Image

- **After excluding `checkout` via `discovery.exclude_instrument`:** everything reconnects immediately — 0 orphans; cart `EmptyCart` 0/6 as root, payment 0/6, email 0/7. (This is our current workaround, but it gives up OBI observability for that service entirely, so it is not a real fix.)

When an application already writes its own `traceparent` (e.g. via the OpenTelemetry SDK), OBI should not add a second one — the SDK-owned `traceparent` must reach the receiver intact.

### Steps to reproduce

1. Deploy the standard OpenTelemetry Demo helm chart (v0.40.7, appVersion 2.2.0) on a single Kubernetes cluster.
2. As a baseline, check the `ParentSpanId` linkage of `checkout`'s downstream server spans in `user_checkout_multi` traces without OBI.
3. Deploy OBI (v0.10.0) as a DaemonSet with `context_propagation: all` and `discovery.instrument` covering the demo namespace (so `checkout` is instrumented) — see config below.
4. Compare the `ParentSpanId` of the checkout-originated downstream server spans (cart / payment / email / shipping) in `user_checkout_multi` traces against the baseline.
5. (Recovery check) Add `checkout` to `discovery.exclude_instrument`, redeploy, and repeat step 4.

### OBI configuration

```yaml
ebpf:
  context_propagation: all
discovery:
  instrument:
    - k8s_namespace: <demo-namespace>   # includes the SDK-instrumented Go sender "checkout"
  # Only for reproduction step 5 (recovery check):
  # exclude_instrument:
  #   - k8s_deployment_name: checkout
  #     k8s_namespace: <demo-namespace>
```

### Relevant logs

```shell

```

### Additional context

**Root cause (code, v0.10.0)**

OBI has two `traceparent` injection paths, and **the guard against a pre-existing header exists in only one of them.**

- **Go uprobe path — no guard.** gRPC: the injection in the `grpcFramerWriteHeaders` return uprobe is gated **only by available buffer space** ([`bpf/gotracer/go_grpc.c:909`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/v0.10.0/bpf/gotracer/go_grpc.c#L909)); it never checks whether a `traceparent` is already present. It always appends and rewrites the 3-byte HTTP/2 frame length ([`go_grpc.c:965`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/v0.10.0/bpf/gotracer/go_grpc.c#L965)). HTTP: same pattern — space check only ([`go_nethttp.c:882`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/v0.10.0/bpf/gotracer/go_nethttp.c#L882)), write without an existing-header check ([`go_nethttp.c:886`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/v0.10.0/bpf/gotracer/go_nethttp.c#L886)).
- **sk_msg path — has a guard.** The sk_msg path used for non-Go traffic checks for an existing `traceparent` via `find_existing_h2_tp` before injecting, and adopts it instead of adding a second one ([`bpf/tpinjector/tpinjector.c:78`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/v0.10.0/bpf/tpinjector/tpinjector.c#L78)).
- **Why only Go:** writing directly into the application's user-space buffer with `bpf_probe_write_user` requires knowing the app's internal offsets, so it is done only for Go binaries. Once a Go gRPC connection is marked as handled by the uprobe (`is_go_grpc_client_conn`), the sk_msg path deliberately bails ([`tpinjector.c:57`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/v0.10.0/bpf/tpinjector/tpinjector.c#L57)). So the unguarded path is the uprobe path, and that path is Go-only.
- **Where the phantom parent comes from:** the span-id in the `traceparent` OBI appends is a fresh span-id OBI generates for itself via `urand_bytes` in `client_trace_parent` ([`bpf/gotracer/go_common.h:318`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/v0.10.0/bpf/gotracer/go_common.h#L318), generated at [`:346`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/v0.10.0/bpf/gotracer/go_common.h#L346)). It differs from the SDK client span's span-id, so a receiver that adopts it ends up with a parent that was never exported — a phantom orphan.

**Relationship to #2276**

#2276 is about the Go path not honoring `context_propagation: disabled`, and is already being worked on. **This issue occurs with propagation enabled and remains even after #2276 is fixed** — coexisting with an SDK-instrumented service while propagation is on requires not "turn injection off" but "don't inject if a `traceparent` is already present." It's also a concrete, reproducible instance of the additive SDK/OBI coexistence goal tracked in #1140.

**Possible directions**

There seem to be a few directions here, and we'd appreciate the maintainers' input.

 - **A. Skip the Go-uprobe injection for processes that do their own SDK tracing (per-PID).**
OBI already attaches uprobes to the app's OTel SDK tracer and checks in-kernel whether a real SDK delegate is installed ([`bpf/gotracer/go_sdk.c:113`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/v0.10.0/bpf/gotracer/go_sdk.c#L113)). We could reuse that as a per-PID flag: right before injecting ([`go_grpc.c`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/v0.10.0/bpf/gotracer/go_grpc.c#L855) / [`go_nethttp.c`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/v0.10.0/bpf/gotracer/go_nethttp.c#L809)), if the PID is marked as "the SDK writes its own `traceparent`", skip the injection.
Pushing a per-PID flag down to the kernel already has a precedent in `valid_pids` ([`bpf/pid/maps/valid_pids.h`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/v0.10.0/bpf/pid/maps/valid_pids.h)), and since these probes compile into the same BPF object it can be done entirely kernel-side. It also fails safe — OBI simply doesn't add its own `traceparent`, while the app's SDK `traceparent` still reaches the receiver intact.
`bpf_probe_write_user` is discussed as effectively DOA in go-instrumentation [#290](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/290), but it still has to be kept for the TLS path, so this uprobe injection path won't go away either — so would it make sense to fix this via A (skip injection for SDK-instrumented processes)?

 - **B. Check the headers for an existing `traceparent` before injecting (the sk_msg approach, [`find_existing_h2_tp`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/v0.10.0/bpf/tpinjector/tpinjector.c#L1359)).**
The catch is that on reused HTTP/2 connections HPACK sends the `traceparent` name literally only on the first request and references it by dynamic-table index afterwards, so scanning the wire won't find it. Because of exactly this limitation, the sk_msg path's own HTTP/2 duplicate check is also best-effort on reused connections (it only catches the first request), and a reliable duplicate check really only holds for HTTP/1.1. So B alone isn't enough.

A looks like the better option to us, but we'd like to know if there's a better approach we've missed.

### Version

v0.10.0

### Kernel version

6.8.0-90-generic

### OS / Distribution

Ubuntu 24.04

### Deployment

Kubernetes

mariomac:
## Summary

Fixes #2732 .

For Go services that were already instrumented with the Go OTEL SDK, we were filtering out the metrics/traces at the userspace level. This was not preventing our kernel-side eBPF Go tracers to keep injecting the traceparent. This resulted on duplicate traceparent injection.

This PR maintains a map with the PIDs of these Go processes that should be injecting its own Traceparent, so we early abort any kernel-side operation regarding traceparent injection.

## Validation

### DIFF
--- a/bpf/gotracer/go_nethttp.c
+++ b/bpf/gotracer/go_nethttp.c
@@ -817,6 +817,13 @@ int obi_uprobe_writeSubset(struct pt_regs *ctx) {
         return 0;
     }
 
+    // The application drives its own OTel SDK traceparent; injecting ours would
+    // append a duplicate traceparent and break downstream traces.
+    if (pid_writes_own_tp()) {
+        bpf_dbg_printk("skipping http tp injection, pid writes its own traceparent");
+        return 0;
+    }
+
     bpf_dbg_printk("=== uprobe/header_writeSubset ===");
 
     void *header_addr = GO_PARAM1(ctx);
@@ -1122,6 +1129,13 @@ on_http2FramerWriteHeaders(struct pt_regs *ctx, off_table_t *ot, u64 stream_id) 
         return;
     }
 
+    // The application drives its own OTel SDK traceparent; the return uprobe
+    // would skip injection anyway, so don't stash the invocation state for it.
+    // Unlike the gRPC path, this uprobe has no other side effects to preserve.
+    if (pid_writes_own_tp()) {
+        return;
+    }
+
     void *framer = GO_PARAM1(ctx);
 
     if (!framer) {
@@ -1231,6 +1245,13 @@ int obi_uprobe_net_http2FramerWriteHeaders(struct pt_regs *ctx) {
 SEC("uprobe/http2FramerWriteHeaders_returns")
 int obi_uprobe_http2FramerWriteHeaders_returns(struct pt_regs *ctx) {
     if (!g_bpf_header_propagation) {
+        return 0;
+    }
+
+    // The application drives its own OTel SDK traceparent; injecting ours would
+    // append a duplicate traceparent and break downstream traces.
+    if (pid_writes_own_tp()) {
+        bpf_dbg_printk("skipping http2 tp injection, pid writes its own traceparent");
         return 0;
     }
 
--- a/bpf/gotracer/go_common.h
+++ b/bpf/gotracer/go_common.h
@@ -107,6 +107,37 @@ struct {
     __uint(max_entries, MAX_CONCURRENT_REQUESTS);
 } ongoing_grpc_transports SEC(".maps");
 
+// go_sdk_tp_pids tracks the PIDs (host tgid) whose OpenTelemetry Go SDK writes
+// its own traceparent into outgoing HTTP/gRPC headers. For these processes OBI
+// must not inject a second traceparent via its uprobes,
+// otherwise the frame ends up carrying two traceparent headers and downstream
+// traces break. The flag is set from the SDK delegate check in go_sdk.c.
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, u32);  // host tgid
+    __type(value, u8); // presence flag
+    __uint(max_entries, MAX_CONCURRENT_REQUESTS);
+    __uint(pinning, OBI_PIN_INTERNAL);
+} go_sdk_tp_pids SEC(".maps");
+
+// Records that the current process drives its own OTel SDK traceparent, so OBI
+// must skip its uprobe-based traceparent injection for it.
+static __always_inline void mark_pid_writes_own_tp() {
+    const u32 tgid = (u32)(bpf_get_current_pid_tgid() >> 32);
+    const u8 one = 1;
+    bpf_map_update_elem(&go_sdk_tp_pids, &tgid, &one, BPF_ANY);
+}
+
+// Returns true when the current process drives its own OTel SDK traceparent
+// injection and OBI should therefore skip its own injection.
+static __always_inline bool pid_writes_own_tp() {
+    if (!g_bpf_traceparent_enabled) {
+        return false;
+    }
+    const u32 tgid = (u32)(bpf_get_current_pid_tgid() >> 32);
+    return bpf_map_lookup_elem(&go_sdk_tp_pids, &tgid) != NULL;
+}
+
 #define SQL_CONN_TYPE_DATABASE_SQL 0 // database/sql (mysql, pq)
 #define SQL_CONN_TYPE_PGX 1          // github.com/jackc/pgx/v5
 
--- a/bpf/gotracer/go_grpc.c
+++ b/bpf/gotracer/go_grpc.c
@@ -828,7 +828,13 @@ int obi_uprobe_grpcFramerWriteHeaders(struct pt_regs *ctx) {
         // The offset will be 0 on first connection through the stream and 9 on subsequent.
         // If we read some very large offset, we don't do anything since it might be a situation
         // we can't handle
-        if (offset < MAX_W_PTR_OFFSET) {
+        if (pid_writes_own_tp()) {
+            // The application drives its own OTel SDK traceparent, so the return
+            // uprobe will skip injection: don't stash the invocation for it. The
+            // connection is already marked above so the sk_msg path also leaves
+            // the app's traceparent intact.
+            bpf_dbg_printk("skipping grpc framer stash, pid writes its own traceparent");
+        } else if (offset < MAX_W_PTR_OFFSET) {
             grpc_framer_func_invocation_t f_info = {
                 .tp = invocation->tp,
                 .framer_ptr = (u64)framer,
@@ -858,6 +864,13 @@ int obi_uprobe_grpcFramerWriteHeaders_returns(struct pt_regs *ctx) {
     }
 
     bpf_dbg_printk("=== uprobe/grpcFramerWriteHeaders_returns ===");
+
+    // The application drives its own OTel SDK traceparent; injecting ours would
+    // append a duplicate traceparent and break downstream traces
+    if (pid_writes_own_tp()) {
+        bpf_dbg_printk("skipping grpc tp injection, pid writes its own traceparent");
+        return 0;
+    }
 
     void *goroutine_addr = GOROUTINE_PTR(ctx);
     off_table_t *ot = get_offsets_table();
--- a/bpf/gotracer/go_sdk.c
+++ b/bpf/gotracer/go_sdk.c
@@ -111,6 +111,10 @@ static __always_inline int tracer_start(struct pt_regs *ctx, u8 check_delegate) 
             sizeof(delegate_ptr),
             (void *)(tracer_ptr + go_offset_of(ot, (go_offset){.v = _tracer_delegate_pos})));
         if (delegate_ptr != NULL) {
+            // Delegate is set, so the application drives its own OTel SDK
+            // tracing. Mark the PID so the Go uprobe traceparent injection
+            // skips it and we don't append a duplicate traceparent.
+            mark_pid_writes_own_tp();
             // Delegate is set, so we should not instrument this call
             return 0;
         }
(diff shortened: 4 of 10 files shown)


bpf/gotracer/go_nethttp.c
```diff
@@ -817,6 +817,13 @@ int obi_uprobe_writeSubset(struct pt_regs *ctx) {
         return 0;
     }
 
+    // The application drives its own OTel SDK traceparent; injecting ours would
+    // append a duplicate traceparent and break downstream traces.
+    if (pid_writes_own_tp()) {
+        bpf_dbg_printk("skipping http tp injection, pid writes its own traceparent");
+        return 0;
```
MrAlias, line 824: `persistConnRoundTrip` has already populated `outgoing_trace_map` before this probe runs. Returning here suppresses only the Go uprobe; in production the shared `tpinjector` is also loaded, and its HTTP/1 `tp_pid` path consumes that still-valid entry and writes the OBI traceparent directly without checking for an existing header. Plaintext SDK-instrumented requests can therefore still receive the duplicate this PR is meant to prevent.

Could we suppress or clear the socket-message injection metadata on the SDK-owned path as well, and exercise the production Go tracer composition in the regression test? The current test constructs `ProcessTracer` with only `goTracer`, so it cannot catch this fallback.

bpf/gotracer/go_nethttp.c
```diff
@@ -1122,6 +1129,13 @@ on_http2FramerWriteHeaders(struct pt_regs *ctx, off_table_t *ot, u64 stream_id)
         return;
     }
 
+    // The application drives its own OTel SDK traceparent; the return uprobe
+    // would skip injection anyway, so don't stash the invocation state for it.
+    // Unlike the gRPC path, this uprobe has no other side effects to preserve.
+    if (pid_writes_own_tp()) {
+        return;
```
MrAlias, line 1136: `setup_http2_client_conn` has already inserted this stream into `http2_req_map`, but this return runs before the normal deletion at the end of the helper. Every SDK-owned HTTP/2 request therefore leaves stale stream state in the shared LRU map; sustained traffic can evict entries for requests that are still in flight and disable correlation or injection for unrelated traffic.

Could we remove the stream entry before returning, or avoid inserting it once the PID is marked?

MrAlias (commented):
The per-PID guard currently suppresses only part of the propagation pipeline: HTTP/1 can still inject through `tpinjector`, and the HTTP/2 skip path leaks stream state. Could we address both paths before merging?

grcevski (commented):
I like the idea, but I just wanted to call out that it won't be 100% accurate, some requests will still fail. For the detection of already exporting OTel, at the moment we rely on seeing an exported span successfully.

So there are three issues with that approach:
1. Until we see the export, we'll keep on injecting the traceparent.
2. The export must succeed, so unsuccessful export doesn't mark the application as exporting traces.
3. Finally, the most tricky part of all, we can't guarantee the application isn't doing context propagation and not exporting traces. Nothing prevents this with the Go OTel SDK, you can propagate trace context and not configure an exporter, or configure the debug to stdout exporter.

I think more viable approach would be to set a uprobe carefully in the http (and maybe http2) client path and see if there's a traceparent. If there's one we can skip. We already have some code like this for extracting the traceparent from server requests, so we need to add something similar to client.

mariomac:
Thanks @MrAlias and @grcevski  for your reviews! I initially wanted to provide a 20%-effort patch that would work in 80% of (Golang) scenarios, despite as you mention there could be situations where this still don't work. I'm happy to rework it to cover 100% of cases.

Just to double-check I understand this part of the code. Regarding Nikola's first point:

> 1. Until we see the export, we'll keep on injecting the traceparent.

I think this should happen if we only updated the `go_sdk_tp_pids` from the Go side (in the export). But we are also [catching this early in the `start_tracer` BPF probe](https://github.com/mariomac/opentelemetry-ebpf-instrumentation/blob/1a0c101d38b9f14824d23d757aee859cd92f345b/bpf/gotracer/go_sdk.c#L113-L119). Shouldn't be enough to detect that the process uses the Go SDK when it starts, before it tries to send anything?

In that case, I'd keep that implementation in addition to the rest of improvements/fixes we can add on top of it.

grcevski:
> Shouldn't be enough to detect that the process uses the Go SDK when it starts, before it tries to send anything?

I still think my problem description under 3. will happen. A user may write code to propagate or add traceparent themselves, maybe not even using the Go SDK, perhaps something else they've hand rolled. I've seen customer built frameworks for this. The problem is the duplication of traceparent, not actually using the Go SDK or exporting traces.

I think we should explore an option that looks at the headers of a client request, because that's all that matters. Before we are about to inject the header, is there another one there.

mariomac:
Closing. Will resubmit a new PR starting from scratch, taking the approach of detecting the traceparent header.

PR #2740, opened 2026-07-21
Add Go runtime histogram metrics

marctc:
## Summary

Add eBPF collection and export for the following Go runtime histogram metrics:

  - `go.memory.gc.pause.duration`
  - `go.schedule.duration`

Reads cumulative histograms directly from the Go runtime and preserves its native bucket layout when exporting them. Collection is available for supported Go versions and is skipped when the required runtime data cannot be resolved safely.

## Validation


pkg/export/otel/metrics_runtime.go
```diff
@@ -271,11 +275,13 @@ func recordRuntimeMetrics(ctx context.Context, metrics *RuntimeMetrics, snapshot
 		return
 	}
 
-	if snapshot.Go != nil {
-		if snapshot.Service.SDKLanguage != svc.InstrumentableGolang {
-			return
+	if snapshot.Service.SDKLanguage == svc.InstrumentableGolang {
```
grcevski, line 354: Just a thought, should we clear the histogram when a pid terminates? It's unlikely that a process pid will be reused, but it can be.

pkg/export/otel/metrics_runtime_histograms.go
```diff
@@ -0,0 +1,180 @@
...
+	data, err := state.histogram.Data()
+	if err != nil {
+		return metricdata.Metrics{}, fmt.Errorf("converting %s histogram: %w", name, err)
+	}
+
+	return metricdata.Metrics{
+		Name: name,
+		Unit: "s",
+		Data: metricdata.Histogram[float64]{
+			Temporality: metricdata.CumulativeTemporality,
```
grcevski: I thought we have a config option for this, cumulative or delta?

grcevski (approved):
This looks great, I left two comments.

bpf/gotracer/maps/runtime.h
```diff
@@ -107,8 +108,31 @@ typedef enum go_runtime_metric_valid {
     go_runtime_metric_valid_cpu_time = 1 << 4,
     go_runtime_metric_valid_memory_used = 1 << 5,
     go_runtime_metric_valid_memory_allocations = 1 << 6,
+    go_runtime_metric_valid_gc_pause_histogram = 1 << 7,
+    go_runtime_metric_valid_schedule_duration_histogram = 1 << 8,
 } go_runtime_metric_valid_t;
 
+enum : u32 {
+    HIST_MAX_BUCKETS = 160,
```
mmat11: nit: rename to `k_hist_max_buckets`

bpf/gotracer/maps/runtime.h
```diff
@@ -107,8 +108,31 @@ typedef enum go_runtime_metric_valid {
...
+};
+
+typedef enum go_runtime_histogram_kind {
+    go_runtime_histogram_kind_gc_pause = 0,
+    go_runtime_histogram_kind_scheduler_latency = 1,
+} go_runtime_histogram_kind_t;
+
+typedef struct go_runtime_histogram_event {
+    u8 type;
+    u8 kind;
```
mmat11: is this kind the same as `go_runtime_histogram_kind`? in that case I would use that type and size the enum to u8

bpf/gotracer/go_runtime.c
```diff
@@ -67,6 +67,46 @@ static __always_inline bool go_runtime_read_offset(void *dst, u32 size, u64 base
...
+
+    go_runtime_histogram_event_t *event =
+        bpf_ringbuf_reserve(&events, sizeof(go_runtime_histogram_event_t), 0);
+    if (!event) {
+        return;
+    }
+
+    event->type = EVENT_GO_RUNTIME_HISTOGRAM;
+    event->kind = kind;
+    event->_pad[0] = 0;
```
mmat11: we aren't reading paddings so we can avoid setting them

pkg/export/otel/metrics_runtime_histograms.go
```diff
@@ -0,0 +1,237 @@
...
+	time      time.Time
+	startTime time.Time
+	histogram runtimemetrics.GoRuntimeHistogramSnapshot
+}
+
+func newGoRuntimeHistogramProducer(temporality metricdata.Temporality) *goRuntimeHistogramProducer {
+	return &goRuntimeHistogramProducer{
+		temporality:  temporality,
+		histograms:   make(map[runtimemetrics.GoHistogramKind]goRuntimeHistogramState, 2),
+		lastProduced: make(map[runtimemetrics.GoHistogramKind]goRuntimeHistogramState, 2),
```
mmat11: if the kind of histograms will grow in the future these numbers need to be bumped, right? I wonder if we should put an enum guard like `runtimemetrics.GoHistogramKindMax` which can be used to grow these maps automatically -- or to keep it simpler, a constant which needs to be updated, WDYT?

marctc: Those were only initial capacity hints; the maps grow automatically. I removed the 2s to avoid implying they need to be kept in sync. WDYT?

pkg/export/otel/metrics_runtime_histograms.go
```diff
@@ -0,0 +1,237 @@
...
+			"metric", name,
+			"error", err)
+		return
+	}
+
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	previous, exists := p.histograms[histogram.Kind]
+	startTime := previous.startTime
```
MrAlias, line 79: This only retains one state for each kind. In Docker and Kubernetes, `service.instance.id` is container-scoped, so two Go PIDs in one container reach this producer with the same resource. Their snapshots overwrite each other here, leaving the exported histogram with whichever PID reported last; delta exports also reset whenever that PID changes.

Could we retain the latest histogram per PID and aggregate live PID populations before producing the container series? Prometheus needs the same treatment in `goRuntimeHistogramCollector`; a two-PID/container test should cover the overwrite and termination cases.

MrAlias (commented):
The histogram collection is on the right track, but it loses data for multi-process containers.

pkg/export/otel/metrics_runtime_histograms.go
mmat11: sounds good, sorry for the delay, I lost the notifications to review this one

mmat11 (approved):
🚀

pkg/export/otel/metrics_runtime_histograms.go
```diff
@@ -0,0 +1,306 @@
...
+			"pid", snapshot.PID,
+			"metric", name,
+			"error", err)
+		return
+	}
+
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	key := goRuntimeHistogramKey{kind: histogram.Kind, pid: snapshot.Service.ProcPID}
```
MrAlias: `tracer.AllowPID` registers `ChildPids` with the parent's `FileInfo`, so `runtimeMetricService` gives every child the same `Service.ProcPID`. Keying here by `Service.ProcPID` makes child snapshots overwrite one another, and `Delete(pid)` cannot evict their state because the process event carries the child PID. Prometheus passes the same value to its collector.

Carry the emitting process's host PID from the ring-buffer event and use it consistently for update/delete in both exporters. Add a child-worker lifecycle test; the current multi-PID tests set `Service.ProcPID` to each snapshot PID and miss this path.

MrAlias (commented):
The per-PID aggregation still loses histogram data for inherited child PIDs.

MrAlias (commented):
The per-PID aggregation fixes concurrent workers, but the termination path still loses or re-counts histogram observations.

pkg/export/otel/metrics_runtime_histograms.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+func (p *goRuntimeHistogramProducer) Delete(pid app.PID) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	for key := range p.histograms {
+		if key.pid == pid {
+			delete(p.histograms, key)
+			delete(p.lastProduced, key)
```
MrAlias, line 125: When delta temporality is selected, deleting both maps here drops every observation recorded after the last `Produce` call if the PID exits before the next collection. The last-PID path has the same gap because `onProcessEvent` deletes this state before removing and shutting down the reporter.

Could we retain the terminating PID's state through one final delta production, or flush its pending delta before deleting the baseline? A test that updates after the first `Produce`, terminates the PID, and then produces again should catch this.

pkg/export/prom/prom_runtime_histograms.go
```diff
@@ -0,0 +1,278 @@
...
+
+func (c *goRuntimeHistogramCollector) DeletePID(pid app.PID) {
+	if c == nil {
+		return
+	}
+
+	c.mu.Lock()
+	for key := range c.histogramSnapshots {
+		if key.pid == pid {
+			delete(c.histogramSnapshots, key)
```
MrAlias, line 136: These snapshots are cumulative, and `snapshot` removes the PID dimension before exposing them. If two PIDs share a label tuple, deleting one here makes every populated bucket and `_count` decrease. Prometheus treats that as a reset, so `rate()` and `increase()` count the surviving PID's existing population again even when it recorded nothing new.

Could we fold the terminated PID's final population into a retired per-series accumulator until the label tuple expires? The current termination test expecting `5 -> 3` should instead cover a scrape before and after termination and assert that the series stays monotonic.

pkg/export/otel/metrics_runtime_histograms.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+}
+
+func (p *goRuntimeHistogramProducer) Delete(pid app.PID) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	for key := range p.histograms {
+		if key.pid == pid {
+			if p.temporality == metricdata.DeltaTemporality {
```
MrAlias, line 96: The default cumulative path skips this branch and deletes the state below. If another PID remains for the same resource, the next cumulative point drops the terminated PID's populations; its start time either stays unchanged or overlaps values already exported from the surviving PID, so this is not a valid cumulative stream.

Could we retain the terminated population for cumulative OTLP output until the reporter expires, as the Prometheus collector now does? The lifecycle test should keep the aggregate at `5` after PID 101 exits, then advance it when PID 202 reports again.

pkg/export/otel/metrics_runtime.go
```diff
@@ -263,11 +285,43 @@ func (r *RuntimeMetricsReporter) reportMetrics(ctx context.Context) {
...
+	}
+}
+
 func (r *RuntimeMetricsReporter) reportRuntimeMetrics(snapshots []runtimemetrics.RuntimeMetricSnapshot) {
 	for _, snapshot := range snapshots {
 		if !r.shouldReportSnapshot(snapshot) {
 			continue
 		}
+		// A snapshot may still be in flight after its process terminated.
+		if !r.snapshotProcessLive(snapshot) {
```
MrAlias, line 320: `instrumentedEventLoop` starts the tracer reader before dispatching `ProcessEventCreated`, while this reporter consumes process events and snapshots from independent channels. A valid snapshot can therefore reach this check before the PID is tracked. `SnapshotFromRingbuf` has already resolved it against `CurrentPIDs`, so a missing exporter tracker entry is not evidence that the process terminated; a short-lived process may never produce another GC-cadence snapshot. Prometheus has the same ordering dependency in `runtimeSnapshotProcessLive`.

Could we reject only PIDs with an observed termination, or otherwise order PID registration ahead of runtime delivery? A snapshot-before-create test should cover this ordering.

MrAlias (commented):
The lifecycle handling still allows cumulative OTLP histograms to regress and valid startup snapshots to be dropped.

pkg/export/otel/pid_tracker.go
```diff
@@ -74,6 +78,17 @@ func (p *PidServiceTracker) TracksPID(pid app.PID) (svc.UID, bool) {
...
 }
 
+func (p *PidServiceTracker) PIDLiveOrUnknown(pid app.PID, uid svc.UID) bool {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+
+	if trackedUID, ok := p.pidToService[pid]; ok {
+		return trackedUID.Equals(&uid)
+	}
+	terminatedUID, terminated := p.terminatedPIDs[pid]
```
MrAlias, line 88: These tombstones treat `(pid, service UID)` as a process generation, but that pair is reused: host instance IDs are derived from the PID, and container instance IDs can remain stable across a respawn. Because the tracer accepts a reused PID before its `ProcessEventCreated` reaches this reporter, its snapshot can arrive first and be rejected here as belonging to the terminated process. A short-lived process can lose its only GC-cadence snapshot.

Could we cover termination → same PID/UID → snapshot-before-create and distinguish the new process from old in-flight data without relying on an unversioned PID tombstone?

MrAlias (commented):
The lifecycle guard now handles first-seen snapshots, but PID reuse can still cause valid runtime data to be dropped.

pkg/export/otel/pid_tracker.go
marctc: Good catch, but this seems like a narrow edge case. Do you consider it a blocker, or can we merge this PR and handle it separately? We’ve had quite a bit of back-and-forth already.

MrAlias: Yeah, not a blocker. We can follow-up on this.

MrAlias (approved):

pkg/export/otel/pid_tracker.go
MrAlias: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2915

PR #2764, opened 2026-07-23
Limit HTTP request capture to payload length

MrAlias:
## Summary

- bound the generic HTTP request-buffer read to the reported payload length
- retain the validated protocol prefix if the bounded helper read still fails
- add a PR-blocking native BPF regression test for short, capped, and failed reads

## Root cause

The Java TLS agent forwards exact-sized plaintext buffers through its ioctl
path. The generic HTTP tail stage nevertheless always asked `bpf_probe_read`
for the full 256-byte trace buffer. For the 74-byte `GET /` and 104-byte
`GET /greeting` requests used by OATS, that read could cross the readable
allocation boundary. A failed probe read zeroed the destination, so user space
received an HTTP record without a method and span metrics failed Weaver's
required `http.request.method` validation.

Whether the fixed-size over-read reached unreadable memory depended on the
adjacent allocation and page layout, which explains why identical main builds
could pass or fail and why retry failure counts varied.

This change reads `min(bytes_len, sizeof(info->buf))`. The 24-byte prefix that
already classified the payload as HTTP is retained as a best-effort fallback
if the bounded read fails for another reason.

## Validation

- generated the complete amd64 and arm64 BPF translations
- passed the native BPF regression test and license-header checks
- passed clang-format and clang-tidy for the changed production translation unit
- passed `run-oats_java_tls_jdk_8` with Weaver validation
- passed `run-oats_java_tls_sync_async` five times with Weaver validation
- confirmed every sync/async run emitted the 74-byte Google request as `GET /`
  with `http.request.method=GET`
- completed independent adversarial reviews focused on eBPF verifier safety,
  test/CI correctness, and root-cause alternatives

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -84,7 +84,7 @@ jobs:
         run: make prereqs
 
       - name: Run checks
-        run: make go-mod-tidy license-header-check notices-update check-go-mod check-ebpf-ver-synced check-config-schema check-config-v2-artifacts check-clean-work-tree
+        run: make go-mod-tidy test-bpf license-header-check notices-update check-go-mod check-ebpf-ver-synced check-config-schema check-config-v2-artifacts check-clean-work-tree
 
   clang-format:
     name: Clang format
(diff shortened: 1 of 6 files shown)


MrAlias:
Superseded by #2766

PR #2766, opened 2026-07-23
fix empty java tls empty method

NimrodAvni78:
## Summary

Fix intermittently missing `http.request.method` on HTTP metrics/spans for Java TLS traffic.

The request-line copy in `__obi_continue2_protocol_http` read a fixed `FULL_BUF_SIZE` (256) bytes from the user buffer:

```c
bpf_probe_read(info->buf, FULL_BUF_SIZE, (void *)args->u_buf);
```

`bpf_probe_read` is all-or-nothing. This is the only `FULL_BUF_SIZE`-sized read in the BPF tree that is not clamped to the actual payload length (every other buffer read clamps first, e.g. the sibling `small_buf` read in `k_tracer_defs.h`). The Java TLS ioctl path (`java_tls.c`) is the one ingress that hands over a tightly sized, flat user buffer with only the first and last byte validated as readable. When a request's first captured chunk is short (a bare request line + a few headers) and sits near the end of a mapped page, reading a fixed 256 bytes straddles an unmapped page, faults, and copies nothing — leaving `info->buf` zeroed. `httpMethodFromBuf` then finds no space in the buffer and returns an empty method, so `Span.Method == ""`.

The generic kprobe path reads kernel skb memory and the OpenSSL uprobe path reads the app's slack-padded heap buffer, so an over-read there lands in mapped memory and the method survives. That is why the loss is Java-TLS-specific and intermittent (depends on the payload's position relative to a page boundary): most spans carry `http.request.method`, some don't, and the empty-method datapoints propagate into the aggregated `http.server.*` / `http.client.*` metrics.

### Change

- Extract the copy into a small `read_request_buf()` helper in `bpf/common/http_types.h` that clamps the read to `args->bytes_len` before `bpf_probe_read`, mirroring the existing `small_buf` idiom. For the Java path `bytes_len == max_len` (already validated readable), so the read can never fault; for every other path reading `min(bytes_len, FULL_BUF_SIZE)` is strictly more correct.
- `protocol_http.h` calls the helper.
- Add a BPF C regression test (`bpf/tests/bpf_http_request_buf_read.c`) that models the kernel all-or-nothing `bpf_probe_read`: a short request line is fully copied under the clamp, the pre-fix fixed-size read faults on that same buffer, and an oversized payload clamps to capacity.
- Add a Go regression test asserting a short SSL request buffer still recovers the method.

## Validation

### DIFF
--- a/bpf/common/http_types.h
+++ b/bpf/common/http_types.h
@@ -5,6 +5,7 @@
 
 #include <bpfcore/vmlinux.h>
 #include <bpfcore/bpf_helpers.h>
+#include <bpfcore/utils.h>
 
 #include <common/connection_info.h>
 #include <common/http_buf_size.h>
@@ -59,6 +60,12 @@ typedef struct call_protocol_args {
     u64 self_ref_parent_id;
     lw_thread_t lw_thread;
 } call_protocol_args_t;
+
+static __always_inline void read_request_buf(http_info_t *info, const call_protocol_args_t *args) {
+    u32 info_buf_len = (u32)args->bytes_len;
+    bpf_clamp_umax(info_buf_len, FULL_BUF_SIZE);
+    bpf_probe_read(info->buf, info_buf_len, (void *)args->u_buf);
+}
 
 // Here we keep information on the packets passing through the socket filter
 typedef struct protocol_info {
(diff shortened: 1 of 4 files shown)


mmat11 (approved):
lgtm

grcevski (approved):
Epic find! LGTM!

PR #2770, opened 2026-07-23
run BPF tests in CI

NimrodAvni78:
## Summary

Run the `bpf/tests` C unit tests in CI.

The `bpf/tests` directory holds native C unit tests for BPF-side logic (`bpf_ssl_read_guard`, `bpf_http2_preface_bounds`, `bpf_kafka_large_buffer`, `bpf_postgres_detection`, `bpf_float64`, `test_trace_util`, `test_large_buffer_loop`, `test_failed_connect_event`), compiled natively against the `bpf/tests/bpfcore` stubs. Nothing ran them: no workflow referenced the directory, and its `Makefile` only compiled the binaries (the `all` / `%: %.c` rules never execute them). As a result these tests provided no automated protection.

### Change

- Add a `test` target to `bpf/tests/Makefile` that compiles **and runs** every test binary, failing on the first non-zero exit.
- Add a dedicated `.github/workflows/bpf_unit_tests.yml` workflow (ubuntu, install clang, `make -C bpf/tests test`), mirroring the `bpf_verifier.yml` pattern and scoped via `paths:` to `bpf/**` (and the workflow file), so it only runs when BPF sources change rather than on every code PR.

Newly added C tests (e.g. those introduced alongside BPF fixes) are then covered automatically.

## Validation

### DIFF
--- a/.github/workflows/bpf_unit_tests.yml
+++ b/.github/workflows/bpf_unit_tests.yml
@@ -1,0 +1,42 @@
+name: BPF unit tests
+
+on:
+  push:
+    branches: [main, release-*]
+  pull_request:
+    branches: [main, release-*]
+    paths:
+      - ".github/workflows/bpf_unit_tests.yml"
+      - "bpf/**"
+  workflow_call:
+    inputs:
+      ref:
+        description: "Git ref (commit, branch, or tag) to check out"
+        required: false
+        type: string
+
+concurrency:
+  group: pr-bpf-unit-tests-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  bpf-unit-tests:
+    name: BPF unit tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ inputs.ref || github.sha }}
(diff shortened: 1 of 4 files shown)


.github/workflows/pull_request.yml
```diff
@@ -118,6 +118,24 @@ jobs:
             exit 1
           fi
 
+  bpf-unit-tests:
```
pinoOgni: I know that these C tests are cheap but they will run on every code PR, not only for bpf changes because the paths filter is broad. 

A possible idea is to create `.github/workflows/bpf_unit_tests.yml` file mirroring the `.github/workflows/bpf_verifier.yml` pattern.

mariomac (approved):
Looks OK to me

mmat11 (approved):

.github/workflows/bpf_unit_tests.yml
```diff
@@ -0,0 +1,42 @@
...
+
+on:
+  push:
+    branches: [main, release-*]
+  pull_request:
+    branches: [main, release-*]
+    paths:
+      - ".github/workflows/bpf_unit_tests.yml"
+      - "bpf/**"
+  workflow_call:
```
pinoOgni, line 11: I think we should wire this block into `release.yml` like `bpf_verifier.yml` is and also add it to the `tests-passed` job's `needs:` list so that's blocks the release (like the `bpf-verifier:`).

pinoOgni (approved):
lgtm, I left a comment.

PR #2772, opened 2026-07-23
h2/grpc: context propagation for pre-existing connections and generic improvements

mmat11:
## Summary

- better SDK cooperation: when the app's gRPC library already writes a `traceparent`, OBI now detects it per request and stays out of the way; previously this was an all-or-nothing per-connection decision, so a single missed write lost the trace
- connections opened before OBI started (persistent channels, connection pools) were invisible; OBI now safely recognizes in-flight HTTP/2 and instruments it
- after the first request, gRPC compresses the `traceparent` header into an HPACK dynamic-table reference that OBI couldn't read, silently breaking trace chains on any persistent connection, now recovered on every request
- injection backs off whenever the wire state isn't exactly what we expect (mid-write buffer flushes, oversized headers, TLS, tunneled protocols)
- socket backfill now runs once per network namespace, so pods with established connections are tracked too
- tests: new Rust/tonic hop in the relay chain with a dedicated persistent-connection (HPACK dyn-table) integration suite

Known limitation: on connections that predate OBI, non-Go services report `*` as the span method name: the method string is fully dynamic-table-indexed and never appears on the wire again, so no wire-level observer can recover it. Go services are unaffected

Manually verified on Go and Python (grpcio) client/server pairs with connections established before OBI attach: spans emitted, parent/child links intact end-to-end

## Validation


grcevski:
This is EPIC! I'll review soon.

bpf/common/h2_defs.h
```diff
@@ -51,3 +73,102 @@ enum {
...
+    }
+
+    switch (type) {
+    case k_h2_frame_data:
+        if ((flags & ~(u8)k_h2_flag_end_stream) || !stream_id) {
+            return 0;
+        }
+        break;
+    case k_h2_frame_headers:
+        // client-initiated streams are odd
```
grcevski: A silly question, what do you mean odd :)? Like strange or the stream_id is always an odd number?

bpf/common/h2_defs.h
```diff
@@ -51,3 +73,102 @@ enum {
...
+    }
+
+    return 1;
+}
+
+// Validates one raw 9-byte frame header, returns payload length via frame_len, 0 to reject
+static __always_inline u8 h2_sniff_frame_header(h2_sniff_state_t *st,
+                                                const unsigned char *d,
+                                                u32 *frame_len) {
+    *frame_len = ((u32)d[0] << 16) | ((u32)d[1] << 8) | (u32)d[2];
```
grcevski, line 303: Does it make sense to check for the frame_len before reading anything, in case it's small unusable frame?

bpf/generictracer/hpack_traceparent_parse.h
```diff
@@ -0,0 +1,112 @@
...
+}
+
+static const u8 k_hex_char_lut[256] = {
+    ['0' ... '9'] = 1,
+    ['a' ... 'f'] = 1,
+    ['A' ... 'F'] = 1,
+};
+
+// Recovers traceparent when the HPACK name reference is a dyn-table index
+static __always_inline u8 find_hpack_traceparent_value(const unsigned char *data,
```
grcevski, line 31: Very cool idea!

bpf/generictracer/hpack_traceparent_parse.h
```diff
@@ -0,0 +1,112 @@
...
+                    continue;
+                }
+                if (data[i + k_hpack_tp_name_offset + k_hpack_tp_name_len] !=
+                    k_hpack_value_len_tp) {
+                    continue;
+                }
+                return try_parse_tp_value(&data[i + k_hpack_tp_val_offset], tp);
+            }
+
+            if (name_len_byte == (k_hpack_tp_name_huffman_len | 0x80)) {
```
grcevski, line 96: Do you think it's remotely possible to fingerprint a traceparent value encoded as huffman? Maybe get excited about fields of certain size, or perhaps the dash encoded as 3 count and the rest all hex?

grcevski: Not saying it should be in this PR :)

bpf/generictracer/protocol_handler.c
```diff
@@ -35,7 +36,9 @@ int obi_handle_buf_with_args(void *ctx) {
     if (args->protocols.http && is_http(args->small_buf, MIN_HTTP_SIZE, &args->packet_type)) {
         bpf_tail_call(ctx, &jump_table, k_tail_protocol_http);
     } else if ((args->protocol_type != k_protocol_type_http) &&
```
grcevski, line 38: Since we attach midstream now, can we possibly get a reversed http2, or see a response thinking it's a request? Maybe it's OK since the userspace will likely reject it.

pkg/internal/ebpf/tpinjector/tpinjector.go
```diff
@@ -37,18 +41,52 @@ type Tracer struct {
...
+	if _, ok := p.seenNetns[inode]; ok {
+		return
+	}
+
+	ok := true
+	for _, it := range p.Iters() {
+		if err := netns.WithNetNS(int(pid), func() error {
+			return it.Run(p.log)
+		}); err != nil {
+			p.log.Error("error running iterator in netns", "pid", pid, "error", err)
```
grcevski, line 117: I thik this operation needs SYS_ADMIN, so if the users didn't supply it they'll get this error non-stop. Maybe best to either report it once, saying you are likely missing cap_sys_admin.

grcevski (approved):
This looks great, I left a few comments and I'm approving since I'm away for a week.

bpf/common/h2_defs.h
mmat11: the comment was badly written -- odd-numbered :D! see: https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-http2-05#section-5.1.1

bpf/common/h2_defs.h
mmat11: we already check at the start of h2_sniff_check_frame, would you like me to move the check here for clarity?

bpf/generictracer/hpack_traceparent_parse.h
mmat11: I think I can do it in a follow-up once the infra is in place

bpf/common/h2_defs.h
grcevski: Up to you, as long as it's checked, I didn't verify :).

bpf/common/h2_defs.h
grcevski: Nice! I didn't know. This could be a way to tell server or client when I work on improving the http2 support in Go.

internal/test/integration/grpc_relay_test.go
```diff
@@ -69,55 +69,48 @@ func TestSuite_GRPCRelay(t *testing.T) {
...
-		require.Equal(ct, http.StatusOK, r.StatusCode)
-		var tq jaeger.TracesQuery
-		require.NoError(ct, json.NewDecoder(r.Body).Decode(&tq))
-		require.NotEmpty(ct, tq.Data)
-	}, time.Minute, time.Second)
-	t.Log("instrumentation ready")
+// waitForRelayInstrumentation blocks until OBI produces spans for the first and last hops
+func waitForRelayInstrumentation(t *testing.T) {
+	waitForServiceSpans(t, "/smoke", "go-entry", time.Minute)
+	waitForServiceSpans(t, "/relay", "dotnet-relay", 2*time.Minute)
```
MrAlias: This accepts readiness after finding a `go-entry` span and any `dotnet-relay` span, but it never establishes that one request reached the intermediate services. When an intermediate process has not been discovered yet, the following subtests start with a fresh trace that stops partway through the relay chain and time out.

Maybe, gate the suite on a single `/relay` trace containing every service in `expectedRelayServices` (or wait for each hop explicitly) before running the strict checks?

MrAlias (commented):
The relay integration suite can begin strict propagation checks before every hop has been discovered. That race is already capable of making the new coverage fail independently of the HTTP/2 behavior under test.

MrAlias (approved):

mmat11:
@MrAlias I added an integration test which found a bug and then I needed to refactor several files in order to fix it due to insns limit. Can I get another pass?

pkg/ebpf/common/http2grpc_transform.go
```diff
@@ -164,16 +164,33 @@ func knownFrameKeys(fr *http2.Framer, hf *http2.HeadersFrame) bool {
...
+		statusLast  = 14
+		indexed     = 0x80
+	)
+
+	if b&indexed != 0 {
+		idx := b &^ byte(indexed)
+		return idx >= statusFirst && idx <= statusLast
+	}
+
+	return b == statusFirst || b == 0x10|statusFirst || b == 0x40|statusFirst
```
MrAlias: Any static `:status` entry (8–14) is valid as a literal name reference. The x/net encoder emits a first `:status: 418` as `0x4e` (name index 14), so this predicate returns false and the response is decoded by `h2c.hdec`; that inserts response fields into the request-side dynamic table before `isResponse` is observed. Subsequent request indices can then resolve against the wrong entries. Could we keep these literal forms out of the request decoder and add a `418` regression test?

MrAlias (commented):
This still permits a valid HPACK encoding to desynchronize the request-side decoder.

MrAlias (approved):

PR #2775, opened 2026-07-24
bpf: restore HTTP server span context on SQL query return

mtanda:
## Summary

`process_sql_return()` in `go_sql.c` calls `obi_ctx__del()` unconditionally when a SQL query returns, clearing `traces_ctx_v1[pid_tgid]` regardless of whether an outer server span (HTTP) was still active on that goroutine.

`traces_ctx_v1` (`bpf/shared/obi_ctx.h`) is keyed only by `pid_tgid`, with no notion of nesting depth. So when a SQL query runs inside an HTTP handler, a very common shape such as logging the response after the query that built it, this unconditional delete clobbers the outer HTTP server span's context too. Any log line the handler writes after the query returns never gets `trace_id`/`span_id` injected by the log enricher (`ebpf.log_enricher`), even though the HTTP span is still active.

## Fix

Before clearing the context, look up `ongoing_http_server_requests` for the current goroutine. If an HTTP server invocation is still ongoing, restore its context instead of clearing it.

## Testing

Verified against a Go (gin + gorm + Postgres) service with `ebpf.log_enricher` enabled: log lines written after a request that runs a SQL query now get `trace_id`/`span_id` injected.

### DIFF
--- a/bpf/gotracer/go_sql.c
+++ b/bpf/gotracer/go_sql.c
@@ -25,6 +25,8 @@
 
 #include <gotracer/go_common.h>
 #include <gotracer/go_str.h>
+#include <gotracer/maps/nethttp.h>
+#include <gotracer/types/nethttp.h>
 
 #include <maps/go_sql.h>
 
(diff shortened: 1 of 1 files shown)


marctc (commented):
Hey @mtanda, thanks a lot for the fix! The patch makes sense to me, but I have a suspicion that this can also happen for other protocols such as in redis, mongo, kafka, and maybe grpc. Perhaps would be good to check for those paths also happens.

Also, we could add an itegration test to validate that behaviour is correct after the fix, WDYT?

mtanda:
@marctc Thanks for the quick response!

Yes, I agree that this is likely to happen with other protocols too.
One of the intentions behind this PR was to keep the fix as minimal as possible, limited to what I could actually verify in my own environment. So it might be better to close this PR and instead consider a more comprehensive fix that covers the other protocols as well.

Also, since it's not immediately obvious when log enrichment isn't working correctly, I agree that adding a test would be valuable.

mmat11:
nice catch... I'm not sure this is the right way to handle this. As @marctc said, this could happen in a lot of other places, I wonder if using a stack and pushing/popping the context instead of deleting it would solve it

marctc:
@mtanda I suggest that you open an issue and we take it from there!

mtanda:
@marctc Yes, I agree. This PR doesn't really solve the actual problem. I just don't have a clear picture yet of how the underlying issue could be fixed, so I opened an issue for it instead.
https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2778

Since I opened the issue from the CLI, it doesn't have the bug or needs triage label applied. Could you please add them? Sorry for the extra step.

PR #2785, opened 2026-07-26
k8s integration: enforce weaver semconv live-check in all OTLP-exporting suites

### LINKED ISSUE #2571: Weaver validation should cover k8s tests

NimrodAvni78:
Resolves https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2571

## Summary

Wires `weaver registry live-check` into all 11 Kubernetes/kind suites that
export telemetry over OTLP, enforcing from day one (no observe mode): any
actionable semconv advisory fails the suite. The 5 remaining k8s suites
(`prom`, `informer_cache`, `netolly_multizone_prom`, `netolly_promexport`,
`netolly_tc_promexport`) export Prometheus directly with no OTLP path, so
weaver cannot see them.

## Review guide

55 files, but only ~10 carry logic; the other ~45 are mechanical repetition.
Suggested reading order:

**Review closely — the actual design (5 files):**
1. `internal/test/integration/components/kube/weaver.go` (new) — the whole
   validation harness: the `weaverRecorder` adapter, the Finish hook, the
   wait-for-HTTP / drain / stop / empty-report-retry sequence. Start here.
2. `internal/test/integration/components/kube/kind.go` — the `WeaverValidation()`
   option, Finish-func wiring, and the exit-code enforcement (incl. the
   `os.Exit(code)` setup-failure fix). ~49 lines.
3. `internal/test/integration/configs/otelcol-config-k8s-weavercol.yml` (new) —
   the intermediate collector: `interval` aggregation + drop-on-overflow. This
   is where the two-hop back-pressure rationale lives (see "How it works").
4. `internal/test/integration/configs/otelcol-config-k8s-weaver.yml` (new) —
   the suite otelcol "tap" that forwards to weavercol while leaving the
   existing prometheus/jaeger assertion paths byte-identical.
5. `schemas/obi/groups/{network,service_graph,k8s}.yaml` (new/extended) —
   semconv declarations; needs a domain eye on attribute names, types, and
   stability. All driven by observed live-check failures (see "Registry
   additions").

**Review once — new infra manifests, mostly declarative wiring (5 files):**
6. `k8s/manifests/08-weaver.yml` (new) — weaver + weavercol pods, registry
   PV/PVC, admin port, `runAsUser: 0`.
7. `k8s/manifests/08-weaver-multi-node.yml` (new) — **diff against #6**; it's
   the same manifest plus `nodeAffinity` pinning to the `otel` zone node.
8. `k8s/manifests/03-otelcol-weaver{,-multi-node}.yml` (new) — otelcol
   manifests that mount the tapped config from #4.
9. `k8s/manifests/00-kind{,-multi-node,-multi-zone}.yml` — +7-9 lines each:
   the `obi-registry` extraMount and the 4320→32320 admin port mapping.

**Skim for consistency — mechanical, no logic (~45 files):**
- **11 suite `*_main_test.go`**: identical 5-line opt-in (swap
  `03-otelcol.yml` → `03-otelcol-weaver.yml`, add `kube.WeaverValidation()`,
  deploy `08-weaver.yml`). Read one (`otel/k8s_main_test.go`); the rest match,
  with multi-node suites using the `-multi-node` manifest variants.
- **~28 `05-*` / `06-*` manifests + `k8s/common/k8s_metrics_testfuncs.go` +
  `configs/obi-config-promscrape.yml`**: pure `deployment.environment` →
  `deployment.environment.name` rename (the testfuncs diff is 30/30 with the
  rest being gofmt column re-alignment). No logic change.

## How it works

- **`08-weaver.yml` / `08-weaver-multi-node.yml`**: in-cluster weaver pod
  (same image pin as compose/OATS) reading the registry from an `obi-registry`
  kind extraMount, writing its report to the shared `testoutput` mount. Admin
  port exposed to the host via kind extraPortMapping 4320→32320. The
  multi-node variant is nodeAffinity-pinned to the `otel`-zone node that
  carries the mount + port mapping. The pod runs `runAsUser: 0`: the shared
  `testoutput` hostPath is root-owned on a fresh CI node, and the weaver
  image's default `uid=1000` cannot create the `weaver-out` subdirectory
  there — same reason the docker-compose weaver sets `user: "0:0"` and OBI
  writes to the mount as root.
- **`03-otelcol-weaver.yml` + `configs/otelcol-config-k8s-weaver.yml`**:
  otelcol variant (same 0.104.0 pin as the base suites) that, besides the
  existing prometheus/jaeger paths, forwards metrics + traces to a small
  dedicated **`weavercol`** collector (contrib 0.155.0, in 08-weaver*.yml)
  which aggregates and relays to weaver. The two-hop design is load-bearing:
  weaver validates every sample and is slow at netolly-scale cardinality;
  when the suite collector exported to weaver directly, its sending queue
  filled and the enqueue failure propagated synchronously through the shared
  OTLP receiver back to OBI, starving the production prometheus path the
  suites assert on (observed: OBI logging "sending queue is full" while test
  queries returned empty). weavercol ingests instantly, so weaver
  back-pressure surfaces there and reaches the suite collector only as an
  asynchronous send failure — dropped and logged, never failing OBI's
  export. weavercol also runs a 10s `interval` aggregation so weaver sees
  each unique series shape instead of every 10ms data point — netolly report
  generation dropped from ~190s to seconds.
- **`kube.WeaverValidation()` option + teardown hook** (components/kube/
  weaver.go): suites opt in via a `NewKind` option; validation runs as the
  first e2e-framework `Finish` func — after all tests, while the cluster is
  still up, before log export and cluster destruction. The validation waits
  for weaver's admin port to answer HTTP (image pull can finish after the
  suite's tests start; a bare TCP dial is not enough because kind's port
  mapping accepts and resets), drains 25s (covers weavercol's interval tick
  + gRPC reconnect backoff), POSTs /stop, fetches the report via the shared
  `weavercheck.FetchReport`, and retries the whole stop/fetch cycle up to 3
  times if the report has zero samples (weaver's pod restarts after /stop
  and OBI keeps emitting, so a weaver that came up mid-suite recovers) —
  guarding against vacuous passes. Archives the report next to the kind
  logs and runs the shared `weavercheck.Validate`.
- **Enforcement is out of band, on purpose**: e2e-framework logs `Finish`
  errors at `klog.V(2)` only and drops them from the exit code ("attempt to
  gracefully clean up. Upon error, log and continue"), and `Finish` funcs
  carry no `*testing.T` — still true in v0.7.0. So validation runs against a
  small `weaverRecorder` adapter implementing `weavercheck.TestingT`
  (failures logged loudly via slog, `FailNow` = `runtime.Goexit` on a
  dedicated goroutine, mirroring `testing.T` semantics for the `require`
  calls inside the shared validator); a recorded failure makes `kube.Kind.Run`
  exit non-zero. The `weaverRecorder` is also the single place the
  `weaver(k8s):` log prefix is applied, so the shared (prefix-less)
  `weavercheck.Validate` output is tagged consistently. The trade-off vs. a
  per-suite last-running test: the failure is package-level (no named failing
  test in `go test` output) in exchange for zero per-suite boilerplate and no
  reliance on test ordering.

## Bugs found by the wiring itself

1. **`kube.Kind.Run` swallowed setup failures** (components/kube/kind.go): a
   suite whose cluster/image/manifest setup failed but whose teardown
   succeeded exited 0 and read as `ok` — CI could silently green-light a
   broken suite. Now `os.Exit(code)` on non-zero.
2. **Deprecated `deployment.environment`** baked into 28 k8s test manifests
   and shared assertions — renamed to the current semconv
   `deployment.environment.name` (weaver flagged the deprecation on
   `target_info`). This includes the snake_case spelling
   `extra_resource_attributes: ["deployment_environment"]` in
   `configs/obi-config-promscrape.yml` (the allowlist OBI's own prometheus
   exporter uses for `target_info`), which the dotted-name sweep initially
   missed — the `prom` suite asserts on that label and fails without it.

## Registry additions (all driven by observed live-check failures)

- `schemas/obi/groups/k8s.yaml` (new): `k8s.owner.name`, `k8s.kind` — OBI's
  k8s decorator output (pkg/transform/k8s.go), observed on `target_info`.
- `schemas/obi/groups/network.yaml`: the 14 `k8s.src.*` / `k8s.dst.*` flow
  endpoint attributes (observed on `obi.network.flow.bytes` in all three
  netolly suites), ref'd on both flow metrics which share the attach path.
  GeoIP attributes deliberately NOT declared (not observed — feature off in
  these suites).
- `schemas/obi/groups/service_graph.yaml`: legacy underscored k8s peer labels
  (`client_k8s_namespace_name`, `server_k8s_namespace_name`,
  `client_k8s_cluster_name`, `server_k8s_cluster_name`), observed on the
  service-graph metrics in the `otel` suite.

`make lint-schema` passes with all additions.

## Suite plumbing changes

- `sharedpidns` routed its traces directly to jaeger, so weaver saw nothing —
  now routes through the weaver-tapped otelcol, which forwards to jaeger
  (assertions unchanged). Flagged by the new empty-report guard.
- Weaver manifests must not start with a `---` separator after the comment
  header: e2e-framework's decoder rejects the empty leading YAML document
  (`Object 'Kind' is missing in 'null'`).

## Validation

### DIFF
--- a/internal/test/integration/components/kube/weaver.go
+++ b/internal/test/integration/components/kube/weaver.go
@@ -231,8 +231,8 @@ func (k *Kind) validateWeaver(parent context.Context, t weavercheck.TestingT) {
 	}
 
 	if k.weaverRequireSpans && report.Statistics.TotalEntitiesByType["span"] == 0 {
-		t.Errorf("weaver report has no span entities — OBI's trace export is not reaching weaver "+
-			"(is OTEL_EXPORTER_OTLP_TRACES_ENDPOINT pointed at the tapped otelcol rather than "+
+		t.Errorf("weaver report has no span entities — OBI's trace export is not reaching weaver " +
+			"(is OTEL_EXPORTER_OTLP_TRACES_ENDPOINT pointed at the tapped otelcol rather than " +
 			"straight at jaeger?); span semconv violations would otherwise go unvalidated")
 	}
 
--- a/internal/test/integration/components/kube/weaver.go
+++ b/internal/test/integration/components/kube/weaver.go
@@ -216,18 +216,18 @@ func (k *Kind) validateWeaver(parent context.Context, t weavercheck.TestingT) {
 		}
 	}
 
-	// A dropped export may have carried the sole sample of a violating shape, so
-	// any drop during the suite — or an unreadable counter — makes the report
-	// untrustworthy.
+	// A first-hop drop may have carried the sole sample of a violating shape, so
+	// any such drop during the suite — or an unreadable counter — makes the
+	// report untrustworthy.
 	if k.tapDropsBaselineErr != nil || finalDropsErr != nil {
 		t.Errorf("could not read the tap's export-failure counters (baseline: %v, teardown: %v) — "+
 			"cannot confirm weaver saw every emitted shape, so the report is untrustworthy",
 			k.tapDropsBaselineErr, finalDropsErr)
 	} else if dropped := finalDrops - k.tapDropsBaseline; dropped > 0 {
-		t.Errorf("the weaver tap dropped %.0f export item(s) during the suite "+
-			"(otelcol_exporter_{send,enqueue}_failed_* across the suite otelcol and weavercol) — "+
-			"weaver may have missed a telemetry shape, so the report cannot be trusted; reduce "+
-			"OBI's emission rate or raise the tap queue sizes", dropped)
+		t.Errorf("the suite otelcol dropped %.0f export item(s) to weavercol during the suite "+
+			"(otelcol_exporter_{send,enqueue}_failed_*) — weaver may have missed a telemetry "+
+			"shape, so the report cannot be trusted; reduce OBI's emission rate or raise the "+
+			"first-hop tap queue size", dropped)
 	}
 
 	if k.weaverRequireSpans && report.Statistics.TotalEntitiesByType["span"] == 0 {
@@ -239,18 +239,14 @@ func (k *Kind) validateWeaver(parent context.Context, t weavercheck.TestingT) {
 	weavercheck.Validate(t, report)
 }
 
-// tapDropCount sums the export failures of both tap hops (suite otelcol ->
-// weavercol -> weaver), erroring if either endpoint can't be scraped.
+// tapDropCount reads the suite otelcol's export failures to weavercol, the
+// first tap hop. Only this hop is a trust signal: weavercol ingests instantly
+// so the first-hop queue never fills, making any failure here a shape that
+// never reached the aggregator. The weavercol -> weaver hop drops by design
+// under weaver back-pressure (otelcol-config-k8s-weavercol.yml), shedding only
+// already-aggregated points weaver has already seen.
 func (k *Kind) tapDropCount(ctx context.Context) (float64, error) {
-	var total float64
-	for _, port := range []int{OtelcolWeaverMetricsHostPort, WeaverColMetricsHostPort} {
-		n, err := exporterFailedCount(ctx, port)
-		if err != nil {
-			return 0, fmt.Errorf("scraping exporter counters on host port %d: %w", port, err)
-		}
-		total += n
-	}
-	return total, nil
+	return exporterFailedCount(ctx, OtelcolWeaverMetricsHostPort)
 }
 
 // exporterFailedCount sums otelcol_exporter_{send,enqueue}_failed_* from a
--- a/internal/test/integration/components/kube/weaver.go
+++ b/internal/test/integration/components/kube/weaver.go
@@ -1,0 +1,322 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package kube // import "go.opentelemetry.io/obi/internal/test/integration/components/kube"
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"runtime"
+	"strconv"
+	"strings"
+	"time"
+
+	"sigs.k8s.io/e2e-framework/pkg/env"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+
+	"go.opentelemetry.io/obi/internal/test/weavercheck"
+)
+
+const (
+	// WeaverK8sAdminHostPort is the host port the kind clusters map weaver's
+	// admin port (4320) to, via the extraPortMapping in manifests/00-kind*.yml.
+	WeaverK8sAdminHostPort = 32320
+
+	// Host ports the kind clusters map the weavercol and suite-otelcol telemetry
+	// endpoints to (00-kind*.yml); validateWeaver scrapes both for tap drops.
+	WeaverColMetricsHostPort     = 32888
+	OtelcolWeaverMetricsHostPort = 32889
+
+	// weaverK8sTimeout bounds the wait-for-weaver + /stop + report-read
+	// sequence. Report generation scales with the number of unique samples
+	// weaver received; the interval processor in
+	// otelcol-config-k8s-weavercol.yml keeps that bounded, but
+	// high-cardinality suites (netolly) still need a generous margin.
+	weaverK8sTimeout = 6 * time.Minute
+
+	// weaverK8sDrainWindow is how long validateWeaver waits after weaver is
+	// reachable before stopping it: it must cover the weavercol `interval`
+	// processor tick (10s, see otelcol-config-k8s-weavercol.yml) plus gRPC
+	// reconnect backoff, so at least one aggregated batch arrives.
+	weaverK8sDrainWindow = 25 * time.Second
+
+	// weaverK8sEmptyReportAttempts is how many /stop + read cycles to try
+	// when the report comes back with zero samples (each cycle restarts the
+	// weaver pod and drains again).
+	weaverK8sEmptyReportAttempts = 3
+
+	// weaverK8sReadyTimeout bounds the pre-test wait for the weaver tap to come
+	// up (see waitForWeaverReady). It is generous because weaver and weavercol
+	// pull their images from the network on a cold CI node.
+	weaverK8sReadyTimeout = 5 * time.Minute
+)
+
+// weaverRecorder adapts weavercheck.TestingT to the suite-teardown context,
+// where no *testing.T exists (e2e-framework Finish funcs run after m.Run()
+// has returned). Failures are recorded and logged at default slog verbosity —
+// deliberately NOT relying on the framework's handling of Finish errors,
+// which are logged at klog V(2) only and dropped from the exit code.
+type weaverRecorder struct {
+	failed bool
+}
+
+func (r *weaverRecorder) Helper() {}
+
+// Logf/Errorf are the single place a "weaver(k8s):" prefix is applied, so the
+// shared weavercheck.Validate output (which is prefix-less, being reused by the
+// non-k8s compose/OATS suites) is tagged consistently. validateWeaver's own
+// messages therefore must NOT repeat the prefix, or it would double up.
+func (r *weaverRecorder) Logf(format string, args ...any) {
+	log().Info("weaver(k8s): " + fmt.Sprintf(format, args...))
+}
+
+func (r *weaverRecorder) Errorf(format string, args ...any) {
+	r.failed = true
+	log().Error("weaver(k8s): " + fmt.Sprintf(format, args...))
+}
+
+// FailNow mirrors testing.T.FailNow for the require calls inside
+// weavercheck.Validate: validateWeaverFinish runs the validation on a
+// dedicated goroutine precisely so this Goexit aborts only the validation.
+func (r *weaverRecorder) FailNow() {
+	r.failed = true
+	runtime.Goexit()
+}
+
+// waitForWeaverReady is a setup func (registered before the suite's tests run,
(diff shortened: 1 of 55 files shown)


mariomac (approved):
Looks Ok!

internal/test/integration/k8s/otel/k8s_main_test.go
```diff
@@ -54,9 +54,13 @@ func TestMain(m *testing.M) {
 		kube.Deploy(testpath.Manifests+"/01-volumes.yml"),
 		kube.Deploy(testpath.Manifests+"/01-serviceaccount.yml"),
 		kube.Deploy(testpath.Manifests+"/02-prometheus-otelscrape.yml"),
-		kube.Deploy(testpath.Manifests+"/03-otelcol.yml"),
+		// weaver-tapped otelcol + in-cluster weaver pod, validated at suite
+		// teardown (enforcing)
+		kube.WeaverValidation(),
```
MrAlias: This validates a tapped collector, but the service deployed immediately below still loads `configs/obi-config.yml`, where `otel_traces_export.endpoint` is `http://jaeger:4318`. The suite's trace traffic never reaches this collector, so Weaver can pass on metrics while missing every span-specific semconv violation.

Can we route traces through otelcol here, as `sharedpidns` now does, and require a span sample in the report for trace-enabled suites?

internal/test/integration/configs/otelcol-config-k8s-weavercol.yml
```diff
@@ -0,0 +1,49 @@
...
+  otlp/weaver:
+    endpoint: weaver:4317
+    tls:
+      insecure: true
+    # weaver's OTLP receiver rejects gzip-compressed payloads
+    compression: none
+    sending_queue:
+      enabled: true
+      queue_size: 500
+      block_on_overflow: false
```
MrAlias, line 38: When Weaver applies back-pressure, this queue is explicitly allowed to drop requests and retries are disabled. `validateWeaver` only checks `len(report.Samples) > 0`, so dropping the sole batch carrying a bad shape still leaves a passing report as long as any other sample arrived.

The tap must either make any dropped export fail validation or preserve at least one sample for every distinct signal/attribute shape before this lossy queue; otherwise the new enforcement can green-light semconv regressions.

MrAlias (commented):
The live-check plumbing can currently pass without observing all of the telemetry it claims to validate.

internal/test/integration/components/kube/weaver.go
```diff
@@ -0,0 +1,351 @@
...
+	adminURL := fmt.Sprintf("http://%s/stop", addr)
+	reportPath := path.Join(testpath.Output, weaverK8sReportSubdir, weaverK8sReportFile)
+	var report *weavercheck.Report
+	for attempt := 1; ; attempt++ {
+		// Bracket the drain window with weavercol's failed-export counters.
+		// weaver is confirmed up (waitForHTTP) at the top of every iteration,
+		// so any increase over this window is genuine back-pressure loss
+		// rather than a weaver-restart transient (those land outside the
+		// window, folded into the next iteration's baseline).
+		dropsBefore, dropErrBefore := weavercolDropCount(ctx)
```
MrAlias: This baseline is captured only after `m.Run` has returned, so send or enqueue failures during the suite are already included in `dropsBefore`. That means the delta can be zero even if Weaver missed the sole sample of a violating shape.

Could we capture the baseline before tests start and compare the lifetime counter here? I also think a scrape failure should make the report untrustworthy rather than skip the dropped-export check.

internal/test/integration/configs/otelcol-config-k8s-weaver.yml
```diff
@@ -0,0 +1,67 @@
...
+      insecure: true
+  otlp/weavercol:
+    endpoint: weavercol:4317
+    tls:
+      insecure: true
+    sending_queue:
+      enabled: true
+      queue_size: 500
+    retry_on_failure:
+      enabled: false
```
MrAlias, line 53: `validateWeaver` currently scrapes only the downstream `weavercol` counters, while this first hop can also drop a batch if `weavercol` becomes unavailable after readiness. In that case, the downstream counters do not observe the loss and the report may still pass.

Could we either make this hop lossless or include this collector's `otelcol_exporter_send_failed_*` and enqueue-failure counters in validation?

internal/test/integration/k8s/manifests/06-instrumented-grpc-client.template.yml
```diff
@@ -22,10 +22,10 @@ metadata:
     # this label will trigger a deletion of OBI pods before tearing down
     # kind, to force OBI writing the coverage data
     teardown: delete
-    deployment.environment: integration-test
+    deployment.environment.name: test
     app.kubernetes.io/version: '3.2.1'
   annotations:
-    resource.opentelemetry.io/deployment.environment: 'to-be-ignored-in-favor-of-annotation'
+    resource.opentelemetry.io/deployment.environment.name: 'to-be-ignored-in-favor-of-annotation'
```
MrAlias: Renaming this key also makes this sentinel, and similar precedence-test sentinels in other manifests, values of the enumerated `deployment.environment.name` attribute. The current `otel` check reports six actionable `undefined_enum_variant` advisories.

Could these fixtures use documented enum values so they still test source selection without producing invalid telemetry?

internal/test/integration/k8s/manifests/06-obi-daemonset.yml
```diff
@@ -79,7 +79,7 @@ spec:
             - name: OTEL_EBPF_DISCOVERY_POLL_INTERVAL
               value: "500ms"
             - name: OTEL_EBPF_METRICS_INTERVAL
-              value: "10ms"
+              value: "1s"
```
MrAlias, line 82: This 10ms to 1s change appears in 18 manifests, including suites that do not deploy Weaver. It also changes the existing test workload and no longer matches the new `weavercol` configuration's assumption that OBI emits every 10ms.

Could we keep the original source cadence and let `weavercol` aggregate it? If the broader cadence change is intentional, it may be easier to evaluate separately with its rationale.

MrAlias (commented):
I found a few gaps in the live-check coverage, along with an enum migration issue that appears to be keeping the `otel` suite red.

internal/test/integration/k8s/otel/k8s_main_test.go
MrAlias: `otel` and `sharedpidns` now route traces through the tap and require a span sample, but this is only partially resolved. The `daemonset`, `daemonset_multi_node`, `daemonset_python`, and `owners` suites still call `WeaverValidation()` without `WeaverRequireSpans()`, while their OBI manifests export traces directly to Jaeger. Those suites can therefore pass Weaver with metrics only and leave their span semconv unchecked.

Could we keep this thread open until the remaining trace-producing suites also route traces through the tap and require spans?

internal/test/integration/k8s/manifests/06-obi-daemonset.yml
NimrodAvni78: we also dealt with the same issue [here](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2008)

```
Metrics export interval: 10ms → 1s
Bumped on 56 weaver-wired composes. The historical 10ms
OTEL_EBPF_METRICS_INTERVAL overwhelmed weaver's per-entity Rego pass — the
otlp/weaver queue overflowed within seconds, every batch in flight was
dropped, and runWeaverValidation consistently raced against an empty
report. 1s drains cleanly, with no observable change in what the test
assertions can see (the production pipeline still polls Prometheus / Jaeger
on the same timing it always did).
```

with a sample every 10ms the weaver tests we're overflowing and draining stalled / data was dropped
i can revert to 10ms in suites that don't deploy weaver but its the same as in the integration tests + OATs that we changed

let me know if you think its fine this way

MrAlias (commented):
The live-check still has unresolved loss-accounting and signal-coverage gaps on this head.

internal/test/integration/k8s/disable_informers/k8s_disable_informers_test.go
```diff
@@ -43,10 +43,14 @@ func TestMain(m *testing.M) {
 		kube.Deploy(testpath.Manifests+"/01-volumes.yml"),
 		kube.Deploy(testpath.Manifests+"/01-serviceaccount.yml"),
 		kube.Deploy(testpath.Manifests+"/02-prometheus-otelscrape.yml"),
-		kube.Deploy(testpath.Manifests+"/03-otelcol.yml"),
+		// weaver-tapped otelcol + in-cluster weaver pod, validated at suite
+		// teardown (enforcing)
+		kube.WeaverValidation(),
```
MrAlias, line 48: `06-obi-daemonset-disable-informers.yml` exports traces directly to Jaeger, while this opt-in does not require a span sample. The suite's `/iping` traffic can therefore produce spans that Weaver never sees, and a metrics-only report still passes.

Could we route this suite's traces through the tapped collector and use `WeaverRequireSpans()` here?

MrAlias (commented):
The latest head still permits the live-check to pass when the downstream tap drops Weaver's only copy of a metric or span shape. `disable_informers` also still sends traces directly to Jaeger without requiring a span in the report.

NimrodAvni78:
@MrAlias do you mind if we merge it now and ill work on finishing all these gaps in a follow up?
since this is blocking a few PRs like telemetry schema coverage

PR #2795, opened 2026-07-27
Add opt-in GenAI error status messages

### LINKED ISSUE #2507: Track opt-in handling for raw GenAI provider error messages
## Context

Follow-up from PR #2503:

- PR: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2503
- Review comment from @grcevski: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2503#discussion_r3477350220

The PR stops copying raw GenAI provider error messages into OTLP span status.message by default. @grcevski raised a valid usability concern: the raw message is often useful for operators, for example quota exceeded, rate limit, or invalid API key diagnostics.

## Security and privacy concern

The raw provider message is provider-supplied free text. OBI cannot reliably know whether it contains only benign diagnostic text or sensitive material such as API key fragments, organization/project/account IDs, tenant metadata, request details, prompt excerpts, or customer usage and billing state.

attributes.select.traces filters span attributes by attribute name. It does not filter arbitrary values, and it does not apply to OTLP status fields. Putting provider text in status.message therefore bypasses the same selection controls users rely on for optional GenAI attributes.

There is a related DB precedent: DB status messages are derived from the optional db.response.error path only after that attribute is selected. GenAI provider messages did not have an equivalent opt-in path.

## OpenTelemetry specification context

- The OpenTelemetry trace API models status as a code plus optional description/message, and the message is only meaningful for Error status. Instrumentation-provided descriptions should be documented and predictable.
- The general recording-errors semantic conventions recommend setting error.type on errored spans and only setting status description when the additional information is not expected to contain sensitive details.
- GenAI semantic conventions model error.type as the low-cardinality provider/client error identifier for failed operations. This fits values like insufficient_quota, authentication_error, invalid_api_key, or rate_limit_error better than raw provider message text.

References:

- https://opentelemetry.io/docs/specs/otel/trace/api/#set-status
- https://opentelemetry.io/docs/specs/semconv/general/recording-errors/#recording-errors-on-spans
- https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-spans.md

## Follow-up work

Decide whether OBI should support exporting raw GenAI provider error messages through an explicit opt-in mechanism.

Possible outcomes:

- Keep the current safe default: status.code = Error, error.type populated, no raw provider text in status.message.
- Add an explicit opt-in config or attribute-selection path for raw provider error text, with documentation that it may contain sensitive provider-supplied data.
- Align the behavior with the DB error path if we choose to expose it.
- Add tests covering default suppression and any opt-in behavior.

MrAlias:
## Summary

- add `gen_ai.response.error` as an exact-selection-only trace control
- copy supported provider error text verbatim to `status.message` only when the resulting span status is `Error`
- keep raw provider text disabled by default and out of sampler and exported span attributes
- preserve existing status classification, `error.type`, database, MCP, and JSON-RPC behavior
- document the privacy implications of enabling provider-supplied error text

## Rationale

Raw provider messages can be valuable for diagnosing quota, authentication, and request failures, but they are unstructured provider-supplied text. They can contain credentials, identifiers, prompts or completions, tenant metadata, request details, customer usage data, billing details, and other sensitive information.

The new control is therefore disabled by default. Operators must explicitly include `gen_ai.response.error`; `*` and `gen_ai.*` do not enable it. Selected messages are copied without additional redaction or truncation, and only for OpenAI, OpenAI-compatible, Anthropic, Gemini, Qwen, AWS Bedrock, and Rerank error responses. The control itself is never exported, and the deprecated `error.message` attribute is not used.

## Validation

- `go test -race ./pkg/export/attributes ./pkg/export/otel/tracesgen ./pkg/appolly/app/request`
- scoped `golangci-lint` for `./pkg/export/attributes` and `./pkg/export/otel/tracesgen`
- Markdown lint

Closes #2507

### DIFF
--- a/devdocs/features.md
+++ b/devdocs/features.md
@@ -51,6 +51,34 @@ chunk.
 | Vector retrieval | `usage.prompt_tokens` or `usage.total_tokens` | Not reported | Most vector stores do not return token usage; it is exported when present. |
 | MCP | Not reported | Not reported | MCP spans do not currently expose token usage. |
 
+## GenAI provider error messages
+
+OBI does not export raw GenAI provider error messages by default. To copy them
+verbatim into the OTLP span `status.message`, explicitly select the
+`gen_ai.response.error` control:
+
+```yaml
+attributes:
+  select:
+    traces:
+      include:
+        - gen_ai.response.error
+```
+
+The control applies to OpenAI, OpenAI-compatible APIs, Anthropic, Google AI
+Studio (Gemini), Qwen, AWS Bedrock, and Rerank responses. It only affects spans
+whose resulting status is `Error`; it does not change status classification or
+the `error.type` attribute. `gen_ai.response.error` is not exported as a span
+attribute, and OBI does not use the deprecated `error.message` attribute.
+
+This sensitive control requires an exact include. Wildcards such as
+`gen_ai.*` and `*` do not enable it. OBI does not additionally redact or
+truncate selected messages. Provider error text can contain credentials,
+account or request identifiers, prompts or completions, tenant metadata,
+request details, customer usage data, account or billing details, and other
+sensitive data. Review backend access, retention, and downstream processing
+before enabling it.
+
 ## Go Instrumentation
 
 Specifically for Go applications, OBI chooses to instrument libraries directly using Uprobes, instead of instrumenting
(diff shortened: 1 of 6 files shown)


mariomac (commented):
Looks good! Pinging copilot for a second pair of "eyes".

mariomac (approved):

PR #2800, opened 2026-07-28
fix(ebpf): never scale array maps in SetupMapSizes

bourbonkk:
## Summary

`SetupMapSizes` scales array maps, but for arrays `max_entries` is the valid index
space, not a capacity with headroom. `valid_pids` is an array bitmap sized by
`k_max_concurrent_pids` (3001) and indexed by generictracer through the mirrored Go
constant, so with `global_scale_factor: -1` it halves to 1500 and every update at or
above that index fails:

```
Error setting up pid in BPF space, sizes of Go and BPF maps don't match
  error="update: key too big for map: argument list too long" i=2995
```

Worse than the log suggests: `pid_matches()` returns 1 (match) on a failed lookup, so
the upper half of the PID hash space silently bypasses the discovery filter.

Fix is to exclude `Array` and `PerCPUArray` in `isResizableMapType`. Type-based rather
than a name denylist, since the invariant holds for every array. No change at the
default `global_scale_factor: 0`. These two types were 0.6 MiB of 294.6 MiB of map
memlock on a test node, so nothing meaningful is lost. `pid_cache` uses the same
constant but stays scalable — it is an LRU hash, so shrinking it only costs hit rate.

Verified with a new test covering shrink and grow (a sibling `Hash` map still resizes),
and on a cluster: 6 of 7 DaemonSet pods logged the error with `global_scale_factor: -1`
and none after removing it. `TestIsResizableMapType` previously asserted `Array` was
resizable; that assertion is updated.

## Validation

- [ ] N/A — map-sizing bug fix, no feature support status change, so [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and the [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) are unchanged.

---

AI policy: I used Claude Code to help analyse the failure and locate the interacting
constants. I reviewed every line, ran the tests, and reproduced both the failure and
the fix on my own cluster.

### DIFF
--- a/pkg/internal/ebpf/convenience/convenience.go
+++ b/pkg/internal/ebpf/convenience/convenience.go
@@ -112,10 +112,13 @@ const (
 
 // isResizableMapType returns true for map types where scaling MaxEntries
 // is meaningful. Excludes special map types whose MaxEntries has fixed
-// semantics (e.g. ProgramArray entries are tail-call slots, not data).
+// semantics (e.g. ProgramArray entries are tail-call slots, not data), and
+// array types, whose MaxEntries is a valid index space rather than a capacity
+// (e.g. valid_pids is indexed using a constant shared with userspace).
 func isResizableMapType(t ebpf.MapType) bool {
 	switch t {
-	case ebpf.ProgramArray, ebpf.PerfEventArray, ebpf.CGroupArray,
+	case ebpf.Array, ebpf.PerCPUArray,
+		ebpf.ProgramArray, ebpf.PerfEventArray, ebpf.CGroupArray,
 		ebpf.ArrayOfMaps, ebpf.HashOfMaps,
 		ebpf.DevMap, ebpf.SockMap, ebpf.CPUMap, ebpf.XSKMap, ebpf.SockHash,
 		ebpf.DevMapHash, ebpf.ReusePortSockArray:
--- a/pkg/internal/ebpf/convenience/convenience_test.go
+++ b/pkg/internal/ebpf/convenience/convenience_test.go
@@ -128,6 +128,32 @@ func TestSetupMapSizes_SkipsNonResizableTypes(t *testing.T) {
 	}
 }
 
+func TestSetupMapSizes_SkipsArrayTypes(t *testing.T) {
+	spec := makeSpec(map[string]*ebpf.MapSpec{
+		"valid_pids":   {Type: ebpf.Array, MaxEntries: 3001},
+		"percpu_array": {Type: ebpf.PerCPUArray, MaxEntries: 1024},
+		"normal":       {Type: ebpf.Hash, MaxEntries: 1024},
+	})
+
+	for _, factor := range []int{-1, 2} {
+		spec.Maps["valid_pids"].MaxEntries = 3001
+		spec.Maps["percpu_array"].MaxEntries = 1024
+
+		SetupMapSizes(spec, factor)
+
+		if got := spec.Maps["valid_pids"].MaxEntries; got != 3001 {
+			t.Errorf("Array must not be resized (factor %d): got %d, want 3001", factor, got)
+		}
+		if got := spec.Maps["percpu_array"].MaxEntries; got != 1024 {
+			t.Errorf("PerCPUArray must not be resized (factor %d): got %d, want 1024", factor, got)
+		}
+	}
+
+	if spec.Maps["normal"].MaxEntries == 1024 {
+		t.Error("Hash map should still have been resized")
+	}
+}
+
 func TestSetupMapSizes_SkipsBelowMinResizableMapSize(t *testing.T) {
 	spec := makeSpec(map[string]*ebpf.MapSpec{
 		"tiny": {Type: ebpf.Hash, MaxEntries: 32},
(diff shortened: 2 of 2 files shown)


pkg/internal/ebpf/convenience/convenience.go
```diff
@@ -113,9 +113,20 @@ const (
 // isResizableMapType returns true for map types where scaling MaxEntries
 // is meaningful. Excludes special map types whose MaxEntries has fixed
 // semantics (e.g. ProgramArray entries are tail-call slots, not data).
+//
+// Array types are excluded because their MaxEntries defines the valid index
```
marctc: Please review AI policy as this gigantic comments are mostly not needed.

marctc:
Thanks for your contribution. I'd suggest that PR description is meant to be human readable. We have many PRs to review and we can't spend so much time reading such walls of text, specially when a PR is only 44 lines of additions.

marctc (changes requested):

pkg/internal/ebpf/convenience/convenience.go
bourbonkk: Sorry about that — I've shortened both the code comments and the PR description. PTAL.

marctc (approved):
Thanks!

pkg/internal/ebpf/convenience/convenience_test.go
```diff
@@ -183,7 +210,7 @@ func TestIsResizableMapType(t *testing.T) {
 		}
 	}
 
-	resizable := []ebpf.MapType{ebpf.Hash, ebpf.Array, ebpf.LRUHash, ebpf.RingBuf}
+	resizable := []ebpf.MapType{ebpf.Hash, ebpf.LRUHash, ebpf.PerCPUHash, ebpf.RingBuf}
```
mmat11: should we also have `LRUCPUHash`, for consistency?

pkg/internal/ebpf/convenience/convenience_test.go
```diff
@@ -128,6 +128,32 @@ func TestSetupMapSizes_SkipsNonResizableTypes(t *testing.T) {
...
 }
 
+func TestSetupMapSizes_SkipsArrayTypes(t *testing.T) {
+	spec := makeSpec(map[string]*ebpf.MapSpec{
+		"valid_pids":   {Type: ebpf.Array, MaxEntries: 3001},
+		"percpu_array": {Type: ebpf.PerCPUArray, MaxEntries: 1024},
+		"normal":       {Type: ebpf.Hash, MaxEntries: 1024},
+	})
+
+	for _, factor := range []int{-1, 2} {
```
pinoOgni: I see that `valid_pids/percpu_array` are reset per iteration but `normal` isn't, so the control assertion depends on the two factors not cancelling out. It holds for `{-1, 2}`(`1024 >> 1 = 512, 512 << 2 = 2048`), but if the factor list ever changes such that they cancel like `{-1, 1}` gives `512 << 1 = 1024` and the test fails with "Hash map should still have been resized" while the code is correct.

We can do something like:
```
for _, tc := range []struct {
	factor     int
	wantNormal uint32
}{
	{-1, 512},
	{2, 4096},
} {
	spec := makeSpec(map[string]*ebpf.MapSpec{
		"valid_pids":   {Type: ebpf.Array, MaxEntries: 3001},
		"percpu_array": {Type: ebpf.PerCPUArray, MaxEntries: 1024},
		"normal":       {Type: ebpf.Hash, MaxEntries: 1024},
	})

	SetupMapSizes(spec, tc.factor)

	if got := spec.Maps["valid_pids"].MaxEntries; got != 3001 {
		t.Errorf("Array must not be resized (factor %d): got %d, want 3001", tc.factor, got)
	}
	if got := spec.Maps["percpu_array"].MaxEntries; got != 1024 {
		t.Errorf("PerCPUArray must not be resized (factor %d): got %d, want 1024", tc.factor, got)
	}
	if got := spec.Maps["normal"].MaxEntries; got != tc.wantNormal {
		t.Errorf("Hash must be resized (factor %d): got %d, want %d", tc.factor, got, tc.wantNormal)
	}
}
```
We construct the spec inside the loop, so every factor starts from the same known state. The expected normal value becomes part of each case, so the control asserts an exact number per factor. 

WDYT?

pinoOgni (changes requested):
Hi @bourbonkk nice catch!

pkg/internal/ebpf/convenience/convenience_test.go
bourbonkk: Good point, added to the test — it already falls through to the resizable default, so this is just closing the coverage gap.

pkg/internal/ebpf/convenience/convenience_test.go
bourbonkk: Agreed, the control assertion shouldn't depend on the factors not cancelling out. Applied your suggestion: the spec is now built inside the loop and each case asserts an exact wantNormal. Thanks for the catch!

mmat11 (approved):

pinoOgni (approved):

mmat11:
@bourbonkk thanks!

PR #2804, opened 2026-07-28
fix(embedding): complete embeddings semconv attributes across all providers

NameHaibinZhang:
## Summary

Aligns GenAI embeddings spans with the OTel GenAI semantic conventions across all embedding paths, and fixes two cases where the dimension count / encoding format were dropped.

## Validation

### DIFF
--- a/pkg/appolly/app/request/span.go
+++ b/pkg/appolly/app/request/span.go
@@ -819,13 +819,46 @@ func (e *VendorEmbedding) OperationName() string {
 	return EmbeddingOperationName
 }
 
+// Dimensions returns the output vector dimension count. It prefers an explicit
+// request dimension ("dimensions" or "output_dimension"), falling back to the
+// length derived from the response payload. Returns 0 when not determinable.
+func (e *VendorEmbedding) Dimensions() int {
+	if e.Input.Dimensions > 0 {
+		return e.Input.Dimensions
+	}
+	if e.Input.OutputDimension > 0 {
+		return e.Input.OutputDimension
+	}
+	return e.Output.Dimensions
+}
+
 // EmbeddingRequest captures the common fields from embedding API requests.
 type EmbeddingRequest struct {
 	Model      string          `json:"model"`
 	Input      json.RawMessage `json:"input"`
 	Dimensions int             `json:"dimensions,omitempty"`
+	// Voyage AI and Cohere use "output_dimension" for the requested vector size.
+	OutputDimension int `json:"output_dimension,omitempty"`
 	// Cohere uses "texts" instead of "input"
 	Texts json.RawMessage `json:"texts,omitempty"`
+	// EncodingFormat is the OpenAI/Voyage/Jina style single-value output format.
+	EncodingFormat string `json:"encoding_format,omitempty"`
+	// EmbeddingTypes is the Cohere v2 style list of output formats.
+	EmbeddingTypes []string `json:"embedding_types,omitempty"`
+}
+
+// EncodingFormats returns the requested output encoding formats, normalizing
+// across providers: OpenAI/Voyage/Jina expose a single "encoding_format"
+// string, while Cohere v2 uses an "embedding_types" array. Returns nil when
+// no format was specified.
+func (r *EmbeddingRequest) EncodingFormats() []string {
+	if len(r.EmbeddingTypes) > 0 {
+		return r.EmbeddingTypes
+	}
+	if r.EncodingFormat != "" {
+		return []string{r.EncodingFormat}
+	}
+	return nil
 }
 
 // InputCount returns the number of input texts in the request.
@@ -853,6 +886,9 @@ type EmbeddingResponse struct {
 	Usage EmbeddingUsage `json:"usage"`
 	// Cohere uses meta.billed_units for token counts
 	Meta *CohereResponseMeta `json:"meta,omitempty"`
+	// Dimensions is the length of a single output vector, derived from the
+	// response payload. Populated by the embedding parser; zero when unknown.
+	Dimensions int `json:"-"`
 }
 
 // EmbeddingUsage captures token usage in embedding responses.
--- a/pkg/appolly/app/request/span.go
+++ b/pkg/appolly/app/request/span.go
@@ -336,6 +336,12 @@ type OpenAIError struct {
 type ToolCall struct {
 	ID   string `json:"id,omitempty"`
 	Name string `json:"name"`
+	// Arguments holds the tool-call input as a JSON string. OpenAI-family
+	// function.arguments is already a JSON string; Anthropic input / Gemini
+	// args / Ollama arguments are JSON objects stored as their raw encoding.
+	Arguments string `json:"arguments,omitempty"`
+	// Type is the tool type, e.g. "function".
+	Type string `json:"type,omitempty"`
 }
 
 type VendorOpenAI struct {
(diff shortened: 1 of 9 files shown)


MrAlias (commented):
Binary Cohere embeddings currently produce an incorrect dimension count.

pkg/ebpf/common/http/embedding.go
```diff
@@ -97,3 +101,41 @@ func EmbeddingSpan(baseSpan *request.Span, req *http.Request, resp *http.Respons
...
+	}
+
+	// Cohere v2 layout: {"embeddings":{"float":[[...]]}}
+	var cohereStyle struct {
+		Embeddings map[string][][]json.Number `json:"embeddings"`
+	}
+	if unmarshalJSON(body, &cohereStyle) {
+		for _, vectors := range cohereStyle.Embeddings {
+			if len(vectors) > 0 {
+				if n := len(vectors[0]); n > 0 {
```
MrAlias: When Cohere returns `binary` or `ubinary`, each array entry packs eight model dimensions, so `len(vectors[0])` is 128 for a 1024-dimensional vector. With both `float` and `binary` requested, map iteration can also choose either count. This is exported as `gen_ai.embeddings.dimension.count`; use the encoding key to expand packed binary values, or prefer a non-packed vector.

NameHaibinZhang: Good catch, thanks! Fixed: the Cohere v2 branch now resolves dimensions per embedding-type key instead of iterating the map. Non-packed vectors (float, int8, uint8) are preferred since their entry count equals the dimension count, and when only binary/ubinary are present the entry count is expanded by 8 (each byte packs eight model dimensions). This also makes the result deterministic when multiple embedding types are requested. Added regression tests covering binary, ubinary, and the float+binary mixed case.

pkg/ebpf/common/http/embedding.go
```diff
@@ -97,3 +101,78 @@ func EmbeddingSpan(baseSpan *request.Span, req *http.Request, resp *http.Respons
...
+// embeddings.{float,int8,...}[][] layout. Returns 0 when not determinable.
+func parseEmbeddingDimensions(body []byte) int {
+	if len(body) == 0 {
+		return 0
+	}
+
+	// OpenAI-style layout: {"data":[{"embedding":[...]}]}
+	var openAIStyle struct {
+		Data []struct {
+			Embedding []json.Number `json:"embedding"`
```
MrAlias: When no explicit dimension is sent, this assumes every OpenAI-shaped embedding is a numeric vector whose element count equals its dimension. Voyage `binary`/`ubinary` vectors are bit-packed, while Voyage and Jina base64 responses are strings; Cohere's `base64` representation is also skipped by the keyed branch below. Valid responses therefore export a dimension that is wrong or absent. Could we derive the dimension from the provider and requested representation, with binary and base64 coverage for each supported provider?

pkg/appolly/app/request/span.go
```diff
@@ -819,13 +951,46 @@ func (e *VendorEmbedding) OperationName() string {
...
 // EmbeddingRequest captures the common fields from embedding API requests.
 type EmbeddingRequest struct {
 	Model      string          `json:"model"`
 	Input      json.RawMessage `json:"input"`
 	Dimensions int             `json:"dimensions,omitempty"`
+	// Voyage AI and Cohere use "output_dimension" for the requested vector size.
+	OutputDimension int `json:"output_dimension,omitempty"`
 	// Cohere uses "texts" instead of "input"
 	Texts json.RawMessage `json:"texts,omitempty"`
+	// EncodingFormat is the OpenAI/Voyage/Jina style single-value output format.
```
MrAlias: Jina uses `embedding_type` (a string or array), not `encoding_format`, and Voyage carries the requested element representation in `output_dtype`. Both fields are ignored here, so valid requests omit part or all of `gen_ai.request.encoding_formats`. Could we model those native fields and normalize them alongside Cohere's `embedding_types`?

pkg/ebpf/common/http/embedding.go
```diff
@@ -97,3 +101,78 @@ func EmbeddingSpan(baseSpan *request.Span, req *http.Request, resp *http.Respons
 
 	return *baseSpan, true
 }
+
+// parseEmbeddingDimensions inspects a raw embedding response body and returns
+// the model dimension count of a single output vector. It supports the OpenAI-style
+// data[].embedding layout (Voyage, Jina) and the Cohere v2
+// embeddings.{float,int8,...}[][] layout. Returns 0 when not determinable.
+func parseEmbeddingDimensions(body []byte) int {
```
MrAlias: This parser requires the complete response envelope before it can use `data[0].embedding`. OBI captures at most 256 KiB, so a valid batch can contain a complete first vector followed by a truncated later vector or field; this then returns zero despite having enough data. Could we recover the first complete vector using the same best-effort approach used for the usage fields?

pkg/ebpf/common/http/qwen_test.go
```diff
@@ -114,6 +116,91 @@ func TestQwenSpan_DashScopeGeneration(t *testing.T) {
...
+	respBody := fmt.Sprintf(
+		`{"data":[{"embedding":"%s","index":0,"object":"embedding"}],"model":"text-embedding-v3","usage":{"prompt_tokens":6,"total_tokens":6},"id":"eb4ea05e-18c5-9554-9234-78a1528e7be3"}`,
+		vec)
+	reqBody := `{"model":"text-embedding-v3","input":["hello"],"encoding_format":"base64"}`
+
+	req := makeRequest(t, http.MethodPost, "https://dashscope.aliyuncs.com/compatible-mode/v1/embeddings", reqBody)
+	resp := makePlainResponse(http.StatusOK, qwenHeaders(), respBody)
+
+	base := &request.Span{}
+	span, ok := QwenSpan(base, req, resp)
```
MrAlias, line 189: This direct `QwenSpan` call bypasses the production detector order. For a headerless HTTP/2 request to DashScope's compatible `/v1/embeddings` endpoint with a `text-embedding-*` model, `OpenAISpan` runs first and claims the request, so this path exports `gen_ai.provider.name=openai` and never reaches Qwen. Could we cover this through `postProcessHTTPSpan` and adjust the detector priority or fallback heuristic?

MrAlias (commented):
Provider-specific representations and the production dispatch path still leave valid embedding requests with incorrect or missing semantic-convention attributes. The response fallback also needs to account for OBI's bounded payload capture.

pkg/ebpf/common/http/embedding.go
NameHaibinZhang: Fixed. parseEmbeddingDimensions is now representation-aware: the requested dtype (from encoding_format, Voyage's output_dtype, or Jina's embedding_type) is threaded into the vector decoding. Numeric arrays with a binary/ubinary dtype are expanded x8 (bit-packed), and base64 string vectors are decoded and sized per dtype (int8/uint8 -> bytes, binary/ubinary -> bytes x8, float -> bytes / 4). Cohere's embeddings.base64 key is handled the same way instead of being skipped. Covered by TestEmbeddingSpan_RepresentationAwareDimensions.

pkg/appolly/app/request/span.go
NameHaibinZhang: Fixed. EmbeddingRequest now models Jina's embedding_type (string-or-array, mirroring the existing stop-sequences pattern) and Voyage's output_dtype. EncodingFormats() normalizes across all provider vocabularies (Cohere embedding_types -> Jina embedding_type -> Voyage output_dtype plus encoding_format -> OpenAI encoding_format), so gen_ai.request.encoding_formats is no longer missing for Jina/Voyage calls. A new RequestedDtype() helper feeds the dimension decoding. Covered by TestEmbeddingRequest_NativeFormatFields.

pkg/ebpf/common/http/embedding.go
NameHaibinZhang: Fixed. The response envelope is now parsed with the existing best-effort jsoniter path (unmarshalJSONBestEffort) and the first vector is captured as json.RawMessage, so a batch response truncated at the 256KiB capture limit after the first complete vector still yields the dimension count. If the truncation lands inside the first vector itself, the strict decode of that raw fragment fails and we report 0 rather than a wrong count. Covered by TestParseEmbeddingDimensions_TruncatedBatch.

pkg/ebpf/common/http/qwen_test.go
NameHaibinZhang: Good catch - the heuristic did have that gap. looksLikeOpenAIBody now yields on DashScope's text-embedding-v<N> model naming and looksLikeQwenBody claims it, so the production detector order routes headerless HTTP/2 DashScope embeddings to Qwen instead of OpenAI. Added TestGenAIHTTP2DashScopeEmbeddingDetectorOrder, which replays the detectors in production order (EmbeddingSpan -> OpenAISpan -> QwenSpan) against a headerless text-embedding-v3 request and asserts OpenAI declines and Qwen claims it.

MrAlias: The detector-order test still does not exercise the production request shape. `parseHTTP2Request` always sets both `Host` and `URL.Host`; once that host is present, `isQwenCompatibleURL` sets `urlDetected`, leaving `maybeQwen` false. The URL-only guard in `qwen.go` still accepts only `qwen*`, so `text-embedding-v3` is rejected after OpenAI yields. I reproduced this on `ce9058b`: with `dashscope.aliyuncs.com` populated, both detectors return false. Could we include the host in this fixture and allow `isDashScopeEmbeddingModel(model)` in that guard?

MrAlias:
Thanks for the follow-up. This is looking close. There was still [one follow-up](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2804#discussion_r3676385288) I identified from the fix to the last review.

pkg/ebpf/common/http/qwen_test.go
NameHaibinZhang: You're right - thanks for the repro. With the host populated (as parseHTTP2Request always does), isQwenCompatibleURL kept maybeQwen false and the URL-only guard rejected text-embedding-v3, so both detectors returned false. Fixed both suggestions: the URL-only guard in QwenSpan now accepts isDashScopeEmbeddingModel(model) alongside the qwen* prefix, and TestGenAIHTTP2DashScopeEmbeddingDetectorOrder now runs two variants - a production shape with Host/URL.Host set to dashscope.aliyuncs.com (mirroring parseHTTP2Request) and the hostless degraded capture. The host variant failed before the guard change and passes after it.

NameHaibinZhang:
@MrAlias all comment addressed,thks

marctc (commented):
nit: i think newer versions of Claude love to add massive comments, do you mind reducing a bit size or use comments only when help to understand the code? :) also some of field in structs have comments and other not (like [here](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2804/changes#diff-0848e3c4c594f62828b016498b280f1044960fe1bd9a7b5063e451d13d06b495R534-R537)), so consistency is always good.

NameHaibinZhang:
> nit: i think newer versions of Claude love to add massive comments, do you mind reducing a bit size or use comments only when help to understand the code? :) also some of field in structs have comments and other not (like [here](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2804/changes#diff-0848e3c4c594f62828b016498b280f1044960fe1bd9a7b5063e451d13d06b495R534-R537)), so consistency is always good.

Fixed, thanks! Trimmed the redundant comments across the PR

marctc (approved):

MrAlias (approved):

PR #2805, opened 2026-07-28
Fix v2 routes conversion

mariomac:
## Summary

Fix V1→V2 config migration losing global `routes` settings. When a V1 config sets a global `routes` block (`unmatched`, `patterns`, `ignored_patterns`, `ignore_mode`), it exports to symmetric `incoming`/`outgoing` V2 policies, but the importer always rebuilt a `Directional` config and left the global `RoutesConfig` fields empty — so the round-trip dropped those fields and the migration contract check rejected them as unsupported. `applyFullV2HTTPRoutes` now collapses identical incoming/outgoing policies back into a single global `RoutesConfig`, which is semantically equivalent but preserves the representation the user authored, making the round-trip lossless. Also updates `TestV2ToRuntimeHTTPRoutesRoundTrip`, which had asserted the old always-directional behavior, to expect the preserved global form while still verifying directional-policy equivalence.

## Validation

### DIFF
--- a/internal/config/convert/import.go
+++ b/internal/config/convert/import.go
@@ -1052,12 +1052,31 @@ func applyFullV2HTTPRoutes(cfg *obi.Config, routes schema.HTTPRoutes) {
 
 	policies := services.DirectionalRoutePolicies{}
 	applyV2HTTPRouteConfig(&policies, routes)
+	// A v1 global routes block exports identical incoming/outgoing policies.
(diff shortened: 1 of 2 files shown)


NimrodAvni78 (approved):

NameHaibinZhang:
Nice

NameHaibinZhang:
can merge this pr ? @NimrodAvni78  make main test success.

PR #2809, opened 2026-07-28
Preserve HTTP server trace through response body

### LINKED ISSUE #1234: Python instrumentation fails to propagate context for internal postgres call
( see https://cloud-native.slack.com/archives/C08P9L4FPKJ/p1769773446698569 )

I am trying to instrument a process that is a http server, which then fires a SQL request via pycopg2 library. The SQL-span is found by instrumentation, but the parentSpanID is 00...000 . For the http part, the trace/span information is correct.

[{"type":"SQLClient","peer":"::1","peerPort":"47880","host":"::1","hostPort":"5432","traceID":"fee813656741ca153963f288b5f17914","spanID":"894a6958f6fb9469","parentSpanID":"0000000000000000","traceFlags":"0","peerName":"python3.14","hostName":"::1","kind":"SPAN_KIND_CLIENT","start":"1769773261046753","handlerStart":"1769773261046753","end":"1769773261046818","duration":"65.463µs","durationUSec":"65","handlerDuration":"65.463µs","handlerDurationUSec":"65","attributes":{"errorCode":"0","errorDescription":"","errorMessage":"","operation":"SELECT","serverAddr":"::1","serverPort":"5432","sqlCommand":"","sqlState":"","statement":"SELECT * FROM tea WHERE kind=('sencha');","table":"tea"}}

Sample server code is at https://gist.github.com/pilhuhn/1f3e853ae025e63a2858ac060c9454ab

This is with the v0.4.1 source download, locally built on RHEL9

marctc:
## Summary

Fixes #1234 where a PostgreSQL span created inside a Python HTTP handler had an empty `parentSpanID`.

The HTTP server trace was removed when response headers were observed, so SQL executed afterward could no longer find its parent. This keeps the trace available for late child spans while still allowing the next request on the thread to replace it. It also cleans up abandoned incomplete requests to avoid stale traces.

## Validation


bpf/generictracer/protocol_http.h
```diff
@@ -344,6 +407,9 @@ static __always_inline void handle_http_response(unsigned char *small_buf,
     } else {
         bpf_dbg_printk("Delaying finish http for large request, orig_len=%d", orig_len);
         info->delayed = 1;
```
MrAlias, line 399: For HTTPS responses, `handle_ssl_buf()` still calls `cleanup_trace_info_for_delayed_trace()` before this path sees the headers. That deletes the matching `server_traces` entry, so this lookup is empty and SQL or other child spans created while the response body is written remain orphaned.

Defer the TLS cleanup until the delayed response finishes, and add the same regression coverage for HTTPS.

MrAlias (commented):
The HTTP lifecycle now differs by transport: TLS retains the old header-time cleanup. That leaves the parent-span regression unresolved for HTTPS workloads.

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -776,6 +777,37 @@ func kernelTime(ktime uint64) time.Time {
...
+				"pid", key.P_key.Pid, "tid", key.P_key.Tid, "ns", key.P_key.Ns, "extraID", key.ExtraId)
+		}
+		return
+	}
+
+	if trace.Tp.TraceId != info.Tp.TraceId || trace.Tp.SpanId != info.Tp.SpanId {
+		return
+	}
+
+	if err := p.bpfObjects.ServerTraces.Delete(&key); err != nil && !errors.Is(err, ebpf.ErrKeyNotExist) {
```
MrAlias: The ownership check and deletion are separate map operations. If a new request replaces this thread key after `Lookup` returns but before `Delete` runs, the timeout cleanup removes the new request's trace and leaves its child spans orphaned.

Could we move this cleanup to a path where removal remains tied to the matching trace identity, or otherwise avoid deleting a reusable key from userspace?

MrAlias (commented):
The TLS lifecycle is now consistent. The userspace timeout path still has an ownership race that can break parent attribution.

pkg/internal/ebpf/generictracer/generictracer.go
marctc: Good catch. I removed the userspace cleanup since the lookup and delete aren’t atomic. Normal cleanup remains in BPF; timeout-only entries may linger until replacement or LRU eviction. A fully safe timeout cleanup needs a separate BPF-owned solution, wdyt?

bpf/generictracer/protocol_http.h
```diff
@@ -344,6 +407,9 @@ static __always_inline void handle_http_response(unsigned char *small_buf,
...
+        if (info->type == EVENT_HTTP_REQUEST) {
+            mark_http_server_thread_trace_response_sent(info);
```
MrAlias: With `ebpf.high_request_volume=true` on plaintext traffic, this delayed branch is skipped and `finish_http()` removes `server_traces` when the headers are observed. That leaves post-header child spans orphaned in this opt-in mode, although the default path covered by this PR is fixed.

Is that an intentional tradeoff for `high_request_volume`? If not, could we keep the parent context alive independently of early HTTP event emission and add coverage for this mode?

MrAlias (commented):
The default Python and TLS paths now cover the reported regression; the remaining concern is limited to the opt-in immediate-emission mode.

bpf/generictracer/protocol_http.h
```diff
@@ -113,31 +113,105 @@ static __always_inline void cleanup_http_info(pid_connection_info_t *pid_conn) {
...
+    return existing;
+}
+
+static __always_inline void cleanup_http_server_thread_trace(http_info_t *info) {
+    trace_key_t t_key = {0};
+    if (!http_server_thread_trace(info, &t_key)) {
+        return;
+    }
+
+    int res = bpf_map_delete_elem(&server_traces, &t_key);
```
MrAlias: `http_server_thread_trace()` validates request A's value, but this delete is a separate operation on the reusable thread key. Once A has set `response_sent`, request B can replace that key; if B's update lands between the lookup and this delete while A's cleanup runs on another CPU, this removes B's parent context and leaves its child spans orphaned.

The window is narrow, but there is no synchronization around the check-and-delete. Could we serialize this transition, or add concurrent cross-thread coverage demonstrating that this interleaving cannot occur?

MrAlias (commented):
The reported regression is now covered across plaintext, TLS, and high-request-volume paths. I still see one narrow concurrency issue in the cleanup lifecycle.

bpf/generictracer/http_server_trace.h
```diff
@@ -0,0 +1,57 @@
...
+}
+
+static __always_inline void
+cleanup_http_server_thread_trace_for_key(http_info_t *info, const trace_key_t *current_key) {
+    trace_key_t t_key = {0};
+    if (!http_server_thread_trace(info, &t_key)) {
+        return;
+    }
+
+    if (info->delayed &&
```
MrAlias, line 47: In the current verifier run, `obi_kprobe_tcp_close` reaches 1,000,001 processed instructions across multiple constant combinations on both `bpf` and `bpf-next`, so generictracer is rejected before loading. Main at the PR's exact base SHA passes those same kernel images, which isolates the regression to this PR.

This ownership guard is on the heavily inlined close path and is the likely place to start, but the logs do not isolate it as the sole cause. Could we profile or restructure this cleanup path and restore the tip-kernel verifier jobs?

MrAlias (commented):
The ownership change closes the cleanup race, but generictracer still does not verify across the configured kernel matrix.

MrAlias (approved):

PR #2810, opened 2026-07-28
Activate Go Auto SDK instrumentation

MrAlias:
## Summary

- activate the global OpenTelemetry Go tracer through the Auto SDK when no SDK delegate is registered and the merged eligibility gate confirms every required helper, symbol, version, checksum, architecture, and `trace.SpanContext` offset
- populate parent and child span contexts from OBI's current goroutine state, track provisional and committed Auto SDK lifecycles, and preserve synthetic instrumentation as the fallback
- attach `Tracer.newSpan` per process after the executable-scoped activation probes succeed, then retire only the matching PID and generation link after BPF acknowledges activation
- emit bounded rich Auto SDK spans through the flexible record ABI, with lifecycle cleanup before every payload drop path
- cover the lower supported OpenTelemetry Go v1.33.0 / Auto SDK v1.1.0 pair and a Renovate-managed rolling-latest fixture

## Motivation

#2819 added the fail-closed Go Auto SDK ABI eligibility and offset gate. #2820 added atomic optional Go probe groups, post-commit process-scoped registration, internal control-event dispatch, and generation-safe executable teardown.

This change builds on those merged prerequisites to complete activation for applications using the global tracer without another registered SDK. If activation cannot be completed safely, the application remains in the existing synthetic instrumentation mode.

## Behavior

The Auto SDK `start` probe reserves provisional lifecycle state before touching application memory. It leaves the SDK's `sampled` argument at its default `true`, writes a nonzero parent context when one exists, always writes the child context, and commits the lifecycle only after all required writes succeed. Root spans receive valid sampled trace and span IDs.

The `ended` probe restores the previous OBI goroutine context and deletes lifecycle state before validating or emitting the bounded JSON payload. Oversized payloads and ring-buffer failures therefore drop only the rich event without leaving committed state behind. Existing synthetic probes suppress output only for committed Auto SDK spans.

`Tracer.newSpan` is attached with `UprobeOptions.PID`. Process start time and the pinned `/proc/<pid>/exe` device/inode are validated around attachment so PID reuse, exec races, and cross-device inode collisions fail closed. BPF acknowledges the matching PID and generation after activation, allowing OBI to detach only that temporary link. The existing three-attempt activation retry policy is unchanged.

The implementation intentionally retains current same-goroutine LIFO semantics and existing duplicate-output behavior across multiple OBI instances. It does not add asynchronous context traversal, trace-native provider activation, cross-instance ownership, failure counters, or new configuration.

## Compatibility

The activation path covers OpenTelemetry Go v1.33.0 through the rolling-latest integration fixture. Unsupported or replaced modules, unreviewed checksums, missing symbols or offsets, unsupported architectures, or unavailable helper support leave activation disabled.

## Validation

- `make generate`
- `make clang-format`
- `make clang-tidy`
- `make lint`
- `make test GOFLAGS=-p=1`
- `make license-header-check`
- `make compile GOFLAGS=-p=1`
- `GOOS=darwin GOARCH=arm64 CGO_ENABLED=0 GOFLAGS=-p=1 go test -vet=off -c ./pkg/internal/ebpf/gotracer`

## Runtime performance

Head activates Go Auto SDK instrumentation. Each `/spans` request creates one application-authored root span and one application-authored child span through the global OpenTelemetry tracer; OBI captures and serializes both. Base leaves the Auto SDK inactive and emits the existing synthetic HTTP span. Head versus base includes Auto SDK activation, span capture, and serialization.

### Setup

- Revisions: base c1f1e06485cb5d0a7c0e584f17ad4675f4dc6dfd; head 3555910967c102b15291bea4eb00df13d894ae6f
- Workload: one binary built from the head `goautosdk/latest` fixture with `OTEL_TEST_VERSION=latest`, reused in all conditions
- Load: `GET http://127.0.0.1:8090/spans`, 500 requests/second, 30-second discarded warm-up, 60-second measurement, 20 preallocated VUs, 100 maximum VUs
- Pinning: application, OBI, and k6 on three physical cores without selected SMT siblings
- Runs: five per condition, interleaved across no OBI, base OBI, and head OBI; application and OBI restarted before each run
- OBI: no OTLP exporter; `OTEL_EBPF_TRACE_PRINTER=counter`; `OTEL_EBPF_BPF_CONTEXT_PROPAGATION=all`

All 15 runs reached 500 requests/second with zero failed requests and zero dropped iterations. `/activation` returned 204 with head and 503 with base or no OBI.

Results are median [minimum–maximum] across five runs.

| Condition | Achieved rps | Avg latency | P50 | P90 | P95 | P99 |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| No OBI | 500.0068 [500.0018–500.0113] | 0.4819 ms [0.4602–0.4945] | 0.4791 ms | 0.6090 ms | 0.6868 ms | 1.7056 ms |
| Base OBI | 500.0071 [500.0042–500.0077] | 0.7334 ms [0.7093–0.7618] | 0.7358 ms | 1.0311 ms | 1.2268 ms | 1.9655 ms |
| Head OBI | 500.0023 [499.9958–500.0036] | 0.8118 ms [0.7779–0.8322] | 0.7673 ms | 1.1177 ms | 1.3596 ms | 2.0363 ms |

| Condition | Application CPU | OBI CPU | Total CPU | Application peak RSS | OBI peak RSS | Sum peak RSS | Processed spans |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| No OBI | 7.00 s [6.54–7.26] | — | 7.00 s [6.54–7.26] | 16.70 MiB | — | 16.70 MiB [14.93–16.89] | — |
| Base OBI | 17.44 s [16.60–18.05] | 1.22 s [1.20–1.26] | 18.70 s [17.80–19.26] | 14.75 MiB | 118.17 MiB | 132.91 MiB [132.49–135.20] | 30,003 |
| Head OBI | 19.98 s [19.30–20.69] | 3.56 s [3.40–3.66] | 23.63 s [22.75–24.35] | 15.23 MiB | 122.06 MiB | 139.01 MiB [133.71–141.40] | 89,889 [89,666–90,003] |

### Deltas

Percentages compare condition medians.

| Comparison | Avg | P50 | P90 | P95 | P99 | Application CPU | Total CPU | Sum peak RSS |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Existing OBI: base vs no OBI | +52.2% | +53.6% | +69.3% | +78.6% | +15.2% | +149.1% | +167.1% | +695.9% |
| **Incremental PR: head vs base** | **+10.7%** | **+4.3%** | **+8.4%** | **+10.8%** | **+3.6%** | **+14.6%** | **+26.4%** | **+4.6%** |
| Total instrumented: head vs no OBI | +68.4% | +60.2% | +83.5% | +98.0% | +19.4% | +185.4% | +237.6% | +732.4% |

Head minus base:

- Average latency: **+0.0783 ms**
- P95 latency: **+0.1328 ms**
- Application plus OBI CPU: **+4.93 seconds per 60-second run**, or 164 µs/request
- Sum peak RSS: **+6.09 MiB**
- Achieved rate: **-0.00097%**

Head processed 89,666–90,003 spans for 30,000–30,001 requests. The text printer showed three spans per request: the application-authored root span, its application-authored child span, and a separate synthetic `HTTP GET /spans` server span enabled by the open-port configuration.

Host: Linux 7.1.3; Intel Core i7-8550U with four physical cores and SMT; Go 1.26.5; clang 22.1.8; k6 2.0.0; Intel P-state with the `powersave` governor. CPU time is the change in process user-plus-system ticks; `perf` was unavailable. Sum peak RSS adds independent process `VmHWM` values and is not a simultaneous memory peak. Results apply to this host and load.

Part of #2258.


MrAlias:
Corrected the compatibility matrix so the lower lane remains pinned at OpenTelemetry Go v1.33.0 / Auto SDK v1.1.0, while the upper lane is now a Renovate-managed `latest` fixture (currently OpenTelemetry Go v1.44.0 / Auto SDK v1.2.1). The fixture-specific rule re-enables patch and minor updates after the repository-wide test dependency patch-disable rule.

I also aligned the oversized-payload assertion with the existing fallback behavior: the test permits an attribute-free synthetic span, rejects any rich span carrying `oversized.value`, and still verifies lifecycle cleanup before the next root span.

Both matrix lanes pass the privileged end-to-end test against the branch’s exact OBI image, and two fresh independent adversarial reviews passed the complete updated PR snapshot.

MrAlias:
Extended this draft with process-scoped Auto SDK activation probe ownership.

- The `start`, `context.WithValue`, and `ended` lifecycle probes remain attached at executable scope. All four activation symbols are still resolved and preflighted atomically, but `Tracer.newSpan` is attached only from `AllowPID` with `UprobeOptions.PID`; no attach cookies are used.
- Each temporary link is keyed by host PID and `go_auto_targets` generation, while its probe recipe and executable ownership are scoped by device/inode. Process start time and the pinned `/proc/<pid>/exe` device/inode are validated around attachment so PID reuse, exec races, and cross-device inode collisions fail closed.
- BPF emits an internal PID/generation acknowledgement after a successful `bpf_probe_write_user` and whenever it later observes `auto_span == true`. The shared ring-buffer context dispatches this event independently of which tracer owns the reader, including before the Go tracer's `Run` method starts.
- Only the exact current PID/generation link is retired. Stale events are ignored, and `BlockPID`, executable teardown, and tracer shutdown close remaining links idempotently without affecting another PID using the same executable.
- The existing three-attempt activation retry policy is unchanged. This does not add activation failure metrics or broader failure counting.

Added focused coverage for PID-scoped options, duplicate `AllowPID`, identity retry, stale generations, PID reuse, same-inode multi-process isolation, cross-device inode collisions, pre-`Run` event dispatch, generic-first shared-ring dispatch, and competing cleanup paths. The Go 1.33 lower-bound and rolling-latest integration fixtures remain in the matrix.

Validation passed: BPF generation, clang formatting/tidy, full lint, Linux compilation, focused package tests, and race tests for the Go tracer, generic tracer, and eBPF orchestration. The complete final diff also passed independent kernel/ABI, concurrency/resource-ownership, and integration/release reviews with no remaining findings.

MrAlias:
I split the reusable prerequisites out of this draft so the activation-specific change can be reviewed independently:

- #2819 adds the fail-closed Go Auto SDK ABI eligibility and offset gate.
- #2820 adds atomic optional Go probe groups, post-commit process-scoped registration, internal control-event dispatch, and generation-safe executable teardown.

This draft remains unchanged while those prerequisites are reviewed. After both merge, I will rebase it onto `main` and reduce it to the activation-specific BPF state machine and context writes, temporary PID-scoped `newSpan` implementation, control-event-driven detach, and the Go 1.33 plus rolling-latest integration matrix.

That sequencing avoids a stacked diff that would continue to show the prerequisite code here before it lands, while giving reviewers two independently buildable changes now.

grcevski (approved):
LGTM! Looks great

MrAlias:
@NimrodAvni78 did you still want to take a look at this? Or, are you okay with merging?

bpf/gotracer/go_sdk.c
```diff
@@ -61,13 +88,39 @@ struct {
...
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __type(key, u32);
+    __type(value, u64);
+    __uint(max_entries, k_max_concurrent_pids);
+    __uint(pinning, OBI_PIN_INTERNAL);
+} go_auto_targets SEC(".maps");
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
```
mmat11, line 108: should these maps be HASH_LRU instead?

mmat11 (commented):
should we run the benchmark in `internal/test/benchmarks` to understand the overhead against main (baseline)?

bpf/gotracer/go_sdk.c
MrAlias: These need to remain `BPF_MAP_TYPE_HASH` maps because eviction would change correctness, not just capacity behavior.

- `go_auto_activation_attempts` enforces the three-attempt limit for each PID/generation. LRU eviction could restore an exhausted attempt budget.
- `go_auto_targets` is managed explicitly by userspace. Evicting an entry would silently deactivate a live target without running its teardown.
- `active_go_auto_spans` retains the previous goroutine context until `End`. Evicting an in-flight entry could prevent that context from being restored.

When capacity is exhausted, failing the new insertion is safer than evicting unrelated live state. The existing end, deactivation, and process-cleanup paths remove these entries explicitly.

MrAlias:
@mmat11: benchmarks were added to the description

mmat11 (approved):
lgtm!

PR #2799, opened 2026-07-28
Document Config v2 migration

MrAlias:
## Summary

- document the Config v1-to-v2 migration workflow for standalone and Collector
  receiver deployments
- provide deterministic input and expected-output fixtures for both deployment
  modes
- document validation, canary, rollback, and release-gate procedures and link
  them from the relevant operator entry points

Closes #1758.

## Motivation

Operators need a reviewable migration path that distinguishes standalone and
Collector receiver configuration shapes, identifies settings that cannot be
preserved safely, and provides a rollback plan before Config v2 is enabled in
production. The committed fixtures make the documented transformation concrete
and reproducible for both deployment modes.

## Validation

- Markdown lint across the repository: 0 issues
- parsed all four migration fixtures as YAML

## Release considerations

This guide documents the migration path without declaring Config v2 released.

- Schema acceptance does not imply runtime support outside the documented
  declarative-provider subset.
- The v1 removal date is not decided.
- The first enabling release notes must link this guide, state supported
  deployment modes and v1 compatibility, and be linked from #2251 and #1133.

PR #2811, opened 2026-07-28
Reject unsupported Config v2 document fields

MrAlias:
## Summary

- reject non-empty Config v2 document fields that the runtime converter cannot preserve instead of silently ignoring them
- keep empty optional document sections accepted
- decode `distribution` explicitly so its presence is available to validation
- add focused coverage for accepted and rejected top-level document shapes

## Why

The Config v2 runtime converter supports only a bounded subset of the OpenTelemetry declarative document. Accepting unsupported fields makes a configuration appear valid even though conversion drops user intent and changes runtime behavior.

This is the first internal, review-sized extraction from #2799. It does not expose a new public surface or publish migration guidance; later stacked changes will cover the remaining conversion, CLI, artifact, and documentation work.

## Validation

- `go test ./internal/config/schema ./internal/config/convert`
- `go test -race ./internal/config/schema ./internal/config/convert`
- `make -o prereqs verify`
- `git diff --check`

### DIFF
--- a/internal/config/convert/import_document.go
+++ b/internal/config/convert/import_document.go
@@ -25,11 +25,8 @@ func DocumentToRuntime(src *schema.Document) (*obi.Config, error) {
 	if src == nil {
 		return nil, errors.New("missing OBI document")
 	}
-	if fields := src.OpenTelemetryExtensionFields(); len(fields) != 0 {
-		return nil, fmt.Errorf(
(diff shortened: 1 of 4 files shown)


mariomac (approved):
Looks good!

mmat11 (approved):

PR #2812, opened 2026-07-28
JS Route harvester: support standard URLPattern API

mariomac:
## Summary

Adds support for the standard URL Pattern API in the JS Route harvester: https://developer.mozilla.org/en-US/docs/Web/API/URL_Pattern_API

## Validation

### DIFF
--- a/pkg/internal/transform/route/harvest/js.go
+++ b/pkg/internal/transform/route/harvest/js.go
@@ -217,13 +217,14 @@ func newFrameworkPatterns() *FrameworkPatterns {
 		// Supports both string literals and regex literals
 		HTTPDispatcher: regexp.MustCompile(`\.on(Get|Post|Put|Patch|Delete|Head|Options|All)\s*\(\s*(?:['"\x60]([^'"\x60]+)['"\x60]|/((?:[^\\,]|\\.)+))`),
 
-		// URLPattern WEB API. The constructor takes the pattern either as a
-		// string or as an init object, optionally through a namespace object:
+		// URLPattern WEB API: matches the opening of the constructor call, whose
+		// arguments are then read until the call closes. The pattern is given
+		// either as a string or as an init object, and either of them may be
+		// spread over several lines:
 		//   new URLPattern("https://example.com/books/:id")
 		//   new URLPattern({ pathname: "/books/:id" })
 		//   new urlpattern.URLPattern("/books/:id", base)
-		// \s covers the line breaks of a call spread over several lines.
-		URLPattern: regexp.MustCompile(`URLPattern\s*\(\s*(?:['"\x60]([^'"\x60]*)['"\x60])?`),
+		URLPattern: regexp.MustCompile(`URLPattern\s*\(`),
 
 		// Matches the 'pathname' member of a URLPattern init object in every
 		// legal property key form: pathname, 'pathname', ["pathname"]
@@ -271,9 +272,9 @@ type RouteExtractor struct {
 	// enableVersioningTypeSeen tracks whether the current enableVersioning call
 	// named an explicit VersioningType (URI is NestJS's default when it didn't)
 	enableVersioningTypeSeen bool
-	// inURLPattern tracks a new URLPattern({...}) call whose init object spans
-	// several lines, until its closing parenthesis
-	inURLPattern bool
+	// urlPatternCall buffers the arguments of the URLPattern call currently
+	// being scanned, nil when no call is open
+	urlPatternCall *urlPatternCall
 
 	// application-level NestJS settings, harvested from any scanned file
 	// (typically main.ts) and applied to Nest routes after the scan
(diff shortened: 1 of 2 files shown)


pkg/internal/transform/route/harvest/js.go
```diff
@@ -213,6 +217,19 @@ func newFrameworkPatterns() *FrameworkPatterns {
 		// Supports both string literals and regex literals
 		HTTPDispatcher: regexp.MustCompile(`\.on(Get|Post|Put|Patch|Delete|Head|Options|All)\s*\(\s*(?:['"\x60]([^'"\x60]+)['"\x60]|/((?:[^\\,]|\\.)+))`),
 
+		// URLPattern WEB API. The constructor takes the pattern either as a
+		// string or as an init object, optionally through a namespace object:
+		//   new URLPattern("https://example.com/books/:id")
+		//   new URLPattern({ pathname: "/books/:id" })
+		//   new urlpattern.URLPattern("/books/:id", base)
+		// \s covers the line breaks of a call spread over several lines.
+		URLPattern: regexp.MustCompile(`URLPattern\s*\(\s*(?:['"\x60]([^'"\x60]*)['"\x60])?`),
```
MrAlias: The scanner feeds this matcher one physical line at a time. With `new URLPattern(\n  "/books/:id",\n  base,\n)`, the first line captures no string and `inURLPattern` then consumes the remaining lines without inspecting a string argument, so no framework route is added.

Please retain the first argument until its literal can be parsed, and cover this layout.

pkg/internal/transform/route/harvest/js.go
```diff
@@ -682,6 +702,62 @@ func sortRouteFragments(fragments []string) {
...
+	for _, constructor := range constructors {
+		e.inURLPattern = true
+		addRoute(urlPatternPathname(constructor[1]))
+	}
+
+	for _, pathname := range e.patterns.URLPatternPathname.FindAllStringSubmatch(line, -1) {
+		addRoute(ensureLeadingSlash(pathname[1]))
+	}
+
+	if strings.Contains(line, ")") {
```
MrAlias: A `)` in any preceding property ends the call. A valid pattern with `protocol: "(https?)"` before `pathname` clears this state before the pathname is seen; it is only harvested by the fallback and is not counted by `FrameworkRoutes`. If compiled output has fragments, `ExtractNodejsRoutes` returns those partial fragments instead and drops this source route.

Track the URLPattern call depth so the state ends only at the outer constructor, and add coverage for this case.

MrAlias (commented):
The supported URLPattern forms can lose route declarations in normal multiline and regex-containing configurations. The harvester can return no source routes or choose unrelated partial compiled fragments.

pkg/internal/transform/route/harvest/js.go
mariomac: Done! The first argument is retained until it can be parsed, so `URLPattern(\n "/books/:id",\n base,\n)` is harvested. Parsing the complete argument list also means only the first argument is considered, so a baseURL or an options object as the second argument is no longer mistaken for the pattern.

pkg/internal/transform/route/harvest/js.go
mariomac: Done! Depth is tracked, and parentheses inside string literals are skipped, so `protocol: "(https?)"` before pathname no longer ends the call. Skipping literals also covers the inverse failure: an unbalanced ( in a string used to leave the state open and swallow the rest of the file.

mariomac:
Thanks for the review @MrAlias ! Addressed the comments and added new test cases.

pkg/internal/transform/route/harvest/js.go
```diff
@@ -682,6 +703,159 @@ func sortRouteFragments(fragments []string) {
...
+		pathname := e.patterns.URLPatternPathname.FindStringSubmatch(args)
+		if pathname == nil || pathname[1] == "" {
+			return ""
+		}
+		return ensureLeadingSlash(pathname[1])
+	}
+
+	if isJSQuote(args[0]) {
+		if pattern := e.patterns.QuotedString.FindStringSubmatch(args); pattern != nil {
+			return urlPatternPathname(pattern[1])
```
MrAlias: The string constructor cannot be reduced to its first literal here. With `new URLPattern("https://example.com/books/:id\\?view=:view#details")`, URLPattern's pathname is `/books/:id`, but the harvester produces `/books/:idviewviewdetails`; with `new URLPattern("books/:id", "https://example.com/api/")`, the pathname resolves to `/api/books/:id`, but this returns `/books/:id`. Both are valid constructor forms and become incorrect complete routes.

Could we parse the string form's pathname, search, and hash boundaries and use a literal base URL when resolving a relative pathname? Please cover both cases.

pkg/internal/transform/route/harvest/js.go
```diff
@@ -213,6 +217,20 @@ func newFrameworkPatterns() *FrameworkPatterns {
...
 		HTTPDispatcher: regexp.MustCompile(`\.on(Get|Post|Put|Patch|Delete|Head|Options|All)\s*\(\s*(?:['"\x60]([^'"\x60]+)['"\x60]|/((?:[^\\,]|\\.)+))`),
 
+		// URLPattern WEB API: matches the opening of the constructor call, whose
+		// arguments are then read until the call closes. The pattern is given
+		// either as a string or as an init object, and either of them may be
+		// spread over several lines:
+		//   new URLPattern("https://example.com/books/:id")
+		//   new URLPattern({ pathname: "/books/:id" })
+		//   new urlpattern.URLPattern("/books/:id", base)
+		URLPattern: regexp.MustCompile(`URLPattern\s*\(`),
```
MrAlias: This also matches `createURLPattern("/not-a-route")`, so a helper call is treated as the standard constructor and adds a framework route. Once `FrameworkRoutes()` is non-zero, `ExtractNodejsRoutes` returns that incorrect complete route before scanning compiled output.

Could we require an actual `new ...URLPattern(` constructor shape while retaining the supported namespace form, and cover the helper-name case?

MrAlias (commented):
The multiline fixes address the earlier failures, but the remaining parsing shortcuts can still promote incorrect paths to complete routes.

pkg/internal/transform/route/harvest/js.go
mariomac: The string pattern is now truncated at the first `?`/`#` that actually opens a component. `?` is ambiguous in URLPattern syntax: after a name, group or wildcard it is the optional modifier and belongs to the pathname, which is why an escaped`\?` is the way to open a search component after one of them. The new scan tracks whether a modifier is legal at each offset, skipping over `:name` tokens and `(...)` groups:
- `new URLPattern("https://example.com/books/:id\?view=:view#details")` → `/books/:id`
- `new URLPattern("https://(sub.)?example.com/books/:id?")` → `/books/:id?` (unchanged)

Regarding to the Base URL resolution, the argument list is now split at the top level and a literal second argument is used as the base URL. A relative pathname replaces the base's last path segment, as URL resolution does:
- `new URLPattern("books/:id", "https://example.com/api/")` → `/api/books/:id`
- `new URLPattern("books/:id", "https://example.com/api")` → `/books/:id`

I applied the same resolution to the init object form, since `{ pathname: "books/:id" }` with a base URL resolves the same way.

A non-literal base (`new URLPattern("books/:id", base)`) is still not statically resolvable and keeps the previous behaviour of harvesting the pathname as given.

pkg/internal/transform/route/harvest/js.go
mariomac: Done. The pattern now requires the `new` operator and keeps the namespace form:

```
\bnew\s+(?:[\w$]+\s*\.\s*)*URLPattern\s*\(
```

`createURLPattern("/not-a-route")` and `new MyURLPattern(...)` no longer match, while `new URLPattern(`, `new   URLPattern (` and `new urlpattern.URLPattern(` still do. Both negative cases are covered.

One limitation worth flagging: the scanner reads one physical line at a time, so a call with `new` and `URLPattern(` on separate lines is not matched. No formatter produces that shape, and buffering a bare `new` across lines would cost more in false positives than it buys

marctc (approved):
Do we need/have integration tests for these features?

mariomac:
@marctc ATM just added unit tests but in a later PR (Deno integration) the integration tests are going to be added

pkg/internal/transform/route/harvest/js.go
```diff
@@ -682,6 +706,308 @@ func sortRouteFragments(fragments []string) {
...
+	basePath := "/"
+	if _, path, hasPath := strings.Cut(authority, "/"); hasPath {
+		basePath = "/" + path
+	}
+	if end := strings.IndexAny(basePath, "?#"); end >= 0 {
+		basePath = basePath[:end]
+	}
+
+	// a relative pathname replaces the last segment of the base pathname
+	return basePath[:strings.LastIndex(basePath, "/")+1] + pathname
```
MrAlias: When the string form has a relative pathname and a non-literal base, this branch invents a root-relative complete route. For `const base = "https://example.com/api/"; new URLPattern("books/:id", base)`, URLPattern's pathname is `/api/books/:id`, but this returns `/books/:id`. The manual concatenation also leaves dot segments unresolved: `../books/:id` against `/api/v1/` becomes `/api/v1/../books/:id` instead of `/api/books/:id`.

Could we only emit a complete route when the base is statically resolvable with URL semantics? Otherwise this declaration should not preempt the compiled-route scan.

pkg/internal/transform/route/harvest/js.go
```diff
@@ -682,6 +706,308 @@ func sortRouteFragments(fragments []string) {
...
+	if len(parsed) > 1 {
+		base = jsStringLiteral(parsed[1])
+	}
+
+	if pattern[0] == '{' {
+		pathname := e.patterns.URLPatternPathname.FindStringSubmatch(pattern)
+		if pathname == nil || pathname[1] == "" {
+			return ""
+		}
+		return resolveURLPatternPathname(pathname[1], base)
```
MrAlias: The object overload does not take its base URL as the second argument: `baseURL` is a member of the init object, while the second argument is options. With `new URLPattern({ pathname: "books/:id", baseURL: "https://example.com/api/" })`, URLPattern's pathname is `/api/books/:id`, but `base` is empty here and the harvester emits `/books/:id`. The new object-plus-second-string test covers a constructor that throws instead of the standard form.

Could we read a literal `baseURL` member from the init object and cover that overload?

pkg/internal/transform/route/harvest/js.go
```diff
@@ -682,6 +706,308 @@ func sortRouteFragments(fragments []string) {
...
+	return append(arguments, strings.TrimSpace(args[start:]))
+}
+
+// jsStringLiteral returns the contents of the string literal that s starts
+// with, or an empty string when s does not start with one.
+func jsStringLiteral(s string) string {
+	if s == "" || !isJSQuote(s[0]) {
+		return ""
+	}
+	return s[1:endOfJSString(s, 0)]
```
MrAlias: `jsStringLiteral` accepts the literal prefix of a larger expression. With `new URLPattern("/books/" + segment, "https://example.com")`, this returns `/books/`, which cleanup promotes to the complete route `/books` even though the runtime pathname may be `/books/:id`.

Could we require the closing quote to end the argument, and reject interpolated templates, before treating it as statically resolvable?

MrAlias (commented):
The latest parsing changes still classify some unresolved or incorrectly resolved URLPattern declarations as complete routes, which can make route discovery return paths the application never serves.

pkg/internal/transform/route/harvest/js.go
mariomac: Done. A relative pathname is now only resolved when the base is statically resolvable; a declared but non-literal base returns no path, so the declaration no longer preempts the compiled-route scan:

- `const base = "..."; new URLPattern("books/:id", base)` → no framework route
- `new URLPattern("/books/:id", base)` → `/books/:id` (an absolute pathname ignores the base)

Resolution goes through `net/url` now, so dot segments are removed with URL semantics: `new URLPattern("../books/:id", "https://example.com/api/v1/")` → `/api/books/:id`. The pathname is escaped and unescaped back by `ResolveReference`, so `:id`, `(\d+)` and friends survive it unchanged. Both cases are covered.

pkg/internal/transform/route/harvest/js.go
mariomac: Good catch — fixed. The init object form now reads a literal `baseURL` member and ignores the second argument, which is the options bag for that overload:

- `new URLPattern({ pathname: "books/:id", baseURL: "https://example.com/api/" })` → `/api/books/:id`
- `new URLPattern({ pathname: "books/:id", baseURL: base })` → no route (same rule as the string form)
- `new URLPattern({ pathname: "books/:id" }, { ignoreCase: true })` → `/books/:id`

The throwing object-plus-second-string test is gone, replaced by these three. The member values are now located by a key-only regex and their source read out by a small object-member scanner, so `pathname` and `baseURL` share the same literal parsing.

pkg/internal/transform/route/harvest/js.go
mariomac: Done. `jsStringLiteral` now requires the closing quote to end the argument, and rejects template literals holding a substitution. A literal that only prefixes a larger expression is no longer statically resolvable, so nothing is harvested:

- `new URLPattern("/books/" + segment, "https://example.com")` → no route
- `new URLPattern({ pathname: "/books/" + segment })` → no route
- `` new URLPattern(`/books/${segment}`) `` → no route

The init object members go through the same function now, so the object form gets the check too. All three are covered.

pkg/internal/transform/route/harvest/js.go
```diff
@@ -682,6 +709,406 @@ func sortRouteFragments(fragments []string) {
...
+	return regexp.MustCompile(
+		`(?:\b` + name + `|['"]` + name + `['"]|\[\s*['"\x60]` + name + `['"\x60]\s*\])\s*:\s*`)
+}
+
+// jsObjectMemberValue returns the source of the value of the object literal
+// member whose key matches key, or an empty string when there is no such
+// member. The value ends at the comma separating it from the next member or at
+// the closing brace of the object.
+func jsObjectMemberValue(object string, key *regexp.Regexp) string {
+	loc := key.FindStringIndex(object)
```
MrAlias, line 1081: Non-blocking: this returns the first textual match at any nesting depth, so `new URLPattern({ metadata: { pathname: "/decoy" }, pathname: "/actual" })` is harvested as `/decoy`, and a later spread can similarly override the selected value.

I don't think this needs to hold up the feature, but could we make the object scan conservative in a follow-up and treat ambiguous initializers as unresolved?

MrAlias (approved):
The supported URLPattern forms are covered well enough to land.

pkg/internal/transform/route/harvest/js.go
mariomac: Ok! I'm wondering if we should replace the whole regexp-based JS harvester by a tokenizer/sub-AST implementation that helps with all these small edge cases also in the rest of route sources

PR #2813, opened 2026-07-28
Limit NestJS route harvesting expansion

MrAlias:
### Motivation

- Prevent untrusted Node.js source from forcing unbounded work in the route harvester by bounding decorator-array extraction and the controller-prefix/version Cartesian expansion.
- The previous logic could extract unlimited quoted strings and emit an unbounded number of `RoutePattern` entries, enabling a resource exhaustion.

### Description

- Add two constants: `maxNestDecoratorValues = 64` and `maxNestRouteVariants = 256` to bound NestJS decorator processing.
- Limit `quotedStrings` to at most `maxNestDecoratorValues` matches instead of using `-1` for unlimited extraction.
- Cap the number of emitted route variants in `flushNestMethod` to `maxNestRouteVariants` and stop emitting further pairs once the cap is reached.
- Add two unit tests (`TestNestJSQuotedStringsAreBounded`, `TestFlushNestMethodCapsRouteVariants`) that exercise the new bounds.

### Testing
- `go test ./pkg/internal/transform/route/harvest`

### DIFF
--- a/pkg/internal/transform/route/harvest/js.go
+++ b/pkg/internal/transform/route/harvest/js.go
@@ -252,8 +252,9 @@ type RouteExtractor struct {
 	// (@Version() may appear above or below @Get()), so the route is emitted
 	// only when the stack ends: at the first non-decorator line, the next
 	// method or controller decorator, or the end of the file.
-	pendingNestMethod *RoutePattern
-	nestRouteVariants int
+	pendingNestMethod    *RoutePattern
+	nestRouteVariants    int
+	nestRouteLimitLogged bool
 	// inEnableVersioning tracks a multi-line app.enableVersioning({...}) call
 	// until its closing parenthesis
 	inEnableVersioning bool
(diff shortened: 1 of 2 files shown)


pkg/internal/transform/route/harvest/js.go
```diff
@@ -413,18 +419,30 @@ func (e *RouteExtractor) flushNestMethod() {
...
+			if !e.appendNestRoute(RoutePattern{
 				Method:  m.Method,
 				Path:    joinNestPaths(prefix, m.Path),
 				File:    m.File,
 				Line:    m.Line,
 				Nest:    true,
 				Version: version,
-			})
+			}) {
+				return
```
NimrodAvni78, line 431: should we warn / debug when we hit the limit?

MrAlias: Added a debug log at the shared NestJS route append gate when the per-file variant cap drops additional routes. It includes the source file and limit and emits once per extractor to avoid attacker-controlled log amplification. `TestFlushNestMethodCapsRouteVariants` verifies repeated over-limit attempts emit one record.

mariomac (approved):
🚀

mmat11 (approved):

Issue #1282, opened 2026-02-11
Filtering (for traces) should allow to skip unaffected spans


pilhuhn:
A filter expression like 

```
filter:
  application:
    server.port:
      match: 8787
```

will only generate a span if the server port is 8787. This has the effect, that internal spans for e.g. for DB-calls triggered from a process are not reported, as the server port for the (targeted) DB is different. 
Of course I could give multiple ports to match, but that requires me to know the internals of observed software.

The filter-matching should probably be limited to e.g. only http-servers and ignore SQLClient spans in the matching.

MrAlias:
This still looks relevant, but it appears to be partly addressed by the Config V2 direction rather than something we should fix only in the current global `filter.application` shape.

Config V2 intended model is protocol-local and signal-scoped filtering, for example:

```yaml
extensions:
  obi:
    capture:
      instrumentation:
        http:
          filters:
            traces:
              server.port:
                equals: 8787
```

With that shape, an HTTP traces filter should apply to HTTP trace spans and should not cause SQL/client child spans to be dropped merely because they do not have the same `server.port`.

There are still a few things to settle before closing this:

- Confirm the intended runtime semantics for protocol/signal-scoped filters in Config V2.
- Preserve v1 migration behavior by fanning out `filter.application` to all protocol/signal filters, while documenting that users can narrow the migrated V2 config afterward.
- Wire the V2 protocol-specific filters into the runtime path once the V2 config path is ready.
- Add a regression test where an HTTP server span matches `server.port`, and a related SQL/DB client span is still emitted.
- Document how this interacts with span metrics and service graph metrics.

So I think this should stay open, but tracked as part of the Config V2 filtering semantics work rather than as an isolated v1 filter change.

PR #2815, opened 2026-07-28
Reject divergent Config v2 signal filters

MrAlias:
## Summary

- reject Config v2 application filters when protocol or signal definitions differ from the canonical HTTP trace filter
- reject metric and trace filter differences for network capture and network stats
- add focused coverage for divergent protocol, single-signal, and network filter configurations

## Why

Config v2 can express filters per protocol and signal, while the runtime stores one shared application filter, one network filter, and one network-stats filter. Conversion previously compared only non-empty HTTP trace and metric filters. A filter defined for one signal, another protocol, or a network path could therefore be accepted and silently collapse into different runtime behavior.

Failing closed preserves user intent until the runtime can represent these distinctions. This is a focused internal extraction from #2799 and is independently mergeable from `main`.

## Validation

- `go test ./internal/config/convert`
- `go test -race ./internal/config/convert`
- `make -o prereqs verify`
- `git diff --check`

### DIFF
--- a/internal/config/convert/import.go
+++ b/internal/config/convert/import.go
@@ -48,7 +48,7 @@ func V2ToRuntime(src *schema.Extension) (*obi.Config, error) {
 	if err := validateV2RuleSelectorFamilies(src.Capture.Rules); err != nil {
 		return nil, err
 	}
-	if err := validateV2HTTPFilters(src.Capture.Instrumentation.HTTP.Filters); err != nil {
+	if err := validateV2SignalFilters(src); err != nil {
 		return nil, err
 	}
 	if err := validateV2HTTPRoutes(src.Capture.Instrumentation.HTTP.Routes, src.Capture.Rules); err != nil {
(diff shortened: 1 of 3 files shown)


mariomac (approved):
Ideally it would be great being able to filter metrics and traces separately, but I understand that this error comes from an internal restriction in Beyla, right? (there is a common filter node).

If so, it LGTM, but I think in the future we can provide filtering for both signals if in the internal pipeline, instead of having a single attribute filter for all the internal spans, we add separate attribute filters, one before the traces exporter and another before the metrics exporter, and we feed them with the different configurations.

MrAlias:
> Ideally it would be great being able to filter metrics and traces separately, but I understand that this error comes from an internal restriction in Beyla, right? (there is a common filter node).
> 
> If so, it LGTM, but I think in the future we can provide filtering for both signals if in the internal pipeline, instead of having a single attribute filter for all the internal spans, we add separate attribute filters, one before the traces exporter and another before the metrics exporter, and we feed them with the different configurations.

Yes, exactly. The runtime currently has a single `filter.application`, applied before spans fan out to the trace and metrics branches, so this PR continues to reject divergent Config v2 filters rather than silently collapse them.

The follow-up is already tracked in #1282. That work should preserve protocol- and signal-scoped filters in the runtime, apply trace filters on the trace branch, and apply metrics filters before span-derived metric aggregation and export. So the proposed split is the right direction, with additional protocol-specific selection needed to fully support the Config v2 model.

PR #2816, opened 2026-07-28
Stabilize YAML export mode ordering

MrAlias:
## Summary

- serialize export modes in the fixed order `metrics`, `traces`, `logs`
- strengthen YAML marshal coverage to assert the emitted sequence order

## Why

`ExportModes.MarshalYAML` iterated over a Go map, so otherwise identical configuration and migration output could list signals in a different order between runs. That nondeterminism creates noisy diffs and prevents generated YAML from being reproducible.

Using an explicit sequence makes output stable without changing which signals are enabled. This focused change is extracted from #2799 and is independently mergeable from `main`.

## Validation

- `go test -race ./pkg/appolly/services`
- `git diff --check`

### DIFF
--- a/pkg/appolly/services/export.go
+++ b/pkg/appolly/services/export.go
@@ -27,6 +27,7 @@ var modeForText = map[string]maps.Bits{
 	"logs":    blockLogs,
 }
 
+// YAML serialization must use a fixed order to keep generated configuration reproducible.
 var orderedExportModes = []string{"metrics", "traces", "logs"}
 
 const (
--- a/pkg/appolly/services/export.go
+++ b/pkg/appolly/services/export.go
@@ -27,6 +27,8 @@ var modeForText = map[string]maps.Bits{
 	"logs":    blockLogs,
 }
 
+var orderedExportModes = []string{"metrics", "traces", "logs"}
+
 const (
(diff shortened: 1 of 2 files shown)


pkg/appolly/services/export.go
```diff
@@ -164,7 +166,8 @@ func (modes ExportModes) MarshalYAML() (any, error) {
 	if modes.blockSignal == blockAll {
 		return node, nil
 	}
-	for text, mode := range modeForText {
+	for _, text := range orderedExportModes {
```
mariomac, line 170: I think is worth adding a comment here (or before the `orderedExportModes` variable) about why are we doing it this way instead of just iterating the `modeForText` map. Otherwise I'm afraid in the future someone else could remove it.

Also, I see this modeForText iteration in many places. Does order care in JSON or Text exports?

mariomac (approved):

pkg/appolly/services/export.go
MrAlias: Added a comment above `orderedExportModes` explaining that YAML serialization needs a fixed sequence to keep generated configuration reproducible (dcbe4737).

The only other `modeForText` iteration is in `JSONSchema`; enum order does not affect schema validation. Text handling only unmarshals values through map lookups, so it does not emit an order.

PR #2820, opened 2026-07-28
Add optional Go probe orchestration

MrAlias:
## Summary

- add optional Go probe groups that preflight every symbol and attach executable-scoped probes atomically with reverse-order rollback
- defer process-scoped probe registration until the executable has committed successfully
- identify executables consistently by `{device, inode}` through discovery, accounting, instrumentation, and teardown
- carry an executable generation through queued deletion events so stale teardown cannot detach a replacement executable
- add shared internal ring-buffer event dispatch for tracer-owned control events

## Motivation

Go Auto SDK activation needs three lifecycle probes at executable scope and a temporary `newSpan` probe scoped to each target PID. The existing probe orchestration does not provide an atomic optional group, a post-commit process-scoped registration hook, or a safe way to consume tracer-internal control events.

This change adds those reusable lifecycle primitives without implementing Auto SDK activation. Tracers that do not implement the optional interfaces keep their existing behavior.

Executable replacement and teardown are serialized. A replacement closes the displaced instrumenter before registering its process-scoped probes, while a queued deletion event carries the old generation and is ignored if a newer instrumenter has committed for the same executable.

## Compatibility

The process-scoped hook uses ordinary PID-filtered uprobes and does not use attach cookies or introduce a newer kernel ABI dependency. The orchestration remains compatible with OBI's minimum supported kernel path.

## Validation

- `make generate`
- focused and race tests for `pkg/ebpf`, discovery, the shared event dispatcher, and the instrumentation event loop
- repeated replacement/stale-unlink lifecycle race tests
- `make lint`
- `make compile`
- Linux arm64 and Darwin amd64/arm64 compile checks

Prerequisite for #2810. Part of #2258.


mariomac (approved):
Looks good! Fine to get it merged once the broken integration tests are fixed.

PR #2824, opened 2026-07-29
refactor(runtime): share the kernel-time helper

pinoOgni:
## Summary

The monotonic-to-wall-clock conversion for bpf event timestamps was duplicated between the JVM runtime-metric parser and generictracer. 

This PR moves it to the `timing` package, next to the `MonoTimeNow` clock it inverts, and have both callers use it. It also adds a direct test for the conversion and the JVM parser tests now assert the stamped time against it within a measured tolerance instead of swapping the clocks.

Moreover, we now drop `timing_darwin.go` because `clock_gettime(CLOCK_MONOTONIC)` works on darwin too.

## Validation


marctc (approved):
cool!

mariomac (approved):

PR #2825, opened 2026-07-29
fix(gotracer): read HTTP/2 StreamID from a register on arm64

mariomac:
`golang_stream_id` read the `StreamID` argument of
`golang.org/x/net/http2.(*Framer).WriteHeaders` from the stack at `SP + 8` on
every architecture. That offset is amd64-only, so on arm64 it produced garbage
and gRPC client context propagation silently stopped working.

## Cause

Where Go places `HeadersFrameParam` depends on the internal ABI's register
budget. Counting the receiver, the call needs 11 integer argument registers:

- **amd64** has 9, so from x/net/http2 v0.45.0 on — when the struct outgrew the
  budget — Go assigns it entirely to the stack. Hence the existing
  `_http2_zero_forty_five_zero` version flag and the `SP + 8` read (the `CALL`
  instruction pushed the return address).
- **arm64** has 16 (R0–R15), so the struct still fits in registers and
  `StreamID` stays in R1.

Consequence on arm64: `grpcFramerWriteHeaders` computed nonsense stream IDs
(`71310922228096`, `4099872`, …) while `NewStream_ret` recorded the correct ones
(1, 3, 7, 9). The `ongoing_streams` lookup never matched, so the uprobe never
wrote the HPACK traceparent. Only the first request on a fresh connection
propagated — via a different path — and every later stream on that connection
started a new trace.

## Fix

Gate the stack read on `__TARGET_ARCH_x86` and fall through to `GO_PARAM2`
otherwise. Also read `StreamID` as the `u32` it actually is, so the value no
longer depends on the struct's alignment padding.

## Test changes

`TestSuite_GRPCRelay` also had a startup race of its own: the only gate was the
HTTP health checks, which prove a service is up but not that OBI discovered its
pid. The `dotnet-relay` warmup loop then drove `/relay` from t=0, creating each
relay's singleton downstream channel before OBI was ready. Since OBI classifies
a connection as HTTP/2 only on the client preface, a channel opened that early
stays un-injectable for its whole lifetime.

`waitForRelayInstrumentation` now waits for all eight services to have spans in
Jaeger before the first `/relay`, driven by `/health` traffic only — that
produces spans without creating any chain connections. The two warmup loops it
replaces are removed.

## Verification

On arm64, five consecutive requests over one persistent channel each produce the
complete eight-service chain; before, only the first reached python-relay:

    docker compose -f docker-compose-grpc-relay.yml restart go-entry
    sleep 28
    for i in 1 2 3; do
      curl -H "Traceparent: 00-a${i}ff…ff-bbbbbbbbbbbbbbbb-01" \
        http://localhost:8080/relay
    done

Also confirmed on the wire with `tcpdump` on port 50051 (OBI doesn't Huffman-encode
the traceparent value, so it's plain ASCII): pre-fix, only request 1's trace ID
appears; post-fix, all of them.

### DIFF
--- a/bpf/gotracer/go_common.h
+++ b/bpf/gotracer/go_common.h
@@ -512,25 +512,39 @@ static __always_inline void process_meta_frame_headers(void *frame, tp_info_t *t
     }
 }
 
+// Reads the StreamID of the HeadersFrameParam passed to
+// golang.org/x/net/http2.(*Framer).WriteHeaders.
+//
+// Whether that struct is register- or stack-assigned depends on the Go internal
+// ABI's register budget. Counting the receiver, the call needs 11 integer
+// argument registers. amd64 offers 9, so from x/net/http2 v0.45.0 on — when the
+// struct grew past the budget — Go assigns it entirely to the stack, with
+// StreamID as its first field one word above SP (the CALL instruction pushed the
+// return address). arm64 offers 16 (R0-R15), so the struct still fits in
+// registers there and StreamID stays in R1.
+// https://github.com/golang/go/blob/master/src/cmd/compile/abi-internal.md
 static __always_inline u64 golang_stream_id(struct pt_regs *ctx, off_table_t *ot) {
+#ifdef __TARGET_ARCH_x86
     const u64 on_stack = go_offset_of(ot, (go_offset){.v = _http2_zero_forty_five_zero});
 
-    if (!on_stack) {
-        return (u64)GO_PARAM2(ctx);
+    if (on_stack) {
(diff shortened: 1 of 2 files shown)


mmat11:
this was changed in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2772/changes#diff-5cbaf9b7e28b1ef94a048b1cfb4a600dbd107a87de1cb023e6cbe57ab0ecae7dR514 as well (and I believe the test flakyness is absent there too)

mmat11 (approved):
lets merge this in order to remove flakyness in main, I'll rebase my PR on top of these changes 👍🏻

PR #2826, opened 2026-07-29
Add support for generic Python asyncio server

### LINKED ISSUE #931: Support python contextvars based context propagation
Based on https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/903, there is no specific python trace context correlation heuristic. The thread based correlation should work well for traditional WSGI frameworks like Flask and Django but not for asyncio.

Rather than focus on asyncio, I'm wondering if correlation can be done by the builtin [contextvars api](https://docs.python.org/3/library/contextvars.html) which provides generic context local storage and is well integrated with asyncio and can be plugged into other concurrency frameworks with [`context.run()`](https://docs.python.org/3/library/contextvars.html#contextvars.Context.run). There is a C API for entering/exiting a context https://docs.python.org/3/c-api/contextvars.html#c.PyContext_Enter.

Let me know if you need more details or it's infeasible. We also use contextvars for the OTel Python context which might make integration with a live SDK feasible.

marctc:
## Summary

This PR adds support for context propagation for generic python asyncio servers.

`asyncio.start_server` creates handler tasks from a plain loop callback. At that point there is no current task, so OBI now uses the owner of the active Python context as the logical parent.

It also removes context-to-task mappings when a `PyContext` is freed. CPython can reuse context addresses, and keeping an old mapping could link a new request to an unrelated task.

Closes #931 

## Validation


bpf/common/trace_parent.h
```diff
@@ -161,10 +161,9 @@ static __always_inline u64 resolve_python_current_task(const trace_key_t *t_key,
     }
 
     // asyncio.to_thread can switch work onto a thread that inherited a context
-    // but has not run task_step, so current_context is the only usable link.
-    const python_context_task_t *context_task = (const python_context_task_t *)bpf_map_lookup_elem(
-        &python_context_task, &thread_state->current_context);
-    const u64 task_id = resolve_python_context_task(context_task);
+    // but has not run task_step, so the context's owner task (cached when the
+    // context was entered) is the only usable link.
+    const u64 task_id = thread_state->current_context_task;
```
mariomac: Caching only the raw `TaskObj*` here removes the version check that `resolve_python_context_task()` previously performed on every fallback lookup.

An `asyncio.to_thread()` worker can continue running after its originating task is cancelled or completed. If CPython reuses that task address while the context is still active, `python_task_state` describes a different request, but `current_context_task` still points to it. Later outbound spans can consequently be attached to an unrelated trace.

Please cache the mapping version alongside the pointer and revalidate it before both consumers use it, or continue resolving through `current_context`.

A regression test where a cancelled `to_thread()` worker keeps running while new tasks force address reuse would cover this case.

bpf/generictracer/python.c
```diff
@@ -149,10 +162,16 @@ int obi_uprobe_context_run(struct pt_regs *ctx) {
...
 
-    // asyncio.to_thread worker has no current_task; look up which task copied this context
+    // asyncio.to_thread worker has no current_task; look up which task copied this context.
+    // The ctx_vars check rejects stale bindings left by a freed context whose
+    // address was recycled, for builds where context_tp_dealloc is not attached.
     const python_context_task_t *context_task =
         (const python_context_task_t *)bpf_map_lookup_elem(&python_context_task, &context);
-    const u64 task_id = resolve_python_context_task(context_task);
+    u64 task_id = resolve_python_context_task(context_task);
+    if (task_id && context_task && context_task->vars != read_python_context_vars(context)) {
```
mariomac: `ctx_vars` is the current HAMT root, not a lifetime-stable identity. CPython replaces this pointer whenever a `ContextVar` is set or removed.

Therefore, a live `Context` that is entered, mutated, and entered again will fail this comparison and have its valid owner discarded. The check is unconditional, so this also happens when the deallocation probe is successfully attached.

This can break the callback-driven case being added here: the second callback running under a reused context has no `current_task`, and tasks created from it lose their logical parent.

Please use an identity mechanism that survives normal context mutation, or keep the stored HAMT pointer synchronized. A test with two consecutive `ctx.run()` calls—where the first changes a `ContextVar` and the second creates a task—would expose the issue.

devdocs/features.md
```diff
@@ -138,4 +117,4 @@ OBI has support for several asynchronous frameworks that allow it to propagate c
 | Ruby Puma Server    |   Ruby    |              N/A | Only works with Puma server                       | Stable
 | Java Thread pool    |   Java    |           JDK 8+ | N/A                                               | Stable
 | Java Virtual Threads |  Java    |          JDK 21+ | Log enrichment is skipped on virtual threads      | Stable
-| Python asyncio      |  Python   |    Python >= 3.9 | Only works with uvloop event loop                 | Stable
+| Python asyncio      |  Python   |    Python >= 3.9 | N/A                                               | Stable
```
mariomac: This row now advertises no limitations, while the detailed documentation in this PR explicitly says that `asyncio.start_server()` handlers cannot be correlated under uvloop because its accept path has no CPython-visible context.

Users consulting the support matrix will interpret `N/A` as support across event-loop implementations.

Please retain that limitation here, for example:

> Callback-driven servers such as `asyncio.start_server()` require the default asyncio event loop.

mariomac (commented):
Great Job! I asked for an automated review and got some comments that is maybe worth considering.

bpf/generictracer/python.c
```diff
@@ -195,6 +215,20 @@ int obi_uretprobe_context_run(struct pt_regs *ctx) {
...
+// so any binding for it is dead.
+SEC("uprobe/libpython3.:context_tp_dealloc")
+int obi_uprobe_context_dealloc(struct pt_regs *ctx) {
+    const u64 id = bpf_get_current_pid_tgid();
+    if (!valid_pid(id)) {
+        return 0;
+    }
+
+    const u64 context = (u64)PT_REGS_PARM1(ctx);
+    bpf_map_delete_elem(&python_context_task, &context);
```
MrAlias: `python_context_task` is shared across all instrumented processes but keyed only by `PyContext*`. Forked workers can allocate live contexts at the same virtual address, so this delete—and the ownerless-copy delete below—can remove another worker’s binding; `python_task_state` has the same raw-pointer collision for the referenced task.

Could we scope both maps by host PID plus address before using deletion as the lifetime boundary?

MrAlias (commented):
The new context-lifetime cleanup still has a multi-process correctness gap.

bpf/common/trace_parent.h
marctc: Hey the concern is valid, but would require more heavy lifting and make the PR bigger/complext. We can open an issue if needed for a follow-up work, wdyt?

bpf/generictracer/python.c
marctc: Same as previous comment. I documented the limitations in features.md

bpf/generictracer/python.c
```diff
@@ -220,16 +258,19 @@ int obi_uprobe_copy_context(struct pt_regs *ctx) {
...
     }
 
     // On the event-loop thread, copy_context still runs inside the task that is
     // serving the request, so bind the new context directly to that task.
     if (thread_state->current_task != k_python_state_none) {
-        map_context_to_task(context, thread_state->current_task);
+        map_context_to_task(id, context, thread_state->current_task);
         return 0;
     }
+    // No owner for this copy; drop any stale binding at this recycled address
```
MrAlias, line 257: At this point a plain callback can still have an owner in `current_context_task`. `loop.call_soon()` captures a fresh context with `copy_context()`, so deleting that binding here means a callback scheduled from another callback loses the owner after one hop; a task created by the second callback then falls back to the thread-local connection.

Could we map the copy to `current_context_task` before treating it as ownerless, and cover `callback -> call_soon -> create_task` with concurrent requests?

MrAlias (commented):
The direct `start_server` path is covered; generic callback propagation still has one handoff gap.

bpf/generictracer/python.c
marctc: This looks like a valid edge case, but multi-hop `call_soon()` callback chains aren’t part of the direct `asyncio.start_server()` path covered by this PR. Do you consider it a blocker, or can we track broader callback propagation separately?

bpf/common/trace_parent.h
MrAlias: I think we could add an issue to track this work, but that would then become a release blocker given I think this is a critical concern. Removing the per-use task-version validation from an existing `asyncio.to_thread()` path causing address reuse that can attach spans to an unrelated trace is a user data corruption issue that we do not want to have happen. I'm not sure it is fair to classify it as a loss of correlation.

I'm also okay with increasing the scope of this PR to include a fix. I'm not sure it would be too much to fix. I think The simplest approach is to stop trusting the cached raw `current_context_task`:

1. Add a helper that looks up `python_context_task[(pid, current_context)]` and calls the existing version-validating `resolve_python_context_task()`.
2. Use it in both consumers:
   - `resolve_python_current_task()`
   - `obi_uprobe_task_init()`
3. Remove `current_context_task` from `python_thread_state`, plus its assignments and cleanup.

That adds one map lookup to the fallback path, but it merely restores the validation and cost that existed before #2826. Verifier risk should therefore be low (right?).

grcevski (approved):
LGTM! I think if supporting call_soon() is not easy, it could be done as a follow-up. As far as I understand, it will simply not cover that case, but we won't get spans bunched up.

Btw, related to python, I noticed a pretty basic case we don't cover well. With uvicorn, if you don't make a handler async and you do async opts, it automatically converts the operation into a background thread handler, and our context propagation breaks. I have an example, I just need to clean it up. I'll open an issue.

marctc:
@MrAlias should we merge this?

MrAlias (changes requested):
No, I don't think the current head is ready to merge. The current head predates [my comment](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2826#discussion_r3731269706), and it still caches the raw `current_context_task` after a one-time validation, so task-address reuse can attach later spans to another request. That needs to be fixed and covered before merge.

To separate the concerns, I'm okay treating the multi-hop `call_soon()` propagation gap as follow-up work if the supported scope is documented accurately; the stale-task data-corruption path is not follow-up work.

marctc:
> No, I don't think the current head is ready to merge. The current head predates [my comment](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2826#discussion_r3731269706), and it still caches the raw `current_context_task` after a one-time validation, so task-address reuse can attach later spans to another request. That needs to be fixed and covered before merge.
> 
> To separate the concerns, I'm okay treating the multi-hop `call_soon()` propagation gap as follow-up work if the supported scope is documented accurately; the stale-task data-corruption path is not follow-up work.

can you check now? :wink:

bpf/common/python_task.h
```diff
@@ -11,31 +11,91 @@
...
-        (const python_task_state_t *)bpf_map_lookup_elem(&python_task_state, &context_task->task);
+        (const python_task_state_t *)bpf_map_lookup_elem(&python_task_state, &task_key);
     if (!task_state || task_state->version != context_task->version) {
+        if (stale) {
+            *stale = 1;
+        }
         return 0;
     }
 
     return context_task->task;
```
MrAlias: My earlier suggested fix was incomplete. `resolve_python_context_task()` validates a version but returns only the task address, so the network path can consume a replacement task's connection if that address is reused before `find_python_owning_server_trace()` looks it up. `python_task_state_t.parent` has the same identity loss: a detached child can later follow the raw parent address into another request. The current counter also resets to `1` when the LRU state entry is absent, allowing an old binding to validate again after eviction and reuse.

Could we introduce a full task reference such as `{address, generation}`, while preserving the existing process-qualified keys, with a generation source that does not reset when an LRU entry disappears? Context mappings and parent links should store the full reference; context resolution should return it; and the owning-trace walk should validate the generation before consuming `.conn` or following each parent hop. `obi_uprobe_task_init()` should allocate a fresh generation, capture the parent's full reference, and use that reference when inheriting its connection. Callers should never downgrade the reference back to a raw address. The regressions should cover reuse between context validation and network consumption, parent completion/reuse before detached-child traversal, and eviction followed by reuse while an old binding remains live.

bpf/common/python_task.h
```diff
@@ -11,44 +11,152 @@
...
+    if (!task_ref || !task_ref->addr || !task_ref->generation) {
+        return NULL;
     }
 
+    const python_addr_key_t task_key = python_addr_key(pid_tgid, task_ref->addr);
     const python_task_state_t *task_state =
-        (const python_task_state_t *)bpf_map_lookup_elem(&python_task_state, &context_task->task);
-    if (!task_state || task_state->version != context_task->version) {
+        (const python_task_state_t *)bpf_map_lookup_elem(&python_task_state, &task_key);
+    if (!task_state || task_state->generation != task_ref->generation) {
```
MrAlias, line 64: My earlier direction to validate the generation before consuming task state was incomplete: this check returns a pointer into an LRU node, and callers read `.conn` or `.parent` later. Replacing a reused task-address entry retires the old node to the LRU free list; another task-state insertion on another CPU can reuse and overwrite that node before those later loads, so a stale reference can still consume unrelated state and join the wrong trace. Could we copy task state through a synchronization/revalidation protocol, or otherwise ensure validated storage cannot be recycled while it is being consumed? The cancelled-worker test completes address reuse before releasing the worker, so it does not exercise node recycling during lookup.

devdocs/features.md
```diff
@@ -166,4 +166,4 @@ OBI has support for several asynchronous frameworks that allow it to propagate c
 | Ruby Puma Server    |   Ruby    |              N/A | Only works with Puma server                       | Stable
 | Java Thread pool    |   Java    |           JDK 8+ | N/A                                               | Stable
 | Java Virtual Threads |  Java    |          JDK 21+ | Log enrichment is skipped on virtual threads      | Stable
-| Python asyncio      |  Python   |    Python >= 3.9 | Only works with uvloop event loop                 | Stable
+| Python asyncio      |  Python   |    Python >= 3.9 | `asyncio.start_server()` is not correlated under uvloop; mutated contexts and cancelled `to_thread` tasks may lose correlation | Stable
```
MrAlias, line 169: This row now documents support for the default asyncio loop, but both Python asyncio entries in `SUPPORT_MATRIX.md` still say that uvloop is required and that propagation only works with uvloop. That is the opposite of the new `start_server` support and leaves the published support contract contradictory. Could we update both support-matrix rows in this PR?

MrAlias (commented):
The generation-bearing references address the ordinary stale-address paths, but recycling a validated LRU node can still race readers. The support documentation also needs to match the implementation and the newly supported event loop.

devdocs/python-asyncio-context-propagation.md
```diff
@@ -135,10 +142,13 @@ Runs when Python activates a `Context` on a thread.
 Responsibilities:
 
 - track which `PyContext*` is active on the current thread,
+- resolve the context's owner task from `python_context_task` and cache it in
+  `python_thread_state.current_context_task`,
```
MrAlias, line 146: `python_thread_state_t` has `current_task`, `current_context`, and `inflight_task`; there is no `current_context_task` cache. `context_run` stores `current_context`, and parent lookup resolves `python_context_task` from that context on each use. Could we update this section and the Phase 1 description below to match the implementation?

bpf/generictracer/python.c
```diff
@@ -255,29 +281,44 @@ int GUARDED_PROG(obi_uprobe_task_init, struct pt_regs *, ctx) {
...
     // Use the parent's connection when it exists. If there is no parent
     // connection yet, fall back to pid_tid_to_conn for the current thread.
     // pid_tid_to_conn is only thread-local and may already point to another
     // request by the time the child task is initialized.
-    if (parent_state && parent_state->conn.port) {
-        task_state.conn = parent_state->conn;
-    } else {
+    if (parent_state_found && parent_state.conn.port) {
+        task_state.conn = parent_state.conn;
+    } else if (parent_resolution != PYTHON_TASK_STALE) {
```
MrAlias, line 363: `PYTHON_TASK_NOT_FOUND` currently means both “there is no active Python owner” and “an active owner could not be resolved.” If `current_task` is set but its task-state entry has been evicted, or `current_context` is set but its binding has been evicted, both task initialization here and `find_python_parent_trace()` treat the result as permission to fall back. Task initialization may persist `pid_tid_to_conn` from another interleaved request, while outbound parent lookup may proceed through the generic fallbacks despite an unresolved active Python identity.

Could we reserve `NOT_FOUND` for no active identity, return `STALE` whenever lookup for an active task or context fails, and fail closed in both consumers?

bpf/common/python_task.h
```diff
@@ -11,44 +11,163 @@
...
 #include <maps/server_traces.h>
 
-static __always_inline u64 resolve_python_context_task(const python_context_task_t *context_task) {
-    if (!context_task || !context_task->task) {
+#include <pid/pid_helpers.h>
+
+// ctx_vars offset in struct _pycontextobject: PyObject_HEAD + ctx_prev.
+// Layout is identical across CPython 3.9-3.14 standard 64-bit builds.
+// See https://github.com/python/cpython/blob/3.14/Include/internal/pycore_context.h#L21-L27
+enum { k_python_context_vars_offset = 16 + 8 };
```
MrAlias, line 20: `24` matches regular GIL-enabled 64-bit CPython 3.9–3.14, but not free-threaded CPython 3.13/3.14, where a 32-byte `PyObject` puts `ctx_vars` at offset 40 and offset 24 reads `ob_type`. Because that type pointer is shared by all `Context` objects, the fallback identity check succeeds across recycled addresses. When the optional deallocation probe is unavailable and the old task generation remains live, a stale binding can therefore be accepted across requests.

Could we narrow and synchronize the documentation to the currently supported build scope in this PR? Runtime layout detection or gating can be tracked in a follow-up issue if preferred; until then, the support contract should exclude unsupported layouts.

bpf/common/python_task.h
```diff
@@ -11,44 +11,163 @@
...
-static __always_inline tp_info_pid_t *find_python_owning_server_trace(u64 task_id) {
+static __always_inline u64 allocate_python_task_generation(void) {
+    const u32 key = 0;
+    u64 *counter = (u64 *)bpf_map_lookup_elem(&python_task_generation, &key);
+    if (!counter) {
+        return 0;
+    }
+
+    __sync_fetch_and_add(counter, 1);
+    return *counter;
```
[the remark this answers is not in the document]
marctc: Generation checks use PID + task address, so duplicate values across different tasks are harmless

internal/test/integration/traces_test.go
```diff
@@ -1693,6 +1710,231 @@ func testPythonAsyncConcurrent(t *testing.T) {
...
+				body, readErr := io.ReadAll(resp.Body)
+				_ = resp.Body.Close()
+				err = readErr
+				if resp.StatusCode != http.StatusOK {
+					err = fmt.Errorf("%s: HTTP %d: %s", url, resp.StatusCode, body)
+				}
+			}
+			results <- err
+		}()
+	}
```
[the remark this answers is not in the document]
marctc: Each iteration gets its own i under this module’s Go version

bpf/generictracer/python.c
```diff
@@ -255,29 +281,44 @@ int GUARDED_PROG(obi_uprobe_task_init, struct pt_regs *, ctx) {
...
+        if (resolve_python_task_ref(id, thread_state->current_task, &parent_ref)) {
+            parent_resolution = PYTHON_TASK_RESOLVED;
+        }
+    } else {
+        parent_resolution =
+            resolve_python_task_from_context(id, thread_state->current_context, &parent_ref);
+    }
+    const python_addr_key_t child_task_key = python_addr_key(id, child_task);
+    const u64 generation = allocate_python_task_generation();
+    if (!generation) {
```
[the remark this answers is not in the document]
marctc: The initializer’s return probe already clears it.

bpf/generictracer/python.c
```diff
@@ -255,29 +281,44 @@ int GUARDED_PROG(obi_uprobe_task_init, struct pt_regs *, ctx) {
     }
 
     thread_state->inflight_task = child_task;
```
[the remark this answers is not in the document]
marctc: The initializer’s return probe already clears it.

MrAlias (approved):
This looks good enough to merge. Thanks for the work 😄 

We can keep the multi-hop `call_soon()` propagation gap and support documentation/guarding support as follow-up work.

grcevski:
So it seems with the latest change all Python AsyncIO tests are failing, I think it's related...

marctc:
> So it seems with the latest change all Python AsyncIO tests are failing, I think it's related...

yeah last changes after MrAlias feedback broke the tests...i have to look into it

marctc:
After applying @MrAlias’s feedback, the integration tests caught another case: a fresh empty context has no task owner, but we were treating it as stale. The new probe records that it has no owner, so we can still find the parent through the connection while keeping the stale-state checks.

PR #2839, opened 2026-07-29
Add Aerospike to v2 instrumentation schema and protocol mappings

MrAlias:
### Motivation

- Aerospike was enabled by default in runtime instrumentations but missing from the v2 schema and `protocolMappings`, making it impossible for operators to disable Aerospike via the v2 config path. This change closes that gap so v2 enablement conversion can control Aerospike selection.

### Description

- Add the `aerospike` protocol name and mapping so `protocolMappings` covers Aerospike and conversion logic can update `InstrumentationAerospike`.
- Expose `aerospike` in the v2 schema model and wire it into export/import conversion paths so `capture.instrumentation.aerospike` is present in v2 documents and conversion completeness checks.
- Update the v2 JSON schema, default example, and docs to include the Aerospike instrumentation block.
- Preserve optional Aerospike omission/defaulting and explicit trace and metric enablement or disablement.
- Gate Aerospike application metrics independently in the OTLP and Prometheus exporters.
- Integrate Aerospike with the shared application-filter validation merged in #2815.

### Filter behavior

Aerospike participates in the shared application-filter validation introduced by #2815. Because the runtime currently has one application filter, Aerospike filters that differ from the canonical shared filter are rejected rather than silently ignored. Full protocol- and signal-scoped runtime filtering is tracked by #1282.

### Testing

- Ran `gofmt` on all changed Go files.
- `go test ./internal/config/schema ./internal/config/convert ./pkg/export/otel ./pkg/export/prom -count=1`; passed.
- `make check-config-v2-artifacts`; passed.
- `make -o prereqs verify`; passed.
- `git diff --check`; passed.
- The repository-pinned Markdown linter passed all 74 Markdown files with zero issues. The `make lint-markdown` wrapper was blocked locally because its container UID could not read this worktree mount, so the same pinned image was run as the worktree owner UID.


marctc (approved):
LGTM!

mmat11 (approved):

PR #2845, opened 2026-07-30
fix(genai): add execute tool args result

NameHaibinZhang:
## Summary

Adds two OTel GenAI semantic-convention attributes to execute_tool spans:

gen_ai.tool.call.arguments — tool call input arguments (JSON string)
gen_ai.tool.call.result — tool call return value (JSON string)
Spec: https://github.com/open-telemetry/semantic-conventions-genai/blob/main/model/gen-ai/spans.yaml

## Validation

## Behavior change

execute_tool spans are now emitted by pairing assistant `tool_calls` with their tool-role results over the request message history, instead of from the `tool_calls` present in an LLM response. Spans are emitted only for tool calls whose result is sent back in a subsequent request, so each execute_tool span describes a real execution and carries both arguments and result. Response-declared tool calls without a captured result no longer produce execute_tool spans.


marctc (approved):
LGTM! (same [comment](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2804#pullrequestreview-4818044801) applies here :)

pkg/appolly/app/request/genai_tool_pairing.go
```diff
@@ -0,0 +1,248 @@
...
+		used[idx] = true
+		out = append(out, ToolCall{
+			ID:        c.id,
+			Name:      c.name,
+			Arguments: jsonRawToAttrString(c.args),
+			Result:    jsonRawToAttrString(results[idx].result),
+		})
+	}
+	return out
+}
```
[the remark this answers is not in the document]
NameHaibinZhang: Fixed. Added an `emptyJSON` helper that treats a JSON `null` literal (after trimming whitespace) the same as an absent payload: tool calls with `null` arguments are skipped, and tool-role results with `null` content are dropped before pairing, so the literal string "null" can no longer reach the exported attributes. This also makes the behavior match the function comment. Covered by the new `TestPairedToolCalls_NullArgumentsAndResult`.

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -458,27 +467,30 @@ var (
...
-		return span.GenAI.Qwen.ToolCalls
-	case span.GenAI.Ollama != nil:
-		return span.GenAI.Ollama.ToolCalls
-	case span.GenAI.OpenAICompatible != nil:
-		return span.GenAI.OpenAICompatible.ToolCalls
-	default:
+	toolCalls := span.GenAI.PairedToolCalls()
+	if len(toolCalls) == 0 {
 		return nil
 	}
```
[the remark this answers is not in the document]
NameHaibinZhang: This behavior change is intentional; I've updated the PR description to call it out explicitly.

Emitting execute_tool spans from the response `tool_calls` was semantically premature: at that point the tool has not run yet (the response merely requests the call), so the span described an execution that may never happen, with timestamps copied from an LLM span that ends before the tool actually executes. Pairing over the request message history guarantees each execute_tool span reflects a real execution and carries both `gen_ai.tool.call.arguments` and `gen_ai.tool.call.result` per the GenAI semantic conventions.

Falling back to the vendor `ToolCalls` when `PairedToolCalls()` returns empty would double-emit in the common multi-turn flow: the turn-N response span would get a name/id-only execute_tool child, and the turn-N+1 request span would get the fully populated one for the same invocation, with no cross-span state available to deduplicate them. Happy to revisit if dropping the name/id-only spans turns out to be a problem in practice.

pkg/appolly/app/request/genai_tool_pairing.go
```diff
@@ -0,0 +1,249 @@
...
+		ToolName   string `json:"tool_name"`
+		Name       string `json:"name"`
+	}
+	if err := json.Unmarshal(raw, &msgs); err != nil {
+		return nil
+	}
+
+	var calls []pendingToolCall
+	var results []pendingToolResult
+	for i := range msgs {
```
MrAlias: Because this scans the entire history, clients that retain the conversation re-emit every completed tool call on later requests. That duplicates `execute_tool` spans and makes their count track history length rather than actual executions.

For these message schemas, the current execution batch is identifiable without cross-request state: find the most recent assistant/model message and pair only its tool calls with results that follow it. If that message has no tool calls, there is nothing new to emit. Could we scope pairing to that suffix?

pkg/appolly/app/request/genai_tool_pairing.go
```diff
@@ -0,0 +1,249 @@
...
+		return nil
+	}
+
+	var calls []pendingToolCall
+	var results []pendingToolResult
+	for i := range contents {
+		for j := range contents[i].Parts {
+			p := contents[i].Parts[j]
+			if p.FunctionCall != nil {
+				calls = append(calls, pendingToolCall{name: p.FunctionCall.Name, args: p.FunctionCall.Args})
```
MrAlias: Gemini function calls and responses can carry IDs, and Gemini 3 requires preserving them for parallel calls. Dropping those IDs and pairing only by name cross-matches two same-name calls when their responses arrive in a different order, so arguments from one execution can be attached to another result.

Could we retain the `id` from both `functionCall` and `functionResponse` and use the existing ID matching path, falling back to names only when the provider genuinely omits it? See [Gemini's function-calling guidance](https://ai.google.dev/gemini-api/docs/generate-content/function-calling?hl=en).

pkg/appolly/app/request/genai_tool_pairing.go
```diff
@@ -0,0 +1,249 @@
...
+		if err := json.Unmarshal(msgs[i].Content, &blocks); err != nil {
+			continue
+		}
+		for j := range blocks {
+			b := &blocks[j]
+			switch b.Type {
+			case "tool_use":
+				calls = append(calls, pendingToolCall{id: b.ID, name: b.Name, args: b.Input})
+			case "tool_result":
+				results = append(results, pendingToolResult{id: b.ToolUseID, result: b.Content})
```
MrAlias: Anthropic `tool_result` blocks can set `is_error: true`, but this parser drops that flag and treats the payload like an ordinary result. Failed tool executions therefore get `gen_ai.tool.call.result` with no error signal, even though that attribute is defined for successful executions.

Could we carry `is_error` through pairing so failed calls omit the success-only result attribute and the generated `execute_tool` span exposes the failure? See [Anthropic's tool-result guidance](https://platform.claude.com/docs/en/agents-and-tools/tool-use/handle-tool-calls) and the [GenAI attribute definition](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/model/gen-ai/registry.yaml).

MrAlias (changes requested):
The request-history pairing re-emits completed tool executions on later turns, and the provider-specific gaps below can also make the resulting execution telemetry inaccurate.

pkg/appolly/app/request/genai_tool_pairing.go
NameHaibinZhang: Fixed in e65bcf27. Pairing is now scoped to the latest assistant/model message in all three schema parsers: if that message carries no tool calls nothing is emitted, and results are only collected from messages that follow it, so retained conversation history no longer re-emits completed tool calls on later requests. Added tests covering both the "final assistant answer" and "new call in latest turn" cases.

pkg/appolly/app/request/genai_tool_pairing.go
NameHaibinZhang: Fixed in e65bcf27. functionCall/functionResponse ids are now preserved and flow through the existing id-matching path, falling back to name-based pairing only when the provider omits ids. Added a test with two parallel same-name calls whose responses arrive out of order.

pkg/appolly/app/request/genai_tool_pairing.go
NameHaibinZhang: Fixed in e65bcf27. is_error is now carried through pairing: failed calls keep their pairing (so the execution is still visible) but omit the success-only gen_ai.tool.call.result attribute, and the generated execute_tool span status is set to Error. Added a test for the is_error path.

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -259,8 +264,8 @@ func generateTracesWithAttributes(
 		s.SetEndTimestamp(pcommon.NewTimestampFromTime(t.End))
 
 		// Create individual execute_tool child spans per tool call (OTel GenAI semconv compliance)
-		if toolCalls := getSpanToolCalls(span); len(toolCalls) > 0 {
-			createToolCallSpans(toolCalls, spanID, traceID, &ss, start, t.End)
+		if len(spanWithAttributes.ToolCalls) > 0 {
+			createToolCallSpans(spanWithAttributes.ToolCalls, spanID, traceID, &ss, start, t.End)
```
MrAlias: The tool result is only observable in request N+1, but this emits the synthetic tool span as a child of N+1 and gives it N+1's inference start/end timestamps. That reverses the causal order and reports model latency as tool duration; request history proves that the tool completed, not when it ran. Could we correlate the result back to the originating turn with actual execution timing, or avoid emitting an `execute_tool` span when those facts are unavailable?

pkg/appolly/app/request/genai_tool_pairing.go
```diff
@@ -0,0 +1,327 @@
...
+// assistant/model message so that clients retaining the full conversation do
+// not re-emit previously completed tool calls on later requests. Only fully
+// paired tool calls are returned.
+func (g *GenAI) PairedToolCalls() []ToolCall {
+	if g == nil {
+		return nil
+	}
+	switch {
+	case g.OpenAI != nil:
+		return pairOpenAIToolCalls(g.OpenAI.Request.Messages)
```
MrAlias: `/v1/responses` carries tool calls and results in `input` items (`function_call` / `function_call_output`) keyed by `call_id`, not in Chat Completions `messages`. `OpenAIInput.Input` is a string and cannot retain the array form, so this receives an empty `Request.Messages` and never emits tool spans for Responses API traffic. Could we preserve the raw `input` and pair the Responses schema when `APIType == "responses"`? [OpenAI Responses API](https://platform.openai.com/docs/api-reference/responses/create)

MrAlias (commented):
The request-history model does not yet produce reliable `execute_tool` telemetry across the APIs this package supports: it assigns prior tool work to the next inference span and ignores request shapes outside Chat Completions messages.

pkg/appolly/app/request/genai_tool_pairing.go
```diff
@@ -0,0 +1,327 @@
...
+	case g.OpenAICompatible != nil:
+		return pairOpenAIToolCalls(g.OpenAICompatible.Request.Messages)
+	case g.Qwen != nil:
+		return pairOpenAIToolCalls(g.Qwen.Request.Messages)
```
MrAlias: Native DashScope generation wraps conversation history under `input.messages`; only compatible-mode requests put `messages` at the top level. Because `QwenSpan` parses the native request through `OpenAIInput`, this field is empty and native Qwen tool executions are silently dropped. Could we retain and parse `input.messages` for the native endpoint? [Qwen Function Calling](https://help.aliyun.com/zh/model-studio/qwen-function-calling)

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -504,6 +510,15 @@ func createToolCallSpans(toolCalls []request.ToolCall, parentSpanID pcommon.Span
...
 			attrs.PutStr(string(attr.GenAIToolCallID), tc.ID)
 		}
+		if tc.Arguments != "" {
+			attrs.PutStr(string(attr.GenAIToolCallArguments), tc.Arguments)
+		}
+		if tc.Result != "" {
+			attrs.PutStr(string(attr.GenAIToolCallResult), tc.Result)
+		}
+		if tc.IsError {
+			sp.Status().SetCode(ptrace.StatusCodeError)
```
MrAlias: Setting span status to Error is not sufficient for an `execute_tool` failure: the v1.41 convention conditionally requires `error.type` whenever the operation ends in error. `is_error` does not provide a provider error code, so this should set a documented low-cardinality fallback rather than deriving a value from the tool result. [GenAI span conventions](https://github.com/open-telemetry/semantic-conventions/blob/v1.41.0/docs/gen-ai/gen-ai-spans.md#execute-tool-span)

pkg/export/otel/tracesgen/tracesgen.go
NameHaibinZhang: The tool runs outside the observed process, so its real execution time isn't visible to eBPF — the request history only proves the call completed by the time the request was sent, not when it ran or for how long. Rather than reuse the N+1 inference window (which reported model latency as tool duration), the execute_tool span is now emitted zero-duration at the request start. Updated TestCreateToolCallSpans to assert the zero-duration timing.

pkg/export/otel/tracesgen/tracesgen.go
NameHaibinZhang: Failed calls now also set error.type = "_OTHER". is_error carries no provider error code, so per the v1.41 convention I use the documented low-cardinality fallback rather than deriving a value from the tool result. Covered by the is_error test.

pkg/appolly/app/request/genai_tool_pairing.go
NameHaibinZhang: The raw input array is now retained (OpenAIInput.InputItems) and, when APIType == "responses", paired through a dedicated Responses parser that matches function_call / function_call_output items by call_id. The same latest-turn scoping applies: only the trailing run of function_call items is paired, and an assistant message after them means the turn already concluded, so retained history doesn't re-emit. Added tests for the happy path, sequential-latest-only, and the completed-turn (no re-emit) case.

pkg/appolly/app/request/genai_tool_pairing.go
NameHaibinZhang: Native DashScope nests the conversation under input.messages, so when the top-level messages is empty the parser now falls back to input.messages; native Qwen tool executions then flow through the existing OpenAI messages pairing. Added a parser test for the nested shape.

pkg/appolly/app/request/genai_tool_pairing.go
```diff
@@ -0,0 +1,399 @@
...
+		return nil
+	}
+
+	lastCall := -1
+	for i := range items {
+		if items[i].Type == "function_call" {
+			lastCall = i
+		}
+	}
+	if lastCall < 0 {
```
MrAlias: Stateful Responses continuations do not include the prior `function_call` in the current `input`; with `previous_response_id` or a Conversation, the client can send only the new `function_call_output`. This leaves `lastCall == -1` and drops the completed execution even though its `call_id` is present. Could we retain enough cross-request state to correlate these outputs, or explicitly limit this instrumentation to manually replayed histories? [OpenAI conversation state](https://developers.openai.com/api/docs/guides/conversation-state)

pkg/appolly/app/request/genai_tool_pairing.go
```diff
@@ -0,0 +1,399 @@
...
+		return emptyJSON(r.result) && !r.isError
+	})
+	if len(calls) == 0 || len(results) == 0 {
+		return nil
+	}
+	used := make([]bool, len(results))
+	var out []ToolCall
+	for i := range calls {
+		c := &calls[i]
+		if emptyJSON(c.args) {
```
MrAlias: Gemini's `FunctionCall.args` field is optional, so an argumentless call can have a valid name, ID, and response while deserializing with empty `args`. Skipping it here drops the entire `execute_tool` span rather than only omitting `gen_ai.tool.call.arguments`. Could we allow missing arguments through pairing and leave the optional attribute unset? [Gemini FunctionCall schema](https://ai.google.dev/api/generate-content#FunctionCall)

MrAlias (commented):
The current request-history model still cannot produce reliable `execute_tool` telemetry: it attaches prior work to the following inference span and drops valid executions when the current payload does not replay a complete call.

pkg/appolly/app/request/genai_tool_pairing.go
NameHaibinZhang: Fixed: pairing no longer drops calls with empty arguments. A call is paired whenever it has a matching result, and missing arguments just leave `gen_ai.tool.call.arguments` unset. Added a Gemini argumentless-call test.

pkg/appolly/app/request/genai_tool_pairing.go
NameHaibinZhang: Documented as an explicit limitation: pairing needs the `function_call` and its `function_call_output` in the same request, so stateful continuations (`previous_response_id` / Conversation) that replay only the output aren't emitted. Correlating them needs cross-request state, which this instrumentation avoids — flagging on the top-level thread for a direction decision.

NameHaibinZhang:
@MrAlias  You're right that request-history pairing can't reconstruct real tool timing or parent the execution to the originating turn: the tool runs outside the observed process, and only its completion is evidenced by a later request. Rather than fabricate timing, I've bounded the distortion — execute_tool spans are emitted zero-duration at the request start (so model latency is never reported as tool duration), and argumentless/failed calls no longer distort or drop.

What remains is inherent to stateless, per-request instrumentation: parenting to the originating turn and true execution timing would need cross-request state, which this PR deliberately avoids. I've scoped the feature as best-effort over replayed histories.

Before going further, which do you prefer: (a) keep it explicitly scoped as best-effort for replayed histories, or (b) drop the synthesized child spans and carry tool-call info on the inference span instead? Happy to take either.

MrAlias:
Thanks—that framing helps. I prefer (b), with one qualification: represent the observed tool calls and responses through `gen_ai.output.messages` and `gen_ai.input.messages`, rather than putting the singular `gen_ai.tool.call.*` attributes directly on the inference span.

A zero-duration child at request N+1 avoids reporting model latency as tool duration, but it still asserts that the tool executed at N+1’s start and under N+1’s inference span. We do not observe either fact. The semantic convention describes `execute_tool` as the actual execution operation and points instrumentation toward the execution site; the model’s tool-call request and the later tool response naturally belong in the inference output and input messages.

OBI already has most of that normalization. I suggest removing the synthetic `execute_tool` spans for provider response/history paths, while leaving directly observed MCP operations alone. We can retain the useful parsing fixes from this PR, wire Responses input arrays into `gen_ai.input.messages` so output-only stateful continuations remain visible, and preserve Gemini call IDs in the normalized messages. These content attributes should remain opt-in.

Cross-request state would improve correlation, but it still could not recover the actual execution timestamps, executor, parent, or status, so I don’t think it would make a synthesized execution span reliable. If inferred completion telemetry is valuable later, we should design it separately as an explicitly opt-in OBI-specific event marked as inferred.

NameHaibinZhang:
@MrAlias Pushed the (b) changes. One thing I'd like to align on before finalizing.

Removing the `execute_tool` span for provider paths does keep the tool call/response *content* (it moves into `gen_ai.output.messages` / `gen_ai.input.messages`), but it drops two things that the span gave us:

- A discrete entry in the trace waterfall for "the model invoked tool X".
- Low-cardinality, directly queryable/aggregatable signals (`gen_ai.operation.name = execute_tool`, `gen_ai.tool.name`) for the provider path. MCP-observed tool spans still have these.

The bigger practical gap is that message content is opt-in, so when content capture is off, provider-path tool usage becomes entirely invisible — whereas the span previously exposed `gen_ai.tool.name` without capturing any content.

I agree we should not fabricate a span that asserts unobserved execution time/parent/executor. But your comment also pointed at a path I like: an explicitly opt-in, `inferred`-marked OBI signal. Would you be open to that — either in this PR or as a follow-up — so provider-path tool usage stays discoverable without claiming a real execution? If you'd rather keep this PR strictly to the messages representation and design the inferred signal separately, that works for me too; I just want to make sure we do not silently lose tool-usage visibility when content is disabled.

MrAlias (approved):

Issue #2801, opened 2026-07-28
PID filter fails open when valid_pids lookup misses, silently disabling discovery filtering


bourbonkk:
### Component

OBI (eBPF instrumentation)

### What happened?

`pid_matches()` treats a failed `valid_pids` lookup as a match:

```c
// bpf/pid/pid.h:31-36
u64 *v = bpf_map_lookup_elem(&valid_pids, &segment);
if (!v) {
    // This is an error of some kind, we should always find the segment
    bpf_dbg_printk("Error looking up PID, segment=%d", segment);
    return 1;
}
```

When the lookup misses, the PID is reported as tracked. `pid_matches()` is called
only from `valid_pid()` (`bpf/pid/pid.h:65,73`), which in turn gates 54 probe
entry points across the generic tracer (`k_tracer.c`, `libssl.c`, `python.c`,
`ruby.c`, `nginx.c`, `jvm.c`, `java_tls.c`, `k_unix_sock.h`), plus `tpinjector`,
`gpuevent` and `common/usdt.h`. So a lookup failure does not degrade one probe —
it disables `discovery` filtering for the affected PIDs everywhere at once, and
those processes get instrumented unconditionally regardless of
`discovery.instrument` / `exclude_instrument`.

The failure is effectively invisible. The only signal is `bpf_dbg_printk`, which
requires `ebpf.bpf_debug` to be enabled, and there is no counter or metric for it.

### What did you expect to happen?

A `valid_pids` lookup miss means the map does not match the userspace view of the
PID hash space, which is a bug rather than a routine condition. It should not
silently widen instrumentation scope. Either the mismatch is rejected at load
time, or the miss is observable.

### Steps to reproduce

Found via #2800. Before that fix, `ebpf.maps_config.global_scale_factor: -1`
shrank `valid_pids` from 3001 to 1500 entries while `rebuildValidPids` still
wrote indices 0..3000, so segments 1500..3000 could never be populated and every
lookup into the upper half of the hash space missed.

1. On a build without #2800, set `ebpf.maps_config.global_scale_factor: -1`
2. Instrument a namespace with `discovery.instrument`
3. Processes whose PID hashes into the upper half of the space are instrumented
   even when they are outside the configured scope

#2800 removes this particular trigger, but not the fail-open behaviour itself.
Any other cause of a missing segment produces the same silent outcome.

### OBI configuration

```yaml
ebpf:
  maps_config:
[...]

PR #2846, opened 2026-07-30
fix(ebpf): enforce the valid_pids index space invariant

bourbonkk:
## Summary

`pid_matches()` treats a failed `valid_pids` lookup as a match, so a map too small for the segment index space would disable discovery filtering across all probes at once. The sizing currently works out with zero headroom — `(k_prime_hash - 1) / 64` is exactly `k_max_concurrent_pids - 1` — but nothing enforced it on either side.

- `bpf/pid/pid.h`: `_Static_assert` on the invariant
- `generictracer.go`: validate the loaded map's `MaxEntries` before writing to it
- `rebuildValidPids` returns an error and fails fast, instead of logging once per segment

Refs #2801.

This hardens against regressions rather than fixing a live fail-open: the lookup cannot currently miss (segment tops out at 3000, the array holds 3001), and #2800 already removed the known trigger. The stale comment on the Go constants pointed at `#define`s that no longer exist, which is how the invariant could drift unnoticed.

## AI usage

Per AI-POLICY §5: Claude Code was used to analyze issue #2801, write these changes, and run the verification below. I reviewed the resulting diff before submitting.

Verification run:
- BPF compile (`-Werror`) for amd64 + arm64
- assert confirmed to fire at `k_prime_hash=192065`, pass at `192064`
- `go vet` clean; `./pkg/internal/ebpf/...` tests pass

## Validation

### DIFF
--- a/pkg/internal/ebpf/generictracer/generictracer.go
+++ b/pkg/internal/ebpf/generictracer/generictracer.go
@@ -133,10 +133,6 @@ func (p *Tracer) rebuildValidPids() error {
 		return nil
 	}
 
-	if err := p.validateValidPidsMap(); err != nil {
-		return err
-	}
-
 	v := p.buildPidFilter()
 
 	p.log.Debug("number of segments in pid filter cache", "len", len(v))
@@ -667,8 +663,11 @@ func (p *Tracer) Run(
 	// At this point we now have loaded the bpf objects, which means we should insert any
 	// pids that are allowed into the bpf map
 	if p.bpfObjects.ValidPids != nil {
+		if err := p.validateValidPidsMap(); err != nil {
+			p.log.Error("BPF PID filter map sizing is invalid, discovery filtering may not be enforced", "error", err)
+		}
 		if err := p.rebuildValidPids(); err != nil {
-			p.log.Error("Error setting up the BPF PID filter, discovery filtering may not be enforced", "error", err)
+			p.log.Error("setting up the BPF PID filter, discovery filtering may not be enforced", "error", err)
 		}
 	} else {
 		p.log.Error("BPF Pids map is not created yet, this is a bug.")
--- a/pkg/internal/ebpf/generictracer/generictracer.go
+++ b/pkg/internal/ebpf/generictracer/generictracer.go
@@ -74,9 +74,8 @@ func New(pidFilter ebpfcommon.ServiceFilter, cfg *obi.Config, metrics imetrics.R
 	}
 }
 
-// Updating these requires updating the constants below in pid.h
-// #define MAX_CONCURRENT_PIDS 3001 // estimate: 1000 concurrent processes (including children) * 3 namespaces per pid
-// #define PRIME_HASH 192053 // closest prime to 3001 * 64
+// Keep in sync with k_max_concurrent_pids (bpf/pid/maps/map_sizing.h) and
+// k_prime_hash (bpf/pid/pid.h), which assert they agree at compile time.
 const (
 	maxConcurrentPids = 3001
 	primeHash         = 192053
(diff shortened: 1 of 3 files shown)


mariomac (approved):
Looks good! Thanks for the contribution. Waiting for an automated review before merging

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -74,9 +74,8 @@ func New(pidFilter ebpfcommon.ServiceFilter, cfg *obi.Config, metrics imetrics.R
 	}
 }
 
-// Updating these requires updating the constants below in pid.h
-// #define MAX_CONCURRENT_PIDS 3001 // estimate: 1000 concurrent processes (including children) * 3 namespaces per pid
-// #define PRIME_HASH 192053 // closest prime to 3001 * 64
+// Keep in sync with k_max_concurrent_pids (bpf/pid/maps/map_sizing.h) and
```
mmat11: nit: can we leave comments explaining why those numbers for those constants?

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -110,24 +109,45 @@ func (p *Tracer) buildPidFilter() []uint64 {
 	return result
 }
 
-func (p *Tracer) rebuildValidPids() {
-	if p.bpfObjects.ValidPids != nil {
-		v := p.buildPidFilter()
+// validateValidPidsMap rejects a map too small to hold every segment index, which
+// would make pid_matches() miss and fail open.
+func (p *Tracer) validateValidPidsMap() error {
+	if got := p.bpfObjects.ValidPids.MaxEntries(); got != maxConcurrentPids {
```
mmat11, line 122: as copilot says, this checks for equality -- question: why do we need this check at all? is it because of map scale factor? if yes, I think we can exclude this specific map from there

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -110,24 +109,45 @@ func (p *Tracer) buildPidFilter() []uint64 {
...
 	}
+
+	return nil
 }
 
 func (p *Tracer) AllowPID(pid app.PID, ns uint32, fi *exec.FileInfo) {
 	p.pidsFilter.AllowPID(pid, ns, fi, ebpfcommon.PIDTypeKProbes)
-	p.rebuildValidPids()
+	if err := p.rebuildValidPids(); err != nil {
+		p.log.Error("Error rebuilding the BPF PID filter", "error", err)
```
mmat11: should we early return here?

also nit: log messages should start lowercase

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -137,7 +157,9 @@ func (p *Tracer) AllowPID(pid app.PID, ns uint32, fi *exec.FileInfo) {
 
 func (p *Tracer) BlockPID(pid app.PID, ns uint32) {
 	p.pidsFilter.BlockPID(pid, ns)
-	p.rebuildValidPids()
+	if err := p.rebuildValidPids(); err != nil {
+		p.log.Error("Error rebuilding the BPF PID filter", "error", err)
```
mmat11: same here, early return?

pkg/internal/ebpf/generictracer/generictracer.go
bourbonkk: Added comments explaining where both values come from in 1a10270c.

pkg/internal/ebpf/generictracer/generictracer.go
bourbonkk: It is not about the scale factor -- array maps are already excluded from `SetupMapSizes` since #2800, so this map is never scaled. The check guards against the Go-side constant drifting from `k_max_concurrent_pids`: the two are duplicated with nothing tying them together at compile time, and the loaded map's `MaxEntries` reflects what the BPF object was actually compiled with. Equality is intentional because both directions misbehave: an undersized map fails open (lookup miss), while an oversized one leaves segments unset, silently filtering out matching PIDs. Reworded the comment and error message accordingly in 1a10270c.

pkg/internal/ebpf/generictracer/generictracer.go
bourbonkk: Done in 1a10270c -- early return so the pid_cache is not touched when the rebuild fails, and lowercased the log messages.

pkg/internal/ebpf/generictracer/generictracer.go
bourbonkk: Done in 1a10270c.

pkg/internal/ebpf/generictracer/generictracer.go
mmat11: thanks

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -110,24 +114,50 @@ func (p *Tracer) buildPidFilter() []uint64 {
...
-			err := p.bpfObjects.ValidPids.Put(uint32(i), segment)
-			if err != nil {
-				p.log.Error("Error setting up pid in BPF space, sizes of Go and BPF maps don't match", "error", err, "i", i)
-			}
+func (p *Tracer) rebuildValidPids() error {
+	if p.bpfObjects.ValidPids == nil {
+		return nil
+	}
+
+	if err := p.validateValidPidsMap(); err != nil {
```
mmat11: this is called for every single allowPid() and blockPid(), to enforce some compile-time constant. Can we move this check somewhere outside where it's checked once during the map lifetime?

bourbonkk: Good point -- moved the validation to Run(), right after the objects are loaded, so it runs once per tracer lifetime (27330662). rebuildValidPids() is now just the map rewrite.

MrAlias (approved):

mmat11 (approved):
thanks!

PR #2848, opened 2026-07-30
Fix instrumentation for executables sharing an inode across filesystems

### LINKED ISSUE #2838: Go tracer reuse keyed on bare inode — only one of N identical-image pods on a node gets uprobes attached
## Summary

`traceAttacher.existingTracers` is keyed by **inode number alone**. On `overlayfs`, sibling containers running the **same image** on one node share the same inode *number* but have **different device** numbers. When N such pods run a **Go** service on a node, the second and subsequent pods collide on the inode key and are routed into the tracer-reuse branch of `getTracer` — and that branch, for a Go tracer, **never attaches uprobes to the new executable**. Result: only the first pod emits spans; the rest emit zero, while metrics and logs both report that all N were instrumented.

Non-Go (Generic) tracers are unaffected because their reuse branch does re-attach.

## Environment

- **OBI:** reproduced against the code in `pkg/appolly/discover/attacher.go` as of **v0.10.0** (line numbers below are v0.10.0). The same logic is present unchanged in the v0.9.0-era code.
- Observed in the field via **Grafana Beyla 3.25.0**, which vendors OBI as a submodule pinned to commit `4bf91071d3ecbc38824bc123a30f149caa1f41f1` (identical `attacher.go` to v0.10.0; the line numbers below match that commit exactly).
- Platform: **AWS EKS / Bottlerocket**, container runtime storage on **overlayfs**.
- Workload: 5 replicas of the same Go image (telegraf) scheduled onto one node.

## Symptom

Five pods, same Go image, same node. Only **one** pod's traffic appeared in the trace backend:

- `telegraf-3` (~23 req/s) → spans present in Tempo
- `telegraf-5` (~560 metrics/s) → **zero** spans, despite being the busiest

## Root cause

`existingTracers` is declared and keyed on inode only:

```go
// pkg/appolly/discover/attacher.go:49
existingTracers     map[uint64]*ebpf.ProcessTracer
```

Lookups/stores use `ie.FileInfo.Ino()` as the sole key (`:147` lookup, `:263` store, `:348` store in `reuseTracer`, `:444`/`:465` in the deletion path).

On overlayfs each container gets its own device, and inode **numbers are reused across layers**, so identical-image siblings collide on the key:

```
pid=721736  dev=1048607  ino=402839243  /usr/bin/telegraf
pid=721760  dev=1048666  ino=402839243   <- same ino, different dev
pid=722361  dev=1048713  ino=402839243   <- same ino, different dev
```

The first pod stores its tracer at `existingTracers[ino]` (and sets `reusableGoTracer`). Each sibling then hits the reuse branch at the top of `getTracer`:

```go
// pkg/appolly/discover/attacher.go:147
if tracer, ok := ta.existingTracers[ie.FileInfo.Ino()]; ok {
    ...
    ta.monitorPIDs(tracer, ie)
    ta.Metrics.InstrumentProcess(ie.FileInfo.ExecutableName())   // :158  (unconditional!)
    if tracer.Type == ebpf.Generic {
        ok = ta.updateTracerProbes(tracer, ie)                   // :164  Generic -> re-attaches
    } else {
        ta.monitorPIDs(ta.reusableGoTracer, ie)                  // :166  Go -> only AllowPID, NO attach
    }
    return ok
}
```

- The **Generic** branch (`:164`) calls `updateTracerProbes` → `tracer.NewExecutableInstance(ie)`, which attaches uprobes to the new executable.
- The **Go** else-branch (`:165-167`) calls only `monitorPIDs(ta.reusableGoTracer, ie)`. `monitorPIDs` just runs `AllowPID` over the new PID + children — it **never calls `NewExecutable`**, so no uprobes are registered for the sibling's binary.

This matters at the kernel level specifically because of the same overlayfs property that causes the collision: a uprobe is registered against a `(file inode object, offset)`, and the sibling's binary is a **different device → different kernel inode object** even though the inode *number* matches. The first pod's uprobes therefore do not fire for the sibling; the sibling needs its own `NewExecutable`.

Note that had the collision *not* happened, the sibling would have fallen through to the correct reuse path lower in `getTracer`:

```go
// :206
if ta.reusableGoTracer != nil {
    return ta.reuseTracer(ta.reusableGoTracer, ie)
}
```

and `reuseTracer` **does** attach:

```go
// :331
func (ta *traceAttacher) reuseTracer(tracer *ebpf.ProcessTracer, ie *ebpf.Instrumentable) bool {
    exe, ok := ta.loadExecutable(ie)
    if !ok { return false }
    if err := tracer.NewExecutable(exe, ie); err != nil {   // :337  attaches uprobes
        ...
    }
    ta.monitorPIDs(tracer, ie)
    ta.existingTracers[ie.FileInfo.Ino()] = tracer
    ...
}
```

So the bare-inode key is what diverts the sibling into the defective `:165-167` else-branch **before** it can reach the correct `:206 → reuseTracer → NewExecutable` path.

For comparison, the sibling ELF-typing cache in `typer.go` already keys correctly on `{Dev, Ino}`:

```go
// pkg/appolly/discover/typer.go
type cacheKey struct {
    Dev uint64
    Ino uint64
}
...
t.instrumentableCache.Get(cacheKey{Dev: execElf.Dev, Ino: execElf.Ino})
```

## Why this is hard to spot (misleads operators)

Neither logs nor metrics reveal the missing spans — only querying the trace backend **per pod** does:

- **`obi_instrumented_processes` / `beyla_instrumented_processes` reports all N.** `Metrics.InstrumentProcess()` is called unconditionally on the reuse path (`:158`), before the Go/Generic split — so OBI *believes* it instrumented every pod.
- **Only one `"instrumenting process"` INFO line appears** (the fresh-attach path). The reuse path logs at **Debug** (`"new process for already instrumented executable"`), so at default log level the siblings are silent.

## Reproduction

1. Build any small Go HTTP service; deploy N≥2 replicas of the **same image** and force them onto **one node** (e.g. `podAntiAffinity: false` / a single-node pool) on a container runtime using overlayfs (EKS/Bottlerocket, containerd default).
2. Run OBI as a DaemonSet instrumenting them.
3. Generate traffic to **every** replica.
4. Query the trace backend per pod: only one replica has spans. `obi_instrumented_processes` shows all N.

Confirm the collision on the node:

```
# WRONG — returns the procfs symlink's own inode (unique per PID); hides the collision:
stat /proc/<pid>/exe

# RIGHT — follows the symlink to the real file; shows same Ino, different Dev across pods:
stat -L /proc/<pid>/exe
```

## Suggested fix

Either of:

1. **Key `existingTracers` on `{Dev, Ino}`** (a `cacheKey`-style struct), matching what `typer.go` already does. This is the more complete fix: it also naturally forces siblings down the fresh-attach path since the collision disappears. The `processInstances` counter and the deletion path (`:444`/`:465`) would need the same key.
2. **Attach on the Go reuse branch**: have the `:165-167` else-branch call `NewExecutable` on the sibling's executable (as `reuseTracer`/`updateTracerProbes` already do) rather than only `monitorPIDs`.

Fix (1) is preferable because it fixes the root cause rather than one symptom of it.

## Workaround

`kubectl rollout restart` — pods land with fresh inodes and start emitting spans within ~60–90s. Not durable; the collision can recur on any node whenever identical-image siblings happen to share an inode number.

## Related

- #1683 — stale inode-keyed cache causing SIGSEGV on binary replacement. Same `existingTracers[Ino]` dedup path; that fix addressed a different failure mode of the same bare-inode key.
- #2638 — references the same `existingTracers[Ino()] = tracer` retention behavior.

Happy to test a patched image or help with a PR.

marctc:
## Summary

Executable reuse was keyed only by inode, but inode numbers are only unique within a filesystem. This caused processes on different devices to be treated as the same executable.

Use device + inode as the executable identity, and only report reused processes as instrumented after attachment succeeds.

Fixes #2838 
Closes grafana/beyla#3027.

## Validation

### DIFF
--- a/bpf/gotracer/go_offsets.h
+++ b/bpf/gotracer/go_offsets.h
@@ -189,18 +189,11 @@ typedef struct go_executable_key {
 
 struct {
     __uint(type, BPF_MAP_TYPE_LRU_HASH);
-    __type(key, u64);           // key: visible inode
-    __type(value, off_table_t); // the offset table
-    __uint(max_entries, MAX_GO_PROGRAMS);
-} go_offsets_map SEC(".maps");
-
-struct {
-    __uint(type, BPF_MAP_TYPE_LRU_HASH);
-    __type(key, u32);           // key: host PID
+    __type(key, go_executable_key_t);
     __type(value, off_table_t); // the offset table
     __uint(max_entries, MAX_GO_PROGRAMS);
     __uint(pinning, OBI_PIN_INTERNAL);
-} go_offsets_by_pid SEC(".maps");
+} go_offsets_map SEC(".maps");
 
 struct {
     __uint(type, BPF_MAP_TYPE_HASH);
@@ -253,14 +246,14 @@ int obi_resolve_go_executable_identity(struct pt_regs *ctx) {
 
 static __always_inline off_table_t *get_offsets_table() {
     struct task_struct *task = (struct task_struct *)bpf_get_current_task();
-    const u32 pid = (u32)(bpf_get_current_pid_tgid() >> 32);
-    off_table_t *offsets = (off_table_t *)bpf_map_lookup_elem(&go_offsets_by_pid, &pid);
-    if (offsets) {
-        return offsets;
+
+    struct inode *inode = BPF_CORE_READ(task, mm, exe_file, f_inode);
+    go_executable_key_t key = {};
+    if (!go_executable_key(inode, &key)) {
+        return NULL;
     }
 
-    const u64 ino = (u64)BPF_CORE_READ(task, mm, exe_file, f_inode, i_ino);
-    return (off_table_t *)bpf_map_lookup_elem(&go_offsets_map, &ino);
+    return (off_table_t *)bpf_map_lookup_elem(&go_offsets_map, &key);
 }
 
 static __always_inline u64 go_offset_of(off_table_t *ot, go_offset off) {
--- a/bpf/gotracer/go_offsets.h
+++ b/bpf/gotracer/go_offsets.h
@@ -6,7 +6,9 @@
 #include <bpfcore/vmlinux.h>
 #include <bpfcore/bpf_helpers.h>
 #include <bpfcore/bpf_core_read.h>
+#include <bpfcore/bpf_tracing.h>
 
+#include <common/pin_internal.h>
 #include <gotracer/go_constants.h>
 
 // To be injected from the user space during the eBPF program load & initialization
@@ -179,15 +181,83 @@ typedef struct off_table {
     u64 table[_last_go_offset];
 } off_table_t;
 
+typedef struct go_executable_key {
+    u64 dev;
+    u64 ino;
+} go_executable_key_t;
+
 struct {
     __uint(type, BPF_MAP_TYPE_LRU_HASH);
-    __type(key, u64);           // key: inode
+    __type(key, u64);           // key: visible inode
     __type(value, off_table_t); // the offset table
     __uint(max_entries, MAX_GO_PROGRAMS);
 } go_offsets_map SEC(".maps");
 
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, u32);           // key: host PID
+    __type(value, off_table_t); // the offset table
+    __uint(max_entries, MAX_GO_PROGRAMS);
+    __uint(pinning, OBI_PIN_INTERNAL);
+} go_offsets_by_pid SEC(".maps");
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __type(key, u8);
+    __type(value, u8);
+    __uint(max_entries, 1);
+    __uint(pinning, OBI_PIN_INTERNAL);
+} go_executable_identity_requests SEC(".maps");
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __type(key, u8);
+    __type(value, go_executable_key_t);
+    __uint(max_entries, 1);
+    __uint(pinning, OBI_PIN_INTERNAL);
+} go_executable_identities SEC(".maps");
+
+static __always_inline bool go_executable_key(struct inode *inode, go_executable_key_t *key) {
+    if (!inode) {
+        return false;
+    }
+
+    key->dev = (u64)BPF_CORE_READ(inode, i_sb, s_dev);
+    key->ino = (u64)BPF_CORE_READ(inode, i_ino);
+    return key->ino != 0;
+}
+
+SEC("kprobe/uprobe_register")
+int obi_capture_go_executable_identity(struct pt_regs *ctx) {
+    const u8 request_key = 0;
+    if (!bpf_map_lookup_elem(&go_executable_identity_requests, &request_key)) {
+        return 0;
+    }
+
+    struct inode *inode = (struct inode *)PT_REGS_PARM1(ctx);
+    go_executable_key_t key = {};
+    if (go_executable_key(inode, &key)) {
+        bpf_map_update_elem(&go_executable_identities, &request_key, &key, BPF_ANY);
+        bpf_map_delete_elem(&go_executable_identity_requests, &request_key);
+    }
+
+    return 0;
+}
+
+SEC("uprobe")
+int obi_resolve_go_executable_identity(struct pt_regs *ctx) {
+    (void)ctx;
+    return 0;
+}
+
 static __always_inline off_table_t *get_offsets_table() {
     struct task_struct *task = (struct task_struct *)bpf_get_current_task();
+    const u32 pid = (u32)(bpf_get_current_pid_tgid() >> 32);
+    off_table_t *offsets = (off_table_t *)bpf_map_lookup_elem(&go_offsets_by_pid, &pid);
+    if (offsets) {
+        return offsets;
+    }
+
     const u64 ino = (u64)BPF_CORE_READ(task, mm, exe_file, f_inode, i_ino);
     return (off_table_t *)bpf_map_lookup_elem(&go_offsets_map, &ino);
 }
--- a/bpf/gotracer/go_offsets.h
+++ b/bpf/gotracer/go_offsets.h
@@ -6,8 +6,6 @@
 #include <bpfcore/vmlinux.h>
 #include <bpfcore/bpf_helpers.h>
 #include <bpfcore/bpf_core_read.h>
-#include <bpfcore/bpf_tracing.h>
-
 #include <common/pin_internal.h>
 #include <gotracer/go_constants.h>
 
@@ -195,22 +193,6 @@ struct {
     __uint(pinning, OBI_PIN_INTERNAL);
 } go_offsets_map SEC(".maps");
 
-struct {
-    __uint(type, BPF_MAP_TYPE_HASH);
-    __type(key, u8);
-    __type(value, u8);
-    __uint(max_entries, 1);
-    __uint(pinning, OBI_PIN_INTERNAL);
-} go_executable_identity_requests SEC(".maps");
-
-struct {
-    __uint(type, BPF_MAP_TYPE_HASH);
-    __type(key, u8);
-    __type(value, go_executable_key_t);
-    __uint(max_entries, 1);
-    __uint(pinning, OBI_PIN_INTERNAL);
-} go_executable_identities SEC(".maps");
-
 static __always_inline bool go_executable_key(struct inode *inode, go_executable_key_t *key) {
     if (!inode) {
         return false;
@@ -219,29 +201,6 @@ static __always_inline bool go_executable_key(struct inode *inode, go_executable
     key->dev = (u64)BPF_CORE_READ(inode, i_sb, s_dev);
     key->ino = (u64)BPF_CORE_READ(inode, i_ino);
     return key->ino != 0;
-}
-
-SEC("kprobe/uprobe_register")
-int obi_capture_go_executable_identity(struct pt_regs *ctx) {
-    const u8 request_key = 0;
-    if (!bpf_map_lookup_elem(&go_executable_identity_requests, &request_key)) {
-        return 0;
-    }
-
-    struct inode *inode = (struct inode *)PT_REGS_PARM1(ctx);
-    go_executable_key_t key = {};
-    if (go_executable_key(inode, &key)) {
-        bpf_map_update_elem(&go_executable_identities, &request_key, &key, BPF_ANY);
-        bpf_map_delete_elem(&go_executable_identity_requests, &request_key);
-    }
-
-    return 0;
-}
-
-SEC("uprobe")
-int obi_resolve_go_executable_identity(struct pt_regs *ctx) {
-    (void)ctx;
-    return 0;
 }
 
 static __always_inline off_table_t *get_offsets_table() {
--- a/bpf/gotracer/go_offsets.h
+++ b/bpf/gotracer/go_offsets.h
@@ -6,7 +6,9 @@
 #include <bpfcore/vmlinux.h>
 #include <bpfcore/bpf_helpers.h>
 #include <bpfcore/bpf_core_read.h>
+#include <bpfcore/bpf_tracing.h>
 
+#include <common/pin_internal.h>
 #include <gotracer/go_constants.h>
 
 // To be injected from the user space during the eBPF program load & initialization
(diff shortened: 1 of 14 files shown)


pkg/appolly/discover/attacher.go
```diff
@@ -154,18 +155,20 @@ func (ta *traceAttacher) getTracer(ie *ebpf.Instrumentable) bool {
...
-		ta.Metrics.InstrumentProcess(ie.FileInfo.ExecutableName())
 		if tracer.Type == ebpf.Generic {
 			// We need to do this because generic tracers have shared libraries. For example,
 			// a python executable can run an SSL and non-SSL application, so it's not enough
 			// to look at the executable, we must ensure this process doesn't have different
 			// libraries attached
 			ok = ta.updateTracerProbes(tracer, ie)
 		} else {
+			// allowing the tracer to forward traces from the new PID and its children processes
+			ta.monitorPIDs(tracer, ie)
```
mmat11: question: why was this moved in the _else_ branch?

mmat11 (approved):
lgtm!

pkg/appolly/discover/attacher.go
marctc: Good question! `monitorPIDs` its already executed in `updateTracerProbes` but only if `NewExecutableInstance` [succeds](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2848/changes#diff-a224f6ad374612d3c476bc4d7487106d8c4a27f616a675f06a14f9f793f54c4bR376-R377). So moving it out of the previous position would make `monitorPIDs` executed even if attachment to probe fails

mmat11: thanks!

lumaks-redox:
One test is failing, not sure if it's blocking the merge though.

marctc:
> One test is failing, not sure if it's blocking the merge though.

Yes it is, it's no flaky. We are investigating because the provided is breaking grpc context propagation. I believe @mmat11 has some work that extends this PR

grcevski (approved):
LGTM! But there are more of these that need to be fixed, in a follow-up. The go offsets map is also keyed off an inode number, as well as the new go channel support.

marctc:
> The go offsets map is also keyed off an inode number, as well as the new go channel support.

Can you take a second look? i address this part
cc @mmat11

lumaks-redox:
Tested on a 6.12.88 node, containerd/overlayfs, three pods of the same image (same visible inode `68323020`, three different overlay device ids). Works — all three get instrumented now. Before, only one did.

Three separate uprobe attach rounds (111 functions each) and even spans across all three pods:

```
138 "hostName":"telegraf-1"
139 "hostName":"telegraf-2"
139 "hostName":"telegraf-4"
```

Zero fallbacks to bare-inode sharing.

One thing worth knowing about what the resolver finds here: it did not resolve to the shared lower layer. Each container has its own copied-up `/usr/bin/telegraf`, so backing inodes came back distinct per pod:

```
pid A  visible_inode=68323020  backing_inode=404413538   <- upper snapshot 205
pid B  visible_inode=68323020  backing_inode=1163961     <- upper snapshot 211
```

I stat'd the layers and those match the upperdir copies exactly (full copies, ~583544 blocks, not metacopy). So these pods never actually shared a backing file and the re-key alone carries it — `go_offsets_by_pid` isn't doing the work in my run. On a setup where containers *do* share the lower-layer file the key would collapse to one and the per-pid map is what has to save it. I didn't exercise that path.

Cosmetic: `pt.log` is set in `Run()`, but `goExecutableKey` runs before that for the first process, so the `pt.log != nil` guard drops the identity log line for the very first executable. I saw 2 of 3 lines and spent a while thinking one had failed.

The single-slot request map behind the `uprobe_register` kprobe made me nervous since it fires for any `uprobe_register` on the box. This node had another eBPF agent and a security sensor attaching uprobes concurrently and it still resolved correctly every time — so it holds up, just flagging the window exists.

bpf/gotracer/go_offsets.h
```diff
@@ -179,15 +181,83 @@ typedef struct off_table {
...
+        return false;
+    }
+
+    key->dev = (u64)BPF_CORE_READ(inode, i_sb, s_dev);
+    key->ino = (u64)BPF_CORE_READ(inode, i_ino);
+    return key->ino != 0;
+}
+
+SEC("kprobe/uprobe_register")
+int obi_capture_go_executable_identity(struct pt_regs *ctx) {
```
mmat11: maybe it's unlikely to happen, but we should filter the process who registers the probe to be OBI, right?

bpf/gotracer/go_offsets.h
```diff
@@ -179,15 +181,83 @@ typedef struct off_table {
...
+    if (go_executable_key(inode, &key)) {
+        bpf_map_update_elem(&go_executable_identities, &request_key, &key, BPF_ANY);
+        bpf_map_delete_elem(&go_executable_identity_requests, &request_key);
+    }
+
+    return 0;
+}
+
+SEC("uprobe")
+int obi_resolve_go_executable_identity(struct pt_regs *ctx) {
```
grcevski: What does this empty probe do?

grcevski (commented):
I have a question, why do we need all these executable identities? I think tracking based on dev and inode should be sufficient. If we have multiple launches from the same binary and some are not instrumented, the PID filter will discard those events.

pkg/internal/ebpf/gotracer/gotracer.go
```diff
@@ -232,27 +246,142 @@ func New(
...
+	p.executableIdentitiesByFile[visibleIdentity] = identity
+	p.executableIdentityMu.Unlock()
+	return identity.Dev, identity.Ino, nil
 }
 
 func (p *Tracer) AllowPID(pid app.PID, ns uint32, fi *exec.FileInfo) {
+	if fi != nil && p.bpfObjects.GoOffsetsByPid != nil {
+		identity := p.executableIdentity(fi)
+		p.executableIdentityMu.Lock()
+		if offsets, ok := p.goOffsetsByExecutable[identity]; ok {
```
grcevski: Yeah I don't think we need this, the go offsets are per file, not pid. And we discard events from pids we don't know.

bpf/gotracer/go_offsets.h
```diff
@@ -179,15 +181,83 @@ typedef struct off_table {
...
+
 static __always_inline off_table_t *get_offsets_table() {
     struct task_struct *task = (struct task_struct *)bpf_get_current_task();
+    const u32 pid = (u32)(bpf_get_current_pid_tgid() >> 32);
+    off_table_t *offsets = (off_table_t *)bpf_map_lookup_elem(&go_offsets_by_pid, &pid);
+    if (offsets) {
+        return offsets;
+    }
+
     const u64 ino = (u64)BPF_CORE_READ(task, mm, exe_file, f_inode, i_ino);
```
grcevski: can we simply not get the device here by doing:
```
const u32 dev = BPF_CORE_READ(task, mm, exe_file, f_inode, i_sb, s_dev);
```

grcevski (commented):
I think I'm missing something in the PR description. It's unclear to me if there's an issue with the userspace not being able to see the correct device when overlay is used or was the change to use PIDs was done for performance reasons. If we wanted to use PIDs, I think we can simply remove all inode mentions at all from the BPF side and if we don't want to process the offsets for same binary mutliple times, we can simply cache them by inode/dev in userspace.

grcevski (changes requested):

bpf/gotracer/go_offsets.h
marctc: The program itself doesn’t do anything. Attaching it forces the kernel to call `uprobe_register.` The kprobe above runs at the start of that function and reads its first argument. We save its device/inode and immediately detach both probes. Do you know a cleaner way to get that identity?

marctc:
> I have a question, why do we need all these executable identities? I think tracking based on dev and inode should be sufficient. If we have multiple launches from the same binary and some are not instrumented, the PID filter will discard those events.

dev,inode is what we use, but overlayfs can give us two different pairs for the same executable: the pair visible through `/proc/<pid>/exe` and the pair used by the kernel for the uprobe. We need the second pair to avoid attaching the same system-wide probe twice.

marctc:
> I think I'm missing something in the PR description.

My bad, i forgot to update it from last changes

grcevski:
> dev,inode is what we use, but overlayfs can give us two different pairs for the same executable: the pair visible through `/proc/<pid>/exe` and the pair used by the kernel for the uprobe. We need the second pair to avoid attaching the same system-wide probe twice.

Hmmm, interesting. Do you mind if we take a look at an example of this? I'm not sure how can they be different and if they are I'm sure there will be a way to get the same id by digging around the proc file system a bit more. Do you have an easy repro I can try and see for myself? Is it the test you have added?

marctc:
After discussing this with Nikola, I removed the extra backing-identity resolver and caches because Go and BPF use the same (dev, ino) values.  I used `bpftool` on two executables with inode 2 and devices 151 and 162; each had 184 links and no duplicate (program, offset) pairs.

pkg/internal/ebpf/gotracer/gotracer.go
```diff
@@ -648,17 +666,20 @@ func (p *Tracer) RegisterOffsets(fileInfo *exec.FileInfo, offsets *goexec.Offset
...
 	ino := fileInfo.Ino()
-	if err := p.bpfObjects.GoOffsetsMap.Put(ino, offTable); err != nil {
-		p.log.Error("setting Go offsets map failed", "pid", fileInfo.Pid(), "ino", ino, "error", err)
+	identity := goOffsetsMapKey(fileInfo)
+	if err := p.bpfObjects.GoOffsetsMap.Put(identity, offTable); err != nil {
+		p.log.Error("setting Go offsets map failed",
+			"pid", fileInfo.Pid(),
+			"ino", ino,
+			"error", err)
 		delete(p.goAutoSDKActivationByIno, ino)
```
grcevski: I think we still need to fix goAutoSDKActivationByIno, right?

grcevski (commented):
This looks pretty good now and the code is much simpler, I left one comment I think we should address, because goAutoSDKActivationByIno looks like is having the same problem. 

I'm approving with the caveat that we either apply the same fix for goAutoSDKActivationByIno or we follow-up with another PR.

grcevski (approved):

mmat11 (approved):

mmat11:
grpc tests are failing again 😞 , maybe let's try to rerun CI after rebasing from main -- are you sure the links for the same image were not duplicated?

marctc:
> grpc tests are failing again 😞 , maybe let's try to rerun CI after rebasing from main -- are you sure the links for the same image were not duplicated?

I tested this on top of current main, and you were right: the links for the same image were duplicated.

The four Go relay containers use the same image. Before the fix, `bpftool` showed the same probe attached four times at the same offset:

```
  6107: perf_event  prog 5689
      uprobe /proc/1010596/exe+0x4f6e60
  6205: perf_event  prog 5689
      uprobe /proc/1010024/exe+0x4f6e60
  6341: perf_event  prog 5689
      uprobe /proc/1010504/exe+0x4f6e60
  6533: perf_event  prog 5689
      uprobe /proc/1010774/exe+0x4f6e60
```

The return probe also had 24 links: six return offsets attached four times.

The fix keeps `{dev, ino}` for offset lookup, but uses the kernel uprobe target to decide whether Go links can be shared. After the fix, `bpftool` shows one entry link:

```
  6739: perf_event  prog 5890
      uprobe /proc/1029707/exe+0x4f6e60
```

The return probe now has six links, one for each return offset. This keeps separate executables independent while attaching the Go probe set only once for containers that resolve to the same kernel target.

grcevski:
Nice!!! Let's wait for green CI and we should be good.

pkg/ebpf/tracer_linux.go
```diff
@@ -374,7 +375,30 @@ func (pt *ProcessTracer) NewExecutable(exe *link.Executable, ie *Instrumentable)
 
 	for _, p := range pt.Programs {
 		p.RegisterOffsets(ie.FileInfo, ie.Offsets)
+	}
+
+	if uprobeKey, ok := pt.resolveUprobeTarget(exe, ie.Offsets); ok {
```
grcevski, line 380: Can you please explain this code? What actually fixed things, it looks like you are doing some sort of de-duplication here.

So what I'm trying to understand is that the original customer issue was saying just one of the 10 programs was intrumented, but now we are trying to work around duplicate instrumentations when it obviously worked before.

Does your integration test fail with main?

mmat11:
lgtm, some unit tests are failing tho

pkg/ebpf/tracer_linux.go
marctc: Yeah, so in #2838 user reported that different Go executables had the same inode number on different filesystems. While this PR was open, #2820 added device and inode keys. The reporter tested main again, but the problem remained.

During review, we checked the links for this new test, but not for the failing gRPC test. When I checked gRPC, I found four copies of every Go probe attached to the same `/relay` file. The four containers look like separate executables, but the kernel registers their uprobes against the same underlying file. Each gRPC call therefore ran the same BPF program four times, which broke context propagation.

The code here fixes that second problem. It creates a temporary probe so the `uprobe_register` kprobe can report the actual file selected by the kernel. If that file already has Go probes, we reuse those links. We still register offsets and process state for every executable. If the kernel selects a different file, we attach a separate set of Go probes.

grcevski: OK, so I think I understand but I think your integration test is maybe not testing the scenario that the user reported. Essentially, if I understand correctly, you are saying with overlayfs we can have:

```
container1:
  dev: 1, inode: 2, exe: a

container2:
  dev: 2, inode: 2, exe: b 

container3:
  dev: 3, inode: 2, exe: a
```

So `container 1` and `container 3` share the same underlying binary, so we instrument them twice. How is your new test different than the gRPC test that failed without deduplication?

grcevski:
I think I need to see the reported dev and inode numbers for the gRPC scenario vs the new test. I think the new test use of tmpfs is forcing a different dev, but what happens in the gRPC test?

pkg/ebpf/tracer_linux.go
marctc: So yes, the new integration test is testing the scenario that the user reported, because the problem depends on two separate executable files having the same inode number on different devices, not on overlayfs itself. 

We use tmpfs to reproduce that identity collision, while the gRPC test covers the opposite case where several visible executable identities resolve to the same kernel uprobe target.

marctc:
> I think I need to see the reported dev and inode numbers for the gRPC scenario vs the new test. I think the new test use of tmpfs is forcing a different dev, but what happens in the gRPC test?

New test:

```
  obi-1 | msg="resolved Go executable identity" pid=2420864 visible_dev=156 visible_ino=2 uprobe_target_dev=156 uprobe_target_ino=2
  obi-1 | msg="resolved Go executable identity" pid=2420879 visible_dev=163 visible_ino=2 uprobe_target_dev=163 uprobe_target_ino=2
```

  gRPC test:
```
  obi-1 | msg="resolved Go executable identity" pid=2423917 visible_dev=209 visible_ino=56360999 uprobe_target_dev=264241153 uprobe_target_ino=56360999
  obi-1 | msg="resolved Go executable identity" pid=2423649 visible_dev=170 visible_ino=56360999 uprobe_target_dev=264241153 uprobe_target_ino=56360999
  obi-1 | msg="resolved Go executable identity" pid=2423171 visible_dev=126 visible_ino=56360999 uprobe_target_dev=264241153 uprobe_target_ino=56360999
  obi-1 | msg="resolved Go executable identity" pid=2423748 visible_dev=183 visible_ino=56360999 uprobe_target_dev=264241153 uprobe_target_ino=56360999
```

grcevski (approved):
LGTM!

I did some looking around and I think this is the best way to solve this problem. I think the issue is the xino translation with containers, whcih picks new device number for all overlaid file systems. I even dug into the overlayfs in Docker and on my host I still see the different device ids. Maybe there's another way of telling that they link to the same file, but at this point this  works well.

PR #2851, opened 2026-07-30
Deno instrumentable type

mariomac:
## Summary

For Deno, we have been reusing `nodejs` instrumentable type but in order to provide full support (followup PRs), we need to make some differences depending on the executable type.

## Validation


pkg/appolly/app/svc/svc.go
```diff
@@ -48,6 +49,11 @@ func (it InstrumentableType) String() string {
 		return semconv.TelemetrySDKLanguageRuby.Value.AsString()
 	case InstrumentableNodejs:
 		return semconv.TelemetrySDKLanguageNodejs.Value.AsString()
+	case InstrumentableDeno:
+		// There is no stable OTEL convention for Deno runtime.
+		// However, Deno sets "deno-rust" in its own native OTEL instrumentation
+		// https://docs.deno.com/runtime/fundamentals/open_telemetry/
+		return "deno-rust"
```
MrAlias, line 56: `deno-rust` is intentional, but `schemas/obi/groups/telemetry.yaml` only extends `telemetry.sdk.language` with `generic`. Weaver flags every Deno sample as `undefined_enum_variant`, which is why `TestSuite_Deno` fails on all three runs.

Could we add `deno-rust` to that override and update its description? The existing `generic` member shows the repository pattern for custom language values.

MrAlias (commented):
The Deno type split is consistent across its current consumers, but the new language value is missing from OBI's semantic-convention registry.

pkg/appolly/app/svc/svc.go
mariomac: Ahh! Thanks for pointing it!

NimrodAvni78 (approved):
maybe i would reference the deno otel docs to mention we didn't invent this convention but follow denos official sdk

pkg/appolly/app/svc/svc.go
MrAlias, line 56: `TestSuite_Deno` reuses `testHTTPTracesKProbes`, but that helper still expects `telemetry.sdk.language=nodejs`. Its `jaeger.Diff` result is discarded and the following assertion rechecks the earlier span diff, so the suite passes without verifying `deno-rust`.

Could we parameterize the expected language and assert the process-tag diff? Weaver accepts both `nodejs` and `deno-rust`, so it cannot catch this regression.

MrAlias (commented):
The type split and schema update are consistent at the current head, but the Deno integration coverage still has a false-positive path.

mmat11:
dumb question: does this work overlap with @MrAlias 's work at https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2957?

MrAlias:
Yeah, not dumb, [it did indeed overlap](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2957#issuecomment-5220313114) 😄:

> To clarify the overlap with #2851: I reviewed that PR, and the shared Deno type split is intentional. This PR uses the same seam for a narrower purpose—keeping Deno out of the Node injector, whose SIGUSR1 inspector path can terminate a Deno process—while preserving the current telemetry language, generic tracing, route harvesting, and configuration behavior.
> 
> @mariomac is away for the next week or two, so I opened this focused fix rather than hold the safety issue behind the broader Deno work. The PRs should still converge: whichever lands second can rebase and retain only its distinct behavior and tests.

Just waiting for @mariomac to return and we can rebase this to include the parts that did not overlap.

bpf/generictracer/protocol_kafka.h
```diff
@@ -73,10 +73,10 @@ enum {
...
     k_kafka_hdr_request_api_version = 2,
     k_kafka_hdr_correlation_id = 4,
-    k_kafka_request_header_fields_without_message_size =
-        k_kafka_hdr_request_api_key + k_kafka_hdr_request_api_version + k_kafka_hdr_correlation_id,
-    k_kafka_min_request_header_size =
-        k_kafka_hdr_message_size + k_kafka_request_header_fields_without_message_size,
+    k_kafka_request_header_fields_without_message_size = k_kafka_hdr_request_api_key +
+        k_kafka_hdr_request_api_version + k_kafka_hdr_correlation_id,
+    k_kafka_min_request_header_size = k_kafka_hdr_message_size +
+        k_kafka_request_header_fields_without_message_size,
```
mariomac: Unrelated to this PR but I had to rerun clang-format so PR checks passed

pkg/appolly/app/svc/svc.go
mariomac: Added!

mmat11 (approved):

PR #2852, opened 2026-07-30
Some improvements for working in Git worktrees

mariomac:
## Summary

When you spin up a new worktree, you will usually get two kinds of failures:
- `testoutput` is not created and some basic checks will fail unless you manually create it.
- Linter requires removing the lint cache (seems a bug in Golangci-lint)

This PR makes sure `testoutput` folder is created before Makefile targets that make use of it, and adds a new target to cleanup the linter cache.

## Validation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -430,8 +430,15 @@ generator-image-build:
 	@echo "### Creating the image that generates the eBPF binaries"
 	$(OCI_BIN) buildx build --load -t $(GEN_IMG) -f generator.Dockerfile  .
 
+
+.PHONY: prepare-integration-test
+prepare-integration-test:
+	@echo "### Removing resources from previous integration tests, if any"
+	rm -rf $(TEST_OUTPUT)/* || true
+	$(MAKE) cleanup-integration-test
+
 .PHONY: cleanup-integration-test
-cleanup-integration-test: clean-testoutput
+cleanup-integration-test:
 	@echo "### Removing integration test clusters"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -423,15 +430,8 @@ generator-image-build:
...
-
-.PHONY: prepare-integration-test
-prepare-integration-test:
-	@echo "### Removing resources from previous integration tests, if any"
-	rm -rf $(TEST_OUTPUT)/* || true
-	$(MAKE) cleanup-integration-test
-
 .PHONY: cleanup-integration-test
-cleanup-integration-test:
+cleanup-integration-test: clean-testoutput
```
MrAlias: `integration-test`, `integration-test-k8s`, and `integration-test-arm` each run `itest-coverage-data` and then call `cleanup-integration-test`. Since this target now depends on the phony `clean-testoutput` target, that final cleanup deletes `itest-covdata.txt` and the test logs that the command just produced.

Could we keep `clean-testoutput` in the pre-test path and leave `cleanup-integration-test` limited to external resources?

MrAlias (commented):
The worktree setup changes are contained, but the integration cleanup path currently discards the artifacts produced by the documented developer targets.

Makefile
mariomac: Oh yeah, good point.

marctc (approved):
Nice, very much needed!

mmat11 (approved):

NimrodAvni78 (approved):

Issue #2849, opened 2026-07-30
Unified ARM CI


skl:
PR #2825 fixed a bug which could've been caught earlier if we had greater coverage for ARM. Historically, before native ARM runners were available, QEMU was used to test the ARM platform.

As QEMU tests are generally much slower, ARM-specific test suites were introduced into OBI only for specific cases and subsets of the full test suite we currently run under x86. This has led to an increasing gap in ARM coverage.

This issue aims to track the work being done to unify the ARM and x86 test suites, with the high-level goals of:

- [x] Adding `ubuntu-24.04-arm` to existing x86 CI workflows that already have an ARM equivalent, increasing ARM coverage gradually by no longer target TestMultiProcess suites only.
- [x] Removing the ARM-specific CI workflows as the above takes precedent.
- [ ] Adding `ubuntu-24.04-arm` to existing x86 CI workflows that do not have an ARM equivalent, further increasing coverage.
- [ ] Adding `ubuntu-26.04-arm` to the newly unified CI workflows, to also (or instead?) cover ARM with kernel version 7.
- [ ] Increase ARM kernel coverage on BPF Verifier workflow (currently host-kernel only)

ARM runners will not be used for every workflow, only for those where ARM behaviour might be different.

## Current state

| Workflow | ARM today | Action |
|---|---|---|
| [bpf_unit_tests.yml](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/.github/workflows/bpf_unit_tests.yml) | none | matrix: #3080 |
| [bpf_verifier.yml](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/.github/workflows/bpf_verifier.yml) | `host-arm64` job (host kernel only) + `generate-bpf-arm` | more ARM kernels: #3318 |
| [generator-image.yml](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/.github/workflows/generator-image.yml), [ghcr_publish_testing_images.yml](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/.github/workflows/ghcr_publish_testing_images.yml) | multi-arch via QEMU buildx (fine for image builds) | none |
| [lint.yml](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/.github/workflows/lint.yml) | none | matrix: #3302 |
[...]

PR #2853, opened 2026-07-30
ci: run unit tests on arm

skl:
## Summary

Part of:

- #2849 

This PR adds ARM to the list of platforms that the unit tests run on, so that they will now run on both x86 and ARM.

Also remove redundant `generate-bpf-arm` step as the existing `generate-bpf` creates both x86 and ARM artifacts already.

## Validation

Unit tests running on both x86 and arm:

https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/30557663403?pr=2853

Screenshot 2026-07-30 at 16 45 25

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -66,16 +66,22 @@ jobs:
           make unit-test-matrix-json >> "$GITHUB_OUTPUT"
 
   test:
-    name: "Unit tests #${{ matrix.id }}"
+    name: "Unit tests #${{ matrix.shard.id }} (${{ matrix.platform.arch }})"
     needs: [matrix, generate-bpf]
     permissions:
       checks: write
       pull-requests: write
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.platform.runner }}
     timeout-minutes: 45
     strategy:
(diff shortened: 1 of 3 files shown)


mariomac (approved):
looks great!

.github/workflows/unit_tests.yml
```diff
@@ -125,4 +131,4 @@ jobs:
         with:
           files: ./testoutput/cover.txt
           flags: unittests
```
pinoOgni, line 133: nit: does it make sense? 
`flags: unittests,${{ matrix.platform.arch }}`?

pinoOgni (approved):
🚀

MrAlias (approved):

.github/workflows/unit_tests.yml
skl: Do we have different coverage per arch? I couldn't find a case of this.

pinoOgni: Mmmh, I found only `pkg/internal/goexec/instructions_amd64.go` and `pkg/internal/goexec/instructions_arm64.go`, so I guess that `flags: unittests` is ok

PR #2854, opened 2026-07-31
fix(ebpf): page-align UserRingbuf max_entries like RingBuf

### LINKED ISSUE #2808: alignMaxEntriesIfRingBuf does not cover UserRingbuf
`alignMaxEntriesIfRingBuf` in `pkg/internal/ebpf/convenience/convenience.go` only checks `ebpf.RingBuf`:

```go
func alignMaxEntriesIfRingBuf(m *ebpf.MapSpec) {
	if m.Type == ebpf.RingBuf {
		m.MaxEntries = roundToNearestMultiple(m.MaxEntries, uint32(os.Getpagesize()))
	}
}
```

`ebpf.UserRingbuf` (`BPF_MAP_TYPE_USER_RINGBUF`) has the same constraint — both types go through `ringbuf_map_alloc` in the kernel, which requires `max_entries` to be a power of 2 and page-aligned. A `UserRingbuf` map would not get the alignment fixup, so scaling one down via a negative `global_scale_factor` could produce an invalid size and fail at load.

This is not reachable today: OBI defines no `UserRingbuf` maps. Filing it so it isn't missed if one is added later.

The fix is presumably just extending the condition to cover both types.

bourbonkk:
## Summary

`alignMaxEntriesIfRingBuf` only covers `ebpf.RingBuf`, but `UserRingbuf` is allocated
through the same kernel path (`ringbuf_map_alloc`), which requires `max_entries` to be
page-aligned. A `UserRingbuf` map scaled down with a negative `global_scale_factor`
would skip the fixup and fail at load.

Not reachable today — OBI defines no `UserRingbuf` maps — so this is a preventive fix
so the case isn't missed if one is added later. Extends the condition to both types and
adds a test covering the alignment for both, plus that other map types are left untouched.

Fixes #2808

## Validation

- [ ] N/A — internal map-sizing fix with no feature support status change, so the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) are unchanged.

---

AI policy: I used Claude Code to draft the change and this description. I reviewed
the code and ran the tests myself.

### DIFF
--- a/pkg/internal/ebpf/convenience/convenience.go
+++ b/pkg/internal/ebpf/convenience/convenience.go
@@ -32,9 +32,9 @@ func roundToNearestMultiple(x, n uint32) uint32 {
 	return (x + n/2) / n * n
 }
 
-// RingBuf map types must be a multiple of os.Getpagesize()
+// RingBuf and UserRingbuf max_entries must be page-aligned: both go through the kernel's ringbuf_map_alloc
 func alignMaxEntriesIfRingBuf(m *ebpf.MapSpec) {
-	if m.Type == ebpf.RingBuf {
+	if m.Type == ebpf.RingBuf || m.Type == ebpf.UserRingbuf {
 		m.MaxEntries = roundToNearestMultiple(m.MaxEntries, uint32(os.Getpagesize()))
 	}
 }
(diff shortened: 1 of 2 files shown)


mmat11 (approved):
we likely won't need this, but it adds negligible complexity, I'm fine merging it

PR #2857, opened 2026-07-31
Improve performance of BPF metrics

### LINKED ISSUE #727: Investigate performance impacts of collecting BPF metrics
The following CPU and memory profiles show that collecting eBPF metrics take a non-negligible amount of time in one of our clusters.

Image

Image

This needs to be further investigated. There are two obvious starting points:

 - the Prometheus reporter
 - `newProgramInfoFromFd` https://github.com/cilium/ebpf/blob/7c861fdc272cd85f23e01805c20512ce2f68219b/info.go#L350

I've noticed elsewhere that creating labels is not cheap - perhaps there's a way to cache that.

With regards to `newProgramInfoFromFd` - collecting BPF metrics involves creating a program ID iterator - which in itself involves inumerous syscalls per program, and on the cilium/ebpf side, some overhead into building the higher-level objects (including I/O).

The obvious workaround is to increase the polling time - but that is a _workaround_. We need to understand what exactly makes it take so long, and then what can be done / optimised.

marctc:
## Summary

Describe the change briefly.

Cache eBPF program and map metadata across metric collections while continuing to discover all existing supported probes and maps on every collection.

| Benchmark | Time/op | Memory/op | Allocations/op |
| --- | ---: | ---: | ---: |
| Probe metrics | 8.1497 ms → 0.9832 ms (**87.94% lower**) | 3,492.9 KiB → 178.3 KiB (**94.90% lower**) | 17,523 → 1,008 (**94.25% lower**) |
| Map metrics | 12.944 ms → 7.586 ms (**41.40% lower**) | 2,908.1 KiB → 295.5 KiB (**89.84% lower**) | 23,900 → 14,010 (**41.40% lower**) |

  Results are benchstat medians from 10 runs with 10 iterations per run on Linux/amd64. All changes were statistically significant (p < 0.001).

Fixes #727 

## Validation

### DIFF
--- a/pkg/export/prom/prom_bpf.go
+++ b/pkg/export/prom/prom_bpf.go
@@ -39,12 +39,13 @@ type BPFCollector struct {
 	mapCache         map[ebpf.MapID]*cachedMap
 	probeMetrics     func() []ProbeMetrics
 	mapMetrics       func() []BpfMapMetrics
+	closeBPFStats    func()
 	mu               sync.Mutex
 	closed           bool
 }
 
 type cachedProgram struct {
-	program   *ebpf.Program
+	supported bool
 	probeType string
 	probeName string
 	probeID   string
@@ -88,7 +89,7 @@ func BPFMetrics(
 	cfg *PrometheusConfig,
 	mpCfg *perapp.MetricsConfig,
 ) swarm.InstanceFunc {
-	return func(_ context.Context) (swarm.RunFunc, error) {
+	return func(ctx context.Context) (swarm.RunFunc, error) {
 		promEnabled := promMetricsEnabled(cfg, mpCfg)
 		internalEnabled := internalMetricsEnabled(ctxInfo.Metrics)
 		if !promEnabled && !internalEnabled {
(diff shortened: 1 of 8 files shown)


pkg/export/prom/prom_bpf_linux.go
```diff
@@ -8,9 +8,18 @@ import (
...
+
+func (bc *BPFCollector) enableBPFStatsRuntime() func() {
+	stats, err := enableStats(unix.BPF_STATS_RUN_TIME)
 	if err != nil {
 		bc.log.Error("failed to enable runtime stats", "error", err)
+		return func() {}
+	}
+
+	return func() {
+		if err := stats.Close(); err != nil {
```
MrAlias, line 21: Closing this FD when `getProbeMetrics` returns stops kernel runtime-stat collection; `cilium/ebpf` documents that accounting runs only while the returned closer remains open. Since collections are separated by the scrape interval, executions during almost the entire interval are never counted, so both execution and latency deltas are severely underreported. Could we enable stats once for the collector lifetime and close it from `close()` instead?

pkg/export/prom/prom_bpf.go
```diff
@@ -305,17 +354,81 @@ func (bc *BPFCollector) getProbeMetrics() []ProbeMetrics {
...
+
+	if !supportedProgramType(info.Type) {
+		bc.closeProgram(id, program)
+		cached := &cachedProgram{}
+		bc.programCache[id] = cached
+		return cached
+	}
+
+	cached := &cachedProgram{
+		program:   program,
```
MrAlias: Keeping this program FD in the cache takes an ownership reference on the kernel program. After its real owner unloads it, the collector keeps it alive, `ProgramGetNextID` continues to see it, and `evictMissingPrograms` can never remove it. That leaves stale programs in the metrics and can pin every supported program observed until collector shutdown. Could we cache only metadata and use a temporary handle for each stats read, or use another scheme that does not extend the program lifetime?

MrAlias (commented):
The cache currently changes kernel statistics and program lifetimes, so the performance gain comes with telemetry correctness and host-lifecycle regressions. Those need to be addressed before merge.

MrAlias (approved):

PR #2858, opened 2026-07-31
NodeJS: simplify script injection

mariomac:
## Summary

Before: converting `_extractorBytes []byte` into a string and then wrap it inside a `((() => { ... })()` clause, before injecting it via `Runtime.eval`.

Now: `_extractorCode string` is a self-contained string and we don't need to prepend/append any string before injecting it.

Reason: A followup Deno extractor code needs to be wrapped differently (into an `async` lambda), so it's simpler keeping the `fdextractor.js` code self-contained so the Go code just injects it without prior processing (and doesn't need to make any difference when we later add a future, async `fdextractor_deno.js`).

## Validation

### DIFF
--- a/pkg/internal/nodejs/fdextractor.js
+++ b/pkg/internal/nodejs/fdextractor.js
@@ -1,120 +1,122 @@
-const STORE = Symbol.for('otel-ebpf-instrumentation.fdextractor');
+(()=>{
+  const STORE = Symbol.for('otel-ebpf-instrumentation.fdextractor');
 
-const net = require('net');
-const fs = require('fs');
+  const net = require('net');
+  const fs = require('fs');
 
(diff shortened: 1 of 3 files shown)


pkg/internal/nodejs/fdextractor.js
```diff
@@ -1,120 +1,122 @@
-const STORE = Symbol.for('otel-ebpf-instrumentation.fdextractor');
-
-const net = require('net');
-const fs = require('fs');
+(()=>{
```
mariomac, line 1: Despite the big Diff, this PR only prepends `(()=>{` and appends `})()` at the end of the file. The code in-between has just been indented 2 spaces.

NimrodAvni78 (approved):

PR #2861, opened 2026-07-31
Dynamic Client: Support isolated netns non-k8s processes for network+stats metrics

damemi:
## Summary

This adds support for non-k8s network and stats metrics via the DynamicSelector. The DynamicSelector currently uses a lookup to map selected PIDs to a matching IP to decide which apps to "include" (export). Today this relies on a k8s store lookup, which won't work in VM/Docker-standalone apps.

This adds an additional fallback into the mapping to look up host network addresses that map to the selected PIDs and uses that if the k8s store is nil.

Based on feedback from the original attempts, this PR limits the scope to host processes that exist in isolated network namespaces, such as docker containers. Bare host PID support will come in a follow up which requires pid-level association for the host network namespace (meaning this PR also doesn't support containers running with host network)

Something else I noticed which may be a follow up is it doesn't look like we export attributes like `service.name` in non-k8s metrics? This would be good to update if possible

## Validation

### DIFF
--- a/pkg/internal/helpers/container/ips.go
+++ b/pkg/internal/helpers/container/ips.go
@@ -1,0 +1,61 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package container // import "go.opentelemetry.io/obi/pkg/internal/helpers/container"
+
+import (
+	"fmt"
+	"net"
+
+	"go.opentelemetry.io/obi/pkg/appolly/app"
+	"go.opentelemetry.io/obi/pkg/internal/netns"
+)
+
+// injectable for testing
+var (
+	withNetNS      = netns.WithNetNS
+	interfaceAddrs = net.InterfaceAddrs
+)
+
+// IPsForPID returns the non-loopback IPv4/IPv6 addresses visible in pid's network
+// namespace. This covers Docker/containerd bridge networking and host-network /
+// bare-metal processes (where the process shares the agent's netns) without
+// requiring a Kubernetes metadata store.
+func IPsForPID(pid app.PID) ([]string, error) {
+	var ips []string
+	err := withNetNS(int(pid), func() error {
+		addrs, err := interfaceAddrs()
+		if err != nil {
+			return fmt.Errorf("listing interface addresses: %w", err)
+		}
+		ips = uniqueNonLoopbackIPs(addrs)
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	return ips, nil
+}
+
+func uniqueNonLoopbackIPs(addrs []net.Addr) []string {
+	seen := make(map[string]struct{}, len(addrs))
+	var out []string
+	for _, addr := range addrs {
+		ipnet, ok := addr.(*net.IPNet)
+		if !ok || ipnet == nil || ipnet.IP == nil || ipnet.IP.IsLoopback() {
+			continue
+		}
+		// Prefer the canonical string form (IPv4 as dotted quad, not :ffff:).
+		ip := ipnet.IP
+		if v4 := ip.To4(); v4 != nil {
+			ip = v4
+		}
+		s := ip.String()
+		if _, dup := seen[s]; dup {
+			continue
+		}
+		seen[s] = struct{}{}
+		out = append(out, s)
+	}
+	return out
+}
--- a/pkg/internal/helpers/container/ips.go
+++ b/pkg/internal/helpers/container/ips.go
@@ -13,17 +13,27 @@ import (
 
 // injectable for testing
 var (
-	withNetNS      = netns.WithNetNS
-	interfaceAddrs = net.InterfaceAddrs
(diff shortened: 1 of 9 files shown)


pkg/internal/helpers/container/ips.go
```diff
@@ -0,0 +1,61 @@
...
+// bare-metal processes (where the process shares the agent's netns) without
+// requiring a Kubernetes metadata store.
+func IPsForPID(pid app.PID) ([]string, error) {
+	var ips []string
+	err := withNetNS(int(pid), func() error {
+		addrs, err := interfaceAddrs()
+		if err != nil {
+			return fmt.Errorf("listing interface addresses: %w", err)
+		}
+		ips = uniqueNonLoopbackIPs(addrs)
```
MrAlias, line 41: For a bare or host-network PID, this returns every non-loopback address in the shared node namespace. `DynamicAppIPs.Allows` then admits any NetO11y/StatsO11y record involving a node address, including traffic from unselected processes, while `DynamicFlowAttrs.rebuild` assigns those shared addresses to the last selected service. Because `CommonAttrs` carries no PID or cgroup identity, nothing downstream can recover ownership. Could we limit this fallback to isolated network namespaces, or add a PID-aware association before claiming bare-host support? Please cover one selected PID sharing a namespace with unselected traffic and multiple selected service identities.

MrAlias (commented):
The namespace fallback is reasonable for isolated containers, but the advertised VM/bare-process path uses shared host IPs as process identity and therefore admits and misattributes unrelated host traffic.

pkg/internal/helpers/container/ips.go
damemi: Sorry for the delay (I was out last week and haven't had a chance to look back at this)

That is a good point and I didn't realize it in my mvp testing of this. Being able to have bare host support would be a big plus for us (we have bare VM users that would like network metrics for specific apps) so adding some PID-aware association seems like the only way to really make that work (can't guarantee their apps will be in an isolated netns I think?). Do you think that's reasonable?

On the specific-app side, a follow up I was hoping to do after this is being able to include `service.name` in non-k8s network metrics as well. Since I didn't have that before, I think that's why I didn't notice this was picking up "everything"

MrAlias: Yes, PID-aware association is the right direction, but it needs to happen before records reach `CommonAttrs`. NetO11y currently aggregates by addresses, ports, protocol, and interface, while StatsO11y also discards process ownership when constructing `CommonAttrs`. At that point, an IP-to-service map cannot distinguish processes sharing the host namespace. Reading the current PID from the TC packet hook would not be reliable either, especially for ingress traffic.

For bare-host support, we need to associate each socket or connection with a selected PID—or a stable cgroup identity—at a process-aware hook, carry that ownership and the local endpoint through NetO11y aggregation and StatsO11y events, and use it for both filtering and `service.name` decoration. The existing IP-based path can remain for Kubernetes and isolated network namespaces, but shared namespace addresses should not be treated as ownership.

If that end-to-end change is too large for this PR, I would scope this PR to isolated network namespaces and remove the bare/host-network claim, then add bare-host support separately. For validation, I think the important cases are a selected process sharing a host IP with unrelated traffic, and multiple selected processes with different service identities, across both NetO11y and StatsO11y.

damemi:
@MrAlias I started working on the full bare-host support we discussed in the sig meetings, and while it does seem very doable it is admittedly a bit to bite off. So I took your advice above and scoped this PR down to just isolated network namespace support (ie, docker containers on a VM) which I think is still a step forward.

That keeps this scope small, and I'm sure it would be good to at least wrap up this PR, so if you'd like to take another look I'll work on part 2 with the full bare process support in a follow up after this. Thanks!

pkg/internal/netns/netns.go
```diff
@@ -21,6 +21,41 @@ func netNSPath(hostPid int) string {
...
+	}
+
+	if pidIno == selfIno {
+		return false, nil
+	}
+
+	hostIno, err := netNSInode(netNSPath(1))
+	if err != nil {
+		// Host PID 1 is not visible; agent vs pid is the only check.
+		return true, nil
```
MrAlias: `IsIsolated` returns true when `/proc/1/ns/net` cannot be inspected, so it has not established that the target differs from the host namespace. If OBI runs in a separate netns and can inspect a selected host process but not PID 1, `IPsForPID` enumerates the host addresses and recreates the unrelated-traffic admission this change is meant to prevent. This should fail closed.

```suggestion
		return false, fmt.Errorf("stat host netns: %w", err)
```

MrAlias (commented):
The isolated-netns scope is close, but one error path still bypasses the host-network exclusion.

pkg/internal/netns/netns.go
damemi: @MrAlias thanks, updated to fail on this

MrAlias (approved):

Issue #1193, opened 2026-01-27
Switch the in_queue and processing to Span events


grcevski:
The internal spans of "in_queue" and "processing" can be verbose sometimes, perhaps span events are better choice for this information.

mlhv:
Hi!

Is this issue still open to work on? I'm hoping to take a look at it.

I was looking through the repo and wanted to learn more about eBPF (I have already contributed a little bit to the opentelemetry-go-contrib repos before and was looking to do the same here as well.)

Thank you!

@NimrodAvni78

marctc:
@mlhv i assigned the issue to you, thanks!

mlhv:
Opened PR for this https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2863.

NimrodAvni78:
@grcevski closing this after the PR as closed, span events are deprecated and the log alternative seems to not match what we want to do

PR #2863, opened 2026-07-31
feat(logs): replace queue/processing spans with a combined log record (opt-in)

mlhv:
## Summary

Every request with an observable queue gap produces two extra full sibling spans, `"in queue"` and `"processing"`, alongside the parent — tripling OTLP span volume for that request. The originating issue proposed span events as a lighter-weight fix, but that API has since been deprecated in favor of the Logs API, so this replaces the sub-spans with a single combined log record instead.

- `pkg/export/otel/tracesgen/tracesgen.go`, `traces.go`: thread a `suppressQueueProcessingSpans` toggle through span generation
- `pkg/export/otel/otelcfg/config_logs.go`, `logsgen/`, `logs.go`: new `otel_logs_export` signal, mirroring the existing `TracesConfig` pattern
- `pkg/obi/config.go`, `pkg/appolly/instrumenter.go`: wire in `Config.QueueProcessingAsLogsEnabled()` and the `OTELLogsReceiver` node
- `internal/test/integration/logs_test.go`: end-to-end toggle-on verification

The toggle defaults off (no behavior change for existing users), and gates on both the `queue_processing_logs` flag _and_ a configured logs endpoint (`Config.QueueProcessingAsLogsEnabled`) — a half-configured toggle keeps the old sub-spans flowing instead of silently dropping the queue/processing timing data. The log record correlates to its parent span via the record's native `trace_id/span_id`, carrying `queue.duration/processing.duration` as attributes.

Refs #1193 

## AI usage

Per AI-Policy §5: Claude Code was used to implement this feature end-to-end (new logs export signal, span-suppression toggle, wiring, and tests) and to run the verification below. Review caught and fixed three correctness issues before this was opened: a logs-receiver enable gate that only checked the common endpoint and ignored the `queue_processing_logs` toggle, a per-service export check using the traces permission instead of the logs one, and a log record that could carry an unmatchable `span_id` on the suppressed-span fallback path. I reviewed the resulting diff before submitting.

Verification run:
- `go test ./pkg/export/otel/... ./pkg/obi/... ./pkg/appolly/...` clean
- Integration suite: toggle-on (`TestSuite_QueueProcessingLogs`) confirms 0 sub-spans + 1 log record; toggle-off regression re-run confirms the existing 3-span behavior is unchanged
- `make lint / make build` clean

## Validation

- [ ] features doc / support matrix - not applicable: both documents cover protocol/library instrumentation matrices, not OTLP export signal config. The new `otel_logs_export` config is documented in the autogenerated `devdocs/config/CONFIG.md`.

### DIFF
--- a/pkg/export/otel/logs.go
+++ b/pkg/export/otel/logs.go
@@ -187,30 +187,11 @@ func getLogsExporter(ctx context.Context, cfg otelcfg.LogsConfig) (exporter.Logs
 	switch proto := cfg.GetProtocol(); proto {
 	case otelcfg.ProtocolHTTPJSON, otelcfg.ProtocolHTTPProtobuf, "":
 		slog.Debug("instantiating HTTP LogsReporter", "protocol", proto)
-		opts, err := otelcfg.HTTPLogsEndpointOptions(&cfg)
+		factory := otlphttpexporter.NewFactory()
+		config, host, queueCfg, retryCfg, err := buildHTTPLogsExporterConfig(cfg, factory)
 		if err != nil {
 			slog.Error("can't get HTTP logs endpoint options", "error", err)
 			return nil, nil, err
-		}
-		factory := otlphttpexporter.NewFactory()
-		config := factory.CreateDefaultConfig().(*otlphttpexporter.Config)
-		config.QueueConfig = getLogsQueueConfig(cfg)
-		config.RetryConfig = getLogsRetrySettings(cfg)
-		config.ClientConfig = confighttp.ClientConfig{
-			Endpoint: opts.Scheme + "://" + opts.Endpoint + opts.BaseURLPath,
-			TLS: configtls.ClientConfig{
-				Insecure:           opts.Insecure,
-				InsecureSkipVerify: cfg.InsecureSkipVerify,
-			},
-			Headers: convertHeaders(opts.Headers),
-		}
-		host := component.Host(emptyHost{})
-		if opts.UnixSocketAddr != "" {
-			mwID := component.MustNewID("obi_uds")
-			config.ClientConfig.Middlewares = []configmiddleware.Config{{ID: mwID}}
-			host = udsHost{extensions: map[component.ID]component.Component{
-				mwID: udsMiddleware{addr: opts.UnixSocketAddr},
-			}}
 		}
 		set := getTraceSettings(factory.Type(), cfg.SDKLogLevel)
 		exp, err := factory.CreateLogs(ctx, set, config)
@@ -218,13 +199,16 @@ func getLogsExporter(ctx context.Context, cfg otelcfg.LogsConfig) (exporter.Logs
 			slog.Error("can't create OTLP HTTP logs exporter", "error", err)
 			return nil, nil, err
 		}
+		// The inner exporter's own queue/retry are disabled by
+		// buildHTTPLogsExporterConfig; queueCfg/retryCfg are applied here so
+		// there is exactly one buffering/retry layer, matching getTracesExporter.
 		wrapped, err := exporterhelper.NewLogs(ctx, set, cfg,
 			exp.ConsumeLogs,
 			exporterhelper.WithStart(exp.Start),
 			exporterhelper.WithShutdown(exp.Shutdown),
 			exporterhelper.WithCapabilities(consumer.Capabilities{MutatesData: false}),
-			exporterhelper.WithQueue(config.QueueConfig),
-			exporterhelper.WithRetry(config.RetryConfig))
+			exporterhelper.WithQueue(queueCfg),
+			exporterhelper.WithRetry(retryCfg))
 		return wrapped, host, err
 	case otelcfg.ProtocolGRPC:
 		slog.Debug("instantiating GRPC LogsReporter", "protocol", proto)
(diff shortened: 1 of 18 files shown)


pkg/export/otel/logs.go
```diff
@@ -0,0 +1,322 @@
...
+
+func (lr *logsOTELReceiver) processSpans(ctx context.Context, exp exporter.Logs, spans []request.Span, traceAttrs map[attr.Name]struct{}, sampler trace.Sampler) {
+	spanGroups := tracesgen.GroupSpans(ctx, spans, traceAttrs, sampler, lr.is, lr.selectorCfg.SensitiveQueryParamsCfg.Effective()...)
+
+	for _, spanGroup := range spanGroups {
+		if len(spanGroup) == 0 {
+			continue
+		}
+		sample := spanGroup[0]
+		if !sample.Span.Service.ExportModes.CanExportLogs() {
```
MrAlias, line 110: `suppressQueueProcessingSpans` is computed once from the global logs configuration, but this receiver applies a second, per-service gate. For a discovered service configured with `exports: [traces]`, the trace receiver still suppresses both timing sub-spans and this branch skips the replacement log, so the queue/processing timing disappears entirely.

Could we make suppression honor the same per-service export decision and cover the traces-only mode in a test?

pkg/export/otel/logsgen/logsgen.go
```diff
@@ -0,0 +1,72 @@
...
+	svcAttrs *svc.Attrs,
+	envResourceAttrs []attribute.KeyValue,
+	nodeMeta *meta.NodeMeta,
+	spans []tracesgen.TraceSpanAndAttributes,
+	reporterName string,
+) plog.Logs {
+	logs := plog.NewLogs()
+	rl := logs.ResourceLogs().AppendEmpty()
+
+	resourceAttrs := tracesgen.TraceAppResourceAttrs(cache, nodeMeta, svcAttrs)
```
MrAlias, line 46: `GenerateTracesWithSelectedResourceAttributes` applies `selectorCfg.SelectionCfg` and `ctxInfo.ExtraResourceAttributes`, but this parallel resource path applies neither. A resource attribute excluded through `attributes.select` is therefore still exported on these logs, while embedding-supplied resource attributes present on the parent trace are missing from the correlated log.

Could we reuse the trace pipeline's filtered resource construction and add a resource-exclusion test?

pkg/export/otel/logs.go
```diff
@@ -0,0 +1,322 @@
...
+				mwID: udsMiddleware{addr: opts.UnixSocketAddr},
+			}}
+		}
+		set := getTraceSettings(factory.Type(), cfg.SDKLogLevel)
+		exp, err := factory.CreateLogs(ctx, set, config)
+		if err != nil {
+			slog.Error("can't create OTLP HTTP logs exporter", "error", err)
+			return nil, nil, err
+		}
+		wrapped, err := exporterhelper.NewLogs(ctx, set, cfg,
```
MrAlias, line 205: `factory.CreateLogs` already builds an exporter helper with `config.QueueConfig` and `config.RetryConfig`; this wrapper installs the same queue/batcher and retry policy again. HTTP logs therefore get nested buffering, backpressure, and retry loops, unlike gRPC and the trace implementation, which disables the inner helpers before wrapping.

Could we keep a single helper layer and exercise the HTTP configuration in a test?

MrAlias (commented):
The replacement is not yet behaviorally equivalent to the spans it suppresses: signal gating, resource policy, and HTTP exporter behavior can all diverge from the corresponding trace pipeline.

marctc (commented):
I understand that this PR is trying to solve #1193, but it's adding the whole log pipeline. Perhaps it's the only way to go, but I'd to in separate PRs.

pkg/export/otel/logs.go
mlhv: Good note - suppression was global so a traces-only service would lose the timing completely. Now gated per-span group on the same `CanExportLogs()` export check the Logs receiver uses. Sub-spans will stay for services that can't get the replacement log. Added a test mixing a traces-only and an all-signals service in one batch. 

commit: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2863/changes/3b6ba814b635924a8b296ea63c86d6eeeef23357

pkg/export/otel/logsgen/logsgen.go
mlhv: Fixed. `GenerateLogs` now runs resource attrs through the same `FilterResourceAttrs/extraResAttrs` path as `generateTracesWithAttributes`, including matching its exact write order. Exported `tracesgen.AddAttrsToMap` so both paths share one implementation. Also added a resource-exclusion test.

commit: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2863/changes/3b3aa58f61b9d863eb73845ab916dbbd3a0389fe, https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2863/changes/e03148c61b442db875bd36317b3513cf6061c008

pkg/export/otel/logs.go
mlhv: Single helper layer applied. Inner exporter's queue/retry are now disabled before `CreateLogs`, and real settings are applied once via the outer wrap. Added a test asserting the inner config is disabled, plus a smoke test for the HTTP configurations.

commit: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2863/changes/7accb1ca88942033f3a66f375fc26e32f0ded4a6

mlhv:
> I understand that this PR is trying to solve #1193, but it's adding the whole log pipeline. Perhaps it's the only way to go, but I'd to in separate PRs.

That's reasonable. I kept the logs plumbing + the actual feature (subspan suppression) together since the plumbing on its own doesn't really do anything. It is a lot to review at once though and a whole new logs pipeline is probably worth its own PR (and the commits are split along the same line, so cutting it into two PRs is fine with me if you'd rather do it that way)

@MrAlias Thank you for the review, I have made fixes accordingly!

Let me know which approach you prefer.

marctc:
> since the plumbing on its own doesn't really do anything. It is a lot to review at once though and a whole new logs pipeline is probably worth its own PR

I mean, you can still configure OBI to export logs with this PR, right? I think its better to separate things yeah, and probably its also worth to add here some integration tests for logs.

MrAlias:
> I mean, you can still configure OBI to export logs with this PR, right? I think its better to separate things yeah, and probably its also worth to add here some integration tests for logs.

Agreed. Let’s split the generic logs export pipeline from the queue/processing behavior. The plumbing PR should include an end-to-end OTLP logs export test; the current debug-exporter test only covers generation and wiring.

mlhv:
Sounds good, splitting it. 

Plumbing PR gets `LogsConfig`, `getLogsExporter` and its HTTP/gRPC/debug branches, and the `LogsConsumer` embedding hook (no receiver, no wiring into the graph yet, since the only thing that would call it right now is the queue/processing fix) For the end-to-end test, I'll mirror `TestOTLPMetricsOverUDS` (real `otelcol` container, not the debug exporter) and drive the exporter directly through `LogsConsumer` to prove real OTLP delivery.

Feature PR (depends on the first one merging) gets `logsgen`, the receiver/wiring, and the trace suppression toggle. Opening the plumbing PR next, will link it here!

grcevski:
My original idea was to use span events, which end up in traces and help debug tracing related latencies directly. I know those became deprecated, so if we are not going to use span events, I prefer that we keep the internal spans, since I don't think for this purpose it would make sense to write the timings of the outer transaction in logs. 

My opinion is, we either use span events (deprecated) or do nothing.

MrAlias:
Thank you for the contribution. We plan to add these values in span event for the time being. Closing.

mlhv:
Hi, sorry I couldn't get to this earlier. I agree with all the discussion above; span events fit better for this use case despite the deprecated notices. Thank you @grcevski and @MrAlias for taking the time to review this and for the clear reasoning!

Hoping to revisit the task maybe in the future.

PR #2866, opened 2026-08-02
Weaver 0.25.1 Modernization: `.weaver.toml` live-check filtering + compact report over HTTP `/stop`

NimrodAvni78:
## Summary

Bumps weaver to `v0.25.1` and modernizes how OBI runs its live-check semantic-convention validation.
Main goal is to minimize custom weaver handling and use its pre-built features, also reduces code on our side:

**Filtering moves into the registry (`.weaver.toml`).** Weaver 0.25 can declare finding filters and level overrides in the registry, so the hand-rolled advice classification in `weavercheck` is gone. `schemas/obi/.weaver.toml` promotes `extends_namespace` / `undefined_enum_variant` to `violation` and drops the known-safe `illegal_namespace` collisions (`server`/`client` address+port, `iface.direction`); a novel collision on any other sample still fails. `Validate` now just enforces "zero violations".

**The report is read over the admin `/stop` HTTP response, made safe by a `compact` format.** The harness POSTs `/stop` and reads the report from the response body (weaver runs with `--output http`) — no shared bind mount or `docker wait`. Weaver's builtin JSON report is ~135 MB for a real multi-language suite (every sample with its full body, each advisory repeated), big enough that weaver can close the connection before the client drains it — the truncation that made earlier HTTP attempts fail. A custom `compact` output template (`schemas/obi/.live_check_templates/compact/`) reduces the body to tens of KB via a jq filter: the statistics block plus a deduplicated findings list (message, level, advice type, count, and the distinct signals that triggered it) — bounded by distinct message/signal pairs, not sample count. That clears the socket buffer before weaver exits, so the body arrives intact. `weavercheck.Report` carries those `Findings` and `Validate` logs the same violation-first, level-grouped breakdown as before.

**Also:** re-enables OATS weaver validation (drops a stray `|| true`); adds a `lint-schema` allowlist clause for 0.25's `DeprecatedIncludeUnreferencedWarning` (OBI still needs `--include-unreferenced`); and groups all six `otel/weaver:` Renovate refs — including `service.yml`, previously tracked by no manager — so they bump together.

## Validation

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -72,6 +72,21 @@
       ],
       "datasourceTemplate": "docker",
       "versioningTemplate": "regex:^(?:(?<compatibility>.+?)-)?v?(?<major>\\d+)\\.(?<minor>\\d+)\\.(?<patch>\\d+)$"
+    },
+    {
+      // internal/test/integration/components/weaver/service.yml is a Compose
+      // service fragment, but its filename doesn't match the docker-compose
+      // manager's compose*.yml pattern, so its digest-pinned image would
+      // otherwise never be updated (it silently lagged a version behind the
+      // other weaver pins). Track it with the same generic image-ref shape used
+      // for the Go files above.
+      "customType": "regex",
+      "managerFilePatterns": ["/^internal\\/test\\/integration\\/components\\/weaver\\/service\\.yml$/"],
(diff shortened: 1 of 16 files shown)


mmat11 (approved):
lgtm

Issue #2778, opened 2026-07-24
bpf: SQL query completion clobbers enclosing HTTP server span context


mtanda:
### Component

OBI (eBPF instrumentation)

### What happened?

`process_sql_return()` in `bpf/gotracer/go_sql.c` calls `obi_ctx__del()` unconditionally when a SQL query returns.

`traces_ctx_v1` (`bpf/shared/obi_ctx.h`) is keyed only by `pid_tgid`, with no notion of span nesting/depth. When a SQL query runs inside an HTTP handler, this unconditional delete also clears the enclosing HTTP server span's context, even though that HTTP span is still active.

A very common shape hits this: logging something after a query that built the response. Because the context was cleared, `ebpf.log_enricher` silently stops injecting `trace_id`/`span_id` into any log line written after the query returns, even though the request is still in-flight.

### What did you expect to happen?

Log lines written after a SQL query completes, but still within the enclosing HTTP server span, should still get `trace_id`/`span_id` injected by `ebpf.log_enricher`.

### Steps to reproduce

1. Instrument a Go HTTP service that uses database/sql (e.g. gin + gorm + Postgres) with `ebpf.log_enricher` enabled.
2. In the HTTP handler, run a SQL query, then log a line after the query returns (e.g. logging the response before writing it).
3. Send a request to the service and inspect the log line's fields.

### OBI configuration

```yaml
ebpf:
  log_enricher:
    services:
    - service:
      - open_ports: "8082"
        exe_path: /usr/local/bin/shop-api
```

### Relevant logs

Before fix (trace_id/span_id missing):
```
{"time":"2026-07-23T10:43:37.119635502Z","level":"INFO","msg":"request completed","method":"GET","path":"/api/v1/products","status":200,"duration_ms":0}
```

After fix (trace_id/span_id present):
```
{"duration_ms":0,"level":"INFO","method":"GET","msg":"request completed","path":"/api/v1/products","span_id":"cd02f93636a93277","status":200,"time":"2026-07-24T04:56:11.694275471Z","trace_id":"5f3bba18b49e28f1d0947d37da94fe0a"}
```

### Additional context

Opened #2775 as a minimal patch for this specific case; it's not a proper fix for the underlying problem.

### Version

v0.10.0

### Kernel version

6.8.0-117-generic

### OS / Distribution

Ubuntu 24.04.4 LTS (colima VM, aarch64)

### Deployment

Docker / container

PR #2875, opened 2026-08-03
test: reproduce log enrichment gap after HTTP->SQL->log sequence

mtanda:
## Summary

Adds an integration test reproducing the log enrichment gap described in #2778: after a SQL query completes inside an HTTP handler, `ebpf.log_enricher` stops injecting `trace_id`/`span_id` into log lines written afterward, because `process_sql_return()` unconditionally clears the `pid_tgid`-keyed span context that the enclosing HTTP span still relies on.

- `go_sql_logger`: a minimal Go HTTP service (net/http + database/sql) that runs a SQL query then logs a line before responding, matching the HTTP -> SQL -> log sequence from the issue.
- `log_enricher_test.go` / `docker-compose-log-enricher.yml` / `suites_test.go`: wires the new component into the log enricher integration test suite and asserts the post-query log line carries `trace_id`/`span_id`.

This is a test-only change to capture the reproduction; it does not include the eBPF fix.

## Test plan
- [ ] `go test -v -run TestLogEnricher -timeout 10m ./internal/test/integration/` (requires Linux/eBPF environment)

Related to #2778


internal/test/integration/docker-compose-log-enricher.yml
```diff
@@ -90,6 +90,16 @@ services:
...
 
+  testserversql:
+    build:
+      context: ../../..
+      dockerfile: internal/test/integration/components/go_sql_logger/Dockerfile
+    image: hatest-testserver-logenricher-sql-go
+    networks:
+      - shared
+    ports:
+      - "8389:8090"
```
MrAlias, line 101: Port 8090 is absent from `obi-config-log-enricher.yml`'s log-enricher service selector, which currently allows only `8380,50051,3030,8085,3040,8388`. Discovery can still select this process for tracing via `OTEL_EBPF_OPEN_PORT`, but `ProcessTracer.AllowPID` skips the log enricher when the independent selector does not match. As a result, this service cannot emit an enriched line and the new tests time out before exercising the SQL-return behavior. Could we add 8090 to the log-enricher allowlist?

internal/test/integration/suites_test.go
```diff
@@ -1113,6 +1113,22 @@ func TestSuite_LogEnricherMultiSegWritev(t *testing.T) {
...
 
+func TestSuite_LogEnricherGoSQL(t *testing.T) {
+	compose, err := docker.ComposeSuite("docker-compose-log-enricher.yml", path.Join(pathOutput, "test-suite-log-enricher-go-sql.log"))
+	require.NoError(t, err)
+
+	compose.Env = append(compose.Env, `OTEL_EBPF_OPEN_PORT=8090`, `OTEL_EBPF_EXECUTABLE_PATH=`)
+	require.NoError(t, compose.Up())
+
+	t.Run("Log Enricher Go SQL", func(t *testing.T) {
+		testLogEnricher(t, logEnricherGoSQLConstants)
```
MrAlias, line 1124: `testLogEnricher` only checks that `trace_id` and `span_id` are present. A SQL-return change that leaves the completed SQL child in `traces_ctx_v1` would pass while correlating the post-query log to the wrong span. Could this reproduction assert that the IDs belong to the still-active HTTP server span, rather than the completed SQL span?

internal/test/integration/suites_test.go
```diff
@@ -1113,6 +1113,22 @@ func TestSuite_LogEnricherMultiSegWritev(t *testing.T) {
...
+	})
+	t.Run("Log Enricher Go SQL manual SDK", func(t *testing.T) {
```
MrAlias, line 1126: This subtest covers a separate gap: the helper documents that `go_sdk.c` never writes manual spans to `traces_ctx_v1`, so restoring context after SQL return cannot make it pass. Since it runs in the normal integration matrix, the SQL regression suite remains red even after #2778 is fixed. Could we remove the manual-SDK path from this PR and track it independently?

MrAlias (commented):
The current wiring does not yet make this a reliable acceptance test for restoring the enclosing HTTP context after SQL completion.

marctc (commented):
Thanks for this! Of course we can't merge the PR as it is because CI would fail. We could technically merge this PR but we wouldn't need to skip the test for now. We can also leave the PR as draft (or closed) and link it to the issue too.

mtanda:
@MrAlias @marctc
Thank you for your feedback. I understand that I need to improve the tests.
Sorry for bothering you both. These tests are included in [#2876](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2876). I will close this PR and continue improving the tests there instead.
I will focus on the SQL span first, and the manual span next.

internal/test/integration/suites_test.go
mtanda: I have created a separate issue for the manual span case: [#2932](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2932).
I also removed the fix for the manual span from this PR: [#2876](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2876).
I will address that point when I open a PR for the manual span.

internal/test/integration/docker-compose-log-enricher.yml
mtanda: Thanks for pointing that out. I've added the port in this PR.
https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2876/changes/fe51270af3287e50eae9a650f9f59082b1982978

internal/test/integration/suites_test.go
mtanda: I've added the test in this commit.
https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2876/changes/810f3771237caf69ba90d7c9a8a3c0d279eb3e29

PR #2876, opened 2026-08-03
fix(gotracer): restore prior trace context after SQL scopes end

### LINKED ISSUE #2778: bpf: SQL query completion clobbers enclosing HTTP server span context

mtanda:
## Summary

Fixes #2778: `process_sql_return()` in `bpf/gotracer/go_sql.c` cleared the
log enricher's trace context unconditionally when a SQL query returned.
`traces_ctx_v1` is keyed only by `pid_tgid`, with no notion of span
nesting, so this unconditional clear also dropped the enclosing HTTP
span's context — even though that span was still active. Any log line
written after the query returned (but still inside the HTTP handler)
silently lost its `trace_id`/`span_id`.

- `fix: restore prior trace context after nested SQL query scopes` —
  `set_sql_info()` now saves whatever context was active before the query
  started; `process_sql_return()` restores it instead of deleting it
  unconditionally.

## Test plan
- [x] `TestSuite_LogEnricherGoSQL` (added in #2875) passes locally:
  `go test -p 1 -failfast -v -run TestSuite_LogEnricherGoSQL -timeout 20m ./internal/test/integration/`

Depends on #2875, which adds the reproduction test this fix makes pass.
Until #2875 merges, this PR's diff includes that test commit; it'll
shrink to just the fix commit after a rebase once #2875 lands.


mmat11:
hi, as written in the issue, this bug is shared with all other protocols in gotracer. I have a tentative implementation for all protocols here https://github.com/coralogix/opentelemetry-ebpf-instrumentation/pull/185/changes but I didn't had the time to test it

mtanda:
@mmat11
Hi, thank you so much for taking the time to look into this and work on a fix! I hadn't noticed your work, and I didn't realize there was a way to fix this across all protocols at once. I was planning to fix them one by one, starting with the ones that have the smallest impact.

One thing I'd like to mention is how this relates to [Auto SDK](https://opentelemetry.io/docs/zero-code/go/autosdk/) support. This wasn't something I raised in the original issue. I only noticed it myself afterward, so I wanted to bring it up here. Since you know the codebase much better than I do, I'd love to hear your thoughts on how your approach might interact with that.

bpf/gotracer/go_sdk.c
```diff
@@ -249,6 +251,9 @@ int obi_uprobe_tracer_Start_Returns(struct pt_regs *ctx) {
             go_addr_key_from_id(&gp_key, (void *)span->parent_go);
             update_tp_parent_go(&gp_key, &span->tp);
 
+            // Make this span the log enricher's active context too.
+            obi_ctx__set(bpf_get_current_pid_tgid(), &span->tp);
```
MrAlias: This only updates the thread that executes the return uprobe, so the manual context is lost as soon as the goroutine parks or is preempted. `runtime.casgstatus` rebuilds `traces_ctx_v1` from the protocol-specific ongoing maps and never consults `go_trace_map` or `active_spans`; inside an HTTP handler it therefore restores the HTTP ancestor while this manual span is still active.

Could we include manual spans in that goroutine-aware restore path and force a scheduler transition in the test?

bpf/gotracer/go_sdk.c
```diff
@@ -282,6 +287,9 @@ int obi_uprobe_nonRecordingSpan_End(struct pt_regs *ctx) {
         go_addr_key_t gp_key = {};
         go_addr_key_from_id(&gp_key, (void *)span->parent_go);
         update_tp_parent_go(&gp_key, &span->prev_tp);
+
+        // Restore the log enricher's context to what it was before this span.
+        obi_ctx__set(bpf_get_current_pid_tgid(), &span->prev_tp);
```
MrAlias: This is not necessarily the context that this call displaced. `prev_tp` came from `go_trace_map`, while an active SQL span exists only in `ongoing_sql_queries`/`traces_ctx_v1`, so a manual span started inside SQL restores the HTTP ancestor here instead of the still-active SQL span. Also, `Span.End` may run on a different goroutine; in that case this writes the original span's parent into that unrelated goroutine's current thread context.

Could we snapshot the actual shared context at `Start` and restore it through the owning goroutine's scheduler lifecycle?

internal/test/integration/log_enricher_test.go
```diff
@@ -923,6 +930,86 @@ func testLogEnricher(t *testing.T, constants testServerConstants) {
...
+// obiSigSpanTraceparent matches OBI's text trace printer output for the
+// "sig" manual span (go.opentelemetry.io/auto/sdk), capturing its own
+// trace_id and span_id from the traceparent field
+// (00-<trace_id>-<span_id>[<parent_span_id>]-<flags>).
+var obiSigSpanTraceparent = regexp.MustCompile(
+	`CUSTOM\(subType=\d+\) \d+ sig\b.*traceparent=\[00-([0-9a-f]{32})-([0-9a-f]{16})\[[0-9a-f]{16}\]-[0-9a-f]{2}\]`)
+
+// testLogEnricherGoSQLManualSDK exercises a "sig" manual span
+// (go.opentelemetry.io/auto/sdk) that runs a SQL query and then logs before
+// ending. go_sdk.c never writes into traces_ctx_v1, so the logenricher has
```
MrAlias: This description is stale in the same patch: `go_sdk.c` now writes `traces_ctx_v1`, and CI expects this test to pass. The assertion message below repeats the old behavior, which will make a future failure point in the wrong direction.

MrAlias (commented):
The SQL save/restore fixes the reported HTTP→SQL→log sequence, but the manual-span path is still thread-local rather than goroutine-aware and can restore the wrong active context. The new test covers only the no-reschedule, manual-around-SQL ordering, so these failures remain hidden despite green CI.

mtanda:
@MrAlias Thanks for the comments. I've learned a lot from them. I now understand how much complexity OBI handles to get this right. It's amazing work.
I also realize that fixing log enrichment for manual spans is more involved than the SQL case, so I'll open a separate issue to track it.

mtanda:
@mmat11 Following the discussion with @MrAlias on this PR, I felt goroutine-related behavior needed more testing. While trying that out on coralogix#185, I found another test case worth reporting.

`testLogEnricherNestedSpans` passes as-is, but I tried a variant where the SQL call runs on its own goroutine instead of the same one as the HTTP handler (handler blocks on a channel until it's done):

```diff
@@ -325,12 +325,20 @@ func main() {

 		jsonLog("nested: after grpc " + id)

-		rows, err := db.Query("SELECT n FROM fake")
-		if err != nil {
+		sqlDone := make(chan error, 1)
+		go func() {
+			rows, err := db.Query("SELECT n FROM fake")
+			if err != nil {
+				sqlDone <- err
+				return
+			}
+			rows.Close()
+			sqlDone <- nil
+		}()
+		if err := <-sqlDone; err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
-		rows.Close()

 		jsonLog("nested: after sql " + id)
```

With that change, the test fails outright. None of the `nested: *` log lines carry trace/span context at all.

Here is my understanding of why, though I am not fully confident in it. `process_sql_return()` in `go_sql.c` deletes this goroutine's own entry from `ongoing_sql_queries` for its `g_key` (the SQL span is ending, so that part is correct), then calls `obi_ctx__restore(pid_tgid, g_key)` with the same `g_key` to look for whatever span should still enclose it, checking the other `ongoing_*_requests` maps in `go_obi_ctx.h`. Since the SQL query runs on a different goroutine than the HTTP handler, the enclosing HTTP span is registered under the handler's own, different `g_key`, not under the SQL goroutine's `g_key`. So none of those lookups find anything, and it falls through to `obi_ctx__del(pid_tgid)`.

If this is a use case that should be supported (a nested call finishing on a separate goroutine from the one that owns the enclosing span), coralogix#185 might not handle it correctly yet. I have not gone beyond reading the code for this, but goroutine-based fan-out for things like SQL calls is common enough in Go that I thought it was worth flagging.

mtanda:
Dropped the manual span fix (`obi_ctx__set()` calls added to `Tracer.Start`/`nonRecordingSpan.End` in `go_sdk.c`) from this PR, along with its test. The PR now only covers the SQL query case in `go_sql.c` (`set_sql_info()` / `process_sql_return()`), which is what #2778 actually reports.

bpf/gotracer/go_sql.c
```diff
@@ -331,7 +331,8 @@ static __always_inline int process_sql_return(void *goroutine_addr, u8 error, u8
         return 0;
     }
     bpf_map_delete_elem(&ongoing_sql_queries, &g_key);
-    obi_ctx__del(bpf_get_current_pid_tgid());
+
+    obi_ctx__restore(bpf_get_current_pid_tgid(), &invocation->tp);
```
mtanda, line 335: Referencing https://github.com/coralogix/opentelemetry-ebpf-instrumentation/pull/185/changes, I tried a small variation here: instead of looking up which span is still active, this preserves the tp already stored in `ongoing_sql_queries` and restores it via `obi_ctx__set()`. That tp was set by `client_trace_parent()` when the query started, so its parent_id already holds the correct enclosing span's id, no extra lookup needed. Not sure this has zero edge cases, but if it holds up, it's a minimal fix with no extra runtime cost.

grcevski: I think this might be the wrong trace information to be restored. What we need to lookup is the ongoing http requests and restore that context.

bpf/shared/obi_ctx.h
```diff
@@ -45,3 +46,18 @@ obi_ctx__set_(const u64 pid_tgid, const tp_info_t *info, obi_ctx_info_t *obi_inf
 static __always_inline long obi_ctx__del(const u64 pid_tgid) {
     return bpf_map_delete_elem(&traces_ctx_v1, &pid_tgid);
 }
+
+// Restores traces_ctx_v1 to the enclosing span's identity encoded in a
+// finished child span's own tp (ip): same trace_id, and parent_id holds the
+// enclosing span's span_id. Deletes the context when there was no enclosing
+// span, so the finished child span can't leak into later work on this thread.
+static __always_inline void obi_ctx__restore(const u64 pid_tgid, const tp_info_t *ip) {
```
MrAlias, line 54: `parent_id` records ancestry, not whether that parent is still active. When an OBI-captured manual span with an HTTP parent encloses SQL, this restores the manual span here; `Span.End()` repairs only `go_trace_map`, so an immediate log is attributed to the ended span.

Could we preserve the context SQL actually displaced in goroutine-owned state, or otherwise validate parent liveness before writing `traces_ctx_v1`?

internal/test/integration/components/go_sql_logger/main.go
```diff
@@ -0,0 +1,73 @@
...
+				sqlDone <- err
+				return
+			}
+			rows.Close()
+			sqlDone <- nil
+		}()
+		if err := <-sqlDone; err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
```
MrAlias, line 54: `process_sql_return()` runs on this child, but the assertion logs later on the handler. The handler's HTTP context is either never displaced or is rebuilt by `runtime.casgstatus`, so this test can pass without validating the restore added by this PR.

Could we keep #2778's synchronous HTTP→SQL→log sequence as the regression case? A separate goroutine case should log after SQL on this goroutine.

```suggestion
		rows, err := db.Query("SELECT * FROM students")
		if err != nil {
			http.Error(w, err.Error(), http.StatusInternalServerError)
			return
		}
		rows.Close()
```

internal/test/integration/log_enricher_test.go
```diff
@@ -923,6 +930,91 @@ func testLogEnricher(t *testing.T, constants testServerConstants) {
...
+		}
+		appLogs := containerLogs(ct, cl, appContainerID)
+		if !assert.NotEmpty(ct, appLogs) {
+			return
+		}
+
+		logIdx := -1
+		// Loop from the end -- it might be possible that OBI wasn't ready to inject
+		// context when the test started, so get the latest request logs every time.
+		for i := len(appLogs) - 1; i >= 0; i-- {
```
MrAlias, line 964: `containerLogs` returns the full container history, and the preceding subtest already guarantees an enriched line in the same container. If this request emits an un-enriched log, this scan skips it and can reuse that earlier request together with its older OBI spans.

Could we tag each request/log with a unique marker, or limit both scans to entries appended after a baseline?

MrAlias (commented):
The SQL-only revision still has a parent-lifecycle bug, and the new coverage can pass without validating the restored context. The PR title and description also need to drop the reverted manual-SDK claims and describe the current implementation.

mmat11:
> @mmat11 Following the discussion with @MrAlias on this PR, I felt goroutine-related behavior needed more testing. While trying that out on [coralogix#185](https://github.com/coralogix/opentelemetry-ebpf-instrumentation/pull/185), I found another test case worth reporting.
> 
> `testLogEnricherNestedSpans` passes as-is, but I tried a variant where the SQL call runs on its own goroutine instead of the same one as the HTTP handler (handler blocks on a channel until it's done):
> 
> ```diff
> @@ -325,12 +325,20 @@ func main() {
> 
>  		jsonLog("nested: after grpc " + id)
> 
> -		rows, err := db.Query("SELECT n FROM fake")
> -		if err != nil {
> +		sqlDone := make(chan error, 1)
> +		go func() {
> +			rows, err := db.Query("SELECT n FROM fake")
> +			if err != nil {
> +				sqlDone <- err
> +				return
> +			}
> +			rows.Close()
> +			sqlDone <- nil
> +		}()
> +		if err := <-sqlDone; err != nil {
>  			http.Error(w, err.Error(), http.StatusInternalServerError)
>  			return
>  		}
> -		rows.Close()
> 
>  		jsonLog("nested: after sql " + id)
> ```
> 
> With that change, the test fails outright. None of the `nested: *` log lines carry trace/span context at all.
> 
> Here is my understanding of why, though I am not fully confident in it. `process_sql_return()` in `go_sql.c` deletes this goroutine's own entry from `ongoing_sql_queries` for its `g_key` (the SQL span is ending, so that part is correct), then calls `obi_ctx__restore(pid_tgid, g_key)` with the same `g_key` to look for whatever span should still enclose it, checking the other `ongoing_*_requests` maps in `go_obi_ctx.h`. Since the SQL query runs on a different goroutine than the HTTP handler, the enclosing HTTP span is registered under the handler's own, different `g_key`, not under the SQL goroutine's `g_key`. So none of those lookups find anything, and it falls through to `obi_ctx__del(pid_tgid)`.
> 
> If this is a use case that should be supported (a nested call finishing on a separate goroutine from the one that owns the enclosing span), [coralogix#185](https://github.com/coralogix/opentelemetry-ebpf-instrumentation/pull/185) might not handle it correctly yet. I have not gone beyond reading the code for this, but goroutine-based fan-out for things like SQL calls is common enough in Go that I thought it was worth flagging.

hi, sorry for the late reply, I picked this up again today, does this PR address this case? if negative, I would still prefer to go with the generic solution -- I'll try to add a test exercising this pattern and report back tomorrow

mtanda:
@mmat11 That sounds like the right direction, thank you for taking this on.
I've been digging into this a bit on my own, and I could well be misunderstanding parts of it, but I wanted to share a couple of points that are on my mind. It depends on how far we want log enrichment's guarantees to extend, but goroutines that split off (e.g. via a channel) and manual spans in general might need some thought too. From what I can tell, the channel case seems to be connected via span links rather than a direct parent-child relationship, so that might be another case worth considering. This part feels genuinely hard to me, and if supporting log enrichment for these cases turns out to be too difficult, it might make sense to just scope them out as unsupported patterns...

bpf/gotracer/go_sql.c
mtanda: The intent was to also cover manual spans, not just HTTP requests, though I'm not sure yet what the right way to handle them is. Hoping there's a good approach for that...

mmat11:
@mtanda sorry again for the delay -- we've decided to close this PR and proceed with the generic implementation. Will drop you a ping once it's ready for review, thanks for catching this!

PR #2880, opened 2026-08-03
Route integration workflows by build inputs

MrAlias:
## Summary

- Route VM integration checks for the Go, configuration, test, schema, action, script, and Docker build inputs they consume.
- Route Java Agent CI when the root dependency image definition changes.
- Keep Markdown-only pull requests out of the VM matrix.
- Invalidate prebuilt Docker artifacts when their source or producer scripts change.

## Why

Several checked-in build inputs could change without scheduling the integration workflow that consumes them. Some newly routed inputs could also reuse cached artifacts produced from older source, allowing CI to pass without exercising the pull request's change.

## Testing

- Ordered path and cache contract assertions — passed
- YAML parsing for both edited workflows — passed
- Focused `actionlint` validation — passed
- `make lint` — passed
- `git diff --check` — passed

### DIFF
--- a/.github/workflows/java-agent.yml
+++ b/.github/workflows/java-agent.yml
@@ -7,6 +7,7 @@ on:
     branches: ["main", "release-*"]
     paths:
       - "pkg/internal/java/**"
+      - "dependencies.Dockerfile"
       - ".github/workflows/java-agent.yml"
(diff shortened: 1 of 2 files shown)


grcevski (approved):
LGTM!

PR #2881, opened 2026-08-03
Harden raw trace record dispatch

MrAlias:
## Summary

- Add deterministic record-boundary, stale-output, lifecycle, resource, and fuzz coverage around the common raw trace dispatcher.
- Reject nil records, oversized TCP large-buffer payloads, negative HTTP/2 capture lengths, and invalid Go OTel attribute counts before unsafe access.
- Preserve partial Go OTel spans when attribute JSON encoding fails; only a malformed attribute count invalidates the record.

## Why

Raw ring-buffer records cross a kernel-to-userspace trust boundary and can be truncated or contain inconsistent length and count fields. Several malformed forms could panic while slicing fixed buffers, while the dispatcher lacked complete boundary and mutation coverage. These checks preserve the existing ignored-zero-span error contract and intentional HTTP event routing.

An out-of-bounds Go OTel attribute count is structural record corruption and rejects the event. A JSON encoding failure does not make the remaining span metadata unusable, so that path retains the previous empty-attributes behavior.

## Testing

- `go test -count=1 ./pkg/ebpf/common` — passed
- `go test -race -count=1 ./pkg/ebpf/common` — passed
- `go vet ./pkg/ebpf/common` — passed
- `go test -count=100 -timeout=10s -run "^TestReadBPFTraceAsSpanReroutesHTTP2WithoutBlocking$" ./pkg/ebpf/common` — passed
- `go test -count=20 -run "^(TestReadBPFTraceAsSpanReroutesHTTP2WithoutBlocking|TestDispatchTraceRecordDoesNotLeakGoroutines|TestReadBPFTraceAsSpanResourceBounds)$" ./pkg/ebpf/common` — passed
- `go test -run "^$" -fuzz=FuzzReadBPFTraceAsSpan -fuzztime=5s ./pkg/ebpf/common` — passed with default parallelism (36 workers, 33,737 executions)
- `go test -run "^$" -fuzz=FuzzReadBPFTraceAsSpan -fuzztime=30s -parallel=1 ./pkg/ebpf/common` — passed in repeated resource-limited campaigns
- `git diff --check` — passed

### DIFF
--- a/pkg/ebpf/common/go_otel_transform.go
+++ b/pkg/ebpf/common/go_otel_transform.go
@@ -33,15 +33,15 @@ func ReadGoOTelEventIntoSpan(record *ringbuf.Record) (request.Span, bool, error)
 	name := cstr(event.SpanName.Buf[:])
 	descr := cstr(event.SpanDescription.Buf[:])
 
-	attrs := ""
-	if a, err := encodedAttrs(event); err == nil {
-		attrs = string(a)
+	attrs, err := encodedAttrs(event)
+	if err != nil {
+		return request.Span{}, true, err
 	}
 
 	return request.Span{
 		Type:          request.EventTypeManualSpan,
 		Method:        name,
-		Statement:     attrs,
+		Statement:     string(attrs),
 		Path:          descr,
 		Peer:          "",
 		PeerPort:      0,
@@ -206,6 +206,9 @@ func autoSpanStatus(status ptrace.StatusCode) (int, error) {
 
 func encodedAttrs(event *GoOTelSpanTrace) ([]byte, error) {
 	size := int(event.SpanAttrs.ValidAttrs)
+	if size > len(event.SpanAttrs.Attrs) {
+		return nil, fmt.Errorf("invalid Go OTel span attribute count: %d exceeds %d", size, len(event.SpanAttrs.Attrs))
+	}
 	if size == 0 {
 		return nil, nil
 	}
--- a/pkg/ebpf/common/go_otel_transform.go
+++ b/pkg/ebpf/common/go_otel_transform.go
@@ -24,6 +24,8 @@ import (
 
 const goAutoSpanJSONMaxLen = 16 * 1024
 
+var errInvalidGoOTelSpanAttributeCount = errors.New("invalid Go OTel span attribute count")
+
 func ReadGoOTelEventIntoSpan(record *ringbuf.Record) (request.Span, bool, error) {
 	event, err := ReinterpretCast[GoOTelSpanTrace](record.RawSample)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/ebpf/common/go_otel_transform.go
```diff
@@ -33,15 +33,15 @@ func ReadGoOTelEventIntoSpan(record *ringbuf.Record) (request.Span, bool, error)
 	name := cstr(event.SpanName.Buf[:])
 	descr := cstr(event.SpanDescription.Buf[:])
 
-	attrs := ""
-	if a, err := encodedAttrs(event); err == nil {
-		attrs = string(a)
+	attrs, err := encodedAttrs(event)
```
pinoOgni, line 38: So, now in `encodedAttrs` we return an error for the bounds violation (corrupted record), but the propagated `json.Marshal` error now also drops the whole span in `ReadGoOTelEventIntoSpan`, where before it was exported with empty attributes. 

Intended to treat both cases identically, or worth keeping the partial span for the marshal-failure case?

pkg/ebpf/common/raw_trace_dispatcher_test.go
```diff
@@ -0,0 +1,365 @@
...
+			parseCtx, cfg, &ringbuf.Record{RawSample: traceRecordBytes(&link)}, filter,
+		)
+		require.NoError(t, err)
+		require.True(t, ignore)
+	}
+	require.NotNil(t, parseCtx.pendingSpanLinks)
+	assert.Equal(t, maxPendingSpanLinks, parseCtx.pendingSpanLinks.cache.Len())
+}
+
+func FuzzReadBPFTraceAsSpan(f *testing.F) {
```
pinoOgni, line 326: I saw in the PR description that you used `-parallel=1`, with the default parallelism, workers would steal each other's events right? If yes, maybe a short comment on the fuzz target noting the `-parallel=1` requirement can be a good idea.

pinoOgni (changes requested):

pkg/ebpf/common/go_otel_transform.go
MrAlias: I separated the two cases in 951889653. `encodedAttrs` now wraps a sentinel only for an out-of-bounds attribute count, and `ReadGoOTelEventIntoSpan` rejects only that structural corruption. Other JSON encoding errors retain the previous behavior and emit the remaining span with empty attributes. The package tests, race tests, and `go vet` pass with this distinction.

pkg/ebpf/common/raw_trace_dispatcher_test.go
MrAlias: `-parallel=1` is not required for channel isolation here. During fuzzing, Go uses `-parallel` to control separate subprocesses, so each worker has its own package-level `MisclassifiedEvents` channel and cannot consume another worker's events. I reran the target without the flag: 36 workers completed 33,737 executions successfully. The original flag only limited local resource use, and the PR description now makes that distinction explicit.

pinoOgni (approved):
lgtm!

PR #2882, opened 2026-08-03
Harden reverse DNS message parsing

MrAlias:
## Summary

- Bound DNS question and answer slice capacity by sections physically representable in the remaining message.
- Reject incomplete sections, short or oversized labels, overlong names, invalid compression offsets, and truncated record data without returning partial results.
- Add compressed A, AAAA, and CNAME coverage, every-byte truncation tests, hostile-count allocation checks, pointer boundaries, and seeded fuzzing.

## Why

The reverse-DNS userspace parser consumed attacker-controlled section counts and malformed wire lengths. Hostile counts could reserve hundreds of kilobytes for a packet with no complete sections, while partial sections and a compression pointer at the message boundary could be accepted. The parser now fails the whole message coherently and keeps work proportional to bytes that can actually contain records.

## Testing

- `go test -race ./pkg/internal/rdns/ebpf/xdp` — passed
- `go test -race ./pkg/internal/rdns/...` — passed
- `go test -count=100 ./pkg/internal/rdns/ebpf/xdp` — passed
- `go test -race -count=20 -run "^TestHostileSectionCountsHaveBoundedAllocation$" ./pkg/internal/rdns/ebpf/xdp` — passed
- `go test -run "^$" -fuzz=FuzzParseDNSMessage -fuzztime=30s ./pkg/internal/rdns/ebpf/xdp` — passed in repeated campaigns
- `go vet ./pkg/internal/rdns/ebpf/xdp` — passed
- `git diff --check` — passed

### DIFF
--- a/pkg/internal/rdns/ebpf/xdp/parser.go
+++ b/pkg/internal/rdns/ebpf/xdp/parser.go
@@ -10,8 +10,15 @@ import (
 	"strings"
 )
 
-// wordSize represents the size of a DNS message word (2 bytes)
-const wordSize = 2
+const (
+	// wordSize represents the size of a DNS message word (2 bytes)
+	wordSize = 2
+
+	maxLabelLength    = 63
+	maxNameLength     = 255
+	minQuestionLength = 1 + 2*wordSize
+	minRecordLength   = 1 + 5*wordSize
+)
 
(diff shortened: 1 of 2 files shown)


grcevski (approved):
LGTM!

PR #2883, opened 2026-08-03
Preserve Config v2 rule semantics

MrAlias:
## Summary

- preserve authored first-match and last-match rule ordering when converting config v2 documents into runtime discovery selectors
- preserve explicit and omitted per-rule export and route refinements across runtime conversion
- reject rule shapes and policy values the runtime cannot represent without changing capture behavior
- reverse exported selector order so runtime-to-v2-to-runtime round trips retain precedence

## Why

The Config v2 schema can express rule combinations whose ordering or refinements were previously changed silently by the runtime converter. These generic conversion semantics are independent of the operator migration guide in #2799 and need focused review before that guide relies on them.

Extracted from #2799.


internal/config/convert/import.go
```diff
@@ -391,40 +430,112 @@ func validateV2RuleSelectorFamilies(rules []schema.Rule) error {
...
+			rule.Action != schema.CaptureActionExclude {
+			return fmt.Errorf(
+				"%s.action: unsupported value %q",
+				path,
+				rule.Action,
+			)
+		}
+		if ruleMatchEmpty(rule.Match) {
+			return fmt.Errorf("%s.match must define at least one selector", path)
+		}
```
skl, line 474: Suggest checking the following case which may fail at OBI startup?

```yaml
capture:
  policy:
    default_action: exclude
  rules:
    - action: include
      match:
        process:
          containers_only: true   # "instrument everything running in a container"
```

internal/config/convert/import.go
```diff
@@ -759,28 +884,54 @@ func regexSelectorFromRule(rule schema.Rule) services.RegexSelector {
...
 		selector.Routes = &services.CustomRoutesConfig{
 			PolicyOverrides: v2DirectionalRoutePolicyOverrides(refine.HTTP.Routes),
 		}
+	} else if completeRoutes {
+		selector.Routes = &services.CustomRoutesConfig{}
 	}
 }
 
-func applyV2RegexRuleRefinement(selector *services.RegexSelector, refine schema.RuleRefinement) {
+func applyV2RegexRuleRefinement(
```
skl: This and `applyV2GlobRuleRefinement` look identical except the signature, there's a chance they could drift in future... not sure if it's worth the abstraction to DRY it up or maybe just a comment to keep these two in sync with each other?

skl (approved):
lgtm, couple of non-blocking questions 🚀

internal/config/convert/import.go
MrAlias: Replaced the two type-specific refinement functions with `v2RuleRefinement`, which computes export modes and routes once before assigning them to either glob or regex selectors (7f7372a9e). This removes the duplicated implementation without adding a selector abstraction.

internal/config/convert/import.go
MrAlias: Added `ContainersOnly` to the non-empty criterion checks in both `GlobDefinitionCriteria.Validate` and `RegexDefinitionCriteria.Validate`. The exact Config v2 example now has a regression test that converts the rule, calls `got.Discovery.Validate()`, and verifies the resulting selector remains containers-only (a95195507). The focused selector package tests and vet pass.

PR #2884, opened 2026-08-03
Separate runnable Config v2 example

MrAlias:
## Summary

- replace the exhaustive Config v2 default-values example with a small, runnable standalone configuration
- retain the complete default mapping as an explicitly non-runnable reference fragment
- validate the runnable example and reference fragment according to their distinct roles
- point the default-parity check and design documentation at the correct artifacts

## Why

The previous default configuration mixed two incompatible purposes. It was useful as a complete field reference but unsafe to present as a runnable example. Separating those artifacts makes the operator path copy-pasteable without losing the audited default mapping.

This cleanup is independent of the migration CLI and guide in #2799, and its large YAML movement is easier to review separately.

Extracted from #2799.


NimrodAvni78 (approved):
LGTM

PR #2885, opened 2026-08-03
Harden StatsO11y ring-buffer decoding

MrAlias:
## Summary

- reject nil and empty stats ring-buffer records before reading the event tag
- exercise all four stats event types with fixed wire-size and offset fixtures
- cover every truncation boundary, TCP I/O count bounds, cancellation, channel closure, and fuzz safety

## Testing

- go test -race ./pkg/internal/statsolly/stats
- 30-second FuzzParseStat run
- go vet ./pkg/internal/statsolly/stats
- linux/arm64 test-package cross-compile

### DIFF
--- a/pkg/internal/statsolly/stats/tracer_ringbuf.go
+++ b/pkg/internal/statsolly/stats/tracer_ringbuf.go
@@ -36,24 +36,14 @@ func NewRingBufTracer(statsMap *ciliumebpf.Map, cfg *config.EBPFTracer) *RingBuf
 	return &RingBufTracer{
 		statsMap: statsMap,
 		cfg:      cfg,
+		forward:  ebpfcommon.ForwardRingbuf(cfg, statsMap, parseStat, nil, rtlog(), nil),
 	}
 }
 
 func (m *RingBufTracer) TraceLoop(out *msg.Queue[[]*ebpf.Stat]) swarm.RunFunc {
-	forward := m.forward
-	if forward == nil {
-		forward = ebpfcommon.ForwardRingbuf(
-			m.cfg,
-			m.statsMap,
-			parseStat,
-			nil, // filter: no batch-level filtering
-			rtlog(),
-			nil, // metrics
-		)
-	}
 	return func(ctx context.Context) {
 		defer out.MarkCloseable()
-		forward(ctx, out)
+		m.forward(ctx, out)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/internal/statsolly/stats/tracer_ringbuf.go
```diff
@@ -38,14 +40,17 @@ func NewRingBufTracer(statsMap *ciliumebpf.Map, cfg *config.EBPFTracer) *RingBuf
...
 func (m *RingBufTracer) TraceLoop(out *msg.Queue[[]*ebpf.Stat]) swarm.RunFunc {
-	forward := ebpfcommon.ForwardRingbuf(
-		m.cfg,
-		m.statsMap,
-		parseStat,
-		nil, // filter: no batch-level filtering
-		rtlog(),
-		nil, // metrics
-	)
+	forward := m.forward
```
pinoOgni: When reading this I saw the field, then the nil-check, and had to go hunting for who populates it and turns out it's only the test. What about doing something like:

```
func NewRingBufTracer(statsMap *ciliumebpf.Map, cfg *config.EBPFTracer) *RingBufTracer {
	return &RingBufTracer{
		statsMap: statsMap,
		cfg:      cfg,
		forward:  ebpfcommon.ForwardRingbuf(cfg, statsMap, parseStat, nil, rtlog(), nil),
	}
}

func (m *RingBufTracer) TraceLoop(out *msg.Queue[[]*ebpf.Stat]) swarm.RunFunc {
	return func(ctx context.Context) {
		defer out.MarkCloseable()
		m.forward(ctx, out)
	}
}
```

This removes the `forward == nil` check from `TraceLoop`, and any `RingBufTracer` built via the constructor is guaranteed a non-nil forward. The test can still override it via a struct literal, and the "nil means production path" convention goes away.  Also, `ForwardRingbuf` now built at constructor time instead of first `TraceLoop` call but I think it is ok.

WDYT?

pinoOgni (changes requested):
Nice catch!

pkg/internal/statsolly/stats/tracer_ringbuf.go
MrAlias: Initialized the forwarder in `NewRingBufTracer` and removed the nil fallback from `TraceLoop` in [4b37683](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2885/commits/4b37683642bbd389b7cf5ef1f875ca06ce1119e8).

pinoOgni (approved):

PR #2887, opened 2026-08-03
Harden watcher event decoding and shutdown

MrAlias:
## Summary

- reject truncated watcher bind records instead of accepting partial payloads
- make Ready and NewPort delivery cancellation-aware while preserving ring-buffer cleanup ownership
- add deterministic fixed-wire, cancellation, callback, and startup lifecycle coverage

## Testing

- `go test -race -count=1 -timeout=3m ./pkg/internal/ebpf/watcher`
- `go test -race -shuffle=on -count=100 -timeout=3m ./pkg/internal/ebpf/watcher`
- `go vet ./pkg/internal/ebpf/watcher`


grcevski (approved):
LGTM!

PR #2893, opened 2026-08-04
Order container store lifecycle updates

MrAlias:
## Summary

- apply container create and delete events to store state before forwarding them
- remove stale namespace and container-ID indexes when a PID is replaced
- serialize same-PID inspection and mutation to prevent stale replacement or resurrection
- add deterministic concurrency tests for blocked fan-out, cancellation, and lifecycle ordering

## Testing

- `go test -race ./pkg/appolly/discover ./pkg/kube`
- `go test -race ./pkg/appolly/discover -run '^TestContainerStoreUpdater' -count=100 -shuffle=on`
- `go test -race ./pkg/kube -run '^TestStoreSerializesConcurrentProcessMutations$' -count=100`
- `go vet ./pkg/appolly/discover ./pkg/kube`
- `make lint`


pkg/appolly/discover/container_updater.go
```diff
@@ -48,11 +48,10 @@ func updateLoop(
 					log.Debug("adding process", "pid", ev.Obj.FileInfo.Pid())
 					store.AddProcess(ev.Obj.FileInfo.Pid())
 				case EventDeleted:
-					// we don't need to handle process deletion from here, as the Kubernetes informer will
-					// remove the process from the database when the Pod that contains it is deleted.
+					store.DeleteProcess(ev.Obj.FileInfo.Pid())
```
MrAlias, line 54:
```suggestion
					// Apply deletion in this stage's event order so a delayed create cannot
					// restore stale PID state after upstream cleanup.
					store.DeleteProcess(ev.Obj.FileInfo.Pid())
```

pkg/appolly/discover/container_updater.go
```diff
@@ -48,11 +48,12 @@ func updateLoop(
 					log.Debug("adding process", "pid", ev.Obj.FileInfo.Pid())
 					store.AddProcess(ev.Obj.FileInfo.Pid())
 				case EventDeleted:
-					// we don't need to handle process deletion from here, as the Kubernetes informer will
-					// remove the process from the database when the Pod that contains it is deleted.
+					// Apply deletion in this stage's event order so a delayed create cannot
```
grcevski, line 52: Do you mind adding a debug log here if we encounter an issue and we need to debug?

pkg/kube/store.go
```diff
@@ -84,8 +84,9 @@ var DefaultResourceLabels = ResourceLabels{
 // - the inspected container.Info objects, indexed either by container ID and PID namespace
 // - a cache of decorated PodInfo that would avoid reconstructing them on each trace decoration
 type Store struct {
-	log    *slog.Logger
-	access sync.RWMutex
+	log           *slog.Logger
+	access        sync.RWMutex
+	processAccess sync.Mutex
```
grcevski, line 89: Do we need the two locks? Maybe all that's needed would be to extend the covered area by accessLock in AddProcess?

grcevski (approved):
LGTM! Minor comments.

pkg/appolly/discover/container_updater.go
MrAlias: Added `log.Debug("deleting process", "pid", ...)` immediately before `DeleteProcess` in 7b850c9ab. This gives the deletion path the same PID-level visibility as creation. The targeted race tests for `pkg/appolly/discover` and `pkg/kube` pass.

pkg/kube/store.go
MrAlias: The locks cover different scopes. `access` protects all store indexes and is also used by metadata readers and writers. `processAccess` serializes `AddProcess` with `DeleteProcess` across the potentially blocking `InfoForPID` lookup. Moving `access.Lock` before that lookup would preserve lifecycle ordering, but it would also block all store readers during container inspection, so I kept the separate lifecycle lock.

mmat11 (approved):

PR #2895, opened 2026-08-04
Harden TC manager lifecycle handling

MrAlias:
## Summary

- serialize TC and TCX shutdown with interface callbacks
- make repeated shutdown and late lifecycle callbacks safe
- surface link and program cleanup failures through a bounded error reporter
- add lifecycle, concurrency, and privileged cleanup coverage

## Validation

- `make lint` with Go 1.25.11
- package race tests
- privileged race tests
- focused lifecycle stress tests (100 repetitions)
- targeted `go vet`
- independent testing, architecture, and security reviews

### DIFF
--- a/pkg/internal/ebpf/tcmanager/error_reporter.go
+++ b/pkg/internal/ebpf/tcmanager/error_reporter.go
@@ -1,0 +1,48 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package tcmanager // import "go.opentelemetry.io/obi/pkg/internal/ebpf/tcmanager"
+
+import "sync"
+
+const errorBufferLen = 16
+
+type errorReporter struct {
+	mutex  sync.RWMutex
+	errors chan error
+	closed bool
+}
+
+func newErrorReporter() errorReporter {
+	return errorReporter{errors: make(chan error, errorBufferLen)}
+}
+
+func (r *errorReporter) channel() chan error {
+	return r.errors
+}
+
+func (r *errorReporter) emit(err error) {
+	r.mutex.RLock()
+	defer r.mutex.RUnlock()
+
+	if r.closed {
+		return
+	}
+
+	select {
+	case r.errors <- err:
+	default:
+	}
+}
+
+func (r *errorReporter) close() {
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+
+	if r.closed {
+		return
+	}
+
+	r.closed = true
+	close(r.errors)
+}
--- a/pkg/internal/ebpf/tcmanager/error_reporter.go
+++ b/pkg/internal/ebpf/tcmanager/error_reporter.go
@@ -10,20 +10,17 @@ import "sync"
 const errorBufferLen = 16
 
 type errorReporter struct {
+	// mutex prevents close from racing with a send to ch.
 	mutex  sync.RWMutex
-	errors chan error
+	ch     chan error
 	closed bool
 }
 
 func newErrorReporter() errorReporter {
-	return errorReporter{errors: make(chan error, errorBufferLen)}
+	return errorReporter{ch: make(chan error, errorBufferLen)}
 }
 
-func (r *errorReporter) channel() chan error {
-	return r.errors
-}
-
-func (r *errorReporter) emit(err error) {
+func (r *errorReporter) enqueue(err error) {
 	r.mutex.RLock()
 	defer r.mutex.RUnlock()
 
@@ -32,7 +29,7 @@ func (r *errorReporter) emit(err error) {
 	}
 
 	select {
-	case r.errors <- err:
+	case r.ch <- err:
 	default:
 	}
 }
@@ -46,5 +43,5 @@ func (r *errorReporter) close() {
 	}
 
 	r.closed = true
-	close(r.errors)
+	close(r.ch)
 }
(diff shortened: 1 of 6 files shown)


pkg/internal/ebpf/tcmanager/error_reporter.go
```diff
@@ -0,0 +1,47 @@
...
+//go:build linux
+
+package tcmanager // import "go.opentelemetry.io/obi/pkg/internal/ebpf/tcmanager"
+
+import "sync"
+
+const errorBufferLen = 16
+
+type errorReporter struct {
+	// mutex prevents close from racing with a send to ch.
```
pinoOgni, line 13: Is this comment needed?

pkg/internal/ebpf/tcmanager/tcxmanager.go
```diff
@@ -102,23 +135,26 @@ func (tcx *tcxManager) Shutdown() {
...
 	}
+	tcx.shutdown = true
 
-	tcx.programs = []*attachedProg{}
-	tcx.links = []*ifaceLink{}
+	tcx.unregisterCallbacksLocked()
+	tcx.cleanupLinksLocked()
 
-	close(tcx.errorCh)
+	tcx.programs = []*attachedProg{}
```
pinoOgni: I see that the tcmanager `Shutdown` is doing `tc.cleanupProgsLocked()`. Can we add  tcxmanager analog of `cleanupProgsLocked`?

pkg/internal/ebpf/tcmanager/netlinkmanager.go
```diff
@@ -32,6 +32,15 @@ type netlinkProg struct {
 	*ebpf.Program
 	name       string
 	attachType AttachmentType
+	closeFn    func() error
+}
+
+func (p *netlinkProg) Close() error {
```
pinoOgni, line 38: What about adding a shared helper in `util.go` instead of having 3 `Close()` for ifaceLink, attachedProg and netlinkProg?

Something like:
```
func closeWith(override func() error, fallback io.Closer) error {
	if override != nil {
		return override()
	}

	return fallback.Close()
}
```
Then we will have:
```
 func (p *netlinkProg) Close() error { return closeWith(p.closeFn, p.Program) }
```

pkg/internal/ebpf/tcmanager/error_reporter.go
```diff
@@ -0,0 +1,47 @@
...
+	mutex  sync.RWMutex
+	ch     chan error
+	closed bool
+}
+
+func newErrorReporter() errorReporter {
```
pinoOgni, line 19: nit: return `*errorReporter`

pinoOgni (changes requested):

pkg/internal/ebpf/tcmanager/error_reporter.go
MrAlias: Why? All uses of this type are `errorReporter` not `*errorReporter`.

pkg/internal/ebpf/tcmanager/error_reporter.go
MrAlias: This kind of explains the whole point of the type. I mean, it isn't explicitly required to function, but its intention is to helps developer understanding of why the type exists.

pinoOgni: Fine by me

pkg/internal/ebpf/tcmanager/error_reporter.go
pinoOgni: My bad, you are right!

pkg/internal/ebpf/tcmanager/tcxmanager.go
MrAlias: TCX shutdown now calls `cleanupProgsLocked()` after `cleanupLinksLocked()`. It closes every registered program, reports close failures through the existing error reporter, and clears the slice. `TestShutdownClosesPrograms` verifies both managers close each program exactly once; the package race suite and 100 cleanup stress repetitions pass.

pkg/internal/ebpf/tcmanager/netlinkmanager.go
MrAlias: Added `closeWith(override, fallback io.Closer)` in `util.go` and changed `netlinkProg`, `attachedProg`, and `ifaceLink` to delegate to it. The focused race suite and `go vet` pass with the shared fallback path.

pkg/internal/ebpf/tcmanager/error_reporter.go
```diff
@@ -0,0 +1,47 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build linux
+
+package tcmanager // import "go.opentelemetry.io/obi/pkg/internal/ebpf/tcmanager"
+
+import "sync"
+
+const errorBufferLen = 16
```
pinoOgni, line 10: nit: add a comment on `errorBufferLen` or `enqueue` noting that when the buffer is full, errors are dropped rather than blocking because callers hold the manager mutex, and that dropped errors are still logged by `emitError`. Without it, the silent drop in enqueue could read as a bug. I see that `TestManagersLogCleanupErrorsWhenErrorChannelIsFull` covers the behavior but IMO a comment is worth it.

pinoOgni (approved):
lgtm! Just a nit.

pkg/internal/ebpf/tcmanager/error_reporter.go
MrAlias: Documented the non-blocking behavior above `enqueue`: callers may hold a manager mutex, so a full buffer drops the error, and `emitError` has already logged it. `go test -race ./pkg/internal/ebpf/tcmanager` and `go vet ./pkg/internal/ebpf/tcmanager` pass.

PR #2896, opened 2026-08-04
Harden flaky-run attempt validation

MrAlias:
## Summary

- validate workflow-run metadata before rerun or cancellation mutations
- reject malformed, ambiguous, fractional, and out-of-range attempt values
- preserve bounded exponent handling while comparing normalized attempts
- add a hermetic 22-case shell contract suite for supervisor behavior

## Validation

- `make test-rerun-flaky`
- contract suite repeated 25 times
- `bash -n`
- ShellCheck
- full `make lint` with Go 1.25.11
- independent testing, architecture, and security reviews

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -322,7 +322,7 @@ compile-cache-for-coverage:
 	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) go build -cover -a -o bin/$(CACHE_CMD) $(CACHE_MAIN_GO_FILE)
 
 .PHONY: test test-rerun-flaky
-test: testoutput test-rerun-flaky
+test: testoutput
 	@echo "### Testing code"
 	KUBEBUILDER_ASSETS="$(shell go tool $(TOOLS_MODFILE) setup-envtest use $(ENVTEST_K8S_VERSION) -p path)" go test -short -race -a ./... -coverpkg=./... -coverprofile $(TEST_OUTPUT)/cover.all.txt
 
--- a/Makefile
+++ b/Makefile
@@ -321,10 +321,13 @@ compile-cache-for-coverage:
 	@echo "### Compiling K8s cache service to generate coverage profiles"
 	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) go build -cover -a -o bin/$(CACHE_CMD) $(CACHE_MAIN_GO_FILE)
 
-.PHONY: test
-test: testoutput
+.PHONY: test test-rerun-flaky
+test: testoutput test-rerun-flaky
 	@echo "### Testing code"
 	KUBEBUILDER_ASSETS="$(shell go tool $(TOOLS_MODFILE) setup-envtest use $(ENVTEST_K8S_VERSION) -p path)" go test -short -race -a ./... -coverpkg=./... -coverprofile $(TEST_OUTPUT)/cover.all.txt
+
+test-rerun-flaky:
+	@./scripts/rerun-flaky_test.sh
 
 .PHONY: test-privileged
 test-privileged: $(ENVTEST) testoutput
--- a/scripts/rerun-flaky.sh
+++ b/scripts/rerun-flaky.sh
@@ -14,13 +14,14 @@
 set -euo pipefail
 
 MAX_ATTEMPTS=2
-MAX_SHELL_ATTEMPT=2147483647
+# Bound untrusted API input to the maximum portable signed 32-bit integer.
+MAX_PORTABLE_SIGNED_INT=2147483647
 
 echo "Evaluating run ${RUN_ID} -- workflow: ${WORKFLOW_NAME}"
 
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -321,11 +321,14 @@ compile-cache-for-coverage:
 	@echo "### Compiling K8s cache service to generate coverage profiles"
 	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) go build -cover -a -o bin/$(CACHE_CMD) $(CACHE_MAIN_GO_FILE)
 
-.PHONY: test
-test: testoutput
+.PHONY: test test-rerun-flaky
+test: testoutput test-rerun-flaky
```
pinoOgni: Why do we need `test-rerun-flaky` here? The script should only run in CI, right?

scripts/rerun-flaky_test.sh
```diff
@@ -0,0 +1,210 @@
...
+  $'{"attempt":1,"name":"Unit tests","jobs":[]}\n{"attempt":1,"name":"Unit tests","jobs":[]}' \
+  1 "$view_call" "Invalid GitHub run response"
+run_case "exponential attempt" "Unit tests" \
+  '{"attempt":1e100,"name":"Unit tests","jobs":[{"name":"unit","conclusion":"failure"}]}' \
+  1 "$view_call" "Invalid GitHub run response"
+run_case "bounded exponential attempt" "Unit tests" \
+  '{"attempt":1e3,"name":"Unit tests","jobs":[{"name":"unit","conclusion":"failure"}]}' \
+  0 "$view_call" "Maximum re-run attempts reached"
+run_case "fractional attempt rounded by jq" "Unit tests" \
+  '{"attempt":1.0000000000000001,"name":"Unit tests","jobs":[{"name":"unit","conclusion":"failure"}]}' \
```
pinoOgni, line 189: This rejection works only on jq ≥ 1.7.

```
jq --version
jq-1.7

jq -n '1.0000000000000001 | tostring'
"1.0000000000000001"

jq --version
jq-1.6

jq -n '1.0000000000000001 | tostring'
"1"
```

scripts/rerun-flaky.sh
```diff
@@ -14,12 +14,45 @@
...
+      else
+        error("invalid run response")
+      end
+  end
+' <<<"$RUN_JSON"); then
+  echo "Invalid GitHub run response." >&2
+  exit 1
+fi
+
+mapfile -t RUN_DETAILS <<<"$PARSED_RUN"
```
pinoOgni: `mapfile` requires bash ≥ 4.

I did a test:
```
bash --version | head -1

GNU bash, version 5.2.21(1)-release (aarch64-unknown-linux-gnu)

bash -c 'mapfile -t arr <<< $'\''a\nb\nc'\''; echo "count=${#arr[@]} first=${arr[0]}"'

count=3 first=a


bash --version | head -1

GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)

bash -c 'mapfile -t arr <<< $'\''a\nb\nc'\''; echo "count=${#arr[@]} first=${arr[0]}"'

bash: mapfile: command not found
count=0 first=

```

pinoOgni: This should work:
```
RUN_DETAILS=()
while IFS= read -r line; do
  RUN_DETAILS+=("$line")
done <<<"$PARSED_RUN"
```

pinoOgni (changes requested):

Makefile
MrAlias: The script only runs on CI, but this runs the test that validates that script.

scripts/rerun-flaky_test.sh
MrAlias: c534fe8fe validates the raw attempt token before fromjson, so jq 1.6 cannot round 1.0000000000000001 to 1 before validation. The 22-case contract suite passes with the official jq 1.6 binary, including the fractional-attempt case.

scripts/rerun-flaky.sh
MrAlias: c534fe8fe replaces mapfile with a while-read loop that populates the array in the current shell. The full 22-case contract suite passes in the official Bash 3.2 container with jq 1.6.

Makefile
pinoOgni: Maybe I'm missing something but `make test` isn't what CI runs (unit tests go through `make run-unit-test-shard`), so as-is the suite only ever executes on developer machines, while the script it validates only ever executes in CI.

Can we move `test-rerun-flaky` to a CI step instead (path-filtered on `scripts/**`) and drop it from the default `test` target?

MrAlias: Updated in 707fbb841. `test-rerun-flaky` now runs once in the unit-test workflow's matrix-setup job, whose PR path filters include `scripts/**`, and it is no longer a prerequisite of `make test`.

pinoOgni (approved):
lgtm!

PR #2899, opened 2026-08-04
Fix config schema $defs collisions between same-named config types

### LINKED ISSUE #2898: `obi-schema` name collision: `CONFIG.md` documents keys that do not exist
### Component

OBI (eBPF instrumentation)

### What happened?

`cmd/obi-schema` identifies Go types by their package-less name — the reflector at [`main.go:332`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/071b7180/cmd/obi-schema/main.go#L332) is built without a `Namer`. Two different types with the same name therefore share one schema definition: the first one reflected wins, and the second is silently replaced by a `$ref` to it.

Two collisions exist in the config graph today:

- `otelcfg.MetricsConfig` ([`config_metrics.go:44`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/071b7180/pkg/export/otel/otelcfg/config_metrics.go#L44)) vs `perapp.MetricsConfig` ([`decfg.go:14`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/071b7180/pkg/export/otel/perapp/decfg.go#L14)) — so `metrics` and `otel_metrics_export` share a definition
- `prom.PrometheusConfig` ([`prom.go:104`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/071b7180/pkg/export/prom/prom.go#L104)) vs `imetrics.PrometheusConfig` ([`iprom.go:24`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/071b7180/pkg/export/imetrics/iprom.go#L24)) — so `prometheus_export` and `internal_metrics.prometheus` share a definition

That produces four kinds of error in the generated reference:

1. **Keys that do not exist.** The `metrics` section documents 14 keys copied from `otel_metrics_export`; the real `metrics` section has only `features`. Likewise `internal_metrics.prometheus` documents 12 keys plus a `native_histogram` subsection, while only `port` and `path` exist ([`iprom.go:24`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/071b7180/pkg/export/imetrics/iprom.go#L24)).
2. **Wrong environment variables.** `internal_metrics.prometheus.port` / `.path` are documented as `OTEL_EBPF_PROMETHEUS_*`; the real variables are `OTEL_EBPF_INTERNAL_METRICS_PROMETHEUS_*` ([`iprom.go:25`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/071b7180/pkg/export/imetrics/iprom.go#L25)). And `otel_metrics_export.features` is documented with `OTEL_EBPF_METRICS_FEATURES`, which that field has no `env` tag for ([`config_metrics.go:78`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/071b7180/pkg/export/otel/otelcfg/config_metrics.go#L78)).
3. **A wrong default.** `prometheus_export.path` is documented as `/internal/metrics`; the real default is `/metrics` ([`config.go:274`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/071b7180/pkg/obi/config.go#L274)). Defaults are written into the shared definition in place, so they leak between the two sections.
4. **A misleading `deprecated` flag.** `metrics.features` is marked `Deprecated: Yes` even though it is the recommended replacement; the flag comes from `otelcfg`'s field comment ([`config_metrics.go:77`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/071b7180/pkg/export/otel/otelcfg/config_metrics.go#L77)) through the shared definition.

Each YAML path resolves to its own definition, so every section documents only its real keys, environment variables and defaults — and a name collision fails schema generation instead of silently producing a wrong document.

### Steps to reproduce

Nothing to run — the committed reference docs already disagree with the config structs. On `main` (commit `071b7180`), and identically in `v0.10.0`, compare:

- the [`metrics` section](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/071b7180/devdocs/config/CONFIG.md?plain=1#L438) with [`perapp.MetricsConfig`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/071b7180/pkg/export/otel/perapp/decfg.go#L14)
- the [`internal_metrics.prometheus` section](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/071b7180/devdocs/config/CONFIG.md?plain=1#L394) with [`imetrics.PrometheusConfig`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/071b7180/pkg/export/imetrics/iprom.go#L24)
- the [`prometheus_export.path` default](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/071b7180/devdocs/config/CONFIG.md?plain=1#L605) with [`DefaultConfig.Prometheus`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/071b7180/pkg/obi/config.go#L274)
- the [`metrics.features` row](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/071b7180/devdocs/config/CONFIG.md?plain=1#L447) marked deprecated with the comment it came from in [`otelcfg.MetricsConfig.DeprFeatures`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/071b7180/pkg/export/otel/otelcfg/config_metrics.go#L78)

### OBI configuration

```yaml

```

### Relevant logs

```shell

```

### Additional context

Suggested direction, in two parts.

**Now — rename the colliding types** so that no two config structs share a bare name. That is enough to fix the generated docs without touching the generator, because its own metadata registries are keyed by bare type name as well. I have this prepared and verified locally and will open a PR shortly, with the details there — including why setting a `Namer` alone is not sufficient, and the naming choices, which I would like feedback on.

**Still needed — a guard** that fails schema generation when two distinct types request the same name. Nothing stops a third collision from landing silently today. That is out of scope for the rename above: doing it properly touches the generator's naming and its metadata registries, so it needs more design and validation than a rename does. Noting it here so it is tracked.

Note that the rename touches exported types, so it is a Go API break for module consumers: allowed in `v0` per `VERSIONING.md`, but a release-note item.

As far as I can tell the change touches no YAML key and no environment variable, so existing user configuration should keep working. If there is a code path or a consumer I have overlooked, I would appreciate you pointing it out.

### Version

v0.10.0

### Kernel version

_No response_

### OS / Distribution

_No response_

### Deployment

Kubernetes

elliehmy:
## Summary

Part of #2898. 
Fixes the two `$defs` collisions described there by renaming the colliding types, and regenerates `devdocs/config`.
Not `Fixes #2898`: the guard that should fail generation on a duplicate name is still needed.

Two commits, separable for review:

1. `perapp.MetricsConfig` → `perapp.GlobalMetricsConfig`, plus the metrics-feature deprecation notices now name the YAML path instead of a Go symbol, since the generator copies those comments verbatim into the reference.
2. `imetrics.PrometheusConfig` → `imetrics.PrometheusEndpointConfig`.

No YAML key and no environment variable changes.

### Feedback wanted on the two names

- **`GlobalMetricsConfig`** matches the package comment's "global and per-service" framing and pairs with the existing `SvcMetricsConfig`. I renamed the newer `perapp` type rather than `otelcfg.MetricsConfig` to keep churn away from the longer-established name — note that `otelcfg` also hardcodes its type name as a log `component` label, so renaming that side would change log output.
- **`PrometheusEndpointConfig`** avoids the stutter that `InternalPrometheusConfig` would create against both the package name (`imetrics`) and the parent type `InternalMetricsConfig`. The type holds only `port` and `path`, so it is effectively endpoint configuration, and the `Prometheus*` prefix keeps it paired with `PrometheusReporter` in the same file.

### Compatibility

Both renamed types are exported under `pkg/`, so this is a source-level break for Go module consumers that construct them directly. `VERSIONING.md` allows that in `v0`, but it is a release-note item and I would rather have maintainers weigh it. The `$defs` key names change too, so any external tooling that references `#/$defs/MetricsConfig` in `config-schema.json` directly would need updating.

## Validation


MrAlias (approved):
cc @NimrodAvni78 @grcevski

NimrodAvni78 (approved):
good catch
LGTM

PR #2909, opened 2026-08-04
feat(config): remove the deprecated `application_jvm` feature alias

pinoOgni:
## Summary

`application_jvm` was kept as a `v0.10` compatibility alias of `application_runtime`, meaning a jvm-named flag enables runtime metrics for every language (including the upcoming nodejs runtime metrics).

This PR removes the deprecated `application_jvm` feature alias and uses `application_runtime` instead.

Also, feature parsing is now strict: an unknown name in `metrics.features` fails at startup with an error listing the valid features, instead of being silently ignored. Without strict parsing, removing the `application_jvm`  alias would have silently disabled metrics for configs still using it.

## Validation

### DIFF
--- a/pkg/export/feature.go
+++ b/pkg/export/feature.go
@@ -5,6 +5,7 @@ package export // import "go.opentelemetry.io/obi/pkg/export"
 
 import (
 	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/invopop/jsonschema"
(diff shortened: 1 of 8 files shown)


pkg/obi/config_test.go
```diff
@@ -510,19 +510,13 @@ jvm_runtime_metrics:
 	assert.Equal(t, 2*time.Second, cfg.JVMRuntimeMetrics.SamplingInterval)
 }
 
-func TestConfig_JVMRuntimeMetricsV010ConfigCompatibility(t *testing.T) {
-	cfg, err := LoadConfig(bytes.NewBufferString(`
+func TestConfig_RemovedApplicationJVMFeatureFailsLoudly(t *testing.T) {
```
marctc: Since we are removing it entirely, do we still need this?

marctc (approved):

pkg/obi/config_test.go
pinoOgni: Yes, here we are testing that the `UnmarshalYAML` error actually aborts `LoadConfig` instead of being swallowed, i.e. the startup failure users will actually see.

I think the function name can be better, something like `TestConfig_UnknownMetricsFeatureFailsStartup` to make that intent explicit and I can reuse `application_jvm`  as the input and add a comment.

pkg/export/feature.go
```diff
@@ -100,16 +96,37 @@ var AppO11yFeatures = FeatureApplicationRED |
...
 	// convert the public data type to the internal representation
 	feats := Features(0)
 	for _, f := range features {
-		feats |= FeatureMapper[f]
+		name := strings.TrimSpace(f)
+		if name == "" {
+			continue
+		}
+		feature, ok := FeatureMapper[name]
+		if !ok {
```
MrAlias, line 123: `application_process` is not in `FeatureMapper`, but five integration compose files still pass it through both `OTEL_EBPF_METRICS_FEATURES` and `OTEL_EBPF_PROMETHEUS_FEATURES`. This branch now aborts startup for those fixtures; the [large-request job](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/30910641333/job/92061453974) logs this exact configuration error and then produces no telemetry.

Could we remove or migrate those entries in this change?

MrAlias (commented):
The strict parser is doing what it advertises, but the repository is not yet internally compatible with that behavior.

pkg/export/feature.go
pinoOgni: Thanks @MrAlias, I thought it was a temporary problem of github and didn't check the logs!

MrAlias (approved):

PR #2921, opened 2026-08-05
fix(otel-detection): don't attribute an ambiguous OTLP export to both signals

### LINKED ISSUE #2765: exclude_otel_instrumented_services: port fallback over-suppresses — a traces-only (or metrics-only) service loses the other signal entirely
## Summary

Follow-up to #2711 / #2712. #2712 fixed the `PeerPort` → `HostPort` comparison so the gRPC port fallback in `exclude_otel_instrumented_services` actually fires. That fix is correct, but it activates a pre-existing design problem that #2711 raised and that was never resolved (the issue was auto-closed by the PR merge):

**When the OTLP `:path` cannot be decoded and detection falls back to port matching, a single OTLP export flags the service for BOTH traces AND metrics — because traces and metrics share the same OTLP port (4317). As a result, a service that exports only one signal via its SDK has OBI suppress the *other* signal too, which nothing else produces → that signal is lost entirely (over-suppression).**

Since #2712 this is now live behavior, not dead code.

### 1. One export span satisfies both predicates

`IsExportMetricsSpan` and `IsExportTracesSpan` each accept a span by *path OR port* (`pkg/appolly/app/request/span.go:2143-2159`):

```go
func (s *Span) IsExportMetricsSpan(...) bool {
    ... return s.isMetricsExportURL() || s.sendsMetricsOnOtelPort(...)
}
func (s *Span) IsExportTracesSpan(...) bool {
    ... return s.isTracesExportURL()  || s.sendsTracesOnOtelPort(...)
}
```

When `:path == "*"` (see §3), the URL matchers return false and both fall through to the port heuristic. But both heuristics converge on the same port (`span.go:2069-2124`):

```
sendsMetricsOnGrpcOtelPort → (no METRICS_ENDPOINT) → sendsOnDefaultGrpcOtelPort → OTLP_ENDPOINT / 4317
sendsTracesOnGrpcOtelPort  → (no TRACES_ENDPOINT)  → sendsOnDefaultGrpcOtelPort → OTLP_ENDPOINT / 4317
```

With a single generic `OTEL_EXPORTER_OTLP_ENDPOINT`, `HostPort == 4317` makes **both** `IsExportMetricsSpan` and `IsExportTracesSpan` return true for the *same* span.

### 2. Sticky flags accumulate across spans

`checkIfExportsOTel` (`pkg/ebpf/common/pids.go:225`) sets one flag per span via `if / else if`, but the flags are set-once/sticky (`EnsureExportsOTel*` in `pkg/appolly/discover/exec/file.go:220`, "returns true if this call flipped the flag"):

```go
if span.IsExportMetricsSpan(...) && fi.EnsureExportsOTelMetrics() {
    pf.reportAvoidedService(fi, "metrics")
} else if span.IsExportTracesSpan(...) && fi.EnsureExportsOTelTraces() {
    pf.reportAvoidedService(fi, "traces")
}
```

Because every export span matches both predicates, the 1st span flips the metrics flag and the 2nd span (metrics already set → short-circuits to the else-if) flips the traces flag. SDKs export both signals periodically, so within seconds **both flags are set.**

### 3. Drop points then suppress both

```go
// RED metrics — pkg/export/otel/metrics.go:896 (and pkg/export/prom/prom.go:938)
return span.Service.Features.AppRED() && !span.Service.ExportsOTelMetrics()
// traces — pkg/export/otel/tracesgen/tracesgen.go:244
return request.IgnoreTraces(span) || span.Service.ExportsOTelTraces() || !acceptSpan(...)
```

Once both flags are set, OBI's own RED metrics and traces for that service are dropped.

## Why `:path` is `"*"` (the trigger)

`pkg/ebpf/common/http2grpc_transform.go:409`:

```go
if path == "" || !validPath.MatchString(path) {
    path = "*"
}
```

`"*"` is a sentinel for "OBI could not decode the path", not malformed traffic. OTLP exporters hold a single long-lived HTTP/2 connection and send an identical `:path` every export, so HPACK compresses it to a dynamic-table index after the first request. OBI attaching mid-stream lacks that dynamic-table state and cannot resolve the index → `"*"`. This is common (and especially frequent with Python `grpcio` exporters), which is exactly why the port fallback exists.

## Reproduction

- Service `currency` exports **traces only** via OTLP gRPC to a generic endpoint `OTEL_EXPORTER_OTLP_ENDPOINT=collector:4317` (no metrics SDK).
- Its trace-export client spans arrive with `Path="*"`, `HostPort=4317`, status UNSET.
- `IsExportTracesSpan` → true (port) ✅ correct.
- `IsExportMetricsSpan` → **also true** (same port) ❌ misclassification.
- `ExportsOTelMetrics` gets set → `metrics.go:896` drops `currency`'s RED metrics.
- Net result: `currency` has **no RED metrics anywhere** — the SDK never sent them and OBI is now suppressing its own. (Symmetric case: a metrics-only service loses traces.)

## Proposed directions

I'd suggest starting with **(1)** as a near-term safeguard against the over-suppression that #2712 activated, and pursuing **(3)** as the long-term fix.

1. **Conservative fallback default (near-term).** When detection relies on the port-only signal (opaque `:path`), do not set both flags from a single match; prefer *not* suppressing when the signal type is ambiguous. In observability, silent data loss is worse than duplication, so the ambiguous case should fail open. This directly contains the regression introduced by #2712 without waiting for a full decoding fix.
2. **Granular suppression.** Suppress only overlapping metric names rather than the whole signal type.
3. **Improve `:path` decoding reliability (long-term).** Track per-connection HPACK dynamic-table state so an indexed `:path` can be resolved, making the port fallback unnecessary and removing the ambiguity at its source.

Happy to bring this to the OBI SIG.

## Environment

- OBI version: `main` (post-#2712)
- Affects: any OTLP gRPC exporter whose `:path` is HPACK-indexed (notably Python `grpcio`)

bourbonkk:
## Summary

Fixes #2765.

When OBI can't decode a gRPC `:path` it falls back to matching the peer port against the process's OTLP endpoint. Traces and metrics share that endpoint, so the fallback made `IsExportMetricsSpan` and `IsExportTracesSpan` both accept the same span. With the sticky per-signal flags, two exports were enough to flag a service for both signals — a service exporting only traces through its SDK had OBI suppress its own RED metrics, which nothing else produced.

The port now counts as evidence only when it identifies one signal: the per-signal `..._ENDPOINT` ports differ, or a per-signal `..._PROTOCOL` rules one out. An ambiguous export sets no flag, so the worst case becomes duplicate telemetry rather than a signal disappearing. Path-based detection is unaffected.

The existing `TestDetectsOTelExport` cases already assert that the two predicates are mutually exclusive (`assert.False` on the opposite predicate), and each port-based case sets the other signal's protocol to `http/protobuf` to keep it unambiguous — so the invariant was intended; only the ambiguous case was uncovered. New cases cover it, plus a `pids.go`-level test that two ambiguous exports set neither flag.

`devdocs/exclude-otel-instrumented-services.md` documents the new fail-open behavior and how to restore suppression (distinct per-signal endpoint ports).

## Validation

`make lint`, `make compile`, and the affected unit tests pass. I also confirmed the new tests fail without the fix.

Per the [Generative AI policy](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/AI-POLICY.md): I used Claude Code to draft the change, tests, and docs; I reviewed and edited all of it, verified the behavior against the code paths described in #2765, and take responsibility for the result.

### DIFF
--- a/devdocs/exclude-otel-instrumented-services.md
+++ b/devdocs/exclude-otel-instrumented-services.md
@@ -71,6 +71,11 @@ as follows:
   `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` / `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`,
   falling back to `OTEL_EXPORTER_OTLP_ENDPOINT`, falling back to
   `default_otlp_grpc_port` (4317).
+- Accept the match only if it fits **one** signal. A single OTLP endpoint carries
+  every signal, so a port that matches traces and metrics equally well identifies
+  neither, and OBI leaves both flags unset. The heuristic still resolves a signal
+  whenever the per-signal `..._ENDPOINT` ports differ or a per-signal
+  `..._PROTOCOL` rules one of them out.
 
 Detection is per-telemetry-type and per-service: a service may be flagged for
 metrics but not traces, or vice versa. Once flagged, the flag is sticky for the
@@ -150,7 +155,24 @@ In practice this means:
   always produce duplicates.
 - The first few requests of a long-lived process are duplicated.
 
-### 2. Suppression is all-or-nothing per service
+### 2. Undecoded gRPC paths on a shared endpoint are never suppressed
+
+OBI cannot always recover a gRPC method: an OTLP exporter holds one long-lived
+HTTP/2 connection and sends an identical `:path` on every export, so HPACK
+compresses it to a dynamic-table index after the first request. OBI attaching
+mid-stream has no dynamic-table state to resolve that index. This is what the
+port fallback exists for, and it is common with Python `grpcio` exporters.
+
+When such a service also sends every signal to one endpoint — the usual
+`OTEL_EXPORTER_OTLP_ENDPOINT` setup — nothing distinguishes a traces export from
+a metrics export, so OBI suppresses neither and duplicates that service's
+telemetry for as long as it runs. This is deliberate: attributing the export to
+both signals would suppress the one the SDK never sends, and no other producer
+would replace it, so an ambiguous signal fails open. Configuring
+`OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` and `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`
+on different ports restores suppression.
+
+### 3. Suppression is all-or-nothing per service
 
 Once a service is flagged, OBI suppresses **every** span / metric it would
 produce for that service, regardless of category. There is no way to say "the
--- a/pkg/appolly/app/request/span.go
+++ b/pkg/appolly/app/request/span.go
@@ -2264,6 +2264,17 @@ func (s *Span) sendsTracesOnOtelPort(defaultOtlpGRPCPort int) bool {
 	}
 }
 
+// A single OTLP endpoint carries every signal, so a port match alone cannot tell
+// which signal a span exports. Attributing an ambiguous export to both would
+// suppress the one the SDK never sends, and nothing else would produce it.
+func (s *Span) sendsOnlyMetricsOnOtelPort(defaultOtlpGRPCPort int) bool {
+	return s.sendsMetricsOnOtelPort(defaultOtlpGRPCPort) && !s.sendsTracesOnOtelPort(defaultOtlpGRPCPort)
+}
+
+func (s *Span) sendsOnlyTracesOnOtelPort(defaultOtlpGRPCPort int) bool {
+	return s.sendsTracesOnOtelPort(defaultOtlpGRPCPort) && !s.sendsMetricsOnOtelPort(defaultOtlpGRPCPort)
+}
+
 func (s *Span) sendsMetricsOnGrpcOtelPort(defaultOtlpGRPCPort int) bool {
 	otlpMetricsProtocol, ok := s.Service.EnvVars[envOTLPMetricsProtocol]
 	if ok && otlpMetricsProtocol != otlpGrpcProtocol {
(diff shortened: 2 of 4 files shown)


devdocs/exclude-otel-instrumented-services.md
```diff
@@ -150,7 +155,24 @@ In practice this means:
   always produce duplicates.
 - The first few requests of a long-lived process are duplicated.
 
-### 2. Suppression is all-or-nothing per service
+### 2. Undecoded gRPC paths on a shared endpoint are never suppressed
```
mmat11, line 158: I'm trying to work around this issue in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2924 by decoding the huffman encoded values -- would that be a complementary workaround to this one?

bourbonkk: Complementary — though maybe not in the direction the question implies.

As far as I can tell the Huffman work in #2924 is scoped to traceparent values(the decode table only covers `[0-9a-f-]`), so `:path` resolution isn't affected, and export detection matches on `Path` via `isTracesExportURL()` / `isMetricsExportURL()`.

The part that does meet this PR is the desync handling. With `MarkUnreliable()` and the `/`-anchored `validPath`, an untrustworthy `:path` now degrades to `*` instead of a plausible-but-wrong value — which is the right call, and it means those requests fall through to the port heuristic. That's exactly the ambiguous case this PR makes safe, so #2924 arguably makes this path more reachable rather than less.

grcevski (approved):
LGTM!

PR #2924, opened 2026-08-05
grpc: decode huffman encoded traceparent values on ingress

mmat11:
## Summary

Also fixes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/discussions/2916

## Validation

### DIFF
--- a/bpf/common/h2_tp_huffman.h
+++ b/bpf/common/h2_tp_huffman.h
@@ -1,0 +1,149 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_helpers.h>
+#include <bpfcore/utils.h>
+
+#include <common/h2_defs.h>
+
+// Huffman traceparent values (RFC 7541 5.2), the encoding gRPC SDKs emit.
+// Every [0-9a-f-] code is 5 or 6 bits, so six bits of lookahead name one character.
+enum {
+    // 55 characters at 5 bits, rounded up to octets
+    k_h2_tp_huff_min = 35,
+    // 55 characters at 6 bits, rounded up to octets
+    k_h2_tp_huff_max = 42,
+    // power of two to keep the index maskable; the zeroed tail keeps lookahead in range
+    k_h2_tp_huff_window = 64,
+    k_h2_tp_huff_mask = k_h2_tp_huff_window - 1,
+    k_h2_tp_huff_lookahead = 6,
+    // set in a table entry whose code is 6 bits wide
+    k_h2_tp_huff_wide = 0x80,
+    // smallest huffman field; scans bounded by the plain sizes would never reach one
+    k_h2_tp_hpack_huff_min = k_hpack_tp_val_offset_huffman + k_h2_tp_huff_min,
+    // widest value offset the scan can hand to the decode program
+    k_h2_tp_huff_at_max = k_hpack_tp_max_scan + k_hpack_tp_val_offset,
+};
+
+// Decode table indexed by 6 bits of lookahead (RFC 7541 Appendix B). A 5-bit code fills
+// two adjacent slots (one per value of the trailing don't-care bit) and stores its plain
+// character; a 6-bit code fills one slot and is flagged k_h2_tp_huff_wide. 0 = no
+// traceparent character starts with these bits; slot 0x3f is the EOS prefix, so EOS
+// rejects here too.
+static const u8 k_h2_tp_huff[64] = {
+    0x30, 0x30, 0x31, 0x31, 0x32, 0x32, 0x61, 0x61, // '0' '1' '2' 'a'
+    0x63, 0x63, 0x65, 0x65, 0x00, 0x00, 0x00, 0x00, // 'c' 'e'
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xad, 0x00, // '-'
+    0x00, 0xb3, 0xb4, 0xb5, 0xb6, 0xb7, 0xb8, 0xb9, // '3'..'9'
+    0x00, 0x00, 0x00, 0xe2, 0xe4, 0xe6, 0x00, 0x00, // 'b' 'd' 'f'
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+};
+
+// Located but not decoded: the decode runs in its own tail-call program
+enum {
+    k_h2_huff_then_finalize = 0,
+    k_h2_huff_then_commit = 1,
+};
+
+typedef struct h2_tp_huff_candidate {
+    u16 at;  // offset of the value within the scanned block, bounded by k_hpack_tp_max_scan
+    u8 len;  // encoded octets, 0 when the scan found none
+    u8 next; // stage to resume, since both server scans can request a decode
+} h2_tp_huff_candidate_t;
+
+enum {
+    // indexed fields commonly alias the plausible length range
+    k_h2_tp_huff_max_candidates = 3,
+};
+
+typedef struct h2_tp_huff_scan {
+    u16 at[k_h2_tp_huff_max_candidates];
+    u8 len[k_h2_tp_huff_max_candidates];
+    u8 count;
+    u8 idx;
+    u8 done;    // scan ran; retries walk the list
+    u16 resume; // candidates at or before this offset were already rejected
+    u8 _pad[2];
+} h2_tp_huff_scan_t;
+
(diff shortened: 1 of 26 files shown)


pkg/ebpf/common/http2grpc_transform.go
```diff
@@ -71,22 +80,59 @@ func getOrInitH2Conn(activeGRPCConnections *lru.Cache[uint64, h2Connection], con
...
+		h2c.hdecRet.MarkUnreliable()
+		return
 	}
 
-	return &v
+	switch {
+	case h2c.lastSeq == 0 || seq <= h2c.lastSeq+1:
+		// baseline, in order, or a duplicate
+	case seq == h2c.lastSeq+2:
+		h2c.pendingSeq = h2c.lastSeq + 1
```
MrAlias: When `seq` jumps from N to N+2 because N+1 was actually lost, this only records `pendingSeq`. `http2FromBuffers` then immediately decodes N+2 with the still-trusted table; if the lost block inserted `x-b` while the decoder still has `x-a` at that dynamic index, N+2 resolves `x-b` as `x-a`.

Could we either mark the table unreliable before decoding this event or buffer it until `pendingSeq` arrives?

bpf/generictracer/protocol_http2.h
```diff
@@ -243,6 +256,16 @@ http2_grpc_end(http2_conn_stream_t *stream, http2_grpc_request_t *prev_info, voi
         bpf_dbg_printk("stream_id = %d", stream->stream_id);
         //dbg_print_http_connection_info(&stream->pid_conn.conn); // commented out since GitHub CI doesn't like this call
 
+        // The ordinal is taken at emission, so user space receives it in ring
+        // buffer order: a gap can only mean a lost event, never reordering.
+        // Incremented before the reserve so a full ring buffer leaves a gap too.
+        http2_conn_info_data_t *h2g =
+            bpf_map_lookup_elem(&ongoing_http2_connections, &stream->pid_conn);
+        if (h2g) {
+            h2g->req_seq++;
```
MrAlias: `req_seq++` is a non-atomic load/add/store on a map value shared across CPUs. Two same-connection request ends can emit the same ordinal; if either event is lost, user space sees no gap and keeps the stale HPACK decoders trusted.

Could we synchronize ordinal allocation—for example, with a spin lock on the map value—so concurrent request ends cannot receive the same sequence number?

bpf/generictracer/hpack_traceparent_parse.h
```diff
@@ -27,19 +29,46 @@ static const u8 k_hex_chars[256] = {
...
+        return 0;
+    }
+
+    const u32 max_pos = data_len - (u32)k_h2_tp_huff_min - 1;
+
+    for (u16 i = 0; i < k_hpack_tp_max_scan && i <= max_pos; i++) {
+        if (h2_tp_huff_len_plausible(data[i])) {
+            huff->at = (u16)(i + 1);
+            huff->len = (u8)(data[i] & (u8)~k_hpack_huffman_flag);
+            return 1;
```
MrAlias: This returns the first byte whose low seven bits are 35–42, but the sweep does not parse HPACK boundaries, so the same bytes can be an earlier indexed field or part of another string. If that candidate fails decoding, the tail-call path never resumes the scan and a valid Huffman traceparent later in the block is ignored.

Could we carry a resume offset and retry after a rejected candidate? A `0xa5` indexed field before the real value reproduces this.

MrAlias (commented):
The expanded traceparent handling is a useful step for gRPC propagation. I found a few edge cases around event sequencing and HPACK candidate selection that could still leave the decoder out of sync or miss a valid context.

bpf/common/h2_tp_huffman.h
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    // RFC 7541 5.2: at most 7 bits of padding, carrying EOS's leading ones
+    const u32 pad = len * 8 - c.bit;
+    if (pad > 7) {
+        return false;
+    }
+    if (pad) {
+        const u8 pad_mask = (u8)((1U << pad) - 1);
+        if ((w[(len - 1) & k_h2_tp_huff_mask] & pad_mask) != pad_mask) {
```
MrAlias: On 6.6, the verifier loses `len`'s lower bound across `bpf_loop` and treats this as a possible `w[-1]` read. The Kernel 6.6-lts check fails here with `invalid access to map value, value_size=64 off=-1 size=1`, so the new HTTP/2 program cannot load on a supported kernel.

Could we re-establish the bound immediately before this read?

bpf/generictracer/hpack_traceparent_parse.h
```diff
@@ -27,19 +29,77 @@ static const u8 k_hex_chars[256] = {
...
+        return 1;
+    }
+
+    const u8 b = c->base[(c->off + i) & (k_kprobes_http2_buf_size - 1)];
+    if (!h2_tp_huff_len_plausible(b)) {
+        return 0;
+    }
+
+    u8 cand = c->scan->count;
+    if (cand >= k_h2_tp_huff_max_candidates) {
```
MrAlias, line 54: This now retries multiple candidates, but the raw-byte sweep still stops after the third plausible length. Three earlier indexed bytes such as `0xa5` fill the list, so a valid Huffman traceparent whose length byte appears fourth is never recorded or decoded. The current decoy test reaches the real value within the three-slot list.

Could we carry a resume offset so rejected candidates do not cap the scan?

bpf/generictracer/protocol_http2.h
```diff
@@ -243,6 +257,23 @@ http2_grpc_end(http2_conn_stream_t *stream, http2_grpc_request_t *prev_info, voi
...
         //dbg_print_http_connection_info(&stream->pid_conn.conn); // commented out since GitHub CI doesn't like this call
 
+        // claimed before the reserve so a dropped event leaves a gap; user
+        // space reorders the one-event ordinal/commit skew; seq 0 = ungated
+        http2_conn_info_data_t *h2g =
+            bpf_map_lookup_elem(&ongoing_http2_connections, &stream->pid_conn);
+        if (h2g) {
+            const u32 base = h2g->req_seq;
+            const u8 claimed = 1;
+            for (u8 i = 1; i <= k_seq_claim_attempts; i++) {
```
MrAlias: The `BPF_NOEXIST` claims prevent duplicates for four concurrent claimers, but five probes can all read the same `base`; the fifth exhausts this loop and keeps `seq == 0`. `gateConnSeq` explicitly bypasses zero, so that event is decoded outside ordinal order. If its ring-buffer reservation then fails, no gap exposes the missing HPACK block and the decoder can keep trusting stale dynamic state.

Could we avoid the ungated fallback and guarantee an ordinal for every claimer?

MrAlias (commented):
The current head still has a supported-kernel load failure, and the bounded fallbacks can bypass the HPACK desynchronization guarantees.

MrAlias (commented):
The verifier and Huffman-scan follow-ups look good. I found two remaining sequencing cases where the HPACK decoder can still get out of sync.

bpf/generictracer/protocol_http2.h
```diff
@@ -243,6 +257,29 @@ http2_grpc_end(http2_conn_stream_t *stream, http2_grpc_request_t *prev_info, voi
         bpf_dbg_printk("stream_id = %d", stream->stream_id);
         //dbg_print_http_connection_info(&stream->pid_conn.conn); // commented out since GitHub CI doesn't like this call
 
+        // claimed before the reserve so a dropped event leaves a gap; user
```
MrAlias: `seq` is assigned in `http2_grpc_end`, so it orders stream completion rather than the HEADERS blocks captured in `http2_grpc_start`. On a multiplexed connection, HEADERS can arrive A then B while B completes first; user space then decodes B against the table from before A and can degrade later method names to `*`.

Could we advance HPACK state in header-arrival order independently of span completion, and cover the inverted-completion case with `h2ConnEncoder`?

bpf/generictracer/protocol_http2.h
```diff
@@ -243,6 +257,29 @@ http2_grpc_end(http2_conn_stream_t *stream, http2_grpc_request_t *prev_info, voi
...
+        // space reorders the one-event ordinal/commit skew
+        http2_conn_info_data_t *h2g =
+            bpf_map_lookup_elem(&ongoing_http2_connections, &stream->pid_conn);
+        if (h2g) {
+            const u8 claimed = 1;
+            prev_info->seq = k_h2_seq_unreliable;
```
MrAlias: The sentinel handles allocation exhaustion only when this event is delivered. If the following `bpf_ringbuf_reserve` fails, no unique ordinal or sentinel reaches user space; the next event can remain contiguous, so `gateConnSeq` keeps the decoders trusted after a dropped HPACK block.

Could we make exhaustion leave a durable missing ordinal or poison state before the reservation?

grcevski (approved):
LGTM! Amazing work.

bpf/generictracer/protocol_http2.h
mmat11: done

bpf/generictracer/protocol_http2.h
mmat11: done -- in a bounded window

PR #2927, opened 2026-08-05
fix(genai): normalize OpenAI Responses content

MrAlias:
## Summary

- normalize Responses text, image, file, audio, and refusal content
- restrict Responses input-item capture to `/v1/responses` endpoints
- encode base64 image data URLs as blob parts instead of remote URIs

## Why

#2845 added Responses input-array capture so tool calls and outputs remain visible in `gen_ai.input.messages`. The shared parser treated every array-valued `input` as Responses items, including embedding batches, and reused Chat Completions decoding for Responses-specific content shapes. This could export embedding arrays outside the message schema and collapse valid multimodal Responses content into raw text.

Responses refusals were also decoded through the `text` field, which emitted an empty content part. Chat Completions image data URLs were represented as remote URIs even though their payload is embedded in the request.

This change keeps endpoint routing and provider-specific content decoding explicit while preserving the existing opt-in content behavior.

## Validation

- `go test ./pkg/appolly/app/request ./pkg/ebpf/common/http`

### DIFF
--- a/pkg/appolly/app/request/genai_normalize_openai.go
+++ b/pkg/appolly/app/request/genai_normalize_openai.go
@@ -8,6 +8,23 @@ import (
 	"strings"
 )
 
+type openAIInputAudio struct {
+	Data   string `json:"data"`
+	Format string `json:"format,omitempty"`
+}
+
+type openAIResponsesContentBlock struct {
+	Type       string            `json:"type"`
+	Text       string            `json:"text,omitempty"`
+	Refusal    string            `json:"refusal,omitempty"`
+	ImageURL   string            `json:"image_url,omitempty"`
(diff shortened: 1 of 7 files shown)


MrAlias:
@NameHaibinZhang PTAL.

NameHaibinZhang:
Nice fix. Routing input-item capture by the /v1/responses path and modeling the Responses content blocks separately from Chat Completions is the right call — the input_image.image_url string-vs-object difference is exactly the kind of thing that quietly collapses multimodal content into raw text. Also good to see the embedding array no longer leaking into gen_ai.input.messages and the refusal content being captured. LGTM. @MrAlias

NameHaibinZhang (approved):

grcevski (approved):
LGTM!

PR #2939, opened 2026-08-06
feat(nodejs): event-loop runtime metrics via injected `perf_hooks` agent

### LINKED ISSUE #2902: Add nodejs runtime metrics
### Component

OBI (eBPF instrumentation)

### Problem

Provide support for nodejs runtime metrics, covering as musch as possible what's defined here https://opentelemetry.io/docs/specs/semconv/runtime/nodejs-metrics/

_No response_

_No response_

### Additional context

_No response_

pinoOgni:
## Summary

This PR adds support for Nodejs runtime metrics https://opentelemetry.io/docs/specs/semconv/runtime/nodejs-metrics/
resolves #2902 

Node has no passive probe path, indeed USDT/DTrace support was [removed](https://github.com/nodejs/node/pull/43652) in Node 19 and V8 internals have no stable ABI, so the only faithful source of the `semconv nodejs.eventloop.*` values is the runtime itself.  We reuse the existing inspector-injected agent as that source: the exported numbers are exactly what `perf_hooks` reports inside the application.

Example:
```
# HELP nodejs_eventloop_utilization_ratio Ratio of time the nodejs event loop was active over the last sampling interval.
# TYPE nodejs_eventloop_utilization_ratio gauge
nodejs_eventloop_utilization_ratio{service_instance_id="obi:1",service_name="nodejs-runtime",service_namespace="integration-test"} 0.017950985290267795
```

Limitations / follow-ups
- Main-thread event loop only (perf_hooks is per-isolate); worker_threads unmeasured; same scope as the OTel Node.js SDK.
- Node.js < 14.10 reports no runtime metrics (agent guard); trace injection is unaffected.
- Injection is single-shot per discovered process; no retry on transient failure (follow-up: retry + failure metric).
- One inspector port per network namespace: among cluster workers only the first binder is injected.
- Delay gauges are attribute-less per semconv:  multiple Node processes of one service instance share those series.
- Fixed 1 s sampling; making it configurable means templating the injected script.
- Injection is gated on the merged feature set of all discovery entries while export is per-service; excluding a Node app via selector drops its metrics but still injects; follow-up PR will make the injector honor per-service features (discussed with @marctc).
- `v8js.*` metrics (GC duration, heap/heap-space memory) are out of scope; the same agent and side channel can carry them in a follow-up.

I tried to reduce the size of the PR using pre-PRs, and this is the best I could manage. I recommend reviewing it commit by commit.

## Validation


grcevski (approved):
LGTM!

One comment I have is that this will add the runtime collection on the nodejs side regardless if the option is enabled or not. The only way I can see this improved is if we split the nodejs agent into two files and what we inject is a merge of the basic + runtime.

I doubt the overhead is justifiable for this complexity, if would be nice to know, but I wouldn't block the PR on it.

pinoOgni:
Hi @grcevski, nice catch and I thought about the mirror case!

I went with a substituted literal instead of splitting the agent into two files, we have the same effect and no merge step. I did 2 commits, so we can decide to adjust or drop them:
- [ec1f661](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2939/commits/ec1f6610bbf74b082d90f41b9f64b600474f4175) handles the traces-only case: the injector substitutes an `RT_ENABLED` literal from the same config gate that sets the `nodejs_runtime_metrics_enabled` bpf constant, so the sampling machinery is never installed when the feature is off. A re-injection with the feature off also tears down what a previous injection installed.
- [5488364](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2939/commits/5488364a2f72e1f420f8b98fc3d8b786940be99c) is the mirror your comment exposed, the metrics-only case: we were installing the full trace-correlation machinery to feed a context map nothing consumes when traces are off. `TRACES_ENABLED` is substituted from the same conditions as `Enabled()`, and the hook handle is now stored so re-injection can disable it (when the agent is injected again into a process where a previous injection already installed things).

pinoOgni:
Note regarding this last commit https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2939/commits/a7296d88cff97b6e61cddb2ac11a1daded33670f : `nodejs_runtime_metrics_enabled ` doubled the `generictracer` verifier cross-product (2^8 to 2^9 program loads), pushing the suite past the 20m CI timeout. 

The constant only gates the decode path in `nodejs.c`, which shares no code with the other constants beyond the debug/traceparent machinery, so crossing it with them verified nothing new. I added a dedicated debug x traceparent x nodejs matrix and I'm working on a new version of the bpf verifier test!

devdocs/runtimes/nodejs.md
```diff
@@ -0,0 +1,102 @@
...
+    - application_runtime
+nodejs:
+  enabled: true
+```
+
+The agent samples at a fixed 1 s interval; there is no sampling-interval
+configuration (unlike `jvm_runtime_metrics.sampling_interval`) because the
+interval is compiled into the injected script.
+
+`nodejs.enabled` defaults to `true`, so the `application_runtime` feature alone
```
marctc: I think this part can be rewritten to avoid mention other runtimes (IMO doesn't help to understand this)

devdocs/runtimes/nodejs.md
```diff
@@ -0,0 +1,102 @@
...
+while Node.js has neither (DTrace support was removed in v19 and V8 internals
+have no stable ABI). Enabling `application_runtime` therefore starts injecting
+the agent into every Node.js process OBI discovers, and each injection is
+logged with `trigger=runtime metrics`. Set `nodejs.enabled: false` to opt out —
+that also disables Node.js trace-context propagation.
+
+## Collection path
+
+Node.js has no USDT probes (DTrace support was removed in Node v19) and V8
+internals have no stable ABI, so neither the JVM nor the Go collection
```
marctc: Same here (LLMs are obsessed to tell you what they are NOT doing 😆 )

devdocs/runtimes/nodejs.md
```diff
@@ -0,0 +1,102 @@
...
+- Injection is single-shot per discovered process; a transient failure at
+  process startup is not retried.
+- The inspector port (9229) is per network namespace, so among processes that
+  share one — `cluster` workers, for instance — only the first to bind it is
+  injected; the others log an attach error and report no metrics.
+- The delay metrics carry no attributes (as specified), so several Node
+  processes of the same service instance write to the same series; only the
+  event-loop time counters are separated per process.
+
+## Snapshot cadence
```
marctc: I think this is already mentioned before, so we could remove it

marctc (commented):
Overall looks good, I added some notes about docs. We discussed in a talk, but IMHO we wouldn't need to add a new flag to enable/disable node js injector if we can write config like this:

```
  discovery:
    instrument:
      - open_ports: 8080          # Java: runtime metrics enabled
        metrics:
          features: [application_runtime]

      - open_ports: 3000          # Node.js: runtime metrics disabled
        metrics:
          features: [application]
```

This can also be handled in a follow-up, but I'm corncerned if we merge it as-is we might potentially need to add deprecation warnings like we had to do with JVM runtime metrics.

pinoOgni:
Hi @marctc as discussed I updated the PR and now the logic is the following:

```
inject         = nodejs.enabled && (traces configured || application_runtime)
TRACES_ENABLED = traces configured
RT_ENABLED     = application_runtime
```
So the already established `nodejs.enabled` flag is used for the injection. If it is `false`, there will be no traces/runtime metrics for nodejs processes, even if traces are configured or `application_runtime` is enabled. Moreover,, OBI now logs a warning for the `application_runtime + nodejs.enabled: false` combination, since it would otherwise silently produce no metrics.

marctc (approved):
great stuff, thanks!

pinoOgni:
Note: I did a rebase after #3005, and all the metrics in `pkg/export/attributes/metric.go` now follow the new pattern. Nothing else changed.

PR #2943, opened 2026-08-06
Harden Config v2 migration

MrAlias:
## Summary

- validate Config v1 migration in the selected standalone or Collector receiver deployment mode
- reject unsupported or lossy settings when Config v2 cannot preserve runtime behavior
- validate generated Config v2 documents and keep deterministic migration golden files with the command tests

## Why

`obi config migrate` should fail closed instead of emitting a document that appears valid but changes runtime behavior. This implementation was split from #2799 so the operator documentation can be reviewed independently.

## Validation

- `go test ./cmd/obi/internal/configcmd`

### DIFF
--- a/cmd/obi/internal/configcmd/configcmd.go
+++ b/cmd/obi/internal/configcmd/configcmd.go
@@ -9,20 +9,29 @@ import (
 	"flag"
 	"fmt"
 	"io"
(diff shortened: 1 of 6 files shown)


grcevski (approved):
LGTM!

PR #2944, opened 2026-08-06
Extend Config v2 parity checks

MrAlias:
## Summary

- map the v1 network flow cache limit to both Config v2 runtime consumers
- exercise the language-detection skip guard through the full parity verification path

## Why

The Config v2 reference should remain aligned with v1 defaults without turning internal language-detection skips into capture exclusions. This checker update was split from #2799 so it can be reviewed as an independent validation change.

## Validation


grcevski (approved):
LGTM!

PR #2945, opened 2026-08-06
Make Collector capture policy explicit

MrAlias:
## Summary

- set `default_action: exclude` in both Config v2 Collector receiver examples
- verify the examples parse as receiver configuration and preserve the intended capture policy

## Why

The examples use include rules to opt matching processes into capture. Making the default exclusion explicit prevents unmatched discovered processes from being captured implicitly. This cleanup was split from #2799 because it is independent of the migration guide.

## Validation

- `go test ./cmd/obi/internal/configcmd`
- parsed both modified files as YAML


grcevski (approved):
LGTM!

PR #2946, opened 2026-08-06
k8s integration: close Weaver live-check gaps

MrAlias:
## Summary

- Fail Weaver validation when either the suite collector or `weavercol` reports new send/enqueue failures. Both counters are baselined before tests and scraped again immediately before the report is stopped; scrape, parse, and counter-reset failures also make the report untrustworthy.
- Route `disable_informers` traces through the tapped collector and require span entities while retaining the existing Jaeger assertion path.
- Retry transient `weavercol` export failures within a bounded 15-second window. The retry budget remains shorter than the final drain, so unresolved exports are counted before validation.
- Replace invalid `deployment.environment.name` precedence sentinels with the documented `development` enum value. Preserve the one-second source metric cadence chosen in #2785.

This follows up on #2785 and closes the remaining false-green paths from #2571. Previously, a non-empty report could pass after either tap hop dropped the only copy of a metric or span shape, and `disable_informers` could pass with metrics even though its traces bypassed Weaver.

## Validation

- `go test ./internal/test/integration/components/kube`
- `go test -c -o /tmp/obi-disable-informers.test ./internal/test/integration/k8s/disable_informers`
- Collector 0.155.0 config validation for `otelcol-config-k8s-weavercol.yml`
- `git diff --check`
- GitHub Actions: 107 checks passed, including Linux/macOS lint and `disable_informers` K8S integration

### DIFF
--- a/internal/test/integration/components/kube/kind.go
+++ b/internal/test/integration/components/kube/kind.go
@@ -71,9 +71,9 @@ type Kind struct {
 	// e2e-framework drops Finish errors from the exit code, so weaver
 	// enforcement has to flow through the Kind itself.
 	weaverFailed bool
-	// Tap export-failure counter captured before tests (waitForWeaverReady),
+	// Tap export-failure counters captured before tests (waitForWeaverReady),
 	// compared at teardown by validateWeaver.
-	tapDropsBaseline    float64
+	tapDropsBaseline    tapDropCounts
 	tapDropsBaselineErr error
 }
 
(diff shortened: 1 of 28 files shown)


NimrodAvni78:
@MrAlias any reason we decrease the interval here from 1s to 10ms
This is done in all tests because it creates many datapoints which both caused events to not reach weaver and the weaver evaluation took a lot longer since it needed to read many more data points

Sorry I just noticed it's a draft.
But it still might be relevant

NimrodAvni78 (approved):
LGTM
there are currently CI failures in main which i dont think this addresses but with a small tweek it can

we can apply the same bounded retry_on_failure (1s/5s/15s, budget < 25s drain) to the first-hop otlp/weavercol exporter too — and if the drops are queue-overflow rather than send-failures, bump the first-hop queue_size

LMK if you want to address this in a separate PR or in this one

mmat11 (approved):

PR #2947, opened 2026-08-06
Fix DNS capture so peer endpoints get named correctly

mifisignal:
## Summary

OBI names the peers an application talks to by watching its DNS lookups. When a
lookup is missed, the peer shows up as a bare IP address instead of a hostname.
These peer names are then reported as "outgoing" instead of the actual name.

I found two separate bugs that make OBI miss a number of DNS lookups. This PR fixes
both, plus a smaller related issue.

## Bug 1: the DNS header was read with the wrong helper

`handle_dns_buf()` parses a buffer that comes from `iovec_memory()`. That
function returns a pointer into a BPF map, which is kernel memory. But the code
read the DNS header out of it with `bpf_probe_read_user()`, the helper for user
memory.

That read always fails. When it fails, the helper zeroes the destination, and
the return value was never checked. So the header came back as all zeroes. The
message body was then copied with a different helper that *does* work, which is
why the captured packet held the real data while the parsed header did not.

Two things break as a result:

1. The transaction ID is always 0. OBI matches a DNS answer to its question
   using the key `(process ID, transaction ID)`. With the ID stuck at 0, every
   lookup from the same process collides on the same key. A lookup that lands
   while a previous one is still cached is thrown away as a duplicate.
2. Every answer is recorded as a question. The flags field is zeroed too, so
   the query/response bit always reads as "query".

Consequently, an app that looks up one name every second had 8 of about
60 lookups reported over a minute. The rest were dropped. After the fix, all
of them are successfully reported.

## Bug 2: unconnected resolver sockets were skipped entirely

musl libc (so every Alpine-based container) and some Java stacks never call
`connect()` on their resolver socket. They use `bind()` plus `sendto()` instead.
These are known as *unconnected sockets*. This [CloudFlare blog post](https://blog.cloudflare.com/everything-you-ever-wanted-to-know-about-udp-sockets-but-were-afraid-to-ask-part-1/)
discusses the subject in detail.

For unconnected sockets, the socket's address pair has no peer in it. `is_dns()`
only looks at that pair to decide whether traffic is DNS, so it never matched,
and OBI missed these lookups completely. For an affected app, no DNS was
captured at all, so no peer could ever be named.

## Bug 3: pairing key collision

Transaction IDs are random, so two lookups from one process can collide by
chance. When that happened, the older entry could suppress the newer answer.
Now a different question name starts a fresh record instead.

## Scope

IPv4 (AF_INET) UDP DNS only. IPv6 DNS transport is still not handled.

### Unit tests

`bpf/tests/bpf_dns_header.c` runs the real `dns.h` against a mocked BPF runtime.
Against the old code it fails with `expected 3178, got 0`, and it also catches
answers being mislabelled as questions. `bpf/tests/bpf_dns_unconnected.c` covers
the unconnected-socket classification.

Including `dns.h` in a host test needs a stub for `common/trace_parent.h`, which
otherwise pulls in the Go tracer and fourteen map definitions. That stub and a
few kernel types are added to the existing `bpf/tests` scaffolding.

`dns_request_transform_test.go` covers the collision guard and the case it
cannot help with, where two lookups share a question name.

### Integration test

`TestSuite_DNSUnconnectedResolver` runs an Alpine workload that resolves a
`valkey` service name and then sends it Redis traffic. The name is served by the
compose network's own DNS, so the test needs no internet access. The Redis
metric acts as a control: if the process was never instrumented, that check
fails first, which separates "DNS capture broke" from "nothing was
instrumented".

The test counts lookups over time rather than just checking one exists. Without
the header fix it sees 8 lookups where it needs 20, and fails.

I ran the full BPF test suite, the Go unit tests, `make lint`, and this
integration suite against `main`, and confirmed the failure direction by
reverting each fix and rebuilding.

## Validation

## AI assistance

Generative AI was used to assist with identifying the bugs and proposing this
change and the accompanying test suite. I personally reviewed every line before
submitting, verified each claim above against captured output, and confirmed the
tests fail without the fixes.

### DIFF
--- a/bpf/generictracer/dns.h
+++ b/bpf/generictracer/dns.h
@@ -217,8 +217,12 @@ static __always_inline u8 handle_dns_buf(const unsigned char *buf,
         return 0;
     }
 
+    // buf is scratch memory from iovec_memory(), so this is a kernel read; a
+    // user read silently zeroes hdr, yielding id 0 and a bogus qr
     struct dnshdr hdr;
-    bpf_probe_read_user(&hdr, sizeof(struct dnshdr), buf);
+    if (bpf_probe_read_kernel(&hdr, sizeof(struct dnshdr), buf) != 0) {
+        return 0;
+    }
 
     const u16 flags = bpf_ntohs(hdr.flags);
     const u8 qr = dns_qr(flags);
@@ -236,7 +240,7 @@ static __always_inline u8 handle_dns_buf(const unsigned char *buf,
         if (req) {
             populate_dns_record(req, p_conn, orig_dport, size, qr, hdr.id, conn_pid);
 
-            bpf_probe_read(req->buf, sizeof(req->buf), buf);
+            bpf_probe_read_kernel(req->buf, sizeof(req->buf), buf);
             bpf_d_printk("sending dns trace [%s]", __FUNCTION__);
             bpf_ringbuf_submit(req, get_flags());
         }
--- a/bpf/generictracer/dns.h
+++ b/bpf/generictracer/dns.h
@@ -84,8 +84,19 @@ static __always_inline u8 dns_ra(u16 f) {
     return (f >> 7) & 0x1;
 }
 
+// RFC 1035 reserved three bits here, but RFC 2535 reassigned the lower two to
+// AD and CD, which DNSSEC-aware resolvers set on ordinary traffic. Only bit 6
+// is still required to be zero, so only bit 6 may be used to reject a message.
 static __always_inline u8 dns_z(u16 f) {
-    return (f >> 4) & 0x7;
+    return (f >> 6) & 0x1;
+}
+
+static __always_inline u8 dns_ad(u16 f) {
+    return (f >> 5) & 0x1;
+}
+
+static __always_inline u8 dns_cd(u16 f) {
+    return (f >> 4) & 0x1;
 }
 
 static __always_inline u8 dns_rcode(u16 f) {
@@ -122,6 +133,62 @@ static __always_inline u16 obi_msg_name_port(struct msghdr *msg) {
         return 0; // IPv6 DNS transport not handled by this fallback
     }
     return bpf_ntohs(sa.sin_port);
+}
+
+// Smallest encodings of a question and of a resource record: the root name
+// (a single zero length octet), then the fixed fields of each.
+enum : u32 {
+    k_dns_min_question_bytes = 1 + 2 + 2,   // name, qtype, qclass
+    k_dns_min_rr_bytes = 1 + 2 + 2 + 4 + 2, // name, type, class, ttl, rdlength
+};
+
+// Whether a datagram is structurally a DNS message. qr alone is one bit, so it
+// says nothing: every payload of at least sizeof(struct dnshdr) has a valid qr,
+// and the socket-identity tier hands this function payloads whose only other
+// evidence is the socket they arrived on. These checks are what separates a DNS
+// message from arbitrary bytes.
+//
+// Deliberately not checked: rcode, because NXDOMAIN and SERVFAIL are lookups
+// worth reporting, and the AD and CD bits, which ordinary DNSSEC-aware
+// resolvers set.
+static __always_inline u8 dns_header_is_plausible(const struct dnshdr *hdr,
+                                                  const u16 flags,
+                                                  const u32 size) {
+    if (dns_opcode(flags) != 0) {
+        return 0; // a lookup is a standard query; IQUERY, STATUS, NOTIFY and UPDATE are not
+    }
+
+    if (dns_z(flags) != 0) {
+        return 0; // still reserved, so a set bit means this is not a DNS header
+    }
+
+    const u16 qdcount = bpf_ntohs(hdr->qdcount);
+    const u16 ancount = bpf_ntohs(hdr->ancount);
+    const u16 nscount = bpf_ntohs(hdr->nscount);
+    const u16 arcount = bpf_ntohs(hdr->arcount);
+
+    if (dns_qr(flags) == k_dns_qr_query) {
+        // a query asks at least one question and answers none of them
+        if (qdcount == 0 || ancount != 0) {
+            return 0;
+        }
+    } else if (qdcount + ancount + nscount + arcount == 0) {
+        // an empty response carries no lookup; an mDNS response may omit the
+        // question, so the sections are counted together rather than separately
+        return 0;
+    }
+
+    // A truncated message describes records that were never sent, so its counts
+    // cannot be held against the bytes on hand.
+    if (dns_tc(flags)) {
+        return 1;
+    }
+
+    const u32 payload = size - sizeof(struct dnshdr);
+    const u32 needed = ((u32)qdcount * k_dns_min_question_bytes) +
+                       (((u32)ancount + nscount + arcount) * k_dns_min_rr_bytes);
+
+    return needed <= payload;
 }
 
 // k_dns_msg_no is a positive answer, not an absence of evidence: the peer is
@@ -362,7 +429,7 @@ static __always_inline u8 handle_dns_buf(const unsigned char *buf,
 
     bpf_d_printk("QR type: %d [%s]", qr, __FUNCTION__);
 
-    if (qr == k_dns_qr_query || qr == k_dns_qr_resp) {
+    if (dns_header_is_plausible(&hdr, flags, size)) {
         conn_pid_t *conn_pid = bpf_map_lookup_elem(&sock_pids, &p_conn->conn);
         if (!conn_pid) {
             bpf_d_printk("can't find connection info for dns call [%s]", __FUNCTION__);
--- a/bpf/generictracer/dns.h
+++ b/bpf/generictracer/dns.h
@@ -20,6 +20,7 @@
 #include <generictracer/k_tracer_defs.h>
 
 #include <maps/sock_pids.h>
+#include <maps/unconn_dns_socks.h>
 
 #include <pid/types/pid_info.h>
 
@@ -136,13 +137,6 @@ static __always_inline u8 is_dns_msg(connection_info_t *conn, struct msghdr *msg
     }
     return 0;
 }
-
-struct {
-    __uint(type, BPF_MAP_TYPE_LRU_HASH);
-    __type(key, u64); // (u64)struct sock *
-    __type(value, u8);
-    __uint(max_entries, 1024);
-} unconn_dns_socks SEC(".maps");
 
 static __always_inline void obi_note_unconn_dns_sock(void *sk) {
     u64 k = (u64)sk;
--- a/bpf/generictracer/dns.h
+++ b/bpf/generictracer/dns.h
@@ -247,10 +247,7 @@ static __always_inline void obi_forget_unconn_dns_sock(void *sk) {
     bpf_map_delete_elem(&unconn_dns_socks, &k);
 }
 
-// Records that sk just sent a DNS query, opening the window in which a nameless
-// answer on this socket may be classified as DNS. Call this only once the sent
-// payload has been parsed as a DNS message, so that a send which never carried
-// DNS — or never left the host — does not open the window.
+// Opens the window in which a nameless answer on sk may be classified as DNS.
 //
 // conn must be the unsorted tuple, so that s_addr/s_port name the local endpoint.
 // The whole record is rewritten unconditionally: one map store, no
@@ -265,6 +262,43 @@ static __always_inline void obi_note_unconn_dns_query(void *sk, const connection
     __builtin_memcpy(fresh.s_addr, conn->s_addr, sizeof(fresh.s_addr));
 
     bpf_map_update_elem(&unconn_dns_socks, &k, &fresh, BPF_ANY);
+}
+
+// Holds a parsed DNS query until udp_sendmsg returns. The payload has been
+// recognised as DNS by this point, but the send may still fail, and a query
+// that never left the host is owed no answer.
+//
+// conn must be the unsorted tuple.
+static __always_inline void
+obi_stage_unconn_dns_query(u64 pid_tid, void *sk, const connection_info_t *conn) {
+    unconn_dns_pending_t pending = {
+        .sk = (u64)sk,
+        .s_port = conn->s_port,
+    };
+    __builtin_memcpy(pending.s_addr, conn->s_addr, sizeof(pending.s_addr));
+
+    bpf_map_update_elem(&unconn_dns_pending, &pid_tid, &pending, BPF_ANY);
+}
+
+static __always_inline void obi_discard_unconn_dns_query(u64 pid_tid) {
+    bpf_map_delete_elem(&unconn_dns_pending, &pid_tid);
+}
+
+// Opens the answer window for a query that was staged and has now been sent.
+// The window is timed from here rather than from the entry probe, so it starts
+// when the datagram actually went out.
+static __always_inline void obi_commit_unconn_dns_query(u64 pid_tid) {
+    const unconn_dns_pending_t *pending = bpf_map_lookup_elem(&unconn_dns_pending, &pid_tid);
+
+    if (!pending) {
+        return;
+    }
+
+    connection_info_t local_conn = {.s_port = pending->s_port};
+    __builtin_memcpy(local_conn.s_addr, pending->s_addr, sizeof(local_conn.s_addr));
+
+    obi_note_unconn_dns_query((void *)pending->sk, &local_conn);
+    obi_discard_unconn_dns_query(pid_tid);
 }
 
 // Reports whether a nameless answer on sk may be attributed to DNS: the socket
--- a/bpf/generictracer/dns.h
+++ b/bpf/generictracer/dns.h
@@ -162,12 +162,11 @@ static __always_inline u8 is_dns_msg(const connection_info_t *conn, struct msghd
     return classify_dns_msg(conn, msg) == k_dns_msg_yes;
 }
 
-// An answer is only expected for as long as a query is outstanding; a resolver
-// that never sees its answer must not leave the socket classified forever.
-enum : u64 { k_unconn_dns_answer_timeout_ns = 10ULL * 1000000000ULL };
-
-// Bounds the state a single socket can accumulate when answers never arrive
-enum : u16 { k_unconn_dns_max_pending = 8 };
+// An answer is only expected for a short while after a query; a resolver that
+// never sees its answer must not leave the socket classified forever. Answers
+// arrive in milliseconds, so this is far longer than needed and still well
+// inside the usual 5s resolver timeout.
+enum : u64 { k_unconn_dns_answer_timeout_ns = 2ULL * 1000000000ULL };
 
 static __always_inline u8 obi_same_local_addr(const unconn_dns_sock_t *state,
                                               const connection_info_t *conn) {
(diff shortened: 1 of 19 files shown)


bpf/generictracer/dns.h
```diff
@@ -97,6 +98,63 @@ static __always_inline u8 is_dns(connection_info_t *conn) {
...
+        if (is_dns_port(p)) {
+            bpf_dbg_printk(
+                "UNCONN_DNS_MSGNAME: classified unconnected UDP DNS via msg_name port=%d", p);
+            return 1;
+        }
+    }
+    return 0;
+}
+
+struct {
```
grcevski: Can we please move this map into the maps folder and import it here? It should also have PIN_INTERNAL flag on it, just in case we ever need to reference it from the other BPF programs.

grcevski (commented):
Thanks for this, looks great!

I made a small pedantic remark about moving the new map into its own file. Also, can you please sign the CLA, so we can approve and merge this.

Thanks.

MrAlias:
@mifisignal please sign the CLA so these changes can be considered.

mifisignal:
@MrAlias I'm working on getting the CLA taken care of. Stand by. :)

bpf/generictracer/dns.h
mifisignal: Done!

mifisignal:
@MrAlias CLA signed now.

bpf/generictracer/k_tracer.c
```diff
@@ -902,7 +907,16 @@ int BPF_KRETPROBE(obi_kretprobe_sock_recvmsg, int copied_len) {
             setup_cp_support_conn_info(&info, false);
             setup_connection_to_pid_mapping(id, &info, orig_dport);
 
-            if (is_dns(&info.conn)) {
+            u8 dns = is_dns_msg(&info.conn, (struct msghdr *)args->msg_ptr);
+            // msg_name did not classify it, but the answer arrived on a socket
+            // seen sending an unconnected UDP DNS query
+            if (!dns && orig_dport == 0 && obi_is_unconn_dns_sock(sock_ptr)) {
```
MrAlias: While this entry remains in the LRU map, one unconnected DNS query makes the fallback treat any later receive on that socket as DNS whenever the tuple has no destination port. That includes a live multi-peer UDP socket whose `msg_name` explicitly names a non-DNS port; the entry can also outlive a closed socket, so a later socket at the same `struct sock *` address inherits the classification. Both paths can send unrelated payloads into `handle_dns_buf()` and produce false DNS telemetry when the payload parses.

Could we make an explicit non-DNS `msg_name` veto the cache and scope fallback state to outstanding DNS use and the socket lifetime instead of retaining raw-pointer membership?

MrAlias (commented):
The DNS header-read correction looks sound, but the receive-side fallback needs tighter scoping.

bpf/generictracer/k_tracer.c
mifisignal: Good catch on both paths. Fixed in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/commit/3f73ef2e1e7d501a804662dcca9fea8b36e3377c.

Classification is now three-valued rather than boolean. An explicit non-DNS peer — from `msg_name`, or from a tuple that names a real destination port — is a positive "not DNS" that deletes the socket's entry, so the multi-peer case you describe retires the classification instead of being overridden by it. Only an unclassifiable receive reaches the fallback.

For lifetime scoping, the map value went from `u8` to a struct holding the local endpoint, an outstanding query count, and the last query timestamp. The receive path claims an answer instead of testing membership: it requires the entry's local address and port to match the socket's current ones, requires a nonzero outstanding count, and requires arrival within 10s. A claim decrements the count and deletes the entry at zero, so the state exists only while a query is actually outstanding. Outstanding queries are capped at 8 so a socket whose answers never arrive can't accumulate credit.

One case remains: a socket freed and replaced at the same `struct sock *` address that rebinds the identical local address and port, within the timeout, while a query is outstanding. Distinguishing that needs a real socket-lifetime identifier — `bpf_get_socket_cookie()` on a `struct sock *` isn't available to kprobes at the 4.18/5.8 floor, and adding a close probe seemed beyond this PR's scope. Happy to take that on if you'd rather close it properly here.

The 10s timeout and the cap of 8 are judgment calls; say the word if you'd prefer different values.

Unit tests in `bpf/tests/bpf_dns_unconnected.c` cover the veto, single-claim semantics, the recycled-pointer and differing-address rejections, both sides of the timeout boundary, and the cap. `TestSuite_DNSUnconnectedResolver` passes locally; leaving the rest of the integration sweep to CI.

grcevski (approved):
LGTM!

grcevski:
I re-ran the tests but they will not succeed. The issue is that with the code changes we now go over the instruction limit for `int BPF_KPROBE(obi_kprobe_udp_sendmsg, struct sock *sk, struct msghdr *msg, size_t len) { @ k_tracer.c:238`

It says:

```
2026-08-17T23:23:28.3163276Z             	; bpf_clamp_umax(iov_size, k_iovec_max_len); @ iov_iter.h:113
2026-08-17T23:23:28.3181619Z             	2272: (b5) if r2 <= 0x2000 goto pc+1          ; R2=0
2026-08-17T23:23:28.3199807Z             	; if (tot_len + iov_size > max_len) { @ iov_iter.h:117
2026-08-17T23:23:28.3217363Z             	2274: (bf) r7 = r2                    ; R2=0 R7=0
2026-08-17T23:23:28.3236546Z             	2275: (0f) r7 += r6                   ; R6=8192 R7=8192
2026-08-17T23:23:28.3245153Z             	2276: (bf) r1 = r7
2026-08-17T23:23:28.3262503Z             	BPF program is too large. Processed 1000001 insn
2026-08-17T23:23:28.3297410Z             	processed 1000001 insns (limit 1000000) max_states_per_insn 34 total_states 27057 peak_states 918 mark_read 0
```

[Full output here](https://productionresultssa1.blob.core.windows.net/actions-results/6f8b4fe7-2af0-4fae-9f36-abe3ca5ccbd3/workflow-job-run-bfdb90da-e9bc-5572-bd63-e2050dd79901/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-08-18T18%3A30%3A44Z&sig=N7zZoekc2vHcJ0rN6Tt74szEPLKvcgdrn970RDwpggY%3D&ske=2026-08-18T22%3A11%3A26Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-08-18T18%3A11%3A26Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-08-18T18%3A20%3A39Z&sv=2025-11-05)

Let me know if you need help with resolving the size issue. One typical way we do this is by reducing loop counts or by splitting the udp helper into two that tail call each other.

mifisignal:
@grcevski is this for a particular kernel version I neglected to test against?

bpf/generictracer/k_tracer.c
```diff
@@ -253,7 +253,19 @@ int BPF_KPROBE(obi_kprobe_udp_sendmsg, struct sock *sk, struct msghdr *msg, size
...
+        // this socket is being used for something else, so it must not stay
+        // classified by an earlier DNS query
+        if (dns == k_dns_msg_no) {
+            obi_forget_unconn_dns_sock(sk);
+        }
+
+        if (dns == k_dns_msg_yes) {
+            // the answer is not classifiable from msg_name, so key on the socket
+            if (orig_dport == 0) {
+                obi_note_unconn_dns_query(sk, &s_args.p_conn.conn);
```
MrAlias: These transitions update socket-wide state from a single datagram before its outcome is known. A DNS send records credit before `udp_sendmsg` returns, so a failed route or send leaves a query that never existed; conversely, a non-DNS send deletes all credit even if a prior DNS query is still in flight. On a multi-peer socket, either path disconnects the fallback state from the actual outstanding work.

Could we stage this classification and commit the state after a successful send while preserving unrelated pending DNS queries? Please cover both a failed DNS send and `DNS query -> non-DNS send -> nameless DNS answer`.

bpf/generictracer/k_tracer.c
```diff
@@ -897,12 +910,37 @@ int BPF_KRETPROBE(obi_kretprobe_sock_recvmsg, int copied_len) {
...
+            // an explicit non-DNS peer retires the socket's classification
+            if (dns_class == k_dns_msg_no) {
+                obi_forget_unconn_dns_sock(sock_ptr);
+            }
+
+            u8 dns = dns_class == k_dns_msg_yes;
+
+            // msg_name did not classify it, but the answer arrived on a socket
+            // with an outstanding unconnected UDP DNS query
+            if (dns_class == k_dns_msg_unknown && orig_dport == 0 &&
```
MrAlias, line 1024: The receive-side transition has the opposite mismatch: a known DNS `msg_name` returns `k_dns_msg_yes` and never consumes credit, while an unknown peer consumes credit here before the iovec is readable or the payload is parsed. Known answers can therefore leave stale credit for unrelated traffic, and a short or unreadable datagram can consume the only credit before the real answer arrives.

Could we separate fallback eligibility from consumption and retire one pending query only after the buffer has been read and parsed as a DNS response, regardless of whether its peer was known or unknown? A combined test for both paths would pin the lifecycle.

bpf/generictracer/dns.h
```diff
@@ -93,10 +96,147 @@ static __always_inline u8 is_dns_port(u16 port) {
...
+static __always_inline void obi_note_unconn_dns_query(void *sk, const connection_info_t *conn) {
+    const u64 k = (u64)sk;
+
+    unconn_dns_sock_t *state = bpf_map_lookup_elem(&unconn_dns_socks, &k);
+
+    if (state && obi_same_local_addr(state, conn)) {
+        if (state->pending < k_unconn_dns_max_pending) {
+            state->pending++;
+        }
+        state->last_query_ns = bpf_ktime_get_ns();
```
MrAlias: `pending` is both bounded and mutated with plain shared-value accesses. Nine live queries are reduced to eight credits by this branch, so after eight nameless replies the ninth remains unclassified; with multiple CPUs, lost increments or a decrement/delete race can drop credit even below that cap.

Could we make the per-socket transition race-safe and represent saturation without treating overflowed live queries as absent? The current sequential cap test codifies the loss instead of covering these interleavings.

MrAlias (commented):
The fallback-state redesign narrows the stale-socket window, but its pending credit is still not a reliable representation of outstanding DNS work. The kernel verifier reruns also remain pending after the reported instruction-limit failure.

mifisignal:
Thanks both — this addresses the state-machine review and the instruction-limit failure.

### The pending counter is gone

All three of the review points traced back to the outstanding-query count, so I
removed it. A socket now records only its local endpoint and the time of its
last DNS query. Those two fields define an eligibility window: while the endpoint
still matches and the query is recent, an answer arriving on that socket with no
peer named is classified as DNS. Nothing is consumed, so the window ends only by
expiring or by being retired when the socket is seen carrying something else.

Updates include:

- **Send-side transitions.** The window is opened only after the sent payload has been
  parsed as a DNS message, so a send that failed or never carried DNS opens nothing. There
  is no credit left for an unrelated send to delete.
- **Receive-side transitions.** Eligibility and consumption are now separate because
  nothing is consumed. A known-peer answer cannot leave state behind, and a short or
  unreadable datagram cannot spend what the real answer needs.
- **Races and saturation.** `obi_note_unconn_dns_query` is now a single unconditional map
  store. There's no read-modify-write cycle, so no lost increments and no cap to saturate.

Part of the design is driven by the fact that musl issues A and AAAA in parallel on
one socket. Retiring the window per answer would leave the second answer unclassified, so
consuming credit per answer is the wrong model, regardless of how the counter is made safe.

The window is now 2s, down from 10s. It only has to span the query-to-answer round trip,
which is milliseconds against a local or in-cluster resolver, and an answer arriving much
later is already near the resolver's own timeout — musl defaults to 5s. A shorter window
narrows the exposure above, at the cost of missing unusually slow answers. The value is a 
judgment call; say the word if you would prefer a different value.

### Socket lifetime

The map is keyed by the `struct sock *` address, which identifies a socket only while it is
alive: once freed, the kernel can hand the same address to an unrelated socket. The recorded
local endpoint rejects a replacement that binds something different, but a replacement
binding the same address and port inherited the classification, because nothing observed the
original socket going away.

A kprobe on `udp_destroy_sock` now drops the entry, so the key means something only for as
long as the socket it names is alive. `bpf_get_socket_cookie()` on a `struct sock *` would
have been the other route, but it requires 5.12, above OBI's 5.8 floor and the RHEL 4.18
backports.

The probe is registered `Required: false`. If the symbol is ever unavailable, the entry
still expires on its own, so the tracer loses the lifetime bound rather than failing to
start.

### Tests

`bpf/tests/bpf_dns_unconnected.c` covers the lifecycle. Two of the new cases fail against
the previous design: parallel answers (only the first was classified) and a receive that
never parsed (it spent the window). The third, window refresh, passes against both and is
a behavioural pin rather than a regression test.

There is also a case for the recycled pointer at an identical endpoint. The existing
recycled-pointer test used a different endpoint, which the endpoint check already rejected,
so it never reached the gap the close probe fills.

`TestSuite_DNSUnconnectedResolver` gains a case for the false-positive class. The workload
pairs each lookup with a decoy exchange on an unconnected socket: a DNS-shaped payload, a
non-DNS peer port, and a reply taken with `recv()` so the receive carries no peer. The
suite requires every reported lookup to name the host the workload actually resolves. I
confirmed it fails when that expectation is flipped.

The suite also covers the risk the close probe introduces. Retiring entries too eagerly
would drop answers, and `every DNS lookup is counted` still passes with the probe attached.

### Verifier

@grcevski — reproduced and fixed. Running the LVH kernels the workflow uses, the previous
head fails on `bpf-next` with the numbers you posted (`Processed 1000001 insn`,
`max_states_per_insn 34`, `total_states 27057`). Both `bpf` and `bpf-next` now pass, along
with 5.10, 5.15, 6.1, 6.6, 6.12, 6.18, the three RHEL kernels and both AL2023 kernels.
`host-arm64` I have left to CI.

No tail-call split was needed. The bookkeeping no longer sits on the path that precedes the
iovec loop, which is what was multiplying verifier states.

grcevski:
Nice! We swiched the BPF CI verifier tests to be on demand because of a recent change that made them slow to run. I'll trigger them manually.

grcevski:
I can confirm the BPF verifier errors are gone.

bpf/generictracer/k_tracer.c
MrAlias: The current head still leaves the socket-wide state disconnected from successful outstanding DNS work:

- `obi_note_unconn_dns_query()` runs from the `udp_sendmsg` entry probe, so a failed DNS send can open the two-second window and let a later nameless, well-formed DNS datagram be reported even though no query was sent.
- An explicit non-DNS send deletes the window while an earlier DNS query may still be in flight, so its later nameless answer is missed.
- The receive path has the same loss: an explicit non-DNS receive deletes the window before a later nameless DNS answer arrives.

The unit test currently codifies the send-side deletion rather than requiring it to preserve eligibility. The integration decoy uses a separate socket only after `resolve_and_ping()` finishes, so there is no regression coverage requiring either interleaving to preserve the DNS window.

Could we stage and commit the state only after a successful send, and keep outstanding DNS eligibility independent of unrelated sends or receives? Please cover a failed DNS send, `DNS query -> non-DNS send -> nameless DNS answer`, and the symmetric receive sequence.

mifisignal: All three are fixed.

The query is now staged at `udp_sendmsg` entry, and a new kretprobe records it
only after the send returns bytes. A send that fails no longer opens the window.
An open window is what lets the next nameless datagram on that socket count as
an answer. This eliminates the risk that a failed send could produce a lookup
that never happened.

Both `k_dns_msg_no` deletes are gone. An unrelated send or receive shouldn't
impact a query that is already in flight, so clearing the window on one threw
away its answer. The answer timeout and `udp_destroy_sock` already limit how
long the window lasts, which is all the early delete was really doing.

The previous unit test checked that a non-DNS send cleared the window. It now
checks that an outstanding query is still eligible afterward. There are also new
tests for a failed send and for both interleavings, in
`bpf/tests/bpf_dns_unconnected.c`.

For the integration tests I build both sequences by hand on one unconnected
socket instead of using musl's. musl fills in `msg_name` when it reads an answer,
so its lookups get classified by the peer port and never reach the socket window
at all. The window is there for clients that read the answer with no peer. The
old decoy also runs on its own socket, and the window is per socket, so it could
never have caught either interleaving. So both sequences read the answer with
`recv()`, which leaves `msg_name` empty. A small responder in its own container
sends the answers back after 500ms. Without that delay, the answer and the
interleaved datagram race each other for the receive queue, and the receive-side
test needs the interleaved one to land first.

Two more things in the branch. `9ac667ed` guards the `udp_destroy_sock` kprobe.
Every kprobe in a subsystem that attaches uprobes has to disable preemption,
because on 6.13 and newer it shares a per-CPU stack with the uprobe programs and
can corrupt it otherwise. `lint-preempt-guard` catches it.

`9a63563b` is a ring buffer fix rather than a DNS fix, and perhaps belongs in
its own PR, but it matters here. The flush that wakes a stalled reader was
suppressing itself on every other tick, so records could sit undelivered for two
flush intervals. That is longer than the DNS request timeout, so on a
lightly-loaded service, a query and its answer reached the pairing stage more
than a timeout apart. The query was then reported as an unanswered `Refused`
lookup, and the answer opened a second span whose duration was meaningless.

One thing to note: Dropping the send-side delete makes the false-positive window
a little wider. After a query and an unrelated send, any nameless datagram that
happens to look like a DNS header gets reported as a lookup, under whatever
question name its bytes happen to spell. `fd1cd665` makes up for that.
`handle_dns_buf` only checked `qr`, which is a single bit, so any payload of
twelve bytes or more counted as a lookup. It now checks the opcode, the reserved
bit, and whether the section counts could fit in the datagram. You can drop that
commit on its own, but if you do, I think the send-side delete needs a second
look. The two go together.

The other thing is that the query event still comes from the entry probe, so a
send that fails still goes into the pairing cache. On musl this works out. The
retry to the next nameserver uses the same transaction id, so the answer pairs
with the first entry. On glibc the common failure happens at `connect()` and
never reaches `udp_sendmsg` at all. What is left is that a lookup which never
succeeds gets reported as `Refused`. That is a server error code for what was
really a local failure. Fixing it means holding the parsed event per thread and
sending it from the kretprobe, and `handle_dns_buf` is shared with the connected
path, so I left it alone.

mifisignal:
@MrAlias ready for followup review

bpf/generictracer/dns.h
```diff
@@ -93,10 +107,230 @@ static __always_inline u8 is_dns_port(u16 port) {
...
+    }
+
+    const u16 qdcount = bpf_ntohs(hdr->qdcount);
+    const u16 ancount = bpf_ntohs(hdr->ancount);
+    const u16 nscount = bpf_ntohs(hdr->nscount);
+    const u16 arcount = bpf_ntohs(hdr->arcount);
+
+    if (dns_qr(flags) == k_dns_qr_query) {
+        // a query asks at least one question and answers none of them
+        if (qdcount == 0 || ancount != 0) {
```
MrAlias: `ANCOUNT != 0` is valid in mDNS queries: [RFC 6762 §7.1](https://www.rfc-editor.org/rfc/rfc6762.html#section-7.1) uses the Answer Section for known-answer suppression, and [§7.2](https://www.rfc-editor.org/rfc/rfc6762.html#section-7.2) continuation packets may also have `QDCOUNT == 0`. Since this code explicitly recognizes port 5353, could we preserve those forms here and cover them in the header tests?

MrAlias (approved):
The fallback lifecycle changes address my earlier concerns. I left one non-blocking mDNS compatibility question in the new header validation.

bpf/generictracer/dns.h
mifisignal: Both forms are preserved now.

 `dns_header_is_plausible` now takes an `mdns` flag. When it's set, I skip the rule that a query must ask a question and answer none. That lets §7.1  known-answer queries and §7.2 continuation packets through. The message still has to describe at least one record, and the section counts still have to fit the bytes on hand.

I matched mDNS on both ends being 5353, not just one. Anything can bind 5353,  and if a site widens `ip_local_port_range` below 5354 at the low end (e.g. 1024:65535), the kernel can hand it out as an ephemeral port. When the socket has no peer in its tuple, I read the peer from `msg_name`. If I can't prove it's mDNS, it falls back to the unicast rule.

In general, though, OBI doesn't support mDNS, and so I'm hesitant to do much more work related to it. I don't think I've ever seen it in a customer environment and no cloud provider that I know of supports it. Until we see customer demand for it, I don't think the juice is worth the squeeze. You?

PR #2949, opened 2026-08-06
Enable the eBPF PID filter before injecting the Java agent

mifisignal:
## Summary

When OBI discovers a new JVM, it tries to inject the Java agent into it. That
injection currently runs before OBI enables the eBPF PID filter for the process.
Until that filter is enabled, none of the process's traffic is captured.

Attaching to a JVM is slow. HotSpot dynamic attach can take up to
`OTEL_EBPF_JAVAAGENT_ATTACH_TIMEOUT` (10s by default) on a JVM that is busy
starting up. During that entire period, a freshly-discovered JVM's own startup
traffic is dropped. If the attach never succeeds, nothing from that process is
captured until the timeout expires.

Capturing events as quickly as possible during startup is important,
because applications often make connections to dependencies immediately at
launch.

This PR enables the PID first and moves the injection after it, into a
goroutine, so discovery no longer waits on the attach.

## Related changes

Moving the injection into a goroutine left `*ebpf.Instrumentable` shared without
synchronization between the injector and the downstream consumers it had just
been sent to. Nothing mutates the fields the injector reads today, so there
would have been no race now, but nothing would have stopped a future change
from introducing one.

The injector only ever needed three values: the instrumentable type, the pid,
and the process's `TMPDIR`. Those are now copied into a `Target` at the call
site and passed by value, avoiding a future possible race.

## Impact on shutdown behavior

In-flight injections are drained with a `WaitGroup` at shutdown. Because
`JavaInjector.NewExecutable` builds its own timeout internally and cannot be
cancelled, shutdown can wait for an injection to time out. There is a `TODO` in
the code describing what threading a context through the injector would take.
I did not do that here to keep the change small, but I am happy to if reviewers
would rather it be fixed now.

## Testing

`TestSuite_JavaDiscoveryEarlyTraffic` runs a JVM started with
`-XX:+DisableAttachMechanism` that sends Redis traffic from the moment it
starts. That flag makes the slow attach deterministic instead of depending on
machine load: the JVM never creates the attach socket, so the injection runs
until the timeout. The compose file sets that timeout to 90s, well past the
60s assertion window.

The suite asserts the workload's Redis operations are recorded:

- with this change: captured in about 7 seconds
- with only the reordering reverted: nothing captured, the assertion fails
  after the full 60 seconds

I also ran `go build`, `go vet`, `make lint`, and `go test -race` on the
discovery package.

The disabled attach mechanism is a stand-in for a busy-starting JVM, which
cannot be reproduced reliably in CI; it drives the same code path for the same
reason, but it is not the literal production case. Separately,
`TestJavaInjector_CopyAgent/error_when_target_directory_not_writable` fails when
the test runs as root, because root ignores directory permissions. That is
pre-existing and not related to this change.

## Validation

## AI assistance

Generative AI was used to identify the issue and assist with this change and its
test suite. I reviewed every line before submitting, verified each claim above
against captured output, and confirmed the test fails without the fix.

### DIFF
--- a/pkg/appolly/discover/attacher.go
+++ b/pkg/appolly/discover/attacher.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"log/slog"
 	"slices"
+	"sync"
 	"time"
 
 	"github.com/cilium/ebpf/link"
@@ -126,6 +127,15 @@ func (ta *traceAttacher) attacherLoop(_ context.Context) (swarm.RunFunc, error) 
 	in := ta.InputInstrumentables.Subscribe(msg.SubscriberName("traceAttacher"))
 	return func(ctx context.Context) {
 		defer ta.OutputTracerEvents.Close()
+		// TODO: cancel an in-flight injection on shutdown (ctx cancellation) or
+		// when a concurrent EventDeleted arrives for the same PID, rather than
+		// only waiting for its internal timeout to elapse. That requires
+		// threading a context.Context through JavaInjector.NewExecutable
+		// (currently it takes only *ebpf.Instrumentable and derives its own
+		// context.Background()-rooted timeout), plus a per-PID cancel-func
+		// registry keyed on the executable key.
+		var javaInjections sync.WaitGroup
+		defer javaInjections.Wait()
 		swarms.ForEachInput(ctx, in, ta.log.Debug, func(instrumentables []Event[ebpf.Instrumentable]) {
 			for _, instr := range instrumentables {
 				ta.log.Debug("Instrumentable", "created", instr.Type, "type", instr.Obj.Type,
@@ -134,15 +144,24 @@ func (ta *traceAttacher) attacherLoop(_ context.Context) (swarm.RunFunc, error) 
 				case EventCreated:
 					ta.resolveServiceMetadata(&instr.Obj)
 					ta.nodeInjector.NewExecutable(&instr.Obj)
-					if ta.javaInjector != nil {
-						if err := ta.javaInjector.NewExecutable(&instr.Obj); err != nil {
-							ta.log.Warn("unable to attach java agent to process, Java TLS telemetry will not work", "pid", instr.Obj.FileInfo.Pid(), "error", err)
-						}
-					}
 
 					ta.processInstances.Inc(executableKey(instr.Obj.FileInfo))
 					if ok := ta.getTracer(&instr.Obj); ok {
 						ta.OutputTracerEvents.Send(Event[*ebpf.Instrumentable]{Type: EventCreated, Obj: &instr.Obj})
+					}
+
+					// Injection blocks for up to the Java attach timeout, so it runs
+					// after the PID is allowed through the eBPF filter, and off the
+					// discovery loop.
+					if ta.javaInjector != nil {
+						javaObj := &instr.Obj
+						javaInjections.Add(1)
+						go func() {
+							defer javaInjections.Done()
+							if err := ta.javaInjector.NewExecutable(javaObj); err != nil {
+								ta.log.Warn("unable to attach java agent to process, Java TLS telemetry will not work", "pid", javaObj.FileInfo.Pid(), "error", err)
+							}
+						}()
 					}
 
 					if instr.Obj.FileInfo.ELF() != nil {
--- a/pkg/appolly/discover/attacher.go
+++ b/pkg/appolly/discover/attacher.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"log/slog"
 	"slices"
-	"sync"
 	"time"
 
 	"github.com/cilium/ebpf/link"
@@ -127,14 +126,14 @@ func (ta *traceAttacher) attacherLoop(_ context.Context) (swarm.RunFunc, error) 
 	in := ta.InputInstrumentables.Subscribe(msg.SubscriberName("traceAttacher"))
 	return func(ctx context.Context) {
 		defer ta.OutputTracerEvents.Close()
-		// TODO: cancel an in-flight injection on shutdown (ctx cancellation) or
-		// when a concurrent EventDeleted arrives for the same PID, rather than
-		// only waiting for its internal timeout to elapse. That requires
-		// threading a context.Context through JavaInjector.NewExecutable
-		// (which currently derives its own context.Background()-rooted
-		// timeout), plus a per-PID cancel-func registry.
-		var javaInjections sync.WaitGroup
-		defer javaInjections.Wait()
+
+		var javaInjections *javaInjectionQueue
+		if ta.javaInjector != nil {
+			javaInjections = newJavaInjectionQueue(ta.log, ta.javaInjector.NewExecutable)
+			javaInjections.start(ctx)
+			defer javaInjections.wait()
+		}
+
 		swarms.ForEachInput(ctx, in, ta.log.Debug, func(instrumentables []Event[ebpf.Instrumentable]) {
 			for _, instr := range instrumentables {
 				ta.log.Debug("Instrumentable", "created", instr.Type, "type", instr.Obj.Type,
@@ -149,19 +148,13 @@ func (ta *traceAttacher) attacherLoop(_ context.Context) (swarm.RunFunc, error) 
 						ta.OutputTracerEvents.Send(Event[*ebpf.Instrumentable]{Type: EventCreated, Obj: &instr.Obj})
 					}
 
-					// Injection blocks for up to the Java attach timeout, so it runs
-					// after the PID is allowed through the eBPF filter, and off the
-					// discovery loop. The target is copied out here so the goroutine
-					// does not share instr.Obj with the consumers it was just sent to.
-					if ta.javaInjector != nil {
-						target := javaagent.InjectionTargetFrom(&instr.Obj)
-						javaInjections.Add(1)
-						go func() {
-							defer javaInjections.Done()
-							if err := ta.javaInjector.NewExecutable(target); err != nil {
-								ta.log.Warn("unable to attach java agent to process, Java TLS telemetry will not work", "pid", target.Pid, "error", err)
-							}
-						}()
+					// Injection blocks for up to the Java attach timeout, so it is
+					// queued after the PID is allowed through the eBPF filter, and
+					// runs off the discovery loop. The target is copied out here so
+					// the injection does not share instr.Obj with the consumers it
+					// was just sent to.
+					if javaInjections != nil {
+						javaInjections.enqueue(javaagent.InjectionTargetFrom(&instr.Obj))
 					}
 
 					if instr.Obj.FileInfo.ELF() != nil {
--- a/pkg/appolly/discover/attacher.go
+++ b/pkg/appolly/discover/attacher.go
@@ -131,9 +131,8 @@ func (ta *traceAttacher) attacherLoop(_ context.Context) (swarm.RunFunc, error) 
 		// when a concurrent EventDeleted arrives for the same PID, rather than
 		// only waiting for its internal timeout to elapse. That requires
 		// threading a context.Context through JavaInjector.NewExecutable
-		// (currently it takes only *ebpf.Instrumentable and derives its own
-		// context.Background()-rooted timeout), plus a per-PID cancel-func
-		// registry keyed on the executable key.
+		// (which currently derives its own context.Background()-rooted
+		// timeout), plus a per-PID cancel-func registry.
 		var javaInjections sync.WaitGroup
 		defer javaInjections.Wait()
 		swarms.ForEachInput(ctx, in, ta.log.Debug, func(instrumentables []Event[ebpf.Instrumentable]) {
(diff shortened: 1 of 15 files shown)


MrAlias:
@mifisignal please sign the CLA so these changes can be considered.

pkg/appolly/discover/attacher.go
```diff
@@ -124,24 +125,42 @@ func (ta *traceAttacher) attacherLoop(_ context.Context) (swarm.RunFunc, error)
...
 					ta.processInstances.Inc(executableKey(instr.Obj.FileInfo))
 					if ok := ta.getTracer(&instr.Obj); ok {
 						ta.OutputTracerEvents.Send(Event[*ebpf.Instrumentable]{Type: EventCreated, Obj: &instr.Obj})
 					}
 
+					// Injection blocks for up to the Java attach timeout, so it runs
+					// after the PID is allowed through the eBPF filter, and off the
+					// discovery loop. The target is copied out here so the goroutine
+					// does not share instr.Obj with the consumers it was just sent to.
+					if ta.javaInjector != nil {
```
grcevski: If hotspot is slow to respond now we will not be instrumenting other applications, correct?

mifisignal: @grcevski can you rephrase the question? Not sure I understand.

grcevski: What I was asking is, does this change the flow of us blocking the instrumentation pipeline, but I don't think it does. I think I was confused by the tracking of java injections. Can you explain better why do you need to count and wait on them?

pkg/internal/java/target.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"go.opentelemetry.io/obi/pkg/appolly/app"
+	"go.opentelemetry.io/obi/pkg/appolly/app/svc"
+	"go.opentelemetry.io/obi/pkg/ebpf"
+)
+
+// Target is everything the injector reads about a process. Callers copy these
+// values out of the Instrumentable up front, so injection can run concurrently
+// with the rest of the discovery pipeline without sharing that object.
+type Target struct {
```
grcevski: I'd rename this struct to something like InjectionTarget and perhaps we can share it with the nodejs injector.

pkg/appolly/discover/attacher.go
mifisignal: Perhaps the comment is confusing. `ta.javaInjector.NewExecutable()` is a blocking call, which is the issue this PR seeks to solve. By moving it into a goroutine, we unblock the instrumenter so it can capture DNS traffic and any other process-related activity.

grcevski: Gotcha, this makes sense now! I like the improvement.

pkg/internal/java/target.go
mifisignal: This patch was written to mimic the style of the NodeJS injector, so that's why it looks how it does.

There's one wrinkle with your suggestion. `NodeInjector.NewExecutable` needs `ie.FileInfo.ELF()` and the discovery loop closes that ELF immediately after the synchronous node call. Putting an `*elf.File` into a struct whose whole purpose is being copied for concurrent use (which is the point of this part of the PR) invites a use-after-close that we're trying to avoid. 

Perhaps this suggestion can be banked for a future PR?

mifisignal: Renamed the struct per your request, along with the filename to match.

mifisignal:
@MrAlias CLA signed now

grcevski (approved):
LGTM!

grcevski:
Thanks for improving this @mifisignal! I guess a bug fix in the same code caused some merge conflicts. Let me know if you can resolve them, if not I can try to if your branch is open.

pkg/appolly/discover/attacher.go
```diff
@@ -124,24 +125,42 @@ func (ta *traceAttacher) attacherLoop(_ context.Context) (swarm.RunFunc, error)
...
+						target := javaagent.InjectionTargetFrom(&instr.Obj)
+						javaInjections.Add(1)
+						go func() {
```
MrAlias: Every discovered JVM can now enter `NewExecutable` concurrently, but `JAttacher` changes euid/egid process-wide and each attacher snapshots and restores those credentials independently. Two JVMs under different UIDs can therefore run under or restore the other attach's identity.

Could we keep the entire credential-changing attach lifetime exclusive, including work that survives `NewExecutable`'s timeout? Serializing only the outer calls is not enough because the inner attach goroutine can outlive that return.

pkg/appolly/discover/attacher.go
```diff
@@ -124,24 +125,42 @@ func (ta *traceAttacher) attacherLoop(_ context.Context) (swarm.RunFunc, error)
...
 	return func(ctx context.Context) {
 		defer ta.OutputTracerEvents.Close()
+		// TODO: cancel an in-flight injection on shutdown (ctx cancellation) or
+		// when a concurrent EventDeleted arrives for the same PID, rather than
+		// only waiting for its internal timeout to elapse. That requires
+		// threading a context.Context through JavaInjector.NewExecutable
+		// (which currently derives its own context.Background()-rooted
+		// timeout), plus a per-PID cancel-func registry.
+		var javaInjections sync.WaitGroup
+		defer javaInjections.Wait()
```
MrAlias: `javaInjections.Wait()` runs before `OutputTracerEvents.Close()`, while `NewExecutable` roots its timeout at `context.Background()`. If `javaagent.attach_timeout` exceeds `shutdown_timeout`—the new integration compose uses 90s versus the default 10s—a stuck attach keeps `traceAttacher` alive past the swarm deadline and produces `CancelTimeoutError`.

Could we pass the run context through the actual attach and cancel and join it, or otherwise cap this drain at the shutdown deadline?

pkg/internal/java/java_inject_notlinux.go
```diff
@@ -10,4 +10,4 @@ package javaagent // import "go.opentelemetry.io/obi/pkg/internal/java"
 type JavaInjector struct{}
 
 func NewJavaInjector(_ any) (*JavaInjector, error) { return nil, nil }
-func (*JavaInjector) NewExecutable(_ any) error    { return nil }
+func (*JavaInjector) NewExecutable(_ InjectionTarget) error { return nil }
```
MrAlias: `gofmt` aligns these declarations, and the macOS lint job is currently failing on this file. Could we apply the formatter here?

```suggestion
func NewJavaInjector(_ any) (*JavaInjector, error)          { return nil, nil }
func (*JavaInjector) NewExecutable(_ InjectionTarget) error { return nil }
```

MrAlias (changes requested):
The early-filter ordering fixes the startup-traffic gap, but the new asynchronous lifecycle is not safe to merge yet. The current head also needs to be brought back to a clean CI state.

pkg/appolly/discover/attacher.go
grcevski: Ah right! I forgot about that aspect, to fix this we'd need some sort of a queue, like a channel where we'll send the processes we need to attach to and they'll still be done in order and one by one.

mifisignal:
Thanks for the feedback! Posted some updates:

Injections now go through a single worker that attaches to one JVM at a time in
discovery order, so the exclusive window is the whole attach lifetime. The
abandoned-attach case is still handled by the existing
`terminated`/`runIfCurrentAttach` guard, which refuses credential changes from
an attach OBI has given up on. Enqueueing is non-blocking, so a stuck attach
still doesn't delay the PID filter.

`NewExecutable` now takes a context and derives its attach deadline from it, so
cancellation abandons an in-flight attach instead of running to
`attach_timeout`. It also refuses to signal a JVM once the caller has given up,
and the worker starts no new attach after cancellation, so the drain returns
within the swarm's cancel timeout even with the integration suite's 90s setting.

New tests cover the invariants: injections never overlap and keep discovery
order, a simulated euid switch/restore per target detects any cross-target
credential leak, enqueue doesn't block behind a stuck attach (including the
full-queue path), and shutdown cancels the in-flight injection without starting
queued ones. I confirmed the first two fail against the previous
goroutine-per-JVM code.

CI on the current head is sitting in `action_required` — could you approve the
run? The only non-lint failure on the prior head was
`TestRuntimeMetricsProm/Go_runtime_metrics_with_Prometheus_export`, a
float-comparison flake in the Go runtime-metrics assertion unrelated to this
change.

pkg/internal/java/java_inject_notlinux.go
mifisignal: Done!

pkg/appolly/discover/attacher.go
mifisignal: Done!

pkg/appolly/discover/attacher.go
mifisignal: Done!

pkg/appolly/discover/attacher.go
```diff
@@ -124,24 +124,36 @@ func (ta *traceAttacher) attacherLoop(_ context.Context) (swarm.RunFunc, error)
...
 					}
 
+					// Injection blocks for up to the Java attach timeout, so it is
+					// queued after the PID is allowed through the eBPF filter, and
+					// runs off the discovery loop. The target is copied out here so
+					// the injection does not share instr.Obj with the consumers it
+					// was just sent to.
+					if javaInjections != nil {
+						javaInjections.enqueue(javaagent.InjectionTargetFrom(&instr.Obj))
+					}
```
MrAlias: This queues every discovered process, although `NewExecutable` only acts on `InstrumentableJava`. If one Java attach is stuck, 100 intervening non-Java discoveries fill the buffer and a later JVM is dropped, permanently losing its Java TLS telemetry.

Could we filter before enqueueing and cover that admission rule in a test?

```suggestion
					if javaInjections != nil && instr.Obj.Type == svc.InstrumentableJava {
						javaInjections.enqueue(javaagent.InjectionTargetFrom(&instr.Obj))
					}
```

MrAlias (commented):
The serialized worker addresses the prior credential and shutdown blockers. Queue admission still lets unrelated process churn crowd out the JVMs this worker exists to inject.

pkg/appolly/discover/attacher.go
mifisignal: Good catch! `NewExecutable` only acts on `svc.InstrumentableJava`, so
non-Java processes were taking slots in a bounded queue that drops on overflow. A
stuck attach plus 100 discoveries of any language evicts a JVM discovered after
them.

I put the filter in `enqueue` rather than at the call site.
`traceAttacher.javaInjector` is a concrete `*javaagent.JavaInjector`, so a test
can't substitute a fake one there — it would have to build the real injector, which
writes the embedded agent jar and only compiles on Linux. `newJavaInjectionQueue`
takes the inject function as a parameter, so a test can pass its own. The rule also
sits next to the buffer it protects. It's covered by
`TestJavaInjectionQueue_OnlyJVMsAreAdmitted`: it holds the worker on a stuck attach,
floods the queue past capacity with non-Java targets, queues a JVM behind them, and
asserts that JVM is still injected and nothing else ever is. It fails without the
filter. The type guard inside `NewExecutable` stays.

`svc.InstrumentableJavaNative` is not admitted, matching what the injector does today.

Unrelated to this PR, but you may want to know: `make lint` fails on
`lint-preempt-guard` at `bpf/gotracer/gotracer.c:34`, and it reproduces on an untouched
checkout of `0f4ec48a`. Looks like the check added in #3059 does not cover that program
on main.

grcevski:
We need another sync with main. There was a conflict with two PRs and we ended up with main that fails lint.

mifisignal:
@grcevski done!

mifisignal:
@MrAlias can you please approve if the changes are satisfactory?

MrAlias (commented):
Sorry for the churn here—I should have caught the queued PID identity issue in the earlier pass. The queue-admission change does address my latest comment.

pkg/internal/java/injection_target.go
```diff
@@ -0,0 +1,29 @@
...
+	"go.opentelemetry.io/obi/pkg/ebpf"
+)
+
+// InjectionTarget is everything the injector reads about a process. Callers
+// copy these values out of the Instrumentable up front, so injection can run
+// concurrently with the rest of the discovery pipeline without sharing that
+// object.
+type InjectionTarget struct {
+	Type svc.InstrumentableType
+	Pid  app.PID
```
MrAlias, line 22: This target is queued by numeric PID only. If the JVM exits and the PID is reused before dequeue, the stale Java target can take an unrelated replacement through the HotSpot path and send it `SIGQUIT` before proving it is a JVM, potentially terminating or core-dumping it.

Could we capture the process start time here and validate it before any attach-side namespace, credential, filesystem, or signal operation? The Go Auto SDK path already follows this pattern. Please cover PID reuse in a regression test.

mifisignal: @MrAlias Done and done. PTAL

MrAlias (commented):
The latest force-push dropped the serialized Java injection queue and context propagation from c9c3524e9, plus the Java-only admission follow-up from 7ba5c08f8. That reopens the credential and shutdown threads; could we restore the three changes together? I have one refinement to my PID-reuse feedback inline.

pkg/internal/java/injection_target.go
```diff
@@ -0,0 +1,47 @@
...
+	StartTime uint64
+}
+
+func InjectionTargetFrom(ie *ebpf.Instrumentable) InjectionTarget {
+	pid := ie.FileInfo.Pid()
+
+	// A failure here means the process is already gone or /proc is unreadable.
+	// Injection will fail on its own, so leave StartTime zero and let the
+	// injector report the error with its own context.
+	startTime, _ := processStartTime(pid)
```
MrAlias, line 39: I should have been explicit about one more invariant in my previous PID-reuse feedback: if this start-time read fails, injection must fail closed. Discarding the error leaves `StartTime` at zero, and `verifyTargetIdentity` treats zero as success, so a recycled PID can still reach the HotSpot attach path and receive `SIGQUIT`.

Could we reject the target when its identity cannot be captured and update the zero-start-time test accordingly?

mifisignal: Whoops! Sorry about those dropped commits. I'm clearly juggling too many issues at once. 🤠 Both are restored now.

Added a fix to ensure we fail safely when the program's start time can't be collected, and test cases to cover it.

MrAlias (approved):

PR #2956, opened 2026-08-07
Restrict nightly release workflow to main

MrAlias:
### Motivation

- Prevent a CI/CD supply-chain authorization flaw where manually-dispatched nightly runs could build, publish, and sign images from arbitrary branch refs by ensuring the nightly publisher only operates on the trusted `main` ref.

### Description

- Add a `github.ref == 'refs/heads/main'` guard to the `compute-version` job and pass `ref: main` into both `nightly-obi` and `nightly-k8s-cache` reusable workflow calls in `.github/workflows/nightly-release.yml` so the publishers check out a trusted ref.

### Testing

- Ran `git diff --check` and a `python3` assertion that verified the `main` ref guard and that `ref: main` is passed into both publisher jobs, and both checks passed.

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -20,7 +20,7 @@ jobs:
   # Derive the dated nightly version string (nightly-<YYYYMMDD>).
   compute-version:
     name: Compute nightly version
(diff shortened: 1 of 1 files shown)


NimrodAvni78 (approved):
Good catch

PR #2957, opened 2026-08-07
Avoid Node.js agent injection into Deno processes

MrAlias:
## What

- Classify Deno separately from Node.js for internal instrumentation decisions.
- Keep Deno on generic tracing and JavaScript route harvesting.
- Preserve `telemetry.sdk.language=nodejs` and the existing `nodejs` route-harvester configuration behavior.
- Add regression coverage for the Node injector boundary and Deno route harvesting.

## Why

Deno was represented internally as Node.js because OpenTelemetry semantic conventions use `nodejs` for JavaScript runtimes. That representation also made Deno eligible for Node-specific agent injection, including the SIGUSR1-based inspector activation path. A Deno process can terminate when it receives that signal.

Separating runtime identity from the exported telemetry language prevents Node-specific injection into Deno without changing its telemetry attributes or generic instrumentation behavior.

## Validation

- `go test -vet=off ./pkg/internal/nodejs`
- `go test -vet=off -run '^TestHarvest(Nodejs|Deno)Routes_' ./pkg/internal/transform/route/harvest`
- `go test -vet=off ./pkg/appolly/app/svc ./pkg/internal/procs`


MrAlias:
To clarify the overlap with #2851: I reviewed that PR, and the shared Deno type split is intentional. This PR uses the same seam for a narrower purpose—keeping Deno out of the Node injector, whose SIGUSR1 inspector path can terminate a Deno process—while preserving the current telemetry language, generic tracing, route harvesting, and configuration behavior.

@mariomac is away for the next week or two, so I opened this focused fix rather than hold the safety issue behind the broader Deno work. The PRs should still converge: whichever lands second can rebase and retain only its distinct behavior and tests.

NimrodAvni78 (approved):
Lgtm

pkg/internal/transform/route/harvest/harvester.go
mmat11:
```suggestion
				h.log.Debug("found application routes", "runtime", runtime, "routes", r.Routes)
```
nit: `language` -> `runtime`

question: do we need `ExtractDenoRoutes()` in place of `ExtractNodejsRoutes()`?

MrAlias: Renamed the local discriminator to runtime, and the log now records the concrete runtime. It uses runtime.String for Node.js and explicitly maps InstrumentableDeno to deno because its String method returns nodejs to preserve `telemetry.sdk.language`. The log therefore reports `runtime=nodejs` or `runtime=deno` without changing exported telemetry attributes.

Deno already reached `ExtractNodejsRoutes` before this type split because it was classified as `InstrumentableNodejs`. This PR keeps that existing dispatch while fixing the Node injector boundary. I opened to #2980 trackstrack work to add Deno-aware command-line discovery and tests with representative Deno invocations.

mmat11 (approved):

grcevski (approved):
LGTM!

Issue #2941, opened 2026-08-06
Node.js injection: retry and failure observability


pinoOgni:
### Component

OBI (eBPF instrumentation)

### Problem

Agent injection is single-shot per discovered process. A transient failure like inspector slow to open after SIGUSR1, seccomp applied at startup only, process busy, permanently disables both trace-context propagation and runtime metrics (not yet merged #2939 ) for that process until it restarts. The only signal is two error lines in the OBI log; there is no metric to alert on and no retry.

### Describe the solution you'd like

- Retry injection with backoff for retryable failures (connection refused /  timeout), bounded attempts.
- An internal metric (e.g. `obi_nodejs_injection_total{result="ok|error"}`) so fleets can alert on silent instrumentation gaps.
- Keep non-retryable outcomes (custom SIGUSR1 handler detected) as a single skip decision.

PR #2975, opened 2026-08-10
tpinjector: attach a no-op SK_SKB verdict on FIONREAD-regression kernels

nishant-sharechat:
### Problem

Kernels carrying commit `929e30f93125` (6.6.128+, 6.12.75+, 6.18.14+, 6.19+) can drop the receive-side readiness wakeup for a socket enrolled in `sock_dir` that has no verdict program: bytes sit in the receive queue but the epoll/poll edge is never delivered.

#2580 (the FIONREAD fixup) addresses the `ioctl(FIONREAD)` half of the same regression, but only helps readers that size their reads via that ioctl (nginx, Java NIO, .NET). Runtimes that block on epoll/poll — Node.js/libuv, the Go netpoller, Python asyncio — get no wakeup and stall. This is the sibling of grafana/beyla#3025 (and #2941).

### Fix

Attach a no-op `SK_SKB` verdict (`SK_PASS`) to `sock_dir`. The kernel then routes enrolled sockets onto the verdict-present receive branch — the path kernel fix `04af4efde58a` repairs — which restores readiness. The program does nothing; the *attachment* is the fix.

It reuses the existing `kernelBreaksFIONREAD()` probe, so it attaches only where the FIONREAD fixup already does and unaffected kernels are untouched. No new config surface, and the generated bindings pick up the program through the normal `make generate` path.

### Validation

Running in production as an equivalent patch (the verdict built at runtime rather than via bpf2go) on affected-kernel nodes across two clusters. On 6.12.85 the HTTP failure rate under context propagation dropped from ~1.8% back to baseline, and the epoll-based stalls disappeared. I have not compiled this exact C form locally (no bpf/clang toolchain on my machine) — CI's `make generate` + build covers that.

### AI disclosure

Prepared with AI assistance (Claude Code). I reviewed every line, validated the approach in production, and take full responsibility for the change per the OBI AI policy.

Refs grafana/beyla#3025

### DIFF
--- a/bpf/tpinjector/tpinjector.c
+++ b/bpf/tpinjector/tpinjector.c
@@ -1160,6 +1160,21 @@ int obi_packet_extender(struct sk_msg_md *msg) {
     return SK_PASS;
 }
 
+// A no-op stream verdict attached to sock_dir. Kernels carrying commit
+// 929e30f93125 (present in 6.6.128+, 6.12.75+, 6.18.14+ and 6.19+) can lose the
+// receive-side readiness wakeup for a socket enrolled in a sockhash that has no
+// verdict program: bytes sit in the receive queue but the epoll/poll edge is
+// never delivered. fionread_fixup.c handles the ioctl(FIONREAD) half of the
+// same regression (nginx, Java NIO, .NET); it cannot help readers that block on
+// epoll/poll (Node.js/libuv, the Go netpoller, Python asyncio). Attaching a
+// verdict program moves enrolled sockets onto the receive path that kernel fix
+// 04af4efde58a targets, which restores the readiness notification. The program
+// passes every skb through unchanged — the attachment is the fix, not the body.
+SEC("sk_skb/verdict")
+int obi_sk_skb_verdict(struct __sk_buff *skb __maybe_unused) {
+    return SK_PASS;
+}
+
 //k_tail_write_msg_traceparent
 SEC("sk_msg")
 int obi_packet_extender_write_msg_tp(struct sk_msg_md *msg) {
--- a/pkg/internal/ebpf/tpinjector/tpinjector.go
+++ b/pkg/internal/ebpf/tpinjector/tpinjector.go
@@ -257,13 +257,30 @@ func (p *Tracer) SocketFilters() []*ebpf.Program {
 }
 
 func (p *Tracer) SockMsgs() []ebpfcommon.SockMsg {
-	return []ebpfcommon.SockMsg{
+	msgs := []ebpfcommon.SockMsg{
 		{
 			Program:  p.bpfObjects.ObiPacketExtender,
 			MapFD:    p.bpfObjects.SockDir.FD(),
 			AttachAs: ebpf.AttachSkMsgVerdict,
 		},
 	}
+
+	// On kernels carrying commit 929e30f93125, sockets enrolled in sock_dir
+	// without a verdict program can lose the receive-side readiness wakeup.
+	// The FIONREAD fixup only repairs ioctl-based readers; a no-op SK_SKB
+	// verdict moves enrolled sockets onto the receive path kernel fix
+	// 04af4efde58a targets, restoring epoll/poll readiness for runtimes that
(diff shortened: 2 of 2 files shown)


grcevski:
This is the exact issue I saw when I attempted to implement the empty verdict on ingress, various kernels failed with the message EOF in few cases. 6.18 LTS is what worries me.

grcevski:
So as per the integration tests, this change breaks all kernels that actually have the bug https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/31417957167. We need to understand why.

nishant-sharechat:
Confirmed — thanks for running the full matrix. It's not just 6.18: it fails on exactly the bug-carrying kernels (6.6, 6.12, 6.18, bpf-next) and passes on the rest, so the verdict truncates on precisely the set it's meant to help.

For context: we've run the runtime-asm equivalent on 6.12.85 in production and it cut the stall failures with no visible EOF wave — but our traffic is small-payload Node.js, and the failing tests are the large-transfer ones, so I don't think we're a counter-example.

To avoid re-walking your ground: was your earlier attempt the parserless BPF_SK_SKB_VERDICT, or a stream_parser+verdict pair? And did the EOF track a specific skb (FIN / zero-length / redirect) rather than plain SK_PASS? Happy to run instrumented variants on our 6.12.85 rig.

mmat11:
just to get some context, is epoll also broken in addition to ioctl(FIONREAD)?

nishant-sharechat:
@mmat11 Yes — epoll readiness is broken independently of `ioctl(FIONREAD)`. On a socket enrolled in the sockhash without a verdict program, bytes land in the receive queue but the epoll/poll edge never fires. That's why #2580's fixup resolved #2941 (nginx, Java NIO, .NET all size reads via the ioctl) while Node.js/libuv, the Go netpoller and Python asyncio stay stuck until their own timeout — they never call it, so there's nothing for an ioctl-layer fixup to correct.

@grcevski — correcting myself from my last comment. I said our small-payload Node.js traffic meant we weren't a counter-example to the large-transfer failures. That was wrong. We reproduced your matrix result in production on 6.12.85.

A gRPC inference path with 2.7–4.3 MB requests, with the verdict attached:

```
cluster.outbound/<model>.upstream_cx_protocol_error: 33
cluster.outbound/<model>.upstream_rq_tx_reset:       33
```

Envoy sees malformed HTTP/2 and resets; grpc-go reports code 13. Client-visible error rate 0.45–0.85/s. Small-payload traffic on the same nodes and kernels was unaffected — a 174 KB path elsewhere in our fleet showed protocol errors at the same rate as a control cluster with propagation disabled.

Four config toggles tracked it exactly. Disabling only the verdict — sockets still enrolled, propagation still on — held clean for an hour: +1 protocol error and 0 new resets across 6 pods, against errors appearing within 2–5 minutes every time it was enabled.

So the matrix failure reproduces outside CI, and the axis is transfer size rather than protocol — BetterStackHQ/collector#142 hit the same truncation on plain HTTP/1.

One observation I can't explain: over that hour with the verdict removed, the stalls did **not** come back on that cluster, even though its clients are Go netpoller. Either the stall window is narrower than we assumed or that path doesn't hit it.

Still happy to run instrumented variants on the 6.12.85 rig — including the stream_parser+verdict pair, or a verdict that only passes below a size threshold, if either is worth testing.

*Prepared with AI assistance; measurements and production validation are my own.*

grcevski:
Hi @nishant-sharechat ,

i was working on this yesterday and the ingress verdict is simply broken for sockets that do splice in the kernel like docker proxy.

is it possible for you to explain how to recreate a reproducer for the stall you are seeing with our existing mitigation? I tried a simple nodejs server and client, but I can't see a problem. Is it dependent on the size of the payload?

grcevski:
I'm going to close this for now, since skb_verdict is broken on many kernels for sockets that do splice. Maybe it worked well in the past, but all modern kernels have issues with it. Let's focus on helping us get a reproduction case.

PR #2977, opened 2026-08-10
integration: netolly GeoIP suite covering src/dst asn+country flow attributes

NimrodAvni78:
## Summary

Adds an integration suite that exercises OBI's GeoIP network-flow attributes
end to end, and declares those attributes in the semantic-convention registry.

When GeoIP is enabled, netolly decorates flows with `src.asn`, `dst.asn`,
`src.country`, and `dst.country`. No suite emitted them before, so they were
never observed by weaver or checked against the schema.

**The suite.** `docker-compose-netolly-geoip.yml` + `TestNetwork_GeoIP` stand up
a server pinned to the MaxMind test address `216.160.83.57` on a dedicated
`216.160.83.0/24` network, plus a client that connects to it. A single
connection produces both a request flow (`dst=…57`) and a response flow
(`src=…57`), so all four attributes get decorated. It uses the MaxMind test
databases already vendored under `internal/test/geoip/`, so it needs no external
data or internet access. The test asserts
`obi_network_flow_bytes_total{dst_asn="AS209",dst_country="US"}` and the `src_*`
equivalents, then weaver live-check validates the emitted shape.

**The schema change.** `schemas/obi/groups/network.yaml` gains the four
attribute declarations and refs them on the flow-bytes / flow-packets metrics.
Without this, weaver flags them as undeclared attributes under the existing
`src`/`dst` namespaces (an `extends_namespace` violation).

No behavioral change to OBI runtime — integration test, config, and schema only.

## Validation

### DIFF
--- a/internal/test/integration/suites_network_geoip_test.go
+++ b/internal/test/integration/suites_network_geoip_test.go
@@ -1,0 +1,39 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package integration
+
+import (
+	"path"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"go.opentelemetry.io/obi/internal/test/integration/components/docker"
+	"go.opentelemetry.io/obi/internal/test/integration/components/promtest"
+)
+
+func TestNetwork_GeoIP(t *testing.T) {
+	compose, err := docker.ComposeSuite("docker-compose-netolly-geoip.yml", path.Join(pathOutput, "test-suite-netolly-geoip.log"))
+	require.NoError(t, err)
+	compose.Env = append(compose.Env, `PROM_CONFIG_SUFFIX=`)
+	require.NoError(t, compose.Up())
+
+	checkGeoIPFlows := func(query string) {
+		pq := promtest.Client{HostPort: prometheusHostPort}
+		require.EventuallyWithT(t, func(ct *assert.CollectT) {
+			results, err := pq.Query(`obi_network_flow_bytes_total` + query)
+			require.NoError(ct, err)
+			require.NotEmpty(ct, results)
+		}, 4*testTimeout, 100*time.Millisecond)
+	}
+
+	checkGeoIPFlows(`{dst_asn="AS209",dst_country="US"}`)
+	checkGeoIPFlows(`{src_asn="AS209",src_country="US"}`)
+
+	runWeaverValidation(t)
+
+	require.NoError(t, compose.Close())
+}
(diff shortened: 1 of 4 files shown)


internal/test/integration/suites_network_geoip_test.go
```diff
@@ -0,0 +1,39 @@
...
+func TestNetwork_GeoIP(t *testing.T) {
+	compose, err := docker.ComposeSuite("docker-compose-netolly-geoip.yml", path.Join(pathOutput, "test-suite-netolly-geoip.log"))
+	require.NoError(t, err)
+	compose.Env = append(compose.Env, `PROM_CONFIG_SUFFIX=`)
+	require.NoError(t, compose.Up())
+
+	checkGeoIPFlows := func(query string) {
+		pq := promtest.Client{HostPort: prometheusHostPort}
+		require.EventuallyWithT(t, func(ct *assert.CollectT) {
+			results, err := pq.Query(`obi_network_flow_bytes_total` + query)
```
MrAlias, line 27: This suite enables only the flow-bytes network feature through the legacy boolean and queries only `obi_network_flow_bytes_total`, so Weaver validates neither the new flow-packets refs nor the existing inter-zone metric. That leaves a real hole: when GeoIP and `network_inter_zone` are both enabled, `NetworkInterZone` selects all four GeoIP attributes by default, but `obi.network.inter.zone.bytes` has no matching refs.

Could we add those four refs and cover the other network metric schemas, either with live signals or a selector-to-registry consistency test?

schemas/obi/groups/network.yaml
```diff
@@ -104,6 +104,30 @@ groups:
...
         examples: ["10.0.0.0/8"]
 
+      # GeoIP decoration of the flow endpoints, attached by the geoip pipeline
+      # stage when GeoIP is enabled. Stored as strings (netolly Metadata is
+      # string-valued). OBI/netolly convention, no upstream equivalent.
+      - id: src.asn
+        type: string
+        stability: development
+        brief: Autonomous System Number of the source endpoint, from GeoIP lookup.
+        examples: ["15169"]
```
MrAlias: The MaxMind path emits nonzero ASNs as `AS%d`, and the new integration test asserts `AS209`, but both ASN examples use the unprefixed form. Could we change both examples to `AS15169` so the registry documents the values OBI emits?

MrAlias (commented):
The end-to-end GeoIP coverage closes the immediate namespace-validation gap and gives us a solid foundation. The remaining feedback is focused on keeping the registry and its validation aligned across the supported network metrics.

internal/test/integration/suites_network_geoip_test.go
NimrodAvni78: good catch, will add those to the schema
regarding a selector-to-registry test its a part of more wide work on measuring telemetry schema coverage which I prefer to keep there

mmat11 (approved):
lgtm

PR #2979, opened 2026-08-10
Deprecate the `application_span` metrics feature

### LINKED ISSUE #2923: deprecate `application_span` (traces_spanmetrics_*) metric feature
currently OBI has 2 ways of exporting "spanmetrics"

1. The `application_span` feature (`FeatureSpanLegacy`) which emits `traces_spanmetrics_*` style metrics (grafana convention)
2. The `application_span_otel` feature (`FeatureSpanOTel`), which emits `traces_span_metrics_*` style metrics ([otel spanmetrics connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/connector/spanmetricsconnector/README.md) convention)

I would say it makes sense for OBI to have a singular otel compliant spanmetrics feature which it will declare and state in its telemetry schema, so i would strive to keep the `application_span_otel` feature and deprecate (and maybe eventually remove) `application_span` feature

NimrodAvni78:
Resolves https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2923

## Summary

OBI has two ways of exporting span metrics:

- `application_span` — emits `traces_spanmetrics_*` (Grafana naming)
- `application_span_otel` — emits `traces_span_metrics_*`, matching the collector-contrib
  [spanmetrics connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/connector/spanmetricsconnector/README.md)

This deprecates both features that emit the Grafana-convention `traces_spanmetrics_*` family, so
OBI converges on a single, OTel-compliant span-metrics feature:

- `application_span` → replaced by `application_span_otel`
- `application_span_sizes` → emits `traces_spanmetrics_size_total` and
  `traces_spanmetrics_response_size_total`, the same legacy family. It has **no OTel-named
  equivalent**; so it is reported as deprecated with no direct replacement rather than implying a swap exists.

Nothing is removed and no metric stops being emitted — existing configurations keep working
exactly as before.

## What changes

**The features are marked deprecated, not disabled.** Both still parse, still enable their
metrics, and still emit `traces_spanmetrics_*`. Users get a startup warning naming the
replacement where one exists.

**JSON schema.** The `Features` item type is now a `oneOf` with the deprecated names in their own
`deprecated: true` branch, so editors and schema validators flag them without rejecting them. The
branch description is generated from the same `deprecatedFeatures` map that drives the warning, so
it names each migration target directly: `application_span (use application_span_otel),
application_span_sizes (no direct replacement)`.

**Telemetry schema.** `traces_spanmetrics_calls_total` and `traces_spanmetrics_latency` are
removed from `schemas/obi/`. The registry now declares only the OTel-canonical names, so the
schema states the contract OBI intends to keep rather than both naming schemes. The OATS cases
that emitted the legacy names move to `application_span_otel` to match — see the reviewer note
below.

**OATS Tests** some OATS tests still emitted the legacy `traces_spanmetrics_calls_total`, that caused weaver validation to fail since we intentionally don't declare them anymore

**Configuration reference.** Regenerated. Deprecated values stay listed, tagged inline:

```
`application_service_graph`, `application_span_otel`, ..., `application_span` (deprecated)
```

## The deprecation warning

Emitted once at startup per deprecated feature, from `Config.validate()`:

```
time=2026-08-07T07:13:16.985Z level=WARN msg="metrics feature is deprecated and will be removed in a future release" feature=application_span use=application_span_otel
time=2026-08-07T07:13:16.986Z level=WARN msg="metrics feature is deprecated and will be removed in a future release" feature=application_span_sizes
```

The check runs against `JoinMetricsConfig().Features`, not the top-level list, because
per-service `discovery.instrument[].metrics.features` sections can enable a feature the top-level
list does not and they drive the exporters through the same joint config
(`pkg/export/prom/prom.go:301`). Checking only `metrics.features` would leave per-service users
silently on a deprecated feature.

`all` / `*` enable both span-metric formats, and the existing conflict resolution drops the legacy
one in favor of OTel, so `application_span` is not reported as deprecated there — the user never
selected it. That resolution path is reachable *only* under `all` / `*`, because setting both
formats explicitly is rejected as a config error, so its message deliberately reports the
resolution without a deprecation notice:

```
level=WARN msg="application_span and application_span_otel cannot be used together, application_span_otel is selected automatically"
```

`application_span_sizes` is not resolved away and keeps emitting, so it *is* reported under
`all` / `*`.

## Follow-up

This PR does not schedule the removal. A later change should drop `FeatureSpanLegacy`,
`FeatureSpanSizes`, `LegacySpanMetrics()`, and the `traces_spanmetrics_*` emission paths in
`pkg/export/otel/metrics.go` and `pkg/export/prom/prom.go` (including
`SpanMetricsRequestSizes` / `SpanMetricsResponseSizes`), and migrate the test configurations
listed above.

## How this was validated

- `go test ./pkg/export/ ./pkg/obi/ ./cmd/config-docs/`
- `golangci-lint run` on the touched packages
- `make check-config-schema` — schema and docs regenerated and in sync
- `make lint-schema` — weaver v0.25.1 registry check
- OATS http suite — the live-check assertion runs in CI; it cannot run locally on a machine whose
  kernel lacks BTF

## Validation


grcevski (approved):
LGTM!

pkg/obi/config.go
```diff
@@ -810,9 +810,16 @@ func (c *Config) validate(context validationContext) error {
...
+		// The user did not pick the legacy format, so this reports the resolution only and
+		// leaves the deprecation notice to the features that stay enabled.
 		if c.Metrics.Features.ResolveSpanMetricsConflict() {
-			slog.Warn("application_span and application_span_otel cannot be used together, application_span_otel is selected automatically")
+			slog.Warn("application_span and application_span_otel cannot be used together," +
+				" application_span_otel is selected automatically")
 		}
+		// Per-service sections can enable features the top-level list does not, and they
+		// drive the exporters through JoinMetricsConfig, so report against the same set.
+		c.warnDeprecatedMetricsFeatures()
```
MrAlias, line 814: `ResolveSpanMetricsConflict` only mutates the top-level feature mask. A per-service `features: [all]` remains `FeatureAll` when `JoinMetricsConfig` adds it back, so `warnDeprecatedMetricsFeatures` reports `application_span` and the shared exporter selects legacy span metric names for every span-metrics-enabled service using it. That contradicts the documented `all`/`*` behavior.

Could we apply the same validation semantics to each per-service mask—reject explicit legacy+OTel and resolve `all`/`*` to OTel—and cover both cases here?

schemas/obi/manifest.yaml
```diff
@@ -10,10 +10,10 @@ description: |
...
     `instance` / `job` scrape-convention attributes
   - Span metrics matching the OTel collector-contrib `spanmetricsconnector`
-    output, in both naming schemes: the OTel-canonical
-    `traces_span_metrics_*` names (emitted when the `application_span_otel`
-    metrics feature is enabled) and the legacy Grafana-style
-    `traces_spanmetrics_*` names (OBI's default `application_span` feature)
+    output, under the OTel-canonical `traces_span_metrics_*` names (emitted
+    when the `application_span_otel` metrics feature is enabled). The legacy
+    Grafana-style `traces_spanmetrics_*` names of the deprecated
+    `application_span` feature are deliberately not declared here
```
MrAlias: `application_span` still parses and both exporters still emit `traces_spanmetrics_*` when `FeatureSpanLegacy` is enabled. Excluding those metrics here makes the registry stop being the complete contract described below; the OTEL Kubernetes fixture already fails Weaver with missing-metric violations, while migrating all four OATS fixtures removes OATS live-check coverage for the legacy path.

Could we keep the legacy metric definitions until the runtime emission is removed (marking them deprecated if Weaver supports it), and retain one legacy OATS case alongside the canonical coverage?

MrAlias (commented):
The deprecation behavior is not yet consistent across the registry and per-service configuration paths.

schemas/obi/groups/spanmetrics.yaml
```diff
@@ -99,15 +99,18 @@ groups:
...
+  # spanmetrics processor output. Emitted by the deprecated `application_span`
+  # metrics feature; the `traces.span.metrics.*` names above are the
+  # replacement (`OTEL_EBPF_OTEL_METRICS_FEATURES=application_span_otel`).
   - id: metric.traces_spanmetrics_calls_total
     type: metric
     metric_name: traces_spanmetrics_calls_total
     instrument: counter
     unit: ""
     stability: development
+    deprecated:
```
MrAlias, line 111: `deprecated` is accepted by the registry linter, but Weaver treats every emitted deprecated metric as a violation. The current K8s OTEL and OATS HTTP checks fail with 10 and 78 violations from these two legacy metrics, and this PR intentionally keeps emitting them during the compatibility window.

Could we keep the legacy definitions without formal `deprecated` metadata (or change the live-check policy in this PR) so declaring the compatibility contract does not leave CI red?

pkg/obi/config.go
```diff
@@ -825,6 +824,60 @@ func (c *Config) validate(context validationContext) error {
...
+}
+
+// resolveSpanMetricsFormats rejects an explicit legacy + OTel combination and resolves the
+// implicit one coming from "all"/"*" in favor of OTel, for every feature list that reaches
+// the exporters. Resolving only the top-level list would let a per-service "all" put every
+// span-metrics service back on the legacy names.
+func (c *Config) resolveSpanMetricsFormats() error {
+	resolved := false
+	for _, features := range c.spanMetricsFeatureMasks() {
+		if features.InvalidSpanMetricsConfig() {
```
MrAlias, line 849: The per-service cases I called out earlier are covered now, but I missed the cross-mask case. These checks compare each mask independently, while both exporters choose metric names from `JoinMetricsConfig()`, which ORs all masks. A top-level `application_span` with a per-service `application_span_otel` therefore passes validation, then `LegacySpanMetrics()` selects legacy names for every service—including the service that requested OTel.

Could we validate the combined mask as well and add a split top-level/per-service test?

MrAlias (commented):
The compatibility path is still not merge-ready: its registry metadata leaves the required live checks red, and the shared exporter can still select a format that disagrees with a per-service request.

schemas/obi/groups/spanmetrics.yaml
NimrodAvni78: Done!

pkg/obi/config.go
NimrodAvni78: Done!

MrAlias (approved):
Non-blocking: the joined-mask check still misses `all`/`*` combined with an explicit legacy mask in another scope. Could we track that split configuration for follow-up coverage?

PR #2981, opened 2026-08-10
cp: avoid creating multiple root spans in case of foreign traceparents

mmat11:
## Summary

When an outgoing request already carries a traceparent that OBI's in-process
context doesn't know (the app injected it itself, an SDK, a metrics scraper,
or a proxy sidecar reusing one context), the client span was emitted with an
empty parent. Every request added one more root span to that trace;
high-volume reusers turned this into thousands of extra roots per trace.
Introduced in #2232.

The client span now adopts the header's span id as its parent, joining the
trace as a child instead of a second root.

## Validation

### DIFF
--- a/bpf/generictracer/protocol_http.h
+++ b/bpf/generictracer/protocol_http.h
@@ -40,8 +40,6 @@
 
 volatile const u32 high_request_volume;
 
-SCRATCH_MEM_SIZED(http_previous_trace_id, TRACE_ID_SIZE_BYTES);
-
 // empty_http_info zeroes and return the unique percpu copy in the map
 // this function assumes that a given thread is not trying to use many
 // instances at the same time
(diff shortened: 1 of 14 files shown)


bpf/generictracer/protocol_http.h
```diff
@@ -574,7 +574,9 @@ __obi_continue_protocol_http_tp(struct pt_regs *ctx,
                 } else if (previous_trace_id &&
                            bpf_memcmp(previous_trace_id, tp_p->tp.trace_id, TRACE_ID_SIZE_BYTES) !=
                                0) {
-                    __builtin_memset(tp_p->tp.parent_id, 0, sizeof(tp_p->tp.parent_id));
+                    // the app injected a foreign trace: its span becomes the
```
grcevski: I see, so if we have a client call, do we now just trust what's in the headers? I think this is what will happen, but this will cause problems with proxies. Istio namely just blindly copies the incoming headers to the outgoing headers, and I think nginx would do the same if no otel instrumentation is loaded. So we'll get the incoming and outgoing spans jumbled.

mmat11: > but this will cause problems with proxies

you are right.. I'm still investigating this issue for which we have N root spans and neverending traces under various circumstances (proxies/scrapers), will get this out of draft when I validate it locally

PR #2982, opened 2026-08-10
Add kernel-side Aerospike protocol detection

pinoOgni:
## Summary

This PR adds the kernel-side for Aerospike and it continues the work done in #2545 

The kernel-classified events dispatch directly to the existing Aerospike parser and the userspace heuristic stays as a fallback for connections attached mid-flight, mirroring the Kafka setup.

Notes: 
- Classified connections keep the generic wire-layer large buffers, so captured data is unchanged and protocol-specific response reassembly (fixing the split-read status limitation) is the planned follow-up.
- Two known trade-offs, both shared with the other kernel-classified protocols: request-only events are no longer flushed on connection close (an operation whose response is never captured no longer produces a span), and a classification sticks for the connection's cache lifetime so a misclassified connection would have its events ignored rather than re-detected.

Limitations/Follow-ups
- Split-read responses (the Java client reads the 8-byte proto header separately from the body) still miss `result_code`: kernel-side response reassembly is the next PR, building on the dispatch hook this PR adds.
- Type-4 compressed frames classify the connection but their bodies remain unparsed (unchanged).

## Validation

### DIFF
--- a/bpf/common/connection_info.h
+++ b/bpf/common/connection_info.h
@@ -23,6 +23,8 @@ enum protocol_type : u8 {
     k_protocol_type_mqtt = 5,
     k_protocol_type_mssql = 6,
     k_protocol_type_sunrpc = 7,
+    // NOTE: 8 and 9 are reserved for NATS and AMQP (userspace placeholders).
+    k_protocol_type_aerospike = 10,
 };
 
 // Struct to keep information on the connections in flight
--- a/bpf/common/connection_info.h
+++ b/bpf/common/connection_info.h
@@ -12,6 +12,7 @@
 
 #include <logger/bpf_dbg.h>
 
+// Keep these values aligned with consts in pkg/ebpf/common/common.go
 enum protocol_type : u8 {
     // Default value, used when the protocol is not known or will be determined/classified
     // in userspace.
@@ -23,7 +24,8 @@ enum protocol_type : u8 {
     k_protocol_type_mqtt = 5,
     k_protocol_type_mssql = 6,
     k_protocol_type_sunrpc = 7,
-    // NOTE: 8 and 9 are reserved for NATS and AMQP (userspace placeholders).
+    k_protocol_type_nats = 8,
+    k_protocol_type_amqp = 9,
     k_protocol_type_aerospike = 10,
 };
 
--- a/bpf/generictracer/protocol_aerospike.h
+++ b/bpf/generictracer/protocol_aerospike.h
@@ -53,7 +53,7 @@ static __always_inline u8 is_aerospike(connection_info_t *conn_info,
         return 0;
     }
 
-    unsigned char hdr[k_aerospike_proto_header_len + 1];
+    unsigned char hdr[k_aerospike_proto_header_len + 1] = {};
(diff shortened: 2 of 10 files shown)


bpf/common/connection_info.h
```diff
@@ -23,6 +23,8 @@ enum protocol_type : u8 {
     k_protocol_type_mqtt = 5,
     k_protocol_type_mssql = 6,
     k_protocol_type_sunrpc = 7,
+    // NOTE: 8 and 9 are reserved for NATS and AMQP (userspace placeholders).
```
grcevski: Maybe we should define them here, so there's no confusion?

bpf/generictracer/protocol_aerospike.h
```diff
@@ -0,0 +1,74 @@
...
+    if (*protocol_type != k_protocol_type_unknown) {
+        return 0;
+    }
+
+    // require the full proto + as_msg headers so the signature check below is meaningful
+    if (data_len < k_aerospike_proto_header_len + k_aerospike_as_msg_header_len) {
+        return 0;
+    }
+
+    unsigned char hdr[k_aerospike_proto_header_len + 1];
```
grcevski: Do you know if the stack is 0 initialised always here? I think when we construct with {} or {0} it does it, but if not then the hdr check below might get some random memory and succeed? Far fetched.

grcevski: I think it will be zero initialised.

grcevski (approved):
LGTM! Minor comments.

bpf/generictracer/protocol_aerospike.h
pinoOgni: Hi @grcevski, I ran a test and it is not zero-initialized, but IMO it's not a problem: the next line is `bpf_probe_read`, which fully overwrites all 9 bytes on success, and on failure we return before reading anything. 
Adding `= {}` would zero 9 bytes on every `is_aerospike` invocation only for them to be immediately overwritten. 

WDYT?

mmat11: I would init it, for consistency (even if in this particular case doesn't matter)

pkg/ebpf/common/tcp_detect_transform.go
mmat11, line 378: question: can we remove this line then? or do we need the fallback like the Kafka case? how likely it is for the connection to be misclassified?

bpf/generictracer/protocol_aerospike.h
pinoOgni: Sure, fine by me.

pkg/ebpf/common/tcp_detect_transform.go
pinoOgni: Yep! We need the fallback for unclassified events like the `matchKafkaFallback` case where we cover obi attached mid-connection and also `protocol_cache` lru eviction mid-connection.

It's not for misclassified connections because those never reach this line, since their events carry the other `protocol_type` and are consumed by that protocol's dispatch.

mmat11 (approved):

PR #2985, opened 2026-08-10
Make integration tests fail fast

MrAlias:
## Summary

- stop Go integration shards after the first reported test failure
- cancel remaining integration matrix jobs when any shard fails
- rely on the workflow supervisor for flaky-test retries instead of retrying inside a partially executed shard

## Motivation

Integration shards currently continue executing tests after a failure is known. In [run 31420248560](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/31420248560/job/93581162625?pr=2984), shard 8 reported its first failing subtest at 20:20:25 UTC but kept running until it was canceled at 20:30:54 UTC.

Adding Go's `-failfast` flag stops later tests and subtests from being scheduled. Enabling GitHub's matrix fail-fast behavior also cancels queued and in-progress sibling shards.

The per-shard `gotestsum` retries are removed because a retry can pass after fail-fast prevented later tests from running, producing incomplete coverage. The existing workflow supervisor still provides a retry for flaky workflow failures.

Cleanup and log-upload steps continue to run so failed shards retain their diagnostics.

## Validation

- ran `actionlint` syntax and expression validation on the modified workflows, excluding existing repository diagnostics
- smoke-tested the `gotestsum` invocation with `-failfast`
- dry-ran the VM integration Make target
- ran `git diff --check`


mmat11 (approved):
lgtm, needs rebase

florianl (approved):

NimrodAvni78 (approved):
HUGE

skl (approved):
lgtm in principle but seems hard to get all CI checks to pass.

Suggest either:

1. addressing some flaky tests in the same PR to make the checks more likely to pass, or:
2. increasing the [MAX_ATTEMPTS of the flaky supervisor](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/28d5467500e496bc5ffda2f1c0c9dbf01d211d28/scripts/rerun-flaky.sh#L16).

.github/workflows/pull_request_integration_tests.yml
```diff
@@ -117,7 +117,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 45
     strategy:
-      fail-fast: false
+      fail-fast: true
```
skl: what if we keep the GitHub fail-fast as `false`? at least then one flaky test doesn't cancel all shards

MrAlias:
Closing as this does not seem viable given our current testing flakiness.

PR #2996, opened 2026-08-11
Add protocol- and signal-scoped filter runtime

MrAlias:
## Summary

- Add an internal application-filter representation scoped by instrumentation and telemetry signal.
- Classify request event types once and reuse that classification for filtering and trace instrumentation selection.
- Apply scoped filters through the existing trace and metric ignore flags so unrelated spans remain available to other exporters.
- Leave legacy `filter.application` behavior and Config V2 conversion unchanged.

## Motivation

The current runtime has one global application filter. Applying an HTTP `server.port` predicate there removes related SQL spans because they do not share the HTTP port. Config V2 already models filters per protocol and signal, but the converter cannot preserve that distinction without a corresponding runtime representation.

This PR establishes that runtime boundary. A follow-up will populate it from Config V2 and remove the converter restriction that currently requires every protocol and signal filter to be identical.

## Behavior

When the internal scoped configuration is populated, an HTTP trace filter affects only HTTP trace spans. Unconfigured protocols and signals pass through unchanged, and trace and metric decisions remain independent. Metric filters use the existing ignore flag honored by application, span, and service graph metrics.

Existing Config V1 filtering continues through the original global filter node.

## Testing

- Classifier coverage for supported event types.
- Gate coverage for protocol and trace/metric independence.
- Pipeline regression coverage for an accepted HTTP parent retaining its SQL child.
- `make verify`.

Refs #1282


grcevski (approved):
LGTM! Sweet, looks great!

PR #2997, opened 2026-08-12
Fail schema generation on duplicate config type names

elliehmy:
## Summary

The remaining half of #2898. #2899 fixed the symptom by renaming the colliding types; this PR adds the guard.

The generator now records the declaring package of every type the reflector names, and fails generation when one name is claimed by more than one package.
`check-config-schema` already runs the generator on every PR, so no workflow change is needed.

## Implementation

The `Namer` hook is used purely for observation. It renames nothing — it returns an empty name so the reflector keeps its default naming, and only records the types passing through. The check itself runs after reflection completes.

I picked this hook because every type that needs a definition name has to pass through it, so there is no need to recompute which types end up in `$defs`. I tried a hand-written `reflect` walk first, but it followed unexported fields and produced false positives on things like `sync.Mutex`.

As a side effect this also covers collisions in the enum registry. The reflector asks for names on more types than end up in `$defs`; the difference is enum types such as `LogLevel`, `Protocol` and `Source`. They produce no definition, but the `enums` registry is keyed by bare name too, so a collision concatenates their values via `append`.

## Limitations

Name collisions among types that are scanned from `packagesToScan` but are not reachable from `obi.Config` are not detected. The reflector never meets those types while the AST scan still collects them, so the env and enum registries can be polluted.

For that to cause actual damage, all of the following must hold:
- the name matches exactly a type that the schema actually uses
- that type carries `env` tags or enum constants, so there is something to pollute
- it is scanned later, since being scanned earlier makes the overwrite harmless

Today the only names declared more than once across the scanned packages are `goRuntimeHistogramKey` and `goRuntimeHistogramState`, both unexported runtime types that never reach the schema. The two collisions that actually broke in #2898 were reachable from both sides, and this guard catches those.

Covering the rest would mean recording the declaring package along the AST scan path as well, which is a larger change than the failure conditions justify, so it is out of scope here.

## What I verified

The generated `config-schema.json` is byte-for-byte identical to the committed file — the guard only observes, so the output does not change.

Temporarily reverting the two renames from #2899 to reproduce a real collision makes generation fail like this:

```
Error: type name declared in more than one package; rename one type per group so that every configuration type has a unique name:
  MetricsConfig
    go.opentelemetry.io/obi/pkg/export/otel/otelcfg
    go.opentelemetry.io/obi/pkg/export/otel/perapp
  PrometheusConfig
    go.opentelemetry.io/obi/pkg/export/imetrics
    go.opentelemetry.io/obi/pkg/export/prom
```

## Generative AI disclosure

Claude Code was used for the design and the implementation. I reviewed the approach and the alternatives, the code and the tests myself, and verified them locally.

## Validation

### DIFF
--- a/cmd/obi-schema/main.go
+++ b/cmd/obi-schema/main.go
@@ -399,6 +399,8 @@ func main() {
 
 	schema := reflector.Reflect(&obi.Config{})
 
+	inlineTypeSchemas := g.buildInlineTypeSchemas(reflect.TypeFor[obi.Config]())
+
 	// A collision makes every later step operate on a schema that already
 	// merged two unrelated types, so stop before post-processing it.
 	if err := g.checkTypeNameCollisions(); err != nil {
@@ -410,7 +412,7 @@ func main() {
 	schema.Description = "JSON Schema for OpenTelemetry eBPF Instrumentation (OBI) configuration"
 
 	// Process inline fields first (merge properties from inline types)
-	g.processInlineFields(schema)
+	g.processInlineFields(schema, inlineTypeSchemas)
 
 	// Process deprecated annotations from comments
 	processDeprecated(schema)
@@ -465,7 +467,9 @@ type jsonSchemaer interface {
 
 // buildInlineTypeSchemas uses reflection to find inline fields that implement JSONSchema().
 // It walks the type hierarchy starting from rootType and returns a map of type name to schema function.
-func buildInlineTypeSchemas(rootType reflect.Type) map[string]func() *jsonschema.Schema {
+// Walking also records the package of every named inline field type, which is
+// what makes those types visible to checkTypeNameCollisions.
+func (g *SchemaGenerator) buildInlineTypeSchemas(rootType reflect.Type) map[string]func() *jsonschema.Schema {
 	result := make(map[string]func() *jsonschema.Schema)
 	visited := make(map[reflect.Type]bool)
 
@@ -510,6 +514,9 @@ func buildInlineTypeSchemas(rootType reflect.Type) map[string]func() *jsonschema
 				for fieldType.Kind() == reflect.Pointer {
 					fieldType = fieldType.Elem()
 				}
+
+				// The reflector never names inline types, so the collision check only sees them from here.
+				g.recordTypeName(fieldType)
 
 				// Check if it implements JSONSchema()
 				if hasJSONSchemaMethod(fieldType) {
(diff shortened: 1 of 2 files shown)


cmd/obi-schema/main.go
```diff
@@ -43,19 +44,81 @@ type SchemaGenerator struct {
...
 
+// newReflector builds the reflector used to generate the schema.
+func (g *SchemaGenerator) newReflector() *jsonschema.Reflector {
+	return &jsonschema.Reflector{
+		RequiredFromJSONSchemaTags: true,
+		AllowAdditionalProperties:  true,
+		ExpandedStruct:             true,
+		FieldNameTag:               "yaml",
+		Mapper:                     g.customMapper(),
+		Namer:                      g.recordTypeName,
```
MrAlias, line 73: `Namer` is not invoked for every reachable named type. For a `yaml:",inline"` field, `reflectStructFields` recurses directly; if the field is a named map such as `services.MetadataGlobMap`, it returns without calling `typeName`. OBI later selects that inline schema by bare name in `buildInlineTypeSchemas` and `processInlineFields`.

A second reachable type named `MetadataGlobMap` can therefore occupy `$defs/MetadataGlobMap`; only that package is recorded, the collision check passes, and `processInlineFields` merges the wrong schema into `GlobAttributes`. Could we record the package identities used by the inline-schema path and add a reachable/reachable inline-name collision test?

MrAlias (commented):
The Namer-based guard misses a reachable inline path, so one class of type-name collisions can still corrupt the generated schema.

cmd/obi-schema/main.go
elliehmy: Thanks — I had assumed every named type passes through`Namer`. 
Fixed in 1276e4e5. The inline walk now records the declaring package, and it runs before the check rather than at merge time, which is why `processInlineFields` takes the map as a parameter now.
Added the reachable/reachable test — it fails without the fix.

MrAlias (commented):
The reachable inline-path collision is now covered, but the new recording step can make schema generation fail because of types the reflector excludes.

cmd/obi-schema/main.go
```diff
@@ -446,6 +515,9 @@ func buildInlineTypeSchemas(rootType reflect.Type) map[string]func() *jsonschema
 					fieldType = fieldType.Elem()
 				}
 
+				// The reflector never names inline types, so the collision check only sees them from here.
+				g.recordTypeName(fieldType)
```
MrAlias, line 540: `buildInlineTypeSchemas` walks every nested field, including non-anonymous unexported fields and fields excluded with `yaml:"-"` or `jsonschema:"-"`; the reflector skips those paths. If an ignored nested struct contains an inline `http.Header` while a reachable `tar.Header` is reflected, this records both packages and `checkTypeNameCollisions` aborts even though only `archive/tar.Header` can affect the schema. Could we make this traversal use the same field-eligibility and tag rules as the reflector, and add a test showing an unreachable inline name does not collide?

elliehmy: Thanks for the careful review — fixed in ac7f436.

MrAlias (approved):

PR #2999, opened 2026-08-12
feat(metrics): allow disabling HTTP body size metrics independently (#2989)

### LINKED ISSUE #2989: Allow disabling the HTTP body size metrics independently of the `application` feature
### Component

OBI (eBPF instrumentation)

### Problem

Enabling `application` in `metrics.features` emits four HTTP body size histograms alongside
the RED duration histograms, and there is no way to turn off only the former:

- `http.server.request.body.size`
- `http.server.response.body.size`
- `http.client.request.body.size`
- `http.client.response.body.size`

These metrics have different requirement levels in the semantic conventions, but they share a single gate. The application feature maps to FeatureApplicationRED, so the RED duration histograms and the HTTP body size histograms are created under that one condition — even though body size is not a RED metric.

| Metric | semconv | OBI gate |
| --- | --- | --- |
| `http.{server,client}.request.duration` | Stable / Recommended | `AppRED()` + `HTTPEnabled()` |
| `http.{server,client}.{request,response}.body.size` | Development / **Opt-In** | **same** |

The feature is defined in [feature.go#L33](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/cad6aa9ac7ebbd0cb7dab91999801aefafbf15cb/pkg/export/feature.go#L33).
The OTLP exporter creates the four histograms only when `Features.AppRED()` is set ([metrics.go#L433](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/cad6aa9ac7ebbd0cb7dab91999801aefafbf15cb/pkg/export/otel/metrics.go#L433)),
and the Prometheus exporter registers them under the same condition ([prom.go#L791](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/cad6aa9ac7ebbd0cb7dab91999801aefafbf15cb/pkg/export/prom/prom.go#L791)).

The [metric requirement level specification](https://opentelemetry.io/docs/specs/semconv/general/metric-requirement-level/) says of Opt-In metrics:

> Instrumentations SHOULD emit the metric **if and only if the user configures the
> instrumentation to do so.**

Today they are emitted whether or not the user asked for them. Dropping `http` from
`otel_metrics_export.instrumentations` does remove them, but it removes HTTP RED with them, so
it is not a way to separate the two.

**OBI already separates the equivalent size metrics elsewhere.**

| Feature | Metrics |
| --- | --- |
| `application_span_sizes` ([feature.go#L64](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/cad6aa9ac7ebbd0cb7dab91999801aefafbf15cb/pkg/export/feature.go#L64)) | `traces_spanmetrics_*_size_total` |
| `application` | RED **+ the four HTTP body size histograms** |

Size metrics are an independent feature on one side and bundled on the other.

**Two secondary points**, neither of which is what the issue is about on its own:

- **Cost** — in our measurements the four body size metrics accounted for roughly 20% of all
  OBI time series.
- **The values themselves** — when `Content-Length` is not available, the value falls back to
  accumulated wire bytes, which include headers ([http_transform.go#L114-L125](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/cad6aa9ac7ebbd0cb7dab91999801aefafbf15cb/pkg/ebpf/common/http_transform.go#L114-L125)),
  so a single series can mix body sizes with header-inclusive byte counts. We think fixing
  that fallback is worth doing in its own right.

Taken together, though, the current state is that a non-trivial share of the metric volume is
spent on values that are hard to rely on. Until the fallback is addressed, it seems reasonable
for users to at least have the option of turning these metrics off.

Any mechanism that lets us keep HTTP RED while dropping the four body size histograms. Two
properties would be enough: the body size metrics can be disabled while `http` instrumentation
stays enabled, and it can be done in OBI's own configuration rather than in a downstream
pipeline.

We don't have a strong preference on the shape. Two ideas, in case they are useful:

**1. A separate feature key**, following the same pattern as `application_span_sizes`:

```yaml
metrics:
  features: [application, application_http_sizes]
```

**2. Split on the `instrumentations` axis** instead, leaving `application` as it is. The size
metrics are HTTP-only, so that axis fits them, and it does not add to the feature list:

```yaml
otel_metrics_export:
  instrumentations: [http, http_sizes]   # `http` alone would give RED only
```

Whether to keep the current behaviour as the default — emitting them for now and dropping them
in a later major release, or making them opt-in from the start — seems like a separate call for
the project to make.

**`attributes.select`** — does not remove metrics. Even with `exclude: ["*"]` on the body size
sections, only *metric* attributes are dropped ([attr_selector.go#L203](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/cad6aa9ac7ebbd0cb7dab91999801aefafbf15cb/pkg/export/attributes/attr_selector.go#L203)).
Resource attributes such as `pod`, `container` and `instance` are governed by a single global
`resource` section and cannot be scoped to one metric. Dropping `http.route` would also collapse responses of very different
sizes into one histogram, which makes the metric meaningless.

**Reducing the histogram buckets** (`buckets.{request,response}_size_histogram`) lowers the
cost, but it is a cardinality workaround and does nothing about the Opt-In requirement.

**Dropping the metrics with the collector's filter processor** works, and it is what we are
doing. It means every OBI user has to add pipeline configuration for metrics the specification
says should not be emitted unless explicitly enabled.

### Additional context

_No response_

KHARSHAVARDHAN-eng:
Fixes #2989

### Problem
Enabling `application` in `metrics.features` currently emits four HTTP body size histograms alongside RED duration metrics without a way to disable only the former:
- `http.server.request.body.size`
- `http.server.response.body.size`
- `http.client.request.body.size`
- `http.client.response.body.size`

### Solution
Introduces an `application_http_sizes` (`FeatureApplicationHTTPSizes`) feature flag to gate HTTP request and response body size metrics independently of `application` RED duration metrics.

- Maps `"application_http_sizes"` in `metrics.features`.
- Updates `DefaultConfig` to include `application_http_sizes` to preserve default behavior while allowing users to specify `metrics.features: [application]` to disable HTTP body size metrics.
- Gates view configuration, meter initialization, collector registration, and observations in both OTLP and Prometheus exporters behind `AppHTTPSizes()`.
- Updates configuration schema and documentation.

### Testing
Added unit tests in `feature_test.go`, `otel/metrics_test.go`, and `prom/prom_test.go` verifying that `features: [application]` emits RED duration metrics while dropping body size histograms.


marctc (changes requested):
Thanks for your contribution. In the original issue there were a few ideas yet to be discussed, so I suggest that before jump into an implementation directly we discuss it there. This is not an immediate change and this approach affects (breaking changes) existing users.

PR #3000, opened 2026-08-12
fix(ebpf): prevent latching incorrect TLS connection socket attribution (#2998)

### LINKED ISSUE #2998: [BUG] TLS client spans are attributed to the wrong socket
### Component(s)

SSL connection binding / connection attribution
(`bpf/common/ssl_helpers.h`, `bpf/generictracer/ssl_defs.h`,
`bpf/generictracer/protocol_tcp.h`, `pkg/ebpf/common/tcp_detect_transform.go`)

### Nature of the bug

When OBI cannot bind an `SSL*` to a real socket, it guesses the connection from the
thread's most recent activity and then permanently latches that guess. On an
event-loop thread, the guess is frequently the *inbound* request being served, so an
outbound client span is attributed to the caller's socket.

Because the latch is never revisited, the error persists for the life of that `SSL*`.
Severity is therefore proportional to connection lifetime:

- Short-lived / churning connections (AWS SDK HTTPS, etc.) self-heal on reconnect, so
  the fault appears as a burst after an OBI restart that decays away.
- Long-lived pooled connections never churn, so they stay wrong indefinitely.

The worst case in our fleet is a Python service holding one pooled TLS Valkey connection.
All of its Redis spans report the wrong peer:

```
service.name      = notification-service
db.system.name    = redis
db.operation.name = GET
server.address    = router          <-- an Apollo Router pod that calls INTO this service
server.port       = 43694           <-- that inbound connection's ephemeral source port
duration          = 18553 ms
```

Redis protocol decoding is correct; command names (`GET`, `HGETALL`, `EXISTS`, `ZRANGE`)
are accurate. However, the connection association, `server.address`, `server.port` and the duration
are all wrong.

### Observations

**1. It is not language- or protocol-specific.** When we query for the signature, we find it across three
Node.js services, on AWS SDK calls (`POST /`), with multiple distinct incorrect ports each
(i.e. several independently mis-latched connections):

```
annotation-service   port=36348  n=13      discussion-service   port=43950  n=9
annotation-service   port=43548  n=8       discussion-service   port=32948  n=3
annotation-service   port=42058  n=7       gateway              port=44434  n=9
annotation-service   port=57734  n=6       gateway              port=57894  n=7
```

In one case, we observed that the misattributed peer is the pod's own node.

**2. The fault bursts on OBI restart and decays as connections churn.** At startup 
every existing TLS connection must be re-latched at once, so a wave of wrong 
guesses is taken and then decays as those connections recycle.

**3. The pooled connection does not decay.** Across three OBI restarts the Valkey
connection is re-latched and is wrong every time, with a fresh wrong port each time:

```
before restart   peer=router:43694
after restart 1  peer=router:59224
after restart 2  peer=router:51916
after restart 3  peer=router:59408    (pod B: peer=10.0.194.213:47132 — a router pod IP)
```

**4. The reported socket does not exist.** Decoding the pod's full `/proc/net/tcp` (17
sockets) shows the reported port on neither side of any connection, while the real Redis
connection is established and healthy:

```
10.0.159.101:32950  ->  10.0.202.229:6379   ESTAB
```

**5. The substitution is directly observable in the eBPF logs.** With
`OTEL_EBPF_BPF_DEBUG=true` (whose `bpf_dbg_printk` output reaches `slog` via the
`debug_events` ringbuf and `pkg/internal/ebpf/logger`), the Python service (`pid=4111`)
has one `SSL*` in the window — its pooled Valkey connection — and it is bound
to an *inbound* socket:

```
=== SSL_write_ex id=4111 ssl=aaaaf58d2160 ===
SSL_buf id=4111 ssl=aaaaf58d2160
SSL conn
[conn] s_ip=10.0.158.80,  s_port=60806
[conn] d_ip=10.0.159.101, d_port=8003      <-- this pod's own LISTEN port
```

`10.0.158.80` is an Apollo Router pod and `8003` is the service's inbound port, so
`ssl_to_conn` holds the router→service connection for an `SSL*` that only ever carries
Redis traffic.

The correct tuple is available and ignored. Moments earlier in the same log, the
kernel path prints the real Redis connection for the same pid:

```
=== kprobe/tcp_sendmsg id=4111, size=96 ===
[conn] s_ip=10.0.159.101, s_port=32950
[conn] d_ip=10.0.202.229, d_port=6379      <-- the real Valkey ENI
```

A read does not repair a wrong latch. The same `ssl=aaaaf58d2160` resolves to the same
wrong tuple under `SSL_read_ex` as under `SSL_write_ex`, because `handle_ssl_buf` only
consults the fallback when `ssl_to_conn` misses — once populated, it is never
re-evaluated. No `SSL conn` in the capture ever resolved to port 6379.

**6. Name resolution is not at fault.** The span reaches `NameResolver` as a Redis client
span and resolves the *server* side to a router pod:

```
type=RedisClient  host_ip=10.0.223.176  host_name=router
                  peer_ip=10.0.159.101  peer_name=notification-service
```

For contrast, a Java Redis client on the same node in the same window resolves correctly:

```
type=RedisClient  host_ip=10.0.243.199  host_name=10.0.243.199
                  peer_ip=10.0.159.105  peer_name=user-service
```

### (Tentative) Root cause analysis

**1. There is no handshake-time SSL→socket binding.** OBI records the `SSL*` on the
`SSL_read`/`SSL_write` uprobe and binds it only if a socket syscall fires while still
inside that uprobe/uretprobe window — `bpf/common/ssl_helpers.h:29-47`, whose comment
calls it "sandwitched between read or write uprobe/uretprobe", reached from
`bpf/generictracer/k_tracer.c:451`. `bpf/generictracer/libssl.c` probes only `SSL_read`,
`SSL_read_ex`, `SSL_write`, `SSL_write_ex`, `SSL_write_ex2` and `SSL_shutdown`; there is no
`SSL_set_fd`, `SSL_set_bio` or `SSL_do_handshake` probe anywhere under `bpf/`. The only
mentions are comments acknowledging the gap
(`bpf/generictracer/maps/pid_tid_to_conn.h:14`: "in case we miss SSL_do_handshake").

Any TLS stack that encrypts into a buffer and writes it later can miss this window.

**2. The fallback guesses, and the guess is latched forever.**
`bpf/generictracer/ssl_defs.h:85` falls back to `pid_tid_to_conn`, keyed on the thread. On
an event-loop thread the connection recorded there is typically the inbound request being
served. `ssl_defs.h:~114` then writes that tuple into `ssl_to_conn` for the remaining life
of the `SSL*`, and nothing re-evaluates it.

**3. The wrong tuple also breaks request/response pairing.** With the tuple wrong, the
opening `SSL_write` is dropped by the server-fd guard at
`bpf/generictracer/protocol_tcp.h:311-321`, so each emitted span pairs command *N+1* with
command *N*'s reply, and userspace `reverseTCPEvent`
(`pkg/ebpf/common/tcp_detect_transform.go:427`) yields a `RedisClient` span carrying the
router's ephemeral port. The reported "duration" is an inter-command idle gap, which is
why the range is 0.13 ms – 119 s. This survives fixing the tuple and needs its own fix.

### Impact

- Fabricated service-graph edges (here `notification-service -> router`, inverting the
  real edge), while the true dependency disappears from the graph.
- Unusable latency data for affected connections.

### Suggested direction

- **Do not latch a guess as if it were a fact.** `ssl_pid_connection_info_t`
  (`connection_info.h:57-61`) has a spare `u8 _pad[6]`, so a `tentative` provenance bit
  costs no size change: set it on fallback-derived latches and re-evaluate on the next
  `TCP_RECV`, so a wrongly write-latched connection self-corrects on its next read rather
  than staying wrong for the connection's life. Also move the `bytes_len <= 0` bail-out
  *above* the latch — a `WANT_READ` zero-byte read, which asyncio does constantly, can
  currently pin a connection permanently. Caveat: mid-connection correction can orphan one
  exchange across two `ongoing_tcp_req` keys.
- **Do not report a fabricated duration** for a reversed event.

The durable fix is probably a handshake-time binding (`SSL_set_fd` / `SSL_set_bio`) so
memory-BIO connections bind at all, rather than relying on the syscall sandwich.
**Question for the maintainers:** Has that been considered already?

`bpf/tests/bpf_ssl_read_guard.c` already mocks the relevant maps including `ongoing_http`,
and `pkg/ebpf/common/redis_detect_transform_test.go` covers the userspace half.

### What I could not determine

Which operation *takes* the initial latch. The capture shows an already-latched `SSL*`, not
a fresh one, so I cannot say from observation whether the first fallback happened on a
write or a read. It matters less than it first appeared: evidence 5 shows a subsequent read
does not repair the entry, so the defect is the absence of re-evaluation rather than the
direction of the initial race.

Relatedly, why one client mis-latches where another does not. Within a single pod pair —
Python 3.14 / uvicorn / uvloop 0.22.1 for both — `aiohttp` HTTPS is attributed correctly
while `redis.asyncio` is not, and I could not establish what makes that deterministic
rather than chance. The latch-without-revisit defect holds either way.

The `bpf_dbg_printk` route above is a cheap reproduction path for a maintainer: one `SSL*`,
one pod, and the wrong tuple is visible in three lines.

### Version / environment

- OBI `v0.10.0` as vendored into Grafana Beyla v3.31.0.
- **Verified against current upstream `main`:** `pkg/ebpf/common/redis_detect_transform.go`
  and `pkg/transform/name_resolver.go` are byte-identical to `main`, so this is not fixed
  by [#2663](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2663),
  which is already included. The eBPF files cited were read from the same tree.
- Kubernetes 1.34 on AWS EKS Auto Mode, Bottlerocket, arm64 (Graviton), kernel 6.18.38.
- Affected observed in: Python 3.14 / uvicorn / uvloop / `redis==8.1.0` (`redis.asyncio`)
  over TLS to ElastiCache Valkey; Node.js services over TLS to DynamoDB and SSM.
- Not observed (in sampling) in: Java Valkey clients, Go clients, Python `aiohttp` HTTPS.
- Config: `name_resolver.sources: [k8s, rdns]`,
  `BEYLA_BPF_CONTEXT_PROPAGATION=headers`, discovery scoped to one namespace.

### AI assistance disclosure

Per the project's Generative AI Policy: this report was drafted with AI assistance. The
source citations were verified by hand, and all cluster evidence above was captured live —
including a DEBUG window explicitly correlated against a known Redis span. Nothing was
compiled or run against a test build.

KHARSHAVARDHAN-eng:
Fixes #2998

### Root Cause
TLS client spans in event-loop applications (e.g. `asyncio`, `Node.js`) were mis-attributed to inbound server sockets because:
1. Zero-byte reads (`WANT_READ`) looked up thread fallback connections (`pid_tid_to_conn`) before checking `bytes_len <= 0`.
2. Once saved into `ssl_to_conn`, fallback guesses were never re-evaluated and remained latched permanently.

### Fix
- Added a `tentative` provenance bit to `ssl_pid_connection_info_t` to differentiate fallback guesses (`tentative = 1`) from confirmed socket bindings (`tentative = 0`).
- Returned early on `bytes_len <= 0` in `handle_ssl_buf` to prevent zero-byte reads from creating or updating `ssl_to_conn`.
- Allowed re-evaluating `ssl_to_conn` when `conn->tentative == 1` so tentative guesses update when confirmed socket info becomes available.

### DIFF
--- a/bpf/common/connection_info.h
+++ b/bpf/common/connection_info.h
@@ -61,7 +61,8 @@ typedef struct http_pid_connection_info {
 typedef struct ssl_pid_connection_info {
     pid_connection_info_t p_conn;
     u16 orig_dport;
-    u8 _pad[6];
+    u8 tentative;
+    u8 _pad[5];
 } ssl_pid_connection_info_t;
 
 typedef struct connection_info_part {
--- a/bpf/common/ssl_helpers.h
+++ b/bpf/common/ssl_helpers.h
@@ -22,6 +22,7 @@ static __always_inline void set_active_ssl_connection(ssl_pid_connection_info_t 
     bpf_dbg_printk("Correlating SSL %llx to connection", ssl);
     dbg_print_http_connection_info(&ssl_conn->p_conn.conn);
 
+    ssl_conn->tentative = 0;
     bpf_map_update_elem(&active_ssl_connections, &ssl_conn->p_conn, &ssl, BPF_ANY);
     bpf_map_update_elem(&ssl_to_conn, &ssl, ssl_conn, BPF_ANY);
 }
--- a/bpf/generictracer/k_tracer.c
+++ b/bpf/generictracer/k_tracer.c
@@ -407,6 +407,7 @@ setup_connection_to_pid_mapping(u64 id, pid_connection_info_t *p_conn, u16 orig_
         ssl_pid_connection_info_t ssl_conn = {0};
         ssl_conn.orig_dport = orig_dport;
         ssl_conn.p_conn = *p_conn;
+        ssl_conn.tentative = 1;
 
         bpf_map_update_elem(&pid_tid_to_conn, &id, &ssl_conn, BPF_ANY);
     }
--- a/bpf/generictracer/ssl_defs.h
+++ b/bpf/generictracer/ssl_defs.h
@@ -34,12 +34,19 @@ handle_ssl_buf(void *ctx, u64 id, ssl_args_t *args, int bytes_len, u8 direction)
         void *ssl = ((void *)args->ssl);
         const u64 ssl_ptr = (u64)ssl;
         bpf_dbg_printk("SSL_buf id=%d ssl=%llx", id, ssl);
+
+        bpf_map_delete_elem(&ssl_to_pid_tid, &ssl_ptr);
+
+        if (bytes_len <= 0) {
+            return;
+        }
+
         ssl_pid_connection_info_t *conn = bpf_map_lookup_elem(&ssl_to_conn, &ssl);
 
-        if (!conn) {
-            conn = bpf_map_lookup_elem(&pid_tid_to_conn, &id);
+        if (!conn || conn->tentative) {
+            ssl_pid_connection_info_t *fallback_conn = bpf_map_lookup_elem(&pid_tid_to_conn, &id);
 
-            if (!conn) {
+            if (!fallback_conn) {
                 // We try even harder, we might have an SSL pointer mapped on another
                 // thread, since tcp_rcv_established was handled on another thread pool.
                 // First we look up a pid_tid by the ssl pointer, which might've been established
@@ -50,9 +57,9 @@ handle_ssl_buf(void *ctx, u64 id, ssl_args_t *args, int bytes_len, u8 direction)
                 if (pid_tid_ptr) {
                     const u64 pid_tid = *pid_tid_ptr;
 
-                    conn = bpf_map_lookup_elem(&pid_tid_to_conn, &pid_tid);
+                    fallback_conn = bpf_map_lookup_elem(&pid_tid_to_conn, &pid_tid);
                     bpf_dbg_printk(
-                        "Separate pool lookup ssl=%llx, pid=%d, conn=%llx", ssl_ptr, pid_tid, conn);
+                        "Separate pool lookup ssl=%llx, pid=%d, conn=%llx", ssl_ptr, pid_tid, fallback_conn);
                 } else {
                     bpf_dbg_printk("Other thread lookup failed for ssl=%llx", ssl_ptr);
                 }
@@ -62,18 +69,14 @@ handle_ssl_buf(void *ctx, u64 id, ssl_args_t *args, int bytes_len, u8 direction)
             // we missed a SSL_do_handshake, update our ssl to connection map to be
             // used by the rest of the SSL lifecycle. We shouldn't rely on the SSL_write
             // being on the same thread as the SSL_read.
-            if (conn) {
+            if (fallback_conn) {
                 bpf_map_delete_elem(&pid_tid_to_conn, &id);
                 ssl_pid_connection_info_t c;
-                bpf_probe_read(&c, sizeof(ssl_pid_connection_info_t), conn);
+                bpf_probe_read(&c, sizeof(ssl_pid_connection_info_t), fallback_conn);
+                c.tentative = 1;
                 bpf_map_update_elem(&ssl_to_conn, &ssl, &c, BPF_ANY);
+                conn = bpf_map_lookup_elem(&ssl_to_conn, &ssl);
             }
-        }
-
-        bpf_map_delete_elem(&ssl_to_pid_tid, &ssl_ptr);
-
-        if (bytes_len <= 0) {
-            return;
         }
 
         if (!conn) {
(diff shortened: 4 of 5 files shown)


marctc:
I’m closing this PR because it does not comply with our contribution and Generative AI policies. The required checklist was not completed.  Before submitting another PR, please:

  - Read and follow CONTRIBUTING.md and AI-POLICY.md.
  - Ensure you understand and can explain every part of the change.
  - Reproduce the issue and provide a failing-before/passing-after test that exercises the actual connection replacement.

This project values contributors who engage with maintainers, respond to feedback, and take ownership of their changes. Fully automated, unverified drive-by contributions are not acceptable.

PR #3001, opened 2026-08-12
fix: cp: stop answered server requests from parenting later client calls

mmat11:
## Summary

When an app handles an incoming request, OBI remembers it so any outgoing calls
made while serving it get nested under it. That memory isn't dropped when the
response is sent — only later, on further activity on that connection or when it
closes. Nothing checked whether the request had already been answered, so
unrelated outgoing calls kept getting attached to it.

A [vmagent](https://docs.victoriametrics.com/victoriametrics/vmagent/) scraping metrics shows it plainly. One incoming `/metrics` request,
answered in 2ms, then 99 unrelated outgoing scrapes nested under it — 53 different
targets over 39 seconds, on an inbound connection that just sat idle.
If that incoming span is never exported, every span in the trace has a missing parent and the trace shows
no root at all.

Fix: don't nest a call under a server request that has already answered. The
`response_sent` flag already exists and is already set when a server response is
sent on the generic path — the parent lookup just never looked at it.

New test `TestServerTraceNotLatchedOntoLaterClientCalls`: a server that answers,
then calls three targets. Nothing is being served at that point, so nesting is
always wrong. Fails on main, passes here.

## Validation


grcevski (approved):
Cool! LGTM! I think this will work and it's a nice safety net for a mistake that happens at eBPF time.

I'd just make the maps configurable and perhaps tied somehow to the concurrency factor we currently use for the BPF maps.

grcevski:
I requested copilot to check if there are any gotchas I missed at line by line level. But overall I like the design.

pkg/ebpf/common/http_transform.go
```diff
@@ -295,6 +297,9 @@ func postProcessHTTPSpan(parseCtx *EBPFParseContext, httpSpan *request.Span, req
 	return *httpSpan
 }
 
+// mirrors the k_parent_status_* enum in bpf/common/trace_parent.h
+const parentStatusConditional = 2
```
pinoOgni, line 301: nit: add this
```
_Static_assert(k_parent_status_conditional == 2, "value mirrored in pkg/ebpf/common (parentStatusConditional)");
```
in `tp_info.h`

pinoOgni (approved):
lgtm!

Issue #2974, opened 2026-08-10
Proposal: single source of truth for metric declaration and export


NimrodAvni78:
## Proposal

OBI exports metrics through two independent code paths: the OTLP exporter
(`pkg/export/otel`) and a Prometheus exporter (`pkg/export/prom`, ~2.9k lines of
non-test code plus ~3.5k lines of tests). Each maintains its own metric names, label sets and
config surface, and they have quietly drifted apart.

Maintaining both in parallel can be a burden, and anyone wanting Prometheus output can already get it by pointing OBI's OTLP exporter at a collector running `prometheusexporter`. 
But on the other hand prometheus is a very popular metrics engine, which we already support, and some users might prefer the pull-based method instead of push. 

This issue does not proposes deprecating / eventually removing the
built-in Prometheus exporter.
But to streamline the underlying code architecture to maintain 1 set of otel metrics, validated and built. but just the existing OTEL infrastructure to transform those to Prometheus metrics at export time

The rest of this issue is what we found while checking whether that is actually correct to do so.

## OTLP <> Prometheus comparison

Every metric OBI can emit was audited, not just the ones in the shared name table. That is
45 definitions in `pkg/export/attributes/metric.go` plus 30 with hard-coded names — span
metrics in both `application_span` and `application_span_otel` modes, service graph, the
three info metrics, and all 17 internal metrics. Each was pushed through
`github.com/prometheus/otlptranslator` — the library `prometheusexporter` uses — and the
result compared to the name OBI's Prometheus exporter actually emits.

**69 of 75 map exactly.** All RED metrics, span metrics, service graph, Go runtime, JVM,
network flows, DNS, GenAI, target/host info and every internal metric round-trip, including
`_seconds`/`_bytes` unit suffixes, `_total` de-duplication, `%`→`percent`, and `{braces}`
units correctly suppressing a suffix. The 6 exceptions are below.

`job` and `instance` also survive exactly: the collector derives
`job = service.namespace/service.name` and `instance = service.instance.id`, matching OBI's
`Attrs.Job()` and `UID.Instance`.

### The 6 that diverge

| OTLP name | unit | instrument | collector derives | OBI Prometheus emits |
|---|---|---|---|---|
| `gpu.cuda.kernel.grid.size` | `1` | histogram | `gpu_cuda_kernel_grid_size` | `gpu_cuda_kernel_grid_size_total` |
[...]

PR #3003, opened 2026-08-12
fix(metrics): sanitize invalid UTF-8 in OTLP attribute values

NimrodAvni78:

## Summary

`sanitizeUTF8ForPrometheus` strips invalid UTF-8 from attribute values, but it is applied only
on the Prometheus path, in `labelValues`. Invalid UTF-8 has previously caused a panic there,
which is why the guard exists.

The same values also go out over OTLP, where nothing strips them. Protobuf string fields must
be valid UTF-8, so marshalling rejects the payload and **the entire metrics export batch is
dropped** — not just the offending datapoint. The Prometheus path has been masking this,
and it becomes the only behaviour once that path is deprecated.

This moves the helper to `attributes.SanitizeUTF8`, shared by both exporters, and applies it
on the OTLP path in `Expirer.recordAttributes` — the single point every span-derived datapoint
attribute passes through:

```go
func sanitizeKeyValue(kv attribute.KeyValue) attribute.KeyValue {
	if kv.Value.Type() != attribute.STRING {
		return kv
	}
	raw := kv.Value.AsString()
	sanitized := attributes.SanitizeUTF8(raw)
	if sanitized == raw {
		return kv
	}
	return kv.Key.String(sanitized)
}
```

Non-string attributes and already-valid strings return the original value unchanged, so the
common path allocates nothing extra.

## Scope

Metric datapoint attributes only. Two related surfaces are deliberately left alone:

- **Resource attributes** (`otelcfg/common.go`) come from Kubernetes and container metadata
  rather than parsed protocol bytes, so they are far less likely to carry invalid UTF-8.
- **Trace span attributes** have the same exposure and no equivalent guard. Worth a follow-up,
  but it is a different code path and does not belong in this change.

Note the fix stays necessary after the exporters are consolidated: it guards the OTLP payload
itself, not the Prometheus rendering.

## Validation


pkg/export/otel/expirer.go
```diff
@@ -95,21 +95,52 @@ func (ex *Expirer[Record, Metric, ValType]) recordAttributes(m Record, extraAttr
...
-		vals = append(vals, extraAttrs[i].Value.Emit())
+		kv := sanitizeKeyValue(extraAttrs[i])
+		keyVals = append(keyVals, kv)
+		vals = append(vals, kv.Value.Emit())
 	}
 
 	return attribute.NewSet(keyVals...), vals
 }
 
+func sanitizeKeyValue(kv attribute.KeyValue) attribute.KeyValue {
```
MrAlias, line 114: `sanitizeKeyValue` is only called from `Expirer.recordAttributes` and the JVM helper, so direct OTLP metric paths still bypass it. `InternalMetricsReporter.InstrumentProcess` records `process.executable.name` from the raw executable basename; Linux filenames can contain invalid UTF-8, which can then poison every internal-metrics export batch while that series remains aggregated.

Could we apply the shared sanitizer at a common OTLP metric-attribute boundary, or cover every direct datapoint path in this scope?

pkg/export/otel/expirer_test.go
```diff
@@ -419,6 +419,63 @@ func TestAppMetricsExpiration_BySvcID(t *testing.T) {
 	}, timeout, 100*time.Millisecond)
 }
 
+// Invalid UTF-8 in an attribute value makes protobuf marshalling of the whole
```
MrAlias: Both lint jobs reject `marshalling` here; the repository's `misspell` check requires `marshaling`.

```suggestion
// Invalid UTF-8 in an attribute value makes protobuf marshaling of the whole
```

MrAlias (commented):
The sanitization boundary is still incomplete across OTLP metric providers, so the same batch-poisoning failure remains reachable outside the main application-metric path.

grcevski (approved):
LGTM!

grcevski:
You need merge with main to get rid of the Java integration test failure.

MrAlias (approved):
🚀

PR #3004, opened 2026-08-12
Adds OBI's emitted-telemetry schema

NimrodAvni78:

## Summary

Adds OBI's emitted-telemetry schema: a semantic-convention registry that declares exactly the metrics and attributes OBI produces over OTLP.

This is a prerequisite to full telemetry schema coverage validation. It is **schema only** — nothing consumes it yet. Reviewing this one is purely "is this the correct description of what OBI emits?"

**What's here**

- `emitted_metrics.yaml` (new): every OBI-emitted OTLP metric and the attributes  it carries, 
narrowed to what OBI actually sets. 
This lets the registry resolve to OBI's true OTLP contract instead of the full upstream surface.
- `obi_internal.yaml`: `target.info` now references the resource attributes OBI attaches to it (`service.instance.id` and the `k8s.*` set), so the  resource-attribute carrier is described rather than empty.
- `dns.yaml`: imports only `dns.lookup.duration` instead of the whole upstream DNS group.
- `README.md`: corrects a stale note on how the group files import and reference upstream definitions.

**Why the metrics are fully re-declared**

`emitted_metrics.yaml` restates each upstream metric with its full attribute list rather than importing it and refining the attributes down to OBI's subset.
That looks like duplication, but it is currently the only option. This is tracked in [open-telemetry/weaver#1667](https://github.com/open-telemetry/weaver/issues/1667)
Until then we redeclare the otel metric defenitions and ref only the attributes we emit
we validate that our schema does not drift from the upstream otel one with `TestOBIMetricOverridesMatchUpstream`

## Validation

### DIFF
--- a/internal/schemacheck/upstream_consistency_test.go
+++ b/internal/schemacheck/upstream_consistency_test.go
@@ -1,0 +1,120 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package schemacheck holds tests that validate the OBI semantic-convention
+// registry against the pinned upstream OpenTelemetry semantic conventions it
+// depends on.
+package schemacheck
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	obiGroupsDir = "../../schemas/obi/groups"
+	upstreamDeps = "../../schemas/obi/.deps"
+)
+
+type metricGroupsFile struct {
+	Groups []struct {
+		Type       string `yaml:"type"`
+		MetricName string `yaml:"metric_name"`
+		Unit       string `yaml:"unit"`
+		Instrument string `yaml:"instrument"`
+		Stability  string `yaml:"stability"`
+	} `yaml:"groups"`
+}
+
+type metricDef struct {
+	unit       string
+	instrument string
+	stability  string
+	source     string
+}
+
+func metricsFromFile(t *testing.T, path string, out map[string]metricDef) {
+	t.Helper()
+	body, err := os.ReadFile(path)
+	require.NoError(t, err)
+	var f metricGroupsFile
+	require.NoErrorf(t, yaml.Unmarshal(body, &f), "parsing %s", path)
+	for _, g := range f.Groups {
+		if g.Type != "metric" || g.MetricName == "" {
+			continue
+		}
+		out[g.MetricName] = metricDef{unit: g.Unit, instrument: g.Instrument, stability: g.Stability, source: path}
+	}
+}
+
+func obiMetrics(t *testing.T) map[string]metricDef {
+	t.Helper()
+	entries, err := os.ReadDir(obiGroupsDir)
+	require.NoError(t, err)
+	out := map[string]metricDef{}
+	for _, e := range entries {
+		if e.IsDir() || filepath.Ext(e.Name()) != ".yaml" {
+			continue
+		}
+		metricsFromFile(t, filepath.Join(obiGroupsDir, e.Name()), out)
+	}
+	return out
+}
+
+func upstreamMetrics(t *testing.T) map[string]metricDef {
+	t.Helper()
+	out := map[string]metricDef{}
+	err := filepath.WalkDir(upstreamDeps, func(path string, d os.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() || filepath.Ext(path) != ".yaml" {
+			return nil
+		}
+		metricsFromFile(t, path, out)
+		return nil
+	})
+	require.NoError(t, err)
+	return out
+}
+
+// TestOBIMetricOverridesMatchUpstream asserts that every OBI metric reusing an
+// upstream semconv metric name declares the same unit, instrument, and stability
+// as the pinned upstream definition.
+//
+// OBI redeclares these metrics instead of importing them because weaver cannot
+// narrow an imported metric's attributes down to the subset OBI emits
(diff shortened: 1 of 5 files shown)


schemas/obi/groups/emitted_metrics.yaml
```diff
@@ -0,0 +1,446 @@
+# OBI-emitted semconv metrics, narrowed to exactly the attributes OBI sets.
+# Each metric ref-s only what OBI emits so the registry resolves (without
+# --include-unreferenced) to OBI's true OTLP contract. Kubernetes attributes
+# are added once their upstream declarations land; per-suite validation keeps
+# --include-unreferenced so forwarded pass-through attributes are tolerated.
+groups:
+  - id: metric.obi.db.client.operation.duration
```
MrAlias, line 7: These standalone `metric.obi.*` groups are not the definitions live-check resolves. With the integration suites' `--include-unreferenced` flag, representative DB, HTTP, and RPC names resolve to the pinned upstream groups and their broader attribute sets; without it, this registry resolves no metrics at all. The problem is reachability, not the OBI namespace: these definitions should remain under an OBI-owned group ID.

Could we wire the namespaced definitions into the registry path that live-check actually resolves, and add a test verifying each standard metric name resolves to OBI's narrowed attributes and local provenance rather than the upstream definition?

schemas/obi/groups/emitted_metrics.yaml
```diff
@@ -0,0 +1,446 @@
...
+    type: metric
+    metric_name: db.client.operation.duration
+    brief: OBI-emitted db.client.operation.duration
+    instrument: histogram
+    unit: "s"
+    stability: stable
+    attributes:
+      - ref: container.id
```
MrAlias, line 15: OBI already emits `container.id` and `container.name` on the metric resource, not on individual data points. The metric `attributes` list models the data-point attribute set; `OpenTelemetryGetters` drops these names because `spanOTELGetters` intentionally has no container cases. Including the refs here—and in the remaining application groups—therefore misrepresents the existing OTLP output; the exporter does not need to change.

Could we remove these refs from the DB, GenAI, HTTP, messaging, and RPC metric declarations?

schemas/obi/groups/emitted_metrics.yaml
```diff
@@ -0,0 +1,446 @@
...
+      - ref: service.namespace
+  - id: metric.obi.go.config.gogc
+    type: metric
+    metric_name: go.config.gogc
+    brief: OBI-emitted go.config.gogc
+    instrument: updowncounter
+    unit: "%"
+    stability: development
+    attributes:
+      - ref: container.id
```
MrAlias, line 70: Runtime service/container metadata is attached to the meter resource by `RuntimeMetricsReporter.newMetricsInstance`; record calls add only runtime-specific data-point attributes. The Go/JVM groups therefore should not list container/service refs, and `go.cpu.time` needs the reachable `go.cpu.detailed_state` attribute emitted by `recordGoRuntimeCPUTime`.

Could we model only those point attributes across the runtime groups?

schemas/obi/groups/obi_internal.yaml
```diff
@@ -73,6 +73,20 @@ groups:
     attributes:
       - ref: instance
       - ref: job
+      - ref: service.instance.id
```
MrAlias, line 76: This addition still leaves the carrier contract incomplete. Both `target.info` and `traces.target.info` are created on the empty-resource system meter, then receive the service resource sets through `WithAttributeSet`; those values are data-point attributes here. `target.info` omits the built-in service, telemetry SDK/distro, host, OS, and reachable metadata fields, while `traces.target.info` declares only `source`.

Could we enumerate the native carrier sets for both metrics and keep only arbitrary user pass-through under the existing tolerance?

internal/schemacheck/upstream_consistency_test.go
```diff
@@ -0,0 +1,120 @@
...
+	obi := obiMetrics(t)
+	upstream := upstreamMetrics(t)
+	require.NotEmpty(t, obi)
+	require.NotEmpty(t, upstream)
+
+	overrides := 0
+	for name, m := range obi {
+		up, ok := upstream[name]
+		if !ok {
+			continue // OBI-custom metric with no upstream definition to match
```
MrAlias, line 106: An emitted override that disappears from the pinned upstream registry is silently reclassified as custom here. A typo or upstream rename in `emitted_metrics.yaml` therefore skips this drift check, while `require.Positive(overrides)` still passes through any other match.

Could we load that override file separately and require every metric name in it to exist upstream, leaving this skip only for genuinely custom metrics?

schemas/obi/groups/emitted_metrics.yaml
NimrodAvni78: after looking at this a bit more i see its quite a large change
I only worked on refining metric definitions with the attributes they emit, but we also need to add resource and spans. all that will basically make us redeclare a large share of the upstream semconv. and also need to test that we are not diverging from it on known metrics. ill try to do this in this PR if i see its a large change then it might need its own PR

schemas/obi/groups/emitted_metrics.yaml
NimrodAvni78: yes you are right, these need to be under a seperate resource object

schemas/obi/groups/emitted_metrics.yaml
NimrodAvni78: same

schemas/obi/groups/obi_internal.yaml
NimrodAvni78: this is actually on purpose, target.info is a metric with no resource which carries all of its resource data on attributes, its mainly used from prometheus style joins. 
discussed on https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2974

NimrodAvni78:
Closing this in favor of #3219

PR #3005, opened 2026-08-12
fix(metrics)!: derive Prometheus metric names from their OTLP definition

NimrodAvni78:

## Summary

OBI declares each metric's name twice: once as an OTLP name and once as a Prometheus name, in
`pkg/export/attributes/metric.go`. The two have drifted apart, so a metric scraped from OBI's
own Prometheus endpoint does not always have the same name as the same metric routed through
OTLP and re-exported in Prometheus format by a collector.

This removes the second declaration. `attributes.Name` now carries `OTEL`, `Unit` and `Type`,
and derives `Prom` from them using `github.com/prometheus/otlptranslator` — the same library
the collector's `prometheusexporter` uses:

```go
HTTPServerDuration = metric(Name{
    Section: "http.server.request.duration",
    OTEL:    "http.server.request.duration",
    Unit:    "s",
    Type:    Histogram,
})
```

Units are no longer restated at the instrument call sites either — they read
`instrument.WithUnit(attributes.HTTPServerDuration.Unit)` instead of a literal — so a metric's
name and unit are each declared exactly once. A name that disagrees with its OTLP form is no
longer expressible, and a derivation failure panics at package init so it cannot reach a
release.

`github.com/prometheus/otlptranslator` was already in the module graph as an indirect
dependency and becomes a direct one. Note this makes it a **runtime** dependency of
`pkg/export/attributes`, not test-only. `go.sum` is unchanged.

### Metric renames

Deriving the names surfaced 6 metrics whose hand-written Prometheus name did not match what a
Prometheus consumer of the OTLP metric produces. Five are cases where the hand-written name
violated Prometheus naming conventions:

| OTLP metric | before | after |
|---|---|---|
| `gpu.cuda.kernel.grid.size` | `gpu_cuda_kernel_grid_size_total` | `gpu_cuda_kernel_grid_size` |
| `gpu.cuda.kernel.block.size` | `gpu_cuda_kernel_block_size_total` | `gpu_cuda_kernel_block_size` |
| `gpu.cuda.memory.copies` | `gpu_cuda_memory_copies_bytes_total` | `gpu_cuda_memory_copies_bytes` |
| `obi.stat.tcp.retransmits` | `obi_stat_tcp_retransmits` | `obi_stat_tcp_retransmits_total` |
| `obi.stat.tcp.failed.connections` | `obi_stat_tcp_failed_connections` | `obi_stat_tcp_failed_connections_total` |

`_total` on a histogram is meaningless and also produced `..._total_bucket`; the two TCP stat
counters were missing the `_total` suffix the convention requires.

### Unit corrections

- `gpu.cuda.memory.copies` declared `WithUnit("1")` while its Prometheus name claimed
  `_bytes`. It records `span.ContentLength`, so the unit is now `By` and the Prometheus name
  is `gpu_cuda_memory_copies_bytes`. `schemas/obi/groups/gpu.yaml` is updated to match, since
  weaver validates declared units against the OTLP ones.
- `obi.queue.capacity.ratio` passed a full sentence to `instrument.WithUnit()` instead of
  `WithDescription()`, so the OTLP `unit` field literally read *"Ratio [0-1] between the
  unread messages of an internal Go channel and its total capacity"*. Any Prometheus
  translation appends the unit to the metric name, which turned this into a 90-character
  metric name. Now `WithDescription(...)` plus `WithUnit("1")`.

## Why these names

Two existing integration suites already assert the collector-derived names rather than the
ones OBI's Prometheus exporter emitted:

- `internal/test/integration/configs/prometheus-config-promscrape-stat.yml` deliberately
  scrapes the collector's exporter (`otelcol:9464`) instead of OBI's own `:8999`, and
  `internal/test/integration/red_test_stat_metrics.go` asserts
  `obi_stat_tcp_failed_connections_total` and `obi_stat_tcp_retransmits_total` — with the
  `_total` suffix OBI's Prometheus exporter did not emit.
- `internal/test/integration/gpu_metrics_test.go` asserts `gpu_cuda_kernel_grid_size`,
  `gpu_cuda_kernel_block_size` and `gpu_cuda_memory_copies` — without the `_total` suffix
  OBI's Prometheus exporter did emit.

The test suite already treated the OTLP-derived names as canonical; the Prometheus names were
the outliers. This moves them onto the names already being asserted.

## Breaking changes

The five renames affect anyone querying those series from OBI's own Prometheus endpoint. Users
consuming these metrics through a collector are unaffected — they were already receiving the
new names.

Relabel rules that rename the new series back to the previous names, for dashboards that
cannot be updated at the same time. Prometheus relabeling renames rather than duplicates, so
this restores the old names instead of exposing both. The three GPU metrics are histograms, so
the rules carry the `_bucket`/`_sum`/`_count` suffixes through:

```yaml
metric_relabel_configs:
  - source_labels: [__name__]
    regex: 'gpu_cuda_kernel_grid_size(.*)'
    target_label: __name__
    replacement: 'gpu_cuda_kernel_grid_size_total${1}'
  - source_labels: [__name__]
    regex: 'gpu_cuda_kernel_block_size(.*)'
    target_label: __name__
    replacement: 'gpu_cuda_kernel_block_size_total${1}'
  - source_labels: [__name__]
    regex: 'gpu_cuda_memory_copies_bytes(.*)'
    target_label: __name__
    replacement: 'gpu_cuda_memory_copies_bytes_total${1}'
  - source_labels: [__name__]
    regex: 'obi_stat_tcp_retransmits_total'
    target_label: __name__
    replacement: 'obi_stat_tcp_retransmits'
  - source_labels: [__name__]
    regex: 'obi_stat_tcp_failed_connections_total'
    target_label: __name__
    replacement: 'obi_stat_tcp_failed_connections'
```

## Validation


MrAlias (approved):

pkg/export/attributes/attr_defs_test.go
```diff
@@ -74,7 +74,9 @@ func TestGoRuntimeDefinitions(t *testing.T) {
 	definitions := getDefinitions(0, NewGroupAttributes(nil))
 	for _, test := range tests {
 		t.Run(test.want.OTEL, func(t *testing.T) {
-			assert.Equal(t, test.want, test.name)
+			assert.Equal(t, test.want.Section, test.name.Section)
```
pinoOgni, line 85: What about `Unit` and `Type`?

schemas/obi/groups/gpu.yaml
```diff
@@ -84,7 +84,7 @@ groups:
     type: metric
     metric_name: gpu.cuda.memory.copies
     instrument: histogram
-    unit: "1"
+    unit: "By"
```
pinoOgni, line 87: Shouldn't this be a breaking change?

pinoOgni (approved):
lgtm!

schemas/obi/groups/gpu.yaml
NimrodAvni78: yes :)
noted in the breaking changes section

pkg/export/attributes/attr_defs_test.go
NimrodAvni78: Done

schemas/obi/groups/gpu.yaml
pinoOgni: Ops, I saw it only in the "Unit corrections" :)

PR #3006, opened 2026-08-12
gotracer: fix duplicate spans on wrapped net.Conn

mmat11:
## Summary

A Go client whose net.Conn is wrapped by a struct that doesn't hold the
connection first can't be read from the request goroutine, so we reported the
same call twice — once with no peer, once with the real address.

Test: a new suite drives vmagent scraping six targets once a second over
pooled keep-alive connections. It asserts both directions — no client span
with server.port 0, and at least five properly attributed spans per target —
so a "fix" that removes the duplicate by dropping the span fails too. Without
the fix it's a clean 1:1, every scrape yielding one span of each kind.

## Validation

### DIFF
--- a/bpf/gotracer/go_net.c
+++ b/bpf/gotracer/go_net.c
@@ -42,6 +42,7 @@
 #include <gotracer/types/net_args.h>
 #include <gotracer/types/nethttp.h>
 
+#include <gotracer/maps/go_persist_conn.h>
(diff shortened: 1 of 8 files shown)


grcevski (approved):
LGTM!

grcevski:
This needs merge with main to get rid the java failure

PR #3011, opened 2026-08-13
Honor context propagation modes in Go clients

### LINKED ISSUE #2276: Disabling context_propagation doesn't disable gRPC traceparent header injection (golang)

MrAlias:
## Summary

Go HTTP/1, HTTP/2, and gRPC clients could inject `traceparent` whenever the kernel supported `bpf_probe_write_user`, even when `context_propagation` was `disabled` or `tcp`. The same global also acted as the capability gate for Go Auto SDK activation.

This change separates those responsibilities:

- wire-header mutation and its dedicated probes require both write-user support and `headers` or `all`;
- `disabled` and `tcp` keep ordinary Go tracing probes but do not inject HTTP/gRPC headers;
- Go Auto SDK activation remains enabled whenever write-user support is available, including with the default disabled propagation mode;
- traceparent parsing behavior is unchanged.

Integration fixtures now state their intended propagation mode explicitly. Unit, verifier, and privileged persistent-connection gRPC coverage exercise the full configuration contract while preserving existing HTTP/2/gRPC safety behavior.

## Validation

- `make generate`
- `make clang-tidy`
- `GOFLAGS=-p=1 make verify`
- `GOFLAGS=-p=1 make build`
- full and race-enabled gotracer unit tests
- serial gotracer BPF verifier matrix
- privileged gRPC disabled/tcp/headers/all wire test
- Go Auto SDK activation, gRPC relay, nested traces, and Go HTTP/2 integration suites

Fixes #2276


pkg/internal/ebpf/gotracer/gotracer.go
```diff
@@ -431,22 +436,24 @@ func (p *Tracer) constants() map[string]any {
...
-		"attr_type_float64":         uint64(attribute.FLOAT64),
-		"attr_type_string":          uint64(attribute.STRING),
-		"attr_type_boolslice":       uint64(attribute.BOOLSLICE),
-		"attr_type_int64slice":      uint64(attribute.INT64SLICE),
-		"attr_type_float64slice":    uint64(attribute.FLOAT64SLICE),
-		"attr_type_stringslice":     uint64(attribute.STRINGSLICE),
-		"g_bpf_traceparent_enabled": true,
-		"g_bpf_loop_enabled":        p.supportsBPFLoop,
+		"g_bpf_debug":                    p.cfg.BpfDebug,
+		"g_bpf_header_propagation":       p.cfg.ContextPropagation.HasHeaders() && writeUserSupported,
```
grcevski: For unencrypted traffic, the wire header injection doesn't need probe write user, so this will limit Go specific context propagation to only platforms that support probe write user..

bpf/gotracer/go_sdk.c
```diff
@@ -424,7 +425,7 @@ int obi_uprobe_tracer_Start_global(struct pt_regs *ctx) {
 SEC("uprobe/tracer_new_span")
 int obi_uprobe_tracer_NewSpan(struct pt_regs *ctx) {
     u64 generation = 0;
-    if (!go_auto_target_generation(&generation) || !g_bpf_header_propagation ||
+    if (!go_auto_target_generation(&generation) || !g_bpf_probe_write_user_enabled ||
```
grcevski, line 428: I think some of our manual trace support can work without probe write user, what we can't do is change the tracer, but simple cases where there's just manual span use will work. I think this change will remove that ability.

pkg/internal/ebpf/gotracer/gotracer.go
MrAlias: Separated the two capabilities in 7b959dbcd. `g_bpf_header_propagation` now follows `headers`/`all` regardless of `bpf_probe_write_user`, so the Go probes keep the correlation state used by tpinjector for plaintext wire injection. Only the HTTP/1, HTTP/2, and gRPC return probes that mutate Go buffers check `g_bpf_probe_write_user_enabled`. The privileged verifier case passes with header propagation enabled and write-user support disabled.

bpf/gotracer/go_sdk.c
MrAlias: Manual spans remain available without `bpf_probe_write_user`. The ordinary global `tracer.Start` and `nonRecordingSpan.End` probes are unconditional in `GoProbes`; the gated `tracer.newSpan` probe is the process-scoped Go Auto SDK activation path that writes the auto-span flag. Added `TestManualSpanProbesDoNotRequireWriteUserSupport` to preserve that distinction.

pkg/internal/ebpf/gotracer/gotracer.go
```diff
@@ -431,22 +436,24 @@ func (p *Tracer) constants() map[string]any {
 		blackBoxCP = uint32(1)
 	}
 
+	writeUserSupported := p.supportsContextPropagation()
```
mmat11: nit: inline this

mmat11 (approved):

grcevski (approved):
LGTM!

PR #3016, opened 2026-08-13
feat(metrics): export eBPF probe latency as a pre-aggregated OTLP histogram

NimrodAvni78:
## Summary

`bpf_probe_latency_seconds` was the only metric on the BPF collector's surface that OBI could emit
over Prometheus but not OTLP. This adds `obi.bpf.probe.latency` (histogram, `s`) and removes
`obi.bpf.probe.executions` and `obi.bpf.probe.latency_seconds_total`, whose values are that
histogram's count and sum.

It uses the SDK `Producer` interface, like the existing `goRuntimeHistogramProducer`: the kernel
exposes only cumulative `run_time_ns` / `run_cnt`, so there are no individual observations to
`Record()`. Bounds come from the already-shared `imetrics.BpfLatenciesBuckets`, and temporality
follows the exporter's preference, with cumulative snapshots differenced when it asks for delta.

The distribution is an approximation: each interval contributes its whole execution count to the
single bucket matching that interval's *average* latency. The kernel only gives totals, but the
attribution to one bucket is OBI's own choice in `updateBuckets`, not a kernel constraint.

## Worth a reviewer's attention

- **`updateBuckets` was only called from the Prometheus `Collect()` path**, so an OTLP-only
  deployment accumulated no buckets at all. `collectInternalMetrics` now calls it. This cannot
  double-count: `BPFMetrics` builds separate `BPFCollector` instances for the two paths, each
  with its own `progs` map.
- **Observations above the largest bound are never recorded** by `updateBuckets`, so
  `bucketCounts` derives the trailing overflow bucket as `count - sum(recorded)`. Without it a
  histogram's buckets would not sum to its own count.

`imetrics.Reporter.BpfProbeStats` gained a `latencyBuckets` parameter; the producer clones it
because the caller keeps mutating it between intervals.

## Breaking changes

Two OTLP metrics removed, both internal meta-metrics about OBI itself:

- `obi.bpf.probe.executions` → the histogram's count (`obi_bpf_probe_latency_seconds_count`)
- `obi.bpf.probe.latency_seconds_total` → its sum (`obi_bpf_probe_latency_seconds_sum`)

Note the sum's Prometheus name changes too (`_sum`, not `_total`), so dashboards reading the total
latency need updating even though the value is preserved.

OBI's own Prometheus internal endpoint is unchanged; aligning `iprom.go` belongs to the
internal-metric reconciliation step. Known limitation, at parity with the counters it replaces:
per-probe state is never pruned, so a program ID that disappears keeps being exported with a
frozen count.

## Validation

### DIFF
--- a/devdocs/bpf-metrics-collection.md
+++ b/devdocs/bpf-metrics-collection.md
@@ -46,8 +46,8 @@ If both paths are active, each path owns a separate collector.
 
 | Prometheus metric | OpenTelemetry metric | Value |
 | --- | --- | --- |
-| `obi_bpf_probe_executions_total` | `obi.bpf.probe.executions` | Probe executions |
-| `obi_bpf_probe_latency_seconds_total` | `obi.bpf.probe.latency_seconds_total` | Total probe runtime in seconds |
+| `obi_bpf_probe_executions_total` | *(none: see `obi.bpf.probe.latency` count)* | Probe executions |
+| `obi_bpf_probe_latency_seconds_total` | `obi.bpf.probe.latency` | Probe latency distribution in seconds |
 | `obi_bpf_map_entries_total` | `obi.bpf.map.entries_total` | Current map entry count |
 | `obi_bpf_map_max_entries_total` | `obi.bpf.map.max_entries_total` | Configured maximum map entries |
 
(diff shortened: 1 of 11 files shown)


marctc (approved):
LGTM thanks!

PR #3019, opened 2026-08-13
Fix effective UID elevation for non-Java processes (#3018)

### LINKED ISSUE #3018: JavaInjector.NewExecutable calls Cleanup with myUID=0 for non-Java processes, elevating non-root agents to euid 0
### Component

OBI (eBPF instrumentation)

### What happened?

When OBI runs as a **non-root** process with ambient `CAP_SETUID`/`CAP_SETGID` and `javaagent.enabled=true`, instrumenting **any non-Java** process (C++, Python, etc.) permanently elevates the OBI agent's effective UID to `0`.
This happens because `JavaInjector.NewExecutable` is invoked for every instrumentable, creates a `JAttacher` whose `myUID`/`myGID` are still the Go zero value (`0`), defers `Cleanup()`, and only then gates the Java attach path. For non-Java processes the attach path is skipped, but the deferred cleanup still runs and executes `syscall.Seteuid(0)` / `syscall.Setegid(0)`.

Instrumenting a non-Java process must not change the OBI process credentials. Cleanup() should either not run, or restore the UID/GID captured by Init().

### Steps to reproduce

1. Run OBI (or an embedding agent that starts `instrumenter.Run` with a dynamic PID selector) as non-root with `CAP_SETUID`/`CAP_SETGID` and `Java.Enabled=true`.
2. Confirm agent identity before instrumentation:
   ```bash
   ps -o user=,pid=,cmd= -C <obi-or-agent-binary>
   # expect non-root, e.g. linux-user

### OBI configuration

```yaml

```

### Relevant logs

```shell

```

### Additional context

_No response_

### Version

_No response_

### Kernel version

_No response_

### OS / Distribution

_No response_

### Deployment

Bare metal / VM

KHARSHAVARDHAN-eng:
## Description
Fixes #3018.

When OBI runs as a non-root process with ambient capabilities (`CAP_SETUID`/`CAP_SETGID`) and `javaagent.enabled=true`, instrumenting any non-Java process (C++, Python, Go, Node, etc.) permanently elevates OBI's effective UID to 0.

This occurred because `JavaInjector.NewExecutable` created a `jvm.JAttacher` (whose `myUID`/`myGID` defaulted to `0`), registered a deferred `attacher.Cleanup()`, and only then checked `if ie.Type == svc.InstrumentableJava`. For non-Java processes, it skipped attach, but the deferred `attacher.Cleanup()` still ran and executed `syscall.Seteuid(0)` / `syscall.Setegid(0)`.

## Fix
1. Updated `JavaInjector.NewExecutable` in `pkg/internal/java/java_inject.go` to return early (`return nil`) when `ie.Type != svc.InstrumentableJava` before initializing `JAttacher` or deferring cleanup.
2. Updated `NewJAttacher` in `pkg/internal/jvmtools/jvm/cmd.go` to initialize `myUID` and `myGID` with `syscall.Geteuid()` and `syscall.Getegid()`.
3. Added unit tests for early return and attacher credential initialization.

### DIFF
--- a/bpf/common/connection_info.h
+++ b/bpf/common/connection_info.h
@@ -61,7 +61,8 @@ typedef struct http_pid_connection_info {
 typedef struct ssl_pid_connection_info {
     pid_connection_info_t p_conn;
     u16 orig_dport;
-    u8 _pad[6];
+    u8 tentative;
+    u8 _pad[5];
 } ssl_pid_connection_info_t;
 
 typedef struct connection_info_part {
--- a/bpf/common/ssl_helpers.h
+++ b/bpf/common/ssl_helpers.h
@@ -22,6 +22,7 @@ static __always_inline void set_active_ssl_connection(ssl_pid_connection_info_t 
     bpf_dbg_printk("Correlating SSL %llx to connection", ssl);
     dbg_print_http_connection_info(&ssl_conn->p_conn.conn);
 
+    ssl_conn->tentative = 0;
     bpf_map_update_elem(&active_ssl_connections, &ssl_conn->p_conn, &ssl, BPF_ANY);
     bpf_map_update_elem(&ssl_to_conn, &ssl, ssl_conn, BPF_ANY);
 }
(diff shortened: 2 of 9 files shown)


marctc:
The PR includes unrelated commits from another PR. Closing for the same reason that i pointed out [here](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3000#issuecomment-5266258083).

grcevski:
Also the changes related to the actual fix don't match what should happen. The issue claims that we do this for all executables, but the existing code clearly checks if the  process is Java.

@KHARSHAVARDHAN-eng this is not a way to engage with our community. If you'd like to work on this project, please take a look at the good first issues, understand the requirement and submit the PR after the fact. We do not accept pure AI generated PRs.

Issue #2769, opened 2026-07-23
Go tracer: prevent duplicate traceparent injection in HTTP/2 and GRPC


mariomac:
### Component

OBI (eBPF instrumentation)

### What happened?

Followup action for bug #2732 and PR #2752.

Pull request #2752 verifies whether an instrumented Golang client (using the `net` standard package) is already injecting the `Traceparent` header in HTTP/1.x. If so, we prevent OBI from injecting again another traceparent and potentially breaking the instrumentation.

We should extend this behaviour to other tracer implementations. In this case, fixing duplicate client-side traceparent injection in generic HTTP/2.x or GRPC tracers.

Unlike https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2276, which makes explicit reference to not injecting traceparents if context_propagation is disabled, this issue addresses the actual duplication of traceparents when context_propagation is enabled.

MrAlias:
## Proposed resolution plan

The best path is to supersede #3021 with a small vertical stack focused on the actual requirement in this issue: identify application-owned `traceparent` fields before HPACK serialization and prevent the existing injectors from adding another field to that stream.

This plan deliberately preserves today's injection behavior when the application did not provide `traceparent`. It does not introduce a new HTTP/2 mutation mechanism.

PR #3020 is already merged and provides the required exact-over-vendored Go symbol selection behavior.

### Shared design and invariants

All protocol and version slices will reuse one positive, exact application-ownership marker.

Conceptually, its key contains:

- process ID plus process start time, so PID reuse cannot inherit ownership;
- the full directional connection identity, without sorting away client/server direction;
- HTTP/2 stream ID.

The value should remain minimal: an ownership sentinel and, if needed, an observation time for bounded stale-entry cleanup. It should not contain application trace context, injection transaction states, audit data, or fallback controls.

The data flow is:

1. Observe an outgoing header field at a stable Go library seam before HPACK serialization.
2. Record only whether the application supplied a field named `traceparent`. Name comparison is case-insensitive; the value is neither copied nor validated.
3. When connection identity and stream ID are known, publish the exact per-stream marker into an internally pinned map shared by gotracer and tpinjector.
4. The existing Go Framer injector checks the marker and skips that stream without consuming the marker.
5. The existing plaintext `sk_msg` injector checks the same marker after parsing the HEADERS stream ID, skips the stream, and may consume the marker.
[...]

PR #3020, opened 2026-08-13
Prefer exact Go symbols over vendor aliases

MrAlias:
## Summary

- prefer the exact requested Go symbol over `/vendor/` aliases regardless of ELF traversal order
- retain a deterministic vendor alias fallback when the exact symbol is absent
- cover exact-only, alias-only, both discovery orders, and repeated aliases

## Why

Go binaries can contain both canonical `golang.org/x/net/http2` symbols and vendored aliases. Probe resolution previously kept whichever matching symbol happened to be discovered first, which could attach a probe to the wrong HTTP/2 implementation.

This establishes the symbol-selection invariant needed by the per-library HTTP/2 ownership probes in the follow-up draft while remaining independently useful to other Go probes.

## Validation

- `go test ./pkg/internal/goexec`
- `make compile`

Relates to #2769.

Stack: 1 of 2. Follow-up #3021 depends on this draft and contains the complete Go HTTP/2 and grpc-go ownership fix. Merge this draft first.

### DIFF
--- a/pkg/internal/goexec/instructions.go
+++ b/pkg/internal/goexec/instructions.go
@@ -83,12 +83,15 @@ func instrumentationPoints(elfF *elf.File, funcNames []string) (map[string]FuncO
 		allSyms, _ = procs.FindExeSymbols(elfF, funcNames)
 	}
 
+	// Prefer usable offsets from exact requested names over vendor aliases,
+	// regardless of symbol-table traversal order. At the same priority, keep the
+	// first usable offset found.
 	allOffsets := map[string]FuncOffsets{}
-	exactOffsets := map[string]bool{}
+	selectedIsExact := map[string]bool{}
 	for _, f := range symTab.Funcs {
-		fName, exact, ok := requestedFunctionName(f.Name, functions)
+		fName, isExact, ok := requestedFunctionName(f.Name, functions)
 		if ok {
-			if exactOffsets[fName] && !exact {
+			if selectedIsExact[fName] && !isExact {
 				continue
 			}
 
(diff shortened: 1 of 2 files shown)


grcevski (commented):
This makes sense, but I'm wondering, should we be attaching the probes then on both symbols at all times. In a sense, there may not be right or wrong symbol, could it be that both can be in use? So the right thing to do is attach to both?

MrAlias:
Yeah, good point. Both copies can be active, and this change does not provide complete coverage for that case. It retains the existing single-target contract while removing its symbol-order dependence. Supporting simultaneous copies requires multi-target resolution and corresponding attachment changes, including compatibility handling for each copy.

The paths needed specifically for #2769 are covered separately in #3021: `x/net` HTTP/2 and the stdlib HTTP/2 implementation use explicit symbols and separate probe groups. I’d prefer to keep this prerequisite scoped to deterministic single-target resolution and handle general multi-copy attachment separately. What do you think?

grcevski:
> The paths needed specifically for https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2769 are covered separately in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3021: x/net HTTP/2 and the stdlib HTTP/2 implementation use explicit symbols and separate probe groups. I’d prefer to keep this prerequisite scoped to deterministic single-target resolution and handle general multi-copy attachment separately. What do you think?

Sure, that makes sense. Make the single choice better and then we can follow-up with multiple probes support.

grcevski (approved):
LGTM!

Issue #2835, opened 2026-07-29
Define and enforce OBI/injector coexistence


MrAlias:
Parent epic: #2828

## Summary

Define and enforce OBI product behavior when standalone OBI runs alongside the OpenTelemetry injector or applications that already emit OpenTelemetry telemetry.

The working Phase 1 ownership model is that OBI covers Go workloads while the injector covers Java, .NET, Node.js, and Python. This issue must confirm that model and define the OBI behavior needed to support it.

## Scope

- Define how OBI detects injector- or SDK-instrumented processes.
- Define signal-specific suppression for traces, RED metrics, span metrics, and runtime metrics.
- Prevent conflicting or duplicate context propagation.
- Define behavior for manually instrumented Go applications.
- Ensure OBI, the injector, and Collector processes are excluded from unintended instrumentation.
- Confirm that Config v2 can express the approved workload ownership model.
- Define the OBI configuration needed to disable Java and Node.js runtime attachment for the Phase 1 profile.
- Define shared exporter, resource identity, service naming, header, and credential expectations that affect OBI.
- Add OBI-level behavioral tests independent of DEB/RPM installation.
- Document supported behavior and limitations for the Packaging integration tests.

## Acceptance criteria

- [ ] Workload ownership is explicitly documented.
- [ ] OBI can express and enforce the Phase 1 ownership model.
- [ ] Trace, RED metric, span-metric, runtime-metric, and propagation behavior is defined.
- [ ] Already instrumented Go applications have documented and tested behavior.
- [ ] OBI, injector, and Collector processes are excluded as intended.
- [ ] Representative OBI-level tests demonstrate that duplicate or conflicting instrumentation is prevented.
- [ ] The resulting contract is consumed by open-telemetry/opentelemetry-packaging#47.

## Consumer

- [Installed package coexistence coverage](https://github.com/open-telemetry/opentelemetry-packaging/issues/47)

PR #3021, opened 2026-08-13
Avoid duplicate Go HTTP/2 traceparents

### LINKED ISSUE #2769: Go tracer: prevent duplicate traceparent injection in HTTP/2 and GRPC

MrAlias:
## Summary

- detect application-owned `traceparent` fields per Go HTTP/2 or grpc-go request before HPACK serialization
- carry exact ownership through the Go framer and plaintext socket-fallback paths
- preserve OBI injection for requests without a traceparent
- cover current and legacy net/http, x/net/http2, and grpc-go layouts

Fixes #2769.

## Root cause

The Go HTTP/2 and grpc-go client uprobes appended an OBI HPACK field without knowing whether that request's application headers already contained `traceparent`. A socket wire scan cannot answer this reliably because HPACK may encode a repeated application field entirely as a dynamic-table index.

## Ownership and injection

The new pre-serialization probes observe header fields at stable Go library seams. They publish an exact stream key containing PID, process start time, the directional connection tuple, and stream ID. Each stream transitions through observing, application-owned, OBI-pending, OBI-written, or fail-closed state.

Version-specific probe groups attach atomically. The direct path commits before the underlying Go writer can consume the frame, including unbuffered TLS writes and terminal CONTINUATION frames. For plaintext fallback, tpinjector injects only from exact pending state, verifies all mutations and readbacks, and marks uncertain rollbacks fail-closed. The generic tracer prefers exact semantic state and does not fall back to the weak port key for a marked Go connection.

Normal request completion removes exact and transient state. LRU maps, process identity, directional keys, and a 30-second freshness bound contain missed-return residue and PID or connection reuse.

## Regression coverage

- current x/net/http2 and net/http, pinned pre-clientStream x/net, and Go 1.17 vendored HTTP/2
- current grpc-go and grpc-go 1.56, including alternate header handlers
- four repeated application-owned calls on persistent connections to exercise HPACK indexing
- no-traceparent controls, mixed multiplexed streams, large header blocks, and terminal CONTINUATION frames
- TLS, plaintext, first unbuffered writes, direct pre-write commits, and forced socket fallback
- exact receiver cardinality/value, peer continuity, probe/audit activation, exported client-span correlation, and cleanup
- host fault tests for partial direct and socket writes, verified rollback, and fail-closed rollback
- existing Go HTTP/1 duplicate and generic HTTP/2 propagation regressions

## Validation

- `make generate`
- `make clang-format`
- `make clang-tidy` (all 243 files)
- `make compile`
- BPF host test suite
- relevant Go unit tests
- privileged `TestHTTP2Go` and `TestHTTP2GoWithHTTPProtocols`
- privileged `TestSuite_GRPCGoTraceparentOwnership`, including current and legacy forced-socket CONTINUATION fallback
- HTTP/1 duplicate regression
- BPF verifier loads for the changed programs

## Stack and scope

This is 2 of 2 and depends on #3020.

The prerequisite is a focused 2-file symbol-resolution invariant (+133/-12). This draft keeps HTTP/2, grpc-go, direct writes, socket fallback, and their shared exact-state tests together because splitting them would separate one ownership state machine into unsafe intermediate changes. Of the 69 files here, 35 are tests, fixtures, or integration selection; the effective diff is approximately half validation coverage.

Unknown or custom-fork library layouts disable the atomic ownership group instead of partially attaching. Locked-down kernels without `bpf_probe_write_user` skip these ownership tests, and kernels without `bpf_msg_pop_data` use a no-mutation socket program. Socket fallback remains best effort when a complete frame is not present in one `sk_msg`; the Go user-buffer path handles normal instrumented connections before that split.

This change does not alter the disabled-propagation configuration semantics from #2276 or claim the broader coexistence contract discussed in #2835.


MrAlias:
Too large. Looking to break this up into reviewable PRs.

PR #3023, opened 2026-08-13
Correlate memory-BIO TLS connections by ciphertext prefix

### LINKED ISSUE #2998: [BUG] TLS client spans are attributed to the wrong socket

mifisignal:
## Summary

Fixes #2998. Outbound TLS client spans are attributed to the wrong socket when the
application's TLS stack uses a memory BIO.

OBI binds an `SSL*` to a connection only when a socket syscall happens while an
`SSL_read`/`SSL_write` uprobe is still on the stack. CPython `asyncio`
(`ssl.MemoryBIO` via `SSLContext.wrap_bio`) and Node.js `TLSWrap` encrypt into a
buffer and let the event loop write it later, so no syscall lands in that window.
OBI then falls back to the thread's most recent connection (on a busy event loop,
that is usually the inbound request being served) and latches it for the life of
the `SSL*`. The result is a client span reporting the caller's address and
ephemeral port as the server it called, which inverts the edge in a service graph.

The proposed change correlates the ciphertext to its socket instead of inferring
the connection from thread state:

- `SSL_set_bio` gives a `BIO*` -> `SSL*` map, so the write BIO can be identified.
  This matters because OpenSSL writes the same record into two or three BIOs on its
  way out; keying on every `BIO_write` would register the same prefix repeatedly.
- `BIO_write` on that write BIO registers a prefix of the outgoing ciphertext
  against the `SSL*`.
- At the socket write, the prefix is matched and the `SSL*` is bound to the real
  connection.

## Known limitation

A connection already established before OBI attaches cannot be correlated: there
was no `SSL_set_bio` to observe, so there is no `BIO*` -> `SSL*` entry and
`BIO_write` returns early. Those connections keep the current fallback behavior.
This clears when the application restarts.

## Validation

New BPF unit tests (`bpf/tests/bpf_tls_record_key.c`, `bpf/tests/bpf_tls_prefix_bind.c`)
cover key derivation, the record-length clamp, and the bind path.

New integration suites (`TestSuitePythonAsyncMemoryBIOTLSClient`,
`TestSuiteNodeMemoryBIOTLSClient`) run a CPython `asyncio` and a Node.js app
that each make one outbound TLS call per inbound request against a single local
TLS upstream peer. Any other peer on a client span is a misattribution. Each app
makes many concurrent requests to make the probability of exhibiting the bug
near certain.

Both suites were run against unpatched `main` and against this branch, same
harness and same load:

| runtime | unpatched `main` | this branch |
|---|---|---|
| Node.js | 1292 of 1504 client spans named a foreign peer (86%) | 0 of 1471 |
| CPython `asyncio` | ~10% of client spans, 114 offending series | 0 of 1434 |

The rate is stochastic — it depends on what the event loop is doing at each
outbound write, and repeated runs of the same workload have ranged from 3% to
12% for CPython. The suites therefore assert that no client span names a peer
other than the upstream over a large sample, rather than testing a threshold.

Worth noting for anyone reproducing this: when a span is mis-attributed,
`server.address` still resolves to the upstream and only `server.port` is
wrong, so the span looks plausible until the port is checked.

A standalone reproduction independent of this test suite is at
https://github.com/mifisignal/obi-tls-attribution-repro

## Validation checklist

- [x] features documentation / support matrix updated as needed

## Generative AI disclosure

Per the project's Generative AI Policy: Generative AI was used in preparing this
submission. Assistance covered exploratory tracing of the fault, drafting the
implementation and tests, and drafting this description.


bpf/generictracer/k_tracer.c
```diff
@@ -448,12 +449,29 @@ int BPF_KPROBE(obi_kprobe_tcp_sendmsg, struct sock *sk, struct msghdr *msg, size
...
                 if (buf) {
+                    const size_t avail = size;
+
                     size = read_msghdr_buf(msg, buf, size);
 
+                    // A memory BIO stack reaches the socket from the event
+                    // loop, after the SSL_write uprobe has returned. Bind it
+                    // here by matching this record against the prefix recorded
+                    // when OpenSSL wrote it to the write BIO.
+                    if (size >= k_tls_hdr_len && tls_record_plausible_start(buf, (u32)size) &&
```
MrAlias: When context propagation installs the `sk_msg` program, `read_msghdr_buf()` can return zero for the bvec-backed send. This match runs before `msg_buffers` supplies the readable ciphertext, while `tcp_rate_check_app_limited` reads that fallback buffer and goes directly to `handle_buf_with_connection(..., NO_SSL, ...)`. The SSL therefore remains on the thread fallback, so a memory-BIO client span can still name the inbound peer.

Could we run the prefix match after whichever buffer source succeeds, including the `tcp_rate_check_app_limited` path? The new integration config leaves context propagation disabled, so it does not cover this case.

bpf/generictracer/tls_prefix.h
```diff
@@ -0,0 +1,209 @@
...
+    k_tls_prefix_max_age_ns = 1000000000ULL,
+};
+
+// Drops the BIO associations an SSL owns, if any.
+static __always_inline void ssl_bios_forget(u32 pid, void *ssl) {
+    const pid_ptr_key_t ssl_key = pid_ptr_key(pid, ssl);
+    ssl_bios_t *bios = bpf_map_lookup_elem(&ssl_to_bios, &ssl_key);
+
+    if (!bios) {
+        return;
```
MrAlias, line 46: `ssl_to_bios` and `bio_to_ssl` are independent LRUs with 4096 and 8192 entries. Once the reverse map evicts an entry, this return leaves the forward BIO ownership behind; after that BIO address is reused, `tls_prefix_register_egress()` can trust the stale pointer and associate a new socket with the freed SSL.

Could we preserve paired ownership across eviction, or validate and discard a forward entry whose reverse ownership is gone? The current reuse test only covers teardown while the reverse entry exists.",

bpf/generictracer/k_tracer.c
mifisignal: Fixed. The plausibility screen moved into `tls_prefix_try_bind`, and the bind now runs wherever a buffer becomes readable: 1/after `read_msghdr_buf`, 2/after the `msg_buffers` fallback in `tcp_sendmsg`, and 3/in `tcp_rate_check_app_limited`. A successful bind skips the plaintext parsers and falls through to `tcp_send_ssl_check`; in `tcp_rate_check_app_limited`, the trailing `else` became an `if (ssl)` so a connection bound inside the fallback block reaches it.
 
Added `TestSuiteNodeMemoryBIOTLSClientContextPropagation`, which runs with `context_propagation: all`; the existing suites keep the plain path.

bpf/generictracer/tls_prefix.h
mifisignal: `tls_prefix_register_egress` now calls `bio_owner_is_current()`: the SSL that a forward entry names must still list that BIO as its rbio or wbio, otherwise the entry is deleted. An entry whose reverse was evicted is no longer correlated for that connection.
 
Added two tests: `test_orphaned_bio_entry_is_discarded` fails against the pre-fix code, and `test_recycled_bio_after_eviction_binds_its_own_ssl` checks that a BIO the new owner has re-registered still correlates.

mifisignal:
Fixed CLA issue and rebased to main.

grcevski (approved):
Looks great! LGTM!

MrAlias (approved):

bpf/common/tls_record.h
```diff
@@ -0,0 +1,161 @@
...
+
+// Builds a correlation key from a buffer that is already readable by the
+// caller. `copied` is how many bytes `src` actually holds; `avail` is the size
+// of the whole transfer the buffer was taken from. Returns 0 if no key could be
+// derived.
+static __always_inline u32 tls_prefix_key_from_buf(tls_prefix_key_t *key,
+                                                   const unsigned char *src,
+                                                   u32 copied,
+                                                   u32 avail) {
+    __builtin_memset(key, 0, sizeof(*key));
```
mmat11: use `bpf_memset`

mmat11 (commented):
Claude tells me `ChangeCipherSpec` could collide for the same process but different connections since a TLS 1.3 ChangeCipherSpec is always the same 6 bytes on the wire (14 03 03 00 01 01), and `tls_record_key_len` clamps to that len.

So if two connections are both still unbound, they register identical keys and the second overwrites the first

bpf/common/tls_record.h
mifisignal: Will fix. Is there any linting that enforces this? If not, you might want to add some.

mifisignal:
Confirmed, and it applies to short plaintext alerts too (close_notify is 7 bytes),
not only ChangeCipherSpec.

Fixed in 9a3788df: `tls_record_key_len` now rejects any record shorter than a header
plus a 16-byte AEAD tag, which is the smallest application record.

Added a regression test for the case, `test_rejects_low_entropy_records` in
`bpf_tls_record_key.c`. It asserts that the literal `14 03 03 00 01 01` is a
well-formed record that is nonetheless not keyed, same for close_notify, and that the
record meets the minimum size requirement.

Finally, updated to use `bpf_memset` per @mmat11.

mifisignal:
Confirmed, and it applies to short plaintext alerts too (close_notify is 7 bytes),
not only ChangeCipherSpec.

Fixes and updates:

`tls_record_key_len` now rejects any record shorter than a header
plus a 16-byte AEAD tag, which is the smallest application record.

Added a regression test for the case, `test_rejects_low_entropy_records` in
`bpf_tls_record_key.c`. It asserts that the literal `14 03 03 00 01 01` is a
well-formed record that is nonetheless not keyed, same for close_notify, and that the
record meets the minimum size requirement.

Also added `test_concurrent_unbound_connections_do_not_cross_bind` in
`bpf_tls_prefix_bind.c`, which drives the scenario you described: two connections
handshaking at once, both unbound, both writing that ChangeCipherSpec through their
own write BIO. It asserts neither registers a key, that the first connection's socket
send correlates nothing, that the second connection's SSL is not bound to the
first's peer, and that each still binds normally on its own application data.

Switched to `bpf_memset` at both sites — the one you flagged in `tls_record.h` and
the scratch record zeroing in `tls_prefix.h`.

bpf/common/tls_record.h
mmat11: We currently don’t have compile time enforcements enabled because there’s a lot of legacy code which uses those builtins; we will need to eventually refactor it before enabling the checks

mmat11 (approved):
lgtm if tests pass

grcevski:
Thanks for this contribution @mifisignal !

PR #3024, opened 2026-08-13
Fix java attach setting OBI uid/gid as 0s

### LINKED ISSUE #3018: JavaInjector.NewExecutable calls Cleanup with myUID=0 for non-Java processes, elevating non-root agents to euid 0

grcevski:
## Summary

OBI java attach could set the uid/gid on the OBI process itself to 0 is undesirable. This PR fixes the flow with the following approach:

- We now pass the cancellation context to the attach helpers, so they would get cancelled if the timeout expired.
- We make the init() and cleanup() idempotent and only run if there was successful save of credentials. This makes the outer cleanup on timeout termination safe to be called when no init happened.
- We skip the whole logic if the process is not Java.

Fixes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/3018

## Validation

### DIFF
--- a/pkg/internal/java/java_inject.go
+++ b/pkg/internal/java/java_inject.go
@@ -18,6 +18,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync/atomic"
 
 	"go.opentelemetry.io/obi/pkg/appolly/app"
 	"go.opentelemetry.io/obi/pkg/appolly/app/svc"
@@ -44,8 +45,9 @@ func (e *JavaInjectError) Error() string {
 }
 
 type JavaInjector struct {
-	log *slog.Logger
-	cfg *obi.Config
+	log             *slog.Logger
+	cfg             *obi.Config
+	currentAttachID atomic.Int64
 }
 
 func NewJavaInjector(cfg *obi.Config) (*JavaInjector, error) {
@@ -57,8 +59,9 @@ func NewJavaInjector(cfg *obi.Config) (*JavaInjector, error) {
 	}
 
 	return &JavaInjector{
-		cfg: cfg,
-		log: slog.With("component", "javaagent.Injector"),
+		cfg:             cfg,
+		log:             slog.With("component", "javaagent.Injector"),
+		currentAttachID: atomic.Int64{},
 	}, nil
 }
 
@@ -116,6 +119,8 @@ func (i *JavaInjector) NewExecutable(ie *ebpf.Instrumentable) error {
 		return nil
 	}
 
+	attachID := i.currentAttachID.Add(1)
+
 	ctx, cancel := context.WithTimeout(context.Background(), i.cfg.Java.Timeout)
 	defer cancel()
 
@@ -127,7 +132,10 @@ func (i *JavaInjector) NewExecutable(ie *ebpf.Instrumentable) error {
 
 	resultChan := make(chan result, 1)
 
-	attacher := jvm.NewJAttacher(i.log)
+	attacher := jvm.NewJAttacher(i.log, attachID, func() int64 {
+		return i.currentAttachID.Load()
+	})
+
 	runtime.LockOSThread()
 	defer runtime.UnlockOSThread()
 
--- a/pkg/internal/java/java_inject.go
+++ b/pkg/internal/java/java_inject.go
@@ -131,6 +131,15 @@ func (i *JavaInjector) NewExecutable(ie *ebpf.Instrumentable) error {
 	runtime.LockOSThread()
 	defer runtime.UnlockOSThread()
 
+	// We need to call cleanup here even though init may not have
+	// happened, to ensure the case when the JVM never responds and
+	// we've terminated with a timeout. Cleanup() is idempotent.
+	defer func() {
+		if err := attacher.Cleanup(); err != nil {
+			i.log.Warn("error on JVM attach cleanup", "error", err)
+		}
+	}()
+
 	// Run the attach procedure in a goroutine, so that we can terminate on stuck attach
 	go func() {
 		defer func() {
--- a/pkg/internal/java/java_inject.go
+++ b/pkg/internal/java/java_inject.go
@@ -148,7 +148,7 @@ func (i *JavaInjector) NewExecutable(ie *ebpf.Instrumentable) error {
 			}
 		}()
 
-		ok, jdk8 := i.verifyJVMVersion(attacher, ie.FileInfo.Pid())
+		ok, jdk8 := i.verifyJVMVersion(ctx, attacher, ie.FileInfo.Pid())
 		if !ok {
 			resultChan <- result{err: &JavaInjectError{Message: "unsupported Java version for OpenTelemetry eBPF instrumentation"}}
 			return
@@ -157,9 +157,9 @@ func (i *JavaInjector) NewExecutable(ie *ebpf.Instrumentable) error {
 		var loaded bool
 		var err error
 		if jdk8 {
-			loaded, err = i.jdkAgentAlreadyLoadedHotspot8(attacher, ie.FileInfo.Pid())
+			loaded, err = i.jdkAgentAlreadyLoadedHotspot8(ctx, attacher, ie.FileInfo.Pid())
 		} else {
-			loaded, err = i.jdkAgentAlreadyLoaded(attacher, ie.FileInfo.Pid())
+			loaded, err = i.jdkAgentAlreadyLoaded(ctx, attacher, ie.FileInfo.Pid())
 		}
 
 		if err != nil {
@@ -182,7 +182,7 @@ func (i *JavaInjector) NewExecutable(ie *ebpf.Instrumentable) error {
 			return
 		}
 
-		if err = i.attachJDKAgent(attacher, ie.FileInfo.Pid(), agentPath); err != nil {
+		if err = i.attachJDKAgent(ctx, attacher, ie.FileInfo.Pid(), agentPath); err != nil {
 			i.log.Error("couldn't attach OpenTelemetry eBPF Java Agent", "pid", ie.FileInfo.Pid(), "path", agentPath, "error", err)
 			resultChan <- result{err: err}
 			return
@@ -289,7 +289,7 @@ func (i *JavaInjector) attachOpts() string {
 	return "=" + strings.Join(opts, ",")
 }
 
-func (i *JavaInjector) attachJDKAgent(attacher *jvm.JAttacher, pid app.PID, path string) error {
+func (i *JavaInjector) attachJDKAgent(ctx context.Context, attacher *jvm.JAttacher, pid app.PID, path string) error {
 	attacher.Init()
 
 	defer func() {
@@ -297,7 +297,7 @@ func (i *JavaInjector) attachJDKAgent(attacher *jvm.JAttacher, pid app.PID, path
 			slog.Warn("error on JVM attach cleanup", "error", err)
 		}
 	}()
-	out, err := attacher.Attach(int(pid), []string{"load", "instrument", "false", path + i.attachOpts()}, false)
+	out, err := attacher.Attach(ctx, int(pid), []string{"load", "instrument", "false", path + i.attachOpts()}, false)
 	if err != nil {
 		i.log.Error("error executing command for the JVM", "pid", pid, "error", err)
 		return err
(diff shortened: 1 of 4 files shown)


grcevski:
I also tested manually with the following approach:

1. I set the capabilities on a freshly built OBI like this:
```
sudo setcap 'cap_sys_admin,cap_sys_ptrace,cap_dac_read_search,cap_net_raw,cap_perfmon,cap_bpf,cap_checkpoint_restore=ep' bin/obi
```
2. I ran a Java application and I stopped it with GDB, so the JVM will never respond.
3. I ran OBI to instrument the application. After the 10 seconds timeout I saw:
```
time=2026-08-13T14:31:15.649-04:00 level=WARN msg="java attach timed out" component=javaagent.Injector timeout=10s pid=235034
time=2026-08-13T14:31:15.650-04:00 level=WARN msg="unable to attach java agent to process, Java TLS telemetry will not work" component=discover.traceAttacher pid=235034 error="java attach timed out"
time=2026-08-13T14:31:15.650-04:00 level=INFO msg="instrumenting process" component=discover.traceAttacher cmd=/home/nino/tools/jdk-21.0.10+7/bin/java pid=235034 ino=36205999 type=java service="" logenricher=false
```
4. I checked the OBI user with:
```
ps -o user=,pid=,cmd= -C obi
```
It was still my own user.

pkg/internal/java/java_inject.go
```diff
@@ -127,75 +131,74 @@ func (i *JavaInjector) NewExecutable(ie *ebpf.Instrumentable) error {
 	runtime.LockOSThread()
 	defer runtime.UnlockOSThread()
 
+	// We need to call cleanup here even though init may not have
+	// happened, to ensure the case when the JVM never responds and
+	// we've terminated with a timeout. Cleanup() is idempotent.
 	defer func() {
 		if err := attacher.Cleanup(); err != nil {
```
MrAlias: This cleanup can run after the in-flight helper calls `Init()` but before `Attach()` switches credentials. It clears `initialized`, so when that helper returns from `Attach()`, its deferred `Cleanup()` becomes a no-op and OBI can remain running as the target user.

Could we keep the full `Init`/`Attach`/`Cleanup` lifecycle owned by the attach goroutine, and have the timeout path wait for credentials to be restored before returning?

MrAlias (commented):
The non-Java guard fixes the original trigger. The timeout path still has a credential-restoration race.

pkg/internal/java/java_inject.go
grcevski: > This cleanup can run after the in-flight helper calls Init() but before Attach() switches credentials. It clears initialized, so when that helper returns from Attach(), its deferred Cleanup() becomes a no-op and OBI can remain running as the target user.

You are right, I didn't think of that one. I'm thinking maybe I don't have to flip the initialized back to false in the Cleanup() call? That way we'll restore the OBI state no matter what. 

> Could we keep the full Init/Attach/Cleanup lifecycle owned by the attach goroutine, and have the timeout path wait for credentials to be restored before returning?

Can you explain better what you mean here? I think you mean I shouldn't bother with the top level clean-up, the one I removed and then added. I was worried this Go routines stays permanently in attach somehow, flips the uid/gid to the process and something blocks it that we won't be able to cancel on. That code makes syscalls and doesn't pass the context everywhere, so if it doesn't cancel we end up with OBI with bad uid/gid.

grcevski: I think I want the following:
- The timeout ending should have a final say, it should restore the original OBI uid/gid and any changes to these shouldn't be possible after we've done that.
- I think maybe I made a mistake with moving the attach off the background. If it times out but eventually succeeds, we don't hold up the pipeline, but eventually we load the java agent or fail.
- Since I turned the actual syscalls to change uid/gid to functions, for testing reasons, I think I can easily add logic in them to say, you are changing them to the non-default, but we timed out, so back-off. I can use a lock there to prevent attach racing with the outer cleanup.

MrAlias: That makes sense, and your first bullet is exactly the invariant I was trying to describe. I was too prescriptive about waiting for the attach goroutine: if the timeout path can make credential restoration final, this concern does not require holding up the pipeline until that goroutine exits.

I don't think leaving `initialized` true by itself is enough, because restoration would still depend on the helper eventually returning. The lock approach in your latest comment sounds good if the same critical section covers the timeout check and both target credential syscalls, and the timeout path uses that lock to mark the attacher timed out and restore the original IDs. Then either attach switches first and timeout restores the IDs, or timeout wins and attach cannot switch them afterward.

In that design, I think `Cleanup()` should still clear `initialized`, so the helper's later defer becomes a no-op. I'd put the synchronization and terminal state on `JAttacher` and keep the package-level syscall variables as test seams.

grcevski: I made the changes based on my plan, except I didn't put attach in the background, I'm not sure if we should gamble with potential goroutine leak for the chance that attach succeeds later.

I introduced new field terminated and that is now what guards everything. I keep initialized as not being unset, because it's really the state of = "did we get the OBI own credentials".

pkg/internal/java/java_inject.go
```diff
@@ -127,75 +131,74 @@ func (i *JavaInjector) NewExecutable(ie *ebpf.Instrumentable) error {
 	runtime.LockOSThread()
 	defer runtime.UnlockOSThread()
 
+	// We need to call cleanup here even though init may not have
+	// happened, to ensure the case when the JVM never responds and
+	// we've terminated with a timeout. Cleanup() is idempotent.
 	defer func() {
-		if err := attacher.Cleanup(); err != nil {
-			slog.Warn("error on JVM attach cleanup", "error", err)
+		if err := attacher.Cleanup(true); err != nil {
```
MrAlias: This timeout can still leave `NewExecutable` blocked on OpenJ9:

`JVM stops replying` → `ctx.Done()` wins → this deferred cleanup runs before the return completes → `detach()` waits without a deadline for `ATTACH_DETACHED` → the timeout error never reaches `traceAttacher`.

Credential restoration happens after `detach()`, so OBI can also remain under the target UID/GID while the discovery loop is stuck. There is a similar pre-connection gap: OpenJ9's blocking `flock(LOCK_EX)` calls do not observe `ctx` and can leave the helper plus its pinned OS thread behind after the deadline.

Could we separate graceful detach from terminal cleanup? Normal completion can send `ATTACH_DETACHED`; once the deadline fires, terminal cleanup should mark the attacher terminated, restore OBI's credentials, and interrupt its local waits without waiting for the JVM protocol. The listener, connection, and acquired locks also need to be published under the same synchronization as the terminal state, so the worker cannot acquire or publish a resource after cleanup has taken ownership. The blocking operations could then use mechanisms such as `LOCK_NB` retries that check `ctx` and socket deadlines derived from the context.

MrAlias (commented):
The credential-restoration race from my prior pass is fixed. OpenJ9 can still keep the configured timeout from returning, leaving process discovery blocked.

pkg/internal/java/java_inject.go
grcevski: Good point. I changed the code to separate the two paths. Essentially, everywhere where Cleanup() will be called gracefully, it's actually where we want to detach from J9. At the Terminate() new call, whcih is in the outer goroutine, I decided to not wait for the graceful detach and simply restore the credentials. This way if any of the internal goroutines finish, we'll eventually call detach and if it's not blocked, we'll clean it up, if it's blocked that goroutine will be stuck. We can work on an extension to prevent this goroutine leak by making more of the attach internals context aware, so they can cancel. But I think that belongs in another PR.

pkg/internal/jvmtools/jvm/cmd.go
```diff
@@ -37,42 +51,91 @@ func NewJAttacher(logger *slog.Logger) *JAttacher {
...
+		return nil
+	}
+
+	var cleanupErr error
+
+	if j.j9attacher != nil {
+		cleanupErr = errors.Join(cleanupErr, j.j9attacher.detach())
+	}
+
+	cleanupErr = errors.Join(cleanupErr, j.restoreCredentialsLocked())
```
MrAlias: After attach A times out, `Terminate()` attempts to restore the original IDs and `traceAttacher` continues to the next event. If A's retained worker later exits, its deferred `Cleanup()` reaches this unconditional restore again. Attach B owns a different `JAttacher`, so the per-instance mutexes do not serialize A's late restore with B's credential switches; A can reset the process-wide euid/egid while B is mid-attach.

Could we make credential ownership process-wide rather than per `JAttacher`? After a successful `Terminate()`, A's stale cleanup must not restore the IDs again. If terminal restoration fails, proceeding to B while A may retry is also unsafe, so that failure and any retry need to be coordinated with future credential switches. A shared credential lock/lease across attachers would make both cases explicit.

MrAlias (commented):
NIce, I like the split between terminal cleanup and graceful OpenJ9 detach; it fixes the blocking timeout path and makes the lifecycle much easier to reason about. The PR is moving in the right direction, though one cross-attacher race remains around the process-wide credential restore.

pkg/internal/jvmtools/jvm/cmd.go
grcevski: Right, yeah, there's that mid attach restoration issue. I don't think I can easily make this process wide, or specific. But I have an idea. I can record on the init the process ID that currently owns the attach sequence. So if late Cleanup() comes back in to restore the credentials to OBI, it can check the owner id and if it's not them, then it will skip the restore of the credentials.

grcevski: OK, I made another update and now I'm using a concept of attachID, which increments with every new attach operation. I prevent the Cleanup to restore the credentials if the active attach ID doesn't match the attachID it had during initialisation. I initially wrote the change by using the PID from FileInfo, but after some thought, I think repeated PIDs would've had the same issue. We shouldn't get the same PID to be looked at twice, but it's theoretically possible. So I think using a completely unrelated ID is better.

pkg/internal/jvmtools/jvm/cmd.go
```diff
@@ -14,65 +14,139 @@ import (
...
+		cleanupErr = errors.Join(cleanupErr, j.j9attacher.detach())
+	}
+
+	activeAttachID := j.activeAttachID()
+
+	// Restore credentials if we are the active attach happening on the pipeline.
+	// If we were delayed and the main loop abandoned us, they would reset the
+	// credentials in Terminate and they might be in process of other JVM attach.
+	if activeAttachID == j.attachID {
+		cleanupErr = errors.Join(cleanupErr, j.restoreCredentialsLocked())
```
MrAlias: `activeAttachID` is sampled before the process-wide credential restore, but nothing serializes that check with the next attach. A stale cleanup can load its own ID here, get descheduled, then attach B increments `currentAttachID` and switches to B’s UID/GID; when A resumes, it still enters this branch and restores OBI’s IDs over B’s attach. The new JVM attach can then fail under the wrong credentials, leaving that process without Java TLS telemetry even though discovery continues. The new test changes the active ID before calling the old cleanup, so it does not cover this check-then-act interleaving. Could we use a process-wide lock or lease that covers the ownership check, both restore syscalls, and the next attacher’s credential switches?

grcevski: ah yeah ok, I'll change how this check is done. I think I can just change the semantics, make it verify current id and then use compare atomically.

grcevski: OK, I made one more fix. Now I use a lock and I delegate the decision on whether to run the function on the main goroutine.

MrAlias (approved):
Awesome work 🚀

PR #3026, opened 2026-08-13
fix(metrics)!: derive internal metric names and drop Prometheus-isms from the OTLP ones

NimrodAvni78:
## Summary

Two related changes to OBI's internal (`obi.*`) metrics.

**1. Four OTLP names end in `_total`**, which denotes a monotonic counter — two of them are gauges.
That suffix belongs to the Prometheus rendering, where it is derived from the instrument type and
unit.

| OTLP name | after | instrument |
|---|---|---|
| `obi.bpf.map.entries_total` | `obi.bpf.map.entries` | gauge |
| `obi.bpf.map.max_entries_total` | `obi.bpf.map.max_entries` | gauge |
| `obi.bpf.network.packets.total` | `obi.bpf.network.packets` | counter |
| `obi.bpf.network.ignored.packets.total` | `obi.bpf.network.ignored.packets` | counter |

**2. Internal metrics now declare their name once**, extending the derive-don't-hand-write approach
to the last place it was missing. They were written twice — the OTLP name in
`pkg/export/otel/metrics_internal.go`, the Prometheus name in `pkg/export/imetrics/iprom.go` — with
nothing keeping them consistent. They are now declared in
`pkg/export/attributes/metric_internal.go` and both exporters read `OTEL` / `Unit` / `Prom` from
there, including the `obi.bpf.probe.latency` histogram, whose name was still a local `var`.

This is a refactor, not a rename: `TestInternalPrometheusNames` pins every derived Prometheus name
against what the exporters emitted before, and they match exactly. It also makes the rename above
self-consistent — previously the two map gauges had to be edited by hand in both files to stay
aligned.

Three metrics stay hand-written in `iprom.go` because they have no OTLP counterpart to derive from:
`obi_prometheus_http_requests_total` (`PrometheusRequest` is a no-op on the OTLP side), and
`obi_bpf_probe_executions_total` / `obi_bpf_probe_latency_seconds_total`, whose OTLP equivalents
were replaced by the latency histogram's count and sum.

## Breaking changes

**All four OTLP names change**, so a backend consuming OBI's OTLP directly sees four renames. On the
Prometheus side it depends on the translation:

- Default (`add_metric_suffixes: true`): only the two map gauges move —
  `obi_bpf_map_entries_total` → `obi_bpf_map_entries`, `obi_bpf_map_max_entries_total` →
  `obi_bpf_map_max_entries`. The counters are unchanged, because `_total` is de-duplicated and
  re-appended.
- With `add_metric_suffixes: false` or a no-suffix / `NoTranslation` strategy, the counters move too.

Both are internal meta-metrics about OBI itself, not application telemetry.

```yaml
metric_relabel_configs:
  - source_labels: [__name__]
    regex: 'obi_bpf_map_entries'
    target_label: __name__
    replacement: 'obi_bpf_map_entries_total'
  - source_labels: [__name__]
    regex: 'obi_bpf_map_max_entries'
    target_label: __name__
    replacement: 'obi_bpf_map_max_entries_total'
```

`metric_relabel_configs` only helps scrape-time consumers. On an OTLP → collector → remote-write
path, use a collector `transform` rule instead.

The un-prefixed `bpf_map_entries_total` on the application endpoint (`prom_bpf.go`) has the same
defect — a point-in-time entry count typed and named as a counter — but it is a different metric
with a different label set, so it is left for a focused fix.

## Not included

Internal metric *label* names (`map_*` → `bpf_map_*`, `process_name` →
`process_executable_name`, `error` → `obi_error`) still differ between the two exporters. That is a
separate, wider breaking change and the last remaining step for internal metrics.

## Testing

`GOOS=linux go build ./...`, `go vet`, `gofmt` and `make lint-schema` clean; `./pkg/export/...`
passes. `TestInternalPrometheusNames` covers all 16 declarations. Assertions updated in
`prom_bpf_test.go` and `internal_metrics_otel_test.go`, the latter now expecting the histogram's
`_count` series rather than the two removed counters; the integration suite was not run locally.

`TestTerminatesOnBadPromPort` is a pre-existing flake — it binds a random port and intermittently
hits `bind: address already in use`, reproducible on repeat runs — so budget a retry rather than
reading it as a regression.

## Validation

### DIFF
--- a/pkg/export/imetrics/iprom.go
+++ b/pkg/export/imetrics/iprom.go
@@ -59,10 +59,12 @@ type PrometheusReporter struct {
 }
 
 func NewPrometheusReporter(cfg *InternalMetricsConfig, manager *connector.PrometheusManager, registry *prometheus.Registry) *PrometheusReporter {
+	internalNames := attributes.NewInternalMetrics(attr.VendorPrefix)
+
 	pr := &PrometheusReporter{
 		connector: manager,
 		tracerFlushes: prometheus.NewHistogram(prometheus.HistogramOpts{
-			Name:                            attributes.InternalTracerFlushes.Prom,
+			Name:                            internalNames.TracerFlushes.Prom,
 			Help:                            "Length of the groups of traces flushed from the eBPF tracer to the next pipeline stage",
 			Buckets:                         pipelineBufferLengths,
 			NativeHistogramBucketFactor:     1.1,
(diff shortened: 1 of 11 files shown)


pkg/export/imetrics/iprom.go
```diff
@@ -61,43 +62,43 @@ func NewPrometheusReporter(cfg *InternalMetricsConfig, manager *connector.Promet
 	pr := &PrometheusReporter{
 		connector: manager,
 		tracerFlushes: prometheus.NewHistogram(prometheus.HistogramOpts{
-			Name:                            attr.VendorPrefix + "_ebpf_tracer_flushes",
+			Name:                            attributes.InternalTracerFlushes.Prom,
```
grcevski: This will break our existing legacy beyla customers. I'm not sure how many still depend on the old internal metrics, but it would be nice to keep the vendor prefix functionality somehow. Maybe you can make this into a function where the prefix is passed in and then used for the runtime metrics.

NimrodAvni78: good catch, done!

grcevski (approved):
LGTM!

PR #3027, opened 2026-08-13
Fix stale go trace map data

### LINKED ISSUE #2948: Unbounded trace merging on the Go tracer — is porting the max_transaction_time bound tracked anywhere?
Follow-up to #2577 (closed, couldn't-reproduce) and #2551. I think the underlying gap is still open, and I have a repro that needs no custom code.

With header context propagation on, I get single traces with thousands of unrelated spans under one trace id. One I pulled had ~2,700 spans across a ~5-hour window: all Beyla-generated, each span sub-second, distinct span ids sharing one trace id — so the long window is merge, not latency. It's just Grafana remote-writing to Mimir under Beyla. Your own stack, no app code.

In #2577 you mentioned the generic path has a bound (`max_transaction_time`) that isn't ported to the Go tracer yet. That fits what I see: each hop is well inside the window, so the per-hop gate never trips and the trace grows forever. I couldn't find an issue tracking that port. Is it tracked somewhere, or worth opening one?

Happy to share the trace and repro steps.

---

*Disclosure: I used an AI assistant to investigate and draft this. The measurements are from my own cluster and I've reviewed all of it.*

grcevski:
## Summary

For the new generic Go tracer path, we had to remove our old probe on the `Runtime.goexit1`. The reason is that this probe typically removes the correlation too quickly and we'd not find correctly any goroutine information for the final send/receive message. There's no downside of leaving tombstones for the goroutines, except when the goroutine is reused. The go routine mapping is fine, but we used to clear also the go_trace_map where the trace context information is stored, and this wasn't happening anymore.

Here's how the goexit1 probe looked like:
```go
SEC("uprobe/runtime_goexit1")
int beyla_uprobe_proc_goexit1(struct pt_regs *ctx) {
    bpf_dbg_printk("=== uprobe/proc goexit1 === ");

    void *goroutine_addr = GOROUTINE_PTR(ctx);
    bpf_dbg_printk("goroutine_addr %lx", goroutine_addr);

    u64 pid_tid = bpf_get_current_pid_tgid();
    u32 pid = pid_from_pid_tgid(pid_tid);

    go_addr_key_t g_key = {.addr = (u64)goroutine_addr, .pid = pid};

    bpf_map_delete_elem(&ongoing_goroutines, &g_key);
    // We also clean-up the go routine based trace map, it's an LRU
    // but at this point we are sure we don't need the data.
    bpf_map_delete_elem(&go_trace_map, &g_key); // <-- IMPORTANT BIT FOR THIS BUG

    return 0;
}
```

So while we benefited from keeping the `ongoing_goroutines` tombstone, we now are not clearing the go_trace_map when we are done with the goroutine.

I re-enabled the `goexit1` probe and saw what was happening with the memcached client:
```
      testserver-87025   [008] ...11 62242.170633: bpf_trace_printk: === uprobe/proc goexit1 === 
      testserver-87025   [008] ...11 62242.170636: bpf_trace_printk: goroutine_addr 2b37366b10e0
      testserver-87025   [008] ...11 62242.170651: bpf_trace_printk: === uprobe/proc goexit1 === 
      testserver-87025   [008] ...11 62242.170654: bpf_trace_printk: goroutine_addr 2b37366b1680
      testserver-87025   [008] ...11 62242.170665: bpf_trace_printk: === uprobe/runtime_newproc1 ===
      testserver-87025   [008] ...11 62242.170668: bpf_trace_printk: creator_goroutine_addr=2b37365e10e0
      testserver-87025   [008] ...11 62242.170675: bpf_trace_printk: === uprobe/runtime_newproc1_return ===
      testserver-87025   [008] ...11 62242.170679: bpf_trace_printk: creator_goroutine_addr=2b37365e10e0
      testserver-87025   [008] ...11 62242.170681: bpf_trace_printk: parent_goroutine=2b37366b0780
      testserver-87025   [008] ...11 62242.170682: bpf_trace_printk: goroutine_addr=2b37366b1680

```
We can see that as soon as `goexit1` happens we create a new goroutine and it has the same exact address, inheriting accidentally now the go trace information.

The fix is to clear the go_trace_map when we create the goroutine, at that point we know we are starting a fresh and we can delete the stale info, without removing it too soon for correlation.

Fixes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2948

## Validation

### DIFF
--- a/bpf/gotracer/go_runtime.c
+++ b/bpf/gotracer/go_runtime.c
@@ -703,6 +703,9 @@ int obi_uprobe_runtime_newproc1(struct pt_regs *ctx) {
         bpf_dbg_printk("can't update map element");
     }
 
+    // Delete any stale info on go_trace_map
+    bpf_map_delete_elem(&go_trace_map, &g_key);
+
     return 0;
 }
 
@@ -757,26 +760,6 @@ int obi_uprobe_runtime_newproc1_return(struct pt_regs *ctx) {
 
 done:
     bpf_map_delete_elem(&newproc1, &c_key);
-
-    return 0;
-}
-
-SEC("uprobe/runtime_goexit1")
-int beyla_uprobe_proc_goexit1(struct pt_regs *ctx) {
-    bpf_dbg_printk("=== uprobe/proc goexit1 === ");
-
-    void *goroutine_addr = GOROUTINE_PTR(ctx);
-    bpf_dbg_printk("goroutine_addr %lx", goroutine_addr);
-
-    u64 pid_tid = bpf_get_current_pid_tgid();
-    u32 pid = pid_from_pid_tgid(pid_tid);
-
-    go_addr_key_t g_key = {.addr = (u64)goroutine_addr, .pid = pid};
-
-    //bpf_map_delete_elem(&ongoing_goroutines, &g_key);
-    // We also clean-up the go routine based trace map, it's an LRU
-    // but at this point we are sure we don't need the data.
-    bpf_map_delete_elem(&go_trace_map, &g_key);
 
     return 0;
 }
--- a/bpf/gotracer/go_runtime.c
+++ b/bpf/gotracer/go_runtime.c
@@ -703,9 +703,6 @@ int obi_uprobe_runtime_newproc1(struct pt_regs *ctx) {
         bpf_dbg_printk("can't update map element");
     }
 
-    // Delete any stale info on go_trace_map
-    bpf_map_delete_elem(&go_trace_map, &g_key);
-
     return 0;
 }
 
@@ -759,6 +756,8 @@ int obi_uprobe_runtime_newproc1_return(struct pt_regs *ctx) {
     }
 
 done:
+    // Delete any stale info on go_trace_map
+    bpf_map_delete_elem(&go_trace_map, &g_key);
     bpf_map_delete_elem(&newproc1, &c_key);
 
     return 0;
--- a/bpf/gotracer/go_runtime.c
+++ b/bpf/gotracer/go_runtime.c
@@ -716,6 +716,10 @@ int obi_uprobe_runtime_newproc1_return(struct pt_regs *ctx) {
 
     bpf_dbg_printk("creator_goroutine_addr=%lx", creator_goroutine_addr);
 
+    // The result of newproc1 is the new goroutine
+    void *goroutine_addr = (void *)GO_PARAM1(ctx);
+    go_addr_key_t g_key = {.addr = (u64)goroutine_addr, .pid = pid};
+
     // Lookup the newproc1 invocation metadata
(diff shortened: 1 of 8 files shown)


bpf/gotracer/go_runtime.c
```diff
@@ -703,6 +703,9 @@ int obi_uprobe_runtime_newproc1(struct pt_regs *ctx) {
         bpf_dbg_printk("can't update map element");
     }
 
+    // Delete any stale info on go_trace_map
+    bpf_map_delete_elem(&go_trace_map, &g_key);
```
MrAlias: `g_key` is built from `creator_goroutine_addr`, which is the runtime goroutine executing `runtime.newproc1`, not the new `*g` whose address may have been recycled. The child address is only available as `GO_PARAM1(ctx)` in the return probe below, so this deletes a different entry and leaves the stale `go_trace_map` value in place.

Could we delete the entry keyed by the returned `goroutine_addr` in `obi_uprobe_runtime_newproc1_return`, before any early exit can bypass the cleanup?

internal/test/oats/memcached/yaml/oats_go_memcached.yaml
```diff
@@ -0,0 +1,58 @@
...
+  - ../docker-compose-obi-go-memcached.yml
+  - ../../weaver/docker-compose-weaver.yml
+input:
+- path: /memcached
+- path: /memcached-error
+- path: /memcached-noreply
+interval: 500ms
+expected:
+  traces:
+  - traceql: '{ .db.operation.name = "SET" && .db.system.name = "memcached" }'
```
MrAlias: These assertions prove that the expected spans exist, but they still pass if goroutine reuse merges independent requests into one trace. Could we add a trace-level assertion that fails without the fix and passes reliably with it? A check that no trace contains more than one matching memcached `SET` looks like a reasonable first approach; this does not need a new general-purpose map lifecycle harness.

bpf/gotracer/go_runtime.c
grcevski: Ah right! I'll make the fix.

internal/test/oats/memcached/yaml/oats_go_memcached.yaml
grcevski: Yeah you are right. I think for that I need to write a normal integration test. I'll check if I can express something like what you suggested above as traceql.

bpf/gotracer/go_runtime.c
grcevski: FIxed.

internal/test/oats/memcached/yaml/oats_go_memcached.yaml
grcevski: FIxed.

mmat11 (approved):

MrAlias (approved):

PR #3028, opened 2026-08-13
Add support for Python runtime metrics

### LINKED ISSUE #2903: Add Python runtime metrics
### Component

OBI (eBPF instrumentation)

### Problem

Provide support for Python runtime metrics, covering as much as possible what's defined here https://opentelemetry.io/docs/specs/semconv/runtime/cpython-metrics/

_No response_

_No response_

### Additional context

_No response_

marctc:
## Summary

This PR adds CPython garbage-collection metrics to the existing `application_runtime` feature.

OBI resolves `_PyRuntime` from the executable or mapped `libpython`. It also validates the CPython version and runtime layout. The controller prefers the `python:gc__done` USDT probe. On Linux `amd64`, it can use a validated internal collector return probe as a fallback.

After each completed GC collection, the eBPF program reads the cumulative counters for generations 0, 1, and 2. It then sends the snapshot through the existing runtime metrics pipeline.

The current implementation supports validated CPython 3.9 through 3.14 releases on Linux `amd64` and `arm64`.

A follow-up PR will add compatibility for CPython builds with stripped symbols, including affected slim and Alpine images.

Fixes #2903

## Validation


pkg/export/attributes/attr_defs.go
```diff
@@ -497,6 +497,27 @@ func getDefinitions(
 			SubGroups:  []*AttrReportGroup{&appAttributes},
 			Attributes: map[attr.Name]Default{},
 		},
+		CPythonGCCollections.Section: {
+			SubGroups: []*AttrReportGroup{&appAttributes},
+			Attributes: map[attr.Name]Default{
+				attr.CPythonGCGeneration: true,
+				attr.ProcessPID:          true,
```
grcevski: Making the PID on by default will cause high cardinality metrics. I think it makes sense for all runtime metrics to have the processPID as attribute, even Go and JVM, so people can differentiate the actual process, but I'm not sure it should be on by default.

This is particularly going to be a problem for Python, where there are no threads in the main process, but threading is achieved by seaprate worker processes. Can you confirm what this currently looks like for the PythonSDK, do they have the process_pid in the metrics by default?

pkg/export/otel/metrics_runtime.go
```diff
@@ -133,14 +166,15 @@ func newRuntimeMetricsReporter(
...
+		resourceKeys:   map[svc.UID]svc.UID{},
 		log:            log,
 		selector:       selectorCfg.SelectionCfg,
 		runtimeEnabled: runtimemetrics.EnabledFeatures(jointMetricsConfig.Features),
 	}
 
-	reporter.reporters, err = otelcfg.NewReporterPool[*svc.Attrs, *RuntimeMetrics](cfg.ReportersCacheLen, cfg.TTL, timeNow,
-		func(id svc.UID, v *RuntimeMetrics) {
-			llog := log.With("service", id)
+	reporter.reporters, err = otelcfg.NewReporterPool[runtimeMetricResource, *RuntimeMetrics](cfg.ReportersCacheLen, cfg.TTL, timeNow,
```
grcevski: I don't understand this change. Can you please clarify why we are changing the reporter key from the service attributes to something based on the selected attributes etc?

pkg/export/prom/prom.go
```diff
@@ -1507,15 +1502,16 @@ func (r *metricsReporter) handleProcessEvent(pe exec.ProcessEvent, log *slog.Log
 	uid := snap.UID
 
 	if pe.Type == exec.ProcessEventCreated {
+		trackedUID, pidTracked := r.pidsTracker.TracksPID(pid)
```
grcevski, line 1548: I don't understand this change again. Getting this code right is very tricky and we have no tests to prove this works or not at the moment. There are many gotchas with properly replacing target info metrics and ensuring there are no duplicates, when we have partial state (before k8s data) and then k8s data arrives. Removing the staleUID check will definitely cause problems.

pkg/export/prom/prom_runtime.go
```diff
@@ -323,29 +328,28 @@ func (r *metricsReporter) deleteRuntimeMetrics(service *svc.Attrs) {
 	}
 
 	labels := r.labelValuesTargetInfo(service)
-	if r.goRuntimeMetrics.memoryLimit == nil {
-		return
+	if r.goRuntimeMetrics.memoryLimit != nil {
```
grcevski, line 340: Just noticed, don't we need to handle here jvm metrics deletion too? I guess this was a bug from before, but we probably need to refactor this and ensure this is handled in each language runtime metrics.

pkg/export/prom/prom_test.go
```diff
@@ -1405,9 +1405,9 @@ func TestHandleProcessEventCreated_EdgeCases(t *testing.T) {
 		}
 		reporter.handleProcessEvent(event2, slog.Default())
 
-		// Service should only be created once initially, then updated once for the same UID
+		// The second PID refreshes target info without deleting runtime metrics for the first PID.
```
grcevski: I think I understand maybe better the changes related to target info, if you add pid to the target info this will cause cardinality explosion. We used to have it and we had to remove it because scaling worker pools would constantly increase cardinality. We shouldn't be adding the pid to the target info, at least not with this PR. If we do, then there must be extensive integration tests on that part.

pkg/internal/cpython/runtime/decode.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func decodeIndirectRings(payload []byte) (runtimemetrics.PythonRuntimeMetricSnapshot, error) {
+	if len(payload) != indirectGILPayloadSize {
+		return runtimemetrics.PythonRuntimeMetricSnapshot{}, fmt.Errorf("ring GC payload size %d, want %d", len(payload), indirectGILPayloadSize)
+	}
+
+	var snapshot runtimemetrics.PythonRuntimeMetricSnapshot
+	for generation, ring := range indirectGILRings {
+		index := int(int8(payload[ring.indexOffset]))
```
grcevski: We need to ensure the payload has enough size here, before we dereference. This should be followed by a test.

pkg/internal/cpython/runtime/decode.go
```diff
@@ -0,0 +1,136 @@
...
+		snapshot.Generations[generation] = values
+	}
+	return snapshot, nil
+}
+
+func decodeCounters(payload []byte, offset int) (runtimemetrics.PythonGCGenerationMetrics, error) {
+	if offset < 0 || offset+inlineRecordSize > len(payload) {
+		return runtimemetrics.PythonGCGenerationMetrics{}, errors.New("short CPython GC field")
+	}
+	collections := int64(binary.LittleEndian.Uint64(payload[offset:]))
```
grcevski: What are these offsets here? Also by reading the code, I can't tell directly if inlineRecordSize is less or greater than 16, so it's unclear if this is legal or illegal access.

pkg/internal/cpython/runtime/decode.go
```diff
@@ -0,0 +1,136 @@
...
+		values, err := decodeCounters(payload, offset+ringCollectionsOffset)
+		if err != nil {
+			return runtimemetrics.PythonRuntimeMetricSnapshot{}, fmt.Errorf("generation %d: %w", generation, err)
+		}
+		snapshot.Generations[generation] = values
+	}
+	return snapshot, nil
+}
+
+func decodeCounters(payload []byte, offset int) (runtimemetrics.PythonGCGenerationMetrics, error) {
```
grcevski: I'd like a test specific to decodeCounters with invalid sizes, negative offsets etc.

pkg/internal/cpython/runtime/layout.go
```diff
@@ -0,0 +1,200 @@
...
+		profile.payloadSize = inlinePayloadSize
+		if !loadDebugOffsets(&profile, versionString) {
+			return layoutProfile{}, false
+		}
+		var ok bool
+		profile.gcGenerationStats, ok = requiredOffset("_gc_runtime_state", "generation_stats", versionString)
+		if !ok {
+			return layoutProfile{}, false
+		}
+	case version == 0x030f00b4:
```
grcevski: It would be nice to name this constant.

pkg/internal/cpython/runtime/manager.go
```diff
@@ -0,0 +1,337 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package runtime // import "go.opentelemetry.io/obi/pkg/internal/cpython/runtime"
+
+import (
```
grcevski: Was it not possible to find a uprobe location where we can trigger collection of Python runtime metrics asynchronously? Like we do for Go and JVM? did you explore this avenue?

As far as I can tell, the Python metrics we collect are all GC related, so the JVM approach we use would be much more efficient. We can put a probe and the sample interval, and if too many GCs happen we ignore until an event happens after the "sample" interval.

pkg/internal/cpython/runtime/memory_linux.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+// processMemory retains the mapped runtime identity used during resolution.
+type processMemory struct {
+	pid           app.PID
+	runtimeDevice uint64
+	runtimeInode  uint64
+}
+
+func newProcessMemory(pid app.PID, runtimeDevice, runtimeInode uint64) *processMemory {
```
grcevski: this can be a general utility for all runtimes, but I think it would be a lot easier to do this from the eBPF side like we do for JVM, if possible. I'd like to see if we've explored this path and why it couldn't work.

grcevski: Apart from the efficiency side, reading memory like this from a running process is inherently unsafe. The reason is that we don't synchronise with the Python runtime. So while we are reading this data, Python could be modifying it. Depends how the memory is laid out, you can read partially written data and get a completely bogus value. From an end user perspective, it would look like the data occasionally jumps to unexpected lows or highs.

The reason why this is not an issue when we do it from the eBPF side, is that we find a place right after the VM has wrtten the values, and the uprobe blocks the thread execution until it finishes, so it effectiely acts as a sync point.

pkg/internal/cpython/runtime/reader.go
```diff
@@ -0,0 +1,216 @@
...
+		return errInconsistentSample
+	}
+	return nil
+}
+
+func resolvePayload(
+	memory remoteProcessMemory,
+	plan readPlan,
+) (uint64, uint64, error) {
+	finalizing, err := readUint64(memory, plan.runtimeAddress+plan.profile.runtimeFinalizing)
```
grcevski: Missing bound checks in many places here.

pkg/internal/cpython/runtime/resolver_linux.go
```diff
@@ -0,0 +1,354 @@
...
+			return nil, err
+		}
+	}
+	return &Target{
+		memory: memory,
+		plan:   readPlan{profile: profile, runtimeAddress: runtimeAddress, identity: identity},
+	}, nil
+}
+
+func (*Resolver) analyzeMappedELF(file *elf.File) (*elfAnalysis, error) {
```
grcevski: We already do a lot of these things while processing instrumentables, it would make sense not to yet again read all exe symbols. I think we should refactor this to be called from the main path where we process symbols and be passed around, like we do for go field offsets.

grcevski (changes requested):

pkg/export/attributes/attr_defs.go
marctc: Python SDK doesn't add `process.pid` by default, so I removed it here as well. Multiple Python workers have independent cumulative counters, so I track them separately internally and aggregate their deltas into the shared service-instance metric.

If per-process runtime metrics are needed, I think `process.pid` should be added consistently for Python, Go, and JVM in a follow-up PR.

pkg/export/otel/metrics_runtime.go
marctc: This resource-key behavior is not specific to Python runtime metrics and should be handled consistently across reporters if needed, removing for now.

pkg/export/prom/prom_runtime.go
marctc: Noted, will do in a follow up PR.

pkg/internal/cpython/runtime/manager.go
marctc: Yes, I explored this. I couldn't a reliable GC probe across the supported Python builds because some builds strip the required symbol, so thats why I went for periodic userspace sampling.

pkg/internal/cpython/runtime/memory_linux.go
marctc: As mentioned above, I also evaluated periodic BPF sampling. Without a reliable in-process trigger, I could not find a way to make an externally triggered BPF program read the target Python process across OBI’s supported environments. 

The current reader therefore uses guarded duplicate reads and rejects inconsistent samples. If you know of a better approach, please let me know.

pkg/internal/cpython/runtime/manager.go
marctc: I checked again, the exact images that don't include the symbols are those builds that use `LDFLAGS="-Wl,--strip-all"` so basically alpine and slim images. Maybe unrelated, but I didn't realize until now that for those images asyncio context propagation wouldn't work either. So we can document it as limitation or keep this strategy to have maximum coverage.

grcevski: I think those will need to be a limitation, ATM, or we can do the same thing you did with the offsets here. Essentially, we can build function offsets database for the python versions and patch based on the saved offsets.

pkg/export/attributes/names/attrs.go
```diff
@@ -205,10 +205,12 @@ const (
...
-	ServiceInstanceID = Name(semconv.ServiceInstanceIDKey)
-	SkipSpanMetrics   = Name("span.metrics.skip")
-	JVMMemoryType     = Name("jvm.memory.type")
-	JVMMemoryPoolName = Name("jvm.memory.pool.name")
+	ServiceInstanceID   = Name(semconv.ServiceInstanceIDKey)
+	SkipSpanMetrics     = Name("span.metrics.skip")
+	JVMMemoryType       = Name("jvm.memory.type")
+	JVMMemoryPoolName   = Name("jvm.memory.pool.name")
+	CPythonGCGeneration = Name("cpython.gc.generation")
+	ProcessPID          = Name(semconv.ProcessPIDKey)
```
grcevski: Do we still need attribute?

pkg/export/prom/prom.go
```diff
@@ -1510,15 +1505,16 @@ func (r *metricsReporter) handleProcessEvent(pe exec.ProcessEvent, log *slog.Log
...
```
grcevski, line 1548: Why do we need this change, I thought you removed the pid split and these prom merics are not related to the runtime metrics.

pkg/export/prom/prom.go
```diff
@@ -1476,16 +1477,10 @@ func (r *metricsReporter) deleteMetricsForService(service *svc.Attrs) {
 	r.deleteRuntimeHistograms(service)
 }
 
-func (r *metricsReporter) deleteMetricsForServicePreservingRuntimeHistograms(service *svc.Attrs) {
```
grcevski, line 1507: Why was this removed? I'm trying to make sense of all the changes here, a bit of annotation from you would help.

pkg/export/prom/prom_python.go
```diff
@@ -0,0 +1,177 @@
...
+		generationLabels := append(append([]string{}, baseLabels...), strconv.Itoa(generation))
+		c.collections.DeleteLabelValues(generationLabels...)
+		c.collectedObjects.DeleteLabelValues(generationLabels...)
+		c.uncollectableObjects.DeleteLabelValues(generationLabels...)
+	}
+
+	c.valuesMu.Lock()
+	defer c.valuesMu.Unlock()
+
+	// Prometheus label values are valid UTF-8, so \xff cannot collide with label
```
grcevski: Can you explain this? Why is the delete in Python more complex than elsewhere? What's the purpose of this `/xff`?

pkg/internal/cpython/runtime/memory_linux.go
grcevski: There's so much code here so it's hard to follow where these duplicate reads are and how it's determining inconsistent samples? Can you point me to this code? At the reader level it just calls readRemote and checks for error. 

What's considered inconsistent sample?

We can read uint64 reliably if it doesn't straddle a cache line boundary, which means the address must be 8 byte aligned. Otherwise, you can read part of the old value and part of the new value.

Also we read multiple values in succession, after we read the first one, we can get preempted and go to sleep, the Python runtime writes all new values, so what we are left is some stale values and some new values.

Why not push all the offsets you have as we do for the go offsets or jvm offsets to eBPF constants and then put a return probe on `PyGC_Collect`? When the collection is about to finish we use the offsets to read the memory and publish them on the ring buffer. I think this is how the JVM runtime metrics work.

Yes it works for python binaries with symbols only, but the official docker images all have symbols and if we want to support stripped binaries we can use the offsets tracker you have to pull the function offsets.

pkg/appolly/discover/attacher.go
```diff
@@ -499,9 +539,15 @@ func (ta *traceAttacher) notifyProcessDeletion(ie *ebpf.Instrumentable) {
 		// to avoid that a new process reusing this PID could send traces
 		// unless explicitly allowed
 		ta.Metrics.UninstrumentProcess(ie.FileInfo.ExecutableName())
-		tracer.BlockPID(ie.FileInfo.Pid(), ie.FileInfo.Ns())
+		tracer.BlockPIDLifecycle(ie.FileInfo.Pid(), ie.FileInfo.Ns(), ie.FileInfo, ie.FileInfo)
```
MrAlias: A recycled fork worker can reach this deletion path after a same-inode worker deletion evicts the executable cache and the next cache miss collapses the replacement worker into its live parent. The allow path records the worker lifecycle with the parent as its service source, but this call passes the worker for both values. The Python controller consequently stages final metrics against the worker rather than the stored parent. Because only the parent received a process-created event, the dynamic gate drops the worker termination and exporters never retire its PID generation, so delayed records or PID reuse can lose or misattribute counters. Please carry the original service source and worker identity through matching creation and termination events, including explicit exporter generation registration and retirement. Add coverage for cache eviction followed by a replacement worker, dynamic selection, and delayed PID reuse.

scripts/python-offsets/main.go
```diff
@@ -0,0 +1,332 @@
...
+		}
+	}
+	changed := false
+	if item.series == "3.12" {
+		for _, name := range [][2]string{{"_PyRuntimeState", "_finalizing"}, {"_PyRuntimeState", "interpreters.main"}, {"PyInterpreterState", "gc"}, {"_gc_runtime_state", "collecting"}} {
+			changed = setNewest(track, name, item.version) || changed
+		}
+	}
+	if item.series == "3.14" {
+		changed = setNewest(track, [2]string{"_gc_runtime_state", "generation_stats"}, item.version) || changed
```
MrAlias: The updater validates all three 3.14 fields, but advances `newest` only for `_gc_runtime_state.generation_stats`. A subsequent 3.14 patch can therefore be marked supported and checkpointed while `_Py_DebugOffsets.gc` and `_Py_DebugOffsets.interpreter_state.gc` remain capped at 3.14.7. Please advance all three validated ranges and add a synthetic updater test covering this case.

MrAlias (commented):
The fork-worker lifecycle handling needs correction before this is safe to merge. A recycled child can lack a matching exporter lifecycle, allowing its final counters to be dropped or misattributed after PID reuse; see the inline comment.

Also add Python to the `application_runtime` table in `SUPPORT_MATRIX.md`. Document the three metrics and the current constraints: non-free-threaded CPython 3.9–3.14, main interpreter only, amd64 USDT or private-symbol fallback, arm64 USDT required, and Experimental stability.

pkg/runtimemetrics/reader.go
```diff
@@ -32,10 +34,15 @@ type RuntimeMetricSnapshot struct {
 	PID        app.PID
 	Generation uint64
 	Time       time.Time
+	// Removed marks a tombstone; its runtime-kind payload remains non-nil for routing.
+	Removed bool
+	// ServiceChanged marks a tombstone that retires the previous service attributes.
+	ServiceChanged bool
```
pinoOgni: I see that `ServiceChanged` is used in `metrics_runtime.go` and in `prom_python.go` but it isn't set anywhere (only in the tests).

internal/test/integration/test_python_elasticsearchclient.go
```diff
@@ -53,13 +53,17 @@ func testElasticsearchSearch(t *testing.T, dbSystemName, queryParam string) {
 func assertElasticsearchOperation(t *testing.T, dbSystemName, op, queryText, index string) {
 	params := neturl.Values{}
 	params.Add("service", comm)
+	params.Add("tags", fmt.Sprintf(
```
pinoOgni: Why this change?

pkg/ebpf/tracer.go
```diff
@@ -29,7 +29,8 @@ type Instrumentable struct {
 
 	// in some runtimes, like python gunicorn, we need to allow
 	// tracing both the parent pid and all of its children pid
```
pinoOgni: very very nit: "child pids"

bpf/generictracer/python_runtime.c
```diff
@@ -0,0 +1,119 @@
...
+
+    struct python_runtime_metric_event *event = bpf_ringbuf_reserve(&events, sizeof(*event), 0);
+    if (!event) {
+        bpf_dbg_printk("python GC event reserve failed pid=%d ns=%d", key.user_pid, key.ns);
+        return 0;
+    }
+    event->type = EVENT_PYTHON_RUNTIME_METRICS;
+    event->pid = key;
+    event->snapshot = snapshot;
+    bpf_ringbuf_submit(event, get_flags());
```
pinoOgni, line 112: This probe fires on every GC completion (like hundreds of times per second in allocation-heavy apps?), and each event pays the ringbuf submit plus the userspace export work.

So, what about adding a sampling interval like `jvm_runtime_metrics.sampling_interval`? For example, we can keep a `last_emit_ns` next to the snapshot in `python_runtime_metric_snapshots` and skip the ringbuf submit when `bpf_ktime_get_ns() - last_emit_ns < interval`, while still writing the snapshot to the map. Since the counters are cumulative and the exit drain reads the map, skipped events lose nothing.

If it makes sense for you we can do it as follow-up and in the meantime document the per-GC cost of `python_gc_done` in `python.md` doc. WDYT?

pinoOgni (changes requested):
Very COOL stuff!

bpf/generictracer/python_runtime.c
marctc: Yeah, I think a follow-up PR is better. Go runtime metrics have the same issue, so I’d leave this out of this PR and address sampling for both runtimes together.

pinoOgni (approved):

bpf/generictracer/python_runtime.c
grcevski: Good point, yeah this is how Java does it, which I think helps. Good to be a follow-up.

pkg/internal/appolly/appolly.go
```diff
@@ -218,6 +218,22 @@ func (i *Instrumenter) instrumentedEventLoop(ctx context.Context, processEvents
 	})
 }
 
+func (i *Instrumenter) dispatchCreatedProcessEvents(instrumentable *ebpf.Instrumentable) {
+	i.handleAndDispatchProcessEvent(exec.ProcessEvent{
+		Type: exec.ProcessEventCreated,
+		File: instrumentable.FileInfo,
+	})
+	for _, child := range instrumentable.ChildFileInfos {
```
grcevski: Hm, I wonder what's the impact of this here on what we make in target_info. So I'm imagining a python program that has 10 children, I think this will make a target info entry for each of the 10 children, while there should be one only. So it's questionable also what the metadata attached to those children is, if it's the same we are ok, we are overwriting the exact same data, but if it differs then we'll make duplicates.

Why did you need this?

scripts/python-offsets/main.go
MrAlias: The range advancement is fixed, but the requested regression test is still missing. Could you add a focused `validate` test covering all three validated 3.14 fields?

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -176,6 +182,31 @@ func (p *Tracer) BlockPID(pid app.PID, ns uint32) {
 	}
 }
 
+// AllowPIDLifecycle registers an exact lifecycle and a possibly shared service source.
+func (p *Tracer) AllowPIDLifecycle(pid app.PID, ns uint32, lifecycle, serviceSource *exec.FileInfo) {
+	p.pidsFilter.AllowPID(pid, ns, serviceSource, ebpfcommon.PIDTypeKProbes)
+	if p.pythonRuntime != nil {
```
grcevski: I wonder if we need these mechanics here. Essentially, all children PIDs will get matched by the same matching criteria as the parent, so we'll get all the PIDs through AllowPID. I think what we need is the opposite of what you are doing here, you need to check if the PID coming through has PPID assigned and group by the PPID. Check how things are done in matcher.

I believe I understand what you are trying to do here, but we'll be effectively adding the same PIDs twice.

grcevski (commented):
This looks pretty good! I am just unclear if we really need changes in how the PIDs are handled. I'm wondering if it's possible to use the fact that we know the parent pid for the children pids. Essentially, all children pids should get matched along with the parent, so we'll see them all coming though in all cases. I don't know if iterating over the child pid infos is neccessary.

marctc:
> This looks pretty good! I am just unclear if we really need changes in how the PIDs are handled. I'm wondering if it's possible to use the fact that we know the parent pid for the children pids. Essentially, all children pids should get matched along with the parent, so we'll see them all coming though in all cases. I don't know if iterating over the child pid infos is neccessary.

Yeah, nice catch...I was treating the typer grouping a child under its parent as if we had lost the child process and its lifecycle. Since every child is matched on its own, that extra path could register the same PID twice and create duplicate `target_info` entries.

Each PID now goes through the normal `AllowPID` path once. For Python workers, we keep the child PID and generation for lifecycle tracking, but use the parent with the same executable as the service  and metadata source.

I udpated the integration tests to add extra workers and check we only export one `target_info`.

grcevski (approved):
LGTM!

grcevski:
There's some sort of a merge error:

```
135.8 # go.opentelemetry.io/obi/pkg/appolly/discover/exec
135.8 pkg/appolly/discover/exec/file.go:39:2: StartTime redeclared
135.8 	pkg/appolly/discover/exec/file.go:35:2: other declaration of StartTime
135.8 pkg/appolly/discover/exec/file.go:59:2: startTime redeclared
135.8 	pkg/appolly/discover/exec/file.go:55:2: other declaration of startTime
135.8 pkg/appolly/discover/exec/file.go:76:3: duplicate field name startTime in struct literal
135.8 pkg/appolly/discover/exec/file.go:140:21: method FileInfo.StartTime already declared at pkg/appolly/discover/exec/file.go:136:21
171.3 make: *** [Makefile:323: compile] Error 1
```

PR #3032, opened 2026-08-14
Strengthen memcached trace-merging coverage

MrAlias:
## Summary

- keep the positive assertion that a memcached SET span is emitted
- reject traces containing more than one matching SET span

This is a test-only follow-up to #3027 for #2948. The previous `count() = 1` TraceQL filter selected healthy traces but did not reject a different merged trace containing multiple SET spans. The separate absence assertion now verifies that no trace contains multiple matching SET spans.

No runtime behavior changes.

## Validation

- `git diff --check`
- `make oats-test-memcached`

### DIFF
--- a/internal/test/oats/memcached/yaml/oats_go_memcached.yaml
+++ b/internal/test/oats/memcached/yaml/oats_go_memcached.yaml
@@ -10,13 +10,17 @@ input:
 interval: 500ms
 expected:
   traces:
-  - traceql: '{ .db.operation.name = "SET" && .db.system.name = "memcached" } | count() = 1'
+  - traceql: '{ .db.operation.name = "SET" && .db.system.name = "memcached" }'
     equals: SET
     attributes:
       db.operation.name: SET
       db.system.name: memcached
       db.query.text: session-key
       server.port: '11211'
+  - traceql: '{ .db.operation.name = "SET" && .db.system.name = "memcached" } | count() > 1'
+    count:
+      min: 0
+      max: 0
   - traceql: '{ .db.operation.name = "GET" && .db.system.name = "memcached" } | count() = 1'
     equals: GET
     attributes:


internal/test/oats/memcached/yaml/oats_go_memcached.yaml
```diff
@@ -10,13 +10,17 @@ input:
...
   traces:
-  - traceql: '{ .db.operation.name = "SET" && .db.system.name = "memcached" } | count() = 1'
+  - traceql: '{ .db.operation.name = "SET" && .db.system.name = "memcached" }'
     equals: SET
     attributes:
       db.operation.name: SET
       db.system.name: memcached
       db.query.text: session-key
       server.port: '11211'
+  - traceql: '{ .db.operation.name = "SET" && .db.system.name = "memcached" } | count() > 1'
```
grcevski, line 20: I tried this, but it didn't work for me, let's see if the tests pass.

grcevski: I think we may need to fix the oats code to be able to terminate when max = 0 and it doesn't find a thing.

MrAlias: yeah, agreed 👍

MrAlias: I was able to have this pass locally and with CI. Not sure why it was failing for you. Can you try running this branch locally to see if this still fails?

mmat11 (approved):

internal/test/oats/memcached/yaml/oats_go_memcached.yaml
grcevski: Nah, it's good. I think I genuinely had a failure, while playing with things I had forgotten to remove the disable specific go tracers and it was messing things up, as we expect, sicne it doesn't know about goroutines.

grcevski (approved):
LGTM!

PR #3037, opened 2026-08-14
gotracer: publish wrapped TLS client connections

MrAlias:
## Summary

- publish the HTTP persistConn socket tuple from the TLS write probe
- add an opt-in custom TLS dialer wrapper to the HTTP test client
- verify wrapped HTTPS requests produce attributed, non-duplicate spans

## Motivation

Follow-up to #3006. A custom `DialTLSContext` can return a wrapper around `*tls.Conn`, which prevents the round-trip probe from reading the connection directly. The TLS write probe can still resolve the underlying socket, but its `ongoing_ssl_ops` entry suppresses the nested `netFD.Write` probe before that probe publishes the connection.

Publishing from the TLS write path gives the round-trip probe the same connection handoff used for cleartext requests and prevents the attributed TLS event from being accompanied by a zero-peer client span.

## Validation

- `make generate`
- `make clang-format`
- `go test ./internal/test/integration/components/httppinger`
- `docker compose -f internal/test/integration/docker-compose-wrapped-tls-conn.yml config --quiet`
- `go test -v -run '^TestWrappedTLSConnClientSpans$' -timeout 10m ./internal/test/integration/`

### DIFF
--- a/bpf/gotracer/go_net_tls.c
+++ b/bpf/gotracer/go_net_tls.c
@@ -23,6 +23,7 @@
 
 #include <gotracer/go_common.h>
 #include <gotracer/go_large_buffer.h>
+#include <gotracer/maps/go_persist_conn.h>
 #include <gotracer/maps/ongoing_ssl_ops.h>
 
 #include <gotracer/types/net_args.h>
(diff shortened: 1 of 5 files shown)


grcevski (approved):
Nice! LGTM!

PR #3039, opened 2026-08-14
Decode HTTP/2 headers in capture order

MrAlias:
## Summary

HTTP/2 streams can finish in a different order than their HEADERS frames were received, but HPACK dynamic tables must be decoded in wire-observation order. Completion-time decoding could therefore resolve indexed fields against the wrong dynamic-table entry.

- decode request and response HPACK blocks when their HEADERS frames are observed, independently for each direction
- assemble completed spans from bounded per-stream metadata and suppress duplicate completions
- fail closed when header state is missing, truncated, evicted, or skipped by a bounded kernel scan
- account for request trailers, response CONTINUATION frames, PUSH_PROMISE, and coalesced frame buffers

## Related

- follow-up to #2924 and the gRPC method decoding report in [Discussion #2916](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/discussions/2916)
- related to #2765, which tracks broader consequences of opaque HPACK-indexed `:path` values; this PR improves decoding order and failure behavior but does not close that issue

## Testing

- `make generate/all`
- `go test ./pkg/ebpf/common ./pkg/internal/ebpf/bhpack -count=1`
- `go test -race -vet=off -p=1 ./pkg/ebpf/common ./pkg/internal/ebpf/bhpack -count=1`
- `clang-tidy generictracer/protocol_http2.h`


internal/test/integration/grpc_relay_test.go
```diff
@@ -308,15 +308,6 @@ func testGRPCRelayChainContextPropagation(t *testing.T) {
 		// Pick the completed trace for the structural checks below.
 		trace = tq.Data[0]
 
-		// all checks inside the loop so partial Jaeger indexing retries
-		relayServerSpans := trace.FindByOperationName("/relay.Relay/Relay", "server")
-		relayClientSpans := trace.FindByOperationName("/relay.Relay/Relay", "client")
-
-		require.GreaterOrEqual(ct, len(relayServerSpans), 6,
-			"should have at least 6 gRPC server spans (one per gRPC relay hop)")
-		require.GreaterOrEqual(ct, len(relayClientSpans), 6,
```
mmat11, line 317: The coalesced-frame check treats any bytes after a HEADERS frame as a desync, including the DATA frame that typically rides in the same write, so it permanently marks both directions of the connection unreliable from the first request onward.

After that the request decoder can't resolve indexed `:path` values, and the new completion path drops those spans instead of falling back to `*`, so on a reused connection the first request is traced and the rest disappear.

Are we sure we want to drop those spans?

MrAlias: The coalesced DATA case no longer poisons either decoder. `h2_next_frame_changes_hpack` now marks only the active direction unreliable, and only when the immediately following frame carries another HPACK block (`HEADERS` or `PUSH_PROMISE`); a following DATA frame leaves the connection state intact.

`TestCoalescedDataDoesNotPoisonHeaderState` covers a `HEADERS + DATA` capture followed by another request on the same connection and verifies that the indexed path still resolves. The affected Go packages, race tests, eBPF generation, `clang-format`, and `clang-tidy` pass.

pkg/ebpf/common/http2grpc_transform.go
```diff
@@ -645,6 +628,176 @@ func http2FromBuffers(parseContext *EBPFParseContext, event *BPFHTTP2Info) (requ
...
+	} else {
+		h2c.hdecRet.MarkUnreliable()
+		response = readResponseMeta(parseContext, connID, event)
+	}
+
+	if stream == nil || !stream.requestSeen {
+		h2c.hdec.MarkUnreliable()
+		return http2EventToSpan(parseContext, &pendingH2Event{event: *event, response: response})
+	}
+	if !stream.requestOK {
```
mmat11, line 765: I think this will still drop the span instead of falling back to `*` -- if we re-add back the assertions on the span count I guess they will fail

MrAlias: `http2FromBuffers` now preserves completions when request headers were observed but produced no trustworthy metadata: it builds the span with `Path="*"` instead of returning `ignore=true`. `TestUnusableRequestHeaderEventFallsBackToWildcard` covers that branch, and the relay integration test restores wildcard-aware server and client span-count assertions. The affected unit suites pass normally and with the race detector; the integration package compiles.

mmat11 (approved):
lgtm if tests pass

PR #3043, opened 2026-08-15
generictracer: stop reporting established idle connections as failed connects

### LINKED ISSUE #3040: Idle connections closed by the local process are reported as failed connects, and the CONNECT span is attached to an unrelated in-flight trace
### Component(s)

Failed-connect detection and trace-context propagation
(`bpf/common/sockaddr.h`, `bpf/generictracer/k_tracer.c`,
`bpf/generictracer/failed_connect.h`, `bpf/generictracer/protocol_tcp.h`,
`bpf/common/trace_parent.h`, `bpf/common/trace_helpers.h`)

### Nature of the bug

A TCP connection that completed its handshake, transferred no data, and
was then closed by the local process is classified as a failed connect. OBI
emits a `CONNECT` client span with `status = error` whose duration is the
socket's entire lifetime. That span is then attached as a child of whatever
request the closing thread happens to be serving, producing a child span that
starts tens of seconds before its own trace's root span.

Connection pools produce these sockets routinely, so a healthy service emits
error spans and healthy traces acquire multi-second durations. In our case a
242ms HTTP 200 became a 33.9s error trace in Tempo.

https://github.com/mifisignal/obi-failed-connect-repro contains test cases
to reproduce the issue.

#### 1. `is_tcp_socket_never_connected` treats zero application bytes as a failed handshake

`bpf/common/sockaddr.h:140-160`:

```c
    if (sk_state == TCP_SYN_SENT || sk_state == TCP_SYN_RECV) return true;
    struct tcp_sock *tp = (struct tcp_sock *)sk;
    const u64 bytes_sent = BPF_CORE_READ(tp, bytes_sent);
    const u64 bytes_received = BPF_CORE_READ(tp, bytes_received);
    return (bytes_sent == 0 && bytes_received == 0);
```

`tcp_sock->bytes_sent` counts *application* bytes passed to `tcp_sendmsg` (RFC
4898 `tcpEStatsPerfHCDataOctetsOut`); `bytes_received` counts payload delivered
to the receive queue. Neither counts SYN, SYN-ACK, or FIN. A connection that
handshook successfully and never carried a request therefore has both at zero,
which this predicate reports as never connected.

The `skc_state` check on the preceding lines already covers a socket that never
left `SYN_SENT`. The byte-counter fallback adds false positives.

#### 2. `established` means "carried data"

The second gate in `kprobe/tcp_close` (`bpf/generictracer/k_tracer.c:744`) reads
as an independent second opinion:

```c
    if (ct && !ct->established && !ct->failed) {
        failed_to_connect_event(&info, k_lw_thread_none, orig_dport, ct->ts);
    }
```

`cp_support_established()` (`k_tracer.c:280-285`) is called from exactly four
sites, all of them **data-transfer** paths: `tcp_sendmsg` (`:451`),
`tcp_rate_check_app_limited` (`:594`), and two in the
`tcp_recvmsg`/`tcp_cleanup_rbuf` handling (`:998`, `:1066`).
`setup_cp_support_conn_info()` (`:292-322`) creates the entry with
`.established = 0` for a client connect (`:305`), and `kretprobe/sys_connect`
(`:325-386`) calls it at `:364` and thereafter only ever writes
`cp_data->failed = 1` (`:378`). No path sets `established`, including
`res == 0`.

Reproduction scenarios `A1` and `A1B` confirm this from the outside: they differ
only in whether `connect()` returned `-EINPROGRESS` or 0, and both fire 6/6. How
the connect returned has no bearing on the outcome, and the gate restates defect
1 instead of checking it.

#### 3. A received FIN distinguishes the firing case from the silent one

Scenarios where the peer closes first produce no spans. With
`OBI_BPF_DEBUG=true`, the `A1` scenario prints the probe's own line at every close:

```
msg="possibly never connected sock: id=10981, sock=ffff0003da031300, ct=ffff800086c0"
msg="TCP connect failed event [failed_to_connect_event]"
```

In the peer-closes cases that line is absent across the whole run, which places
`is_tcp_socket_never_connected` itself as the branch returning false, before the
`established` gate is reached. A case in which the client sleeps through the
peer's FIN and never calls `recv` is also silent, so the application's reads are
not what changes the predicate's answer.

Both cases carry zero application bytes, and they differ only in a FIN arriving
before the close. Reading the counters the predicate reads, from a `bpftrace`
probe on `tcp_close`, over two sockets that differ only in which side closed:

```
dport=9101 state=1 bytes_sent=0 bytes_received=0 segs_out=2 segs_in=1   # local process closes
dport=9102 state=8 bytes_sent=0 bytes_received=1 segs_out=3 segs_in=2   # peer sent FIN first
```

A FIN consumes a sequence number, and the kernel folds that advance into
`tp->bytes_received`, moving the counter from 0 to 1 on a connection that
carried no application data. `state=1` is `TCP_ESTABLISHED` and `state=8` is
`TCP_CLOSE_WAIT`, so neither socket is caught by the `skc_state` check and the
byte clause decides both outcomes.

The consequence is that the classification turns on whether a peer's FIN arrived
before `close()`, which is independent of whether the connection was
established. It also decides who is exposed: a client that closes its own idle
pooled connections is affected, and a client whose peer hangs up first is
shielded.

#### 4. The span is attached to whatever the closing thread is serving

`failed_to_connect_event` (`bpf/generictracer/protocol_tcp.h:235-253`) calls
`tcp_get_or_set_trace_info(...)` at `:250`. Since `req->direction == TCP_SEND`,
this takes the client branch (`protocol_tcp.h:76-86`) into `find_parent_trace`
(`bpf/common/trace_parent.h:276-336`), which resolves `server_traces` by the
current thread's `trace_key_t` (`:322`, via `find_parent_process_trace`,
defined `:112-140`). The fallback at `:328-333` reads the thread key stored at
connect time and has the same shape one step removed. Neither considers whether
the socket belongs to that request.

The guard that would reject this is `should_be_in_same_transaction`
(`bpf/common/trace_helpers.h:40-49`):

```c
    if (child_tp->ts < parent_tp->ts) return 0;
    const u64 diff = child_tp->ts - parent_tp->ts;
    return diff < max_transaction_time;
```

It admits the relationship because the timestamp validating it differs from the
timestamp used as the span start. `failed_to_connect_event` sets `req->tp.ts` to
`trace_ts`, which it passes as the close time (`protocol_tcp.h:243, 246`),
while `start_monotime_ns` is the connect time
(`bpf/generictracer/failed_connect.h:28, 38`). The guard sees a child a few
microseconds newer than the parent and admits it; the span is then emitted
starting seconds earlier than its own trace root.

With `trace_ts` set to `connect_ts`, the guard's first line would reject the
relationship and `init_new_trace` would give the event a fresh trace. That is
what happens in `A1`, where the closing thread has no request in flight and the
span carries `parentSpanId = (none)`.

This defect applies to genuine failed connects as well.

### Corroboration from OBI's other detector

OBI's later failed-connection detector, added in `38119709` ("Introduce
StatsO11y (#1572)"), uses a different signal. `bpf/statsolly/tp_tcp.c:125-130`:

```c
    // Trust sk_err: err==0 means the kernel saw no problem ...
    // Exception: aborted connect (TCP_SYN_SENT -> TCP_CLOSE) never established, still a failure.
    if (err == 0 && args->oldstate != TCP_SYN_SENT) { return 0; }
```

Running both detectors over the same sockets in one run, the generic tracer
emitted 6 `CONNECT` error spans for the idle connections while StatsO11y
reported 0. For a connect dying in `SYN_SENT` both reported 6. Each of those
runs also produced one StatsO11y stat for `169.254.169.254`, unreachable from
inside the container, which shows the detector was live and reporting genuine
failures while it passed over the sockets the generic tracer called failures.

### Observations

An Apollo Router 2.10 (Rust/hyper) pod issuing GraphQL subgraph calls to two
Node/Express peers and one Python/uvicorn peer over plaintext HTTP/1.1 emitted
88 `CONNECT` spans, all `status = error`, over three hours, against roughly
36,000 successful subgraph client calls. Durations ran from 5088ms to
61529ms. No actual connection errors actually occurred.

Sampling `ss -tanoi` inside both router pods at 3s intervals for 358s per pod
found these sockets in the kernel, established and idle, with every counter
frozen:

```
ESTAB 0 0 10.0.242.34:53898 172.20.244.160:4001 timer:(keepalive,58sec,0)
   cwnd:10 bytes_acked:1 segs_out:2 segs_in:1 ... lastsnd:1470  lastrcv:1470  lastack:1470
   cwnd:10 bytes_acked:1 segs_out:2 segs_in:1 ... lastsnd:19480 lastrcv:19480 lastack:19480
```

`segs_out:2` is SYN plus the handshake ACK, `segs_in:1` is the SYN-ACK, and
`bytes_acked:1` is the SYN's own sequence byte. There are no `data_segs_out` /
`data_segs_in` fields, and `rcvmss:536` was never revised upward. A working
pooled socket to the same peer showed `bytes_sent:88038 bytes_received:87768
segs_out:264 data_segs_out:150`. Across 240 snapshots, `SYN-SENT` to any peer
port was 0.

All observed zero-byte sockets went `ESTAB → TIME_WAIT` directly, with zero
`CLOSE-WAIT` observations, so the router was the active closer. Apollo Router
builds its subgraph client on `hyper_util::client::legacy::Client`, whose pool
races a checkout against a new connect and, when the checkout wins, spawns the
started connect to completion and inserts it into the pool as an idle
connection (`hyper-util/src/client/legacy/client.rs:404-450`). Any HTTP client
pool with this shape produces these sockets.

### Impact

- Unrelated healthy traces are corrupted. A 242 ms HTTP 200 is reported as a
  33.9s error trace. This defeats latency analysis, `status = error` trace
  search, and duration-based tail-sampling or alerting, silently, because the
  application's own logs are clean.
- A real connect failure is indistinguishable from a routine pool reap.

### (Tentative) Suggested fixes

1. Drop the `bytes_sent == 0 && bytes_received == 0` clause from
   `is_tcp_socket_never_connected`, or replace the predicate with the
   `statsolly` test (`sk_err`, plus `oldstate == TCP_SYN_SENT`).
2. Set `established` at connect completion, so the `tcp_close` gate becomes an
   independent check. This needs to cover the non-blocking path, where
   `connect()` returns `-EINPROGRESS` and no current probe observes the
   handshake completing.
3. Make `tp.ts` equal `start_monotime_ns` in `init_failed_connect_tcp_req`, so
   the existing `should_be_in_same_transaction` guard applies to the span's real
   start time. This alone removes the cross-trace attachment for both
   false-positive and genuine failed-connect events.

Independently, an invariant that a child span may not start before its parent
would be worth having; nothing currently enforces it.

### Version / environment

- Reproduced on `otel/ebpf-instrument:main`, digest
  `sha256:6f38eb3e3c5d556773002855a660e0e96b787f630bd3e6886aabfb3e36c574db`,
  revision `7b96052`, aarch64 (Graviton3, Ubuntu 24.04). All three scenarios
  give the same result on kernel `6.17.0-1019-aws` and on kernel
  `7.0.0-1010-aws`.
- Originally observed on Grafana Beyla v3.31.0, vendoring OBI at `7d018fe`
  ("Release v3.31.0 artifacts"), kernel 6.18.39 aarch64, Kubernetes 1.36 on AWS
  EKS Auto Mode. `bpf/common/sockaddr.h` is byte-identical between that build
  and upstream `main` (`d6f4f521`, 2026-08-14), and the `tcp_close` /
  `failed_to_connect_event` logic is materially identical.
- The feature originates in `4dac5549` "Report failed connects (#737)"
  (2025-10-07). The commit body is the bare PR title, so there is no recorded
  justification for the byte-counter clause. If it was meant to catch a case the
  state check misses, a comment on that line would help.
- The production build carries local patches to DNS capture in `generictracer`
  (see [beyla#2998](https://github.com/grafana/beyla/issues/2998)). A diff
  against the upstream base confirms none of them touch any file in the
  failed-connect path, and the reproduction above runs on a stock upstream
  image.

### AI assistance disclosure

Per the project's Generative AI Policy: this report was drafted with AI
assistance. The source analysis, the reproduction, the live socket captures, and
the comparison against upstream `main` were gathered from a running system and a
local checkout, and reviewed by a human before filing.

mifisignal:
Fixes #3040.

## Symptom

A TCP connection that completed its handshake, transferred no data, and
was closed by the local process was incorrectly classified as a failed connect. 
OBI emitted a `CONNECT` client span with `status = error` spanning the socket's 
whole lifetime, attached to whatever request the closing thread was serving. Client
connection pools produce these sockets routinely, so healthy services emitted
error spans and healthy traces acquired multi-second durations; the reporting
case measured a 242ms HTTP 200 arriving in Tempo as a 33.9s error trace.

## Changes

**Classification** (`bpf/common/sockaddr.h`). `is_tcp_socket_never_connected()`
treated `bytes_sent == 0 && bytes_received == 0` as a failed handshake, but
those counters are application payload only — a connection that established and
never carried a request reads 0 on both.

The predicate now uses two values the kernel maintains for the socket's
lifetime: its state and `bytes_acked`. `TCP_SYN_SENT` and `TCP_SYN_RECV` mean
the handshake never finished. `TCP_CLOSE` is ambiguous; a socket reaches it 
both when a connect fails and when a working connection is closed normally. 
`bytes_acked` distinguishes the two. A SYN consumes one byte of sequence space, 
and the kernel adds that byte once the peer acknowledges it, so any socket whose 
handshake completed reads at least 1 even if it never carried data, and one whose 
handshake did not reads 0. Thus, refused, reset and timed-out connects stay 
correctly reported.

This also removes an asymmetry the byte counters introduced: a FIN advances
`bytes_received` to 1 for its own sequence byte, so of two identical idle
sockets, the one closed by this process fired and the one whose peer hung up
first did not.

**Establishment** (`bpf/generictracer/k_tracer.c`). `cp_support_data_t.established`
was set only from the four data-transfer paths, so the `tcp_close` gate that
reads it repeated the byte test rather than checking it independently.
`kretprobe/sys_connect` now sets it when `connect()` returns 0. The
`-EINPROGRESS` path completes in softirq, where no probe of ours observes it, so
those sockets are still decided by their state at close time.

**Span timestamp** (`bpf/generictracer/failed_connect.h`). `tp.ts` was the close
time while the span's `start_monotime_ns` was the connect, so
`should_be_in_same_transaction()` validated the parent against a timestamp taken
microseconds earlier and admitted any request in flight. `tp.ts` is now the
connect, the instant the span reports, so the existing guard rejects a parent
that did not exist when the connect was made. This applies to failed connects as
well as the false positives.

**Parent retirement** (`bpf/common/trace_parent.h`). Correcting the timestamp
makes the guard reject more parents, and a rejected parent was always marked
invalid. That's correct for a parent left behind by a transaction that ended long
ago, and incorrect for one that started after its candidate child: it is serving a
live transaction, and retiring it would drop the traces of the requests it is
still handling. Only the first case now retires the parent.

## Tests

`bpf/tests/test_failed_connect_classify.c` covers the predicate; the
established-idle case fails against the previous code.
`test_failed_connect_event.c` gains an assertion that the trace timestamp equals
the span start. The test stubs gain `struct tcp_sock`, the TCP state enum, the
`sock_common` fields the predicate reads, `BPF_CORE_READ_INTO` and a
`bpf_endian.h`.

`TestSuite_FailedConnect` drives a Go workload (`skip_go_specific_tracers`, so
it lands in the generic tracer) that closes idle pooled connections from inside
a request handler while a second goroutine connects to a port nobody listens on.
It asserts the workload is instrumented, a refused connect is still reported, no
`CONNECT` span names the idle peer's port, and no span starts before its parent.

Without this fix, the suite reports idle sockets as `CONNECT` errors, each
`CHILD_OF` a `processing` span starting 3s after it, and both new assertions
fail. With the fix, all four now pass, as do `TestSuiteGoGeneric`,
`TestSuiteNestedTracesMaxTransactionTime` and `TestExistingSocketsDetection`,
which cover the propagation and transaction-window paths the `trace_parent.h`
change touches. Verified on kernel 7.0.0-1010-aws, aarch64.

## AI assistance disclosure

Per the project's Generative AI Policy: this change was developed with AI
assistance. The fix, the unit tests, and the integration suite were written
against a local checkout, the test runs recorded above were executed on a
dedicated host, and all of it was reviewed by a human before submission.

### DIFF
--- a/bpf/common/trace_helpers.h
+++ b/bpf/common/trace_helpers.h
@@ -49,6 +49,16 @@ static __always_inline u8 should_be_in_same_transaction(const tp_info_t *parent_
     return diff < max_transaction_time;
 }
 
+// Tells apart the two reasons should_be_in_same_transaction() rejects a
+// candidate parent. A parent left behind by a transaction that ended long ago
+// is stale and the caller retires it. A parent that started after the child
+// did is simply not this child's parent: it belongs to a transaction still in
+// flight, and retiring it would drop the trace of the requests it is serving.
+static __always_inline u8 parent_trace_is_stale(const tp_info_t *parent_tp,
+                                                const tp_info_t *child_tp) {
+    return child_tp->ts >= parent_tp->ts;
+}
+
 static __always_inline void init_new_trace(tp_info_t *tp) {
     bpf_d_printk("Generating new traceparent id [%s]", __FUNCTION__);
     new_trace_id(tp);
--- a/bpf/common/trace_parent.h
+++ b/bpf/common/trace_parent.h
@@ -363,10 +363,12 @@ find_trace_for_client_request_with_t_key(const pid_connection_info_t *p_conn,
         bpf_dbg_printk("Found existing server tp for client call");
 
         if (!should_be_in_same_transaction(&server_tp->tp, tp)) {
-            bpf_dbg_printk("Parent and child are too far apart, marking server trace as invalid");
+            bpf_dbg_printk("Parent and child are too far apart, discarding the parent trace");
             bpf_dbg_printk(
                 "%lld >>> %lld (max: %lld)", tp->ts, server_tp->tp.ts, max_transaction_time);
-            server_tp->valid = 0;
+            if (parent_trace_is_stale(&server_tp->tp, tp)) {
+                server_tp->valid = 0;
+            }
             return 0;
         }
 
@@ -404,10 +406,12 @@ find_parent_trace_for_client_request_with_t_key(const pid_connection_info_t *p_c
         bpf_dbg_printk("Found existing server tp for client call");
 
         if (!should_be_in_same_transaction(&server_tp->tp, tp)) {
-            bpf_dbg_printk("Parent and child are too far apart, marking server trace as invalid");
+            bpf_dbg_printk("Parent and child are too far apart, discarding the parent trace");
             bpf_dbg_printk(
                 "%lld >>> %lld (max: %lld)", tp->ts, server_tp->tp.ts, max_transaction_time);
-            server_tp->valid = 0;
+            if (parent_trace_is_stale(&server_tp->tp, tp)) {
+                server_tp->valid = 0;
+            }
             return 0;
         }
 
--- a/bpf/common/sockaddr.h
+++ b/bpf/common/sockaddr.h
@@ -137,24 +137,42 @@ parse_sockaddr_info(const u32 pid, struct sockaddr *addr, connection_info_part_t
     return false;
 }
 
-static __always_inline bool is_tcp_socket_never_connected(struct sock *sk) {
-    if (!sk) {
-        return true;
-    }
-
-    // Read the socket state
-    const u8 sk_state = BPF_CORE_READ(sk, __sk_common.skc_state);
-
-    // Socket was never connected if it's in these states:
+// Decides whether a socket's handshake ever completed, from the two values the
+// kernel keeps for its whole lifetime.
+//
+// A socket that completed its handshake never returns to TCP_SYN_SENT or
+// TCP_SYN_RECV, so either state answers the question on its own. Every other
+// state except TCP_CLOSE is only reachable after the handshake completed.
+//
+// TCP_CLOSE is reached both ways: tcp_done() puts a connect that was refused,
+// reset or timed out there, and so does a close that ran to completion after a
+// normal exchange. bytes_acked separates them, because tcp_snd_una_update()
+// counts the SYN's own sequence byte when the peer acknowledges it. It is
+// therefore at least 1 for every socket whose handshake completed, whatever the
+// connection carried afterwards, and 0 for one whose handshake did not. A
+// socket the process never connected also reads 0 here; the callers gate on a
+// connect-time entry, so it never reaches this predicate.
+static __always_inline bool tcp_never_connected(u8 sk_state, u64 bytes_acked) {
     if (sk_state == TCP_SYN_SENT || // Connection attempt in progress
         sk_state == TCP_SYN_RECV    // SYN received but not established
     ) {
         return true;
     }
 
-    struct tcp_sock *tp = (struct tcp_sock *)sk;
-    const u64 bytes_sent = BPF_CORE_READ(tp, bytes_sent);
-    const u64 bytes_received = BPF_CORE_READ(tp, bytes_received);
+    if (sk_state != TCP_CLOSE) {
+        return false;
+    }
 
-    return (bytes_sent == 0 && bytes_received == 0);
+    return bytes_acked == 0;
+}
+
+static __always_inline bool is_tcp_socket_never_connected(struct sock *sk) {
+    if (!sk) {
+        return true;
+    }
+
+    const u8 sk_state = BPF_CORE_READ(sk, __sk_common.skc_state);
+    const struct tcp_sock *tp = (const struct tcp_sock *)sk;
+
+    return tcp_never_connected(sk_state, BPF_CORE_READ(tp, bytes_acked));
 }
--- a/bpf/generictracer/failed_connect.h
+++ b/bpf/generictracer/failed_connect.h
@@ -15,7 +15,6 @@ static __always_inline void init_failed_connect_tcp_req(tcp_req_t *req,
                                                         u16 orig_dport,
                                                         u64 connect_ts,
                                                         u64 end_ts,
-                                                        u64 trace_ts,
                                                         u64 extra_id,
                                                         const pid_info *pid) {
     __builtin_memset(req, 0, sizeof(*req));
@@ -35,5 +34,10 @@ static __always_inline void init_failed_connect_tcp_req(tcp_req_t *req,
     req->pid = *pid;
     req->buf[0] = '\0';
 
-    req->tp.ts = trace_ts;
+    // The event is built at close time, but the span starts at the connect.
+    // tp.ts is what should_be_in_same_transaction() compares against a
+    // candidate parent's start, so it has to be the same instant the span
+    // reports, or a connect made before the parent request existed passes a
+    // check meant to reject exactly that.
+    req->tp.ts = connect_ts;
 }
--- a/bpf/generictracer/k_tracer.c
+++ b/bpf/generictracer/k_tracer.c
@@ -362,6 +362,15 @@ int BPF_KRETPROBE_GUARDED(obi_kretprobe_sys_connect, int res) {
 
         setup_cp_support_conn_info(&info.p_conn, true);
 
+        // connect() returning 0 means the handshake completed, so the socket
+        // cannot be a failed connect from here on, whether or not it ever
+        // carries data. The non-blocking path returns -EINPROGRESS and
+        // completes in softirq, where no probe of ours observes it, so for
+        // those the socket's own state at close time remains the only answer.
+        if (res == 0) {
+            cp_support_established(&info.p_conn);
+        }
+
         tracked_connection_t t_conn = {
             .time = bpf_ktime_get_ns(),
             .direction = TCP_SEND,
--- a/bpf/generictracer/protocol_tcp.h
+++ b/bpf/generictracer/protocol_tcp.h
@@ -230,7 +230,7 @@ static __always_inline void failed_to_connect_event(pid_connection_info_t *pid_c
         const u64 event_ts = bpf_ktime_get_ns();
         const u64 extra_id = extra_runtime_id();
         init_failed_connect_tcp_req(
-            req, pid_conn, orig_dport, connect_ts, event_ts, event_ts, extra_id, &pid);
+            req, pid_conn, orig_dport, connect_ts, event_ts, extra_id, &pid);
 
         bpf_dbg_printk("TCP connect failed event");
 
--- a/bpf/tests/bpfcore/bpf_core_read.h
+++ b/bpf/tests/bpfcore/bpf_core_read.h
@@ -26,6 +26,16 @@
         __r;                                                                                       \
     })
 
+// Copies a zero value of the field's type, so an array field lands in an array
+// destination the way the real macro's bpf_probe_read() does.
+#define BPF_CORE_READ_INTO(dst, src, ...)                                                          \
+    do {                                                                                           \
+        __typeof__(___bpf_apply(___bpf_arrow, ___bpf_narg(__VA_ARGS__))(src,                       \
+                                                                        ##__VA_ARGS__)) __v = {};  \
+        (void)(src);                                                                               \
+        __builtin_memcpy((dst), &__v, sizeof(__v));                                                \
+    } while (0)
+
 #define BPF_CORE_READ_STR_INTO(dst, src, ...) ((void)(src), 0)
 #define bpf_core_field_exists(field) (0)
 #define bpf_core_enum_value_exists(t, v) (0)
--- a/bpf/tests/bpfcore/bpf_endian.h
+++ b/bpf/tests/bpfcore/bpf_endian.h
@@ -1,0 +1,11 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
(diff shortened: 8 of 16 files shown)


bpf/common/trace_helpers.h
```diff
@@ -48,6 +48,16 @@ static __always_inline u8 should_be_in_same_transaction(const tp_info_t *parent_
...
 }
 
+// Tells apart the two reasons should_be_in_same_transaction() rejects a
+// candidate parent. A parent left behind by a transaction that ended long ago
+// is stale and the caller retires it. A parent that started after the child
+// did is simply not this child's parent: it belongs to a transaction still in
+// flight, and retiring it would drop the trace of the requests it is serving.
+static __always_inline u8 parent_trace_is_stale(const tp_info_t *parent_tp,
+                                                const tp_info_t *child_tp) {
+    return child_tp->ts >= parent_tp->ts;
```
grcevski, line 59: I'm not sure about this check... It's normal for a child timestamp to be higher than the parent timestamp. Do we want to check the reversed?

bpf/common/trace_parent.h
```diff
@@ -348,10 +348,12 @@ find_trace_for_client_request_with_t_key(const pid_connection_info_t *p_conn,
         bpf_dbg_printk("Found existing server tp for client call");
 
         if (!should_be_in_same_transaction(&server_tp->tp, tp)) {
-            bpf_dbg_printk("Parent and child are too far apart, marking server trace as invalid");
+            bpf_dbg_printk("Parent and child are too far apart, discarding the parent trace");
             bpf_dbg_printk(
                 "%lld >>> %lld (max: %lld)", tp->ts, server_tp->tp.ts, max_transaction_time);
-            server_tp->valid = 0;
+            if (parent_trace_is_stale(&server_tp->tp, tp)) {
+                server_tp->valid = 0;
```
grcevski, line 370: Can we add an additional dbg printf here, so in the logs we know we now marked it as invalid?

bpf/common/trace_helpers.h
grcevski: Oh I think we are using this only in the context of the should be in the same transaction. I think in that case we need to change the name and make it more specific, since it's very likely someone will just use that elsewhere.

grcevski (approved):
LGTM! Minor comments about the naming on thing, otherwise thanks for this patch.

bpf/common/trace_parent.h
mifisignal: Done!

bpf/common/trace_helpers.h
mifisignal: Done!

grcevski:
Hm, I don't understand the test failure, must be a BPF verifier issue. Let me kick of some tests.

mifisignal:
@grcevski any idea what's going on with these test failures? they don't look related to the work.

grcevski:
No idea, first time I'm seeing this happen. They die right away and BPF verifier tests looked all good. Can you force push and sync with main, I hope that's enough to clear whatever is happening to the CI.

mifisignal:
@grcevski we've got some flapping issues caused by timeouts getting triggered:

Two failures today, 32478912585 and 32511703701, are both a single kernel (al2023-6.1.172-216.329.amzn2023) hitting `panic: test timed out after 50m0s` in the `gotracer/Bpf` matrix. Individual program loads are taking 15–31s each inside QEMU on that kernel, and the 1024-combination gotracer cross-product doesn't finish. The job's timeout-minutes: 60 and the test's 50m deadline are simply too tight  for the slowest emulated kernel. Fixing it means either trimming that cross-product (it is the largest matrix in the test by an order of magnitude) or raising both budgets.

I'm unsure why some of these kernel builds are failing. I can't replicate them on my own build infrastructure. On the verifier host, at `a78a4951` with a fresh make docker-generate, `TestBPFVerifierWithConstants` passes 622/622, as does current main (`bbf2d296`). A -race build of the same test also passes  lean with no data races.

mifisignal:
@grcevski not sure if this is in your purview, but the kernel verification matrix testing would go much, much faster if the project were able to use a cloud instance with nested virtualization support. My validations on a c7i.8xlarge with nested virturalization enabled (kvm support) are very fast.

mmat11 (approved):

PR #3055, opened 2026-08-17
logenricher: don't lose log lines of short-lived processes

mmat11:
## Summary

The enricher blanks the original log line in the writing process' buffer and
re-writes the enriched line asynchronously, opening the destination lazily as
/proc/<pid>/fd/<n> at write time. If the process exits before the async writer
gets there, the open fails with ENOENT and the line is dropped. The
line is lost entirely, and OBI logs an ERROR for every one:

    failed to open log file for writing ... /proc/160146/fd/2: no such file or directory

Open the destination at capture time instead, while the process is still
alive: an open file description keeps the log pipe writable after the process
exits, so the async write still lands. The residual window (process dies
between the write syscall and the ring buffer drain) is inherent to dying
processes, so that ENOENT is now logged at debug level instead of error.

## Validation

### DIFF
--- a/pkg/internal/ebpf/logenricher/logenricher.go
+++ b/pkg/internal/ebpf/logenricher/logenricher.go
@@ -349,11 +349,46 @@ func (p *Tracer) handleLogEvent(record *ringbuf.Record) (request.Span, bool, err
 		return request.Span{}, true, nil
 	}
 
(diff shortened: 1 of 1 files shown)


grcevski (approved):
LGTM!

PR #3065, opened 2026-08-17
Attach Go probes to all matching symbol copies

### LINKED ISSUE #3063: Attach Go probes to all matching symbol copies
## Summary

Go binaries can contain both canonical symbols and one or more `/vendor/`
aliases, and more than one copy can be active. The current Go offset contract
resolves each requested function to one `FuncOffsets`, so only one copy receives
probes.

#3020 makes that single-target choice deterministic: usable exact names take
precedence over vendor aliases, with a stable alias fallback. It intentionally
does not cover binaries where multiple matching copies execute.

## Scope

- Extend Go symbol resolution to retain every usable exact or vendored match for
  a requested function.
- Update probe attachment to attach to every resolved copy.
- Apply compatibility and signature handling per copy so one incompatible match
  does not suppress instrumentation of another compatible match.
- Deduplicate matches that resolve to the same attachment offsets.
- Keep resolution and attachment behavior deterministic across symbol-table
  traversal orders.

## Acceptance criteria

- A binary containing active canonical and vendored copies of a requested Go
  function is instrumented at both copies.
- Multiple vendored copies are each instrumented when compatible.
- An incompatible copy does not prevent a compatible copy from being attached.
- Duplicate offsets do not produce duplicate attachments.
- Tests cover exact plus vendored matches, multiple aliases, compatibility
  differences, deduplication, and different discovery orders.

## Context

- Review raising the multi-copy case:
  https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3020#pullrequestreview-4938435312
- Follow-up discussion and proposed separation of scope:
  https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3020#issuecomment-5295789468
- #3021 covers the immediate #2769 HTTP/2 paths with explicit symbols and
  separate probe groups. This issue tracks the general multi-copy solution.

MrAlias:
## Summary

- retain every usable canonical or `/vendor/` match for each requested Go function
- attach ordinary probes to every compatible resolved copy
- partition atomic probe groups by canonical or vendor namespace
- sort and deduplicate attachment offsets independently of symbol-table traversal order

Fixes #3063.

## Why

Go binaries can contain active canonical and vendored copies of the same requested function. The existing contract established in #3020 selected one deterministic match, so other executable copies remained uninstrumented.

This changes the offset contract from one result per requested name to an ordered list that preserves the actual resolved symbol identity. Probe descriptors are cloned per copy, keeping per-copy start and return offsets independent.

## Invariants and compatibility

- Binaries with one usable match retain the existing attachment behavior.
- Canonical and vendored copies are all retained when they have distinct attachment addresses.
- Matches sharing a start address are merged so they cannot create duplicate entry or return attachments.
- Return-probe compatibility is evaluated per copy; a copy without usable return offsets does not suppress another compatible copy.
- Atomic groups are formed only from symbols in the same canonical or vendor namespace. An incomplete group in one namespace does not suppress a complete group in another, and symbols from different copies are never combined.

## Validation

- `go test ./pkg/internal/goexec ./pkg/ebpf ./pkg/appolly/discover ./pkg/internal/ebpf/gotracer`
- `make compile`
- targeted `golangci-lint` for the four affected packages
- `git diff --check`

## Scope

This PR completes the general multi-copy resolution and attachment contract. It intentionally does not include the HTTP/2 traceparent ownership state machine or generated eBPF changes from #3021; those remain separate from this symbol-resolution work.

### DIFF
--- a/pkg/appolly/discover/typer_test.go
+++ b/pkg/appolly/discover/typer_test.go
@@ -71,11 +71,11 @@ func TestLoadAllGoFunctionNamesIncludesConditionalGoTracerSymbols(t *testing.T) 
 }
 
 func TestContextWithValueDoesNotQualifyGoProxy(t *testing.T) {
-	assert.True(t, isGoProxy(&goexec.Offsets{Funcs: map[string]goexec.FuncOffsets{
+	assert.True(t, isGoProxy(&goexec.Offsets{Funcs: map[string][]goexec.FuncOffsets{
 		"runtime.chansend1": {},
 		"context.WithValue": {},
 	}}))
-	assert.False(t, isGoProxy(&goexec.Offsets{Funcs: map[string]goexec.FuncOffsets{
+	assert.False(t, isGoProxy(&goexec.Offsets{Funcs: map[string][]goexec.FuncOffsets{
 		"runtime.chansend1":                {},
 		"context.WithValue":                {},
 		"net/http.serverHandler.ServeHTTP": {},
--- a/pkg/ebpf/instrumenter.go
+++ b/pkg/ebpf/instrumenter.go
@@ -15,6 +15,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"strings"
 	"sync"
 	"syscall"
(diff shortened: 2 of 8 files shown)


grcevski (approved):
LGTM! Thanks for following up with this.

PR #3066, opened 2026-08-17
Support protocol-scoped filters in Config V2

### LINKED ISSUE #1282: Filtering (for traces) should allow to skip unaffected spans

MrAlias:
## Summary

- Translate each Config V2 instrumentation filter into the runtime's protocol- and signal-scoped filter representation added by #2996.
- Export scoped runtime filters back to their matching Config V2 protocol and signal sections.
- Preserve V1 migration coverage for legacy application filters that are fanned out across Config V2.

## Motivation

Config V2 already models application filters independently for each protocol and telemetry signal. The runtime converter nevertheless required every protocol's trace and metric filters to match the HTTP trace filter because the runtime previously had only one global application filter.

That restriction prevented Config V2 from using the scoped filtering behavior now available in the runtime.

## Behavior

An HTTP trace filter now affects only HTTP trace spans. HTTP metrics and spans produced by unconfigured protocols continue through the pipeline unchanged. Different protocols and signals can use different filter definitions.

Legacy Config V1 `filter.application` behavior remains global. Network capture and network stats filters remain shared between their trace and metric representations because their runtime filters are still global.

## Testing

- Config V2 import coverage for single-signal, cross-signal, cross-protocol, Aerospike, and legacy migration cases.
- Runtime-to-V2 export coverage for scoped filters.
- V1 migration command coverage.
- `make verify`.

Closes #1282

### DIFF
--- a/cmd/obi/internal/configcmd/configcmd.go
+++ b/cmd/obi/internal/configcmd/configcmd.go
@@ -746,6 +746,7 @@ func migrationAlias(path string) bool {
 	}
 	for _, prefix := range []string{
 		"executable_path",
+		"filter.application",
 		"open_port",
 		"target_pids",
 		"network.enable",
--- a/internal/config/convert/export.go
+++ b/internal/config/convert/export.go
@@ -195,7 +195,7 @@ func protocolInstrumentation(
 ) schema.ProtocolInstrumentation {
 	return schema.ProtocolInstrumentation{
 		Enabled: protocolEnabled(tracesInstrumentations, metricsInstrumentations, appMetricsEnabled, mapping),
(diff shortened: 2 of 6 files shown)


grcevski (approved):
LGTM!

PR #3072, opened 2026-08-18
prommgr: serve ports registered after the first StartHTTP call

### LINKED ISSUE #3069: Cannot reliably serve internal_metrics and prometheus_export at the same time: StartHTTP latches, so one port never listens
### Component(s)

`pkg/export/connector` (PrometheusManager), internal metrics, network flows Prometheus export

### What happened

A single OBI DaemonSet configured to export traces over OTLP plus network metrics on `prometheus_export.port` plus internal metrics on `internal_metrics.prometheus.port` only opens one of the two Prometheus ports. Which one it opens is decided per process start, so identically configured pods of the same DaemonSet end up split between working and broken, with no error in the logs.

On a 19-node DaemonSet, 4 pods opened both ports and 15 opened only the internal metrics port. Every pod runs the same image and the same config. On the affected pods the network metrics port is genuinely not listening:

```
dial tcp4 127.0.0.1:9090: connect: connection refused
```

while a working pod on the same DaemonSet serves 594 `beyla_network_flow_bytes_total` lines on that port. Affected pods log `opening prometheus scrape endpoint` for the internal metrics port only. The network flow metrics from those pods never reach the scraper and nothing reports a failure.

### Cause

`PrometheusManager.StartHTTP` latches on a single boolean and, on the first call, opens listeners only for what is registered at that moment (`pkg/export/connector/prommgr.go:71-95`):

```go
func (pm *PrometheusManager) StartHTTP(ctx context.Context) {
	pm.mt.Lock()
	defer pm.mt.Unlock()
	if pm.started {
		return
	}
	pm.started = true
	...
	for port, paths := range pm.registries {
		...
		listenAndServe(ctx, port, mux)
	}
}
```

Any `Register` arriving after that first call stays in `pm.registries` and is never served. No error, no warning, the port just never listens.

Two callers race for the first call:

- the internal metrics reporter registers and starts early in startup (`pkg/export/imetrics/iprom.go:187` and `:195`)
- the network flows exporter registers and starts late, when its pipeline is built (`pkg/export/prom/prom_net.go:145` and `:153`, the latter in a goroutine)

The outcome is asymmetric rather than a coin flip. If the internal metrics reporter reaches `StartHTTP` first, `prometheus_export.port` never opens. If the flows exporter's `Register` lands first, both ports open, because the loop covers every registered port.

### What did you expect to happen

Traces and Prometheus metrics should be usable together in one instance. Both configured Prometheus ports should listen, or startup should fail loudly.

### Steps to reproduce

Configure one instance with all three exports:

```yaml
otel_traces_export:
  endpoint: http://tempo:4318
internal_metrics:
  prometheus:
    port: 6060
prometheus_export:
  port: 9090
  features:
    - network
network:
  enable: true
```

Run it as a DaemonSet, then check which ports log `opening prometheus scrape endpoint`, or curl both ports on each pod. Larger node counts make it easier to see, since the split shows up across pods.

### Workaround

Remove `internal_metrics.prometheus.port` entirely. The reporter selection at `pkg/instrumenter/instrumenter.go:269-286` then falls through to the no-op branch, the flows exporter becomes the only `StartHTTP` caller, and its `Register` sits a few lines above its `StartHTTP` under the same mutex, so the port opens deterministically. The cost is losing internal metrics.

`internal_metrics.exporter: disabled` is not sufficient. The case at `:273` is `config.InternalMetrics.Exporter == imetrics.InternalMetricsExporterPrometheus || config.InternalMetrics.Prometheus.Port != 0`, so a non-zero port keeps the Prometheus reporter alive regardless of the exporter setting.

Pointing both exporters at the same port does not help either. `registries` is keyed on (port, path), so the late registration's path returns 404 instead of the connection being refused. Same outage, harder to spot.

### Suggested fix

Track which ports have already been served and let a later `StartHTTP` open listeners for newly registered ports, instead of latching on one boolean.

Alternatively make the contract explicit and fail or warn when `Register` is called after `started` is true. The doc comment on `StartHTTP` already states the ordering requirement, but nothing enforces it.

Happy to put up a PR for the first option if that direction looks right.

### Version

Observed with Grafana Beyla 3.31.0. Its `.obi-src` pins `7d018fe` in grafana's OBI fork, which diverges from upstream at merge base `071b718` (2026-08-02).

Confirmed present on upstream `main` at `4a6614e`. `prommgr.go`, `iprom.go`, `prom_net.go` and `instrumenter.go` are unchanged between `071b718` and `4a6614e`.

Kubernetes on AWS EKS, Bottlerocket.

### AI assistance disclosure

Per the project's Generative AI Policy: this report was drafted with AI assistance. The cited source locations were read and verified against the commits named above, and the pod counts and the refused connection are from a live cluster of mine.

mlushpenko:
## Summary

Fixes #3069.

`StartHTTP` latched on a single bool and opened listeners only for the registries present on the first call, so a `Register` arriving afterwards was kept in `pm.registries` but never served. The port silently never listened.

That happens with both `internal_metrics.prometheus.port` and `prometheus_export.port` set: the internal metrics reporter and the network flows exporter each register and start independently, and when the reporter gets there first the flows export port never opens. On a 19-node DaemonSet 15 pods lost that port while 4 kept it, same image and config.

This tracks served ports instead, so a later `StartHTTP` opens listeners for ports that are not served yet, and repeated calls stay harmless.

Added `prommgr_test.go` (the package had no tests). The first test fails on `main` and passes with the change:

```
# before
--- FAIL: TestStartHTTP_ServesPortRegisteredAfterFirstStart (1.05s)
--- PASS: TestStartHTTP_ServesAllPortsRegisteredBeforeStart
--- PASS: TestStartHTTP_IsIdempotentPerPort

# after
ok  	go.opentelemetry.io/obi/pkg/export/connector
```

Also ran `-race`, `go vet`, and `golangci-lint run pkg/export/connector/...` clean. After `make docker-generate`, the packages that touch this code pass too:

```
ok  go.opentelemetry.io/obi/pkg/export/connector
ok  go.opentelemetry.io/obi/pkg/export/prom
ok  go.opentelemetry.io/obi/pkg/export/imetrics
ok  go.opentelemetry.io/obi/pkg/instrumenter
```

## Validation

---

*Disclosure: I used an AI assistant while writing this. I have reviewed the change and the test, and the cluster numbers are from my own deployment.*

### DIFF
--- a/pkg/export/connector/prommgr.go
+++ b/pkg/export/connector/prommgr.go
@@ -29,8 +29,9 @@ func log() *slog.Logger {
 // PrometheusManager allows exporting metrics from different sources (instrumented metrics, internal metrics...)
 // sharing the same port and path, or using different ones, depending on the configuration provided by the registrars.
 type PrometheusManager struct {
-	mt      sync.Mutex
-	started bool
+	mt sync.Mutex
+	// ports that already have a listener, so a later StartHTTP only opens the new ones
+	served map[int]struct{}
 	// key 1: port. Key 2: path
 	registries maps.Map2[int, string, *prometheus.Registry]
 
@@ -66,19 +67,23 @@ func (pm *PrometheusManager) Register(port int, path string, collectors ...prome
 	reg.MustRegister(collectors...)
 }
 
-// StartHTTP serves metrics in background. Its invocation won't have effect if it has been invoked previously,
-// so invoke it only after you are sure that all the collectors have been registered via the Register method.
+// StartHTTP serves metrics in background. Safe to call repeatedly: each call opens listeners for
+// registered ports that are not served yet, so a late Register still gets its port served.
 func (pm *PrometheusManager) StartHTTP(ctx context.Context) {
 	pm.mt.Lock()
 	defer pm.mt.Unlock()
-	if pm.started {
-		return
+
+	if pm.served == nil {
+		pm.served = map[int]struct{}{}
 	}
-	pm.started = true
 
 	log := log()
-	// Creating a serve mux for each port
+	// Creating a serve mux for each port not served yet
 	for port, paths := range pm.registries {
+		if _, ok := pm.served[port]; ok {
+			continue
+		}
+		pm.served[port] = struct{}{}
 		mux := http.NewServeMux()
 		for path, registry := range paths {
 			log.With("port", port, "path", path).Info("opening prometheus scrape endpoint")
--- a/pkg/export/connector/prommgr.go
+++ b/pkg/export/connector/prommgr.go
@@ -30,8 +30,12 @@ func log() *slog.Logger {
 // sharing the same port and path, or using different ones, depending on the configuration provided by the registrars.
 type PrometheusManager struct {
 	mt sync.Mutex
-	// ports that already have a listener, so a later StartHTTP only opens the new ones
-	served map[int]struct{}
+	// mux of each port that already has a listener, so a path registered later can be
+	// added to the running mux instead of needing a second listener
+	muxes map[int]*http.ServeMux
+	// (port, path) pairs already wired into a mux. http.ServeMux panics on a duplicate
+	// pattern, so a repeated StartHTTP must skip them.
+	handled maps.Map2[int, string, struct{}]
 	// key 1: port. Key 2: path
 	registries maps.Map2[int, string, *prometheus.Registry]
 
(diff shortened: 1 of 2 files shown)


marctc (commented):
Hey, thanks for the fix! This fixes late registrations on new ports, but a new path on an already running port still returns 404. Can we cover that case too?

mlushpenko:
Good catch, thanks. You're right, skipping the whole port meant a path registered afterwards never reached the mux.

Pushed 831905e: the mux is now kept per port and a later `Register` adds its handler to the running one. `http.ServeMux` takes its own write lock so that's safe while it serves, but it panics on a duplicate pattern, so there's a `handled` set of (port, path) to keep repeated `StartHTTP` calls idempotent.

Added a test for it, which fails on the previous commit:

```
--- FAIL: TestStartHTTP_ServesPathRegisteredOnAlreadyServedPort
```

Four tests now cover late port, late path on a live port, everything registered up front, and repeated calls. `pkg/export/connector`, `pkg/export/prom`, `pkg/export/imetrics` and `pkg/instrumenter` all pass, lint clean.

marctc (approved):
LGTM thanks!

grcevski (approved):
LGTM!

PR #3076, opened 2026-08-18
fionread: disable CP on failed fixup attachment

mmat11:
## Summary

Currently, on failed fixup attachment for ioctl(FIONREAD), we just log an error and proceed. This PR disables context propagation as a safety measure in order to not break external apps

## Validation

### DIFF
--- a/pkg/internal/ebpf/tpinjector/fionread_check.go
+++ b/pkg/internal/ebpf/tpinjector/fionread_check.go
@@ -115,7 +115,7 @@ func sockhashFIONREADProbe(cookies *ebpf.Map) (bool, error) {
 
 func (p *Tracer) kernelBreaksFIONREAD() bool {
 	p.fionreadOnce.Do(func() {
-		broken, err := sockhashFIONREADProbe(nil)
+		broken, err := p.fionreadProbe(nil)
 		if err != nil {
 			// an inconclusive probe must not enable propagation on its own; the
 			// end-to-end check in sockhashSafe makes the final call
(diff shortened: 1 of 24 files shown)


NimrodAvni78 (approved):
LGMT!

pkg/internal/ebpf/tpinjector/fionread_check.go
```diff
@@ -151,19 +154,43 @@ func loadableFIONREADFixup() (*ebpf.CollectionSpec, error) {
...
+// kernel propagation stays off unless the compensation is verified end to end: losing
+// propagation is preferable to breaking the instrumented applications
+func (p *Tracer) sockhashSafe() bool {
+	p.sockhashOnce.Do(func() {
+		if !p.kernelBreaksFIONREAD() {
+			p.sockhashOK = true
+			return
+		}
+
+		if !p.fionreadFixupEnabled {
```
pinoOgni: If we are here it means that `LoadSpecs` could not load the fixup bundle. If "measured-broken": kernel bugged for real, propagation would break apps so it is correct to disable. If "assumed-broken": kernel may be healthy and we disable propagation anyway which is the safe, lossy choice.

pkg/internal/ebpf/tpinjector/fionread_check.go
```diff
@@ -151,19 +154,43 @@ func loadableFIONREADFixup() (*ebpf.CollectionSpec, error) {
...
-		p.log.Error("FIONREAD compensation is ineffective (attach failed or blocked?); " +
-			"applications sizing reads via FIONREAD (nginx, Java, .NET) may stall or " +
-			"truncate transfers; set context_propagation: disabled " +
-			"(OTEL_EBPF_BPF_CONTEXT_PROPAGATION=disabled) to avoid impact")
-		return
-	}
-	p.log.Info("FIONREAD compensation verified")
+// every bail-out has the same consequence, so each reason is reported with a common explanation
+func (p *Tracer) reportPropagationDisabled(reason string, args ...any) {
+	p.log.Error("context propagation is disabled: "+reason+". This kernel misreports "+
```
pinoOgni, line 159: When this function is called in this if
```
		if err != nil {
			p.reportPropagationDisabled("the BPF compensation cannot be verified", "error", err)
			return
		}
```
it means that the second probe could not run either and if the first probe errored too nothing was ever measured and the kernel may be healthy right? If yes, what about adding something like "..or could not be verified to correctly report" to the error log?

grcevski (approved):
LGTM!

mmat11:
tests are failing because some compose files are missing tracefs and context propagation is now correctly failing -- will add tracefs mount to all compose files

mifisignal:
There needs to be a doc update to https://opentelemetry.io/docs/zero-code/obi/security/ relating to the new /sys/kernel/tracing mount requirement.

PR #3082, opened 2026-08-18
Reassemble split Aerospike responses in kernel

pinoOgni:
## Summary

This PR adds large buffer to Aerospike and it fixes the Aerospike limitation where clients that read a response in multiple `recv()` calls (like the java client who reads 8-byte protocol header first and then the body) never exposed the response `result_code` so failed operations were reported as successful.

The solution consists to track the first response frame per connection and delays the span event until the frame is complete, shipping the bytes to userspace as app-layer large buffers. Clients that read responses in a single call are unaffected.

## Validation


internal/config/convert/import.go
```diff
@@ -1206,6 +1206,10 @@ func applyFullV2Instrumentation(cfg *obi.Config, instrumentation schema.Instrume
 	cfg.EBPF.BufferSizes.Kafka = instrumentation.Kafka.BufferSize
 	cfg.EBPF.KafkaTopicUUIDCacheSize = instrumentation.Kafka.TopicUUIDCacheSize
 
+	if instrumentation.Aerospike != nil {
```
mmat11, line 1209: question: why does this need a nil check?

pinoOgni: Basically `Aerospike` is a pointer field in the structure `instrumentation` and was defined as a pointer because it's the only section added after v2 documents already existed. A pointer can be nil so we need the guard.

mmat11 (approved):
lgtm

grcevski (approved):
LGTM!

PR #3084, opened 2026-08-18
generictracer: report an unobserved response as unknown, not as a 499 error

### LINKED ISSUE #3067: [BUG] CLIENT spans carry a synthesized `http.response.status_code = 499` for calls that returned 200
## Nature of the bug

OBI can sometimes emit an HTTP CLIENT span carrying `http.response.status_code =
499` and span status `Error` for a request that succeeded. The peer's own server
span for the same call reports `200`, and both applications' logs record
success. The span also reports `http.response.body.size = 0`, and its duration
is the socket's lifetime rather than the request's. No 499 is ever sent or
received on the wire: the value is synthesized in `force_finish_http()` when OBI
reaches `tcp_close` having never observed the response.

## Discussion

A trace span is a story about what happened. One such story is, "gateway sent a
POST to router. The router answered 499. The call failed, in 3.6 ms."

This framing is nonsensical. First, there's no such HTTP error as 499; it's an
nginx logging convention that indicates the client socket was shut down before
the respose could be returned. In reality, the call returned 200 after about a
second, and 3.6 ms is when OBI stopped watching rather than when the request
ended. 

What it should say instead is, "client sent an HTTP request at T. I observed the
request. I never observed the outcome, because the connection went away. Duration
was at least X; the true duration and result are unknown."

The current code can't express the latter story. Instrumentation is an observer,
and observers lose track when a buffer is unreadable or a socket is torn down
mid-flight. A model offering only "worked" or "failed" forces the observer to
pick one or the other in the face of ambiguity. Fortunately the OTel model provides 
a span status of `Unset` to mean that no explicit judgment was made. The 
problem is that OBI cannot reach it because `status == 0` is used to mean
"still reading."

Additionally, per [OTel HTTP
semconv](https://opentelemetry.io/docs/specs/semconv/http/http-spans/),
`http.response.status_code` is required "If and only if one was received/sent".
The bug manifests itself when OBI's 499 doesn't correspond to received status
code, and so populating the attribute when this happens is a spec violation.

## Symptom

A client sends a request, the peer answers 200, the client reads the whole
response and logs success. OBI emits a CLIENT span with
`http.response.status_code = 499`, `status = STATUS_CODE_ERROR`, and
`http.response.body.size = 0`. A second OBI instance watching the peer emits a
SERVER span reporting 200 for the same call.

The duration measures the socket's lifetime rather than the request's, running
from the request write to `bpf_ktime_get_ns()` at `tcp_close`: a median of
64.9 ms against a median actual call of 3.1 ms in one scenario, and 6 full
seconds when the connection sits in a pool that is never reaped.

## Suspected cause

`force_finish_http()` (`protocol_http.h:200-216`):

```c
    if (!high_request_volume) {
        if (!http_info_complete(info)) {
            info->resp_len = 0;
            info->end_monotime_ns = bpf_ktime_get_ns();
            info->status = 499;
        }
    }

    finish_http(info, pid_conn, current_key);
```

`finish_http()` emits only when `http_info_complete(info)` holds, and that
predicate is:

```c
return (info->start_monotime_ns != 0 && info->status != 0 && info->pid.host_pid != 0);
```

A request whose response was never observed has `status == 0`, so it is silently
dropped. `force_finish_http()` is reached from
`force_finish_possible_delayed_http_request()` on the socket-teardown path: the
connection is going away and everything known about that request is about to be
lost. Yet the fields set include the current time, `resp_len = 0` for the
response it never saw, and a non-zero `status` to get past the `status != 0`
gate.

The neighbouring helper confirms the overload:

```c
static __always_inline bool still_reading(http_info_t *info) {
    return info->status == 0 && info->start_monotime_ns != 0;
}
```

The status field is thus serving as both a state-machine flag ("have I seen the
response yet") and the reported HTTP status. Under that design, "finished,
outcome unknown" is unrepresentable: `0` is taken by in-flight, and every other
value is a status code.

We suspect the intent was not to report a client abort. It was to avoid losing
the span, and 499 is the least-wrong non-zero value when the meaning is "the
connection went away before I saw a response" — `0` means in-flight, and any 5xx
would falsely blame the server. `high_request_volume` gating the whole branch
fits the same reading: with that option set the fabrication is skipped and the
span is dropped instead. That is a deliberate trade-off between losing a record
and inventing a status, made inside a struct that offered only those two
options.

Nothing in #767 mentions status semantics, so whether
the author foresaw that 499 would surface as an OTel `Error` on a CLIENT span is
unknown.

## Reproduction

https://github.com/mifisignal/obi-phantom-499-repro

### How it is triggered

The following two conditions must coincide:

1. OBI does not observe the response, so the `http_info_t` keeps
   `status == 0`. The reproduction induces this by re-segmenting the response so
   its first TCP segment is one byte, with the rest 2 ms later: `tcp_cleanup_rbuf`
   returns early on `copied_len <= 1`, and the remainder arrives in a read that
   does not begin with a status line.
2. `tcp_close` fires while that incomplete entry is still in `ongoing_http`.

Condition 2 publishes the error, and the rate precisely tracks the close rate.
Holding the segmentation constant and varying only the client's connection pool:

| App's outbound pool | Phantom 499s |
|---|---|
| connection per request | 8,433 of 8,435 (99.98%) |
| keep-alive, 250 ms idle reaping | 216 of 9,422 (2.29%) |
| keep-alive, never reaped | 2 of 9,475 (0.02%) — the pair closed at shutdown |

We know that a missed response plus a`tcp_close` deterministically produces the
phantom 499, and that the defect is entirely absent when the response is
observed. We don't know, though, what exactly caused the missed responses that
are necessary to make the bug appear. 

## Evidence

The same call from both sides. `check.py` pairs each phantom span with the peer
SERVER span whose whole lifetime falls inside it: 8,433 of 8,433 matched.

```
CLIENT (app)   GET  server.address=relay:9100
               http.response.status_code=499  http.response.body.size=0
               status=STATUS_CODE_ERROR       duration=18.244ms

SERVER (peer)  GET  url.path=/r  client.address=172.19.0.4
               http.response.status_code=200  http.response.body.size=227
               status=unset                   duration=4.660ms
```

Both applications' access logs, same traffic:

```
{"role":"app", "msg":"call","id":1,"peer_status":200,"peer_bytes":64,"duration_ms":30.33}
{"role":"peer","msg":"request","id":1,"status":200,"bytes":64,"aborted":false,"response_completed":true}
```

Aggregates for that run: 8,435 calls made, 8,435 returned 200, 0 failed; the peer
served 8,435 requests with 0 aborted or partial; all 8,435 peer SERVER spans
report 200.

## Impact

Healthy requests are published as errors. Error-based trace search returns them,
tail-sampling policies keyed on span status retain them preferentially, and the
service graph draws a failing edge between two healthy services. The inflated
duration compounds it, pulling the same spans into latency-based sampling and
alerting.

RED metrics are affected as well. These spans carry no `span.metrics.skip`
attribute — verified on the exported spans rather than assumed — and with
`OTEL_EBPF_METRICS_FEATURES=application` the collector receives
`http.client.request.duration`, `http.client.request.body.size` and
`http.client.response.body.size` series labelled
`http.response.status_code=499` alongside the 200 series.

Because the rate follows connection churn rather than request failures, the
result is a steady low-percentage error floor on a healthy service.

## Version / environment

```
OBI     otel/ebpf-instrument:main
        sha256:6f38eb3e3c5d556773002855a660e0e96b787f630bd3e6886aabfb3e36c574db
        Version=7b96052 Revision=7b96052
host    EC2 c7g, Ubuntu 24.04, aarch64 (Graviton)
kernel  7.0.0-1010-aws
BTF     /sys/kernel/btf/vmlinux present
docker  29.7.2
```

Generic tracer, plaintext HTTP/1.1, Node workloads on both ends. The code quoted
above is unchanged on upstream `main` at commit `4a6614e` (2026-08-17), verified
against a fresh clone rather than the tested image.

## AI assistance disclosure

Per the project's Generative AI Policy: this report was drafted with AI
assistance. The reproduction, the parameter sweep behind the trigger analysis,
and every span, log line and count quoted here were produced by running the
attached code on the system described above, and were reviewed by a human before
submission.

mifisignal:
Fixes #3067.

`force_finish_http()` sets `status = 499` when `tcp_close` arrives with a
request still in `ongoing_http` and where no response was read. Userspace
publishes that as `http.response.status_code` on a CLIENT span, and per OTel
HTTP semconv, a 4xx on a CLIENT span should map to `Error`. This bug causes
calls that actually returned 200 to be reported as failures. 499 is nginx's
logging convention indicating that the client closed the connection prematurely.
It's not returned to a client (the client is no longer a client!), so OBI
must not synthesize one as though it actually happened.

The underlying issue is that `status` carries two meanings at once: `0` means
"no response yet", and anything else is the observed HTTP status code. There is
no value for "finished, but nothing was received", so `force_finish_http()`
substituted 499 — which is then published as `http.response.status_code`, an
attribute semconv reserves for a status that was actually received or sent. It
had no better option: `finish_http()` emits only when `status != 0`, so a record
with no observed response would otherwise be dropped in silence.

## Changes

**BPF.** `http_info_t` gains `unobserved_response`, consuming one of the two
bytes of the struct padding. `force_finish_http()` sets it and leaves `status`
at 0. Emit paths now gate on `http_info_emittable()` to avoid dropping the spans
silently, which is the regression #767 added `force_finish_http()` to fix. The
span status is decided from the marker rather than from `status == 0`, since
`HTTPSpanStatusCode()` maps a zero status to `Error` and would otherwise
relocate the same false error from the status code onto the span.
`http_info_complete()` keeps its prior meaning, because
`should_ignore_unreadable()` and the incomplete-server-trace cleanup use it to
ask whether the response arrived, not whether the record can be published.

To determine whether the peer sent anything back, it now saves the kernel's byte
counter when the request starts and reads it again at force-finish. Client
records use `bytes_received`; server records use `bytes_sent`, because a
server's response goes out rather than in. The comparison is `> snapshot + 1`,
not `> snapshot`, since a FIN consumes a sequence number.

**Trace export.** That counter splits a force-finished request into two cases: the
peer sent something back and OBI failed to parse it, or nothing came back at all.
Neither reports a status code, since nothing was received; and neither reports an
error, since OBI has no evidence of failure. They differ in whether the duration is
a measurement:

- If the peer sent bytes OBI did not parse, the outcome is unknown: status `Unset`,
  `obi.http.response.observed=false`. The response arrived at an unknown moment,
  so the span contributes no duration. The call is still counted in
  `traces_service_graph_request_total`, and never in `..._failed_total`.
- If nothing came back, the request ended when the local process closed the
  socket. In this case, the duration is valid, and the span participates in
  metrics normally. The status is still `Unset`, though, because OBI cannot
  distinguish a peer that failed from a client that stopped waiting.

## Resource impact

Adding 8 bytes to `http_info_t` consumes some extra ring buffer space and
memory allocated for the `ongoing_http` map:

| | before | after |
|---|---|---|
| `sizeof(http_info_t)` | 416 B | **424 B** |
| ring-buffer per record (incl. header) | 424 B | 432 B (**+1.89%**) |
| `ongoing_http` (LRU hash, 30 000 preallocated) | — | **+234 KiB** |
| at 1k / 10k requests per second | — | +7.8 / +78 KiB/s |

## Test measurements

Reproduction: https://github.com/mifisignal/obi-phantom-499-repro — `DURATION_S=40
CONCURRENCY=2`, stock and patched images both built on the host from base
`4a6614e6`, measured at branch HEAD `a5a048eb`. Cells read `stock → fix`.

| scenario | app calls returning 200 | CLIENT spans | spans with `499` |
|---|---|---|---|
| C0 (control, response observed) | 10555 → 10659 | 10555 → 10659 | 0 → 0 |
| S1 (connection per request) | 8160 → 8237 | 8160 → 8237 | 8159 → **0** |
| S2 (keep-alive, 250 ms reaping) | 9356 → 9365 | 233 → 227 | 232 → **0** |
| S3 (keep-alive, never reaped) | 9359 → 9393 | 3 → 3 | 2 → **0** |

Every call in every scenario returned 200 at the application. The spans that
carried `499`/`STATUS_CODE_ERROR` on the stock build carry no
`http.response.status_code` at all on the fix build, with span status unset, in
exactly the counts stock reported as 499 — and the control's spans still carry
`http.response.status_code=200` on both builds. Client spans survive at the same
rate per call (C0 1.000 on both; S1 99.99% stock vs 99.98% fix).

Note that S2 and S3 emit a CLIENT span for only a small fraction of calls on
both builds. That is a separate pre-existing defect: when a new request arrives on
a reused connection, `get_or_set_http_info()` hands the previous request's
still-open record to `finish_http()` — which drops it for the same `status != 0`
reason — then overwrites it. This change does not address that; it is
characterized separately at
https://github.com/mifisignal/obi-connection-reuse-span-loss-repro.

## Testing

`bpf/tests/test_http_unobserved_response.c` covers the states the predicates
distinguish, the FIN tolerance, reused connections, and that the marker alone
does not rescue a record missing the fields a span needs. `bpf_ssl_read_guard`
asserts the TLS read and write paths pass no socket, pinning the limitation
described under "What this does not settle" below. Go tests cover the transform, the
status mapping, the trace attributes and both metrics paths, each paired with a
control asserting an ordinary request is unchanged on the same axis. The
transform test asserts the span is emitted.

`TestServiceGraphMetrics_UnknownOutcomeIsCountedWithoutADuration` tests that an
unknown-outcome span increments the service-graph request counter and does not
add a latency sample.

`TestSuite_UnobservedResponse` tests that a call whose response OBI never parses
is still reported, but has no `http.response.status_code`, is not an error,
and carries the observation marker, while ensuring that an ordinary observed
call is unaffected.

## Considerations

A span with no observed response still carries `http.response.status_code` in
the RED series, because `span_getters.go` derives that label from `Span.Status`
unconditionally — so a call in the second case above is counted with a status of
`0`. Traces omit the attribute and metrics label it zero. That is less misleading
than a 499, but it is new behavior.

On TLS uprobe and unix-socket paths there is no `struct sock *`, so the snapshot
stays zero and a request reads as "nothing came back". On a reused keep-alive
connection whose earlier request was answered, that misreads as the peer having
answered. It withholds a duration rather than inventing one, which is the safer
direction, but it argues for plumbing the socket through those paths too.

## AI assistance disclosure

Per the project's Generative AI Policy: this change was developed with AI
assistance. The reproduction, the measurements above and the source analysis were
produced against a running system on a dedicated host, and reviewed by a human
before submission.

### DIFF
--- a/internal/test/integration/configs/obi-config-response-observation.yml
+++ b/internal/test/integration/configs/obi-config-response-observation.yml
@@ -10,6 +10,16 @@ routes:
   unmatched: path
 otel_traces_export:
   endpoint: http://jaeger:4318
+prometheus_export:
+  port: 8999
+metrics:
+  features:
+    - application
+    # The service graph is emitted from inside the span-metrics path, so that
+    # feature has to be on for the graph series to exist at all.
+    - application_span_otel
+    - application_service_graph
+    - application_host
 discovery:
   # The workload is a Go binary, and the code under test is in the generic
   # tracer, which is where every non-Go workload lands.
--- a/internal/test/integration/configs/obi-config-response-observation.yml
+++ b/internal/test/integration/configs/obi-config-response-observation.yml
@@ -1,0 +1,19 @@
+trace_printer: text
+log_level: debug
+ebpf:
+  # The pooled calls are made after the request that triggered them has already
+  # answered, so they arrive conditionally parented and are held until their parent
+  # settles or this bound elapses. The default of five minutes outlives the suite,
+  # which would leave the spans unexported rather than late.
+  max_transaction_time: 15s
+routes:
+  unmatched: path
+otel_traces_export:
+  endpoint: http://jaeger:4318
+discovery:
+  # The workload is a Go binary, and the code under test is in the generic
+  # tracer, which is where every non-Go workload lands.
+  skip_go_specific_tracers: true
+  services:
+    - name: responseobservationclient
+      exe_path: responseobservationclient
--- a/pkg/ebpf/common/http_transform.go
+++ b/pkg/ebpf/common/http_transform.go
@@ -103,28 +103,31 @@ func httpInfoToSpanLegacy(info *HTTPInfo) request.Span {
 	return span
 }
 
-// The kernel's enum http_response_observation. Nonzero means the record was
-// force-finished at socket teardown with no response read, so it carries no status.
+// The kernel's enum http_response_observation. Nonzero means no probe read the
+// response, so the record carries no status.
 const (
 	bpfResponseParsed   = 0
 	bpfResponseReceived = 1
 	bpfResponseSilent   = 2
+	bpfResponseUnread   = 3
 )
 
-// markResponseObservation copies the kernel's observation onto the span. Both
-// nonzero values mean no response was read, so Status is cleared in either
-// case.
+// markResponseObservation copies the kernel's observation onto the span. Every
+// nonzero value means no response was read, so Status is cleared in each case.
 //
-// They differ in whether the end timestamp is usable. ResponseReceived took it
-// when instrumentation stopped watching, which can be long after the response
-// arrived, so the duration is discarded. ResponseSilent took it from the close
-// that ended the request, so the duration is preserved.
+// They differ in whether the end timestamp is usable. ResponseReceived took it when
+// instrumentation stopped watching, which can be long after the response arrived, so
+// the duration is discarded. ResponseSilent took it from the close that ended the
+// request and ResponseUnread from the response's own bytes, so both durations are
+// preserved.
 func markResponseObservation(span *request.Span, event *BPFHTTPInfo) {
 	switch event.ResponseObservation {
 	case bpfResponseParsed:
 		return
 	case bpfResponseSilent:
 		span.ResponseObservation = request.ResponseSilent
+	case bpfResponseUnread:
+		span.ResponseObservation = request.ResponseUnread
 	// An unrecognized value withholds the duration, which asserts less than publishing
 	// one.
 	case bpfResponseReceived:
--- a/pkg/ebpf/common/http_transform.go
+++ b/pkg/ebpf/common/http_transform.go
@@ -69,7 +69,7 @@ func httpInfoToSpanLegacy(info *HTTPInfo) request.Span {
 		scheme = "https"
 	}
 
-	return request.Span{
+	span := request.Span{
 		Type:              request.EventType(info.Type),
 		Method:            info.Method,
 		Path:              removeQuery(info.URL),
@@ -97,6 +97,44 @@ func httpInfoToSpanLegacy(info *HTTPInfo) request.Span {
 		Statement:    scheme + request.SchemeHostSeparator + info.HeaderHost,
 		ProtoVersion: info.ProtoVersion,
 	}
+
+	markResponseObservation(&span, &info.BPFHTTPInfo)
+
+	return span
+}
+
+// The kernel's enum http_response_observation. Nonzero means the record was
+// force-finished at socket teardown with no response read, so it carries no status.
+const (
+	bpfResponseParsed   = 0
+	bpfResponseReceived = 1
+	bpfResponseSilent   = 2
+)
+
+// markResponseObservation copies the kernel's observation onto the span. Both
+// nonzero values mean no response was read, so Status is cleared in either
+// case.
+//
+// They differ in whether the end timestamp is usable. ResponseReceived took it
+// when instrumentation stopped watching, which can be long after the response
+// arrived, so the duration is discarded. ResponseSilent took it from the close
+// that ended the request, so the duration is preserved.
+func markResponseObservation(span *request.Span, event *BPFHTTPInfo) {
+	switch event.ResponseObservation {
+	case bpfResponseParsed:
+		return
+	case bpfResponseSilent:
+		span.ResponseObservation = request.ResponseSilent
+	// An unrecognized value withholds the duration, which asserts less than publishing
+	// one.
+	case bpfResponseReceived:
+		fallthrough
+	default:
+		span.ResponseObservation = request.ResponseReceived
+		request.SetIgnoreDurations(span)
+	}
+
+	span.Status = 0
 }
 
 func httpRequestResponseToSpan(parseCtx *EBPFParseContext, event *BPFHTTPInfo, req *http.Request, resp *http.Response) request.Span {
@@ -162,6 +200,8 @@ func httpRequestResponseToSpan(parseCtx *EBPFParseContext, event *BPFHTTPInfo, r
 		},
 		Statement: scheme + request.SchemeHostSeparator + headerHost,
 	}
+
+	markResponseObservation(&httpSpan, event)
 
 	return postProcessHTTPSpan(parseCtx, &httpSpan, req, resp)
 }
--- a/pkg/export/otel/metrics.go
+++ b/pkg/export/otel/metrics.go
@@ -939,225 +939,296 @@ func (r *Metrics) record(span *request.Span, mr *MetricsReporter) {
 
 	ctx := trace.ContextWithSpanContext(r.ctx, trace.SpanContext{}.WithTraceID(span.TraceID).WithSpanID(span.SpanID).WithTraceFlags(trace.TraceFlags(span.TraceFlags)))
 
-	// Data point timestamps are the collection time, not the span end time: the OTel
-	// metrics API takes no per-measurement timestamp. Span-accurate timing lives in
-	// traces and in the exemplars attached through ctx.
+	// A record finished without its response ends when something other than the
+	// response ended it, so its duration describes more than the request it names.
+	// Every instrument reached through recordDuration stands on that duration; the
+	// body sizes do not, and are recorded on their own terms.
+	measured := !request.IgnoreDurations(span)
+
 	if otelMetricsAccepted(span) {
-		switch span.Type {
-		case request.EventTypeHTTP:
-			// JSON-RPC over HTTP gets recorded as RPC server metrics
-			if span.SubType == request.HTTPSubtypeJSONRPC && mr.is.GRPCEnabled() {
-				grpcDuration, attrs := r.grpcDuration.ForRecord(span)
-				grpcDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			} else if mr.is.HTTPEnabled() {
-				httpDuration, attrs := r.httpDuration.ForRecord(span)
-				httpDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
+		if measured {
+			r.recordDuration(span, mr, ctx, duration)
+		}
 
-				httpRequestSize, attrs := r.httpRequestSize.ForRecord(span)
-				httpRequestSize.Record(ctx, float64(span.RequestBodyLength()), instrument.WithAttributeSet(attrs))
+		r.recordBodySizes(span, mr, ctx)
+	}
 
-				httpResponseSize, attrs := r.httpResponseSize.ForRecord(span)
-				httpResponseSize.Record(ctx, float64(span.ResponseBodyLength()), instrument.WithAttributeSet(attrs))
-			}
-		case request.EventTypeGRPC:
-			if mr.is.GRPCEnabled() {
-				grpcDuration, attrs := r.grpcDuration.ForRecord(span)
-				grpcDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			}
-		case request.EventTypeGRPCClient:
-			if mr.is.GRPCEnabled() {
-				grpcClientDuration, attrs := r.grpcClientDuration.ForRecord(span)
-				grpcClientDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			}
-		case request.EventTypeSunRPCClient:
-			if mr.is.SunRPCEnabled() {
-				grpcClientDuration, attrs := r.grpcClientDuration.ForRecord(span)
-				grpcClientDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			}
-		case request.EventTypeSunRPCServer:
-			if mr.is.SunRPCEnabled() {
-				grpcDuration, attrs := r.grpcDuration.ForRecord(span)
-				grpcDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			}
-		case request.EventTypeHTTPClient:
-			// HTTP client subtypes that are database calls get recorded as db client metrics
-			if mr.is.DBEnabled() && (span.SubType == request.HTTPSubtypeSQLPP || span.SubType == request.HTTPSubtypeElasticsearch) {
-				dbClientDuration, attrs := r.dbClientDuration.ForRecord(span)
-				dbClientDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			} else if span.SubType == request.HTTPSubtypeJSONRPC && mr.is.GRPCEnabled() {
-				grpcClientDuration, attrs := r.grpcClientDuration.ForRecord(span)
-				grpcClientDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			} else if span.SubType == request.HTTPSubtypeAWSS3 && mr.rpcClientRecorded() {
-				grpcClientDuration, attrs := r.grpcClientDuration.ForRecord(span)
-				grpcClientDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			} else if span.SubType == request.HTTPSubtypeAWSSQS && request.IsSQSMessagingClientOperation(span) && mr.msgPublishRecorded() {
-				msgPublishDuration, attrs := r.msgPublishDuration.ForRecord(span)
-				msgPublishDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			} else if mr.is.GenAIEnabled() && request.IsGenAISubtype(span.SubType) {
-				genAIClientDuration, attrs := r.genAIClientDuration.ForRecord(span)
-				genAIClientDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-				if tokens, reported := span.GenAIInputTokenCount(); reported {
-					genAIInputTokenUsage, attrs := r.genAIInputTokenUsage.ForRecord(span)
-					genAIInputTokenUsage.Record(ctx, float64(tokens), instrument.WithAttributeSet(attrs))
-				}
-				if tokens, reported := span.GenAIOutputTokenCount(); reported {
-					genAIOutputTokenUsage, attrs := r.genAIOutputTokenUsage.ForRecord(span)
-					genAIOutputTokenUsage.Record(ctx, float64(tokens), instrument.WithAttributeSet(attrs))
-				}
-			} else if mr.is.HTTPEnabled() {
-				httpClientDuration, attrs := r.httpClientDuration.ForRecord(span)
-				httpClientDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-				httpClientRequestSize, attrs := r.httpClientRequestSize.ForRecord(span)
-				httpClientRequestSize.Record(ctx, float64(span.RequestBodyLength()), instrument.WithAttributeSet(attrs))
-				httpClientResponseSize, attrs := r.httpClientResponseSize.ForRecord(span)
-				httpClientResponseSize.Record(ctx, float64(span.ResponseBodyLength()), instrument.WithAttributeSet(attrs))
-			}
-		case request.EventTypeRedisClient:
-			if mr.is.RedisEnabled() {
-				dbClientDuration, attrs := r.dbClientDuration.ForRecord(span)
-				dbClientDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			}
-		case request.EventTypeRedisServer:
-			if mr.is.RedisEnabled() {
-				dbServerDuration, attrs := r.dbServerDuration.ForRecord(span)
-				dbServerDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			}
-		case request.EventTypeSQLClient:
-			if mr.is.SQLEnabled() {
-				dbClientDuration, attrs := r.dbClientDuration.ForRecord(span)
-				dbClientDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			}
-		case request.EventTypeSQLServer:
-			if mr.is.SQLEnabled() {
-				dbServerDuration, attrs := r.dbServerDuration.ForRecord(span)
-				dbServerDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			}
-		case request.EventTypeMongoClient:
-			if mr.is.MongoEnabled() {
-				dbClientDuration, attrs := r.dbClientDuration.ForRecord(span)
-				dbClientDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			}
-		case request.EventTypeCouchbaseClient:
-			if mr.is.CouchbaseEnabled() {
-				dbClientDuration, attrs := r.dbClientDuration.ForRecord(span)
-				dbClientDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			}
-		case request.EventTypeMemcachedClient:
-			if mr.is.MemcachedEnabled() {
-				dbClientDuration, attrs := r.dbClientDuration.ForRecord(span)
-				dbClientDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			}
-		case request.EventTypeMemcachedServer:
-			if mr.is.MemcachedEnabled() {
-				dbServerDuration, attrs := r.dbServerDuration.ForRecord(span)
-				dbServerDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			}
-		case request.EventTypeAerospikeClient:
-			if mr.is.AerospikeEnabled() {
-				dbClientDuration, attrs := r.dbClientDuration.ForRecord(span)
-				dbClientDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-			}
-		case request.EventTypeKafkaClient, request.EventTypeKafkaServer:
-			if mr.is.KafkaEnabled() {
-				switch request.MessagingOperationTypeOf(span.Method) {
-				case request.MessagingSend:
-					msgPublishDuration, attrs := r.msgPublishDuration.ForRecord(span)
-					msgPublishDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-				case request.MessagingProcess:
-					msgProcessDuration, attrs := r.msgProcessDuration.ForRecord(span)
-					msgProcessDuration.Record(ctx, duration, instrument.WithAttributeSet(attrs))
-				}
-			}
-		case request.EventTypeMQTTClient, request.EventTypeMQTTServer:
-			if mr.is.MQTTEnabled() {
-				switch request.MessagingOperationTypeOf(span.Method) {
-				case request.MessagingSend:
-					msgPublishDuration, attrs := r.msgPublishDuration.ForRecord(span)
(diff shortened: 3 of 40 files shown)


pkg/export/otel/metrics_svc_graph.go
```diff
@@ -353,6 +353,13 @@ func (r *SvcGraphMetrics) record(span *request.Span, mr *SvcGraphMetricsReporter
 	t := span.Timings()
 	duration := t.End.Sub(t.RequestStart).Seconds()
 
+	// The request counter is its own instrument, so a span whose duration was never
```
grcevski, line 356: While we don't want to store these unmeasurable spans in the metrics, I think they might be useful for the service graphs, right? It would tell us that there was something happening from this side to the other, the connectivity is there, there's just no duration.

grcevski:
This is really a great solution! I like this a lot more than the 499 we were putting before.

mifisignal:
FYI - this PR isn't ready for review yet. The comments aren't up to my standards, and there are overlapping commits. I'll follow up when I feel good about it.

pkg/export/otel/metrics_svc_graph.go
mifisignal: For a client span, `serviceGraphTotal` is additionally gated on `ClientSpanToUninstrumentedService()` (which is preexisting behavior). When the peer is instrumented, its own server span draws the edge. So server spans always count; client spans count when the peer isn't instrumented. In short, there should be no change to the service graph.

mifisignal: In the current shape, the edge is untouched. `traces_service_graph_request_total` is recorded exactly as before; the `ignoreDurations` guard wraps only `serviceGraphClient` and `serviceGraphServer`. Server spans count the edge unconditionally, and client spans count it when the peer isn't instrumented, both unchanged from today. Connectivity is preserved, with no duration.

Two things about the graph do change:
  - `..._failed_total` no longer counts these calls. Today the fabricated 499 is a 4xx on a CLIENT span, which HTTPSpanStatusCode maps to Error, so an edge whose responses all go unparsed is drawn as failing while the peer's own SERVER span reports 200 for the same calls. With the status Unset they count as traffic on the edge and never against it.
  - The two latency histograms lose the samples whose duration was never measured. Those were the socket's lifetime rather than the request's, so they were inaccurate.

mifisignal:
Ready for review @grcevski

mifisignal:
Added commits to address a second failure mode that this PR didn't originally cover. 

When the connection is reused instead of closed, the record keeps `status == 0`,
the next request on that connection overwrites it, and the call is never
reported at all. It yield one span per *connection* instead of one per
*request*, and none while the connection stays open. Server spans go the same
way, which is worse, since service maps and RED metrics are built from them.

The marker introduced here helps resolve the problem. A buffer arriving in the response
direction with no status line in it is now recorded as the response, unparsed, ending the
record at its own arrival time instead of being treated as more of the earlier request. The
record therefore survives the next request on the connection, reports no status, and
carries a duration measured from the response's own bytes rather than from a teardown. A
request still in flight when the next one takes the connection is reported too, without a
duration it cannot vouch for. The duplicate check no longer swallows the next call on a
hot pooled connection.

Two fixes to the integration harness I added earlier in this PR are in there as well:
several services declared the same build for one image tag, which races and, on newer
Compose, makes a service that only references the tag try to pull it; and the Jaeger query
passed no limit, so it silently capped at 20 traces.

Reproduction: https://github.com/mifisignal/obi-connection-reuse-span-loss-repro

grcevski:
It seems that the  TestSuite_ResponseObservation test is still hitting some issues.

pkg/export/otel/metrics_svc_graph.go
grcevski: I think this change might actually be affecting the OATS http test.

mifisignal:
@grcevski Rebased onto main. Two things came out of it:

First, I found a link failure in the BPF unit tests. This PR adds a `struct sock
*` argument to `handle_buf_with_connection`, so the test stub in
`bpf/tests/generictracer/k_tracer_defs.h` records it in a `test_last_sock`
symbol that each test file defines. The new `bpf_tls_prefix_bind.c` test from
#3023 landed in between and defines the other `test_last_*` symbols but not that
one, so it no longer linked. Added the missing definition.

Second, I found a weaver violation in the OATS http job. This PR emits a new
span attribute, `obi.http.response.observed`, when no response was seen. Semconv
only allows `http.response.status_code` if a response was received, so the
attribute marks that the status is missing on purpose. It was never declared in
`schemas/obi/`, which is where OBI declares everything it emits, so live-check
reported it as an unknown key and failed the job with 126 violations. Declared
it in `registry.obi.traces` as a development-stability boolean.

I checked the fix by running live-check against the registry with a sample that
carried both the new attribute and a made-up key.

internal/test/integration/configs/obi-config-response-observation.yml
```diff
@@ -0,0 +1,13 @@
+trace_printer: text
```
MrAlias, line 1: The suite inherits the five-minute `max_transaction_time`, but this assertion waits only 60 seconds. Generic-tracer calls marked `ParentConditional` remain in `SettleConditionalParents` past the assertion window; the current CI retries captured all 60 records before that stage but exposed only 12–33 to Jaeger. Could we set a suite-specific transaction bound below `testTimeout`, or otherwise isolate the pooled worker from conditional parents? The expected span count should remain unchanged.

pkg/export/otel/metrics.go
```diff
@@ -1349,6 +1349,12 @@ func (mr *MetricsReporter) onSpan(spans []request.Span) {
 		if !s.Service.Features.AppOrSpan() || request.IgnoreMetrics(s) {
 			continue
 		}
+		// The span's duration feeds every instrument this path reaches, so it is
+		// dropped whole. The service graph counts it instead; its request counter
+		// stands on its own.
+		if request.IgnoreDurations(s) {
```
MrAlias: `request.IgnoreDurations` returns before `traces.host.info` is recorded. A service with `FeatureApplicationHost` and only `ResponseReceived` traffic therefore loses the host-info gauge, even though that gauge does not use span duration and traces are still exported. Could we record host info before this check and add the unmeasured-span case to `TestAppMetrics_TracesHostInfo`?

MrAlias (commented):
The response-observation path needs two lifecycle boundaries tightened: captured spans must reach the integration assertion within the suite window, and duration suppression must not discard unrelated host metadata.

mifisignal:
@MrAlias Thanks for the feedback. Some more things to know:

The suite config now sets `ebpf.max_transaction_time: 15s`, below `testTimeout`. The expected span count is unchanged.

**Host info.** The gauge is now recorded before any duration gating, and `TestAppMetrics_TracesHostInfoUnmeasuredSpans` covers it.
 
**Body sizes.** The same `continue` also skipped `reporter.record`, which records the request and response size histograms. Those are now separated. The request size is published even when the duration is withheld: the request was sent, and its length was read from it. The response size is not. An unmeasured record carries a zeroed length, so publishing it would report an empty response for a call whose response was never seen, which is the same fabrication as the 499. Instead of guarding every duration site, we moved the sizes out of the duration switch in both exporters. The span-metrics size counters stay with the calls counter, since that family is read together.
 
**New coverage.** The suite never produced `http_response_received`, which is the only state that sets IgnoreDurations — the relay case is unread and the resetter is silent. So nothing exercised any of this end to end. Added a call the workload bandons before reading its answer, and added Prometheus assertions: 1/the service graph counts the edge but publishes no client latency; 2/the RED duration series is absent for that port and present for the relay port; and 3/the request size is published for it and the response size is not.

pkg/export/otel/metrics.go
mifisignal: Good catch. The gauge doesn't use duration and it's independent of the per-service reporter, so the `AppHost()` record now happens before the `IgnoreDurations` guard, and `TestAppMetrics_TracesHostInfo_UnmeasuredSpan` pins it with a  `SetIgnoreDurations` span. The Prometheus `traces_host_info` is process-event-driven and was never affected; this was OTel-only.

internal/test/integration/configs/obi-config-response-observation.yml
mifisignal: Went with the suite-specific bound. The suite's config now sets `ebpf.max_transaction_time: 15s`, against the 60s `testTimeout`, so a conditionally-parented pooled call whose parent never arrives is released roughly 15s after it's held rather than sitting for the default five minutes. The pooled calls also run from a background goroutine after the triggering request has answered, and each run closes idle connections at its tail so the last record isn't waiting on a sweep. Expected span count is unchanged at 60 per role.

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -544,20 +544,30 @@ func messagingOperationAttrs(method string) []attribute.KeyValue {
...
 }
 
+// appendHTTPResponseStatus reports the status only when one was seen: semconv requires
+// http.response.status_code "if and only if one was received/sent".
+func appendHTTPResponseStatus(attrs []attribute.KeyValue, span *request.Span) []attribute.KeyValue {
+	if span.ResponseObservation == request.ResponseParsed {
+		return append(attrs, request.HTTPResponseStatusCode(span.Status))
+	}
+
+	return append(attrs, attribute.Bool(string(attr.OBIHTTPResponseObserved), false))
```
MrAlias: This attribute is emitted here but is missing from `schemas/obi/groups/traces.yaml`, so Weaver will reject these spans. Please add this under `registry.obi.traces.attributes`:

```yaml
      - id: obi.http.response.observed
        type: boolean
        stability: development
        brief: >
          Set to false when OBI did not observe the HTTP response, and absent otherwise.
        examples: [false]
```

Please run the Weaver live check after adding it.

pkg/ebpf/common/http_transform.go
```diff
@@ -160,6 +201,8 @@ func httpRequestResponseToSpan(parseCtx *EBPFParseContext, event *BPFHTTPInfo, r
 		Statement: scheme + request.SchemeHostSeparator + headerHost,
 	}
 
+	markResponseObservation(&httpSpan, event)
```
MrAlias: By the time this line runs, `http.ReadResponse` has succeeded and `httpSpan.Status` is userspace-parsed. When the parsed response has a positive `Content-Length`, `ResponseLength` is also userspace-derived rather than taken from `event.RespLen`. Calling `markResponseObservation` clears the parsed status based on the earlier kernel classification. Keep the parsed outcome; only preserve the unreliable-duration flag:

```suggestion
if event.ResponseObservation == bpfResponseReceived {
	request.SetIgnoreDurations(&httpSpan)
}
```

Make the test prove which source won: parse a response with `Content-Length: 17`, set `event.RespLen` to a different nonzero value, then expect `ResponseParsed`, status 200, response length 17, and an ignored duration.

pkg/export/otel/metrics.go
```diff
@@ -928,218 +928,291 @@ func (r *Metrics) record(span *request.Span, mr *MetricsReporter) {
...
+	default:
+		return
+	}
 
-				gblock, attrs := r.gpuKernelBlockSize.ForRecord(span)
-				gblock.Record(ctx, float64(span.SubType), instrument.WithAttributeSet(attrs))
+	size, attrs := requestSize.ForRecord(span)
+	size.Record(ctx, float64(span.RequestBodyLength()), instrument.WithAttributeSet(attrs))
+
+	if request.IgnoreDurations(span) {
```
MrAlias: `IgnoreDurations` only says whether the end time is usable. For OTel RED response-size histograms, use response-field knownness instead:

```suggestion
if span.ResponseObservation != request.ResponseParsed {
	return
}
```

Add both HTTP server and client cases with a nonzero response length: `ResponseParsed` emits the response-size histogram; `ResponseReceived`, `ResponseSilent`, and `ResponseUnread` do not. The request-size histogram must remain present in all four cases.

pkg/export/prom/prom.go
```diff
@@ -969,6 +969,66 @@ func (r *metricsReporter) collectMetrics(ctx context.Context) {
...
+		requestSize, responseSize = r.httpClientRequestSize, r.httpClientResponseSize
+		requestAttrs, responseAttrs = r.attrHTTPClientRequestSize, r.attrHTTPClientResponseSize
+	default:
+		return
+	}
+
+	r.observeHistogram(requestSize.WithLabelValues(labelValues(span, requestAttrs)...).Metric,
+		float64(span.RequestBodyLength()), span)
+
+	if request.IgnoreDurations(span) {
```
MrAlias: Prometheus has the same response-size bug. Gate this field on the observation state, not duration:

```suggestion
if span.ResponseObservation != request.ResponseParsed {
	return
}
```

Add the matching server and client cases for all four observation states. Parsed responses emit the response-size histogram; the other three do not; request-size histograms remain present.

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -544,20 +544,30 @@ func messagingOperationAttrs(method string) []attribute.KeyValue {
...
 	case request.EventTypeHTTP:
 		attrs = []attribute.KeyValue{
-			request.HTTPResponseStatusCode(span.Status),
 			request.ClientAddr(request.PeerAsClient(span)),
 			request.ServerAddr(request.SpanHost(span)),
 			request.ServerPort(span.HostPort),
 			request.HTTPRequestBodySize(int(span.RequestBodyLength())),
 			request.HTTPResponseBodySize(span.ResponseBodyLength()),
 		}
+		attrs = appendHTTPResponseStatus(attrs, span)
```
MrAlias: Status is conditional now, but `http.response.body.size` is still added unconditionally for both server and client traces. Please remove response size from the two base attribute slices, rename this helper to `appendHTTPResponseAttrs`, and make it own both response fields:

```go
func appendHTTPResponseAttrs(attrs []attribute.KeyValue, span *request.Span) []attribute.KeyValue {
	if span.ResponseObservation != request.ResponseParsed {
		return append(attrs, attribute.Bool(string(attr.OBIHTTPResponseObserved), false))
	}

	return append(attrs,
		request.HTTPResponseBodySize(span.ResponseBodyLength()),
		request.HTTPResponseStatusCode(span.Status),
	)
}
```

Metrics need the same status-knownness rule. In `pkg/appolly/app/request/span_getters.go`, make the status getter invalid for non-parsed responses:

```go
case attr.HTTPResponseStatusCode:
	getter = func(s *Span) attribute.KeyValue {
		if s.ResponseObservation != ResponseParsed {
			return attribute.KeyValue{}
		}
		return HTTPResponseStatusCode(s.Status)
	}
```

The existing Prometheus wrapper will map that invalid value to an empty fixed label. Add server and client trace cases with nonzero seeded status and body size: parsed keeps both; the other three omit both and keep `obi.http.response.observed=false`.

MrAlias (commented):
Response observation still needs one invariant applied consistently: userspace-parsed fields win, and exporters only publish response fields they actually know.

pkg/export/otel/metrics.go
```diff
@@ -928,218 +928,291 @@ func (r *Metrics) record(span *request.Span, mr *MetricsReporter) {
...
-				gcalls, attrs := r.gpuKernelCallsTotal.ForRecord(span)
-				gcalls.Add(ctx, 1, instrument.WithAttributeSet(attrs))
+		if measured {
+			r.recordDuration(span, mr, ctx, duration)
+		}
+
+		r.recordBodySizes(span, mr, ctx)
+	}
+
+	if measured {
```
MrAlias: `FeatureSpanSizes` is independent of `FeatureSpanMetrics`, so this outer duration guard drops a known request-size counter for every `ResponseReceived` span. The calls/latency pairing does not apply to the size counters.

Please split the behavior exactly as follows:

- `SpanMetrics && measured`: record latency and calls.
- `SpanMetrics && !measured`: record neither.
- `SpanSizes`: always record request size.
- `SpanSizes && ResponseParsed`: record response size.

Mirror the split in Prometheus. Add one OTel test and one Prometheus test with only `FeatureSpanSizes` configured on both the exporter and the span service. Set `ResponseObservation: ResponseReceived` and call `request.SetIgnoreDurations(&span)`. Each test must find the request counter and not find the response counter.

grcevski:
I think we need to add `obi.http.response.observed` to traces.yaml to make sure weaver doesn't complain.

grcevski:
I noticed maybe small issue with `self_referencing_request`. This helper is sort of unrelated to the change you are making. It's used to detect when a client request points to the same server, so technically we try to match the correct trace id.

The problem is that we are now not cleaning up the requests that are `unread`. So these client requests will stick around in the map until kicked out.

So I thought of the following scenario:

request A -> client call C1 -> unparsed response

request B -> client call C2 -> (on the same socket, say keep-alive, or just unlucky)

What we'll end up now is:

```
Expected: A -> C1    B -> C2
Actual:   A -> C1
               | -> C2
               B -> empty
```

The reason is that C2 will check self-referencing request and will find the stale info from C1 and then take it as gospel.

So the issue is that we check `self_referencing_request` before we do any sort of resetting of the trace information for a given connection tuple.

The `self_referencing_request` was never intended for client calls, it's there to ensure a server call on the same process can take the traceparent information from the client call. But we don't check that, since before we assumed we must have deleted the finished client call.

What we need to do is check inside `self_referencing_request` that the request we are concerned with is a server request, before we go on an look up a client request. You can pass in this info by an argument and get it like:

```
request_type_by_direction(direction, packet_type) == EVENT_HTTP_REQUEST
```

mifisignal:
All of this is addressed. Summary of what changed:

**Parsed fields now win.** `http_transform.go` has two paths instead of one. When `http.ReadResponse` succeeded, `keepParsedResponse` keeps the parsed status and length and only drops the duration for `bpfResponseReceived`. The unparsed path still clears the status. There's a test that parses a response with `Content-Length: 17` while `event.RespLen` says something else, and checks 17 wins.

**Exporters only publish response fields they know.** Response size in OTel and Prometheus is now gated on `ResponseObservation != ResponseParsed`, not on the duration. `tracesgen` has a single `appendHTTPResponseAttrs` helper that owns both status and body size, and the status getter in `span_getters.go` returns an invalid attribute when the response wasn't parsed.

**Span sizes are split from span metrics.** They're separate features, so the duration guard no longer covers both. Latency and calls need `SpanMetrics && measured`; request size only needs `SpanSizes`; response size needs `SpanSizes && ResponseParsed`. Same split in Prometheus. Both have a test with only `FeatureSpanSizes` on, an unmeasured span, and a check that the request counter is there and the response counter isn't.

**`self_referencing_request` is server-only now.** It takes `direction`, derives the incoming request type, and only looks up a client record for a server request. So C2 on a kept-alive socket can't pick up C1's stale parent. There's an integration test for exactly that sequence.

**Schema.** The attribute is in `schemas/obi/groups/traces.yaml`. One difference from the suggestion: I made it a string enum, `obi.http.response.observation`, instead of a boolean `obi.http.response.observed`. A boolean only says "not parsed", but there are three different ways that happens, and they mean different things — whether the duration is usable depends on which one. The enum keeps that. I can switch to the boolean if you'd rather.

Ran the Weaver live check (it's built into the response-observation compose stack, since that's the only suite emitting the attribute). No violations. `make lint-schema` passes. `TestSuite_ResponseObservation` passes, 22 subtests.

grcevski (approved):
LGTM!

MrAlias (approved):
Nice work 😄 

Please resolve the current conflicts with `main` and refresh the PR description: it still references the removed `unobserved_response`/`obi.http.response.observed` names, says connection-reuse loss is out of scope, names a test file that no longer exists, and describes the superseded status-`0` metric behavior.

mmat11:
the metrics refactor + the rebase lost the changes of a couple of PRs merged yesterday (https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3217 https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3303)
would it be possible to revert the refactor of metrics.go and only do minimal changes that affect the HTTP duration path?

mifisignal:
Restored those two branches (and a stale-TODO comment #3303 had cleaned up) in the latest push. Kept the `recordDuration`/`recordBodySizes` split . Audited the other touched files for the same failure mode; nothing else was affected.

Sorry about that! Should be copacetic now.

mmat11:
> Restored those two branches (and a stale-TODO comment #3303 had cleaned up) in the latest push. Kept the `recordDuration`/`recordBodySizes` split . Audited the other touched files for the same failure mode; nothing else was affected.
> 
> Sorry about that! Should be copacetic now.

I still see the refactor in metrics.go; is that necessary?

mifisignal:
@mmat11 I updated metrics.go to make the updates as minimal as possible. Is this better?

mmat11 (approved):

PR #3091, opened 2026-08-19
Emit `db.server.operation.duration` for server-side DB spans

### LINKED ISSUE #3031: Inconsistent metrics for DB server-side spans: Redis/Memcached emit `db.client.operation.duration`, SQL emits nothing
### Component

OBI (eBPF instrumentation)

### Problem

Hi everyone, I'm opening this issue because I found this `TODO` in [metrics_test.go#L224](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/export/otel/metrics_test.go#L224) (// Redis server GET (TODO is this a bug?)).

Basically, OBI handles server-side database spans inconsistently in both metric exporters:

`EventTypeRedisServer` and `EventTypeMemcachedServer` spans are recorded under `db.client.operation.duration` ([metrics.go#L981](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/export/otel/metrics.go#L981), [metrics.go#L1001](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/export/otel/metrics.go#L1001), [prom.go#L1078](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/export/prom/prom.go#L1078), [prom.go#L1094](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/export/prom/prom.go#L1094)).
`EventTypeSQLServer` spans have no case in either exporter, so they are silently dropped from metrics. Traces are consistent: all of these get `SpanKindServer`.

Two problems with the current Redis/Memcached behavior:

- Semconv defines `db.client.operation.duration` for database client instrumentations; there is no `db.server.* duration` metric. Recording a server-side measurement there attributes the metric to the server's service as if it were a client.
- When both sides are instrumented, the same operation is counted twice under the same metric name.

There is also an argument for it: when only the server process is instrumentable (external or uninstrumented clients), the server span is the only visibility into DB latency, and emitting nothing loses that signal. So SQL server spans should behave the same, and they don't. 

Pick one convention and apply it to all DB protocols in both exporters
1. all DB server spans emit `db.client.operation.duration`, with the server-only-visibility rationale documented; or
2. no DB server spans emit it (semconv-strict), accepting the visibility loss when clients aren't instrumented.
3. any other ideas?

Ether direction is a user-visible metrics change, so I think we should decide how to proceed before opening a PR.
I'm open to suggestions.

_No response_

### Additional context

_No response_

pinoOgni:
## Summary

resolves #3031 

This PR adds a new `db.server.operation.duration` histogram and it also moves Redis/Memcached server spans to it. SQL server spans now emit it too. 

Notes:
- Breaking change: dashboards reading server-side Redis/Memcached measurements from `db.client.operation` must switch to this new metric. 
- Integration test: added `TestSuite_PythonRedisServerSide` that asserts per-operation `db_server_operation_duration_seconds` counts and the client-metric is absent. Memcached/SQL server-side integration variants are omitted becayse they share the same one-case switch path redis covers; can be added on request.
- Follow-up: the `db.server.operation.duration` metric is OBI-custom, indeed upstream semconv only defines the client one. A semconv proposal will follow, and the registry entry documents the contract in the meantime.

## Validation

### DIFF
--- a/internal/test/integration/docker-compose-python-redis-server.yml
+++ b/internal/test/integration/docker-compose-python-redis-server.yml
@@ -1,0 +1,113 @@
+version: "3.8"
+
+services:
+  redis:
+    build:
+      context: ../../../internal/test/integration/components/redis/
+      dockerfile: Dockerfile
+    image: redis
+
+  testserver:
+    build:
+      context: ../../../internal/test/integration/components/pythonredis/
+      dockerfile: Dockerfile
+    image: hatest-testserver-python-redis
+    ports:
+      - "${TEST_SERVICE_PORTS}"
+    depends_on:
+      otelcol:
+        condition: service_started
+      redis:
+        condition: service_started
+
+  # OBI instruments the redis server process (not the python client), so the
+  # emitted metrics are the server-side db.server.operation.duration family.
+  obi:
+    build:
+      context: ../../..
+      dockerfile: ./internal/test/integration/components/obi/Dockerfile
+    volumes:
+      - ./configs/:/configs
+      - ./system/sys/kernel/security:/sys/kernel/security
+      - ../../../testoutput:/coverage
+      - ../../../testoutput/run-python-redis-server:/var/run/obi
+    image: hatest-obi
+    privileged: true # in some environments (not GH Pull Requests) you can set it to false and then cap_add: [ SYS_ADMIN ]
+    network_mode: "service:redis"
+    pid: "service:redis"
+    environment:
+      OTEL_EBPF_CONFIG_PATH: "/configs/obi-config-redis.yml"
+      GOCOVERDIR: "/coverage"
+      OTEL_EBPF_TRACE_PRINTER: "text"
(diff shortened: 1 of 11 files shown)


grcevski (approved):
LGTM!

One question (which can be a follow-up), could this same thing be a problem for Aerospike? Since we added the client protocol, if we instrument the Aerospike server itself, would it be similar?

mmat11 (approved):

pinoOgni:
> One question (which can be a follow-up), could this same thing be a problem for Aerospike? Since we added the client protocol, if we instrument the Aerospike server itself, would it be similar?

Yep it's in my todo list.  Right now Aerospike is client-side only and we drop server-side events so an instrumented Aerospike server doesn't double-report, and there's no `EventTypeAerospikeServer`. I'll add server-side Aerospike spans, and they'd need the same `db.server.operation.duration` treatment.

PR #3097, opened 2026-08-19
Add `server.port` attribute to `db.client.operation.duration`

pinoOgni:
## Summary

This PR simply adds `server.port` as a default attribute on the `db.client.operation.duration` metric.
[Otel secomv](https://opentelemetry.io/docs/specs/semconv/db/database-metrics/) recommends both `server.address` and `server.port` on this metric.

We are following the semconv conditional requirement: reported by default only for a known port other than the DBMS default and only when server.address is set. A user selection that explicitly includes `server.port` (exact name or glob, e.g. include: ["*"]) reports any known port, including the DBMS default; an unknown port is always omitted, never emitted as 0.

## Validation

### DIFF
--- a/pkg/export/attributes/attr_defs.go
+++ b/pkg/export/attributes/attr_defs.go
@@ -390,6 +390,7 @@ func getDefinitions(
 			SubGroups: []*AttrReportGroup{&appAttributes},
 			Attributes: map[attr.Name]Default{
 				attr.ServerAddr:       true,
+				attr.ServerPort:       true,
 				attr.DBOperation:      true,
 				attr.DBSystemName:     true,
 				attr.ErrorType:        true,
--- a/internal/test/integration/red_test_python_redis.go
+++ b/internal/test/integration/red_test_python_redis.go
@@ -38,6 +38,8 @@ func testREDMetricsForPythonRedisLibrary(t *testing.T, testCase TestCase) {
 	for _, span := range testCase.Spans {
 		require.EventuallyWithT(t, func(ct *assert.CollectT) {
 			var err error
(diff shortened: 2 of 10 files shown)


pkg/export/attributes/attr_defs.go
```diff
@@ -390,6 +390,7 @@ func getDefinitions(
 			SubGroups: []*AttrReportGroup{&appAttributes},
 			Attributes: map[attr.Name]Default{
 				attr.ServerAddr:       true,
+				attr.ServerPort:       true,
```
MrAlias, line 393: The database semconv makes `server.port` conditional: a known non-default port is included by default only when `server.address` is set. A known default port, or a known port without an address, can still be available through explicit inclusion; an unknown port should be omitted entirely.

With this enabled unconditionally, the Redis test adds the default port `6379` to the default metric set, while a DB duration measurement whose connection lookup leaves `HostPort` unset emits `server.port=0`. Could we apply the distinction above? Tests should cover default and explicit selection for a non-default port with an address, a default port, a known port without an address, and a zero port.

MrAlias (changes requested):
The default handling of `server.port` needs one adjustment to match the database semantic convention.

pkg/export/attributes/attr_defs.go
pinoOgni: Thanks @MrAlias, you are right! I applied your suggestions.

grcevski (approved):
LGTM!

mmat11 (approved):

MrAlias (approved):

PR #3098, opened 2026-08-19
fix: keep conditional parent settlement coherent

MrAlias:
## Summary

- mark every span emitted from a TCP event with its captured conditional-parent state
- settle conditional parents before attribute, instrumentation, or dynamic-signal filters can separate a child from its parent
- detach stale parents without changing trace IDs that may already have propagated downstream

## Why

The conditional-parent handling added in #3001 only marked the primary span returned by a TCP protocol parser. Kafka, Redis, NATS, AMQP, and Memcached can emit additional spans before that primary span returns, so those spans bypassed settlement with an unconditional stale parent.

Settlement also ran after filtering. A filter that retained a client but discarded its server parent left the client unprovable and eventually detached it even when it started before the server ended.

Finally, detaching a stale parent assigned the client a new trace ID after the original ID might already have propagated. Downstream spans then referenced a client that was no longer present in their trace. Preserving the trace ID keeps the propagated topology coherent while removing the stale parent relationship.

## Validation

- `make generate`
- `go test -race ./pkg/ebpf/common ./pkg/appolly`
- `go tool -modfile=internal/tools/go.mod golangci-lint run ./pkg/appolly/... ./pkg/ebpf/common/... --timeout=6m`


pkg/appolly/parent_settle.go
```diff
@@ -146,18 +145,12 @@ func settle(span request.Span, parentEnd int64) request.Span {
...
 
-// reroot detaches a span whose claimed parent had already finished when the
-// span started: it becomes its own trace. When context propagation put the
-// original ids on the wire, downstream services keep them: their spans then
-// group under the stale trace instead of following this one, which is where
-// today's merging already leaves them. Correcting the wire would require
-// knowing the request had ended before its bytes left, which only the
-// response's content can say.
-func reroot(span request.Span) request.Span {
-	span.TraceID = idgen.RandomTraceID()
```
grcevski, line 160: Hm, do we need to remove this line that calculated the new TraceID?

grcevski (commented):
I think the related tests are failing, it might be because we don't reset the traceID anymore?

pkg/appolly/parent_settle.go
MrAlias: The removal was intentional, but the failing integration tests mean I haven’t established that it is correct across the full pipeline. The goal is to avoid changing a trace ID that may already have propagated downstream, because that would split the child from its descendants. I’m checking the failing tests now to determine whether they encode the old re-rooting behavior or expose a real regression; I’ll either update the expectations with evidence or restore the trace-ID reset.

grcevski (approved):
LGTM!

PR #3089, opened 2026-08-19
Discover java service names like OpenTelemetry Java

grcevski:
## Summary

When not running in Kubernetes/Docker, OBI gets only executable metadata from the instrumented processes. Various other OTel SDKs pull the name based on language metadata found in metadata files or the file name itself.

This initial PR adds name resolution for Java services as per the OTel Java rules:

- SpringBootServiceNameDetector (spring boot specific)
- MetadataResourceExtractor (jar encoded metadata)
- JarServiceNameDetector (service name based on the jar file name)

The ordering rule is: Spring Boot name , then manifest title, then JAR basename, then keep it as `java`.

## Validation

- [X] I have read and followed the [contributing guidelines](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/CONTRIBUTING.md)
- [ ] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed.

PR #3099, opened 2026-08-19
nodejs: extract meaningful node application names when other metadata is missing

grcevski:
## Summary

This PR adds support for extracting node metadata to name the applications, in case we lack higher level metadata like Kubernetes, Docker, Cloud vendor. It's similar to what I did for [Java](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3089).

It prefers the metadata encoded in the various node application deployment files. When the name is scoped like @scope/name we make the scope the namespace and the name the service.name. It extracts the service.version like Java does and if all fails it falls back on the script name. 

Keeping it as draft initially because I built on top of the Java PR, which isn't merged yet.

## Validation


pkg/internal/nodejstools/launch.go
```diff
@@ -0,0 +1,186 @@
...
+	"--http-parser":                       {},
+	"--icu-data-dir":                      {},
+	"--import":                            {},
+	"--input-type":                        {},
+	"--inspect-port":                      {},
+	"--inspect-publish-uid":               {},
+	"--loader":                            {},
+	"--localstorage-file":                 {},
+	"--max-http-header-size":              {},
+	"--max-old-space-size-percentage":     {},
```
mmat11, line 60: Claude tells me: plain `--max-old-space-size`, `--stack-size`, `--max-semi-space-size` and the rest of the V8 numeric flags are absent.

In general, any flag that's missing from here leaves its value in the arg stream and the next iteration returns that value as the entrypoint. Worth guarding the result (e.g. require a .js/.mjs/.cjs/.ts suffix or an existing path) rather than relying on the list staying exhaustive

grcevski: Ah yeah, I think it's version dependent and development build dependent, I guess I should've actually tried to see what gets printed as options. If you put an option that doesn't exist, node won't start, but I like your suggestion, I can make the parsing so much simpler and less error prone if a new option is added in the future.

grcevski: OK, I made it a lot simpler. I still parse a few command line options, because now .js can be a module import and we want to skip those. I should skip with <space> syntax and = syntax.

pkg/internal/nodejstools/launch.go
```diff
@@ -0,0 +1,57 @@
...
+			}
+			if nodeOptionConsumesScriptArgument(arg) {
+				skipNext = true
+				continue
+			}
+			if strings.HasPrefix(arg, "-") {
+				continue
+			}
+		}
+		if hasNodeScriptExtension(arg) {
```
MrAlias: The suffix filter fixes the numeric-flag case, but it also drops valid extensionless entrypoints and keeps scanning application arguments until one happens to end in `.js`. For `node /opt/orders/server --config worker.js`, Node executes `/opt/orders/server` while this returns `worker.js`; `--entry-url file:///opt/orders/main.js` similarly produces no entrypoint because its value is discarded. Both `ResolveServiceMetadata` and `FindNodeJSAppDir` consume this result, so service identity and route harvesting can target the wrong package/tree. Could we preserve these valid entrypoint forms and cover them with caller-level regression tests?

MrAlias (commented):
The Node metadata discovery still needs to preserve valid entrypoint forms.

pkg/internal/nodejstools/launch.go
grcevski: Right, good point. I remember now why I thought I should parse all possible options. A user can do `node app` and if there's a file `app.js` it will pick it up. I think I'm going to go back to my extensive parser and remove the debug options.

pkg/internal/nodejstools/launch.go
grcevski: Btw, @mmat11 https://nodejs.org/api/cli.html --max-old-space-size,  --max-semi-space-size are there, stack-size was removed in 18 (https://github.com/orgs/nodejs/discussions/5014) but it's still valid to consider, if it's invalid option node won't start.

pkg/internal/nodejstools/launch.go
grcevski: I reverted and refined a bit my earlier version, that version handles both of these cases well. I've added regression tests with those two exact scenarios and a bit more.

pkg/internal/nodejstools/launch.go
grcevski: I pushed an updated version that guards against unknown new options that might get added to future node versions. I watch for dashed options and if they are not known and they are followed by a non-dashed option, I bail and end the scan, defaulting to just node.

mmat11: @grcevski can we add a test with `node --max-old-space-size 4096 app.js` in TestParseNodeLaunch?

grcevski: Ah nice, you caught a bug! let me look into where it breaks.

grcevski: Actually that should be invalid, I just checked. The --max-old-space-size is not a valid skip next option and you wouldn't be able to run `node` with it as such. What we'd do is name the app 4096, but this is what happens when I launch it like that:

```sh
node --max-old-space-size 4096 app.js 
Error: illegal value for flag --max-old-space-size of type size_t
Try --help for options
node: bad option: --max-old-space-size
node: bad option: --harmony-import-attributes
node: bad option: --js-source-phase-imports
```

when it's used correctly:
```
node --max-old-space-size=4096 app.js
Process PID: 4187404
Send SIGUSR1 to toggle profiling: kill -SIGUSR1 4187404
0
Server running on port 3030
```

So I don't think we'll find a service that has a command line like the one you described. 

I can add safety checks if you think it's required. For example, if we decide to pick a name based on the file name argument, then there must be a file in the folder with `.js/.mjs/.cjs/.ts` suffix. I can validate that and reject 4096. WDYT?

grcevski: I added a safety check for the local file. My thinking was, while the current semantics for the node option is "not skip next" they can potentially change that in the next release and we wouldn't know. So now `ResolveServiceMetadata` ensures that when we found the script we think is the app starting point, we validate if it's a valid script file, and if not we default back to node.

mmat11: if it's been tested then it's already good for me! just wanted to make sure

mmat11 (approved):
lgtm

skl (approved):
lgtm!

pkg/internal/nodejstools/launch.go
MrAlias: The parser now preserves the command-line form from this thread, but the filesystem guard validates it with a different resolution model than Node. With Node 22, an existing extensionless file and a directory resolved through `package.json` or `index.js` are valid entrypoints; an extensionless argument also resolves a `.js` file, but not `.mjs`, `.cjs`, or `.ts`. `nodeScriptExists` skips the exact extensionless file and directory, while the current test matrix treats all four source suffixes as inferred matches. Without a valid package name, the rejected valid cases leave the service name unset and it defaults to `node` instead of the entrypoint basename.

Could we align this guard and its tests with Node's resolution order: exact file first, then the runtime-supported extension fallbacks, then directory resolution? In particular, the exact extensionless case should be accepted, and the explicit `.mjs`/`.cjs`/`.ts` cases should be tested separately from extension inference.

grcevski: Good catch! I pushed an update, let me know if this is what you had in mind. I believe you were suggesting we follow the logic in here: https://github.com/nodejs/node/blob/v22.23.0/lib/internal/modules/cjs/loader.js#L683-L784

pkg/internal/nodejstools/launch_test.go
```diff
@@ -0,0 +1,99 @@
...
+				"--title", "orders",
+				"--watch-path", "src",
+				"app.js",
+			},
+			entryPoint: "app.js",
+		},
+		{name: "underscore option", args: []string{"--inspect_port", "9230", "app.js"}, entryPoint: "app.js"},
+		{name: "V8 option with attached value", args: []string{"--stack-trace-limit=20", "app"}, entryPoint: "app"},
+		{name: "attached config file", args: []string{"--experimental-config-file=node.config.json", "app"}, entryPoint: "app"},
+		{name: "default config file", args: []string{"--experimental-config-file", "app"}, entryPoint: "app"},
```
MrAlias, line 54: Non-blocking follow-up: Node 22.23.2 consumes the path after `--experimental-config-file` (the same as the `=path` form), but this test makes that path the entrypoint because the option is absent from `optionsWithValues`. When that config file is extensionless and no explicit or package name supplies the identity, the new exact-file guard accepts it and names the service after the config; an absolute config path in another directory also makes `FindNodeJSAppDir` scan that directory instead of the script passed next.

I think we can take a look into addressing this by restoring this option to `optionsWithValues` and covering`--experimental-config-file config server.js with server.js as the expected entrypoint in a follow-up PR once this merges..

MrAlias (approved):
🚀

PR #3105, opened 2026-08-20
nodejs: export manual span kind; honor explicit span.end(t) within bounded skew

NimrodAvni78:
## Summary

Two small fidelity follow-ups to #2661.

**Span kind.** The payload already carried `kind`, but every manual span was exported as `Internal`. The decoder now maps the OTel JS kind enum to Go's `trace.SpanKind` (the enums are offset by one; invalid values fall back to `Internal`) and sets it on the span. No exporter change was needed — the traces pipeline already derives the exported kind from `span.SpanKind`, the same way Go Auto SDK spans do.

**Explicit end timestamps.** `span.end(t)` previously ignored its argument; the sentinel's BPF ktime was always the end time. The bridge now serializes a valid explicit end (`Date`, epoch millis, or `[seconds, nanos]` hrtime — all handled defensively, never throwing) as an optional `endWallNs` payload field. The decoder translates it wall→monotonic (same approach as the Go Auto SDK path) and honors it only within a 5-minute skew of the BPF ktime anchor; outside that bound it falls back to the sentinel time. The bound exists because the rest of the pipeline assumes one monotonic time domain — a frozen or misconfigured app clock must not corrupt trace timing.

Payload docs updated; the span-kind limitation removed from the devdoc.

## Validation


mmat11 (approved):

pkg/ebpf/common/node_otel_transform.go
```diff
@@ -94,6 +97,13 @@ func ReadNodeSpanEventIntoSpan(record *ringbuf.Record) (request.Span, bool, erro
...
 	}
 	end := int64(event.EndKtime)
+	if rec.EndWallNs != "" {
+		if wallEnd, err := strconv.ParseInt(rec.EndWallNs, 10, 64); err == nil {
+			if monoEnd, ok := nodeExplicitEnd(wallEnd, end); ok {
+				end = monoEnd
+			}
+		}
+	}
 	start := end - durNs
```
MrAlias: `Span.end(endTime)` changes the span's end; it does not move its start. This decoder replaces `end` with the explicit timestamp and then subtracts the bridge's elapsed `durNs`. If a span starts at `t0` and `end(t0+500ms)` is called one second later, the expected interval is `t0` to `t0+500ms`; this code exports `t0-500ms` to `t0+500ms` instead. Every accepted explicit time that differs from the actual `end()` call time therefore shifts the start and reports the wrong duration. Could we translate the serialized `startNs` (or otherwise recompute the duration from the original start and explicit end) instead of preserving `durNs` by moving the start?

pkg/internal/nodejs/spanbridge.js
```diff
@@ -216,6 +216,48 @@
...
+        return msToNs(t.getTime());
+      }
+      if (typeof t === 'number') {
+        // Matching @opentelemetry/core: a number below performance.timeOrigin
+        // is a performance.now()-style offset, not an epoch timestamp.
+        const origin = globalThis.performance && globalThis.performance.timeOrigin;
+        if (typeof origin === 'number' && Number.isFinite(t) && t < origin) {
+          return msToNs(origin + t);
+        }
+        return msToNs(t);
```
MrAlias, line 243: `performance.timeOrigin` can be slightly ahead of `Date.now()` after a clock adjustment. In that state, a valid epoch-millisecond `endTime` falls below `timeOrigin`, this code adds the origin a second time, and the decoder rejects the resulting timestamp through its five-minute skew guard, silently falling back to the sentinel time. `@opentelemetry/core` 2.10 uses `time < performance.timeOrigin / 2` specifically to distinguish relative readings without misclassifying epoch milliseconds; could we use the same check and add the just-below-origin case to the bridge test?

MrAlias (approved):

PR #3106, opened 2026-08-20
nodejs: nest eBPF client spans under correct manual spans

NimrodAvni78:
## Summary

Follow-up to #2661, closing the main structural gap vs the Go manual-span path: 
OBI's automatic client spans parented to the request's *server* span,
so an outgoing HTTP call made inside a manual span rendered as its sibling instead of its child.

The span bridge now publishes manual-span context transitions through`/dev/null/obi-mspan/<traceId><spanId>`. On push, 
BPF overrides the **span id** of the thread's `traces_ctx_v1` entry (the trace id is untouched), saving the server context in a small map; on pop — and on every per-callback `-ctx` refresh — the base context is restored. Client-span parenting and downstream traceparent propagation already read `traces_ctx_v1`, so both nest under the active manual span with no consumer-side changes.

Integration coverage extends the `/manual` suite: an outgoing client call made inside `checkout` must be its child (not the server span's), and the existing flat assertions still hold for spans outside manual scope.

## Validation


bpf/common/trace_parent.h
```diff
@@ -350,6 +351,32 @@ static __always_inline tp_info_pid_t *find_parent_trace(const pid_connection_inf
...
+static __always_inline u8 nodejs_manual_reparent_client(u64 pid_tgid, tp_info_t *tp) {
+    const obi_ctx_info_t *shadow = bpf_map_lookup_elem(&node_manual_ctx_shadow, &pid_tgid);
+    if (!shadow) {
+        return 0;
+    }
+    const obi_ctx_info_t *live = bpf_map_lookup_elem(&traces_ctx_v1, &pid_tgid);
+    if (!live) {
+        return 0;
+    }
+    for (u8 i = 0; i < TRACE_ID_SIZE_BYTES; ++i) {
```
grcevski: Maybe use bpf_memcmp here for consistency and brevity, like we do elsewhere:

    if (bpf_memcmp(existing->tp.trace_id, info->tp.trace_id, TRACE_ID_SIZE_BYTES) != 0 ||

bpf/generictracer/nodejs.c
```diff
@@ -122,6 +126,18 @@ nodejs_v8_parse_heap(const unsigned char *payload, u64 len, u64 *vals, u8 *name_
     return 0;
 }
 
+// True if the context carries a non-zero trace id. An all-zero trace id is used
+// as the shadow map's "no base existed" marker (real request trace ids are
+// random 16-byte values and are never all zero).
+static __always_inline bool ctx_has_trace_id(const obi_ctx_info_t *c) {
+    for (u8 i = 0; i < TRACE_ID_SIZE_BYTES; ++i) {
```
grcevski: Maybe use :
```
static __always_inline u8 valid_trace(const unsigned char *trace_id) {
    return *((u64 *)trace_id) != 0 || *((u64 *)(trace_id + 8)) != 0;
}
```
I think that was the intent, right?

pkg/internal/nodejs/spanbridge.js
```diff
@@ -483,6 +552,56 @@
...
+          if (hookDepth === 0 && overrideEmitted) {
+            emitPop();
+            overrideEmitted = false;
+          }
+        } catch (_) {
+          // never let the hook throw into the app
+        }
+      },
+    });
+    mspanHook.enable();
```
grcevski, line 626: Should we save an error state in a variable if we've caught any exceptions and skip calling .enable() here?

bpf/common/trace_parent.h
```diff
@@ -416,6 +451,14 @@ find_parent_trace_for_client_request_with_t_key(const pid_connection_info_t *p_c
         }
 
         *tp = server_tp->tp;
+        // A manual-span parent may legitimately never be exported (payload
+        // over budget, SDK handoff mid-span, span outliving the transaction
+        // window); a conditional parent would then be rerooted out of the
+        // trace. Ship immediately instead: worst case is a dangling parent id
+        // inside the correct trace.
+        if (nodejs_manual_reparent_client(pid_tgid, tp)) {
```
grcevski: This helper is called only from the tpinjector. Do we need to consider manual spans here? I think this will mess up the parenting of the client request. I think I don't fully understand the change.

grcevski: OK, I get it, but I think you need to change the helper slightly. At the time of the client call and the server call the parenting is slightly different. So I think you need something like this here:
```
__builtin_memcpy(tp->trace_id, server_tp->tp.trace_id, sizeof(tp->trace_id));
__builtin_memcpy(tp->parent_id, server_tp->tp.span_id, sizeof(tp->parent_id));

if (nodejs_manual_parent_span_id(pid_tgid, tp->trace_id, tp->parent_id)) {
    return k_parent_status_live;
}
```

grcevski: while in the server function you'd call, which matches what the helper does today:

```
*tp = server_tp->tp;

if (nodejs_manual_parent_span_id(pid_tgid, tp->trace_id, tp->span_id)) {
    return k_parent_status_live;
}
```

I added an extra argument.

pkg/internal/nodejs/spanbridge.js
```diff
@@ -160,6 +174,34 @@
...
+  // Once the app's SDK owns telemetry — via a wrapped setter (yielded) or a
+  // registration through an unwrapped copy (detectRegistryHandoff) — the bridge
+  // must stop pointing the kernel context map at its own spans, or stale
+  // overrides would point eBPF client spans at bridge span ids that are no
+  // longer exported. Handoff drops the override rather than skipping the emit:
+  // skipping would latch whichever override was already in the map.
+  const emitOverride = (span) => {
+    if (yielded) return;
+    if (detectRegistryHandoff()) {
+      emitPop();
```
grcevski: I think there's a slight chance that you can get yielded between the handoff and emitPop, and emitPop won't pop the context, since it checks for yielded. I'm fine with this being a limitation, we'll mess up one span, but I don't know if it's possible to emitPop() in yieldToApp just before  we switch yielded = true?

grcevski (commented):
I think it would be useful to add a test with context propagation on, to prove my understanding wrong if nothing else :).

internal/test/integration/suites_test.go
```diff
@@ -424,6 +424,21 @@ func TestSuite_NodeJS(t *testing.T) {
...
+// tpinjector rather than by the generic tracer, so manual-span nesting runs
+// through a different helper. Same assertions, second code path.
+func TestSuite_NodeJSManualSpansPropagation(t *testing.T) {
+	compose, err := docker.ComposeSuite("docker-compose-nodejs-manual-prop.yml",
+		path.Join(pathOutput, "test-suite-nodejs-manual-prop.log"))
+	require.NoError(t, err)
+
+	compose.Env = append(compose.Env, `OTEL_EBPF_OPEN_PORT=3030`, `OTEL_EBPF_EXECUTABLE_PATH=`, `NODE_APP=app`)
+	require.NoError(t, compose.Up())
+	t.Run("HTTP manual spans (OTel API bridge)", testHTTPTracesNodeManualSpans)
```
MrAlias, line 438: This dedicated propagation suite calls `testHTTPTracesNodeManualSpans` immediately after `compose.Up()`, before OBI has necessarily discovered and instrumented the Node process. On the current head, the amd64 integration job hit `connection reset by peer` on the first `/manual` request in the initial run and both retries, so this coverage never reached the parenting assertions and CI remains red. Could we wait for instrumentation readiness here—for example, with `waitForTestComponents(t, "http://localhost:3031")`—before exercising `/manual`?

bpf/maps/node_manual_ctx_shadow.h
```diff
@@ -0,0 +1,27 @@
...
+// reads the base for root-span parenting, the pop marker restores it, and the
+// per-callback '-ctx' / '-noreqctx' refresh clears it (each callback re-derives
+// the base and the bridge re-applies the override right after).
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, u64); // pid_tgid
+    __type(value, obi_ctx_info_t);
+    __uint(max_entries, 1 << 14); // sized with traces_ctx_v1, whose entries it shadows
+    __uint(pinning, OBI_PIN_INTERNAL);
+} node_manual_ctx_shadow SEC(".maps");
```
MrAlias, line 27: `node_manual_ctx_shadow` has no exit cleanup. If a thread or process exits while an override is active, its `pid_tgid` entry remains in this LRU map; if that ID is reused, `handle_manual_ctx` treats the stale shadow as the new thread's saved base. That can re-anchor its manual spans and live log context to the previous trace, while automatic client spans lose the intended manual-span nesting. Could we delete this entry alongside `obi_ctx__del(id)` in `obi_kprobe_sys_exit`, as the `java_vt_threads` cleanup already does for TID reuse?

MrAlias (commented):
The parenting changes line up across the generic-tracer and tpinjector paths, but the new shadow context needs exit cleanup, and the propagation suite currently fails before reaching its assertions.

bpf/maps/node_manual_ctx_shadow.h
grcevski: Since this is nodejs, I think it's almost always going to be the same thread, the event loop. So it's sufficient to clean-up in sys_exit, but I think he's already cleaning it up in `handle_ctx_clear`. Are you thinking of the case when we somehow miss this event?

grcevski (approved):
LGTM!

bpf/maps/node_manual_ctx_shadow.h
MrAlias: Would `handle_ctx_clear` run if the process terminates while the override is active? It only runs on a later request-to-no-request callback, but a synchronous `process.exit()` can terminate before the bridge’s `after` hook or another `before` callback. Wouldn’t that leave the `pid_tgid` entry behind?

I think my original `sys_exit` suggestion is incomplete too. Does `obi_kprobe_sys_exit` run when Node’s main process exits through `exit_group`, or when a fatal signal uses `do_group_exit`? Would `sched:sched_process_exit` be a better cleanup point, since those paths converge on `do_exit`?

grcevski:
This PR needs a rebase with main. I believe all comments are addressed.

grcevski:
Hey @NimrodAvni78, can we resurrect this and move it along?

bpf/common/trace_parent.h
```diff
@@ -367,6 +368,43 @@ static __always_inline tp_info_pid_t *find_parent_trace(const pid_connection_inf
     return 0;
 }
 
+// Node.js manual-span nesting: when a bridge manual span is active, its shadow
```
rafaelroquetto: nit: I'd remove this comment altogether - it's tedious, and the code is much simpler than this makes it appear (and faster to read than the comment itself)

bpf/common/trace_parent.h
```diff
@@ -367,6 +368,43 @@ static __always_inline tp_info_pid_t *find_parent_trace(const pid_connection_inf
...
+// The returned parent may legitimately never be exported (payload over budget,
+// SDK handoff mid-span, span outliving the transaction window). Callers ship it
+// as live rather than conditional anyway: a conditional parent would be rerooted
+// out of the trace, and the worst case here is a dangling parent id — in this
+// trace, or in the bridge's own if the manual span outlived the connection and
+// fell back to the bridge trace id when it ended.
+static __always_inline u8 nodejs_manual_parent_span_id(u64 pid_tgid,
+                                                       const unsigned char *trace_id,
+                                                       unsigned char *span_id_out) {
+    const obi_ctx_info_t *shadow = bpf_map_lookup_elem(&node_manual_ctx_shadow, &pid_tgid);
```
rafaelroquetto: And here I'd add a comment about why this "gate" is necessary - a one liner (i.e. `if there's no shadow, then it's not a nodejs ctx`)

bpf/common/trace_parent.h
```diff
@@ -433,6 +475,12 @@ find_parent_trace_for_client_request_with_t_key(const pid_connection_info_t *p_c
         }
 
         *tp = server_tp->tp;
+
+        // This tp IS the parent: create_trace_info copies parent_tp.span_id
```
rafaelroquetto: Ditto - this comment is just confusing / arguing some sort of implicit context.

bpf/generictracer/nodejs.c
```diff
@@ -147,15 +154,107 @@ static __always_inline int handle_async_switch(char *buf, const u64 pid_tgid) {
...
+    } else {
+        const tp_info_pid_t *tp = trace_info_for_connection(conn, TRACE_TYPE_SERVER);
+        if (tp && tp->valid) {
+            obi_ctx__set(pid_tgid, &tp->tp);
+        } else {
+            obi_ctx__del(pid_tgid);
+        }
+    }
+
+    // Each callback re-derives the base context, so drop any stale manual-span
```
rafaelroquetto: ditto (although perhaps I'd keep the first sentence, not sure).

bpf/generictracer/nodejs.c
```diff
@@ -147,15 +154,107 @@ static __always_inline int handle_async_switch(char *buf, const u64 pid_tgid) {
...
+//
+// The log enricher (logenricher.c) also reads traces_ctx_v1, so while a manual
+// span is active, log lines correlate to it rather than to the server span.
+static __always_inline int handle_manual_ctx(const char *path, const u64 pid_tgid) {
+    // One byte longer than the longest valid payload, so an over-long path is
+    // distinguishable: bpf_probe_read_user_str returns the same count for an
+    // exact-length payload and for a longer one it truncated, and a foreign
+    // process could otherwise have residual bytes decoded as a plausible
+    // override attributed to itself. Same reasoning as handle_runtime_metrics.
+    unsigned char hexbuf[k_mspan_hex_len + 2] = {};
```
rafaelroquetto: The comment says 1 byte long but it adds 2 here - why? Is it the null terminator?

bpf/generictracer/nodejs.c
```diff
@@ -40,6 +43,10 @@ enum {
     k_max_fd_digits = 4,
     // strlen("/dev/null/obi-span/") — the JSON span payload starts here
     k_span_payload_offset = 19,
+    // strlen("/dev/null/obi-mspan/") — the manual-span override payload starts here
+    k_mspan_payload_offset = 20,
+    // hex chars in an override payload: trace id + span id, two chars per byte
+    k_mspan_hex_len = 2 * (TRACE_ID_SIZE_BYTES + SPAN_ID_SIZE_BYTES),
```
rafaelroquetto: isn't this the same as `TRACE_ID_CHAR_LEN + SPAN_ID_CHAR_LEN`?

The comment would be more informative if it actually described the payload itself, e.gf. `[00000000000000000000000000000000][0000000000000000]` or something like that.

bpf/generictracer/nodejs.c
rafaelroquetto: nit: you might consider putting this on a PER_CPU_ARRAY/scratch mem to take pressure off the stack.

bpf/generictracer/nodejs.c
```diff
@@ -147,15 +154,107 @@ static __always_inline int handle_async_switch(char *buf, const u64 pid_tgid) {
...
+    if (n != k_mspan_hex_len + 1) {
         return 0;
     }
 
-    const tp_info_pid_t *tp = trace_info_for_connection(conn, TRACE_TYPE_SERVER);
-    if (tp && tp->valid) {
-        obi_ctx__set(pid_tgid, &tp->tp);
+    obi_ctx_info_t sentinel = {};
+    decode_hex(sentinel.trace_id, hexbuf, 2 * TRACE_ID_SIZE_BYTES);
+    decode_hex(sentinel.span_id, hexbuf + 2 * TRACE_ID_SIZE_BYTES, 2 * SPAN_ID_SIZE_BYTES);
```
rafaelroquetto:
```suggestion
    decode_hex(sentinel.trace_id, hexbuf, TRACE_ID_CHAR_LEN);
    decode_hex(sentinel.span_id, hexbuf + TRACE_ID_CHAR_LEN, SPAN_ID_CHAR_LEN);
```

bpf/generictracer/nodejs.c
```diff
@@ -147,15 +154,107 @@ static __always_inline int handle_async_switch(char *buf, const u64 pid_tgid) {
...
+
+    // Save the pre-override base on the first override of this sync block. A
+    // missing live entry is recorded as an all-zero "no base existed" marker so
+    // pop / span-end can tell it apart from a real server context.
+    obi_ctx_info_t *shadow = bpf_map_lookup_elem(&node_manual_ctx_shadow, &pid_tgid);
+    if (!shadow) {
+        obi_ctx_info_t base = {};
+        const obi_ctx_info_t *live = obi_ctx__get(pid_tgid);
+        if (live) {
+            bpf_memcpy(&base, live, sizeof(base));
```
rafaelroquetto:
```suggestion
            base = *live;
```

PR #3107, opened 2026-08-20
logenricher: only consume pipes registered as tracked stdout/stderr

mmat11:
## Summary

Since https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3055, the log enricher opens `/proc/<pid>/fd/<fd>` at event-capture time and caches the open file for 30 minutes. For a shell command-substitution pipe, that cached fd is the pipe's write end: the child exits, the shell never sees EOF, and every `$(...)` in an enriched process tree hangs until cache eviction — then reads the enriched log line as the command's output. Found on a live cluster: an instrumented shell script froze mid-$() for 10+ minutes and resumed only when the OBI pod exited.

Only touch pipes that are actually log destinations. When a PID is allowed, userspace registers the pipe inodes behind its stdout/stderr in a new log_pipes BPF map; pipe_write skips any pipe whose inode is not registered — non-log pipes are no longer NULed, captured, opened, or written to. Pipe events carry the inode, and userspace writes through a live tracked owner's `/proc` fd rather than the possibly-dead writer's — which also delivers short-lived writers' log lines more reliably, the original goal of #3055.

## Validation

### DIFF
--- a/bpf/logenricher/types.h
+++ b/bpf/logenricher/types.h
@@ -46,11 +46,14 @@ typedef struct log_event {
 const log_event_t *log_event__unused __attribute__((unused));
 
 // bare inode numbers collide across filesystems (pipefs vs on-disk FIFOs,
-// and get_next_ino() wraps at 32 bits), so pipes are keyed by (ino, dev)
+// and get_next_ino() wraps at 32 bits), so pipes are keyed by (ino, dev).
+// dev holds a 32-bit kernel dev_t; u64 keeps the key free of implicit padding
 typedef struct log_pipe_key {
     u64 ino;
     u64 dev;
 } log_pipe_key_t;
+
+const log_pipe_key_t *log_pipe_key__unused __attribute__((unused));
 
 enum tty_driver_type___new {
     TTY_DRIVER_TYPE_SYSTEM,
--- a/bpf/logenricher/types.h
+++ b/bpf/logenricher/types.h
@@ -33,9 +33,11 @@ enum {
 static const char k_newline = '\n';
 
 typedef struct log_event {
+    u64 ino;
     u32 tgid;
     u32 len;
     u32 fd;
+    u32 _pad;
     obi_ctx_info_t ctx;
     u8 file_path[k_pts_file_path_len_max];
     u8 log[];
--- a/bpf/logenricher/types.h
+++ b/bpf/logenricher/types.h
@@ -37,13 +37,20 @@ typedef struct log_event {
     u32 tgid;
     u32 len;
     u32 fd;
-    u32 _pad;
+    u32 dev; // kernel dev_t of the pipe's superblock, disambiguates ino across filesystems
     obi_ctx_info_t ctx;
     u8 file_path[k_pts_file_path_len_max];
     u8 log[];
 } log_event_t;
 
 const log_event_t *log_event__unused __attribute__((unused));
+
+// bare inode numbers collide across filesystems (pipefs vs on-disk FIFOs,
+// and get_next_ino() wraps at 32 bits), so pipes are keyed by (ino, dev)
+typedef struct log_pipe_key {
+    u64 ino;
+    u64 dev;
+} log_pipe_key_t;
 
 enum tty_driver_type___new {
     TTY_DRIVER_TYPE_SYSTEM,
--- a/pkg/internal/ebpf/logenricher/logenricher.go
+++ b/pkg/internal/ebpf/logenricher/logenricher.go
@@ -13,6 +13,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"slices"
 	"strconv"
 	"sync"
 	"unsafe"
@@ -40,6 +41,7 @@ import (
 type LogEvent struct {
 	orig    BpfLogEventT
 	logLine string
+	dest    string
 }
 
 type Tracer struct {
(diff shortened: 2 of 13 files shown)


grcevski (approved):
LGTM!

pkg/internal/ebpf/logenricher/logenricher.go
```diff
@@ -270,6 +276,8 @@ func (p *Tracer) AllowPID(pid app.PID, ns uint32, fi *exec.FileInfo) {
 		p.log.Error(err.Error())
 	}
 
+	p.registerLogPipes(uint32(pid))
```
MrAlias: Non-blocking follow-up: `registerLogPipes` snapshots fd 1 and 2 when the PID is allowed. If the process later redirects stdout or stderr to a new pipe, that pipe remains unenriched while its inode is unregistered. A cached write handle to the old pipe is not retired on redirection either, so it can delay EOF until cache expiry, eviction, or `BlockPID`.

This should not hold up the demonstrated command-substitution safety fix, but could we track reconciling the current fd 1 and 2 targets with their registered inodes and retiring stale map entries and cache handles?

MrAlias (approved):
This fixes the demonstrated command-substitution interference. The remaining stdout/stderr replacement lifecycle can be handled as a non-blocking follow-up.

pkg/internal/ebpf/logenricher/logenricher.go
mmat11: done in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3107/changes/f7a1f81f8268e29bd53f62d52c01d2062e318bf6

bpf/logenricher/types.h
```diff
@@ -33,16 +33,25 @@ enum {
 static const char k_newline = '\n';
 
 typedef struct log_event {
+    u64 ino;
     u32 tgid;
     u32 len;
     u32 fd;
+    u32 dev; // kernel dev_t of the pipe's superblock, disambiguates ino across filesystems
```
pinoOgni, line 40: `dev` in `log_pipe_key_t` is `u64` and here it is `u32`

pkg/internal/ebpf/logenricher/logpipes_test.go
```diff
@@ -0,0 +1,564 @@
...
+func newLogPipesMap(t *testing.T, maxEntries uint32) *ebpf.Map {
+	t.Helper()
+
+	if err := rlimit.RemoveMemlock(); err != nil {
+		t.Skipf("removing memlock failed: %v", err)
+	}
+	m, err := ebpf.NewMap(&ebpf.MapSpec{
+		Name:       "log_pipes_test",
+		Type:       ebpf.Hash,
+		KeySize:    16,
```
pinoOgni: nit: `uint32(unsafe.Sizeof(pipeKey{}))` so the test fail  if the key grows

pkg/internal/ebpf/logenricher/logenricher.go
```diff
@@ -40,6 +42,15 @@ import (
 type LogEvent struct {
 	orig    BpfLogEventT
 	logLine string
+	dest    string
+	pin     pipeKey // destination identity the write must land on; zero when unpinned
+}
+
+// pipe identity as the BPF map sees it: inode number plus the kernel dev_t of
+// its superblock, since bare inode numbers collide across filesystems
+type pipeKey struct {
```
pinoOgni: We can add `-type log_pipe_key_t` to the `//go:generate` so we have a generated `BpfLogPipeKeyT` as the map key and avoid hand-mirrored code.

pkg/internal/ebpf/logenricher/logenricher.go
```diff
@@ -289,7 +312,188 @@ func (p *Tracer) AllowPID(pid app.PID, ns uint32, fi *exec.FileInfo) {
...
+		return pipeKey{}
+	}
+
+	return pipeKey{Ino: st.Ino, Dev: kernelDev(st.Dev)}
+}
+
+// registerLogPipes records the pipes currently behind stdout/stderr, retiring
+// registrations the process redirected away from. The stats run unlocked so
+// slow /proc access never stalls the event hot path
+func (p *Tracer) registerLogPipes(pid uint32) {
```
pinoOgni, line 368: AFAIU right now if an app writes to a `fd` that is not 1 or 2 (and that was enriched before this PR) is silently dropped right? Not sure if this your desired behavior, if yes we can add a note in the doc with "only stdout/stderr pipes are enriched; if not we may scan the process's whole `fs` table for pipes or make the `fd` set configurable (not sure if it makes sense..)

pinoOgni (changes requested):

pkg/internal/ebpf/logenricher/logenricher.go
mmat11: yeah that's the desired behavior - safety net. I added a line in the devdocs

pinoOgni (approved):

pkg/internal/ebpf/logenricher/logenricher.go
```diff
@@ -289,7 +314,188 @@ func (p *Tracer) AllowPID(pid app.PID, ns uint32, fi *exec.FileInfo) {
...
+		p.pidPipes[pid] = make(map[int]pipeKey)
+	}
+	p.pidPipes[pid][fd] = key
+}
+
+// callers must hold pipesMU
+func (p *Tracer) removePipeFD(pid uint32, fd int, key pipeKey) {
+	delete(p.pidPipes[pid], fd)
+	// a held write end would rob readers of EOF once the fd is gone
+	p.fdCache.Remove(procFdPath(pid, fd))
```
MrAlias, line 426: When this owner redirects or exits while an event is still queued, removing the cache entry closes the descriptor warmed by `handleLogEvent`. `handle` then retries only the stale `e.dest`, so the already-intercepted line is dropped even if another registered owner still exposes the pipe.

Could we keep an in-flight reference until the queued write completes, or resolve fresh candidates from `e.pin` in `handle` before dropping?

MrAlias (commented):
The new pipe registration direction is sound, but the redirection lifecycle still has an in-flight ownership gap.

pkg/internal/ebpf/logenricher/logenricher.go
mmat11: done

MrAlias: The retry covers the case where a second owner remains, but the queued event still does not own the descriptor warmed at capture time. If this is the sole owner, `BlockPID` closes it and removes the only candidate before `handle` runs, so the already-intercepted final line is dropped. `removePipeFD` can also close the returned file between `reopenLogDestination` and `Write`. Could we keep an in-flight reference until the queued write completes and cover the sole-owner teardown case?

mmat11: done in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3107/changes/52b64d9b2637abb412dac4fff2c3b2d3648e8477

PR #3108, opened 2026-08-20
fix(metrics)!: derive internal metric labels from OTLP attributes, and fix bpf_map_entries

NimrodAvni78:
## Summary

The bpf and internal metrics carried different label names per exporter: OTLP used semconv-style keys (`bpf.map.id`, `process.executable.name`), the Prometheus exporters hand-wrote their own (`map_id`, `process_name`). Each attribute is now declared once in `pkg/export/attributes/names/attrs.go` and both paths read from it — OTLP uses the key, Prometheus derives the label via the existing `attr.Name.Prom()` — so the two cannot drift. (`obi.version` / `obi.revision` are still built inline in `metrics_net.go` and `metrics_stats.go` for resource attributes; unifying those is a separate change.)

`bpf_map_entries_total` is also fixed while its labels move: it was gauge data typed as a Counter, named `_total`, carrying the map capacity as a `max_entries` label. It becomes the gauge pair `bpf_map_entries` / `bpf_map_max_entries`, mirroring the internal `obi_bpf_map_*` metrics. `Describe()` now advertises both map descs, which it never did while `Collect()` emitted them.

Two labels stay hand-written because they have no OTLP counterpart to derive from: `obi_prometheus_http_requests_total{port, path}` (`PrometheusRequest` is a no-op on the OTLP side), and `obi_internal_build_info`'s const labels — `build_info` is a Prometheus idiom whose labels are its interface, and those names must keep matching `obi_build_info` and `obi_kube_cache_internal_build_info`.

These OTLP keys are declared in `schemas/obi/groups/obi_internal.yaml`, and `TestInternalOTelMetrics` live-checks the emitted `obi.*` surface against the registry, so a namespaced attribute the registry does not declare fails the suite.

## Breaking changes

Prometheus label names, one metric rename (`bpf_map_entries_total`), one OTLP attribute rename (`obi.error` to `error.type`), and one label *value* change (the export-error class, below).

The remaining OTLP keys keep their spelling but are now derived from `attr.VendorPrefix` rather than hardcoded `obi.`, so a build that vendors OBI and overrides the prefix emits its own prefix on these attributes too — consistent with how the internal metric *names* already behave.

| metric | before | after |
|---|---|---|
| `obi_otel_{metric,trace}_export_errors_total` | `error` | `error_type` |
| `obi_instrumented_processes`, `obi_instrumentation_errors_total` | `process_name` | `process_executable_name` |
| `obi_bpf_probe_executions_total`, `obi_bpf_probe_latency_seconds_total` | `probe_id`, `probe_type`, `probe_name` | `bpf_probe_id`, `bpf_probe_type`, `bpf_probe_name` |
| `obi_bpf_map_entries`, `obi_bpf_map_max_entries` | `map_id`, `map_name`, `map_type` | `bpf_map_id`, `bpf_map_name`, `bpf_map_type` |
| `bpf_probe_latency_seconds` | `probe_id`, `probe_type`, `probe_name` | `bpf_probe_id`, `bpf_probe_type`, `bpf_probe_name` |
| `bpf_map_entries_total` (Counter, `max_entries` label) | `map_id`, `map_name`, `map_type`, `max_entries` | `bpf_map_entries` + `bpf_map_max_entries` (Gauges) with `bpf_map_id`, `bpf_map_name`, `bpf_map_type` |

The export-error label is a semantics change, not just a rename. It carried `err.Error()` — the raw message, embedding endpoints and deadlines — so every distinct failure minted a new series. It now carries semconv `error.type`: the canonical gRPC status code string where the error has one, `DEADLINE_EXCEEDED` / `CANCELLED` for context and timeout failures, the type of the underlying `net.Error` for transport failures, and `_OTHER` for anything else. Every check goes through `errors.Is` / `errors.As`, so it sees past the wrapping the OTLP exporters add; an error matching none of them is bucketed rather than reported by Go type, because the type of an unrecognized error names the wrapper that carried it and not the failure. `pkg/internal/errtype` owns the code spelling and the `_OTHER` fallback so the span pipeline and the internal metrics cannot spell the same condition two ways; `request.GRPCStatusCodeString` and `request.ErrorTypeOther` now delegate to it. This drops the OBI-invented `obi.error` attribute (including from the registry) in favour of the semconv attribute `obi_instrumentation_errors_total` already uses; the transformation is recorded under "Pending transformations" in `site/README.md` for release prep.

Unchanged because they already matched: `subscriber`, `service_name`, `service_namespace`, `telemetry_type`, and `otel_metric_overflow` — the last is pinned to `otel.metric.overflow` by a test rather than derived at init, so it stays a compile-time constant.

Also in the diff: three new suffix constants (`VendorGoarchSuffix`, `VendorGoosSuffix`, `VendorGoversionSuffix`) for the build-info attributes, and two corrected `Help` strings on the map and instrumentation-error metrics.

## Validation


pkg/export/imetrics/errortype.go
```diff
@@ -0,0 +1,71 @@
...
+		if joined, ok := err.(interface{ Unwrap() []error }); ok {
+			errs := joined.Unwrap()
+			if len(errs) == 0 || errs[0] == nil {
+				return err
+			}
+			err = errs[0]
+			continue
+		}
+
+		if fmt.Sprintf("%T", err) != "*fmt.wrapError" {
```
mmat11: is there a better way to compare this error?

pkg/export/imetrics/errortype.go
```diff
@@ -0,0 +1,71 @@
...
+// ExportErrorType classifies an export failure into a low-cardinality error.type value.
+// The error message is unsuitable as an attribute: it embeds endpoints and deadlines, so
+// it mints a new series per distinct failure.
+func ExportErrorType(err error) string {
+	if err == nil {
+		return errtype.Other
+	}
+
+	if st, ok := status.FromError(err); ok && st.Code() != codes.OK {
+		if code := errtype.GRPCCode(int(st.Code())); code != "" {
```
mmat11, line 31: question: is this function used for grpc errors only?

mmat11 (approved):
lgtm, I see this contains breaking changes -- should we keep this open until we release v0.12.3 so this can be released with v0.13.0?

skl (approved):
lgtm, ty!

pkg/export/imetrics/errortype.go
NimrodAvni78: grpc and HTTP, so the fallback is HTTP

PR #3112, opened 2026-08-20
python: extract service metadata for known frameworks

grcevski:
## Summary

Similar pr to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3099 and https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3089. This change now extracts metadata from various typical sources for Python service names. Mainly it relies on `pyproject.toml` but it also provides command line extractor for various popular frameworks, like: django, gunicorn, uvicorn, flask, fastapi, celery etc.

Reguires https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3099 to be merged first because I built on top of it.

Full support added for:
- gunicorn
- uvicorn
- hypercorn
- daphne (used in django)
- uwsgi
- waitress
- flask
- fastapi
- django
- celery

Some of the changes I made are refactor from java and nodejs to reuse the directory walking code.

## Validation


pkg/internal/pythontools/launch.go
```diff
@@ -0,0 +1,1039 @@
...
+			return pythonLaunch{target: args[i+1], targetKind: targetFile}
+		case strings.HasPrefix(arg, "--wsgi-file="):
+			return pythonLaunch{target: strings.TrimPrefix(arg, "--wsgi-file="), targetKind: targetFile}
+		}
+	}
+	return pythonLaunch{}
+}
+
+func parseFlask(args []string, env map[string]string) pythonLaunch {
+	target := lastLongOptionValue(args, "--app", env["FLASK_APP"])
```
marctc: Flask can also be launched with `-A` as alias for `--app`. It also works sumply with `flask run` but I'm not sure if we can do much in this case.

marctc (approved):
LGTM, left a minor comment!

pkg/internal/pythontools/launch.go
```diff
@@ -0,0 +1,1039 @@
...
+				return false
+			}
+		} else if r != '_' && !unicode.IsLetter(r) && !unicode.IsDigit(r) {
+			return false
+		}
+	}
+	return value != ""
+}
+
+func firstApplicationReference(values []string) string {
```
MrAlias: `firstApplicationReference` applies one `module:object` grammar to both Gunicorn and Hypercorn, but neither launcher is that narrow. Gunicorn accepts `company.orders.wsgi` and defaults the object to `application` ([loader](https://github.com/benoitc/gunicorn/blob/26.2.0/gunicorn/util.py#L412-L420)); Hypercorn accepts a no-colon target with default `app`, file targets such as `src/api.py:app`, and `asgi:`/`wsgi:` prefixes ([loader](https://github.com/pgjones/hypercorn/blob/0.18.0/src/hypercorn/utils.py#L81-L110)). All of those valid commands currently leave the launch without a target. Could we keep the option scanning shared but parse each launcher's application spec separately, with table cases for these forms?

pkg/internal/pythontools/launch.go
```diff
@@ -0,0 +1,1039 @@
...
+	}
+	positionals, ok := gunicornPositionals(args)
+	if !ok {
+		return pythonLaunch{}
+	}
+
+	applyGunicornSettings(fields, &settings)
+	applyGunicornSettings(args, &settings)
+
+	launch := pythonLaunch{fallbackName: cleanValue(settings.name)}
```
MrAlias: `searchPaths` keeps chdir first and appends `--pythonpath` in command-line order. Gunicorn instead establishes chdir and then inserts each path at `sys.path[0]`, so `--pythonpath=/one,/two` searches `/two`, `/one`, then chdir ([chdir](https://github.com/benoitc/gunicorn/blob/26.2.0/gunicorn/app/base.py#L83-L90), [pythonpath](https://github.com/benoitc/gunicorn/blob/26.2.0/gunicorn/app/base.py#L227-L233)). Duplicate module names can therefore select the wrong project's metadata; `python -I -m gunicorn` also leaves the resolver without Gunicorn's default cwd. Could we mirror the effective order and cover both cases in `TestParsePythonLaunch`?

pkg/internal/pythontools/launch.go
```diff
@@ -0,0 +1,1039 @@
...
+	}
+	return positionals[0], true
+}
+
+func parseUWSGI(args []string) pythonLaunch {
+	for i := 0; i < len(args); i++ {
+		arg := args[i]
+		switch {
+		case (arg == "-w" || arg == "--module") && i+1 < len(args):
+			return pythonLaunch{target: args[i+1], targetKind: targetModule}
```
MrAlias: `parseUWSGI` returns on the first recognized target, so it cannot reproduce the effective launch configuration. uWSGI also accepts `--wsgi` as a module alias, `--file` as a file alias, and `--pythonpath`/`--python-path`/`--pp` as import roots ([module aliases](https://uwsgi-docs.readthedocs.io/en/latest/Options.html#wsgi), [file alias](https://uwsgi-docs.readthedocs.io/en/latest/Options.html#file), [path options](https://uwsgi-docs.readthedocs.io/en/latest/Options.html#pythonpath)). The aliases currently produce no target, while a module reachable only through a path option loses its project metadata. Could we collect the relevant settings before constructing `pythonLaunch`, preserving uWSGI's effective path precedence, and cover each alias plus repeated paths?

pkg/internal/pythontools/launch.go
```diff
@@ -0,0 +1,1039 @@
...
+			}
+			continue
+		}
+		parsed.positionals = append(parsed.positionals, arg)
+	}
+	return parsed, true
+}
+
+func parseDjango(args []string, env map[string]string) pythonLaunch {
+	target := lastLongOptionValue(args, "--settings", env["DJANGO_SETTINGS_MODULE"])
```
MrAlias: `parseDjango` keeps only the settings target and loses two import roots. Django prepends its global `--pythonpath` value ([Django](https://github.com/django/django/blob/6.0/django/core/management/base.py#L92-L95)); when Python runs `/srv/orders/manage.py`, it also prepends `/srv/orders` unless safe-path mode is active ([Python](https://docs.python.org/3/using/cmdline.html#interface-options)). From another cwd, either root can be the only way to import the settings package, so the declared project metadata is missed. Could we retain the last `--pythonpath` and the `manage.py` directory when safe-path mode permits, with regression cases for both?

pkg/internal/pythontools/launch.go
```diff
@@ -0,0 +1,1039 @@
...
+		return parseUWSGI(args), true
+	case "waitress", "waitress-serve", "waitress_serve":
+		return parseWaitress(args), true
+	case "flask":
+		return parseFlask(args, env), true
+	case "fastapi":
+		return parseFastAPI(args), true
+	case "django", "django-admin", "django_admin":
+		return parseDjango(args, env), true
+	case "celery":
```
MrAlias, line 171: The dispatcher still has the process environment here, but `parseCelery` cannot see it. Celery uses `CELERY_APP` as the environment default for `-A` ([CLI documentation](https://docs.celeryq.dev/en/stable/reference/cli.html#celery)); `CELERY_APP=company.orders celery worker` therefore yields no target or project metadata. Could we initialize from the environment and let command-line `-A`/`--app` override it, with tests for the environment-only and override cases?

MrAlias (commented):
The separation between launcher parsing and metadata resolution is a good boundary, and the shared file helpers keep the Java and Node changes contained. The current adapters still flatten framework-specific target and import-path semantics, so supported launch commands can resolve the wrong project or produce no declared metadata.

pkg/internal/pythontools/launch.go
grcevski: Ah I didn't know that, let le see if I can cover it. I think I can.

grcevski:
@marctc and @MrAlias  I believe I've addressed all of the comments. I added the changes commit by commit.

MrAlias (approved):

PR #3122, opened 2026-08-21
Limit Kafka Produce topic fan-out

MrAlias:
### Motivation

- The Produce request parser previously trusted the untrusted declared topic count and appended every parsed topic, enabling attacker-controlled per-request fan-out that can exhaust memory and CPU.
- A small per-topic encoding can encode millions of topics in a single Kafka packet, so the parser must bound work and downstream allocations to preserve availability.

### Description

- Add a `const maxProduceTopics = 100` and enforce it in `pkg/internal/ebpf/kafkaparser/produce.go` by capping the parsed `topicsLen` before the parsing loop. 
- Keep existing parsing semantics for each topic; only the number of topics parsed per request is limited. 
- Add `TestParseProduceRequestLimitsTopics` in `pkg/internal/ebpf/kafkaparser/produce_test.go` to exercise an attacker-controlled large declared topic count and assert that only `maxProduceTopics` topics are returned.

### Testing

- `go test ./pkg/internal/ebpf/kafkaparser`
- `git diff --check`
- `make verify`

### DIFF
--- a/pkg/internal/ebpf/kafkaparser/produce.go
+++ b/pkg/internal/ebpf/kafkaparser/produce.go
@@ -19,6 +19,8 @@ type ProduceRequest struct {
 	Topics []*ProduceTopic
 }
 
+const maxProduceTopics = 100
+
 var (
 	errInvalidProduceRecordsLength = errors.New("invalid produce records length")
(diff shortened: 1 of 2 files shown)


grcevski (approved):
LGTM!

PR #3123, opened 2026-08-21
Reject split span metric format conflicts

MrAlias:
## Summary

- Track legacy and OTel span metric formats across configuration scopes after resolving `all` and `*`
- Reject split legacy/OTel selections even when their joined mask reconstructs `FeatureAll`
- Cover both top-level and per-service orderings

Follow-up to the non-blocking review note on #2979.

## Testing

- make generate
- make verify
- make compile


grcevski (approved):
LGTM!

PR #3124, opened 2026-08-21
Fix Go netFD read map cleanup

MrAlias:
### Motivation

- Prevent a resource-exhaustion bug where skipped Go `netFD` reads inserted into the fixed-size `ongoing_fd_reads` BPF hash map were not deleted by the read-return probe, allowing an attacker to fill the map and disrupt tracing. 

### Description

- In `obi_uprobe_netFdReadRet`, emit optional HTTP large-buffer data as before but then delete the `ongoing_fd_reads` entry for the goroutine so skipped reads do not leak. 
- Adjusted the large-buffer emission condition to only call `send_http_large_buffers_if_needed` when `http_large_buffer_skip(len)` is false and a valid `net_ptr->byte_ptr` exists. 
- This preserves large-buffer capture behavior while ensuring the pinned `BPF_MAP_TYPE_HASH` does not accumulate stale skip entries. 

### Testing

- `make generate`
- `make clang-format`
- `git diff --check`
- `make clang-tidy`


grcevski (approved):
LGTM!

PR #3125, opened 2026-08-21
Fix log enricher buffer suppression order

MrAlias:
### Motivation

- The BPF log enricher previously overwrote application write buffers before confirming the enriched event was accepted by the ring buffer, which could silently drop application logs on `bpf_ringbuf_output` failure.
- The change restores a fail-open fallback by only mutating user memory after the ring-buffer enqueue succeeds.

### Description

- Split read vs suppression logic by introducing `copy_ubuf`/`copy_iovec` to copy user bytes into the event payload without touching user memory, and `suppress_ubuf`/`suppress_iovec` to perform `bpf_probe_write_user` only when appropriate.
- Update `__write` to call the copy helpers, call `bpf_ringbuf_output`, return immediately on error, and only call the suppress helpers after a successful ring-buffer submission.
- Preserve existing limits and safety checks (`bpf_clamp_umax`, segment/length caps, newline termination) and keep original enrichment payload layout and sizing logic.

### Testing

- `clang-format -i bpf/logenricher/logenricher.c`
- `cd bpf && clang-tidy logenricher/logenricher.c`
- `make generate`
- `make compile`
- `go test ./pkg/internal/ebpf/logenricher`
- `go test -tags=bpf_verifier_tests -run 'TestBPFVerifierWithConstants/logenricher' ./pkg/internal/ebpf/verifier`

### DIFF
--- a/bpf/logenricher/logenricher.c
+++ b/bpf/logenricher/logenricher.c
@@ -47,10 +47,7 @@ static __always_inline bool pid_tracked(const struct task_struct *task) {
     return tracked != NULL;
 }
 
-static __always_inline u32 consume_ubuf(log_event_t *e,
-                                        struct iov_iter *from,
-                                        void *ubuf,
-                                        const char *fill) {
+static __always_inline u32 copy_ubuf(log_event_t *e, struct iov_iter *from, void *ubuf) {
     const size_t count = BPF_CORE_READ(from, count);
     u32 to_copy = (u32)count;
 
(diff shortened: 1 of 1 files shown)


skl (approved):
lgtm

mmat11:
github isn't picking up slash commands 🙃

MrAlias:
It looks like CI is pretty backed up. The run is queued FWIW: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/32756173719

mmat11 (approved):

PR #3126, opened 2026-08-21
Limit Java TLS buffer allocation to payload size

MrAlias:
### Motivation

- The Java TLS extraction path previously always allocated a `MAX_SIZE` (16 KiB) backing array for flattened fresh `ByteBuffer[]` inputs, which increased worst-case retained plaintext when buffers were stored in `SSLStorage` and introduced a memory exhaustion risk. 
- The goal is to avoid allocating large arrays when the actual payload is much smaller while preserving the existing `MAX_SIZE` cap for legitimately large payloads.

### Description

- Compute the required `bufSize` by summing the `remaining()` bytes from the source buffers up to `MAX_SIZE` and allocate the destination `ByteBuffer` to that size instead of always allocating `MAX_SIZE` in `flattenFreshByteBufferArray`. 
- Keep the existing `MAX_SIZE` cap so large payloads still truncate to the current maximum. 
- Update unit tests in `ByteBufferExtractorTest` to assert zero-capacity for null/empty/all-null fresh-array cases and to assert capacity equals the actual copied payload size for small inputs. 

### Testing

- `make java-spotless-apply`
- `cd pkg/internal/java && gradle :agent:test --tests io.opentelemetry.obi.java.instrumentations.util.ByteBufferExtractorTest -PnativeOnly=true`
- `make java-test`


grcevski (approved):
Nice! LGTM!

PR #3127, opened 2026-08-21
Keep clang-tidy compatible with LLVM 22

MrAlias:
## Summary

- exclude LLVM 22 style checks that conflict with established eBPF conventions
- preserve externally discoverable eBPF program and map symbols
- guard `memchar_pos` against null pointer arithmetic

## Motivation

The `readability-*` and `misc-*` check groups automatically pick up checks added or expanded by newer LLVM releases. With LLVM 22.1.8, `make clang-tidy` reports 580 errors that do not occur with the repository's current LLVM 18 tooling.

Most of those diagnostics come from five style checks that were not part of the lint policy when the broad check groups were introduced. In particular, `misc-use-internal-linkage` recommends making eBPF programs and maps static even though loaders discover those symbols externally. The boolean-conversion and parentheses checks similarly flag established kernel and verifier-oriented C patterns without identifying behavioral defects.

Explicitly excluding those checks keeps the existing lint policy stable across LLVM versions. The remaining analyzer diagnostic identifies possible pointer arithmetic on a null `haystack`; checking it before entering the scan removes that undefined behavior.

## Compatibility and performance

The lint exclusions do not affect generated code. Older clang-tidy versions tolerate the negative check entries, including names they do not implement.

For non-null buffers, `memchar_pos` retains its existing behavior. A null buffer now returns the existing `INVALID_POS` sentinel before pointer arithmetic. The added inline guard has negligible runtime cost and may be optimized away when the caller proves the pointer is non-null. This change does not affect the ABI, eBPF map layouts, or minimum supported kernel.

## Testing

- `make clang-tidy` (LLVM 18.1.3, all 250 files)
- `clang-format --dry-run --Werror bpf/common/tc_common.h`
- `clang-tidy bpf/common/tc_common.h`
- `git diff --check`


grcevski (approved):
LGTM!

PR #3130, opened 2026-08-21
logenricher: serialize cache publication with retirement

MrAlias:
## Summary

- serialize registered pipe cache publication with pipe retirement
- retain event ownership without restoring a cache reference after teardown
- add unprivileged regression coverage for post-retirement publication and EOF

## Context

#3107 made queued events own their destination descriptors until the write completes. A cache miss could still race with owner retirement: teardown could remove an absent cache key immediately before the capture path published its newly opened descriptor. The event reference would be released after the write, but the late cache reference would keep the pipe writer open until eviction or the 30-minute TTL.

This change publishes registered pipe destinations while holding the same lifecycle lock used by retirement. If the destination was retired while it was opening, the event reference remains valid for the captured write, but no cache reference is created.

## Validation

- `make generate`
- `go test -race ./pkg/internal/ebpf/logenricher`
- `go test -race -run TestRetiredPipeDestinationIsNotCached -count=20 -v ./pkg/internal/ebpf/logenricher`

### DIFF
--- a/pkg/internal/ebpf/logenricher/logenricher.go
+++ b/pkg/internal/ebpf/logenricher/logenricher.go
@@ -505,18 +505,19 @@ func (p *Tracer) pipeRegistered(key pipeKey) bool {
 
 // identity to pin the tty fallback destination with; rejects an unregistered
 // pipe (app IPC)
-func (p *Tracer) fallbackDest(path string) (pipeKey, bool) {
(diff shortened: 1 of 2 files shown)


grcevski (approved):
LGTM! This looks solid, I think it should be in 0.12.2

NimrodAvni78 (approved):

grcevski:
We have second approval, I'm going to merge and this will land in 0.12.2

Issue #2940, opened 2026-08-06
`v8js.*` runtime metrics (GC, heap, heap spaces)


pinoOgni:
### Component

OBI (eBPF instrumentation)

### Problem

Add support for the `v8js.*` runtime metrics (GC, heap, heap spaces) described here https://opentelemetry.io/docs/specs/semconv/runtime/v8js-metrics/

PR #3131, opened 2026-08-21
Add Nodejs `v8js.*` runtime metrics (GC duration and heap spaces)

pinoOgni:
## Summary

Part of #2940 

This PR adds the semconv `v8js.*` metric set (GC duration and heap spaces), 

It uses the already injected agent to observe GC cycles using a `PerformanceObserver` (one histogram observation per collection) and walks `v8.getHeapSpaceStatistics()` on its existing 1s tick, reporting (as already done for `nodejs.eventloop`) over the `uv_fs_access` side channel. 

Notes:
- Everything rides the existing machinery: same `application_runtime` feature and per-service gating, same BPF constant, same runtime-metrics queue and exporters; only a new optional `buckets.v8js_gc_duration_histogram` override.
- All metrics work from the existing Nodejs 14.10 floor: the GC kind is read from `PerformanceNodeEntry.detail` (16+) with a fallback to the pre-16 entry.kind accessor.

Limitations / possible follow-ups:
- Only the well-known `v8js.heap.space.name` values are exported (`new_space`, `old_space`, `code_space`, `map_space`, `large_object_space`). Semconv allows custom values, but the extra spaces V8 reports are engine-version-dependent (for example on my machine with Node 26 I have 13 spaces like `code_large_object_space`, `shared_large_object_space`, `read_only_space` and so on)  and the weaver validation flags them, so they are dropped before export. If we ever want them, we may propose the missing members to semconv, the OBI change is then trivial.

Follow-up:
- `v8js.resource.active` is intentionally out of scope in this PR (different source, no GC/heap relation).

I recommend reviewing it commit by commit.

## Validation


MrAlias (approved):

bpf/generictracer/nodejs.c
```diff
@@ -224,6 +279,115 @@ static __always_inline int handle_runtime_metrics(const char *path, const u64 pi
...
+    if (!payload) {
+        return 0;
+    }
+    const long len =
+        bpf_probe_read_user_str(payload, k_v8_heap_payload_read_len, path + k_v8_payload_offset);
+    if (len <= 0) {
+        return 0;
+    }
+
+    if (kind_char == 'g') {
```
marctc: a comment and/or constant here and in line 352 would be great

marctc (approved):
LGTM, great stuff! just added minor comment

bpf/generictracer/nodejs.c
pinoOgni: Thanks!

PR #3132, opened 2026-08-21
deno: extract service metadata of deno apps

grcevski:
## Summary

This is a follow-up of https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3099, where we do the same thing but for Deno JavaScript apps. I also fixed route harvesting for deno, where command like `deno run` would have been incorrectly interpreted as `node run` and we'd think run is run.js.

## Validation


pkg/internal/denotools/service_metadata.go
```diff
@@ -0,0 +1,469 @@
...
+	return metadata
+}
+
+func metadataMatches(metadata serviceMetadata, requireName bool) bool {
+	if requireName {
+		_, _, ok := parseProjectName(metadata.Name)
+		return ok
+	}
+	return strings.TrimSpace(metadata.Version) != ""
+}
```
skl, line 312: Given an explicitly configured service name:

```
OTEL_SERVICE_NAME=worker
```

and the following directory structure:

```
  /app/deno.json      {"name":"outer","version":"1.0.0"}
  /app/src/deno.json  {"name":"worker"}
  /app/src/main.ts
```

`/app/src/deno.json` does not match because it has no `version`. Lookup continues to `/app/deno.json`, and `1.0.0` is incorrectly attached to the explicitly named service.

The tests already establish that a nearer named project must not borrow an outer version:

```go
t.Run("nearer named config does not borrow an outer version", func(t *testing.T) {
```

- [service_metadata_test.go:72-84](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3132/changes#diff-2b0ae2b074f6e764176fc1919b971c8a9b9c19a80cf41fa5fbf5cffa8dd63030R72)

pkg/internal/transform/route/harvest/js.go
```diff
@@ -1588,6 +1589,18 @@ var (
...
+	return findJSAppDir(pid, func(args []string) string {
+		return nodejstools.ParseNodeLaunch(args).EntryPoint
+	})
+}
+
+func FindDenoAppDir(pid app.PID) (string, error) {
+	return findJSAppDir(pid, func(args []string) string {
+		return denotools.ParseDenoLaunch(args).EntryPoint
+	})
+}
```
skl, line 1610: Entrypoints with protocol/scheme prefixes (supported by `deno run`, see docs below) result in the working directory being scanned instead of the app root directory.

Given:

```shell
deno run file:///opt/app/main.ts
```

1. `ParseDenoLaunch` stores the entrypoint unchanged:
  • `pkg/internal/denotools/launch.go:326-328`
  • The test at `launch_test.go:24` explicitly expects the raw `file:///...` string.
2. `FindDenoAppDir` returns that raw value to `findJSAppDir`:
  • `pkg/internal/transform/route/harvest/js.go:1597-1600`
3. `findJSAppDir` passes it directly to `FindScriptDirectory`:
  • `js.go:1614-1616`
4. `FindScriptDirectory` handles an entrypoint specially only if it begins with `/`:
  • `harvester.go:197-212`
5. Since `file:///opt/app/main.ts` begins with `file:`, it falls through to:
  • `harvester.go:214-219`
  • This returns root + cwd, not root + /opt/app.
6. `ExtractDenoRoutes` then scans that directory:
  • `js.go:1691-1696`
  • `js.go:1708`


The same fallback occurs for `https://...`, `jsr:...`, and `npm:...`, all of which ParseDenoLaunch intentionally preserves as raw entrypoint specifiers.

For those entrypoints, the working directory does not identify the entrypoint source, so scanning it can produce unrelated routes.

Docs:

- https://docs.deno.com/runtime/run/#run-from-a-url-or-stdin
- https://docs.deno.com/runtime/reference/cli/run/#terminate-run

skl (approved):
lgtm, couple of edge cases (could be follow-up)

pkg/internal/denotools/service_metadata.go
grcevski: Nice catch! I'll take a look and fix this.

grcevski: I think this isn't a problem because if the OTEL_SERVICE_NAME is set, we'll have the inner UID.Name set on the service before we come to resolve. I added this test and it passed. I think it describes what you had in mind. FileInfo should've parsed any env variables by the time we reach this resolve code:

```
	t.Run("nearer named config does not borrow an outer version", func(t *testing.T) {
		root := t.TempDir()
		writeDenoFile(t, filepath.Join(root, "app", "deno.json"), []byte(`{"name":"outer","version":"1.0.0"}`))
		writeDenoFile(t, filepath.Join(root, "app", "src", "deno.json"), []byte(`{"name":"worker"}`))
		writeDenoFile(t, filepath.Join(root, "app", "src", "main.ts"), nil)
		fileInfo := mockDenoProcess(t, root, []string{"run", "src/main.ts"}, nil)
		fileInfo.SetAutoServiceName("my-service")

		err := ResolveServiceMetadata(fileInfo)

		require.NoError(t, err)
		assert.Equal(t, "my-service", fileInfo.ServiceAttrs().UID.Name)
		assert.Empty(t, fileInfo.ServiceAttrs().Metadata[serviceVersion])
	})
```

The reason is that we check `	resolveName := service.UID.Name == "` which will be false in this case and then we'll try to find the version, which doesn't exist in the preferred src/deno.json, so we come out without version.

grcevski: No, wait, I'm wrong. My test doesn't work.

pkg/internal/transform/route/harvest/js.go
grcevski: Good find, I thought initially you wanted me to support these extra prefixes, but what you are saying is we should special case file, because that's legit.

pkg/internal/denotools/service_metadata.go
grcevski: Fixed

pkg/internal/transform/route/harvest/js.go
grcevski: Fixed.

PR #3135, opened 2026-08-22
feat(java): report incompatible preloaded agents

### LINKED ISSUE #913: Add ability to report error on loading the Java agent on version mismatch
We might create incompatible new OBI version that will not work with the old Java OBI agent. When we fail to load the new OBI agent, because of pre-existing agent, we log an appropriate error. 

One way of achieving this would be to look up a version string for the Java agent once we try to load the new agent, find if a prior version is loaded (we do this already), read the version string and report and error if the version is incompatible.

salman-frs:
## Summary

A JVM cannot replace a Java agent while the process is running. OBI currently
skips injection when it finds its agent already loaded, even if that agent came
from an incompatible OBI release.

This change adds a compatibility version to the Java agent. The agent publishes
the version through the `otel.obi.java.agent.version` system property. When OBI
finds an existing agent, it reads the property with `jcmd VM.system_properties`
and compares it with the version in its embedded agent JAR.

A matching version keeps the existing behavior and skips reinjection. A missing
or different version returns a `JavaInjectError` and logs both the loaded and
expected versions so the operator can restart the process.

Both OBI and the Java agent read the version from
`obi-java-agent-version.properties`. The Java agent README documents when this
compatibility version must change.

The committed `embedded/obi-java-agent.jar` is unchanged. The build and release
targets rebuild it from source, while committing it here would add a 5 MB binary
diff.

Fixes #913.

## Validation

- `go build ./...`
- `go test ./pkg/internal/java/... ./pkg/appolly/discover/...`
- `go tool -modfile=internal/tools/go.mod golangci-lint run ./pkg/internal/java/... --timeout=6m`
- `gradle build`
- `make lint-markdown`
- Manual end-to-end testing with HotSpot JDK 8 and JDK 21:
  - inject an agent into a clean JVM
  - skip reinjection when the versions match
  - report an error when the versions differ
  - report an old agent without a version as `unknown`

### DIFF
--- a/pkg/internal/java/README.md
+++ b/pkg/internal/java/README.md
@@ -135,6 +135,20 @@ p.getInt(1 + 36)       → 1    (Buffer length at offset 37)
 p.getByte(1 + 36 + 4)  → 42   (Data byte at offset 41)
 ```
 
+### 4. Java agent compatibility
+
+A JVM cannot unload or replace a Java agent. The agent therefore publishes its version in the
+`otel.obi.java.agent.version` system property. OBI reads this property with
+`jcmd VM.system_properties` and compares it with the version in the embedded agent JAR.
+
+If the versions do not match, OBI reports an error and does not instrument the process. Restart
+the process to load the agent included with the current OBI version.
+
+The version is stored in
+`agent/src/main/resources/obi-java-agent-version.properties`. Increment it when a change breaks
+compatibility between OBI and the Java agent. Do not increment it for compatible changes, such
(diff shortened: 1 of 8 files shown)


MrAlias (approved):

salman-frs:
Changed `NewExecutable` to take an `InjectionTarget`, so the version check now uses `target.Pid` instead of `ie`. It wouldn't have compiled after merge otherwise.

grcevski (approved):
LGTM!

PR #3136, opened 2026-08-23
Conform to OTEL messaging semantic conventions

NimrodAvni78:
## Summary

Brings OBI's messaging spans in line with the messaging semantic conventions:

- **Emits `messaging.operation.name` on the Kafka / MQTT / NATS / AMQP trace path**, which the conventions mark `Required` on every messaging span. It was already a default-on attribute and already emitted on AWS SQS spans; the four broker protocols were the gap.
- **Stops emitting the deprecated `publish` operation type.** Semconv v1.41 deprecates `publish` in favour of `send`, so `messaging.operation.type` now reports `send` where it previously reported `publish`.

The two attributes carry different information, so they are now derived separately: `operation.name` is the system's own name for the call, while `operation.type` is the five-member semconv enum used to aggregate across brokers. They coincide for Kafka and for the consume side, and differ for MQTT / NATS / AMQP, which name the call `publish` but whose type is `send`. This mirrors how the Java and JS instrumentations derive the pair.

| System | `operation.name` | `operation.type` |
| --- | --- | --- |
| Kafka producer | `send` | `send` |
| MQTT / NATS / AMQP producer | `publish` | `send` |
| consumers | `process` | `process` |

`publish` is still a *declared* member of `messaging.operation.type` in v1.41, marked `deprecated: {renamed_to: send}`, so emitting it was never a live-check violation and nothing was failing. This is spec hygiene ahead of the deprecated member eventually being removed, not a fix for a broken build.

## Breaking changes

Five surfaces are visible to anything consuming OBI's messaging telemetry:

- `messaging.operation.type` reports `send` instead of `publish` on producer spans.
- Kafka producer span names change from `publish {topic}` to `send {topic}`, following the operation name. MQTT, NATS and AMQP span names are unchanged.
- The `messaging.operation.name` label on `messaging.client.operation.duration` changes from `publish` to `send` for Kafka producers, re-keying those series. MQTT, NATS and AMQP keep `publish`.
- The debug / JSON exporter prints `operation: send` for Kafka spans where it printed `operation: publish` (`spanAttributes()` surfaces `s.Method` directly).
- Span metrics re-key: `traces.span.metrics.calls` / `.duration` are grouped by `span.name`, which `spanMetricAttributes()` emits unconditionally (span metrics have no attribute selection). Every Kafka producer series moves from `span_name="publish {topic}"` to `span_name="send {topic}"`.

Dashboards, alerts or saved queries that match the old values need updating.

## Notes for reviewers

- Producer-side dispatch goes through `request.MessagingOperationTypeOf(span.Method)` rather than matching `send` and `publish` side by side. The normalisation lives in one function, so each call site matches the single canonical `send`, and adding a protocol that names the call something else does not mean touching every switch.
- `pkg/ebpf/common/http/aws_sqs.go` swaps its bare `"send"` / `"receive"` literals for the `request.MessagingSend` / `request.MessagingReceive` constants (the latter added for it). Behaviour-neutral, and included so the SQS path names these values the same way the code this PR touches does.
- Kafka consumer spans keep `operation.type: process`. The conventions model a broker fetch as a Receive span (`type: receive`, name `poll`), which is what the Java instrumentation emits; changing that also changes span kind mapping, so it is left for a separate change.

## Validation

### DIFF
--- a/internal/test/integration/red_test_kafka.go
+++ b/internal/test/integration/red_test_kafka.go
@@ -87,10 +87,10 @@ func testREDMetricsPythonKafkaOnly(t *testing.T) {
 			Comm:    "python3.14",
 			Spans: []TestCaseSpan{
 				{
-					Name: "publish my-topic",
+					Name: "send my-topic",
 					Attributes: []attribute.KeyValue{
 						attribute.String("span.kind", "producer"),
-						attribute.String("messaging.operation.type", "publish"),
+						attribute.String("messaging.operation.type", "send"),
 						attribute.String("messaging.destination.name", "my-topic"),
 						attribute.String("messaging.client.id", "kafka-python-producer-1"),
 						attribute.String("messaging.destination.partition.id", "0"),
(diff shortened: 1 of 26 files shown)


skl (approved):
lgtm

MrAlias (approved):

PR #3145, opened 2026-08-24
ruby: extract service metadata for rails applications

grcevski:
## Summary

This is the Ruby version of the similar PRs I've been making to extract meaningful service metadata from applications, when there's none available from higher levels, like k8s, cloud or docker. Needs https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3099 to be merged before.

I've added an AST based grammar parser. Initially I had LLM generated gemspec and ruby file parsers when I was working on the draft, but it would be bad idea to keep them, there are many edge cases and good luck with reviewing them. Fortunately there's a new pure Go based AST project and I was able to use it. I added makefile switches to reduce the import impact on our binary, since it has parsers for many languages. 

We can use this approach for more route harvesters and not limit to what we have today, i.e. Java and nodejs.

## Validation


pkg/appolly/discover/attacher.go
```diff
@@ -143,7 +147,7 @@ func (ta *traceAttacher) attacherLoop(_ context.Context) (swarm.RunFunc, error)
 					"exec", instr.Obj.FileInfo.CmdExePath(), "pid", instr.Obj.FileInfo.Pid())
 				switch instr.Type {
 				case EventCreated:
-					ta.resolveServiceMetadata(&instr.Obj)
+					ta.resolveServiceMetadata(ctx, &instr.Obj)
```
MrAlias, line 150: `resolveServiceMetadata` runs before `getTracer` in the single `ForEachInput` loop, so one Ruby lookup that reaches this timeout prevents every process behind it from being attached or deleted for five seconds. `ResolveServiceMetadata` also returns without joining its filesystem goroutine, allowing timed-out work to accumulate. Could we keep metadata-before-attachment ordering without blocking the shared loop, and cover a stalled Ruby lookup followed by an unrelated process event?

Makefile
```diff
@@ -16,6 +16,10 @@ RELEASE_REVISION ?= $(shell git rev-parse --short HEAD )
 BUILDINFO_PKG ?= go.opentelemetry.io/obi/pkg/buildinfo
 TEST_OUTPUT ?= ./testoutput
 RELEASE_DIR ?= ./dist
+# RUBY_GRAMMAR_TAGS are needed to reduce the gotreesitter dependency we have for parsing Ruby files.
+# We can use gotreesitter for other languages in the future, so we can expand these here. ATM, we use
+# the Ruby parser for determining the application name, but we can use it for route harvesting as well.
+RUBY_GRAMMAR_TAGS := grammar_subset,grammar_subset_ruby
```
MrAlias, line 22: `RUBY_GRAMMAR_TAGS` only affects these Make targets. The supported Collector paths still run `go build ./collector/...` and `ocb` without the tags, so those distributions embed gotreesitter's default all-grammar blob—about 16 MiB more than the Ruby-only probe. Could we make the subset part of the supported Collector build path and add a size regression, rather than relying on each downstream consumer to discover the required `GOFLAGS`?

MrAlias (commented):
The metadata extraction is well-scoped, but its attachment-loop placement and the Collector build footprint still need to be resolved.

pkg/internal/rubytools/ruby_tree.go
```diff
@@ -0,0 +1,134 @@
...
+
+type rubySyntaxTree struct {
+	source   []byte
+	language *gotreesitter.Language
+	tree     *gotreesitter.Tree
+}
+
+func parseRubyTree(source []byte) *rubySyntaxTree {
+	language := grammars.RubyLanguage()
+	tree, err := gotreesitter.NewParser(language).ParseStrict(source)
```
mmat11, line 19: claude says:

```ParseStrict takes no context, so the 5s attacher timeout only makes ResolveServiceMetadata return while the parse goroutine keeps running — a benign 480 KB Ruby file (a = b repeated) takes 110s and 306 MB here, and parse time is quadratic, so the 2 MiB maxRubyMetadataBytes cap allows roughly 35 minutes of uncancellable CPU per discovered process; please wire SetTimeoutMicros/SetCancellationFlag to the context and lower the byte cap.```

File: pkg/internal/rubytools/ruby_tree.go [hunk 4]
```diff
@@ -0,0 +1,134 @@
...
+)
+
+type rubySyntaxTree struct {
+	source   []byte
+	language *gotreesitter.Language
+	tree     *gotreesitter.Tree
+}
+
+func parseRubyTree(source []byte) *rubySyntaxTree {
+	language := grammars.RubyLanguage()
```
mmat11, line 18: also claude:

`grammars.RubyLanguage() panics (not errors) when the binary is built with grammar_subset but without grammar_subset_ruby — verified: gotreesitter: failed to load grammar "ruby.bin": ... not embedded in this grammar_subset build — and there is no recover on this path, so a downstream subsetting for a different language crashes the agent on its first Ruby process; suggest a sync.OnceValue with recover that degrades to skipping Ruby metadata.`

pkg/appolly/discover/attacher.go
grcevski: I think this is possible, but it will require a large refactor of how we process metadata. I would only do this in a follow-up.

grcevski: I decided to scope down the implementation to Rails only, rather than any Ruby app. We don't need the expensive parser anymore and most services we'll see from users will likely be Rails if they are written in Ruby.

Makefile
grcevski: I decided to scope down the implementation to Rails only, rather than any Ruby app. We don't need the expensive parser anymore and most services we'll see from users will likely be Rails if they are written in Ruby.

pkg/internal/rubytools/ruby_tree.go
grcevski: I decided to scope down the implementation to Rails only, rather than any Ruby app. We don't need the expensive parser anymore and most services we'll see from users will likely be Rails if they are written in Ruby.

pkg/internal/rubytools/ruby_tree.go
grcevski: I decided to scope down the implementation to Rails only, rather than any Ruby app. We don't need the expensive parser anymore and most services we'll see from users will likely be Rails if they are written in Ruby.

pkg/internal/rubytools/rails.go
```diff
@@ -0,0 +1,80 @@
...
+			continue
+		}
+
+		class := strings.TrimPrefix(match[1], "::")
+		if class == "Application" {
+			return railsServiceName(module)
+		}
+
+		return railsServiceName(class)
+	}
```
MrAlias, line 101: The Rails integration removes `OTEL_SERVICE_NAME` and expects `testapi` from `module Testapi`, while the project directory is `/build`, so falling back to the directory would change the intended result. The scanner currently returns on the first match and keeps only one flat `module`; a closed nested module can therefore produce the wrong name, and an inactive declaration can beat the real application.

Could we keep the bounded scanner but make it fail closed by maintaining a stack for the recognized Ruby scopes, popping it on `end`, treating candidates inside conditional or other executable scopes as ambiguous, scanning to EOF, and returning a name only when exactly one candidate remains?

Can we also add tests for closed nested modules, conditional declarations, and multiple candidates?

pkg/internal/rubytools/service_metadata.go
```diff
@@ -0,0 +1,184 @@
...
+	configPath := filepath.Join(dir, "config")
+	configInfo, err := os.Lstat(configPath)
+	if errors.Is(err, os.ErrNotExist) {
+		return "", false, nil
+	}
+	if err != nil {
+		return "", false, fmt.Errorf("checking Rails config directory %q: %w", configPath, err)
+	}
+	if configInfo.Mode()&os.ModeSymlink != 0 {
+		return "", true, nil
```
MrAlias: "A symlinked `config` directory is treated as a Rails project before `config/application.rb` is checked. A non-Rails Ruby service that deploys configuration through a directory symlink therefore gets the containing directory as an auto-derived `service.name`.

Could we return `rails=false` unless a Rails application marker is established, while still refusing to follow unsafe symlinks?

pkg/internal/rubytools/rails.go
grcevski: I think I understand the issue you found, it's something like this, right? 

```ruby
module Orders       # records Orders
  module Admin      # Orders is still active, but we'll find Admin instead, since it's closer to the Application
  end
  class Application < Rails::Application
```
This can never be perfect and we need to weigh in the simple parser vs bring in the proper AST as before and all the downside of this. 

I'm going to add some additional rudimentary scope parsing, handling the usual ruby scope syntaxt, but there are many things still in the ruby syntax that will break it (such as multiline comments etc), and I don't think it's worth supporting. The Ruby generator template will not do any of this https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/config/application.rb.tt. So we are adding complexity for very little payback.

At the end of the day if people get a wrong name, they can always supply the correct one with an env variable.

pkg/internal/rubytools/service_metadata.go
grcevski: Good catch. Fixed.

grcevski (commented):
Added additional scope handling, but only the basic one.

skl (approved):
lgtm

pkg/internal/rubytools/rails.go
MrAlias: Limiting this to the conventional Rails-generated structure and falling back for nested or unusual syntax sounds reasonable to me. My remaining concern is multiple candidates: the scanner still returns on the first application declaration, so an earlier inactive or stray declaration can beat the real one. What about scaning to EOF and returning a name only when there is exactly one safe candidate, with a multiple-candidate test? That would keep the parser bounded without requiring an AST.

grcevski: I could do that, but how realistic is that someone will have two module declarations with a Rails::Application in a file, is that even valid? At the end of the day, isn't better we find one of them instead of just saying Ruby? At least it's something, no?

MrAlias: That’s a fair question. Multiple Rails applications are supported, although they’re probably unusual in a typical `config/application.rb`. I don't real have a good enough sense to say how realistic. My concern is less about that specific layout and more that a second match means the scanner may not know which declaration is active.

Since the fallback is normally the project-directory name, would you be open to scanning to EOF and treating multiple candidates as ambiguous? It seems like a small safeguard while keeping the parser intentionally simple.

grcevski: OK, I added a counter, if it's > 1 then we fall back on the directory instead.

mmat11 (approved):
lgtm

PR #3147, opened 2026-08-24
Support Go 1.27 in CI

MrAlias:
## Summary

- Declare Go 1.27.0 as the preferred toolchain in `go.mod`, allowing the existing unit-test workflow to continue using a single latest Go version through `go-version-file`.
- Upgrade golangci-lint to v2.13.1 so lint CI also runs with Go 1.27.
- Raise the lint tools module's minimum Go version to 1.26 and apply the formatter updates required by the new linter release.
- Build the OBI generator image with Go 1.27.0.
- Re-enable Kubernetes dependency updates that were blocked by the former Go 1.25 generator toolchain.

## Context

This is the first stage of the repository's documented Go upgrade process. The Go 1.27 generator image must be merged and published before a follow-up can update `GEN_IMG`, the Docker generator tag, and the remaining maintained `go.mod` files to a Go 1.26 minimum.

The root `go 1.25.11` language floor remains unchanged in this PR so the currently published generator image can still run with `GOTOOLCHAIN=local`.

## Testing

- `actionlint .github/workflows/unit_tests.yml .github/workflows/lint.yml`
- `make generator-image-build`
- `make docker-generate`
- `GOTOOLCHAIN=go1.27.0 make test`
- `GOTOOLCHAIN=go1.27.0 make lint`
- Affected package tests with Go 1.27.0
- Verified the published Go 1.25.11 generator accepts the updated modules


MrAlias:
Updated the unit-test workflow to preserve the existing architecture and shard matrix. It now runs only on Go 1.27.0; the additional Go-version dimension and version-specific cache/artifact naming have been removed.

internal/test/integration/log_enricher_test.go
```diff
@@ -212,6 +212,7 @@ func testLogEnricherNodeJS(t *testing.T) {
 					return
 				}
 				req.Header.Set("traceparent", fmt.Sprintf("00-%s-%s-01", tp.traceID, tp.parentID))
+				req.Close = true
```
mmat11, line 215: question: why was this needed?

pkg/internal/sqlprune/sqlparser_test.go
```diff
@@ -68,9 +68,12 @@ func TestSQLExtraction(t *testing.T) {
 						customers c
 					WHERE c.is = p.id AND c.inactive IS NULL
 				) as bananas`: {op: "SELECT", table: "customers"},
-			"SELECT 1.2":                              {op: "SELECT", table: ""},
-			"SELECT 0xdeadBEEF":                       {op: "SELECT", table: ""},
-			"SELECT A + B":                            {op: "SELECT", table: ""},
+			"SELECT 1.2": {op: "SELECT", table: ""},
```
mmat11, line 71: was this required by some linter?

pkg/runtimemetrics/reader_test.go
```diff
@@ -558,7 +558,9 @@ type runtimeMetricFilter struct {
 func (f runtimeMetricFilter) AllowPID(app.PID, uint32, *exec.FileInfo, ebpfcommon.PIDType) {}
 func (f runtimeMetricFilter) BlockPID(app.PID, uint32)                                     {}
 func (f runtimeMetricFilter) ValidPID(app.PID, uint32, ebpfcommon.PIDType) bool            { return false }
-func (f runtimeMetricFilter) Filter(spans []request.Span) []request.Span                   { return spans }
+
```
mmat11, line 561: same here, required by some linter?

internal/test/integration/log_enricher_test.go
MrAlias: Go 1.27 changed the HTTP/1 transport to drain an unread response body when it is closed, which can return the connection to the idle pool. These tests assume each trace context uses a new connection; reuse made later requests exercise persistent-connection correlation instead. Setting req.Close keeps each request on a separate connection so the tests remain scoped to log enrichment.

pkg/internal/sqlprune/sqlparser_test.go
MrAlias: Yes. This came from upgrading golangci-lint to v2.13.1, which brings in gofumpt v0.11.0. gofumpt rewrites these short map entries with blank lines, and the repository runs it through golangci-lint fmt.

pkg/runtimemetrics/reader_test.go
MrAlias: Yes, this is the same gofumpt v0.11.0 formatter change. It separates Filter from the aligned one-line methods above and CurrentPIDs below; golangci-lint fmt requires this layout.

mmat11 (approved):
lgtm

skl (approved):
ty

PR #3155, opened 2026-08-24
feat(otel): add opt-in gzip compression for OTLP traces export

### LINKED ISSUE #3154: OTLP traces exporter discards the collector's gzip default and offers no compression setting
The OTLP traces exporter always sends uncompressed payloads, and there is no configuration option to change that. Two separate causes, at `a0f822b`.

**The exporter default is thrown away.** `pkg/export/otel/traces.go:289` (HTTP) and `:344` (gRPC) assign a fresh `ClientConfig` struct literal over the one `CreateDefaultConfig()` just returned. That zeroes every field the literal omits, `Compression` among them, and both factories set it on purpose:

- `otlphttpexporter@v0.151.0/factory.go:47` — `clientConfig.Compression = configcompression.TypeGzip`
- `otlpexporter@v0.158.0/factory.go:43` — same

Both protocol branches do it, so switching protocol is not a workaround.

**There is no config surface.** `TracesConfig` at `pkg/export/otel/otelcfg/config_traces.go:25` has no compression field, so a user cannot turn it on even knowing about the above.

The metrics path is not affected. It builds SDK exporters (`pkg/export/otel/otelcfg/exporter.go:69` and `:81`), and those read `OTEL_EXPORTER_OTLP_COMPRESSION` themselves.

Measured on a production deployment, roughly 150 instrumented nodes across two clusters,
with the exporter in its current uncompressed state:

- `sum(rate(beyla_otel_trace_exports_total[10m]))` across the agent fleet: about 24,800 spans/s
- at the receiver, `sum(rate(tempo_distributor_bytes_received_total[10m]))` is about 17.4 MB/s
  over `sum(rate(tempo_distributor_spans_received_total[10m]))` of about 24,800 spans/s, so
  roughly 700 bytes per span
- that is about 58 GB/hr of uncompressed OTLP payload crossing the network

The two span rates are measured independently at the sender and the receiver and agree, which
is the cross-check that the byte figure is sound. For anyone whose collector sits across a
network boundary, or across an availability zone, that is a large amount of egress that gzip
would cut substantially.

I have a fix ready if it is welcome: set the individual `ClientConfig` fields instead of replacing the struct, so the factory default survives, plus a `compression` option using the OTel env var names. Worth noting the spec default for `OTEL_EXPORTER_OTLP_COMPRESSION` is none, while the collector exporter you build on chooses gzip. Keeping the exporter default seemed like the least surprising behaviour, but happy to default it to none instead if you prefer the spec reading, since the discarded default is a bug either way.

*Disclosure: AI-assisted. An assistant did the investigation and wrote the code and tests. Every claim above is pinned to a file:line reference or a test that can be re-run, rather than asked to be taken on trust. I am accountable for the contribution and will answer for it in review.*

mlushpenko:
Fixes #3154.

`TracesConfig` had no compression setting, so OTLP trace export was always uncompressed with no way to turn it on.

Adds a `compression` option with `OTEL_EXPORTER_OTLP_TRACES_COMPRESSION`, and `OTEL_EXPORTER_OTLP_COMPRESSION` as the common fallback, matching how endpoint and protocol are already split in `TracesConfig`. Accepted values are `none` and `gzip`, validated by the struct tag, so a typo fails at startup rather than silently.

The default is unchanged: omitting the option sends uncompressed, which matches both OBI's current behaviour and the spec default for `OTEL_EXPORTER_OTLP_COMPRESSION`. Both protocol branches still build `ClientConfig` as an explicit literal, so this does not pull in any other collector default.

The metrics path needs no change. It builds SDK exporters, which read `OTEL_EXPORTER_OTLP_COMPRESSION` themselves.

## Verification

`pkg/export/otel/traces_compression_test.go` asserts the wire, not the config struct: an `httptest` collector records `Content-Encoding`, and a gRPC `StatsHandler` reads `stats.InHeader.Compression`, since `grpc-encoding` is consumed by the transport and never reaches the handler's metadata. Cases: unset sends uncompressed, `gzip` compresses, `none` is explicit off, and the traces key wins over the common one.

`go test ./pkg/export/otel/... ./pkg/obi/...` is green, `golangci-lint run ./pkg/export/otel/...` is clean, and `make check-config-schema` passes with the regenerated `config-schema.json` and `CONFIG.md` included.

*Disclosure: AI-assisted. An assistant did the investigation and wrote the code and tests. Every claim above is pinned to a file:line reference or a test that can be re-run, rather than asked to be taken on trust. I am accountable for the contribution and will answer for it in review.*


pkg/export/otel/otelcfg/config_traces.go
```diff
@@ -30,6 +30,12 @@ type TracesConfig struct {
 	Protocol       Protocol `yaml:"protocol" env:"OTEL_EXPORTER_OTLP_PROTOCOL"`
 	TracesProtocol Protocol `yaml:"-" env:"OTEL_EXPORTER_OTLP_TRACES_PROTOCOL"`
 
+	CommonCompression string `yaml:"-" env:"OTEL_EXPORTER_OTLP_COMPRESSION"`
+	// TracesCompression is the compression applied to exported OTLP payloads. Accepted
```
marctc: Instead of this comment, we could add some validation here. We use oneof validation tags for other enum-like configuration fields. Could we add one here and to CommonCompression? If supported codecs depend on the selected protocol, we need protocol-specific validation instead

marctc:
> ## On the default
> 
> Keeping gzip is what the exporter you build on already intends. Note the OTel spec default for `OTEL_EXPORTER_OTLP_COMPRESSION` is none, so if you would rather follow the spec reading I am happy to flip the default to none in this PR. The discarded factory default is a bug either way, and the config field lets users choose.


I'm not super familiar with OTLP, but do you see any risks of enabling this by default? could any setup break because of this?

pkg/export/otel/otelcfg/config_traces.go
```diff
@@ -30,6 +31,12 @@ type TracesConfig struct {
 	Protocol       Protocol `yaml:"protocol" env:"OTEL_EXPORTER_OTLP_PROTOCOL"`
 	TracesProtocol Protocol `yaml:"-" env:"OTEL_EXPORTER_OTLP_TRACES_PROTOCOL"`
 
+	CommonCompression string `yaml:"-" env:"OTEL_EXPORTER_OTLP_COMPRESSION" validate:"omitempty,oneof=none gzip zstd zlib snappy deflate lz4 x-snappy-framed"`
```
MrAlias: `OTEL_EXPORTER_OTLP_COMPRESSION` is shared by traces and metrics, but this validation allows values the metrics exporter cannot honor. With no signal-specific override, `zstd` makes traces use zstd while Go metrics silently use no compression.

Could we restrict the common value to `none`/`gzip`, the portable intersection supported by both signals?

pkg/export/otel/otelcfg/config_traces.go
```diff
@@ -30,6 +31,12 @@ type TracesConfig struct {
...
+	// TracesCompression is the compression applied to exported OTLP payloads. Left unset,
+	// the OTLP exporter's own default of gzip applies. gRPC accepts a narrower set than
+	// HTTP, which Validate checks once the protocol is known.
+	TracesCompression string `yaml:"compression" env:"OTEL_EXPORTER_OTLP_TRACES_COMPRESSION" validate:"omitempty,oneof=none gzip zstd zlib snappy deflate lz4 x-snappy-framed" jsonschema:"type=string,enum=none,enum=gzip,enum=zstd,enum=zlib,enum=snappy,enum=deflate,enum=lz4,enum=x-snappy-framed"`
```
MrAlias: Even when trace-specific, these additional codecs expand OBI's public configuration surface beyond the reported gzip issue. OTLP receivers are only required to support `none`/`gzip`, and OBI has no capability negotiation or fallback when an extension is rejected.

Please keep `TracesCompression` to `none`/`gzip` in this PR. Additional codecs should come through a separate proposal with a concrete use case and receiver-compatibility contract.

pkg/export/otel/traces.go
```diff
@@ -286,13 +285,15 @@ func getTracesExporter(ctx context.Context, cfg otelcfg.TracesConfig, im imetric
...
 		config.QueueConfig = configoptional.None[exporterhelper.QueueBatchConfig]()
 		config.RetryConfig = disabledRetry
-		config.ClientConfig = confighttp.ClientConfig{
-			Endpoint: opts.Scheme + "://" + opts.Endpoint + opts.BaseURLPath,
-			TLS: configtls.ClientConfig{
-				Insecure:           opts.Insecure,
-				InsecureSkipVerify: cfg.InsecureSkipVerify,
-			},
-			Headers: convertHeaders(opts.Headers),
+		config.ClientConfig.Endpoint = opts.Scheme + "://" + opts.Endpoint + opts.BaseURLPath
```
MrAlias: Mutating the factory-created `ClientConfig` preserves more than its gzip default. HTTP now also inherits the Collector's timeout, write-buffer, connection-pool, and forced HTTP/2 defaults; the gRPC branch below likewise inherits its write buffer and `round_robin` balancer. Those are unrelated behavior changes and will drift with future Collector upgrades.

Could we retain OBI's explicit `ClientConfig` construction and set only `Compression`?

pkg/export/otel/traces.go
```diff
@@ -476,3 +479,12 @@ func convertHeaders(headers map[string]string) configopaque.MapList {
 	}
 	return opaqueHeaders
 }
+
+// An empty value keeps whatever the exporter factory set. UnmarshalText rejects codecs
+// the collector cannot encode.
+func setCompression(dst *configcompression.Type, value string) error {
+	if value == "" {
+		return nil
```
MrAlias: An empty value cannot always mean “keep the factory default” here. OTel declarative config v2 defines omitted compression as `none`, while OBI's v2 converter rejects explicit compression; a v2 user therefore gets gzip on omission and has no way to opt out.

Could we wire compression through v2 import/export/migration and preserve its omission semantics before changing this default?

MrAlias (changes requested):
The gzip default addresses the reported bandwidth problem, but the shared compression setting is inconsistent across signals, the trace setting broadens the public codec surface beyond this fix, and the implementation changes unrelated Collector defaults and config-v2 semantics. Those boundaries need to be narrowed before this is ready.

pkg/export/otel/otelcfg/config_traces.go
mlushpenko: Both keys are now `oneof=none gzip`. The Go metrics SDK only defines NoCompression and GzipCompression, so any other value on the shared key would have left metrics uncompressed.

pkg/export/otel/otelcfg/config_traces.go
mlushpenko: Narrowed to none/gzip. That also removes the need for protocol-specific validation, so the Validate method is gone.

pkg/export/otel/traces.go
mlushpenko: Back to the explicit literal with only `Compression` set, so none of those other defaults come along.

pkg/export/otel/traces.go
mlushpenko: Omitted compression is `none` again, so v2 semantics are unchanged and there is nothing to migrate. Wiring compression through v2 alongside a default change can be a separate PR if that is the direction you prefer.

pkg/export/otel/otelcfg/config_traces.go
mlushpenko: Added `oneof` on both keys. They are limited to none/gzip now, which is the set both protocols and the metrics SDK support, so protocol-specific validation is not needed after all.

mlushpenko:
Default stays none in this PR, so nothing changes for existing setups.

@marctc on the risk: the one to watch with gzip on by default is a receiver that ignores Content-Encoding. CPU is not a concern, gzip on OTLP protobuf runs a few hundred MB/s per core and a busy node here exports around 100 KB/s. Anyone who wants it can set the new option.

Scope is now narrowed per the review: none/gzip only on both keys, and OBI keeps its explicit ClientConfig. Changing the default, along with wiring compression through config v2, can be a follow-up.

grcevski:
Hi @mlushpenko , can you please merge with main when you can, we had some temporary instability with the tests and your branch picked them up.

mlushpenko:
Merged main, thanks for the heads up.

MrAlias (approved):
The implementation now keeps OBI's uncompressed default and addresses the earlier configuration boundaries. The title and description still describe preserving the exporters' gzip default, so could we update them to match the current opt-in `none`/`gzip` behavior?

mlushpenko:
Updated the title and description to match the current behaviour: opt-in `none`/`gzip`, default stays uncompressed. Thanks for the review.

marctc (approved):

PR #3156, opened 2026-08-24
Preserve application traceparents in Go HTTP/2

MrAlias:
## Summary

- observe application-supplied `traceparent` names before HPACK encoding in current Go standard-library and `golang.org/x/net/http2` clients
- publish a positive, exact per-stream ownership marker shared by gotracer and tpinjector, and gate both existing injection paths on that marker
- cover TLS and plaintext persistent connections, unowned controls, and mixed owned/unowned multiplexed streams

Relates to #2769.

This supersedes only the current Go HTTP/2 slice of #3021 and intentionally leaves the existing injection mechanics unchanged.

## Validation

Passed:

- `make -C bpf/tests test`
- `go test -v -run '^TestHTTP2Go$' -timeout 15m ./internal/test/integration/`
- `go test -v -run '^TestHTTP2GoWithHTTPProtocols$' -timeout 15m ./internal/test/integration/`
- Docker-backed BPF generation and targeted clang-format/clang-tidy
- `make compile`


bpf/gotracer/go_nethttp.c
```diff
@@ -1281,11 +1315,72 @@ static __always_inline void setup_http2_client_conn(void *goroutine_addr,
...
                 bpf_map_update_elem(&http2_req_map, &s_key, &goroutine_addr, BPF_ANY);
+                if (owned_ref_published) {
+                    bpf_map_update_elem(&http2_owned_stream_by_framer, &s_key, &owned_ref, BPF_ANY);
+                }
             }
         }
     }
 }
 
+static __always_inline bool is_traceparent_field_name(const unsigned char *name) {
```
grcevski: maybe we can remove this helper and use `if (stricmp((const char *)temp, "traceparent", W3C_KEY_LENGTH))` like in go_common.h?

grcevski (approved):
LGTM! I only made a minor comment to use the helper we have to compare `traceparent`.

bpf/gotracer/go_nethttp.c
MrAlias: Removed `is_traceparent_field_name` and now call `stricmp((const char *)name, "traceparent", W3C_KEY_LENGTH)` directly in `obi_uprobe_http2ClientConnWriteHeader` in 25e05fe0b. Targeted amd64/arm64 gotracer generation, the BPF unit suite, clang-format, and clang-tidy pass.

bpf/gotracer/go_nethttp.c
mmat11: To build the map key, the code assumes the connection is TLS and reads the underlying TCP connection out of it. For plaintext HTTP/2 there is no TLS wrapper, so that read goes past the end of the object and picks up whatever is next in memory. The helper that turns this into an address doesn't validate what it gets, so the garbage can come back as a valid-looking connection with a wrong or empty address. The map key then doesn't match what the sk_msg side computes, and the app's header gets duplicated anyway. The read existed before, but this PR is the first thing that depends on it being right.
The proper fix is to check whether the connection is TLS first (the client connection already tracks that), unwrap it only in that case, and use the connection directly for plaintext. As a cheaper fix for now, we could at least reject results where the address or either port is empty

MrAlias: Changed `setup_http2_client_conn` in 622331933 to read the client connection TLS-state offset and unwrap `tconn` only when TLS is present. Plaintext HTTP/2 now passes the direct TCP connection to `get_conn_info`, so its ownership marker uses the same socket key as the `sk_msg` path.

The offset database now covers x/net and vendored standard-library layouts, including the x/net v0.12-to-v0.15 transition. Both `TestHTTP2Go` variants pass their TLS and plaintext ownership cases.

mmat11 (approved):
lgtm

PR #3157, opened 2026-08-24
Fix Go HTTP/2 propagation for immediate writers

### LINKED ISSUE #3151: Preserve Go HTTP/2 injection with unbuffered writers
### Component

OBI Go HTTP/2 context propagation

### Problem

The existing direct HTTP/2 injector mutates the Framer buffer from a `WriteHeaders` return probe. With a buffered writer, the bytes may still be available to change. With an unbuffered or immediately flushing writer, the underlying `Writer.Write` may already have consumed the frame before the return probe executes, causing propagation to be missed even though injection appeared to run.

PR #3021 introduced interior pre-write probes and architecture-specific instruction discovery. Those are possible implementation techniques, not requirements. This issue should first establish the failure and find the smallest stable library seam.

This reliability gap is independent of the duplicate-header ownership fix in #2769.

### Required behavior

When context propagation is enabled and an application does not supply `traceparent`, supported Go HTTP/2 clients must emit exactly one OBI header even when their writer does not retain the frame after `Write` returns.

### Investigation scope

1. Add a minimal current-layout fixture with a genuinely unbuffered or immediately flushing writer.
2. Demonstrate whether TLS and plaintext lose direct injection at the current return-probe boundary.
3. Keep a normal buffered writer as a control.
4. Identify the earliest stable exported or internal function boundary where the complete frame is available and has not yet been consumed.
5. Prefer ordinary entry/return probes or existing structure offsets.
6. Consider a versioned interior probe only if no stable symbol seam can satisfy the behavior.
7. If instruction discovery is unavoidable, cover amd64 and arm64, reject unknown instruction layouts atomically, and document the maintenance boundary.

Padding reservation may be used if the demonstrated fix needs capacity, but it is not independently required.

### Out of scope

- Prescribing amd64/arm64 instruction scanning before the reproduction is established.
- Direct user-write transaction safety, which has its own contract.
- `sk_msg` socket rollback.
- Application-owned header detection from #2769.
- Permanent audit/listener infrastructure or broad production force-fallback controls.

### Definition of done

- A pre-fix test reliably demonstrates the affected writer behavior.
- The selected attachment seam is justified against simpler alternatives.
- Supported current layouts inject exactly one OBI header for buffered and unbuffered TLS/plaintext requests without an application header.
- Unknown/custom layouts do not partially attach any new probe group.
- Existing application-owned-header suppression, tracing, metrics, and disabled propagation remain unchanged.
- Architecture-specific code, if any, has focused decoding/boundary tests on every supported architecture.

MrAlias:
Fixes #3151

## Summary

- reproduce Go HTTP/2 propagation with the current x/net layout using buffered, one-byte immediate-write, plaintext, and TLS fixtures
- reserve HPACK field capacity while `WriteHeaders` still owns the frame, then commit the traceparent at `endWrite` entry before the underlying writer can consume it
- discover the narrow padding boundary on amd64 and arm64 and attach the padding/endWrite probes atomically only for recognized x/net or stdlib layouts
- retain the existing `WriteHeaders` return probe as the buffered fallback

## Timing boundary

The existing return probe runs after `Framer.endWrite` has called the underlying writer. A one-byte `bufio.Writer` therefore bypasses its own buffer for HEADERS frames and lets both plaintext and TLS connections consume the frame before the return probe can mutate it. Before this change, buffered requests carried one traceparent while immediate-write requests carried none.

`Framer.endWrite` entry is the earliest stable ordinary function boundary where the complete frame exists and has not yet reached the writer. It does not, however, guarantee 66 bytes of spare backing capacity. The only interior probe is consequently limited to the compiler's `PadLength` decision in `WriteHeaders`: it temporarily requests 66 bytes of padding so Go grows and populates the frame normally. The ordinary `endWrite` entry probe replaces that reserved padding with one HPACK traceparent field and changes its pad length to zero before the write.

The instruction boundary decoder recognizes the current amd64 and arm64 layouts. The padding and commit probes form an atomic group, so an unknown layout attaches neither probe. Existing return-probe behavior remains available for layouts that already have sufficient buffered capacity.

The concurrent application-owned traceparent path suppresses creation of the framer invocation before `WriteHeaders`. Both new probes require that invocation, so an application-owned stream reserves no padding and receives no OBI field.

## Validation

Passed:

- `make generate`
- `make clang-format`
- `make clang-tidy`
- `make -C bpf/tests test`
- `GOARCH=arm64 go test -c ./pkg/internal/goexec`
- focused Go tests for goexec, probe-group resolution, gotracer, and discovery
- privileged BPF verifier suite: `go test -tags=bpf_verifier_tests ./pkg/internal/ebpf/verifier/...`
- privileged HTTP/1, HTTP/2, and gRPC propagation tests; HTTP/2 covers buffered and immediate writers over plaintext and TLS, and verifies exactly one valid traceparent when enabled and none when disabled
- `make compile`


bpf/gotracer/go_nethttp.c
```diff
@@ -1440,6 +1441,212 @@ int GUARDED_PROG(obi_uprobe_net_http2FramerWriteHeaders, struct pt_regs *, ctx)
...
+    const u64 stack_offset = go_offset_of(ot, (go_offset){.v = pad_offset});
+    if (stack_offset == (u64)-1 || stack_offset < 34 || stack_offset >= 512) {
+        return 0;
+    }
+
+    unsigned char *pad_ptr = (unsigned char *)PT_REGS_SP(ctx) + stack_offset;
+    u32 stream_id = 0;
+    u8 original_pad = 0;
+    u64 fragment_len = 0;
+    long err = bpf_probe_read_user(&stream_id, sizeof(stream_id), pad_ptr - 34);
```
grcevski: Can we please name these constants?

pkg/internal/goexec/instructions.go
```diff
@@ -155,6 +165,12 @@ func storeFunctionOffset(
 		}
 
 		existing.Returns = sortedUniqueOffsets(append(existing.Returns, offs.Returns...))
+		if existing.PadStart == 0 {
```
grcevski, line 174: I know we check for -1 in the ebpf side, where do we set this to -1 if there are uninitialized, I can't seem to find the code, likely I'm missing it.

grcevski: Oh I found them, nvm:
	offTable.Table[goexec.FramerPadLengthStackPos] = missingGoOffset
	offTable.Table[goexec.FramerPadLengthStackVendoredPos] = missingGoOffset

grcevski (approved):
LGTM! Very cool approach!

I left minor pedantic comment.

pkg/internal/goexec/instructions.go
MrAlias: Confirmed. `RegisterOffsets` initializes both framer padding entries to `missingGoOffset` in `pkg/internal/ebpf/gotracer/gotracer.go`, then `setFramerPaddingOffset` replaces an entry only when a recognized layout supplies a padding offset. No code change is needed for this thread.

bpf/gotracer/go_nethttp.c
MrAlias: Named all three stack-layout values in 680598f2c: `k_h2_pad_length_to_stream_id_offset`, `k_h2_pad_length_to_fragment_len_offset`, and `k_h2_pad_length_stack_offset_limit`. The same constants now drive the bounds check and pointer arithmetic. `make generate/all`, `make clang-tidy`, and `make -C bpf/tests test` pass.

bpf/gotracer/go_nethttp.c
```diff
@@ -1440,6 +1441,221 @@ int GUARDED_PROG(obi_uprobe_net_http2FramerWriteHeaders, struct pt_regs *, ctx)
...
+            ? commit_http2_reserved_padding(buf, n, f_info)
+            : append_http2_traceparent_to_framer(framer, wbuf_pos, buf, n, cap, f_info);
+    if (committed) {
+        bpf_map_delete_elem(&framer_invocation_map, &g_key);
+    }
+    return 0;
+}
+
 SEC("uprobe/http2FramerWriteHeaders_returns")
 int GUARDED_PROG(obi_uprobe_http2FramerWriteHeaders_returns, struct pt_regs *, ctx) {
```
mmat11, line 1787: The old return hook is still active as a fallback and doesn't know padding was reserved. If the new send hook doesn't run (for example, the compiler inlines that function), the frame goes out with 66 zero bytes and the fallback appends a second traceparent. The receiver decodes the zeros as garbage headers and resets every request on the connection.
Fix: in the fallback, check whether padding was reserved and not yet filled, and do nothing if so. Before installing the new hooks, verify the send function is actually called from the frame writer; if it isn't, don't install them

MrAlias: The return fallback now skips framer_invocation_map entries with reserved padding, so it cannot append a second traceparent after the pre-flush hook reserved the frame tail. The optional pre-flush group also records direct call targets from WriteHeaders and is installed only when that exact library copy calls the matching endWrite symbol; an inlined or otherwise unused endWrite leaves the group disabled.

bpf/gotracer/go_nethttp.c
```diff
@@ -1567,6 +1567,222 @@ int GUARDED_PROG(obi_uprobe_net_http2FramerWriteHeaders, struct pt_regs *, ctx)
     return 0;
 }
 
+static __always_inline void
+make_http2_traceparent_field(unsigned char field[k_h2_tp_hpack_huffman_size], const tp_info_t *tp) {
+    field[0] = 0;
+    field[1] = sizeof(tp_encoded) | 0x80;
+    __builtin_memcpy(field + 2, tp_encoded, sizeof(tp_encoded));
```
mmat11: use `bpf_memcpy`

bpf/gotracer/go_nethttp.c
```diff
@@ -1567,6 +1567,222 @@ int GUARDED_PROG(obi_uprobe_net_http2FramerWriteHeaders, struct pt_regs *, ctx)
...
+                             field,
+                             sizeof(field)) != 0) {
+        return false;
+    }
+
+    const u8 consumed_padding = 0;
+    return bpf_probe_write_user(pad_length_ptr, &consumed_padding, sizeof(consumed_padding)) == 0;
+}
+
+static __always_inline bool append_http2_traceparent_to_framer(
```
mmat11, line 1718: `The append path can oversize a HEADERS frame and kill the connection.
The fallback that appends the traceparent at endWrite allows the frame to be up to header + 16384 bytes before adding its 66-byte field, so a request whose header block lands within 66 bytes of the peer's max frame size (16384 is both the default and the RFC minimum) goes out oversized. The reserve path does this math correctly (16384 − 66 − 5 − 1); the append guard just misses the − 66.`

pkg/internal/goexec/instructions.go
```diff
@@ -205,7 +223,15 @@ func staticSymbolOffsets(fName string, allSyms map[string]procs.Sym, ilog *slog.
 			ilog.Error("error finding returns for symbol", "symbol", fName, "offset", s.Off-s.Prog.Off, "size", s.Len, "error", err)
 			return FuncOffsets{}, false
 		}
-		return FuncOffsets{Start: s.Off, Returns: returns}, true
+		var callTargets []uint64
+		if isFramerWriteHeaders(fName) {
```
mmat11: this is a generic function, I don't think we should handle struct specific offsets in here

pkg/internal/goexec/instructions.go
```diff
@@ -241,13 +267,41 @@ func findFuncOffset(f *gosym.Func, elfF *elf.File) (FuncOffsets, bool, error) {
 			if err != nil {
 				return FuncOffsets{}, false, fmt.Errorf("finding function return: %w", err)
 			}
-			return FuncOffsets{Start: off, Returns: returns}, true, nil
+			var callTargets []uint64
+			padStart := uint64(0)
+			padOffset := uint64(0)
+			if isFramerWriteHeaders(f.Name) {
```
mmat11: same here, I think we should handle this more generically

pkg/internal/goexec/instructions_amd64.go
```diff
@@ -47,3 +47,99 @@ func FindReturnOffsets(baseOffset uint64, data []byte) ([]uint64, error) {
...
+}
+
+func FindPadStartOffset(baseOffset uint64, data []byte) (uint64, uint64, error) {
+	type decoded struct {
+		index int
+		inst  x86asm.Inst
+	}
+	instructions := make([]decoded, 0, len(data)/4)
+	for index := 0; index < len(data); {
+		if isENDBRXX(data[index:]) {
```
mmat11: the decode-walk (ENDBR skip + x86asm.Decode + advance) is written three times in this file (returns, call targets, pad boundary); a shared iterator taking a per-instruction callback would keep the ENDBR workaround and the error convention in one place

pkg/internal/goexec/instructions_amd64.go
```diff
@@ -47,3 +47,99 @@ func FindReturnOffsets(baseOffset uint64, data []byte) ([]uint64, error) {
...
+					targets = append(targets, uint64(target))
+				}
+			}
+		}
+		index += instruction.Len
+	}
+	return targets, nil
+}
+
+func FindPadStartOffset(baseOffset uint64, data []byte) (uint64, uint64, error) {
```
mmat11, line 73: could we add a short comment explaining what this does?

pkg/internal/goexec/instructions_arm64.go
```diff
@@ -27,3 +29,66 @@ func FindReturnOffsets(baseOffset uint64, data []byte) ([]uint64, error) {
...
+		if target >= 0 {
+			targets = append(targets, uint64(target))
+		}
+	}
+	return targets, nil
+}
+
+func FindPadStartOffset(baseOffset uint64, data []byte) (uint64, uint64, error) {
+	const (
+		loadByteImmediateMask = 0xffc00000
```
mmat11, line 55: could we add a short comment on these constants explaining what they are?

bpf/gotracer/go_nethttp.c
MrAlias: ccda9053c uses `bpf_memcpy` for the encoded traceparent name. `make generate` and `make clang-tidy` pass.

bpf/gotracer/go_nethttp.c
MrAlias: `ccda9053c` subtracts the 66-byte traceparent field from the fallback's maximum accepted length and uses that same bound for the verifier clamp. Appending the field can no longer push the HEADERS payload over 16384 bytes. `make generate` and the BPF unit suite pass.

pkg/internal/goexec/instructions.go
MrAlias: `eb5c160d9` removes the Framer-name branch from `staticSymbolOffsets`. The ELF-symbol path now delegates to `analyzeFunctionOffsets`, which collects return sites, call targets, and stack-field patterns uniformly.

pkg/internal/goexec/instructions.go
MrAlias: `eb5c160d9` makes `findFuncOffset` use the same generic instruction analyzer. Neither ELF lookup path contains struct- or Framer-specific branching now.

pkg/internal/goexec/instructions_amd64.go
MrAlias: `eb5c160d9` centralizes ENDBR skipping, `x86asm.Decode`, error reporting, and instruction advancement in `walkX86Instructions`; return, call-target, and padding analysis now supply callbacks.

pkg/internal/goexec/instructions_amd64.go
MrAlias: `eb5c160d9` documents that `FindPadStartOffset` recognizes the compiler sequence that loads the `PadLength` byte from the stack and branches when it is zero.

pkg/internal/goexec/instructions_arm64.go
MrAlias: `eb5c160d9` documents the constants as fields from the LDRB immediate encoding used to recognize a byte load from SP.

mmat11 (approved):

grcevski:
Latest changes seem to have a BPF bug:
```
misaligned stack access off 0+0+-86 size 8
processed 187 insns (limit 1000000) max_states_per_insn 0 total_states 14 peak_states 14 mark_read 10
2026/09/08 17:08:39 WARN couldn't load tracer component=ebpf.ProcessTracer error="loading and assigning BPF objects: loading spec 0: loading and assigning BPF objects: assign values: field ObiUprobeHttp2FramerEndWrite: program obi_uprobe_http2FramerEndWrite: load program: permission denied: misaligned stack access off 0+0+-86 size 8 (230 line(s) omitted)" required=true
```

bpf/gotracer/go_nethttp.c
MrAlias: I reverted this to `__builtin_memcpy`. This destination is `field + 2`, so it is only 2-byte aligned. Our `bpf_memcpy` implementation copies this 8-byte value through `__it_mob`, which casts the destination to `u64 *`; that generated an 8-byte stack store at offset `-86`, and the kernel verifier rejected `obi_uprobe_http2FramerEndWrite` on both amd64 and arm64.

For this unaligned wire-format field, Clang lowers `__builtin_memcpy` to four aligned 16-bit stores. I verified the generated instructions, the exact gotracer verifier configuration with header propagation enabled, and the privileged HTTP/2 immediate-write test. All pass with the builtin copy.

PR #3158, opened 2026-08-24
Make sk_msg HTTP/2 mutation rollback-safe

### LINKED ISSUE #3153: Make sk_msg HTTP/2 mutation rollback-safe
### Component

OBI `sk_msg` HTTP/2 context propagation

### Problem

The plaintext HTTP/2 injector changes a socket message through a sequence of helpers such as `bpf_msg_push_data`, `bpf_msg_pull_data`, frame-length writes, HPACK writes, and potentially `bpf_msg_pop_data` during rollback. A failure after the message has been resized can leave an invalid HTTP/2 frame or depend on rollback helpers that are unavailable or unreliable on part of OBI's supported kernel matrix.

PR #3021 explored transactional socket mutation, but this is independent of the duplicate-header behavior tracked in #2769.

### Required safety property

If `sk_msg` injection cannot complete, the injector must either restore the original message exactly or avoid mutation. It must never pass a partially resized or internally inconsistent HTTP/2 message to the peer.

### Scope

- Document the current socket-mutation sequence and every helper failure boundary.
- Preflight all conditions that can be checked before `bpf_msg_push_data`.
- Define the commit point for the inserted HPACK bytes and HTTP/2 frame length.
- Check every helper and memory-write result.
- Verify rollback instead of assuming `bpf_msg_pop_data` or a compensating write succeeded.
- Detect or gate helper availability across Linux 5.8+, supported RHEL-family 4.18 backports, and the repository's verifier matrix.
- Use a no-mutation path when safe rollback cannot be guaranteed.
- Add BPF host tests that fail each push, pull, write, readback, and pop step.
- Add privileged plaintext coverage proving peer continuity after each simulated failure and on a subsequent request.

### Out of scope

- Go user-buffer mutation.
- Application-owned `traceparent` detection from #2769.
- General-purpose audit maps/listeners or production configuration added solely to force test branches.

Fault injection is part of this issue's definition of done, not a separate issue.

### Definition of done

- Every socket mutation boundary has an explicit, tested failure outcome.
- A failed injection leaves byte-for-byte original framing or performs no mutation.
- Unsupported helper combinations select a verifier-safe no-mutation program/path.
- The peer connection remains usable after every injected failure.
- Successful plaintext injections still produce exactly one valid `traceparent`.
- Existing protocol detection, metrics, tracing, and disabled-propagation behavior remain unchanged.
- BPF host tests, verifier/kernel tests, targeted integrations, generation, formatting, clang-tidy, `make verify`, and `make build` pass.

MrAlias:
## Summary

- treat plaintext HTTP/2 `sk_msg` injection as a checked socket-message transaction
- preflight stable bounds and frame state before `bpf_msg_push_data`, then publish the enlarged frame length only after the HPACK bytes are ready
- restore the pushed range and original frame length on every pre-commit failure, with checked retries and readback instead of assuming rollback succeeded
- gate mutation on verifier support for `bpf_msg_pop_data`, replacing the mutator with a scan-only program when rollback is unavailable
- cover helper failures with host mocks and an attached privileged test that verifies restoration plus a successful follow-up write on the same peer connection

## Safety contract

The three-byte HTTP/2 frame-length store is the socket-message commit point. The inserted HPACK field is already in place, and no fallible socket helper or message-memory write follows it. The checked trace-map update after commit is bookkeeping and cannot alter the wire bytes.

Before commit, any failure removes the exact pushed range, verifies the original message size, restores the original frame length, and verifies that length by readback. A failed first rollback operation is retried. Bytes are passed only after rollback is verified. If rollback remains uncertain, `SK_DROP` makes the application write fail and the connection must be discarded rather than exposing a partial frame. The per-stream `written` flag is published only after the wire transaction commits.

## Kernel compatibility

The loader checks `bpf_msg_pop_data` with `features.HaveProgramHelper`, so supported RHEL-family 4.18 backports are handled independently of upstream versioning. No socket program is attached during startup.

The attached `SK_MSG` semantic probe is limited to the privileged test. It injects failures at the preflight pull, push, post-push pull, frame pull, rollback pop, and rollback pull boundaries, then checks byte-for-byte restoration and a successful follow-up injection on the same connection.

If the helper capability check fails, the collection substitutes a verifier-safe program that resumes HTTP/2 frame scanning without push, pull, pop, or message-memory writes. Protocol detection, tracing, metrics, existing on-wire traceparents, and configurations with header propagation disabled retain their existing paths.

## Validation

- `make generate`
- `clang-format` on the modified C test sources
- `make clang-tidy`
- `make -C bpf/tests test`
- `go test ./pkg/internal/ebpf/tpinjector`
- `go test -count=1 -v -exec='sudo -n' -tags=privileged_tests -run '^TestH2MutationRollback$' ./pkg/internal/ebpf/tpinjector`
- `go test -count=1 -v -parallel 1 -exec='sudo -n' -tags=bpf_verifier_tests -run '^TestBPFVerifierWithConstants/tpinjector/Bpf/' -timeout 20m ./pkg/internal/ebpf/verifier`
- `make verify`
- `make build`

Fixes #3153


grcevski (approved):
LGTM! Very cool!

bpf/tests/testdata/h2_mutation_peer.c
```diff
@@ -0,0 +1,69 @@
+// Copyright The OpenTelemetry Authors
```
mmat11, line 1: can we move this test data under bpf/tests?

pkg/internal/ebpf/tpinjector/h2_mutation_check_test.go
```diff
@@ -0,0 +1,246 @@
...
+
+	for _, boundary := range boundaries {
+		if err := verifyH2MutationBoundary(&objects, boundary.mask); err != nil {
+			return fmt.Errorf("%s boundary: %w", boundary.name, err)
+		}
+	}
+	return nil
+}
+
+func verifyH2MutationBoundary(objects *BpfH2MutationProbeObjects, faultMask uint64) error {
```
mmat11: these look very complicated/invasive probes, is there a way to check these capabilities without attaching programs?

pkg/internal/ebpf/tpinjector/h2_mutation_check_test.go
```diff
@@ -0,0 +1,246 @@
...
+	var objects BpfH2MutationProbeObjects
+	if err := spec.LoadAndAssign(&objects, nil); err != nil {
+		return fmt.Errorf("loading rollback probe: %w", err)
+	}
+	defer objects.Close()
+
+	attach := link.RawAttachProgramOptions{
+		Target:  objects.Sockets.FD(),
+		Program: objects.H2MutationPeer,
+		Attach:  ebpf.AttachSkMsgVerdict,
```
mmat11, line 55: /cc @grcevski was this the bugged datapath?

mmat11 (commented):
Claude findings:

- Dropping the message kills the application's connection. When rollback can't be verified, the program tells the kernel to drop the message. That makes the application's write call fail with a permission error, and Go treats the connection as dead. The PR describes this as harmless and "scoped"; it isn't. The docs also describe a mitigation that was deleted in the last commit.
How to fix: make an explicit decision. Either keep the drop and document it as a visible write error for the app, or pass the message through and mark the stream so nothing touches it again. Remove the dead mitigation code and fix the docs.

- Most of the verification does nothing. After writing bytes, the code reads them back to verify them. But nothing could have changed them in between, so the readbacks can never fail in production. They add about 400 BPF instructions for no safety.
How to fix: keep only the readback after the undo step, where the message really did change. Remove the rest.

pkg/internal/ebpf/tpinjector/h2_mutation_check_test.go
grcevski: Ugh yeah, you are right, I missed that. Depending on which sockets we attach this too, it will fail the sockets. I found that if the socket did splice() then the verdict breaks it.

MrAlias:
Addressed the review-summary findings in 99cc4de5f:

- Kept `SK_DROP` as an explicit fail-closed outcome for uncertain rollback. Passing a possibly resized or partially rewritten HTTP/2 frame would corrupt the peer connection. The dead `bpf_msg_apply_bytes` scoping code is removed, and the documentation now states that the application write fails and the connection must be discarded.
- Removed the HPACK and frame-length readbacks that immediately followed ordinary memory stores, along with their synthetic fault boundaries. The restoration readback after rollback remains.
- Removed the runtime loopback attachment. Startup now uses the verifier helper-capability check; the semantic probe is limited to the privileged test.

`make verify`, the privileged same-peer test, and all tpinjector verifier variants pass.

pkg/internal/ebpf/tpinjector/h2_mutation_check_test.go
MrAlias: Removed this attach path from production startup. The loopback `SK_MSG` program now runs only in `TestH2MutationPeerContinuity`; normal tracer loading performs the verifier capability check without attaching a socket program. The privileged continuity test and all tpinjector verifier variants pass.

bpf/tests/testdata/h2_mutation_peer.c
MrAlias: Moved the probe fixture to `bpf/tests/testdata/h2_mutation_peer.c`. The `testdata` subdirectory keeps the eBPF-only source out of the top-level host-test wildcard, and the generated probe now carries the `privileged_tests` build tag.

pkg/internal/ebpf/tpinjector/h2_mutation_check_test.go
MrAlias: Runtime startup now checks `asm.FnMsgPopData` with `features.HaveProgramHelper` and does not load or attach the loopback program. The attach-based semantic probe moved to `h2_mutation_check_test.go` and is compiled only for `privileged_tests`.

bpf/tests/testdata/h2_mutation_peer.c
```diff
@@ -0,0 +1,67 @@
...
+    "\x00\x0btraceparent\x37"
+    "00-0123456789abcdef0123456789abcdef-0123456789abcdef-01";
+
+static __always_inline bool mutation_fault(u32 boundary) {
+    const u32 zero = 0;
+    const u64 *mask = bpf_map_lookup_elem(&fault_mask, &zero);
+    return mask && (*mask & (1ULL << boundary));
+}
+
+#define H2_SOCKET_TRANSACTION_FAULT(boundary) mutation_fault(boundary)
```
mmat11: why do we have this as a define and not a function?

bpf/tpinjector/h2_write_transaction.h
```diff
@@ -0,0 +1,154 @@
...
+    k_h2_socket_boundary_push,
+    k_h2_socket_boundary_write_pull,
+    k_h2_socket_boundary_frame_pull,
+    k_h2_socket_boundary_rollback_pop,
+    k_h2_socket_boundary_rollback_pull,
+    k_h2_socket_boundary_rollback_readback,
+} h2_socket_transaction_boundary_t;
+
+#ifndef H2_SOCKET_TRANSACTION_FAULT
+#define H2_SOCKET_TRANSACTION_FAULT(boundary) false
```
mmat11: is this a test-only thing? I would avoid adding complexity here

bpf/tpinjector/tpinjector.c
```diff
@@ -1888,3 +1849,15 @@ int obi_packet_extender_write_h2_tp(struct sk_msg_md *msg) {
                         k_h2_tp_hpack_size);
     return SK_PASS;
 }
+
+// Kernels that cannot verify bpf_msg_pop_data load this program in place of the mutator.
```
mmat11, line 1853: what are the implications of this kernels? will grpc CP just have no rollback or could there be more complications?

pkg/internal/ebpf/tpinjector/h2_mutation_check_test.go
mmat11: much better now!

bpf/tests/testdata/h2_mutation_peer.c
MrAlias: Removed `H2_SOCKET_TRANSACTION_FAULT` and the fault map entirely. The host transaction test now fails the mocked `bpf_msg_pull_data`, `bpf_msg_push_data`, and `bpf_msg_pop_data` helpers directly, so the production transaction header no longer needs this test-only macro.

bpf/tpinjector/h2_write_transaction.h
MrAlias: Removed the test boundary enum and every synthetic fault branch from `h2_write_transaction.h`. Host tests now exercise rollback by failing the socket-helper mocks directly; the privileged probe is limited to checking exact mutation bytes and a second write on the same connection. The host suite, privileged continuity test, and all tpinjector verifier variants pass.

bpf/tpinjector/tpinjector.c
MrAlias: The loader disables the `sk_msg` HPACK mutator entirely when the verifier rejects `bpf_msg_pop_data`; it does not run mutation without rollback. The replacement program only resumes HTTP/2 frame scanning, so this socket-level injection path is unavailable on that kernel while detection, tracing, existing on-wire traceparents, and other propagation paths remain unchanged.

pkg/internal/ebpf/tpinjector/h2_mutation_check_test.go
mmat11: I don't think there's any value in TestH2MutationPeerContinuity if it's code we don't use anyway -- I think we can delete both this file and h2_mutation_privileged_test.go, correct me if I am wrong

bpf/tests/testdata/h2_mutation_peer.c
```diff
@@ -0,0 +1,52 @@
...
+    __type(key, u32);
+    __type(value, u32);
+    __uint(pinning, OBI_PIN_INTERNAL);
+} invocations SEC(".maps");
+
+static const unsigned char expected_hpack[k_h2_tp_hpack_size] =
+    "\x00\x0btraceparent\x37"
+    "00-0123456789abcdef0123456789abcdef-0123456789abcdef-01";
+
+#include <tpinjector/h2_write_transaction.h>
```
pinoOgni: Can we write this include at the beginning of the file?

bpf/tests/bpf_h2_socket_write_transaction.c
```diff
@@ -0,0 +1,204 @@
...
+    msg->size = message_len;
+    msg->data = message;
+    msg->data_end = message + message_len;
+    return 0;
+}
+
+#define bpf_msg_pull_data test_msg_pull_data
+#define bpf_msg_push_data test_msg_push_data
+#define bpf_msg_pop_data test_msg_pop_data
+#include <tpinjector/h2_write_transaction.h>
```
pinoOgni: Also this one

MrAlias: Moved the transaction include ahead of the host mock state and helper implementations in 911c51ea7. The three wrapper declarations and macro substitutions must precede the include so the shared inline transaction calls the host mocks.

bpf/tests/testdata/h2_mutation_peer.c
MrAlias: Moved the transaction include ahead of the fixture maps, constants, and helper implementations in 911c51ea7. The test-only helper declarations and macro substitutions remain immediately before it so the shared inline transaction binds to fault-injecting wrappers without adding hooks to the production header.

pkg/internal/ebpf/tpinjector/h2_mutation_check_test.go
MrAlias: `h2_write_socket_transaction` is used by the production `obi_packet_extender_write_h2_tp` program; this file is the test-only program that attaches the same helper to a socket. I kept the test and expanded it in 911c51ea7 to inject all six retryable helper failures, verify exact restoration at the peer, and then verify a successful mutation on the same connection.

skl (approved):
lgtm

bpf/tpinjector/h2_write_transaction.h
```diff
@@ -0,0 +1,128 @@
...
+
+    unsigned char *data = msg->data;
+    if (!data || (void *)data + k_h2_tp_hpack_size > msg->data_end) {
+        return h2_restore_socket_message(
+                   msg, original_size, frame_offset, payload_len, inject_offset)
+                   ? k_h2_socket_transaction_rollback_verified
+                   : k_h2_socket_transaction_rollback_uncertain;
+    }
+
+    __builtin_memcpy(data, expected, k_h2_tp_hpack_size);
```
mmat11: use `bpf_memcpy`

bpf/tpinjector/h2_write_transaction.h
```diff
@@ -0,0 +1,128 @@
...
+        popped = bpf_msg_pop_data(msg, inject_offset, k_h2_tp_hpack_size, 0) == 0;
+    }
+    if (!popped) {
+        return false;
+    }
+    if (msg->size != original_size) {
+        return false;
+    }
+
+    bool pulled = bpf_msg_pull_data(msg, frame_offset, frame_offset + 3, 0) == 0;
```
mmat11: question: why does it try to pull twice here? also what's `3`?

same for `popped` some lines up

bpf/tpinjector/tpinjector.c
```diff
@@ -1847,36 +1817,27 @@ int obi_packet_extender_write_h2_tp(struct sk_msg_md *msg) {
...
 
-    // past the push the inserted bytes exist but hold no field, so every exit has to undo
-    if (bpf_msg_pull_data(msg, inject_offset, inject_offset + k_h2_tp_hpack_size, 0) != 0) {
-        h2_undo_tp_push(msg, frame_offset, inject_offset, payload_len);
-        return SK_PASS;
+    const h2_socket_transaction_outcome_t outcome =
+        h2_write_socket_transaction(msg, frame_offset, payload_len, inject_offset, expected);
+    if (outcome == k_h2_socket_transaction_rollback_uncertain) {
+        bpf_d_printk("dropping message after uncertain h2 rollback [%s]", __FUNCTION__);
+        return SK_DROP;
```
mmat11, line 1830: will this drop the packet?

bpf/tpinjector/h2_write_transaction.h
MrAlias: Rollback gets one retry because it is the last chance to restore the original message before rejecting the application write. A repeated helper failure returns `rollback_uncertain`.

I replace `3` with `k_h2_frame_len_field_size` in ff8ce8ba8.

bpf/tpinjector/tpinjector.c
MrAlias: Yes—more precisely, `SK_DROP` rejects the current socket message. Meaning the current socket message is dropped, but not the TCP connection. The kernel returns `EACCES` from `sendmsg` and leaves the socket open. The HTTP/2 transport should retire that connection because the dropped HEADERS frame can leave the HPACK encoder and decoder states out of sync. Whether the RPC is retried on a new connection depends on the client implementation and retry policy; it is not guaranteed.

This is intentional: passing a possibly resized or partially rewritten HTTP/2 frame would corrupt the peer connection. The failure behavior is [documented here](https://github.com/MrAlias/opentelemetry-ebpf-instrumentation/blob/ff8ce8ba8d8cd81112e1b179fb320fd022af27c4/devdocs/grpc-context-propagation.md#socket-mutation-transaction).

mmat11 (approved):

PR #3159, opened 2026-08-24
Make Go HTTP/2 Framer injection fail safely

### LINKED ISSUE #3150: Make Go HTTP/2 user-buffer injection fail safely
### Component

OBI Go HTTP/2 context propagation

### Problem

The existing Go HTTP/2 Framer injector mutates several pieces of user memory with separate `bpf_probe_write_user` calls: the appended HPACK field, the writer length, and the HTTP/2 frame length. If an early write succeeds and a later write fails, the frame and writer metadata can disagree.

PR #3021 explored transactional direct writes, but that work is independent of the application-owned-header fix in #2769 and should be evaluated on its own.

### Required safety property

A failed direct injection must either:

- leave the original HTTP/2 frame valid and sendable; or
- commit one complete, internally consistent `traceparent` injection.

The implementation must not assume that a sequence of user-memory writes is atomic or that rollback itself cannot fail.

### Scope

- Inventory every user-memory mutation in the existing Go HTTP/2 and grpc-go Framer injection paths.
- Check and propagate the result of every `bpf_probe_write_user` call.
- Preflight buffer capacity, offsets, and all state needed before the first mutation.
- Establish an explicit commit order. Prefer writing new bytes before publishing writer/frame lengths when that leaves the original frame authoritative until commit.
- Define fail-closed behavior for cases where exact restoration cannot be guaranteed.
- Keep outgoing trace state consistent with the actual mutation result.
- Add focused BPF host tests that fail each mutation step independently.
- Add privileged TLS and plaintext coverage proving that a simulated failure does not corrupt the connection and that a later request still succeeds.
- Validate the changed programs with the supported verifier/kernel matrix.

### Out of scope

- Application-owned `traceparent` detection from #2769.
- Moving injection to a pre-`Writer.Write` interior probe or solving unbuffered-writer timing.
- `sk_msg` socket mutation.
- Permanent audit maps, listeners, or production fallback controls added only for tests.

Fault injection is part of this issue's definition of done, not a separate feature.

### Definition of done

- Every direct mutation step has a tested failure path.
- No tested failure produces a mismatched HTTP/2 payload length, writer length, or partial HPACK field.
- The peer connection remains usable after every injected failure.
- Successful injections still produce exactly one valid `traceparent`.
- Existing HTTP/2 metrics, gRPC tracing, and disabled-propagation behavior remain unchanged.

MrAlias:
## Summary

- make the net/http HTTP/2 and grpc-go direct Framer injectors share one checked user-buffer transaction
- preflight frame identity, bounds, capacity, offsets, and writer state before the first mutation
- reconcile failed metadata writes to an exact pristine or committed state and fail closed when neither can be verified
- keep per-stream outgoing injection state aligned with the observed transaction result
- cover every mutation boundary and persistent plaintext/TLS connection continuity

## Safety property

The previous injectors appended an HPACK field and updated the Go writer and HTTP/2 frame lengths with separate unchecked or partially checked `bpf_probe_write_user` calls. A later failure could therefore publish metadata that did not describe the bytes the peer would receive.

The shared transaction now writes and verifies the complete HPACK field while the original lengths remain authoritative. It then publishes the three-byte frame length and finally the Go writer length. Every write result is checked and the resulting metadata is read back. On failure, reconciliation verifies an exact original state, attempts exact rollback, or finishes one complete commit. If neither state can be proven, the result is uncertain and further socket mutation for that stream is suppressed.

Direct injection is bypassed without mutation when offsets, capacity, frame type, stream ID, `END_HEADERS`, payload length, or writer state do not match the supported single-frame layout. This does not add CONTINUATION support or change application-owned-header detection.

## Validation

- `make generate`
- `make clang-format`
- `make clang-tidy`
- `make -C bpf/tests test`
- `go test ./pkg/internal/ebpf/gotracer`
- `go test -exec=sudo -v -count=1 -timeout 10m -tags=privileged_tests -run "^(TestGRPCClientTraceparentRespectsPropagationMode|TestGRPCClientTraceparentWriteFailureContinuity|TestHTTP2ClientTraceparentWriteFailureContinuity)$" ./pkg/internal/ebpf/gotracer`
- `go test -exec=sudo -v -count=1 -timeout 20m -tags=bpf_verifier_tests ./pkg/internal/ebpf/verifier/...`
- `go test -exec=sudo -v -count=1 -timeout 10m -parallel 1 -tags=bpf_verifier_tests -run "TestBPFVerifierWithConstants/gotracer/Bpf" ./pkg/internal/ebpf/verifier/...`
- `SKIP_DOCKER_BUILD=1 go test -v -count=1 -run "^(TestHTTP2Go|TestHTTP2GoWithHTTPProtocols)$" -timeout 10m ./internal/test/integration/`

Fixes #3150.


bpf/gotracer/go_grpc.c
```diff
@@ -758,18 +759,25 @@ int GUARDED_PROG(obi_uprobe_grpcFramerWriteHeaders, struct pt_regs *, ctx) {
...
 
     if (!w_ptr) {
         bpf_dbg_printk("w ptr is 0");
         return 0;
     }
 
-    const u32 conn_ptr_pos =
+    const u64 conn_ptr_pos =
         go_offset_of(ot, (go_offset){.v = _grpc_transport_buf_writer_conn_pos});
+    if (conn_ptr_pos == (u64)-1) {
```
grcevski, line 774: Do we set -1 for _grpc_transport_buf_writer_conn_pos by default? I think we probably should set `missingGoOffset` for all offsets and write a helper or something. It has always been unsafe with 0 as the default.

grcevski (approved):
LGTM! One comment on ensuring that we really set -1 to these offsets.

bpf/gotracer/go_grpc.c
MrAlias: Initialized the three grpc-go `bufWriter` fields—including `_grpc_transport_buf_writer_conn_pos`—to `missingGoOffset` before applying discovered values. The first version also included shared net/http and `bufio.Writer` fields, but PR CI exposed existing probes that consume those entries without guarding `-1`; commit f90617646 narrows the sentinel group accordingly. `TestMissingGoGRPCBufWriterOffsetsUseSentinel` covers the grpc-go fields and verifies that the shared writer fields retain their existing zero default. The focused Go tracer suite and the isolated grpc-go/net/http HTTP/2 continuity cases pass.

mmat11 (approved):
lgtm -- I think the PR needs a rebase and I'd run at least the verifier tests on #3159 #3158 #3157 #3156

PR #3162, opened 2026-08-24
nodejs: fix config file entrypoint parsing

MrAlias:
## Summary

- classify `--experimental-config-file` as an option that consumes its following value
- preserve the application script for automatic service identity and route harvesting
- cover the separated config-file syntax with a focused regression test

## Validation

- `go test ./pkg/internal/nodejstools`
- `make verify`

### DIFF
--- a/pkg/internal/nodejstools/launch_test.go
+++ b/pkg/internal/nodejstools/launch_test.go
@@ -52,6 +52,7 @@ func TestParseNodeLaunch(t *testing.T) {
 		{name: "V8 option with attached value", args: []string{"--stack-trace-limit=20", "app"}, entryPoint: "app"},
 		{name: "attached config file", args: []string{"--experimental-config-file=node.config.json", "app"}, entryPoint: "app"},
 		{name: "separate config file", args: []string{"--experimental-config-file", "config", "server.js"}, entryPoint: "server.js"},
+		{name: "default config file", args: []string{"--experimental-default-config-file", "app"}, entryPoint: "app"},
 		{name: "removed Node 16 option", args: []string{"--experimental-fetch", "app"}, entryPoint: "app"},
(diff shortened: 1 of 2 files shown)


pkg/internal/nodejstools/launch_test.go
```diff
@@ -51,7 +51,7 @@ func TestParseNodeLaunch(t *testing.T) {
 		{name: "underscore option", args: []string{"--inspect_port", "9230", "app.js"}, entryPoint: "app.js"},
 		{name: "V8 option with attached value", args: []string{"--stack-trace-limit=20", "app"}, entryPoint: "app"},
 		{name: "attached config file", args: []string{"--experimental-config-file=node.config.json", "app"}, entryPoint: "app"},
-		{name: "default config file", args: []string{"--experimental-config-file", "app"}, entryPoint: "app"},
+		{name: "separate config file", args: []string{"--experimental-config-file", "config", "server.js"}, entryPoint: "server.js"},
```
pinoOgni, line 54: We can re-add the "default config file" (correct) case:
```
{name: "default config file", args: []string{"--experimental-default-config-file", "app"}, entryPoint: "app"},
``` 
So it guards the boolean flag from ever being mistakenly added to `optionsWithValues`

pinoOgni (changes requested):

pkg/internal/nodejstools/launch_test.go
MrAlias: Restored the `--experimental-default-config-file app` case in `TestParseNodeLaunch`. It verifies that the boolean flag does not consume the application entrypoint, alongside the separate-value `--experimental-config-file` regression case.

MrAlias:
Closing in favor of #3200

PR #3167, opened 2026-08-25
Add Aerospike server spans

pinoOgni:
## Summary

This PR allows produce Aerospike server spans (it follows the redis/sql logic).

Notes:
- server spans report the response `result_code` without any capture buffer configuration (the server writes responses in a single send).
- also fixes a pre-existing gap: `EventTypeAerospikeClient` was missing from the service-graph kind mapping, so aerospike client spans have had the wrong service-graph kind since the parser landed.
- a new integration suite instruments the asd container and asserts the five server spans end to end, including a `KEY_EXISTS_ERROR` one.

Follow-up:
- metric routing is not included: server events don't yet observe `db.server.operation.duration`. That lands in a follow-up PR.

## Validation

### DIFF
--- a/devdocs/protocols/tcp/aerospike.md
+++ b/devdocs/protocols/tcp/aerospike.md
@@ -148,6 +148,20 @@ plus, for writes, the per-op type bytes:
 The span name follows the OTel database convention `{operation} {target}`, e.g.
 `GET test.users` (`{db.operation.name} {db.namespace}.{db.collection.name}`).
 
+### Client- and server-side spans
+
+The span type follows the observed role: instrumenting an application that uses
+an Aerospike client produces `client` spans, while instrumenting the Aerospike
+server process (`asd`) produces `server` spans. Both roles share the span name
+convention and the attribute set above. When both peers are instrumented, one
+operation yields one client span and one server span with distinct kinds — the
+server span nests under the client span in the same trace.
+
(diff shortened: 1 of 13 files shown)


NimrodAvni78 (commented):
do we allow getting only client spans and not server spans for other protocols?
if not i think we should and maybe even default to only client ones
WDYT?

pinoOgni:
> do we allow getting only client spans and not server spans for other protocols? if not i think we should and maybe even default to only client ones WDYT?

Right now role selection happens through process selection, for every protocol: server spans only appear if obi is attached to the server process, and discovery selectors can exclude it. The `instrumentations:` list is role-blind, so client-only is achievable today; just via process selection, not via a flag.

That said, I'm thinking that one process can emit spans of several kinds, and there process exclusion is too blunt:
- obi can be deployed as a daemonset: you may still want telemetry from the `aerospike` pods, but not one server span per db operation.
- a process that is both server and client like a proxy, or a db node calling other nodes where you want one role's spans but not the other's.

So what about doing it as a span-kind filter rather than a per-protocol flag: it operates on a property every span already has, so one implementation covers all protocols.

We can discuss it offline and I can open an issue.

MrAlias (approved):

grcevski (approved):
LGTM!

PR #3168, opened 2026-08-25
remove service.name and service.namespace from default metric labels

NimrodAvni78:
## Summary

OBI reports `service.name` and `service.namespace` twice: as resource attributes, which the semantic conventions require, and again as metric-level attributes. This makes the metric-level copy opt-in instead of on by default, so what OBI emits matches the conventions.

The resource attributes are untouched.

## Restoring the old behaviour
For users who want/need this behavior:

`attributes.extra_group_attributes` gains an `app` group:

```yaml
attributes:
  extra_group_attributes:
    app: [service.name, service.namespace]
```

`attr_defs.go` already passed `extraGroupAttributes[GroupApp]` into the group, so only the parser case and the config schema enum were missing. Distributions that need the flattened form restore it with config rather than a patch.

## Scope

This affects both exporters, not just OTLP. `spanPromGetters` delegates to `spanOTELGetters`, so the two share one set of defaults, and there is no per-exporter attribute selection today.

**This is a breaking change for Prometheus users.** `service_name` and `service_namespace` are no longer emitted as labels on application metric series, so existing dashboards, alerts and recording rules that filter or group by them stop matching after an upgrade. Restore the previous behaviour with the `extra_group_attributes` config above, or migrate the queries to join through `target_info`.

Joining without the labels is possible today: the Prometheus exporter puts `job` and `instance` on `target_info`, and `job` is `namespace/name` (`svc.Attrs.Job()`), so service identity is recoverable.

### Families that keep the labels

Three code paths build their label sets directly instead of going through `AttrSelector`, so they are unaffected by this change and by `extra_group_attributes`:

| Family | Label source |
| --- | --- |
| `jvm_*`, `nodejs_*` runtime metrics | `runtimeServiceLabels()` |
| `go_*` runtime metrics | `labelNamesTargetInfo()` |
| `traces.span.metrics.*` data points | `spanMetricAttributes()` |

Work on them will be a follow up since it includes changes outside this scope

## Tests

- New coverage asserts the two attributes are absent from the default set, still present in the full set, and restored by the `app` group.
- Two existing expectations updated for the new default.
- `obi-config-perapp.yml` sets the restore key: that suite filters metric series by `service_name` and has no wildcard `select`, so it also keeps the opt-in path exercised in CI. The other suites that filter by `service_name` already use `select: {"*": {include: ["*"]}}`, which matches these attributes regardless of their default.

## Validation

### DIFF
--- a/pkg/export/attributes/attr_defs.go
+++ b/pkg/export/attributes/attr_defs.go
@@ -68,7 +68,7 @@ func getDefinitions(
 		map[attr.Name]Default{
 			attr.Instance:         true,
 			attr.Job:              true,
-			attr.ServiceNamespace: true,
+			attr.ServiceNamespace: false,
 		},
 		extraGroupAttributes[GroupPrometheus],
 	)
--- a/pkg/export/attributes/attr_defs.go
+++ b/pkg/export/attributes/attr_defs.go
@@ -243,9 +243,10 @@ func getDefinitions(
 		extraGroupAttributes[GroupAppKube],
 	)
 
-	// ServiceName and ServiceNamespace are reported both as resource and metric attributes, as
-	// the OTEL definition requires that it is reported as resource attribute,
-	// but Cloud providers may extract this from the metric
+	// The semantic conventions define service.name and service.namespace as
+	// resource attributes, and OBI reports them there. They are also available
+	// as metric-level attributes for backends that read service identity off
+	// the series instead of joining through target_info.
 	appAttributes := NewAttrReportGroup(
 		false,
 		[]*AttrReportGroup{
@@ -254,8 +255,8 @@ func getDefinitions(
 			&appContainerAttributes,
 		},
 		map[attr.Name]Default{
-			attr.ServiceName:      true,
-			attr.ServiceNamespace: true,
+			attr.ServiceName:      false,
+			attr.ServiceNamespace: false,
 		},
 		extraGroupAttributes[GroupApp],
 	)
--- a/devdocs/config/CONFIG.md
+++ b/devdocs/config/CONFIG.md
@@ -59,7 +59,7 @@ Attributes configures the decoration of some extra attributes that will be added
 
 | YAML Path | Type | Env Var | Default | Values | Deprecated | Description |
 |---|---|---|---|---|---|---|
-| `attributes.extra_group_attributes` | [`ExtraGroupAttributesMap`](#extragroupattributesmap) |  |  |  |  | Map of attribute group names to arrays of attribute names. Only 'k8s_app_meta' is currently supported as a key. |
+| `attributes.extra_group_attributes` | [`ExtraGroupAttributesMap`](#extragroupattributesmap) |  |  |  |  | Map of attribute group names to arrays of attribute names. Supported keys are 'app' and 'k8s_app_meta'. |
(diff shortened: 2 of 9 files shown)


mariomac (approved):
Well done!

pkg/export/attributes/attr_defs.go
```diff
@@ -243,9 +243,10 @@ func getDefinitions(
...
 	)
 
-	// ServiceName and ServiceNamespace are reported both as resource and metric attributes, as
-	// the OTEL definition requires that it is reported as resource attribute,
-	// but Cloud providers may extract this from the metric
+	// The semantic conventions define service.name and service.namespace as
+	// resource attributes, and OBI reports them there. They are also available
+	// as metric-level attributes for backends that read service identity off
+	// the series instead of joining through target_info.
 	appAttributes := NewAttrReportGroup(
```
pinoOgni, line 250: I see that `prometheusAttributes` still has `attr.ServiceNamespace: true` this means that  with Prom enabled, we are removing only `service_name`. Is it intentional?

pinoOgni (commented):
nit: the PR description "Prometheus users who filter metric series by service_name need the config above." can be modified to make it clearer that this PR changes the default exported labels so dashboards filtering on `service_name` break on upgrade. We should also add it to the release note.

pkg/export/attributes/attr_defs.go
NimrodAvni78: updated description and added a breaking change label

pkg/export/attributes/attr_defs.go
```diff
@@ -254,8 +255,8 @@ func getDefinitions(
 			&appContainerAttributes,
 		},
 		map[attr.Name]Default{
-			attr.ServiceName:      true,
-			attr.ServiceNamespace: true,
+			attr.ServiceName:      false,
```
MrAlias, line 258: These defaults only affect metric families built through `AttrSelector`. Prometheus runtime metrics bypass this group: Go derives labels from `labelNamesTargetInfo`, while JVM and Node.js use `runtimeServiceLabels`, which hardcodes both service labels. Those series therefore retain `service_name` and `service_namespace`, and the new `app` group cannot restore or remove them consistently.

Could we route the runtime collectors through their attribute definitions, or explicitly scope and document the exception? Please add Prometheus runtime coverage for the chosen behavior.

MrAlias (commented):
Prometheus label behavior remains inconsistent across metric families, so the documented breaking-change migration does not match the current output.

pkg/export/attributes/attr_defs.go
NimrodAvni78: Yeah apparently runtime metrics work through a different attribute selection process
I documented the scope to exclude runtime metrics for this change and take it as a follow up since its a bigger change

pkg/export/attributes/attr_defs.go
pinoOgni: I think I explained myself badly. 
Basically, `prometheusAttributes` is not modified here and it is a subgroup of `appAttributes`, active whenever Prom export is on, and it defaults `job`, `instance`, and `service_namespace` onto every metric series. The `Default()` (of `attr_selector.go`) merge only adds on `true`; a parent's `false` can't remove a subgroup's `true`. So, for Prometheus users this pr removes only `service_name` while `service_namespace` remains a per-series label, still "duplicated" relative to `target_info`. 

If it is intentional we should document it and change the PR title.

pinoOgni: Oh sorry, I didnt see the `attr.ServiceNamespace: false,` in `prometheusAttributes`! Nice!

pinoOgni (approved):
lgtm!

PR #3169, opened 2026-08-25
declare span-size metrics and span-metric data-point attributes

NimrodAvni78:
## Summary

Declares the two span-size counters OBI emits but the registry never described, and adds the data-point attributes the existing span metrics carry.

Two independent gaps:

- The five data-point attributes below were genuinely under-declared for span metrics that today's suites already emit.
- The span-size counters were emitted by no test configuration at all, because nothing enabled `application_span_sizes`. This PR turns that feature on in the per-app suite, so the declarations land together with the coverage that exercises them.

## What is declared

`traces_spanmetrics_size_total` and `traces_spanmetrics_response_size_total` — counters emitted by the `application_span_sizes` feature. Both keep underscore names on the OTLP side, so the emitted name and the Prometheus series name are identical. They are created via `Float64Counter` with no `instrument.WithUnit`, so both declare an empty unit even though they measure bytes.

## Data-point attributes

The four existing span-metric groups declared only their span dimensions. They also carry `service.name`, `service.instance.id`, `service.namespace`, `telemetry.sdk.language` and `host.id` as data-point attributes, because `spanMetricAttributes()` builds an explicit attribute set rather than relying on the resource. Those five are now declared on every span-metric group, including the two new ones.

## Test coverage

- `perapp_features_test.go` asserts both size counters for all six instrumented services. The queries pin the `exported` label, so the OTEL and Prometheus subtests cannot pass on each other's series — the per-app Prometheus scrapes both `otelcol:9464` and `obi:8999`.
- `obi-config-perapp.yml` enables `application_span_sizes` globally so every service emits the size counters.

## Validation


schemas/obi/groups/spanmetrics.yaml
```diff
@@ -131,3 +147,50 @@ groups:
       - ref: span.kind
       - ref: status.code
       - ref: source
+      - ref: service.name
+      - ref: service.instance.id
+      - ref: service.namespace
+      - ref: telemetry.sdk.language
+      - ref: host.id
+
+  - id: metric.traces_spanmetrics_size_total
```
MrAlias, line 164: When this lands on current `main`, these two groups remain ordinary `development` metrics even though `FeatureSpanSizes` and `application_span_sizes` are deprecated with no replacement, and the adjacent legacy groups are marked deprecated. The overview also says span metrics are emitted by `application_span_otel`, which does not enable these counters. That leaves the resolved registry inconsistent about both lifecycle and gating.

Could we mark both size metrics `deprecated` with `reason: obsoleted` and a no-replacement note, then update the span-metrics overview and manifest to document the separate `application_span_sizes` gate?

MrAlias (commented):
The emitted attribute set and counter definitions match the exporters. The remaining issue is keeping the registry's lifecycle and feature-gating contract consistent with `main`.

MrAlias (approved):

grcevski (approved):
LGTM!

schemas/obi/groups/spanmetrics.yaml
NimrodAvni78: done!

PR #3146, opened 2026-08-24
bpf: classify TCP client/server by the local port, not the sorted d_port

mmat11:
## Summary

`handle_unknown_tcp_connection` classified a connection as server-side by asking `is_listening` about the sorted `d_port`, so a client with any local listener on that port got `is_server=true` and its large buffers stored with request/response packet types swapped, dropping generic TCP spans (see also #3116); fixed by checking the local port instead

## Validation

- [x] I have read and followed the [contributing guidelines](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/CONTRIBUTING.md)
- [ ] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed.

PR #3170, opened 2026-08-25
pids: keep blocked PIDs valid for a grace period so their final events drain

mmat11:
## Summary

Events from an exiting process can sit in the ringbuffer for up to the reader flush interval (3s). Process exit triggers `BlockPID()`, which removes the PID immediately, so those buffered events were silently dropped by `ValidPID()`.

A short-lived client (one-shot mongosh, psql, a migration job) loses all of its spans; a long-running service loses the last ones before shutdown. Found while investigating the missing mongosh traces reported in #3146.

`BlockPID` now marks the entry instead of deleting it: events keep passing for a 5s grace period, then expire.

## Validation

### DIFF
--- a/pkg/ebpf/common/pids.go
+++ b/pkg/ebpf/common/pids.go
@@ -33,9 +33,10 @@ var readNamespacePIDs = procs.FindNamespacedPids
 
 var pidsFilterMonoNow = timing.MonoTimeNow
 
-// removed entries are kept this long so spans captured before the removal can
-// still drain through the reader and batching stages; it only bounds memory,
-// admission is decided by comparing span timestamps against removedAt
+// Spans are filtered as soon as the ring buffer reader parses them, at most two
+// flush intervals after capture. Removed entries stay this long so those spans
+// still find their process; this only bounds memory, admission is decided by
+// comparing span timestamps against removedAt
 const pidRemovalRetention = 5 * time.Minute
 
 type PIDInfo struct {
@@ -45,6 +46,10 @@ type PIDInfo struct {
 	// zero while the process lives; the CLOCK_MONOTONIC instant it was
 	// blocked at otherwise, comparable against BPF span timestamps
 	removedAt time.Duration
+	// the process that held this pid before, kept while its spans may still arrive
+	previous *PIDInfo
+	// when that previous process was removed: older spans are not this process's
+	since time.Duration
 }
 
 type ServiceFilter interface {
@@ -98,18 +103,32 @@ func (pf *PIDsFilter) ValidPID(userPID app.PID, ns uint32, pidType PIDType) bool
 
 	if ns, nsExists := pf.current[ns]; nsExists {
 		if info, pidExists := ns[userPID]; pidExists {
-			return info.pidTypes&pidType != 0
+			// any generation still draining keeps its tracer type valid, so its
+			// last events are parsed; Filter then picks the generation by time
+			for g := &info; g != nil; g = g.previous {
+				if g.pidTypes&pidType != 0 {
+					return true
+				}
+			}
 		}
 	}
 
 	return false
 }
 
-// spans of a removed pid still pass if captured while it lived: they were
-// merely waiting in the reader and batching stages. Later timestamps mean the
-// pid was reused by a process we were not told to track
-func capturedWhileAllowed(info *PIDInfo, span *request.Span) bool {
-	return info.removedAt == 0 || span.End <= int64(info.removedAt)
+// The generation of the process that was alive when the span was captured, nil
+// when there is none: the span is newer than the last known process, or older
+// than the ones still kept
+func generationFor(info *PIDInfo, span *request.Span) *PIDInfo {
+	for g := info; g != nil; g = g.previous {
+		if g.removedAt != 0 && span.End > int64(g.removedAt) {
+			return nil
+		}
+		if g.since == 0 || span.End > int64(g.since) {
+			return g
+		}
+	}
+	return nil
 }
 
 func expired(info *PIDInfo, now time.Duration) bool {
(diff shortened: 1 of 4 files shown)


daniel-percona:
Thank you for working on this. Here is a follow-up response to the request from https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3146#issuecomment-5411188717. This is the same workflow I used in that PR, but with this fix instead. Steps below and logs attached:

config: `/etc/obi-mongosh-fixed.yaml`

```
discovery:
  instrument:
    - exe_path: /usr/bin/mongosh

log_level: DEBUG
trace_printer: text

ebpf:
  max_transaction_time: 5s
  protocol_debug_print: true
  buffer_sizes:
    tcp: 4096

meter_provider:
  features:
    - application

otel_traces_export:
  endpoint: http://127.0.0.1:4318
  protocol: http/protobuf
  instrumentations:
    - mongo

prometheus_export:
  port: 8999
  path: /metrics
  instrumentations:
    - mongo
``` 


```
$ sudo /usr/local/bin/obi-3170 -config=/etc/obi-mongosh-fixed.yaml 2>&1 | tee /tmp/obi-3170.log
```

Workload:

```
$ mongosh 'mongodb://*****:*****@127.0.0.1:27018/obi_mongosh_test?authSource=admin&directConnection=true' --quiet --eval '
print("OBI_TEST_START_WAITING");");
sleep(15000);obi_shell_test.insertOne({source: "obi-matt-test", sequence_id: Date.now()}));
print("OBI_TEST_RUNNING_COMMANDS");ource: "obi-matt-test"}).limit(1).toArray());
printjson(db.obi_shell_test.insertOne({source: "obi-matt-test", sequence_id: Date.now()}));tt-test"}}));
printjson(db.obi_shell_test.find({source: "obi-matt-test"}).limit(1).toArray());
printjson(db.obi_shell_test.updateOne({source: "obi-matt-test"}, {$set: {updatedBy: "obi-matt-test"}}));
printjson(db.obi_shell_test.deleteOne({source: "obi-matt-test"}));
print("OBI_TEST_END");
'
OBI_TEST_START_WAITING
OBI_TEST_RUNNING_COMMANDS
{
  acknowledged: true,
  insertedId: ObjectId('6a8da04d6988a5db283e8d33')
}
[
  {
    _id: ObjectId('6a8da04d6988a5db283e8d33'),
    source: 'obi-matt-test',
    sequence_id: 1787666509477
  }
]
{
  acknowledged: true,
  insertedId: null,
  matchedCount: 1,
  modifiedCount: 1,
  upsertedCount: 0
}
{
  acknowledged: true,
  deletedCount: 1
}
OBI_TEST_END
```

logs:

```
strings /tmp/obi-3170.log | grep -E 'instrumenting process.*mongosh|obi-matt-test|obi_shell_test|MongoClient|labelValues="\[ (insert|find|update|delete) mongodb '
time=2026-08-25T10:01:41.741-04:00 level=INFO msg="instrumenting process" component=discover.traceAttacher cmd=/usr/bin/mongosh pid=573692 ino=12594237 type=cpp service="" logenricher=false
obi_shell_test
obi-matt-test
obi_shell_test
obi-matt-test
obi-matt-test
obi_mongosh_test.obi_shell_test
obi_shell_test
obi-matt-test
obi-matt-test
obi_shell_test
obi-matt-test
```

jaeger:

```
$ curl -s 'http://127.0.0.1:16686/api/services' | jq .
{
  "data": [
    "jaeger",
    "obi-go-mongo-test",
    "mongosh"
  ],
  "total": 3,
  "limit": 0,
  "offset": 0,
  "errors": null
}


$ curl -s 'http://127.0.0.1:16686/api/traces?service=mongosh&limit=20' | jq .
{
  "data": [],
  "total": 0,
  "limit": 0,
  "offset": 0,
  "errors": null
}


$ for op in insert find update delete; do   echo "=== mongosh operation: $op ===";   curl -s "http://127.0.0.1:16686/api/traces?service=mongosh&operation=$op&limit=20" | jq '.data | length'; done
=== mongosh operation: insert ===
0
=== mongosh operation: find ===
0
=== mongosh operation: update ===
0
=== mongosh operation: delete ===
0
```

Full logs attached: 
[obi3170.log.tar.gz](https://github.com/user-attachments/files/31423426/obi3170.log.tar.gz)

daniel-percona:
The above was ran without combining both PRs. Once I combined them both it worked, please see https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3146#issuecomment-5412797819

pkg/ebpf/common/pids.go
```diff
@@ -29,10 +30,20 @@ const (
 // choose to filter traces using whether the User-space or Host-space PIDs
 var readNamespacePIDs = procs.FindNamespacedPids
 
+var pidsFilterNow = time.Now
+
+// blocked pids stay valid for this long: events of an exiting process can sit
+// in the kernel ring buffers up to the reader flush interval, and dropping the
+// pid immediately silently discards its final spans
+const pidRemovalGracePeriod = 5 * time.Second
```
MrAlias: `flushInterval` is not a three-second upper bound here. Because the ticker flushes only after `lastReadAt` is already older than one interval, a low-volume record arriving just after another read can wait until the following tick—almost six seconds—before parsing. `Filter` then runs again only when the userspace batch flushes, adding the configurable `batch_timeout` (or waiting for a full batch when it is zero). This five-second expiry can therefore drop the same final spans this change is meant to preserve.

Could we tie PID retirement to the actual ring-buffer and batch drain, and cover a record delayed across both stages?

pkg/ebpf/common/pids.go
```diff
@@ -180,6 +195,7 @@ func (pf *PIDsFilter) addPID(pid app.PID, nsid uint32, fi *exec.FileInfo, t PIDT
 		pidInfo.fileInfo = fi
 		pidInfo.pidTypes |= t
 		pidInfo.otherKnownPids = allPids
+		pidInfo.removedAt = time.Time{}
```
MrAlias: `pidInfo` may be a tombstone for a different process that reused this PID. Clearing `removedAt` revives its old `pidTypes` too, so a generic-to-Go reuse can re-enter the generic BPF PID filter on the next rebuild even though the new process was registered only for Go tracing.

Could we replace tombstoned state on `AllowPID`, while continuing to OR tracer types only for a live entry, and cover cross-type reuse?

MrAlias (commented):
The tombstone direction addresses the immediate exit race, but its expiry and revival rules still need to match the reader pipeline and PID reuse lifecycle.

pkg/ebpf/common/pids.go
MrAlias: The timestamp comparison is the right direction, and resetting `PIDInfo` fixes the cross-type revival I called out. My earlier suggestion to replace the tombstone was incomplete, though: if the PID is reused before an old buffered event reaches `Filter`, resetting the entry discards the old process's cutoff and `fileInfo`. That event then passes against the new live entry and is attributed to the new service.

The invariant we need is that generation-specific metadata survives until the event is filtered. The fixed five-minute retention does not guarantee that while `batch_timeout` permits zero or values above five minutes. Could we either filter and decorate before userspace batching while retaining the previous generation across PID reuse, or explicitly drain and acknowledge the reader and pending batch before retiring it?

The tests should cover a pre-block span delivered after the retention window and a pre-block span filtered after the same PID is re-allowed. Both should retain the old service attributes, while a post-reuse span should receive the new attributes.

grcevski (approved):
LGTM!

mmat11:
@MrAlias can you give another pass here? I think I've addressed everything that was left

MrAlias (approved):

PR #3177, opened 2026-08-25
fix: reject inverted ranges in IntEnum port and PID config parsing

immanuwell:
## Summary

`open_port` / `target_pids` / `open_ports:` accept inverted ranges like `9000-1000`, no error. 
regex only checks shape, not order, so it just matches nothing, forever, silently

Repro:
```
OTEL_EBPF_OPEN_PORT=9000-1000 bin/obi
```
Parses fine, but `Matches()` needs `Start <= n <= End`, never true when `End < Start`. 
OBI starts up happy and just never finds the process

Fix: reject `end < start` in `UnmarshalText`, next to the existing syntax check. 
Added a test case for it

## Validation

- [ ] not a core feature change, no docs update needed

### DIFF
--- a/pkg/appolly/services/criteria.go
+++ b/pkg/appolly/services/criteria.go
@@ -270,6 +270,9 @@ func (p *IntEnum) UnmarshalText(text []byte) error {
 		e.Start, _ = strconv.Atoi(strings.TrimSpace(parts[0]))
 		if len(parts) > 1 {
 			e.End, _ = strconv.Atoi(strings.TrimSpace(parts[1]))
+			if e.End < e.Start {
+				return fmt.Errorf("invalid int enum %q: range end must not be less than start (%d-%d)", val, e.Start, e.End)
+			}
 		}
 		p.Ranges = append(p.Ranges, e)
 	}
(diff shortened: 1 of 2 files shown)


pkg/appolly/services/criteria.go
```diff
@@ -270,6 +270,9 @@ func (p *IntEnum) UnmarshalText(text []byte) error {
 		e.Start, _ = strconv.Atoi(strings.TrimSpace(parts[0]))
 		if len(parts) > 1 {
 			e.End, _ = strconv.Atoi(strings.TrimSpace(parts[1]))
+			if e.End < e.Start {
```
MrAlias, line 273: Both `strconv.Atoi` errors are discarded, so this ordering check does not reject every inverted range accepted by the regex. For example, `999999999999999999999-888888888888888888888` overflows both endpoints to `MaxInt`; the comparison sees equal values, `UnmarshalText` returns nil, and the invalid configuration becomes a one-value range matching `MaxInt`.

Please reject either conversion error before comparing the endpoints and add coverage for overflowing range bounds.

MrAlias (commented):
The ordinary inverted-range case is fixed, but overflowing bounds leave the validation incomplete.

MrAlias:
@immanuwell are you still able to work on this?

PR #3181, opened 2026-08-25
emit obi.queue.capacity.ratio and declare it

NimrodAvni78:
## Summary

`obi.queue.capacity.ratio` has never been emitted. The instrument, the `imetrics.Reporter` method and both exporters all exist, and `Send` calls the gauge on every message — but `msg.NewQueue` defaults it to a no-op stub, and nothing in production ever passed `msg.InternalMetrics(...)` to replace it.

This wires the reporter through, corrects the ratio, names the remaining subscribers, and declares the metric.

## Changes

- `QueueFromConfig` takes an `imetrics.Reporter` and passes `msg.InternalMetrics` to every queue it builds. All 48 call sites already had `ctxInfo.Metrics` in scope. A nil reporter is tolerated, since some callers build a `global.ContextInfo` without one.
- The ratio is now `unread/capacity`, was `(unread+1)/capacity` — which never read 0 on an idle queue and reached 1.02 at the default buffer of 50, i.e. it was wrong exactly where the metric matters. Both the declared brief and the `Reporter` doc already specified `[0-1]`.
- Named the three subscribers that passed no `msg.SubscriberName`: `transform.DockerProcessEventDecorator` and `flow.ProtocolFilter` (both branches). They shared a single `(unnamed)` series, so Docker-without-Kubernetes running a netolly protocol filter had two unrelated queues writing to the same series. An unnamed `Subscribe` now falls back to the queue's name.
- Declared `obi.queue.capacity.ratio` and its `subscriber` attribute in the registry.

## Tests

`TestQueueFromConfig_ReportsBufferUtilization` fails against the old factory, which reported nothing. `TestMetrics_CapacityGauge_Saturation` fills a buffer and pins the gauge at exactly 1, failing against the old `+1` expression; existing gauge tests had `+1` baked into their expected values and were corrected. `obi_queue_capacity_ratio` is added to `internalMetricsExpected`, so the integration suite proves the series reaches Prometheus rather than only that the recorder was called.

## Behaviour change

Deployments with internal metrics enabled start receiving a new gauge, one series per subscriber name, sampled on every send. `NoopReporter` deployments are unaffected. No compatibility concern: the metric was never emitted, so no existing series changes meaning.

## Validation

### DIFF
--- a/internal/test/integration/internal_metrics_otel_test.go
+++ b/internal/test/integration/internal_metrics_otel_test.go
@@ -32,6 +32,7 @@ var internalMetricsExpected = []string{
 	"obi_bpf_map_entries",                 // gauge, per eBPF map
 	"obi_bpf_map_max_entries",             // gauge, per eBPF map
 	"obi_bpf_probe_latency_seconds_count", // histogram, needs BPF run-stats + traffic
+	"obi_queue_capacity_ratio",            // gauge, per pipeline queue subscriber
 }
 
 // TestInternalOTelMetrics brings up OBI with internal_metrics.exporter=otel so
--- a/pkg/appolly/discover/finder.go
+++ b/pkg/appolly/discover/finder.go
@@ -85,7 +85,7 @@ func (pf *ProcessFinder) Start(ctx context.Context, opts ...ProcessFinderStartOp
 		opt(&startConfig)
 	}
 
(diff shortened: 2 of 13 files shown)


MrAlias (approved):

grcevski (approved):
LGTM!

PR #3184, opened 2026-08-25
Bind Java injection to stable process identities

MrAlias:
## Summary

- Capture process start time during executable inspection and acquire a stable process descriptor before tracer setup or queueing.
- Open credentials, namespaces, working directories, temporary directories, and target roots relative to that descriptor, and signal through pidfd_send_signal.
- Abort and join canceled HotSpot and OpenJ9 attach work, including response and detach reads.
- Preserve the existing bounded FIFO queue while closing process descriptors on every admission and shutdown path.

## Why

The asynchronous attach path added around #2949 validated process start time only after potentially slow tracer setup. It then used the numeric PID again for credential lookup, namespace entry, target-root writes, and SIGQUIT. A recycled PID could therefore redirect later attach operations after a successful validation. Cancellation also returned from the serialized worker while an inner scanner or socket read could remain active.

This change establishes one invariant: Java injection proceeds only when the process identity captured during inspection can be converted into a stable proc descriptor before asynchronous work begins. All PID-sensitive resources and signals come from that descriptor, and cancellation does not return until the inner attach work has stopped. Systems without stable descriptor signaling fail closed. The implementation does not depend on pidfd_open, preserving compatibility with supported 4.18 kernel backports that expose pidfd_send_signal.

## Validation

- `make build`
- `make verify`
- `make compile`


pkg/internal/jvmtools/jvm/cmd.go
```diff
@@ -238,14 +301,15 @@ func (j *JAttacher) attachInNamespace(ctx context.Context, pid, nspid, targetUID
...
 
-	tmpPath := util.GetTmpPath(attachPid)
-
 	// Make write() return EPIPE instead of abnormal process termination
 	signal.Ignore(syscall.SIGPIPE)
 
 	if err := ctx.Err(); err != nil {
 		return nil, err
 	}
+	if err := process.Alive(); err != nil {
```
grcevski, line 310: This is really nitpicking, but theoretically (I mean really theoretically) we can check here and then the process can die and new one can replace it. We could check if the process.Alive after we attach but before we send the command.

pkg/appolly/discover/java_injection_queue.go
```diff
@@ -64,11 +68,31 @@ func (q *javaInjectionQueue) start(ctx context.Context) {
 					q.log.Warn("unable to attach java agent to process, Java TLS telemetry will not work",
 						"pid", target.Pid, "error", err)
 				}
+				if err := target.Close(); err != nil {
```
grcevski: Do we also need target.Close() on ctx.Err() != nil?

pkg/internal/jvmtools/jvm/cmd.go
MrAlias: Added a second liveness check after the attach socket is established and immediately before the command write. Both the HotSpot and OpenJ9 paths now call `process.Alive()` before transmitting, with focused coverage in `cmd_hotspot_test.go` and `cmd_openj9_test.go`.

pkg/appolly/discover/java_injection_queue.go
MrAlias: Dequeued target ownership now lives in `injectTarget`, which defers `target.Close()` before checking `ctx.Err()`. The process handle is therefore closed whether injection runs or cancellation wins after channel selection. `TestJavaInjectionQueue_ClosesDequeuedTargetAfterCancellation` covers that path.

grcevski (approved):
LGTM!

PR #3189, opened 2026-08-26
Fix confmap decoding of YAML sequences into text-unmarshalable types

### LINKED ISSUE #3188: metrics.features as a YAML list fails when OBI runs as a collector receiver
### Component

OBI (eBPF instrumentation)

### What happened?

Running OBI as a collector receiver, giving `metrics.features` as a YAML list makes config loading fail, so the collector never starts:

```
'metrics.features' cannot parse value as 'export.Features': strconv.ParseUint: invalid syntax
```

The same value as a comma-separated string (`features: "application,application_span_otel"`) works.
The list form also works in standalone mode, when the config is loaded from a YAML file. Only the receiver path rejects it.

The list form should be accepted:

- `Features.JSONSchema()` declares it as `array` (`pkg/export/feature.go:149`)
- `Features.UnmarshalYAML` accepts only a sequence (`pkg/export/feature.go:218`)
- The repo's own v1 example uses a list (`devdocs/config/version-2.0/examples/migration-v1.yaml:17`)

### Steps to reproduce

1. Build a collector that includes the obi receiver (`examples/otel-collector`).
2. Start it with the config below (legacy layout, no `version: "2.0"`).
3. It fails while loading the configuration.

Without a collector or a cluster:

```go
cfg := obi.DefaultConfig
err := cfg.Unmarshal(confmap.NewFromStringMap(map[string]any{
    "metrics": map[string]any{"features": []any{"application", "application_span_otel"}},
}))
// err: 'metrics.features' cannot parse value as 'export.Features': strconv.ParseUint: invalid syntax

// the comma-separated form decodes fine
cfg = obi.DefaultConfig
err = cfg.Unmarshal(confmap.NewFromStringMap(map[string]any{
    "metrics": map[string]any{"features": "application,application_span_otel"},
}))
// err == nil
```

### OBI configuration

```yaml
receivers:
  obi:
    metrics:
      features:
        - application
        - application_span_otel
```

### Relevant logs

```shell
'' parse legacy OBI receiver config: decoding failed due to the following error(s):

'metrics.features' cannot parse value as 'export.Features': strconv.ParseUint: invalid syntax
```

### Additional context

Two decode hooks in `(*Config).Unmarshal` (`pkg/obi/config.go:491`) are registered in the wrong order.

| hook | what it does |
|---|---|
| `TextUnmarshallerHookFunc` | when the value is a **string**, calls the target's `UnmarshalText` |
| `stringSliceToTextUnmarshalerHookFunc` | joins a **list of strings** with commas |

Handling a list needs *join, then `UnmarshalText`*. The registration is the other way round:

```go
DecodeHook: mapstructure.ComposeDecodeHookFunc(
    mapstructure.StringToTimeDurationHookFunc(),
    mapstructure.TextUnmarshallerHookFunc(),        // hands strings to UnmarshalText
    stringSliceToTextUnmarshalerHookFunc(),         // joins the list into a string  <- too late
    inlineMetadataHookFunc(),
),
```

`ComposeDecodeHookFunc` feeds each hook the previous one's output. A list therefore passes straight through `TextUnmarshallerHookFunc` (not a string), and the comma-separated string produced after it has no hook left to consume it. `Features.UnmarshalText` never runs, and `WeaklyTypedInput` ends up parsing the string as a `uint`.

`stringSliceToTextUnmarshalerHookFunc` was added for exactly this list form (`pkg/obi/config.go:555`):

> handles types like Features and ExportModes that have UnmarshalYAML for YAML sequences but also
> support comma-separated text via UnmarshalText

So this is an ordering slip rather than a design decision, unchanged since #1053.
`metrics.features` is just the case we hit — any type implementing `UnmarshalText` alongside a
sequence-accepting `UnmarshalYAML` breaks the same way.

**Suggested fix** — swap the two hooks:

```go
DecodeHook: mapstructure.ComposeDecodeHookFunc(
    mapstructure.StringToTimeDurationHookFunc(),
    stringSliceToTextUnmarshalerHookFunc(),
    mapstructure.TextUnmarshallerHookFunc(),
    inlineMetadataHookFunc(),
),
```

The list is joined first and the next hook hands the result to `UnmarshalText`; the comma-separated form passes through the first hook untouched and keeps working. Happy to open a PR with the reorder plus a `Config.Unmarshal` regression test covering both shapes.

The hook order is unchanged from v0.5.0, where it was introduced, through current `main`, so every release shipping the collector receiver is affected.

### Version

v0.10.0

### Kernel version

_No response_

### OS / Distribution

_No response_

### Deployment

Kubernetes

elliehmy:
Fixes #3188

## What

Swap the registration order of two decode hooks.

`stringSliceToTextUnmarshalerHookFunc` ran after `mapstructure.TextUnmarshallerHookFunc`.
`ComposeDecodeHookFunc` feeds each hook the previous one's output, so the comma-separated string the slice hook produces had no hook left to pass it to `UnmarshalText`.

## Why

Running OBI as a collector receiver, giving `metrics.features` as a list stops the collector from starting.

```yaml
receivers:
  obi:
    metrics:
      features:
        - application
        - application_span_otel
```

```
'metrics.features' cannot parse value as 'export.Features': strconv.ParseUint: invalid syntax
```

The list is the shape `Features.JSONSchema()` declares as `array` and `Features.UnmarshalYAML` accepts. It works in standalone mode; only the receiver path rejects it.

The slice hook was added to support this shape in the first place — its doc comment names `Features` and `ExportModes` — so this is an ordering slip rather than intended behaviour, unchanged since #1053.

Any type implementing `UnmarshalText` alongside a sequence-accepting `UnmarshalYAML` breaks the same way.

## Tests

`TestUnmarshalConfmapSequences` covers the sequence and the comma-separated form for both types the hook's doc comment names — `metrics.features` and `discovery.instrument[].exports`. Both sequence subtests fail without the reorder.

#### Generative AI disclosure

Per AI-POLICY.md §5: an AI assistant was used to investigate the decode-hook chaining, produce the reproduction, and draft the test. I reviewed the change and the test, confirmed the failure and the fix against `pkg/obi` and `collector`, and take responsibility for the contribution.

### DIFF
--- a/pkg/obi/config.go
+++ b/pkg/obi/config.go
@@ -501,8 +501,11 @@ func (c *Config) Unmarshal(component *confmap.Conf) error {
 		WeaklyTypedInput: true,
 		DecodeHook: mapstructure.ComposeDecodeHookFunc(
 			mapstructure.StringToTimeDurationHookFunc(),
-			mapstructure.TextUnmarshallerHookFunc(),
+			// ComposeDecodeHookFunc feeds each hook the previous hook's output, so the
+			// slice-joining hook must run before TextUnmarshallerHookFunc, which only
+			// fires on strings.
 			stringSliceToTextUnmarshalerHookFunc(),
+			mapstructure.TextUnmarshallerHookFunc(),
 			inlineMetadataHookFunc(),
 		),
 	})
--- a/pkg/obi/config_test.go
+++ b/pkg/obi/config_test.go
@@ -21,6 +21,8 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"go.opentelemetry.io/collector/confmap"
+
 	"go.opentelemetry.io/obi/pkg/appolly/meta"
 	"go.opentelemetry.io/obi/pkg/appolly/services"
 	"go.opentelemetry.io/obi/pkg/config"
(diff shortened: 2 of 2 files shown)


MrAlias (approved):

grcevski (approved):
LGTM!

PR #3191, opened 2026-08-26
fix(export): surface the span's remote parent to the trace sampler

### LINKED ISSUE #3190: parentbased_* trace samplers never see the remote parent: ShouldSample receives the exporter lifecycle context
## What happened

With `track_request_headers: true` and a `parentbased_always_off` trace sampler, OBI exports **zero spans** — even when every incoming request carries a sampled W3C `traceparent` (in our case injected by an Envoy gateway tracing at 100%).

Switching only the sampler to `always_on` makes the same spans appear in the backend **correctly joined to the incoming trace** (correct trace ID, `parentSpanId` pointing at the gateway's client span). So the remote parent is read and propagated onto the exported spans — but the *sampling decision* never learns about it.

## Root cause

`pkg/export/otel/traces.go`:

```go
func (tr *tracesOTELReceiver) processSpans(ctx context.Context, exp exporter.Traces, spans []request.Span, ...) {
	spanGroups := tracesgen.GroupSpans(ctx, spans, traceAttrs, sampler, ...)
```

`pkg/export/otel/tracesgen/tracesgen.go`:

```go
sr := spanSampler().ShouldSample(trace.SamplingParameters{
	ParentContext: ctx,           // <-- the receiver's lifecycle context, same for every span
	Name:          span.TraceName(),
	TraceID:       span.TraceID,
	...
})
```

`ParentContext` is the long-lived context of the export loop, which never contains a span context. The span's own remote parent (populated from the incoming `traceparent` into the `request.Span`) is used for span assembly but never surfaced to the sampler. As a result `ParentBased(...)` always takes its **root** branch:

- `parentbased_always_off` → drops 100% of spans, including ones that would join an existing sampled trace;
- `parentbased_always_on` → indistinguishable from `always_on`;
- `parentbased_traceidratio` → ratio-samples spans that belong to already-sampled traces, producing torn traces.

In other words, all `parentbased_*` sampler names currently behave exactly like their root halves, which defeats their purpose and breaks the natural configuration for gateway-anchored sampling ("let the edge decide, OBI follows").

## Expected

Before calling `ShouldSample`, construct the remote parent from the span's parent data when present, e.g.:

```go
parentCtx := ctx
if span.ParentSpanID.IsValid() {
	parentCtx = trace.ContextWithRemoteSpanContext(ctx, trace.NewSpanContext(trace.SpanContextConfig{
		TraceID:    span.TraceID,
		SpanID:     span.ParentSpanID,
		TraceFlags: trace.FlagsSampled, // or the flags carried by the incoming traceparent
		Remote:     true,
	}))
}
```

(The incoming sampled flag would ideally be carried on `request.Span` alongside the IDs if it isn't already.)

## Environment / evidence

- Beyla v3.28.0 (executed by Grafana Alloy v1.19.0 as a subprocess), kernel 6.8 (AKS), `track_request_headers: true`, `context_propagation: disabled`.
- With `parentbased_always_off`: OBI's own `http_server_request_duration_seconds_count` counted 98 gateway-routed requests on one node while `beyla_otel_trace_exports_total` stayed at 0.
- With `always_on` and no other change: the same workload's spans appear in Tempo joined to the gateway's traces (verified `parentSpanId` == the gateway's client span).

Happy to provide fuller traces or test the fix.

tete17:
## Summary

Fixes #3190.

`GroupSpans` passed the exporter's lifecycle context as `ParentContext` to every `ShouldSample` call, so the sampler never saw the remote parent that the span itself carries. All `parentbased_*` samplers therefore took their **root** branch for every span:

- `parentbased_always_off` dropped 100% of spans — including spans joining an already-sampled trace (our production case: an Envoy gateway sampling at 100% and injecting `traceparent`, OBI configured to follow the edge decision → zero spans exported while OBI's own RED metrics counted the requests);
- `parentbased_always_on` was indistinguishable from `always_on`;
- `parentbased_traceidratio` ratio-sampled spans belonging to already-sampled traces, producing torn traces.

The fix builds the remote parent `SpanContext` from data the span already carries — `TraceID`, `ParentSpanID`, and `TraceFlags` (the incoming traceparent's flags byte, parsed by the BPF side) — and hands it to the sampler when a parent is present. Root spans still present no parent, so root-half behavior is unchanged, and `normalizeTraceContext` already guarantees roots get `TraceFlags=1` with no `ParentSpanID`.

One deliberate choice worth reviewer attention: the parent is marked `Remote: true` unconditionally. For spans whose parent is another eBPF-observed span of the same process there is no in-process OTel context either way, and the default `ParentBased` composition treats local and remote sampled parents identically, so the decision is unaffected; distinguishing them would require plumbing OBI doesn't have today.

## Validation

- [x] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed. *(No doc change needed — this makes the already-documented `parentbased_*` sampler names behave as documented.)*

New tests in `tracesgen_test.go`:

- `TestGroupSpansSamplerSeesRemoteParent` — the sampler receives a valid, remote, sampled parent for joined spans and no parent for roots;
- `TestGroupSpansParentBasedSampling` — `parentbased_always_off` keeps joined-sampled, drops joined-unsampled, drops roots; `parentbased_always_on` keeps roots.

Both fail on `main` without the fix and pass with it (`go test ./pkg/export/otel/tracesgen/`).

Real-world validation: with Beyla v3.28.0 (via Grafana Alloy v1.19.0), the same workload that exported zero spans under `parentbased_always_off` exports correctly-joined spans under `always_on` — the spans carry the right parent, only the sampling decision was blind to it. Full evidence in #3190.

### DIFF
--- a/internal/test/integration/configs/obi-config-parentbased-sampler.yml
+++ b/internal/test/integration/configs/obi-config-parentbased-sampler.yml
@@ -1,0 +1,26 @@
+routes:
+  patterns:
+    - /basic/:rnd
+  unmatched: path
+ebpf:
+  track_request_headers: true
+otel_metrics_export:
+  endpoint: http://otelcol:4018
+otel_traces_export:
+  endpoint: http://jaeger:4318
+  sampler:
+    name: parentbased_always_off
+discovery:
+  disabled_route_harvesters: ["nodejs"]
+  services:
+    - namespace: multi-k
+      name: service-a
+      open_ports: 5000
+attributes:
+  kubernetes:
+    cluster_name: my-kube
+  select:
+    http_server_request_duration_seconds_count:
+      exclude: ["server_address"]
+    "*":
+      include: ["*"]
--- a/internal/test/integration/docker-compose-parentbased-sampler.yml
+++ b/internal/test/integration/docker-compose-parentbased-sampler.yml
@@ -1,0 +1,103 @@
+version: "3.8"
+
+services:
+  nodeserver:
+    build:
+      context: ../../..
+      dockerfile: internal/test/integration/components/nodemultiproc/Dockerfile
+    image: nodemultiproc
+    ports:
+      - "5000:5000"
+      - "5001:5001"
+      - "5002:5002"
+      - "5003:5003"
+    environment:
+      LOG_LEVEL: DEBUG
+    depends_on:
+      otelcol:
+        condition: service_started
+      jaeger:
+        condition: service_started
+
+  obi:
+    build:
+      context: ../../..
+      dockerfile: ./internal/test/integration/components/obi/Dockerfile
+    command:
+      - --config=/configs/obi-config-parentbased-sampler.yml
+    volumes:
+      - ./configs/:/configs
+      - ./system/sys/kernel/security:/sys/kernel/security
+      - ../../../testoutput:/coverage
+      - ../../../testoutput/run-multi:/var/run/obi
+    image: hatest-obi
+    privileged: true # in some environments (not GH Pull Requests) you can set it to false and then cap_add: [ SYS_ADMIN ]
+    pid: "host"
+    environment:
+      GOCOVERDIR: "/coverage"
+      OTEL_EBPF_METRICS_FEATURES: "application,application_span_otel"
+      OTEL_EBPF_TRACE_PRINTER: "text"
+      OTEL_EBPF_METRICS_INTERVAL: "1s"
+      OTEL_EBPF_BPF_BATCH_TIMEOUT: "10ms"
+      OTEL_EBPF_LOG_LEVEL: "DEBUG"
+      OTEL_EBPF_BPF_DEBUG: "TRUE"
+      OTEL_EBPF_HOSTNAME: "obi"
+      OTEL_EBPF_OTLP_TRACES_BATCH_TIMEOUT: "1ms"
+      OTEL_EBPF_INTERNAL_METRICS_PROMETHEUS_PORT: 8999
+      OTEL_EBPF_INTERNAL_METRICS_PROMETHEUS_PATH: /metrics
+    ports:
+      - "8999:8999"
+
+  # OpenTelemetry Collector
+  otelcol:
+    image: otel/opentelemetry-collector-contrib:0.159.0@sha256:1f2c54a30e713fac6b3ae77a1ec84010c2007e29ced8ec666214fc2f6739c1cc
+    container_name: otel-col
+    deploy:
+      resources:
+        limits:
+          memory: 125M
+    restart: unless-stopped
+    command: ["--config=/etc/otelcol-config/otelcol-config-4017-weaver.yml"]
+    volumes:
+      - ./configs/:/etc/otelcol-config
+    ports:
+      - "4017" # OTLP over gRPC receiver
+      - "4018:4018" # OTLP over HTTP receiver
+      - "9464" # Prometheus exporter
+      - "8888" # metrics endpoint
+    depends_on:
+      prometheus:
+        condition: service_started
+      jaeger:
+        condition: service_started
+      weaver:
+        condition: service_healthy
+
+
+  weaver:
+    extends:
+      file: components/weaver/service.yml
+      service: weaver
+    volumes:
+      - ../../../schemas/obi:/obi-registry:ro
+  prometheus:
+    image: quay.io/prometheus/prometheus:v3.14.0@sha256:5ce7540c3c00ef4ab0c9d2c995c6a5b9c421f44b4a115d97a2c7af3b1c21cbb0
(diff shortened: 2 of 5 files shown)


marctc (changes requested):
This looks functionally equivalent to #416 and #444. I could not get that approach working end to end. The exporter only sees the completed span and the `TraceFlags` set by BPF. Changing the exporter alone cannot keep the sampling decision consistent across root and child spans. I think the decision must happen in BPF before the flags propagate. Can you add an end to end integration test that proves this works with real captured spans?

tete17:
Thanks for the pointer to #416/#444 — I read both before responding, and you're right about the general case. I've scoped the PR accordingly rather than defending the unconditional version.

**Where I agree:** for OBI-originated trace contexts, the flags byte is written at request time, before any sampling decision exists, so it is unconditionally `sampled` and carries no information. No exporter-side change can recover a decision that was never made. That covers client/producer spans (local parent) and OBI→OBI propagated hops — those genuinely need the decision to happen in eBPF before the flags propagate, i.e. #2922.

**Where I think an exporter fix still stands:** spans whose parent arrived on the wire from an actual sampler — a gateway or an SDK-instrumented upstream. There the decision exists and is encoded in the flags, and honoring it is standard W3C `ParentBased` semantics. That is precisely the #3190 scenario: Envoy sampling 100% at the edge, OBI configured `parentbased_always_off` to follow the edge decision — and exporting nothing at all, because the sampler is never shown the parent the span itself carries.

**What the PR now does (updated):** the remote parent is presented only for **server and consumer** spans — the kinds whose `ParentSpanID` comes from an incoming trace context. Client and producer spans keep root-half behavior, exactly as on `main`, with a comment explaining why and what completes them. So relative to `main` this is strictly additive: no span that lacks a wire parent changes behavior, and the root/child inconsistency you hit in #444 (local children detaching from their root's fate) cannot be introduced because local parents are never presented.

It also composes with #2922 rather than competing with it: once the eBPF decision rides the propagated flags, the wire-parent trust this PR implements becomes correct for OBI→OBI hops with no further exporter change — #2922 needs an exporter that honors parent flags anyway.

**On the integration test:** agreed, and working on it — adapting the e2e harness from #444 to the current test layout, targeting exactly the wire-parent claims: request with a sampled `traceparent` → span exported and joined; unsampled `traceparent` → dropped; no `traceparent` → dropped, all under `parentbased_always_off`. I'll push it to this branch.

(Also amended the commit author so EasyCLA can link it.)

tete17:
Integration test added as requested: `TestParentBasedSamplerWireParent`, real captured spans through the full pipeline (instrumented Node service → OBI → Jaeger), following the shape of the suite in #444.

Setup: global `parentbased_always_off` sampler, `ebpf.track_request_headers: true`, and three request patterns against the instrumented service — a fixed `traceparent` with flags `01`, a fixed `traceparent` with flags `00`, and no `traceparent`.

Asserted via the Jaeger API:

- only the trace carrying the **sampled** wire parent is exported;
- it keeps the incoming trace ID, and the exported span **references the exact parent span ID** from the `traceparent` header (not re-rooted);
- the unsampled trace and the root requests export nothing.

Results, run locally on kernel 6.12 with the same compose harness CI uses:

| build | result |
|---|---|
| `main` (verified fix absent from source before image build) | **FAIL** — no spans exported at all: `spans joining a sampled trace (traceparent flags 01) must be exported` |
| this branch | **PASS** (85s) |

Two honest notes from getting it working:

1. The test needs `ebpf.track_request_headers: true` — without it OBI never parses the incoming header, generates a fresh context per request, and everything is a root (which `parentbased_always_off` then drops). That's the read side and it's off by default; the fix only makes the sampler honor what the reader already extracted. This mirrors exactly how we hit the bug in production (Grafana Alloy/Beyla behind an Envoy gateway sampling 100%).
2. The failure mode on `main` in this e2e is precisely the one reported in #3190: OBI's own trace printer shows the spans carrying the right wire parent, while zero spans reach the backend.

The test is picked up automatically by `scripts/generate-integration-matrix.sh` (verified in the generated matrix output), so it runs in the standard integration shards.

marctc:
`TestMultiProcessAppCPTCPOnly` confirms my previous concern. It consistently loses a downstream server span. Server or consumer span does not prove that `TraceFlags` contains a real upstream sampling decision. The parent can come from BPF propagation, so constructing a remote parent in the exporter is not sufficient.

This needs more investigation before asking AI to propose another fix. Also, I am closing this PR because it does not meet our AI policy requirements for human review, understanding, and testing.

PR #3192, opened 2026-08-26
fix: route klog output through slog

skl:
## Summary

OBI uses `k8s.io/client-go` informers for Kubernetes metadata, but didn't previously configure `klog`. This resulted in all `client-go/klog` output going to stderr, while OBI's own `slog` output went to stdout. This inconsistency causes downstream log consumers of OBI difficultly in determining log levels.

This PR addresses the above by introducing a new `klogbridge` package, with the sole purpose of directing `klog` output to `slog`, resulting in consistent log output:

1. Stream — stdout (same as OBI’s slog), not raw stderr
2. Format — text/JSON from OBI’s config
3. Level filtering — OBI’s `log_level` applies

## Validation

### DIFF
--- a/pkg/kube/informer_provider.go
+++ b/pkg/kube/informer_provider.go
@@ -21,7 +21,6 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 
 	"go.opentelemetry.io/obi/pkg/export/imetrics"
-	"go.opentelemetry.io/obi/pkg/kube/klogbridge"
 	"go.opentelemetry.io/obi/pkg/kube/kubecache/meta"
 	"go.opentelemetry.io/obi/pkg/kube/kubeflags"
 )
@@ -143,7 +142,6 @@ func (mp *MetadataProvider) Get(ctx context.Context) (*Store, error) {
 }
 
 func (mp *MetadataProvider) getInformer(ctx context.Context) (meta.Notifier, error) {
-	klogbridge.Install()
 	if mp.informer != nil {
 		return mp.informer, nil
 	}
--- a/pkg/kube/informer_provider.go
+++ b/pkg/kube/informer_provider.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 
 	"go.opentelemetry.io/obi/pkg/export/imetrics"
+	"go.opentelemetry.io/obi/pkg/kube/klogbridge"
 	"go.opentelemetry.io/obi/pkg/kube/kubecache/meta"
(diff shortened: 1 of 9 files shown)


pkg/kube/informer_provider.go
```diff
@@ -142,6 +143,7 @@ func (mp *MetadataProvider) Get(ctx context.Context) (*Store, error) {
 }
 
 func (mp *MetadataProvider) getInformer(ctx context.Context) (meta.Notifier, error) {
+	klogbridge.Install()
```
MrAlias: These runtime calls can execute after the process has active klog users. `SetSlogLogger` is explicitly not thread-safe, and concurrent logging reproduces a data race. The collector also performs client-go work before reaching this lazy call. Install the bridge during each owning application's logger initialization—including the collector—and avoid mutating klog from public runtime APIs.

pkg/kube/klogbridge/klogbridge.go
```diff
@@ -0,0 +1,29 @@
...
+
+// Install routes all k8s.io/klog output (client-go informers, reflectors, etc.)
+// through the process-default slog logger, so it honors OBI's log level,
+// format, and output stream instead of klog's default raw-stderr format.
+//
+// Safe to call multiple times; only the first call has an effect. Call after
+// slog.SetDefault so the captured logger matches the embedding binary's config.
+func Install() {
+	klogBridgeOnce.Do(func() {
+		klog.SetSlogLogger(slog.Default().With("component", "k8s.client-go"))
```
MrAlias: `SetSlogLogger` does not make slog the only level gate. Legacy `klog.Warning*` records become `INFO` and are dropped when OBI uses `WARN`; legacy `klog.V(n)` records are pre-filtered by klog, so OBI `DEBUG` still misses them. client-go v0.35.4 uses both APIs. Account for these mappings and add WARN and DEBUG coverage.

MrAlias (changes requested):
The bridge is localized, but its process-global installation needs to happen during application initialization so the embedded Collector path cannot race with active logging. The klog level-mapping limitation should be handled or scoped explicitly.

pkg/kube/informer_provider.go
skl: Done in c010df6f3

pkg/kube/klogbridge/klogbridge.go
skl: Done in c010df6f3

grcevski (approved):
LGTM!

MrAlias (approved):

PR #3194, opened 2026-08-26
Otel compliance:  Add missing otel network attributes

NimrodAvni78:
## Summary

Emits three semconv network attributes OBI already had the data for but never wrote out.

| Attribute | Level (semconv 1.41.0) | Source |
| --- | --- | --- |
| `network.peer.address` | Recommended | `span.Peer` on server spans, `span.Host` on client spans |
| `network.peer.port` | Recommended — if `network.peer.address` is set | `span.PeerPort` / `span.HostPort`, same split |
| `network.protocol.version` | Recommended | parsed from the HTTP/1 request line; `2` for HTTP/2 and gRPC |

All three are registered in `Traces.Section`, so each can be turned off individually via `attributes.select.traces.exclude`. On by default, since semconv asks for them.

## network.peer.*

This recovers discarded information rather than filling a field: `server.address` and `client.address` prefer a resolved K8s/DNS name, so whenever resolution succeeded the socket address was lost — you could not tell which pod served a request.

Emitted only when the value parses as an IP. Some paths fall back to the `Host` header, which is a name, and semconv defines the attribute as an IP or Unix socket address; `server.address` already carries the name. DNS and failed-connect spans are excluded, the former because its branch inverts the client/server mapping.

The client/server split follows the existing `HostAsServer` / `PeerAsClient` conventions per event type.

## network.protocol.version

`Span.ProtoVersion` is a `uint8` enum next to the existing `Flags uint8`, reusing padding the struct already had. No BPF change: the HTTP/1 buffer path reads it from the request line (the same buffer `httpMethodFromBuf` already parses), the payload-extraction path takes it from `net/http`'s parsed request, and the HTTP/2 transform sets it directly, covering gRPC.

It is set only on those paths, never in the shared post-processor, so presence is deterministic per instrumentation rather than depending on whether a request buffer happened to be recoverable. Go uprobe spans come from a C struct with no version field, so they omit it.

Emitted only on plain HTTP spans: an HTTP subtype carries a different protocol, and the SQL++ branch in particular builds a deliberately DB-only attribute set.

`network.protocol.name` needs nothing — conditionally required only "If not `http`", and here it always is.

## Validation


grcevski (approved):
LGTM!

PR #3195, opened 2026-08-26
Otel compliance: Clamp http.request.method; emit method_original, user_agent.original, db.query.summary

NimrodAvni78:
## Summary

Clamps `http.request.method` to the semconv enum and emits three attributes OBI already had the data for but never wrote out.

| Attribute | Level (semconv 1.41.0) | Source |
| --- | --- | --- |
| `http.request.method_original` | Conditionally required — if and only if it differs from `http.request.method` | the wire method, when outside the semconv enum |
| `db.query.summary` | Recommended — if the instrumentation can generate a summary | `span.DBQuerySummary`, already computed by `sqlprune.SQLQuerySummary` |
| `user_agent.original` | Recommended on server spans, opt-in on client | `req.UserAgent()` off the parsed request, wherever payload extraction runs |

All three are registered in `Traces.Section`, so each can be turned off individually via `attributes.select.traces.exclude`, the same lever `url.query` uses.

`user_agent.original` and the other HTTP transport attributes are emitted only on plain HTTP spans: an HTTP subtype carries a different protocol.

## http.request.method clamping

A method outside the enum becomes `_OTHER`, with the wire value on `http.request.method_original` when it differs from `_OTHER` itself. Matched case-sensitively per spec, so `get` clamps. `OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS` overrides the enum, as semconv requires of any instrumentation that clamps — a comma-separated list replacing the defaults.

changes both the `http.request.method` span and metric attribute, and `span.name` (`"PURGE /x"` becomes `"HTTP /x"`, semconv's form)

We were not limiting the values of `http.request.method` previously, No test app previously sent a non-standard method, which is why weaver never caught it and failed.

```
unclamped (PURGE): undefined_enum_variant — value 'PURGE' is not documented
clamped   (_OTHER):   no violation
```

#### Known gap

- `method_original` can be truncated on the Go uprobe path: `k_method_max_len` is 7, so `PROPFIND` arrives as `PROPFIN`. Pre-existing, and no enum member is affected — the longest (`CONNECT`, `OPTIONS`) are exactly 7. Previously that truncated value landed on `http.request.method` itself, so this is strictly an improvement.

## Validation


pkg/appolly/app/request/http_method.go
```diff
@@ -0,0 +1,79 @@
...
+var HTTPMethodOther = semconv.HTTPRequestMethodOther.Value.AsString()
+
+// Semconv mandates this variable name for overriding the enum: a comma-separated
+// list replacing the defaults below.
+const envKnownMethods = "OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS"
+
+// Taken from the semconv enum members rather than copied as literals, so a
+// changed value fails to compile instead of silently clamping a valid method.
+// Method names are case-sensitive, so the set is matched exactly.
+var defaultKnownHTTPMethods = methodSet(
```
grcevski, line 26: I think our eBPF generic tracer will not detect any of the non-standard ones. We check for GET,POST,PUT,DELETE,PATCH,HEAD I think. It will work for Go.

grcevski (commented):
I think this is good idea, I think though we need another detector for HTTP on the TCP path. Since we'll miss from eBPF the HTTP methods we don't know of and then we'll need to detect them again.

pkg/appolly/app/request/http_method.go
NimrodAvni78: yeah you'r right here, currently it only works for the go path (and JVM SSL as well i think)
I think it still makes sense to use the HTTP methods for detection in kernel space to narrow down our filters, which means we will never forward non standard ones

But maybe as a follow up we can load the list of known HTTP methods from the config into kernel space and use it for detection (with making sure performance stays the same)
WDYT?

pkg/ebpf/common/http/http_enrichment.go
```diff
@@ -128,6 +130,11 @@ func (e *HTTPEnricher) processHeaders(
 	for name, values := range headers {
 		action, rule := e.resolveHeaderAction(name, scope, span)
 		if action == config.HTTPParsingActionExclude {
+			// An excluded header must not resurface through a semconv attribute
```
mmat11, line 133: this will still appear in http-based protocols which don't enrich the span (eg. graphql, aws, etc.) -- is this the desired behavior?

NimrodAvni78: @mmat11 i think this kinda relates to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3217
so which one merges sooner we can rebase to it

pkg/appolly/app/request/http_method.go
grcevski: Yeah it might be good idea, right now the code is very much optimised to detect only certain types in the generic tracer path.

grcevski (approved):
LGTM!

PR #3196, opened 2026-08-26
Otel compliance: improve and unify error.type across spans and metrics

NimrodAvni78:
## Summary

Emits `error.type` on spans, and makes the metric label report the same value instead of a literal `"error"`.

`error.type` is Conditionally Required by semconv when a request fails, and was simply absent from spans. It is registered in `Traces.Section`, so it can be turned off via `attributes.select.traces.exclude`.

## Values

Set only when `SpanStatusCode` already reports failure, always from a closed set:

| Span | Value |
| --- | --- |
| HTTP | status code (`503`) |
| gRPC | status name (`UNAVAILABLE`) |
| DB | server error code (`WRONGTYPE`) |
| DNS | RCODE (`NXDomain`) |
| GenAI | provider error (`rate_limit_error`) |
| anything else that failed | `_OTHER` |

The HTTP status is used only for plain HTTP spans. An HTTP subtype is a different protocol carried over HTTP, so it reports its own parsed error, or nothing when it has none.

Excluded: messaging spans (`SpanStatusCode` never marks them failed), failed-connect spans (`status` is never assigned, so `_OTHER` would be a constant label), and ONC/Sun RPC, whose status is meaningful (`denied`, or `AcceptStat + 1`) — mapping it properly is a follow-up.

## Behavior change: the metric label

The metrics pipeline had its own getter returning the literal `"error"` for everything except DNS and memcached. we now use the same mapping for both.

Label values change on the four metric families that carry `error.type`: `db.client.operation.duration`, `db.server.operation.duration`, `dns.lookup.duration` and `gen_ai.client.operation.duration`. In practice that means DB metrics for Redis, Mongo, Couchbase, Aerospike and SQL go from `"error"` to the real code; DNS and memcached already reported real values and are unaffected. HTTP and RPC metrics do not carry `error.type` at all, so they are untouched.

## Validation


pkg/appolly/app/request/span_error_type.go
```diff
@@ -0,0 +1,70 @@
...
+	return ErrorTypeOther
+}
+
+// parsedErrorType reports the error a protocol parser extracted from the
+// payload, across every field the parsers write it to.
+func parsedErrorType(span *Span) string {
+	if span.DBError.ErrorCode != "" {
+		return span.DBError.ErrorCode
+	}
+	if span.SQLError != nil && span.SQLError.SQLState != "" {
```
grcevski: I think we should fall back on the numeric code when sqlstate is missing.

The spec says:
"These systems SHOULD set the db.response.status_code to a known vendor-specific error code. If only SQLSTATE is available, it SHOULD be used."

Also from mysql:

"
Additionally, error.type SHOULD match db.response.status_code. MySQL’s ERR_Packet specification confirms error_code is unconditional, while SQLSTATE is conditional on CLIENT_PROTOCOL_41.
"

pkg/appolly/app/request/span_error_type.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+	case EventTypeGRPC, EventTypeGRPCClient:
+		if code := GRPCStatusCodeString(span.Status); code != "" {
+			return code
+		}
+	case EventTypeDNS:
+		return dnsparser.RCode(span.Status).String()
+	}
+
+	return ErrorTypeOther
```
grcevski, line 40: I think we need to consider others here, SunRPC at least?

NimrodAvni78: Good catch, Done! I looked at other places and i dont think i missed that we actually extract error info about

pkg/appolly/app/request/span_error_type.go
NimrodAvni78: Done

grcevski (approved):
LGTM!

PR #3199, opened 2026-08-26
gotracer: support nested spans in shared ctx

### LINKED ISSUE #2778: bpf: SQL query completion clobbers enclosing HTTP server span context

mmat11:
## Summary

Fixes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2778

Previous PRs https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2775 https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2875 https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2876

/cc @mtanda 

## Validation


MrAlias (changes requested):
The implementation does not yet preserve the active-context invariant across span begin/end, scheduling, overflow, and generic restoration: same-kind spans need distinct identities, stackless header state must survive rescheduling, overflow must retain a recoverable parent, `g_dead` must clear goroutine-owned state, and server restoration must follow live request ownership rather than timestamp buckets. Until every path restores the enclosing context without reviving stale state, this is not safe to merge.

Please also remove and ignore `internal/test/integration/components/go_simple_grpc/go_simple_grpc`: it is an unused 17 MB Mach-O binary, while the component image builds `/src/grpclog` from `main.go` and copies only that output.

bpf/gotracer/go_kafka_go.c
```diff
@@ -69,8 +69,9 @@ int GUARDED_PROG(obi_uprobe_writer_write_messages_ret, struct pt_regs *, ctx) {
 
     // Drop the goroutine-keyed traceparent so casgstatus can't re-install this
     // produce's context after the request ends (issue #2046).
+    const tp_info_t *tp = bpf_map_lookup_elem(&produce_traceparents_by_goroutine, &g_key);
+    go_obi_ctx__end(&g_key, k_obi_ctx_kafka_produce, tp);
```
MrAlias, line 73: `TestSuite_GoKafkaTraceparent` passes on the base commit, but this head fails the initial run and both reruns: 61 of 63 post-produce worker logs had a trace ID initially, and all 61 did on each rerun. Since these logs run after `WriteMessages` returns with a background context, the goroutine is keeping or reinstalling stale trace context for unrelated work. Could we fix the lifecycle path that leaves or restores this context after the produce returns and keep this regression green?

bpf/gotracer/go_obi_ctx.h
```diff
@@ -0,0 +1,182 @@
...
+        bpf_memset(fresh, 0, sizeof(*fresh));
+        fresh->frames[0].tp = *tp;
+        fresh->frames[0].kind = kind;
+        fresh->depth = 1;
+        bpf_map_update_elem(&obi_ctx_stacks, g_key, fresh, BPF_ANY);
+        return;
+    }
+
+    if (st->depth >= k_obi_ctx_max_depth) {
+        // full: drop above the outermost, which must stay recoverable
```
MrAlias: Retaining the outermost frame here does not keep it recoverable once an evicted invocation returns. With five nested same-kind SQL calls, the oldest inner frame is discarded; after the retained inner calls unwind, its return falls back to the newest SQL frame and removes the still-running outer context, leaving subsequent logs unenriched. Could we track outstanding evictions so an end can never consume a different invocation's frame, and extend the same-kind test past `k_obi_ctx_max_depth`?

MrAlias (commented):
The context stack still needs strict begin/end ownership: overflow can let an evicted return remove the retained outer context, and the Kafka return path now fails the existing stale-context regression on every run.

bpf/gotracer/go_kafka_go.c
mmat11: yeah I noticed this locally, still investigating

grcevski:
There seems to be verifier issue with 5.15:
```
obi-1                   |  R0=map_value(id=0,off=0,ks=16,vs=240,imm=0) R1=map_value(id=0,off=0,ks=16,vs=240,imm=0) R2=inv(id=21,umin_value=4,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R3_w=map_value(id=0,off=0,ks=16,vs=240,umax_value=168,var_off=(0x0; 0xf8),s32_max_value=248,u32_max_value=248) R4=inv(id=34,umax_value=3,var_off=(0x0; 0x3)) R5=inv3 R6=fp-112 R7=inv(id=0) R8=inv(id=0) R9=map_value(id=0,off=0,ks=16,vs=240,imm=0) R10=fp0 fp-16=mmmmmmmm fp-24=mmmmmmmm fp-32=mmmmmmmm fp-40=mmmmmmmm fp-48=mmmmmmmm fp-56=mmmmmmmm fp-64=mmmmmmmm fp-72=mmmmmmmm fp-80=mmmmmmmm fp-88=mmmmmmmm fp-96=mmmmmmmm fp-104=mmmmmmmm fp-112=mmmmmmmm fp-120=mmmmmmmm fp-128=mmmmmmmm fp-136=mmmmmmmm fp-144=mmmmmmmm fp-152=mmmmmmmm fp-160=mmmmmmmm fp-168=mmmmmmmm fp-176=mmmmmmmm fp-184=mmmmmmmm fp-192=mmmmmmmm fp-200=mmmmmmmm fp-208=mmmmmmmm
obi-1                   | ; return (u32)(stack_hi - PT_REGS_SP(ctx));
obi-1                   | 1482: (bf) r5 = r8
obi-1                   | 1483: (67) r5 <<= 32
obi-1                   | 1484: (77) r5 >>= 32
obi-1                   | ; if (top->kind == kind && top->stack_off == stack_off) {
obi-1                   | 1485: (5d) if r3 != r5 goto pc+21
obi-1                   | 
obi-1                   | from 1485 to 1507: R0=map_value(id=0,off=0,ks=16,vs=240,imm=0) R1=map_value(id=0,off=0,ks=16,vs=240,imm=0) R2=inv(id=21,umin_value=4,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R3_w=inv(id=0,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R4=inv(id=34,umax_value=3,var_off=(0x0; 0x3)) R5_w=inv(id=0,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R6=fp-112 R7=inv(id=0) R8=inv(id=35) R9=map_value(id=0,off=0,ks=16,vs=240,imm=0) R10=fp0 fp-16=mmmmmmmm fp-24=mmmmmmmm fp-32=mmmmmmmm fp-40=mmmmmmmm fp-48=mmmmmmmm fp-56=mmmmmmmm fp-64=mmmmmmmm fp-72=mmmmmmmm fp-80=mmmmmmmm fp-88=mmmmmmmm fp-96=mmmmmmmm fp-104=mmmmmmmm fp-112=mmmmmmmm fp-120=mmmmmmmm fp-128=mmmmmmmm fp-136=mmmmmmmm fp-144=mmmmmmmm fp-152=mmmmmmmm fp-160=mmmmmmmm fp-168=mmmmmmmm fp-176=mmmmmmmm fp-184=mmmmmmmm fp-192=mmmmmmmm fp-200=mmmmmmmm fp-208=mmmmmmmm
obi-1                   | ; return;
obi-1                   | 1507: (b7) r3 = 4
obi-1                   | ; if (st->depth >= k_obi_ctx_max_depth) {
obi-1                   | 1508: (2d) if r3 > r2 goto pc-61
obi-1                   | 
obi-1                   | from 1508 to 1448: R0=map_value(id=0,off=0,ks=16,vs=240,imm=0) R1=map_value(id=0,off=0,ks=16,vs=240,imm=0) R2=inv(id=21,umin_value=4,umax_value=3,var_off=(0x0; 0x7)) R3_w=inv4 R4=inv(id=34,umax_value=3,var_off=(0x0; 0x3)) R5_w=inv(id=0,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R6=fp-112 R7=inv(id=0) R8=inv(id=35) R9=map_value(id=0,off=0,ks=16,vs=240,imm=0) R10=fp0 fp-16=mmmmmmmm fp-24=mmmmmmmm fp-32=mmmmmmmm fp-40=mmmmmmmm fp-48=mmmmmmmm fp-56=mmmmmmmm fp-64=mmmmmmmm fp-72=mmmmmmmm fp-80=mmmmmmmm fp-88=mmmmmmmm fp-96=mmmmmmmm fp-104=mmmmmmmm fp-112=mmmmmmmm fp-120=mmmmmmmm fp-128=mmmmmmmm fp-136=mmmmmmmm fp-144=mmmmmmmm fp-152=mmmmmmmm fp-160=mmmmmmmm fp-168=mmmmmmmm fp-176=mmmmmmmm fp-184=mmmmmmmm fp-192=mmmmmmmm fp-200=mmmmmmmm fp-208=mmmmmmmm
obi-1                   | ; obi_ctx_frame_t *frame = &st->frames[st->depth & (k_obi_ctx_max_depth - 1)];
obi-1                   | 1448: (27) r2 *= 56
obi-1                   | 1449: (0f) r0 += r2
obi-1                   | last_idx 1449 first_idx 1480
obi-1                   | regs=4 stack=0 before 1448: (27) r2 *= 56
obi-1                   | regs=4 stack=0 before 1508: (2d) if r3 > r2 goto pc-61
obi-1                   | regs=c stack=0 before 1507: (b7) r3 = 4
obi-1                   | regs=4 stack=0 before 1485: (5d) if r3 != r5 goto pc+21
obi-1                   | regs=4 stack=0 before 1484: (77) r5 >>= 32
obi-1                   | regs=4 stack=0 before 1483: (67) r5 <<= 32
obi-1                   | regs=4 stack=0 before 1482: (bf) r5 = r8
obi-1                   | regs=4 stack=0 before 1481: (61) r3 = *(u32 *)(r3 +48)
obi-1                   | regs=4 stack=0 before 1480: (55) if r5 != 0x3 goto pc+26
obi-1                   |  R0_r=map_value(id=0,off=0,ks=16,vs=240,imm=0) R1_r=map_value(id=0,off=0,ks=16,vs=240,imm=0) R2_r=invP(id=21,umin_value=4,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R3_rw=map_value(id=0,off=0,ks=16,vs=240,umax_value=168,var_off=(0x0; 0xf8),s32_max_value=248,u32_max_value=248) R4_rw=invP(id=34,umax_value=3,var_off=(0x0; 0x3)) R5_rw=inv(id=0,umax_value=255,var_off=(0x0; 0xff)) R6_r=fp-112 R7=inv(id=0) R8_rw=inv(id=0) R9=map_value(id=0,off=0,ks=16,vs=240,imm=0) R10=fp0 fp-16=mmmmmmmm fp-24=mmmmmmmm fp-32=mmmmmmmm fp-40=mmmmmmmm fp-48=mmmmmmmm fp-56=mmmmmmmm fp-64=mmmmmmmm fp-72_r=mmmmmmmm fp-80_r=mmmmmmmm fp-88_r=mmmmmmmm fp-96_r=mmmmmmmm fp-104_r=mmmmmmmm fp-112_r=mmmmmmmm fp-120=mmmmmmmm fp-128=mmmmmmmm fp-136=mmmmmmmm fp-144=mmmmmmmm fp-152=mmmmmmmm fp-160=mmmmmmmm fp-168=mmmmmmmm fp-176=mmmmmmmm fp-184=mmmmmmmm fp-192=mmmmmmmm fp-200=mmmmmmmm fp-208=mmmmmmmm
obi-1                   | parent didn't have regs=4 stack=0 marks
obi-1                   | last_idx 1479 first_idx 1445
obi-1                   | regs=4 stack=0 before 1479: (71) r5 = *(u8 *)(r3 +52)
obi-1                   | regs=4 stack=0 before 1478: (0f) r3 += r5
obi-1                   | regs=4 stack=0 before 1477: (bf) r3 = r0
obi-1                   | regs=4 stack=0 before 1476: (27) r5 *= 56
obi-1                   | regs=4 stack=0 before 1475: (bf) r5 = r4
obi-1                   | regs=4 stack=0 before 1474: (57) r4 &= 3
obi-1                   | regs=4 stack=0 before 1473: (07) r4 += -1
obi-1                   | regs=4 stack=0 before 1472: (bf) r4 = r2
obi-1                   | regs=4 stack=0 before 1447: (05) goto pc+24
obi-1                   | regs=4 stack=0 before 1446: (15) if r2 == 0x0 goto pc+1
obi-1                   | regs=4 stack=0 before 1445: (79) r8 = *(u64 *)(r10 -176)
obi-1                   |  R0_r=map_value(id=0,off=0,ks=16,vs=240,imm=0) R1_r=map_value(id=0,off=0,ks=16,vs=240,imm=0) R2_r=invP(id=21,umin_value=4,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R3=inv0 R4_w=inv3 R5=inv3 R6_r=fp-112 R7=inv(id=0) R8_w=inv(id=0) R9_w=map_value(id=0,off=0,ks=16,vs=240,imm=0) R10=fp0 fp-16=mmmmmmmm fp-24=mmmmmmmm fp-32=mmmmmmmm fp-40=mmmmmmmm fp-48=mmmmmmmm fp-56=mmmmmmmm fp-64=mmmmmmmm fp-72_r=mmmmmmmm fp-80_r=mmmmmmmm fp-88_r=mmmmmmmm fp-96_r=mmmmmmmm fp-104_r=mmmmmmmm fp-112_r=mmmmmmmm fp-120=mmmmmmmm fp-128=mmmmmmmm fp-136=mmmmmmmm fp-144=mmmmmmmm fp-152=mmmmmmmm fp-160=mmmmmmmm fp-168=mmmmmmmm fp-176_r=mmmmmmmm fp-184=mmmmmmmm fp-192=mmmmmmmm fp-200=mmmmmmmm fp-208=mmmmmmmm
obi-1                   | parent already had regs=4 stack=0 marks
obi-1                   | ; frame->tp = *tp;
obi-1                   | 1450: (79) r2 = *(u64 *)(r6 +40)
obi-1                   | 1451: (7b) *(u64 *)(r0 +40) = r2
obi-1                   | R0 invalid mem access 'inv'
obi-1                   | processed 534 insns (limit 1000000) max_states_per_insn 1 total_states 39 peak_states 39 mark_read 15
```

mmat11:
@grcevski yep, I have a fix locally, I'm doublechecking it works properly

skl (approved):
lgtm, suggest updating devdocs [here](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/f1fec54701445568b2b4bc01920324149dea6df9/devdocs/trace-log-correlation.md?plain=1#L91-L101) based on the changes

bpf/gotracer/go_obi_ctx.h
```diff
@@ -0,0 +1,237 @@
...
+        return;
+    }
+
+    const u32 idx = obi_ctx__find(st, obi_ctx__depth(st), kind, tp);
+
+    if (idx < k_obi_ctx_max_depth && (tp || st->overflow == 0)) {
+        st->depth = idx;
+        st->overflow = 0;
+    } else if (st->overflow > 0) {
+        st->overflow--;
```
MrAlias, line 219: When the stack is overflowed, every unmatched end decrements `overflow`, regardless of `kind`. `ClientConn.Close` always reaches `go_obi_ctx__end(..., k_obi_ctx_grpc_client, NULL)`, so closing a connection inside more than four nested spans consumes another span's overflow entry; after the actual returns, the next NULL end pops a retained frame early and logs under the wrong outer span. Could we keep enough identity for overflowed frames to reject unrelated ends and add this sequence to `bpf_go_obi_ctx.c`?

MrAlias (commented):
The main lifecycle fixes are in place, but overflow accounting can still restore the wrong enclosing span when an unrelated end probe fires.

bpf/gotracer/go_obi_ctx.h
mmat11: fixed

bpf/gotracer/go_grpc.c
```diff
@@ -551,8 +557,8 @@ int GUARDED_PROG(obi_uprobe_ClientConn_Close, struct pt_regs *, ctx) {
     go_addr_key_t g_key = {};
     go_addr_key_from_id(&g_key, goroutine_addr);
 
+    go_obi_ctx__end(&g_key, k_obi_ctx_grpc_client, NULL);
```
MrAlias: `ClientConn.Close` is not tied to the invocation it removes. The probe ignores `GO_PARAM1(ctx)` and unconditionally performs a kind-only end plus deletion of the goroutine-keyed invocation. If an interceptor closes connection B while an RPC on connection A is active, the stored branch pops A's frame; the overflow branch consumes A's gRPC overflow count. The SQL-overflow test cannot exercise either same-kind collision.

Could we make a receiver mismatch return without touching either `obi_ctx_stacks` or `ongoing_grpc_client_requests`, and replace the single `BPF_ANY` goroutine slot with state that preserves each active invocation?

For example, a per-goroutine LIFO can cover nested Invoke calls, while RecvMsg may need correlation by `*clientStream`. Then each Invoke/RecvMsg completion can remove its exact record and call `go_obi_ctx__end` with that record's `tp`. `ClientConn.Close` cannot distinguish two active RPCs on the same connection, so it should not issue a kind-only end; if it remains a cleanup path, it needs that same invocation identity. Once end paths own an invocation, missing or foreign lookups can safely be no-ops.

Can we also cover the stored and overflowed gRPC frames with test(s) and verify that the later RPC completion removes its own frame?

grcevski: I don't think I understand this comment @MrAlias. If an interceptor closes the gRPC connection, isn't it game over for the client tracking? What do you mean by "stored branch pop A's frame"? Do you mind explaining in a bit more detail?

grcevski (approved):
LGTM!

bpf/gotracer/go_grpc.c
MrAlias: Sure, I'll give it a try.

Closing connection A is terminal for its RPCs, but it should not be terminal for their telemetry bookkeeping: those calls still unwind through the `Invoke`/`RecvMsg` probes, where we need A's invocation record to emit the failed span and end by A's `tp`. The collision I meant involves two `ClientConn`s on the same goroutine:

1. `A.Invoke(...)` enters on goroutine G. We store A's invocation in `ongoing_grpc_client_requests[G]` and begin A's gRPC context.
2. A unary interceptor for A, still on G, calls `B.Close()` on a different connection.
3. The current `Close` probe does not check receiver B. It uses only G, so it deletes A's invocation and calls `go_obi_ctx__end(G, grpc_client, NULL)`, which likewise identifies only the kind.
4. `B.Close()` returns and the interceptor continues A. B's close does not terminate A's RPC, but A's tracking record is already gone. When A completes, its completion probe cannot emit A's span or end by A's `tp`.
 
By “stored branch,” I meant the first branch in `go_obi_ctx__end`. If A fit in `frames[]`, `obi_ctx__find` finds A by kind and `st->depth = idx` removes A's frame and anything above it. If the stack was already full when A began, A has no stored frame; `go_obi_ctx__begin` increments `overflow[k_obi_ctx_grpc_client]` instead, and `Close(B)` takes the second branch and decrements A's count. In both cases, closing B consumes state belonging to A.

So if `Close` is on A, A's completion still needs its telemetry record; if `Close` is on B, it must not touch A's state at all.

grcevski: I think this scenario is more general problem with the way we correlate things. I agree that it's a problem, but I'm not sure it is in the context of the uprobes we are setting. These probes are specific to a given grpc connection and the way the go grpc code is written today, there's 1:1 mapping between a goroutine and connection. Tracking which connection it was requires that we change all keys to be a pair of goroutine/connection. It's not a bad idea, but perhaps our of scope for this PR.

MrAlias: Yeah, I agree that the broader nested gRPC correlation problem should be handled separately. 👍 

For this PR, my request is narrower: `ClientConn_Close` must not remove state for a different connection merely because the close occurs on the same goroutine.

The 1:1 assumption does not hold. The uprobe is attached to the `ClientConn.Close` method, not to a particular instance, and receives the actual receiver through `GO_PARAM1(ctx)`. Also, [`ClientConn` may be used concurrently](https://github.com/grpc/grpc-go/blob/v1.83.1/Documentation/concurrency.md#clients), and [`ClientConn.Invoke` calls application interceptor code synchronously](https://github.com/grpc/grpc-go/blob/v1.83.1/call.go#L29-L38). A valid call path is therefore `A.Invoke()` → interceptor → `B.Close()` on the same G. The current probe sees B but ignores that receiver and deletes and kind-ends A's G-keyed state.

Could we compare `GO_PARAM1(ctx)` with the stored `invocation->cc` and return without modifying `ongoing_grpc_client_requests` or the context stack when there is no invocation or the receiver does not match? If so, can we also add a regression test for the same-goroutine A/B scenario, covering stored and overflowed gRPC context and verifying that A's client span is still exported and the surrounding context is restored?

The existing test closes the connection only after `Invoke` has completed and does so on another goroutine, so it does not exercise this collision. The broader single-slot limitation can be tracked separately; a `(goroutine, connection)` key would not fully solve recursive calls on the same connection anyway.

mmat11: done, added tests

MrAlias (approved):

mtanda:
Thank you for taking the time to fix this, @mmat11. Really appreciate it!

Issue #2905, opened 2026-08-04
Expand JVM runtime metrics


marctc:
### Component

OBI (eBPF instrumentation)

### Problem

Currently we have a basic subset of JVM metrics, but we could use jvm agent with jmx to get the rest of metrics: https://opentelemetry.io/docs/specs/semconv/runtime/jvm-metrics/

PR #3203, opened 2026-08-26
Add JVM class, thread, and CPU runtime metrics 

marctc:
## Summary

This PR adds Java agent collection and export for the following JVM runtime metrics:

  - `jvm.class.loaded`
  - `jvm.class.unloaded`
  - `jvm.class.count`
  - `jvm.thread.count`
  - `jvm.cpu.time`
  - `jvm.cpu.count`
  - `jvm.cpu.recent_utilization`

The values come from the JDK platform MXBeans and are sent through the existing Java agent `ioctl` side channel and runtime metrics pipeline.

Like Node.js runtime metrics, these metrics require the injected agent. `javaagent.enabled` is enabled by default. Setting it to `false` disables the class, thread, and CPU metrics. The existing HotSpot memory metrics continue  working through their USDT probes.

The agent reuses the existing `jvm_runtime_metrics.sampling_interval` configuration and its 1 second default. The same interval controls both HotSpot memory sampling and Java agent collection.

  ### Benchmark

A JDK 21 JMH reference run used one pinned CPU, five JVM forks, three warmup iterations, and seven measured iterations per fork. 

  | Sampling interval | Approximate one-core usage | Allocation rate |
  | --- | ---: | ---: |
  | 10 ms | 0.603% | 6.38 MB/s |
  | 100 ms | 0.060% | 638 KB/s |
  | 1 s (default) | 0.006% | 63.8 KB/s |
  | 5 s | 0.0012% | 12.8 KB/s |
  | 10 s | 0.0006% | 6.4 KB/s |

The benchmark measures management-bean collection and packet encoding. It does not measure initial agent attachment, instrumentation, or the JNI ioctl call.

A follow-up PR will add `jvm.gc.duration`, the remaining metric from the current JVM runtime semantic conventions.

Part of #2905

## Validation


pkg/runtimemetrics/reader.go
```diff
@@ -513,3 +526,15 @@ func SnapshotFromJVMRuntimeEvent(event appruntime.JVMRuntimeEvent) RuntimeMetric
...
+func SnapshotFromJVMRuntimeEvent(event appruntime.JVMRuntimeEvent) RuntimeMetricSnapshot {
+	values := event.Values
+	return RuntimeMetricSnapshot{
+		Service: event.Service,
+		PID:     event.PID,
+		Time:    event.Time,
+		JVM: &JVMRuntimeMetricSnapshot{
+			RuntimeValues: &values,
+		},
+	}
```
MrAlias, line 661: `JVMRuntimeEvent` loses the process lifetime here. A snapshot queued before termination can reach Prometheus after final-service cleanup; JVM snapshots bypass `runtimeSnapshotProcessLive`, so the sample recreates the three non-expiring counters. After same-service PID reuse, OTLP also accepts the old snapshot because its generation is zero. Could we carry the runtime-metric generation into this snapshot and gate Prometheus on `Service.ProcPID` under `runtimeMu`? `snapshot.PID` is namespace-scoped and is not the process-tracker key.

pkg/export/otel/metrics_jvm.go
```diff
@@ -54,11 +78,84 @@ func setupJVMRuntimeMeters(ctx context.Context, m *jvmRuntimeMetrics, meter inst
...
+		if now.Sub(m.lastExpiration) >= m.ttl {
+			m.runtimeEntries.DeleteExpired()
+			m.lastExpiration = now
+		}
+
+		entryKey := strconv.Itoa(int(snapshot.Service.ProcPID))
+		entry := m.runtimeEntries.GetOrCreate(
+			[]string{entryKey},
+			func() *jvmRuntimeEntry { return &jvmRuntimeEntry{} },
+		)
```
MrAlias: This expires a live JVM's cumulative baseline using the metric TTL. If `sampling_interval` exceeds that TTL while another event keeps the service reporter alive, the next snapshot creates an empty entry and adds the JVM's full class and CPU lifetime again. Could we retain the baseline for the process generation until termination and cover this configuration in a test?

pkg/export/prom/prom_runtime.go
```diff
@@ -332,29 +394,50 @@ func (r *metricsReporter) deleteRuntimeMetrics(service *svc.Attrs) {
...
-	r.goRuntimeMetrics.memoryUsed.DeleteLabelValues(append(append([]string{}, labels...), "other")...)
-	r.goRuntimeMetrics.deleteCounter(
-		r.goRuntimeMetrics.memoryAllocated,
-		attributes.GoRuntimeMemoryAllocated.Prom,
-		labels,
+	// Keep the shared JVM counters while any PID still belongs to this
+	// service. Delete them only when the service loses its final PID.
+	if r.jvmRuntimeMetrics.classLoaded == nil ||
+		r.pidsTracker.ServiceLive(service.UID) {
+		return
```
MrAlias, line 447: This preserves every source baseline until the service loses its final PID. When one JVM exits while another keeps the UID live, the departed PID's `metric+labels+ProcPID` entry survives; if that PID is reused and the replacement's first value exceeds the stale baseline, `delta` subtracts the previous JVM's count. Could we retire the exiting PID's source entries on every process deletion while keeping the shared `CounterVec` until the final PID exits?

MrAlias (commented):
The new JVM counters need to use the process generation as their lifecycle boundary; the current queue and cleanup behavior can replay or undercount cumulative values after process exit or PID reuse.

pkg/internal/java/agent/src/main/java/io/opentelemetry/obi/java/JVMRuntimeMetrics.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Consumer;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+final class JVMRuntimeMetrics {
+  static final long DEFAULT_SAMPLING_INTERVAL_NANOS = TimeUnit.SECONDS.toNanos(10);
+  private static final Logger logger = Logger.getLogger("JVMRuntimeMetrics");
```
grcevski: We had issues with the logger class in some places when dynamically loading. Use System.err.println instead.

pkg/internal/java/java_inject.go
```diff
@@ -51,7 +51,12 @@ type JavaInjector struct {
 }
 
 func NewJavaInjector(cfg *obi.Config) (*JavaInjector, error) {
+	log := slog.With("component", "javaagent.Injector")
 	if !cfg.Java.Enabled {
+		if cfg.AppRuntimeMetricsEnabled() {
+			log.Warn("application_runtime is enabled but the Java agent is disabled " +
+				"(javaagent.enabled=false): Java agent-backed JVM runtime metrics will not be collected")
```
grcevski: We should probably say which metrics will not be collected, e.g. class loading and cpu

grcevski (approved):
LGTM!

bpf/generictracer/java_tls.c
```diff
@@ -163,6 +167,39 @@ int BPF_KPROBE_GUARDED(obi_kprobe_sys_ioctl) {
...
+        }
+
+        bpf_memset(event, 0, sizeof(*event));
+        if (bpf_probe_read_user(
+                &event->loaded_class_count, k_jvm_runtime_metrics_payload_len, uarg + 1) != 0) {
+            bpf_ringbuf_discard(event, 0);
+            return 0;
+        }
+
+        // Event identity
```
pinoOgni: nit: remove it

pkg/export/otel/metrics_jvm.go
```diff
@@ -75,6 +168,46 @@ func (m *jvmRuntimeMetrics) record(snapshot runtimemetrics.RuntimeMetricSnapshot
...
+	}
+	if delta > 0 {
+		metric.Add(ctx, int64(delta))
+	}
+
+	value := current
+	*previous = &value
+}
+
+func recordJVMRuntimeFloatCounter(
```
pinoOgni, line 193: Is this function not the same as`recordJVMRuntimeCounter`?

pkg/export/prom/prom_jvm_test.go
```diff
@@ -24,7 +25,7 @@ import (
 	"go.opentelemetry.io/obi/pkg/runtimemetrics"
 )
 
-func TestRuntimeMetricsReporterRecordsJVMMemoryPoolUsed(t *testing.T) {
+func TestRuntimeMetricsReporterRecordsJVMMetrics(t *testing.T) {
```
pinoOgni, line 28: is this test too big? Maybe we can split it into subtests?

devdocs/config/CONFIG.md
```diff
@@ -407,13 +407,13 @@ AvoidedServicesConfig controls the avoided-services internal metric.
 | `javaagent.attach_timeout` | `duration` | `OTEL_EBPF_JAVAAGENT_ATTACH_TIMEOUT` | `10s` | `30s`, `5m`, `1ms`, etc |  |  |
 | `javaagent.debug` | `boolean` | `OTEL_EBPF_JAVAAGENT_DEBUG` | `false` |  |  |  |
 | `javaagent.debug_instrumentation` | `boolean` | `OTEL_EBPF_JAVAAGENT_DEBUG_INSTRUMENTATION` | `false` |  |  |  |
-| `javaagent.enabled` | `boolean` | `OTEL_EBPF_JAVAAGENT_ENABLED` | `true` |  |  |  |
+| `javaagent.enabled` | `boolean` | `OTEL_EBPF_JAVAAGENT_ENABLED` | `true` |  |  | Turns on the Java injector agent, used for TLS tracing, virtual thread correlation, and agent-backed runtime metrics. Setting it to false disables class, thread, and CPU runtime metrics. HotSpot memory metrics remain available. |
 
 ## `jvm_runtime_metrics`
```
pinoOgni, line 412: I think that `SUPPORT_MATRIX.md` needs to be updated.

pinoOgni (commented):
Nice stuff!  I have a question.

AFAIU the counters (`JVMClassLoaded` `JVMClassUnloaded` and `JVMCPUTime`) are built to aggregate per `jvmRuntimeEntryKey` i.e. `pid` and `generation` while the gauges `JVMClassCount`, `JVMCPUCount` and `JVMCPURecentUtilization` are not like that and they are created with `nil` attributes and `JVMThreadCount` only with the daemon `bool`. It means that every PID under the same UID writes the same series (which it should be fine because `svc.UID.Instance` is `namespace.pod.container` in kubernetes and it is the container name in docker and to have a collision we should have two jvm processes in one container...). 

The question is: shouldn't the creation of counters and gauges be the same?

pkg/export/otel/metrics_jvm.go
marctc: yes but other one receives a `instrument.Int64Counter`

marctc:
> Nice stuff! I have a question.
> 
> AFAIU the counters (`JVMClassLoaded` `JVMClassUnloaded` and `JVMCPUTime`) are built to aggregate per `jvmRuntimeEntryKey` i.e. `pid` and `generation` while the gauges `JVMClassCount`, `JVMCPUCount` and `JVMCPURecentUtilization` are not like that and they are created with `nil` attributes and `JVMThreadCount` only with the daemon `bool`. It means that every PID under the same UID writes the same series (which it should be fine because `svc.UID.Instance` is `namespace.pod.container` in kubernetes and it is the container name in docker and to have a collision we should have two jvm processes in one container...).
> 
> The question is: shouldn't the creation of counters and gauges be the same?

The counters and gauges use the same exported service attributes. The PID and generation key calculates each process counter delta before adding it to the shared series. Gauges need no baseline and follow the existing Go scalar-gauge behavior. If we want process-aware scalar aggregation, I suggest that a separate change covers both Go and JVM.

pinoOgni:
> > Nice stuff! I have a question.
> > AFAIU the counters (`JVMClassLoaded` `JVMClassUnloaded` and `JVMCPUTime`) are built to aggregate per `jvmRuntimeEntryKey` i.e. `pid` and `generation` while the gauges `JVMClassCount`, `JVMCPUCount` and `JVMCPURecentUtilization` are not like that and they are created with `nil` attributes and `JVMThreadCount` only with the daemon `bool`. It means that every PID under the same UID writes the same series (which it should be fine because `svc.UID.Instance` is `namespace.pod.container` in kubernetes and it is the container name in docker and to have a collision we should have two jvm processes in one container...).
> > The question is: shouldn't the creation of counters and gauges be the same?
> 
> The counters and gauges use the same exported service attributes. The PID and generation key calculates each process counter delta before adding it to the shared series. Gauges need no baseline and follow the existing Go scalar-gauge behavior. If we want process-aware scalar aggregation, I suggest that a separate change covers both Go and JVM.

Oh ok, thanks!

pinoOgni (approved):

MrAlias (approved):

PR #3213, opened 2026-08-27
Derive `ProtocolType` constants from the generated BPF enum

pinoOgni:
## Summary

This PR replaces the hand-written `ProtocolType*` iota block with aliases of the bpf2go generated consts, so the Go values are derived from the compiled `protocol_type` C enum instead of being paired with it by hand. 

In this way we avoid of having an unmatched value that falls through the classified-protocol dispatch switch and events silently degrade to userspace heuristic detection, which produces identical spans with no error and no failing test.

Notes:
- Nothing outside `pkg/ebpf/common` uses these constants, so the change is contained to the package but external code importing `pkg/ebpf/common` and comparing these constants against raw `uint8` values gets a compile error. It is just a one-line fix per site.

## Validation

### DIFF
--- a/bpf/common/common.c
+++ b/bpf/common/common.c
@@ -16,7 +16,8 @@
 
 #include <common/common.h>
 
-// Force emitting struct http_request_trace into the ELF for automatic creation of Golang struct
+// Force emitting these types into the ELF so bpf2go generates their Go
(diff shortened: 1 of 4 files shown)


mariomac (approved):
Very smart!

mmat11 (approved):

PR #3217, opened 2026-08-27
Correctly attribute HTTP subtype spans + classify to correct metrics

NimrodAvni78:
## Summary

HTTP client subtypes that have their own semantic convention stop being reported as HTTP, in both spans and metrics. An S3 call is an S3 call, a chat completion is a GenAI call, an SQS send is a messaging operation.

OBI already names these spans by the domain convention (`s3.PutObject`, `chat gpt-4`, `search my_index`, `tools/call get_weather`). Only the attributes and the metric routing still described the HTTP exchange that carried the operation. This makes them agree.

## Spans

| Subtype | Before | After |
| --- | --- | --- |
| plain HTTP, GraphQL | full HTTP surface | unchanged |
| Elasticsearch | full HTTP surface | `url.full` + `http.request.method` only |
| AWS S3, AWS SQS | HTTP + domain | domain only |
| GenAI (OpenAI, Anthropic, Gemini, Qwen, Ollama, OpenAI-compatible, Bedrock, embeddings, rerank, retrieval) | HTTP + domain | domain only |
| MCP | HTTP + domain | domain only |
| JSON-RPC | HTTP + domain | domain only |
| SQL++ | domain only | unchanged |

Removed from the affected subtypes: `url.full`, `url.scheme`, `url.query`, `http.request.method`, `http.response.status_code`, `http.request.body.size`, `http.response.body.size`. `server.address`, `server.port` and `service.peer.name` stay on every subtype.

Elasticsearch keeps `url.full` and `http.request.method` because `span.db.elasticsearch.client` marks both `requirement_level: required`. It is the only span group in the upstream registry that requires HTTP attributes.

Server spans are untouched. OBI enriches the HTTP server span in place with MCP, JSON-RPC and GraphQL attributes, and that span genuinely is an HTTP server request. `span.graphql.server` in particular defines no peer or network attributes at all, so it cannot stand alone. Whether OBI should emit a nested domain span under the HTTP server span is a separate question.

## Metrics

Two subtypes were falling through to `http.client.request.duration` despite carrying full domain attributes:

- **AWS S3** now records `rpc.client.call.duration`, matching how JSON-RPC is already handled. S3 spans carry `rpc.system.name=aws-api` and `rpc.method` from the AWS SDK conventions.
- **AWS SQS** now records `messaging.client.operation.duration`, alongside Kafka, MQTT, NATS and AMQP. Upstream defines that metric for operations initiated by a producer or consumer client, with `send`, `receive` and `ack` as example operation names, so it covers both SendMessage and ReceiveMessage.

`DeleteMessage` is the SQS ack, so it now maps to `messaging.operation.type=settle` — a member of the upstream enum (`create`, `send`, `receive`, `process`, `settle`) that OBI previously left unmapped, which meant the span omitted the attribute entirely. SQS spans also now carry `messaging.system`, which they never emitted: the metric data point said `messaging.system=aws_sqs` while the span it came from said nothing.

Instrument creation was widened so it covers every case that can route to it: the RPC client histogram is now created under `grpc || sunrpc || http`, and the messaging histograms under `mq || http`. so `sqs` can actually emit queuing metrics, if no sqs spans are emitted the metric also wont emit

## Not included

MCP metrics are deliberately left alone. They are defined in the [gen-ai semconv](https://github.com/open-telemetry/semantic-conventions-genai/blob/67dff024110be5bd9f318006e733f4078e0f4c97/model/mcp/metrics.yaml#L19) and OBI currently does not import it.
MCP client calls are recorded as `gen_ai.client.operation.duration` and MCP server calls fall through to `http.server.request.duration`. This might be an interesting features but is seperate from this PR

## Compatibility

This removes attributes from exported spans and moves S3 and SQS traffic to different metrics. Consumers filtering GenAI or AWS spans on `url.full` or `http.response.status_code`, or reading S3/SQS latency from `http.client.request.duration`, need to move to the domain attributes and metrics. The attribute removals have to be recorded as a schema transformation in the next schema version at release prep.

## Validation


pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -656,23 +675,32 @@ func traceAttributesSelectorInternal(span *request.Span, optionalAttrs map[attr.
...
+		transport := httpClientTransportScope(span.SubType)
+
 		attrs = []attribute.KeyValue{
-			request.HTTPResponseStatusCode(span.Status),
-			request.HTTPUrlFull(url),
-			semconv.URLScheme(scheme),
 			request.ServerAddr(host),
 			request.PeerService(request.PeerServiceFromSpan(span)),
 			request.ServerPort(span.HostPort),
-			request.HTTPRequestBodySize(int(span.RequestBodyLength())),
```
grcevski, line 666: I think the sizes might be useful for people in any case, but maybe there's a replacement. For example, while I was working at Elastic, how big the Elasticsearch response is, was the first thing we'd ask when debugging slow search response. We'd lose that now with this change.

NimrodAvni78: well elastic conventions dont have anything for this
general db conventions have `db.response.returned_rows` which might be applicable but not the exact same
if you think its necessary we can keep it and deviate from conventions a bit

grcevski: I don't have a very strong opinion about it, I just think it's useful to debug requests. I guess for returned rows we'll need to parse the response from Elasticsearch, is that feasible?

I wonder what other think. I personally think we should deviate from the spec if we have valuable info to share with the users that we already collect.

NimrodAvni78: ok ill wait for more reviews on this

pkg/export/prom/prom.go
```diff
@@ -1068,8 +1068,11 @@ func (r *metricsReporter) observe(span *request.Span) {
 			case r.is.DBEnabled() && (span.SubType == request.HTTPSubtypeSQLPP || span.SubType == request.HTTPSubtypeElasticsearch):
 				r.observeHistogram(r.dbClientDuration.WithLabelValues(labelValues(span, r.attrDBClientDuration)...).Metric, duration, span)
 			case span.SubType == request.HTTPSubtypeJSONRPC && r.is.GRPCEnabled():
-				// JSON-RPC client calls over HTTP get recorded as RPC client metrics
 				r.observeHistogram(r.grpcClientDuration.WithLabelValues(labelValues(span, r.attrGRPCClientDuration)...).Metric, duration, span)
+			case span.SubType == request.HTTPSubtypeAWSS3:
```
mmat11, line 1072:
```
The new S3 and SQS cases skip the enablement check every neighbour case has.
With http, grpc, sunrpc and mq all disabled, those histograms stay nil.
OBI then crashes on the first S3 or SQS client call.

Fix: guard both cases with r.is.HTTPEnabled(), and do the same in metrics.go.
```

sorry for the claudish

pkg/export/otel/tracesgen/tracesgen.go
NimrodAvni78: @grcevski iv'e added response size back as an attribute for elsaticsearch/opensearch

pkg/appolly/app/request/span.go
```diff
@@ -219,6 +220,21 @@ func MessagingOperationTypeOf(operationName string) string {
 	return operationName
 }
 
+// IsMessagingClientOperation reports whether an AWS SQS span describes a
+// producer or consumer operation. Queue administration calls carry no
+// messaging.operation.type and are not messaging client operations.
+func IsMessagingClientOperation(span *Span) bool {
```
mmat11: nit: I'd put `SQS` in this function name to signal it's not related to generic spans but to SQS ones

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -546,6 +546,25 @@ func messagingOperationAttrs(method string) []attribute.KeyValue {
...
 
+type httpTransportScope int
+
+const (
+	httpTransportAll httpTransportScope = iota
+	httpTransportRequestOnly
+	httpTransportNone
+)
+
+func httpClientTransportScope(subType int) httpTransportScope {
```
mmat11, line 563: should this be kept updated by future protocols? in that case I think it would be better to have a comment explaining what's this used for

mmat11 (approved):
lgtm

PR #3219, opened 2026-08-27
Declare OBIs full telemetry schema

NimrodAvni78:
## Summary

Declares the full contract of what OBI emits: 84 metrics and 15 span shapes, restructured from flat files into one directory per domain under `schemas/obi/groups/`.

This head was 66 commits behind `main`, so it also catches the declarations up with emission changes that landed since, and adds a provenance test so the registry cannot drift from upstream unnoticed.

## Emission changes now described

- **`error.type`** is added to both messaging metrics, which now carry it. `rpc.client.call.duration` already declared it; the server metric has no path to it in `getDefinitions`, so it stays undeclared until the Go side is widened.
- **`v8js.resource.active`** is added to the Node.js imports. It is an upstream metric OBI emits unchanged, so it is imported rather than redeclared, like the other fourteen there.
- **`obi.error` is gone**, replaced by `error.type` on the internal export-error metrics. Already recorded under "Pending transformations" in `devdocs/telemetry-schema.md`.
- **`span.obi.messaging.client`** is a forward declaration. Messaging span kind now follows the operation rather than the connection side, so `receive` and `settle` map to client. No transform currently produces either for Kafka, MQTT, NATS or AMQP, and SQS takes the HTTP client path, so nothing emits this shape yet — it is declared so that the first emitter does not land undeclared.

## Attributes that are declared but not emitted

`httpClientTransportScope` returns `httpTransportNone` for every GenAI subtype, S3, SQS and MCP, which gates off `url.full`, `url.scheme`, `url.query`, `http.request.method` and the status-code and body-size attributes. Those seven are removed from `span.obi.gen_ai.client` and `span.obi.aws.client`. Elasticsearch is `httpTransportRequestOnly`, so `span.obi.elasticsearch.client` keeps `url.full` and `http.request.method` and loses the other five.

## `gen_ai.provider.name`

The enum gains six members: `ollama`, the four OpenAI-compatible gateways SUPPORT_MATRIX lists (`litellm`, `vllm`, `localai`, `openrouter`), and `custom`.

Most provider values come from OBI's own detection tables and are enumerable. One was not: `openai_compatible.gateways[].provider` is user configuration, reported verbatim, so any value could be emitted and any value outside the enum is a live-check violation.

That configuration is meant to name a gateway OBI does not know, so rejecting an unknown name would defeat its purpose. Instead the value is normalized where the attribute is read: a name with no member reports as `custom`, and the gateway stays identifiable through `server.address` on the same span. Existing configurations keep working and no value outside the contract is emitted.

## MCP

`span.obi.mcp.client` is a new group. MCP is not a GenAI operation — `IsGenAISubtype` excludes it, because it is a tool and resource protocol rather than a model provider and has no `gen_ai.provider.name` to report — so the attributes only `mcpAttributes` emits (the `mcp.*` set, `gen_ai.tool.*`, `gen_ai.prompt.name`) move off `span.obi.gen_ai.client` onto it.

MCP seen on the server side stays on `span.obi.http.server`, which OBI enriches in place; that span genuinely is an inbound HTTP request.

## Live-check resolution

`--include-unreferenced` is removed from the seven places weaver `live-check` is invoked, so the enforced contract is the set of groups some signal actually references. Three consequences follow:

- Two attribute groups no signal references — `registry.obi.exception` and `registry.obi.traces_resource` — drop out of that resolution. `exception.message` and `otel.scope.name` are expected to resolve from their upstream signal roots instead; the weaver-validated suites are what confirm it.
- `deployment.environment.name` gains a `missing_attribute` / `extends_namespace` filter in `.weaver.toml`. It is forwarded from the workload rather than set by OBI, so it belongs to no OBI group and would otherwise surface as a violation. This is a deliberate, permanent widening of what live-check tolerates.
- Node.js and V8 metrics are imported in `nodejs/imports.yaml` rather than left to resolve implicitly.

`generate-schema-docs.sh` keeps the flag deliberately: the published reference is a superset that documents what OBI declares, including groups live-check cannot hold it to.

## Carrier membership

Live-check resolves an emitted attribute against the resolved registry's global attribute map, not against the group describing the span it arrived on, so an attribute dropped from one carrier while still declared on another leaves CI green. `TestEmittedSpanAttributesMatchDeclaredGroup` in `internal/schemacheck` closes that gap for `span.obi.elasticsearch.client`: it drives `TraceAttributesSelector` with a span populated in every field the Elasticsearch path reads and asserts the emitted keys are exactly what the group declares. It fails in both directions — removing `http.response.body.size` from the group, or removing the exporter's append. The harness is a table, so further carriers can be pinned the same way.

## Published reference

`make generate-schema-docs` rendered only attributes and metrics, so the span declarations would have published nothing. It gains a spans page, documenting the span kind alongside each shape. The reference is now 21 attribute groups, 84 metrics and 15 spans.

That surfaced a latent bug in `schema-docs.jq`: `examples` is a list in most declarations but a bare scalar in some upstream ones, and iterating it blindly aborted the whole render. The attributes and metrics pages never reached one; a span attribute does.

## Validation

Neither document records per-signal declarations, so there is nothing there to update.

`make lint-schema`, `make check-schema-files`, `make test-schema` and `make check-config-schema` pass, as do the unit suites. Tests are added for spans-page selection by provenance and kind, a scalar `examples` declaration which fails without the coercion, the gateway-provider normalization, and a guard asserting the Go value space and the declared enum still agree, which fails if a provider is added to one and not the other. The weaver-validated integration, OATS and k8s suites exercise the live-check change and have not been run locally; CI covers them. They are the check that matters most here — whether `exception.message` and `otel.scope.name` still resolve once their groups leave the resolution is a question only live-check answers.


NimrodAvni78:
Might have to change once any of https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3217 https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3196 https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3195 https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3194 get in

skl (approved):
lgtm

schemas/obi/groups/gen_ai/spans.yaml
```diff
@@ -0,0 +1,76 @@
...
+    brief: OBI GenAI client span.
+    attributes:
+      - ref: http.request.method
+      - ref: http.response.status_code
+      - ref: url.full
+      - ref: url.scheme
+      - ref: url.query
+      - ref: http.request.body.size
+      - ref: http.response.body.size
+      - ref: gen_ai.provider.name
```
MrAlias, line 19: This new span contract references a closed provider enum, but two supported emission paths produce values outside it: native Ollama emits `ollama`, while `openai_compatible.gateways[].provider` accepts an unrestricted string and falls back to `custom` (the exporter tests explicitly cover `litellm` and `custom`). Because Weaver promotes `undefined_enum_variant` to a violation, live-check will reject those emitted values. Could we include Ollama and make the configured gateway-provider value space valid here (or constrain the configuration to an enumerated set), with schema coverage for both paths?

MrAlias (commented):
The declaration coverage is much stronger, but the emitted contract and its published reference are not yet internally consistent.

schemas/obi/manifest.yaml
```diff
@@ -7,8 +7,7 @@ description: |
   the standard semconv set:
 
   - The `target.info` meta-metric family (Prometheus / OpenMetrics
-    `target_info` heritage, emitted in OTel dot notation over OTLP) and the
-    `instance` / `job` scrape-convention attributes
+    `target_info` heritage, emitted in OTel dot notation over OTLP)
```
MrAlias, line 10: Current `main` now generates and publishes the registry reference under `site/docs/` with `make generate-schema-docs`, and `AGENTS.md` requires committing that output whenever the registry changes because CI does not verify it. This head predates that generator and leaves the current 17-group/44-metric pages unchanged, so merging these declarations would publish a stale telemetry reference. Could we sync with current `main`, run the generator, and include the resulting `site/docs` updates?

NimrodAvni78:
@MrAlias fixed both, and also added all the recent schema changes made in main

grcevski:
There are some conflicts with main it seems now, maybe recent merges caused this.

schemas/obi/groups/db/spans.yaml
```diff
@@ -0,0 +1,67 @@
...
+  # protocols covered by span.obi.db.client never emit http.* / url.*, which is
+  # why this is a separate group rather than extra refs on the shared one.
+  - id: span.obi.elasticsearch.client
+    type: span
+    span_kind: client
+    stability: development
+    brief: OBI Elasticsearch client span, detected from HTTP client traffic.
+    attributes:
+      - ref: http.request.method
+      - ref: url.full
```
MrAlias, line 38: `traceAttributesSelectorInternal` still appends `http.response.body.size` for an HTTP client span when `span.SubType == HTTPSubtypeElasticsearch` and `span.Elasticsearch != nil`, even though the common transport path is limited to request-only attributes for this subtype. This group therefore omits an attribute OBI emits, and the generated span reference omits it too. The Elasticsearch integration sends these spans through Weaver, but live-check validates this key against the resolved registry’s global attribute map rather than enforcing membership in this carrier. Since `http.response.body.size` is declared elsewhere, CI remains green despite the mismatch.

Could we either keep `http.response.body.size` in this group or remove the Elasticsearch-specific append and update the exporter test? In either case, could we add a contract assertion that keeps this carrier and the exporter’s Elasticsearch attribute set in sync?

site/docs/spans.md
```diff
@@ -0,0 +1,454 @@
...
+| `db.namespace` | string | stable | The name of the database, fully qualified within the server address and port. | customers; test.users |
+| `db.operation.name` | string | stable | The name of the operation or command being executed. | findAndModify; HMSET; SELECT |
+| `db.query.text` | string | stable | The database query being executed. | SELECT * FROM wuser_table where username = ?; SET mykey ? |
+| `db.response.status_code` | string | stable | Database response status code. | 102; ORA-17002; 08P01; 404 |
+| `db.system.name` | enum | stable | The database management system (DBMS) product as identified by the client instrumentation. | other_sql; softwareag.adabas; actian.ingres; aws.dynamodb; aws.redshift; azure.cosmosdb; intersystems.cache; cassandra; … |
+| `elasticsearch.node.name` | string | development | Represents the human-readable identifier of the node/instance to which a request was routed. | instance-0000000001 |
+| `error.type` | enum | stable | Describes a class of error the operation ended with. | timeout; java.net.UnknownHostException; server_certificate_invalid; 500 |
```
MrAlias: This new reference renders `error.type` as an enum for Elasticsearch, while the GenAI row in the same generated page and `x.obi.error` declare it as a string. The generator is publishing whichever duplicate definition Weaver embedded in each carrier, so the per-signal contract is internally contradictory. The docs script explicitly accepts nondeterministic override resolution and does not verify its output in CI, which lets this remain green.

Could we make the renderer substitute OBI’s `x.obi.*` override definition before formatting carrier attributes and add a focused test that asserts every OBI carrier renders `error.type` as `string`?

MrAlias (commented):
The refreshed registry resolves the earlier provider and publication gaps, but cross-layer validation still permits the emitted and generated contracts to diverge.

schemas/obi/groups/gen_ai/spans.yaml
```diff
@@ -61,3 +61,5 @@ groups:
       - ref: network.peer.address
       - ref: network.peer.port
       - ref: network.protocol.version
+      - ref: user_agent.original
```
MrAlias, line 64: The four specialized client groups now declare `user_agent.original`, but each successful AWS, GenAI, MCP, and Elasticsearch transform returns from `postProcessHTTPSpan` before `HTTPEnricher.Enrich` populates `RequestHeaders`. `userAgentAttributes` only falls back to `span.UserAgent` for server spans, so these client shapes cannot emit this attribute even when selected. Should we just remove the four refs for client spans, or did we actually want to capture these?

NimrodAvni78: right!
fixed!

schemas/obi/groups/db/spans.yaml
MrAlias: The carrier now includes `http.response.body.size`, which resolves the emitted-attribute mismatch. The requested cross-layer assertion is still missing: Weaver only verifies that this key exists globally, so removing it from `span.obi.elasticsearch.client` would still leave CI green. Could we add that assertion before resolving this thread?

NimrodAvni78: Added`TestEmittedSpanAttributesMatchDeclaredGroup` in internal/schemacheck. It drives TraceAttributesSelector with a fully-populated Elasticsearch client span and asserts the emitted keys are exactly what span.obi.elasticsearch.client declares, so removing http.response.body.size from either side now fails. It runs in the standard unit test job, and the harness is a table so other carriers can be added.

MrAlias (approved):
Awesome work 👏

PR #3212, opened 2026-08-27
ran 'go fix'  to modernize code to Go 1.26

mariomac:
## Summary

We recently updated the Go version from 1.25 to 1.26. Running `go fix ./...` to apply some automatic modernizations of the coding style and standard API usage.

Also added a linter option to make sure we keep modernized code style.

## Validation

- [ ] I have read and followed the [contributing guidelines](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/CONTRIBUTING.md)
- [ ] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed.

PR #3221, opened 2026-08-27
Support Go 1.27 target instrumentation

MrAlias:
## Summary

- recover fully qualified interface implementation addresses from Go 1.27 runtime metadata when stripped binaries do not expose named `go:itab` symbols
- generate the required runtime ABI field offsets, type sizes, and constants from released Go toolchains through the existing offset update flow
- register the standard library's relocated `net/http/internal/http2` probes, ownership probes, and struct offsets
- unwrap the typed gRPC credential and TLS connection layers before resolving the underlying socket
- defer application-wrapped TLS clients to the existing writer handoff instead of accepting an empty socket tuple from the direct path
- restore Go 1.27 build and test images so integration and OATS jobs exercise these compatibility paths

## Background

Updating the target applications to Go 1.27 exposed two changes in the Go toolchain that OBI depends on. The linker now stores unnamed itabs contiguously in `.go.type`, and runtime type display names no longer include the complete package path. The standard library's bundled HTTP/2 implementation also moved from `net/http` to `net/http/internal/http2`, changing the symbols and struct ownership used by OBI's probes.

The upgrade also exposed connection-unwrapping assumptions that are independent of the `google.golang.org/grpc` module layout. TLS-enabled gRPC servers can wrap the socket in credential and TLS interfaces before reaching `net.TCPConn`. Application-wrapped TLS clients can make the direct `persistConn` read appear successful with an empty tuple, masking the valid tuple published later by the TLS writer. The new type-address checks unwrap only recognized layers and preserve the existing writer handoff for unknown wrappers.

The original Go image update exposed these failures and was reverted in #3206. The source modernization in #3212 allows OBI itself to build with the newer toolchain, but it does not address instrumentation of Go 1.27 target binaries.

## ABI maintenance and compatibility

The stripped-binary fallback depends on private Go runtime ABI facts that the field-only offset tracker does not provide. A companion generator extracts the required field offsets, type sizes, and constants from each supported Go release's DWARF and writes them in the existing versioned offset format.

The normal field offsets remain in `offsets.json`, while the additional ABI facts are generated into `go_abi_offsets.json`. This separation is intentional in this PR: the existing tracker cannot emit type sizes or constants, and its writer replaces its complete output, so running both generators against one file would cause one result set to erase the other.

Both artifacts are updated by the same `make update-offsets` target, and the scheduled workflow only commits after both generators succeed. They are embedded from the same repository commit. Runtime decoding also rejects an ungenerated future Go minor instead of silently applying stale ABI metadata.

Patch releases use the generated ABI for their Go minor version. Existing named-itab discovery and the existing standard-library and `x/net/http2` probe groups remain unchanged for older targets; the new moduledata and internal HTTP/2 paths are gated to Go 1.27 and later.

Follow-up #3234 tracks making the two generation paths atomic or adding explicit cross-artifact version validation. Consolidating them here would require extending or replacing the tracker's orchestration while preserving its version discovery, caching, and third-party module behavior, which is broader than the Go 1.27 compatibility fix.

Follow-up #3235 tracks dynamically discovering a complete runtime ABI from target DWARF before using generated data. This PR intentionally uses generated ABI facts for both stripped and non-stripped binaries; stripped binaries cannot support dynamic discovery, and an unknown future Go minor is rejected safely until that follow-up is implemented.

## Validation

- `make docker-generate`
- `make lint`, including the Linux and macOS CI jobs
- `go test -short -race -count=1 ./pkg/internal/ebpf/gotracer ./pkg/internal/goexec`
- `go test -race -count=1 -run '^TestWrappedTLSConnClientSpans$' -v ./internal/test/integration`
- regenerated `go_abi_offsets.json` twice and verified byte-for-byte identical output
- verified stripped amd64 and arm64 Go 1.27 binaries resolve itabs, all required internal HTTP/2 symbols, and their generated struct offsets


bpf/gotracer/go_grpc.c
```diff
@@ -55,11 +55,30 @@ static __always_inline void grpc_server_conn_info(void *tr, connection_info_t *c
     }
 
     off_table_t *ot = get_offsets_table();
+    void *conn_type = NULL;
     void *conn_ptr = NULL;
-    bpf_probe_read_user(&conn_ptr,
-                        sizeof(conn_ptr),
-                        (void *)(tr + go_offset_of(ot, (go_offset){.v = _grpc_st_conn_pos}) +
-                                 k_go_iface_data_offset));
+    void *conn_iface = (void *)(tr + go_offset_of(ot, (go_offset){.v = _grpc_st_conn_pos}));
```
grcevski, line 60: Hm, interesting, so it's one more level of nesting to get to the info. It sounds like we need to version check to support older go?

MrAlias: The older version support should already be handled by the runtime type checks, so we shouldn't need a Go-version check. We only unwrap when the interface’s concrete type matches `*credentials.syscallConn` or `*tls.Conn`; otherwise, the original connection pointer path is preserved.

The extra nesting is also tied to the gRPC connection wrapper rather than Go 1.27 specifically, so checking the concrete type is more accurate than checking the Go version.

grcevski: Oh sorry, I messed up, I confused iface with type.

bpf/gotracer/go_grpc.c
```diff
@@ -55,11 +55,30 @@ static __always_inline void grpc_server_conn_info(void *tr, connection_info_t *c
...
+    bpf_probe_read_user(&conn_type, sizeof(conn_type), conn_iface);
+    bpf_probe_read_user(&conn_ptr, sizeof(conn_ptr), conn_iface + k_go_iface_data_offset);
+
+    const u64 syscall_conn_type = go_offset_of(ot, (go_offset){.v = _grpc_syscall_conn_type_off});
```
grcevski: Do we want to use the trick of -1 for the unset new offsets here (and elsewhere), since we started doing it, or it's not possible for this offset to ever be 0?

bpf/gotracer/go_nethttp.c
```diff
@@ -1798,15 +1798,26 @@ int GUARDED_PROG(obi_uprobe_persistConnRoundTrip, struct pt_regs *, ctx) {
...
             sizeof(tls_state),
             (void *)(pc_ptr + go_offset_of(ot, (go_offset){.v = _pc_tls_pos}))); // find tlsState
         bpf_dbg_printk("conn_conn_ptr=%llx, tls_state=%llx", conn_conn_ptr, tls_state);
 
+        if (tls_state) {
+            void *conn_type = NULL;
+            bpf_probe_read_user(&conn_type, sizeof(conn_type), conn_iface);
+            const u64 tls_conn_type = go_offset_of(ot, (go_offset){.v = _tls_conn_type_off});
+            if (tls_conn_type && conn_type != (void *)tls_conn_type) {
+                bpf_dbg_printk("wrapped TLS client connection, waiting for writer handoff");
```
grcevski, line 1816: Can you explain why we reject the connection here, what does it mean that we are waiting for the writer handoff? Does it reattempt on connection upgrade to TLS?

grcevski (approved):
Good stuff, I just have some minor suggestions and one question.

bpf/gotracer/go_grpc.c
MrAlias: These values are interface type/itab addresses rather than struct field offsets. A valid address cannot be zero, so zero is an unambiguous missing value and matches the existing `ITypes` behavior. The `-1` sentinel is needed for optional field offsets because zero can be a valid field position. Using `-1` here would also require changing the truthiness guards, particularly the wrapped-TLS check.

I can clarify the naming/comment so this distinction is visible. Specifically, I can rename them to `GrpcSyscallConnTypeAddress`/`TLSConnTypeAddress` and the BPF entries to `*_type_addr`. What do you think?

bpf/gotracer/go_nethttp.c
MrAlias: We are not rejecting the connection or the request here; we are only abandoning the direct eBPF memory walk.

When `tlsState` is present but `persistConn.conn` is not exactly `*tls.Conn`, the application has wrapped the TLS connection in an unknown `net.Conn` implementation. We cannot safely interpret that wrapper’s memory using the `tls.Conn` layout. Doing so can produce a bogus connection tuple that masks the correct fallback (this is addressing an existing bug in our implementation uncovered in the fixes for Go 1.27 upgrade).

“Waiting for the writer handoff” means relying on the existing write-side correlation: this probe has already associated the request goroutine with the `persistConn`. Later, `persistConnWriter.Write` associates the writer goroutine with that same `persistConn`, and the `crypto/tls.(*Conn).Write/netFD.Write` probe obtains the real socket tuple and publishes it back to the request. The `RoundTrip` return probe uses that tuple when the direct lookup is absent.

There is no retry and this is not related to upgrading a connection to TLS. TLS is already established here; we are deferring socket discovery until the normal request write reaches a location where the underlying connection is known.

pkg/internal/goexec/itab.go
```diff
@@ -5,15 +5,51 @@ package goexec // import "go.opentelemetry.io/obi/pkg/internal/goexec"
...
+	go127KindPointer     = 22
+	go127KindSlice       = 23
+	go127KindStruct      = 25
+	go127UncommonArray   = 72
+	go127UncommonChan    = 64
+	go127UncommonDefault = 48
+	go127UncommonMap     = 136
+	go127UncommonOneWord = 56
+	go127UncommonWithPkg = 80
+	maxGoTypeNameLen     = 4096
```
MrAlias: These are pretty unmaintainable. Looking into updating out offset tooling to support these values being dynamically discovered.

bpf/gotracer/go_nethttp.c
grcevski: right ok, so we don't see the connection here and we hope we can pick it up on the uprobes that handle the request. makes sense.

bpf/gotracer/go_grpc.c
grcevski: I think it's fine as is, you are right. It's up to you if you want to name them or put a comment saying this cannot be 0 if they exist.

bpf/gotracer/go_nethttp.c
MrAlias: exactly 😄

bpf/gotracer/go_grpc.c
MrAlias: Updated in 0e35d4d38. I renamed the Go table entries to `GrpcSyscallConnTypeAddress`/`TLSConnTypeAddress`, the BPF identifiers to `*_type_addr`, and the local variables to `*_type_addr`. This keeps zero as the missing-address sentinel while making it clear these values are runtime addresses rather than struct field offsets.

grcevski:
Ship it!

PR #3227, opened 2026-08-28
Gate Go runtime metrics probes behind the `application_runtime` feature 

pinoOgni:
## Summary

Go runtime metrics uprobes were attached to every instrumented Go binary regardless of user configuration and the resulting events discarded in userspace when the `application_runtime` metric features was off.

This PR adds the check with `AppRuntimeMetricsEnabled()` (used also by the generictracer) and skips probe attachment, per-process target registration, and per-binary availability detection when the feature is disabled. In this way when the feature is off, instrumented Go apps no longer execute the related bpf runtime metrics program on every GC cycle or emit ring-buffer events that the userspace throws away. 

## Validation


marctc (approved):
Thanks, nice catch!

grcevski (approved):
LGTM!

PR #3228, opened 2026-08-28
store-demo: add an Aerospike-backed recentlyviewed service

### LINKED ISSUE #2532: Add aerospike to obi-store demo

bhargav0605:
## Summary

Closes #2532.

The store demo had no Aerospike workload, so the instrumentation added in #2545 wasn't demonstrated anywhere. This adds `recentlyviewed`, a small Java service that records product views in Aerospike, called by `frontend` from the product page.

Browsing a product now produces a single trace across all three tiers — `frontend` (Go) → `recentlyviewed` (Java) → Aerospike — with no SDK in any of them:

trace-tree

The Aerospike span carries full DB semantic conventions, and OBI attributes it to its Kubernetes workload with no extra configuration:

attributes

### Notes for reviewers

A few choices worth flagging, since several aren't obvious from the diff:

- **The OBI image bump is required, not incidental.** The demo pinned `v0.9.0`, which predates Aerospike support (first shipped in `v0.11.0`). Moved to `v0.12.2`, reusing the digest the other examples already use.

- **`ebpf.buffer_sizes.aerospike` is also required.** Once a connection is classified as Aerospike the generic tcp buffer no longer applies to it, and this setting defaults to `0` — so without it nothing is captured and no spans are produced. Same reason `docker-compose-aerospike.yml` sets it.

- **The service lives under `services/`, not `app/src/`**, to leave the vendored Online Boutique tree pinned at v0.10.5 unmodified. `fix_architecture.py` now searches both roots, and recognises `pom.xml` as a Java marker.

- **It speaks HTTP rather than gRPC.** gRPC would mean a proto file and generated stubs in the vendored frontend; the Aerospike traffic is the point of the service, and `frontend` already serves HTTP.

- **The `frontend` change is optional and best-effort** — `mapOptionalEnv`, a one second timeout, warn-and-continue on failure, matching the existing `getRecommendations` treatment. Verified by scaling `recentlyviewed` to zero replicas: the product page still returns 200.

`db.query.text` is intentionally not enabled. The client sets `sendKey`, but OBI emits the primary key only when the attribute is explicitly selected, and it is application data.

<details>
<summary>The demo storefront driving the traffic</summary>

store

</details>

## Validation

- [ ] N/A — this adds a demo workload rather than changing a core feature, so `devdocs/features.md` and `SUPPORT_MATRIX.md` are unchanged.

Verified end to end on a `kind` cluster. `test_fix_architecture.py` and `fix_architecture.py --check` pass, and `make lint-markdown` is clean.

---

Generative AI (Claude) was used to explore the codebase, draft the service and manifests, and help diagnose why spans initially didn't appear. I ran and verified everything on my own cluster, and reviewed and revised the code before submitting.

### DIFF
--- a/examples/store-demo/app/src/frontend/handlers.go
+++ b/examples/store-demo/app/src/frontend/handlers.go
@@ -154,6 +154,9 @@ func (fe *frontendServer) productHandler(w http.ResponseWriter, r *http.Request)
 		renderHTTPError(log, r, w, errors.Wrap(err, "could not retrieve product"), http.StatusInternalServerError)
 		return
 	}
+
+	fe.recordProductView(r.Context(), log, id)
+
 	currencies, err := fe.getCurrencies(r.Context())
 	if err != nil {
 		renderHTTPError(log, r, w, errors.Wrap(err, "could not retrieve currencies"), http.StatusInternalServerError)
@@ -204,6 +207,36 @@ func (fe *frontendServer) productHandler(w http.ResponseWriter, r *http.Request)
 	})); err != nil {
 		log.Println(err)
 	}
+}
+
+// recordProductView tells the recentlyviewed service that this product was
+// looked at, which is what gives the demo its Aerospike traffic.
+//
+// Optional and best-effort, like getRecommendations below: the service may not
+// be deployed, and a failure here must never affect the product page.
+func (fe *frontendServer) recordProductView(ctx context.Context, log logrus.FieldLogger, id string) {
+	if fe.recentlyViewedSvcAddr == "" {
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, time.Second)
+	defer cancel()
+
+	url := fmt.Sprintf("http://%s/viewed/%s", fe.recentlyViewedSvcAddr, id)
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, nil)
+	if err != nil {
+		log.WithField("error", err).Warn("failed to build recently-viewed request")
+		return
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		log.WithField("error", err).Warn("failed to record recently-viewed product")
+		return
+	}
+	defer resp.Body.Close()
+	// Drain so the connection can be reused.
+	_, _ = io.Copy(io.Discard, resp.Body)
 }
 
 func (fe *frontendServer) addToCartHandler(w http.ResponseWriter, r *http.Request) {
--- a/examples/store-demo/app/src/frontend/handlers.go
+++ b/examples/store-demo/app/src/frontend/handlers.go
@@ -155,7 +155,9 @@ func (fe *frontendServer) productHandler(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	fe.recordProductView(r.Context(), log, id)
+	if r.Method == http.MethodGet {
+		fe.recordProductView(r.Context(), log, id)
+	}
 
 	currencies, err := fe.getCurrencies(r.Context())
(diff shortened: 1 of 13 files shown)


examples/store-demo/k8s/03-obi-values.yaml
```diff
@@ -3,8 +3,10 @@ namespaceOverride: obi-store-demo
 image:
   registry: docker.io
   repository: otel/ebpf-instrument
-  tag: v0.9.0
-  digest: sha256:26f82b148dfe8cb0530561ab72a3cb5490b3ae5df556a33c27984af2e28542cf
+  # Aerospike instrumentation shipped in v0.11.0 (#2545); v0.9.0 predates it.
```
mmat11: I think we can remove this comment

examples/store-demo/k8s/frontend.yaml
```diff
@@ -80,6 +80,8 @@ spec:
             value: "checkoutservice:5050"
           - name: AD_SERVICE_ADDR
             value: "adservice:9555"
+          - name: RECENTLYVIEWED_SERVICE_ADDR
```
mmat11:
```suggestion
          - name: RECENTLY_VIEWED_SERVICE_ADDR
```

examples/store-demo/k8s/03-obi-values.yaml
bhargav0605: Sure, removed it.

examples/store-demo/k8s/frontend.yaml
bhargav0605: I agree, I have corrected it.

MrAlias (commented):
The new service and topology integration look coherent. The frontend recording hook needs to preserve the existing route's safe-method behavior.

examples/store-demo/app/src/frontend/handlers.go
```diff
@@ -154,6 +154,9 @@ func (fe *frontendServer) productHandler(w http.ResponseWriter, r *http.Request)
 		renderHTTPError(log, r, w, errors.Wrap(err, "could not retrieve product"), http.StatusInternalServerError)
 		return
 	}
+
+	fe.recordProductView(r.Context(), log, id)
```
MrAlias: `productHandler` serves both GET and HEAD, so a HEAD request now writes the product to Aerospike and updates `viewed_at`. Crawlers and `curl -I` can therefore create views even though HEAD is a safe method.

Could we keep the side effect on GET and cover the HEAD path?

```suggestion
	if r.Method == http.MethodGet {
		fe.recordProductView(r.Context(), log, id)
	}
```

bhargav0605: Thank you @MrAlias , I have applied your suggestion in 9aafc1d7.

mmat11 (approved):
lgtm

examples/store-demo/k8s/03-obi-values.yaml
```diff
@@ -3,8 +3,8 @@ namespaceOverride: obi-store-demo
 image:
   registry: docker.io
   repository: otel/ebpf-instrument
-  tag: v0.9.0
-  digest: sha256:26f82b148dfe8cb0530561ab72a3cb5490b3ae5df556a33c27984af2e28542cf
+  tag: v0.12.2
```
MrAlias:
```suggestion
  tag: v0.13.0
```

MrAlias (approved):

examples/store-demo/k8s/03-obi-values.yaml
bhargav0605: Done in ad9c0e42, digest updated to match.

PR #3237, opened 2026-08-28
Add JVM garbage collection duration metric

### LINKED ISSUE #2905: Expand JVM runtime metrics

marctc:
## Summary

This PR adds Java agent collection and export for `jvm.gc.duration`.

The agent registers notification listeners on the JVM garbage collectors. Each completed collection sends its duration through the existing Java agent `ioctl` channel.

This metric requires the injected Java agent. `javaagent.enabled` is enabled by default.

 Setting `javaagent.enabled` to `false` disables GC duration, class, thread, and CPU metrics. The existing HotSpot memory metrics remain available.  GC notifications are event-driven.

The default histogram boundaries use the semantic-convention advisory values. Users can change them through `buckets.jvm_gc_duration_histogram`.

This PR follows #3203, which added the other agent-backed JVM runtime metrics.

Fixes #2905

## Validation


MrAlias (approved):
The GC event path and exporter coverage look consistent, but the user-facing JVM docs and disabled-agent warning still describe agent-backed collection as class, thread, and CPU metrics only. Please document `jvm.gc.duration` and its event-driven behavior in `devdocs/runtimes/jvm.md`, `SUPPORT_MATRIX.md`, and the Java/JVM configuration descriptions, and update the `javaagent.enabled=false` warning and test while resolving the current conflicts with main.

pkg/ebpf/common/pids.go
```diff
@@ -144,7 +144,11 @@ func (pf *PIDsFilter) CurrentPIDs(t PIDType) map[uint32]map[app.PID]svc.Attrs {
 		cVal := map[app.PID]svc.Attrs{}
 		for kv, vv := range v {
 			if vv.pidTypes&t != 0 && vv.removedAt == 0 {
-				cVal[kv] = vv.fileInfo.UnsafeServiceAttrs()
+				serviceSource := vv.fileInfo
+				if source := serviceSource.RuntimeMetricServiceSource(); source != nil {
```
grcevski: I don't think I understand this change. Why do we need to give out a different service metadata for the JVM metrics? Was this a bug from before?

marctc: This was an attempt to fix CI issues but I think it was just flaky. Reverting now, thanks for spotting...

grcevski (approved):
LGTM!

PR #3238, opened 2026-08-28
Generate Go offsets atomically

### LINKED ISSUE #3234: Make Go ABI offset generation atomic
## Problem

#3221 adds generated Go runtime ABI facts alongside the existing generated `offsets.json`. The files are separate because `go-offsets-tracker` discovers only struct-field offsets and replaces its complete output, while the ABI generator also needs type sizes and constants.

The current workflow is safe for #3221: `make update-offsets` runs both generators, the scheduled workflow only commits after both succeed, and runtime decoding rejects an ungenerated future Go minor. However, maintaining two versioned artifacts for the same moduledata decoder creates a longer-term consistency risk. Their Go version coverage could drift, and an interrupted local update can temporarily leave only one artifact refreshed.

## Goal

Make Go field-offset and runtime-ABI generation atomic, or enforce an equivalent invariant that prevents mismatched Go version coverage from being published or consumed.

## Possible approaches

- extend `go-offsets-tracker` to support type sizes and constants, allowing one generated artifact;
- add one orchestration command that collects both result sets and invokes the tracker writer once;
- retain separate artifacts but stage updates atomically and add explicit cross-artifact version validation.

The implementation should avoid duplicating the tracker's version discovery, caching, and third-party module behavior unless the benefits justify owning that logic.

## Acceptance criteria

- a mismatch between moduledata offset coverage and runtime ABI coverage fails generation or validation;
- adding a stable Go minor updates all required metadata together;
- existing standard-library and third-party offset tracking and caching remain intact;
- an ungenerated future Go minor cannot silently use stale runtime ABI metadata;
- `make update-offsets` and the scheduled update workflow use the resulting atomic or validated path.

MrAlias:
## Summary

- replace the separate field-offset and runtime-ABI commands with one OBI-owned orchestrator that reuses the tracker packages
- consolidate Go ABI facts into `offsets.json` and publish the generated file with one validated atomic rename
- reject mismatched ABI and `runtime.moduledata` Go version coverage during generation
- preserve version-gated standard-library caching and existing third-party collection behavior
- inspect the x/net legacy HTTP/2 structs when generation runs under Go 1.27, where x/net v0.57+ otherwise exposes wrapper types

## Details

The generator collects all field offsets and runtime ABI facts before invoking the tracker writer once. It writes to a same-directory temporary file, verifies that every ABI fact and the required `runtime.moduledata` fields have identical coverage, and only then renames the file over the embedded artifact.

Standard-library fields are grouped by their configured version bounds so fields introduced in newer Go releases do not invalidate cache entries for older releases. Third-party discovery and caching continue through `go-offsets-tracker`. Generator-owned temporary storage is removed on success and failure.

Go 1.27 makes recent x/net HTTP/2 implementations wrappers around the standard library. Generation selects the `http2legacy` implementation so the tracker can inspect the external package's real structs; the standard-library collector covers the Go 1.27 replacement structs.

The runtime now reads ABI facts from the existing embedded `offsets.json`, and the separate `go_abi_offsets.json` artifact and tracker CLI tool dependency are removed.

Fixes #3234

## Validation

- `make update-offsets` twice, including x/net v0.57 and v0.58; the second run reproduced the same Git diff
- `make docker-generate`
- `make verify`


skl (approved):
ty

PR #3240, opened 2026-08-28
Support Python runtime metrics for stripped builds

marctc:
## Summary

This follow-up extends CPython runtime metrics to stripped Linux `amd64` builds, including Docker Official Python `slim` and `alpine` images.

When USDT and internal GC symbols are unavailable, OBI starts from the exported `PyGC_Collect` symbol and inspects the bounded machine code to locate `collect` or  `gc_collect_main`. CPython 3.9-3.12 use the repeated callback call shape. CPython 3.13-3.14 use the thread-state lookup and collector argument shape.

The resolver fails closed when the ELF layout or collector target cannot be validated. Linux `arm64` support remains unchanged and requires USDT.

### Reviewer note

The matching code includes short GDB-style disassembly excerpts. The `>` marker shows which instruction each branch is processing, so the code can be followed against the inspected machine-code flow.

## Validation


MrAlias (commented):
The stripped-image coverage validates the intended resolver paths, but the new ELF parser lets workload-controlled metadata drive a memory-sized allocation in the OBI process.

pkg/internal/cpython/runtime/probe_linux_amd64.go
```diff
@@ -6,29 +6,470 @@
...
+	return size, nil
+}
+
+// pythonFunctionStartEntries reads the fixed-format search table from .eh_frame_hdr.
+func pythonFunctionStartEntries(file *elf.File) (uint64, []byte, error) {
+	section := file.Section(".eh_frame_hdr")
+	if section == nil || section.ReaderAt == nil || section.Size > math.MaxInt || file.ByteOrder == nil {
+		return 0, nil, fmt.Errorf("%w: missing CPython function-start table", errUnsupportedLayout)
+	}
+	data := make([]byte, section.Size)
```
MrAlias, line 60: The section size comes from the mapped runtime, but we allocate it in full before validating the header or entry count. A workload can map a CPython-shaped ELF with an oversized `.eh_frame_hdr` and force the OBI process into a correspondingly large allocation or OOM during discovery. Could we reject oversized tables before allocating, as `legacyVersionFromELF` already does for `.rodata`?

pkg/internal/cpython/runtime/probe_linux_amd64.go
```diff
@@ -6,29 +6,471 @@
...
+	// The TLS lookup returns its block in RAX; [RAX+8] holds the current PyThreadState.
+	//
+	//	26814e: lea rdi,[rip+...]
+	//	268155: call __tls_get_addr@plt
+	//	> 26815a: mov rbx,[rax+8]
+	if state.tlsStage == tlsCallComplete {
+		destination, register := inst.Args[0].(x86asm.Reg)
+		source, memory := inst.Args[1].(x86asm.Mem)
+		if inst.Op == x86asm.MOV && register && memory &&
+			source.Base == x86asm.RAX && source.Disp == 8 {
```
pinoOgni: nit: name constant and a comment about why 8.

pinoOgni: I run some tests and AFAIU `Disp == 8` works because mimalloc's `_mi_heap_default (.tdata)` occupies TLS offset 0 on both Docker 3.13 and 3.14 builds `(readelf -s ... | grep TLS)`. 

If there is a `--without-mimalloc` build, `_Py_tss_tstate` will be at `0` and the matcher fails closed we have no wrong data, just no metrics on that build.

So, are we sure about `Disp == 8` or we can drop it and just use the later  `(rdi=tstate, esi=2, edx=2)` check that should be enough?

pinoOgni: This if accepts any `mov REG,[rax+8]` regardless of index register and so it means that `mov rbx,[rax+rcx*8+8]` would also be taken as the `_Py_tss_tstate` load, right?.  The real instruction is a fixed-field load and never has an index. What about adding `source.Index == 0`?

I dont think it is a real-world case observed but just makes the proof match what it means.

pkg/internal/cpython/runtime/probe_linux_amd64.go
```diff
@@ -6,29 +6,471 @@
...
+		return 0, nil, err
+	}
+	// The header contains four encoding bytes, a 4-byte .eh_frame pointer, and a 4-byte entry count.
+	// These encodings select version 1, PC-relative sdata4, udata4, and data-relative sdata4.
+	if len(data) < ehFrameHeaderSize ||
+		!bytes.Equal(data[:len(supportedEHFrameHeaderEncoding)], supportedEHFrameHeaderEncoding[:]) {
+		return 0, nil, fmt.Errorf("%w: unsupported CPython function-start table", errUnsupportedLayout)
+	}
+	count := uint64(file.ByteOrder.Uint32(data[ehFrameHeaderSize-ehFrameEncodedFieldSize : ehFrameHeaderSize]))
+	if count == 0 || count > uint64((len(data)-ehFrameHeaderSize)/ehFrameTableEntrySize) ||
```
pinoOgni: is `count > uint64((len(data)-ehFrameHeaderSize)/ehFrameTableEntrySize)` needed?

pkg/internal/cpython/runtime/probe_linux_amd64.go
```diff
@@ -74,14 +516,232 @@ func recognizedCollectorName(name string) bool {
...
+
+	// Some builds place the repeated-call pattern in a direct PyGC_Collect wrapper:
+	//
+	//	PyGC_Collect: call <wrapper>
+	//	<wrapper>:    call 278110  // callback
+	//	              call 1fbdd0  // gc_collect_main
+	//	              call 278110  // callback
+	//
+	// Inspect each local wrapper once, matching the single call level seen in CPython.
+	for _, callee := range rootCalls {
```
pinoOgni: very very nit: `rootCalls` is the ordered list of every direct call in `PyGC_Collect` and if it
calls the same local helper twice, the loop disassembles it twice. We can do:
```
for _, callee := range slices.Compact(slices.Sorted(slices.Values(rootCalls))) {
```

grcevski (approved):
LGTM!

MrAlias (approved):

PR #3241, opened 2026-08-28
Add IPv6 support to eBPF reverse DNS

loglapa:
## Summary

- inspect UDP DNS responses over IPv4 and IPv6, including bounded traversal of common IPv6 extension headers
- accept valid A and AAAA answers in the reverse-DNS cache while rejecting malformed address lengths
- constrain parsing and ring-buffer submission to the declared IP and UDP payload boundaries

Relates to grafana/beyla#1769. The reverse-DNS XDP implementation now lives in OBI and is consumed by Beyla through its pinned OBI dependency.

## Validation

- [x] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed. No documentation update is needed because this extends the existing reverse-DNS mode without changing configuration or telemetry schema.
- generated the rdns BPF objects for amd64 and arm64 with clang 18 and `-Wall -Werror`
- `go test ./pkg/internal/rdns/... ./pkg/internal/pipe/rdns/...`
- `go test -race ./pkg/internal/rdns/ebpf/xdp`
- `go vet ./pkg/internal/rdns/...`
- `clang-format-18 --dry-run --Werror bpf/rdns/rdns_xdp.c`
- `clang-tidy-18 bpf/rdns/rdns_xdp.c -- -target bpfel -D__TARGET_ARCH_arm64 -I bpf -std=gnu17 -O2`

## AI use disclosure

Codex assisted with codebase analysis, test drafting, and review. I reviewed the final patch and validation results in detail and take responsibility for the contribution.

### DIFF
--- a/bpf/rdns/rdns_xdp.c
+++ b/bpf/rdns/rdns_xdp.c
@@ -7,6 +7,7 @@
 #include <bpfcore/bpf_builtins.h>
 #include <bpfcore/bpf_endian.h>
 
+#include <common/protocol_defs.h>
 #include <logger/bpf_dbg.h>
 
 // Reverse DNS implementation by means of XDP packet inspection.
@@ -51,6 +52,17 @@ enum { RB_RECORD_LEN = 256 };                        // Maximum record length fo
 enum { DNS_PORT = 53 };                              // Standard DNS port
 enum { UDP_HDR_SIZE = sizeof(struct udphdr), DNS_HDR_SIZE = 12 }; // Header sizes
 
+// IPv6 next-header values used while walking the extension-header chain.
+enum ipv6_next_header {
+    IPV6_HOP_BY_HOP = 0,
+    IPV6_ROUTING = 43,
+    IPV6_FRAGMENT = 44,
+    IPV6_AUTHENTICATION = 51,
+    IPV6_DESTINATION_OPTIONS = 60,
+};
+
+enum { IPV4_FRAGMENT_MASK = 0x3fff };
+
 static __always_inline __u8 get_bit(__u8 word, __u8 offset) {
     return (word >> offset) & 0x1;
 }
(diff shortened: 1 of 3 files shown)


bpf/rdns/rdns_xdp.c
```diff
@@ -51,6 +52,17 @@ enum { RB_RECORD_LEN = 256 };                        // Maximum record length fo
 enum { DNS_PORT = 53 };                              // Standard DNS port
 enum { UDP_HDR_SIZE = sizeof(struct udphdr), DNS_HDR_SIZE = 12 }; // Header sizes
 
+// IPv6 next-header values used while walking the extension-header chain.
```
rafaelroquetto: nit: remove this comment, it does not add any meaningful context

bpf/rdns/rdns_xdp.c
```diff
@@ -76,40 +88,140 @@ static __always_inline void *ctx_xdp_data_end(struct xdp_md *ctx) {
...
+    case ETH_P_IPV6:
+        return ipv6_udp_header((void *)(eth + 1), data_end);
+    default:
+        return NULL;
+    }
 };
 
 // Validates and calculates the size of a DNS question section
-static __always_inline __u32 validate_qsection(struct xdp_md *ctx, const unsigned char *data) {
+static __always_inline __u32 validate_qsection(const unsigned char *data,
```
rafaelroquetto, line 193: why is this change required?

bpf/rdns/rdns_xdp.c
```diff
@@ -138,9 +251,8 @@ static __always_inline __u32 validate_qsection(struct xdp_md *ctx, const unsigne
 // Submits a DNS packet to the ring buffer for user space processing.
 // Uses a bounded loop with per-iteration packet bounds checks so the BPF verifier
 // can track each access, avoiding bpf_xdp_load_bytes which requires kernel 5.18+.
-static __always_inline void submit_dns_packet(struct xdp_md *ctx, const unsigned char *const data) {
-    const unsigned char *end = ctx_xdp_data_end(ctx);
-
+static __always_inline void submit_dns_packet(const unsigned char *const data,
```
rafaelroquetto, line 229: ditto

rafaelroquetto (changes requested):
There are a few changes that appear unrelated I don't fully understand why they are part of this PR.

Overall, the idea is correct, but this is refactoring too much of the original architecture with seemingly no good reason - you shouldn't have to touch the ipv4 code, instead, I'd expect the current `udp_header` function to be turned into `udp_header_ipv4` (and analogous `ip_header` -> `ip4_header`)

Then, I'd hope we have analogous functions, namely `udp_header_ipv6` and `ip6_header`.

Finally, your toplevel `udp_header` has the right idea, and this should be the main fork between chosing to go down ipv4 or ipv6 paths. Just pass the context downstream like we already do for ipv4.

I am not sure we need validate_udp_header - `parse_dns_response` and `dns_response_tracker` provide enough sanity checks already.

For the ipv6 header iteration, I think it can be simplified a bit as well - have a look at https://github.com/rafaelroquetto/opentelemetry-ebpf-instrumentation/blob/187d8078598384b59a8075b906bc43e3aa83d335/bpf/socktracer/ingress.c#L106 - in a nutshell, we can iterate up until we find the UDP header - no need to validate it - it's pretty similar to what you have

bpf/rdns/rdns_xdp.c
```diff
@@ -138,9 +251,8 @@ static __always_inline __u32 validate_qsection(struct xdp_md *ctx, const unsigne
...
+                                              const unsigned char *const end) {
```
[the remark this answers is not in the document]
loglapa: Fixed in 6ea62ec1. Ring-buffer records now prepend the actual DNS payload length in network byte order; Go validates it and slices `RawSample` to that length before parsing. A regression test puts a valid-looking A address in the stale tail after a truncated payload and verifies that it is rejected.

bpf/rdns/rdns_xdp.c
loglapa: Removed in 6ea62ec1.

bpf/rdns/rdns_xdp.c
loglapa: The explicit endpoint is required to use the boundary derived from `udp->len` rather than the XDP frame end, which can include bytes beyond the declared UDP payload. 6ea62ec1 also carries that exact payload length in the ring-buffer record so user space cannot parse the fixed-size stale tail.

bpf/rdns/rdns_xdp.c
loglapa: This accounts for both QTYPE and QCLASS: the DNS question suffix contains two 16-bit fields, while the previous code advanced and bounds-checked only one. Using four bytes keeps multi-question traversal aligned and verifies the complete question against the UDP payload boundary.

loglapa:
Updated in 24aa8a3a to follow the suggested architecture: the original IPv4 path is now only renamed to ip4_header/udp_header_ipv4 and otherwise left intact; the IPv6 path uses analogous ip6_header/udp_header_ipv6 helpers; the top-level udp_header is the sole EtherType fork and passes ctx downstream. I also removed validate_udp_header and moved the UDP payload boundary check into dns_response_tracker, which already owns packet-length validation. The IPv6 extension walk is now the smaller bounded header walk described in the review. Targeted reverse-DNS tests, the xdp race test, go vet, and diff/format checks pass locally.

bpf/rdns/rdns_xdp.c
rafaelroquetto: I meant the change to the params list

bpf/rdns/rdns_xdp.c
```diff
@@ -119,9 +206,10 @@ static __always_inline __u32 validate_qsection(struct xdp_md *ctx, const unsigne
         ++data;
 
         if (len == 0) {
-            size += sizeof(__u16); // account for QTYPE and QCLASS
+            const __u32 question_fields_size = 2 * sizeof(__u16); // QTYPE and QCLASS
```
rafaelroquetto, line 209: I see the bug, but this is a convoluted way to write sizeof(__u32)

bpf/rdns/rdns_xdp.c
rafaelroquetto: This is not how this works (copilot even elaborated down below)

bpf/rdns/rdns_xdp.c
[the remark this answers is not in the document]
rafaelroquetto: also this is not how any of this was designed - the validation is just a sanity check to avoid shipping false positives to userspace

rafaelroquetto:
hey @loglapa I appreciate the effort, but there are multiple issues with this approach - in particular, I don't believe there's a full understanding of how this code is meant to work, which is an expectation when we review PRs.

My recommendation is for you to go back and truly try to understand the spirit of this particular part of the code if you'd like to contribute to OBI. Throwing an AI agent at the problem and doing a shallow human pass on top of it is not enough - there's a reason we are explicit about having full understanding of the changes you are proposing in our AI policy - otherwise, PRs like this generate a lot of maintanence burden and review churn.

I will close this one, as it cannot be merged as is and the current approach is just incorrect. I'd be more than happy to review a saner PR on those lines - feel free to reach out on our slack channel if you have any questions regarding the implementation.

PR #3268, opened 2026-09-01
Make the HTTP body size metrics opt-in behind a new application_sizes feature

### LINKED ISSUE #2989: Allow disabling the HTTP body size metrics independently of the `application` feature

elliehmy:
## Summary

Closes #2989, with the `application_sizes` feature @grcevski suggested [there](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2989#issuecomment-5454111272). Review settled on keeping `application` as a bundle rather than moving the histograms out of it.

The goal is to make the four HTTP body size histograms (`http.{server,client}.{request,response}.body.size`) switchable on their own. The semantic conventions mark them [Opt-In](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestbodysize) while the RED duration histograms are Recommended, so a single feature covering both left no way to choose between them.

Three names in `metrics.features`:

| features | RED | body size |
| --- | --- | --- |
| `application` | on | on |
| `application_red` | on | off |
| `application_sizes` | — | on |

This mirrors how `stats` bundles `stats_tcp_rtt` and `stats_tcp_io`. `application` keeps meaning what it meant before, so **existing configurations and the defaults behave exactly as they do today.** This is not a breaking change.

`application_sizes` on its own is rejected at validation: the size histograms are emitted from the HTTP application metric pipeline, so alone they would publish nothing at all, and saying so at startup seemed better than silently exporting no telemetry. A defined per-service list replaces the global one instead of extending it, so the global mask and every defined per-service mask are validated independently.

## Config v2

Config v2 gets `capture.instrumentation.http.enabled.body_size_metrics`. `ProtocolEnablement` is shared by every protocol and these histograms only exist for HTTP, so HTTP gets its own enablement type that inlines the shared one rather than widening it.

The key is derived from the v1 `DefaultConfig`, so omitting it emits both exactly as v1 does, and `false` turns the sizes off. It converts both ways, and `obi config migrate` now carries `application_sizes` across instead of reporting it as outside the contract.

Unlike v1, `body_size_metrics: true` next to `metrics: false` is not treated as an error; the sizes simply follow `metrics`. The defaults overlay leaves that `true` indistinguishable from one that came from the default, so rejecting it would turn away a configuration over a value the user never wrote.

## Tests

Coverage includes the feature mapping and its backwards compatibility, the rejected combinations (global, per-service, and the v2 key), the v2 conversion both ways, the migration round trip, and each exporter emitting or omitting the four histograms.

The exporter tests do not stop at counting configured views and instruments: they drive real spans through the reporter so the recording paths run too. Under `application_red` the size instruments are never created, so a missing guard on the recording side is a nil dereference.

Existing tests that use the `FeatureApplicationRED` constant directly were updated to enable the sizes alongside it. Configurations that name `application` are unaffected.

## Validation

Neither of those documents enumerates metrics features. The generated config reference and the config v2 migration limitations table are updated. Ran `make lint`, `make verify` and `make check-config-v2-artifacts` locally.

## Generative AI disclosure

Claude Code was used to locate the exporter, config and test call sites, apply the edits, and draft this description. I reviewed and edited every change, ran the tests and lint locally, and take responsibility for the content.

---

The design changed during review. The final v1 and v2 configuration and defaults are laid out in [this comment](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3268#issuecomment-5630307792); the discussion above covers how it got there.


pkg/obi/config.go
```diff
@@ -845,6 +848,18 @@ func (c *Config) validate(context validationContext) error {
 	return nil
 }
 
+// validateAppSizeMetrics rejects application_sizes without the application RED metrics.
+// The body size histograms are emitted from the same HTTP metric pipeline, so on their own
+// they would produce no telemetry at all.
+func (c *Config) validateAppSizeMetrics() error {
+	features := c.JoinMetricsConfig().Features
```
MrAlias: `JoinMetricsConfig` only proves that the two flags appear somewhere in the configuration. A defined per-service feature list replaces the global list in `makeServiceAttrs`, so the newly accepted `application`-global/`application_sizes`-per-service case gives that service only `FeatureApplicationSizes`; `otelMetricsAccepted` then rejects all of its application metrics because `AppRED()` is false.

Could we validate the global feature mask and every defined per-service mask independently? The added test should expect this split configuration to fail unless both features are present in the same effective list.

devdocs/config/version-2.0/migration.md
```diff
@@ -393,7 +393,7 @@ include:
 | `ebpf.log_enricher.services` | Correlation filtering is not supported in v2. Do not broaden annotation silently; keep v1 or redesign the deployment. |
 | `attributes.sensitive_query_params` | No v2 field exists. Do not remove a redaction setting without an equivalent privacy review. |
 | `discovery.exclude_otel_instrumented_services_span_metrics` | No independent v2 selector exists for this legacy span-metrics exception. |
-| `metrics.features` values other than application RED, basic network flow, and the individual network-stat features | Span metrics, service graphs, application host/runtime metrics, inter-zone/network-packet variants, and eBPF metrics are not represented by current v2 enablement. |
+| `metrics.features` values other than application RED, basic network flow, and the individual network-stat features | Span metrics, the HTTP body size histograms, service graphs, application host/runtime metrics, inter-zone/network-packet variants, and eBPF metrics are not represented by current v2 enablement. |
```
MrAlias: Making the size histograms opt-in removes them from every default configuration, including v2, but v2 has no way to opt them back in: `applyV2MetricsEnablement` only sets `FeatureApplicationRED`, and migration rejects `application_sizes`. This also makes the PR's release-note instruction to add `application_sizes` unusable for v2 deployments.

Could we add a v2 representation for this feature? Issue #2989's production report specifically calls out that v2 already lacks the existing mitigations and asks for an obvious v2 home.

MrAlias (changes requested):
The exporter split is sound, but configuration validation admits a nonfunctional per-service combination and config v2 has no way to restore the now-disabled body-size metrics.

internal/config/convert/export_test.go
```diff
@@ -1104,6 +1131,12 @@ func fieldByYAMLName(value reflect.Value, name string) (reflect.Value, bool) {
 		if field.PkgPath != "" {
 			continue
 		}
+		if field.Anonymous && strings.Contains(field.Tag.Get("yaml"), ",inline") {
+			if inlined, ok := fieldByYAMLName(value.Field(i), name); ok {
+				return inlined, true
+			}
+			continue
+		}
```
elliehmy, line 1139: `fieldByYAMLName` resolves struct fields by YAML name but did not follow embedded `,inline` structs, which the YAML output flattens. 
Existing walks never had to cross one: they start below the embed, as in `value(t, doc.TracerProvider, ...)` where Go resolves the promoted field at compile time.
`["http", "enabled", "traces"]` is the first lookup that has to cross an embed by name, once `HTTPInstrumentation.Enabled` inlines `ProtocolEnablement`.

pkg/obi/config.go
elliehmy: Fixed in ec6e79e7. The global mask and every defined per-service mask are now validated independently, and the split-configuration test expects the rejection instead of accepting it.

devdocs/config/version-2.0/migration.md
elliehmy: Added in 5b9a5b2b as `capture.instrumentation.http.enabled.body_size_metrics`. 
It converts both ways, so obi config migrate now carries `application_sizes` across instead of reporting it as outside the contract.

marctc (commented):
I understand the split, but is really necessary to have `application_sizes` deactivated by default? this a big breaking change for many existing users.

devdocs/config/version-2.0/examples/default-values-reference.fragment.yaml
```diff
@@ -208,6 +208,8 @@ extensions:
           enabled:
             traces: true
             metrics: true
+            # Opt-In in the semantic conventions, so off unless requested.
+            body_size_metrics: false
```
mariomac: I agree with Marc, this should be enabled by default to avoid breaking changes.

mariomac:
I see a problem here. Setting application_sizes as a new entry for the features section would break current installations. I'd do:
- For config v1, if `application` feature is set, it enables both application and application_sizes metrics.
- If user wants to get application but not application_sizes metrics, force it to move to config v2, which offers more fine-grained selection.

elliehmy:
Thanks for the reviews.

Whether the default should be on or off is the question I left open in the description for a maintainer call. From a user's point of view I lean towards off, so let me lay out why.

@marctc @mariomac I understand the concern as `application` no longer implying the body size histograms: a user who leaves `features: [application]` untouched gets no config error, no startup failure and no warning, but four metrics silently disappear.

I think a release note covers that. OBI already has the mechanism — the [`Behavior changes` section of the v0.13.0 release notes](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/releases/tag/v0.13.0) carries two changes of the same kind:

- `service.name` / `service.namespace` are no longer default metric labels — #3168
- Go runtime metrics require enabling the `application_runtime` feature — #3227

So I would expect the following to be enough for users to notice:

- name the four metrics that stop being collected by default
- state that `application_sizes` has to be added to `metrics.features` to keep them
- for config v2, `capture.instrumentation.http.enabled.body_size_metrics: true`

That is my opinion, and the maintainers' call matters more — I will follow whichever direction you decide.

---

That said, @mariomac's suggestion to leave v1 undivided and offer the fine-grained selection only in config v2 has a limitation: it assumes users can move to v2, and today many cannot.

On current main ([`79ff8e85`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/commit/79ff8e856887e0d82007f93d4a7a4bfffce3e006)) the set of metric features that survive the v1-to-v2 migration is limited — [devdocs/config/version-2.0/migration.md](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/79ff8e856887e0d82007f93d4a7a4bfffce3e006/devdocs/config/version-2.0/migration.md#L396) says:

> Span metrics, service graphs, application host/runtime metrics, inter-zone/network-packet variants, and eBPF metrics are not represented by current v2 enablement.

We run OBI as the collector receiver with `features: application,application_span_otel`, so under that plan we could not turn the body size histograms off without giving up span metrics. The mitigation #2989 asked for would stay unavailable to v1 deployments.

---

So I also considered the following shape. It keeps existing installations intact while leaving v1 a way to opt out — `application` becomes a bundle, with the individual names offered alongside it:

| features | result |
| --- | --- |
| `application` | RED + body size — same as today |
| `application_red` | RED only |
| `application_sizes` | body size only |

`stats` already has this structure — `stats` is the bundle and `stats_tcp_rtt`, `stats_tcp_io` and friends are the individual names. The motivation is the same one the [code comment](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/79ff8e856887e0d82007f93d4a7a4bfffce3e006/pkg/export/feature.go#L56-L60) gives:

> If overhead is a concern, enable the lower-frequency metrics individually and opt into `stats_tcp_io` explicitly.

Taking the expensive item out of the bundle is the same kind of change as this one. The cost is one extra feature name, and existing `features: [application]` configurations are unaffected.

MrAlias (approved):

grcevski:
Hi @elliehmy, I like your proposal for:
```
application	RED + body size — same as today
application_red	RED only
application_sizes	body size only
```

Would that work for you? Is that something you can perhaps implement in this PR?

elliehmy:
@grcevski Yes, implemented. Here is where the configuration and the defaults ended up in v1 and v2.

## config v1 — `metrics.features`

Existing configurations keep working: `application` still enables both. The [three names](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/087e088f8dfdc3a258f5a51cbd60ae2a9856898d/pkg/export/feature.go#L79) map like this:

```yaml
metrics:
  features:
    - application
```

To turn off only the body sizes:

```yaml
metrics:
  features:
    - application_red
```

| features | RED | body size |
| --- | --- | --- |
| unset (default, `application`) | on | on |
| `application_red` | on | off |
| `application_red, application_sizes` | on | on |
| `application_sizes` alone | config error | |

`application_sizes` on its own is [rejected](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/087e088f8dfdc3a258f5a51cbd60ae2a9856898d/pkg/obi/config.go#L859). The size histograms are emitted from the HTTP application metric pipeline, so on their own they publish nothing at all, and saying so at startup seemed better than silently exporting no telemetry:

```
application_sizes needs the application RED metrics enabled in the same features list:
use 'application', or add 'application_red' next to 'application_sizes'
```

## config v2 — `capture.instrumentation.http.enabled`

Aligned with v1, through a [new HTTP-only key](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/087e088f8dfdc3a258f5a51cbd60ae2a9856898d/internal/config/schema/instrumentation.go#L62). Omit it and both are emitted, as before.

```yaml
extensions:
  obi:
    version: "2.0"
    capture:
      instrumentation:
        http:
          enabled:
            traces: true
            metrics: true
            body_size_metrics: false   # true when omitted
```

| setting | RED | body size |
| --- | --- | --- |
| key omitted (default) | on | on |
| `body_size_metrics: true` | on | on |
| `body_size_metrics: false` | on | off |
| `metrics: false` | HTTP off | off |

The v2 defaults are derived from the v1 `DefaultConfig`, so the unconfigured behavior of the two formats stays in sync on its own.

Unlike v1, v2 does not treat `body_size_metrics: true` next to `metrics: false` as an error; the sizes simply follow `metrics`. The defaults overlay leaves that `true` indistinguishable from one that came from the default, so rejecting it would turn away a configuration over a value the user never wrote.

## Tests

Tests that use the `FeatureApplicationRED` constant directly were updated to enable the sizes alongside it. Configurations that name `application` are unaffected.

grcevski:
Hi @elliehmy, can you please sync with main to pick up a fix for the failing test?

pkg/export/otel/metrics_test.go
```diff
@@ -206,6 +206,33 @@ type InstrTest struct {
 	extraColl            int
 }
 
+// The HTTP body size histograms are Opt-In in the semantic conventions, so
+// application_red must set up the RED views without them.
+func TestOtelMetricOptions_BodySizeFeature(t *testing.T) {
```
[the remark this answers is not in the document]
elliehmy: Added in `0023382f` — the OTel test now drives spans through the reporter instead of only counting views, so removing either recording guard panics the `application_red` case.

grcevski (approved):
LGTM!

PR #3269, opened 2026-09-01
Print unnamed cidrs as plain strings

pinoOgni:
## Summary

This PR prints `cidrs` how they are written in the configuration. In brief, `network.cidrs` and `stats.cidrs` accept either a plain CIDR string or a `{cidr, name}` map, but `log_config: yaml|json` always printed the map form, adding `name: ""` to every entry the user wrote as a plain string.

## Validation

### DIFF
--- a/pkg/internal/pipe/cidr/cidr.go
+++ b/pkg/internal/pipe/cidr/cidr.go
@@ -40,6 +40,19 @@ func (d Definition) Label() string {
 	return d.CIDR
 }
 
+// plainDefinition drops the Definition method set, so marshaling it does not
+// recurse back into MarshalYAML.
+type plainDefinition Definition
+
+// MarshalYAML emits back the shape the definition was written in: a plain CIDR
+// string when no name was given, a "cidr"/"name" mapping otherwise.
+func (d Definition) MarshalYAML() (any, error) {
+	if d.Name == "" {
+		return d.CIDR, nil
+	}
+	return plainDefinition(d), nil
+}
+
 // Definitions contains a list of CIDRs to be set as the "src.cidr" and "dst.cidr"
 // attribute as a function of the source and destination IP addresses.
 // It supports two YAML formats:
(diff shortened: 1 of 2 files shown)


pkg/internal/pipe/cidr/cidr.go
```diff
@@ -40,6 +40,19 @@ func (d Definition) Label() string {
...
 }
 
+// plainDefinition drops the Definition method set, so marshaling it does not
+// recurse back into MarshalYAML.
+type plainDefinition Definition
+
+// MarshalYAML emits back the shape the definition was written in: a plain CIDR
+// string when no name was given, a "cidr"/"name" mapping otherwise.
+func (d Definition) MarshalYAML() (any, error) {
+	if d.Name == "" {
```
MrAlias, line 50: A mapping with only `cidr` is already accepted (see `TestUnmarshalYAML_NamedCIDRs`), but it has an empty `Name` and is emitted as a scalar here. That means `log_config` still changes the shape for a valid mapping such as `- cidr: 172.16.0.0/12`.

Could we retain the scalar-versus-mapping form through both `LoadConfig` and `Config.Unmarshal`? A regression test should cover the emitted YAML and JSON from each configuration path.

MrAlias (changes requested):
This loses the configured representation for a supported nameless mapping; the loaders need to retain that distinction before this can fulfill the PR's stated behavior.

pkg/internal/pipe/cidr/cidr.go
pinoOgni: Hi @MrAlias, nice catch! I worked on the fix but it's touching many files and I'm not sure how to move forward, two ideas:

1. Change the PR title to something like "Print unnamed cidrs as plain strings", which is what the change actually does now and then push the fix in follow-up.
2. Push the fix now.

WDYT?

mariomac (approved):

pkg/internal/pipe/cidr/cidr.go
MrAlias: Yeah, option (1) is probably the way to go for now 👍

grcevski: I adjusted the PR title. I think we should be good to merge this.

grcevski (approved):
LGTM!

MrAlias (approved):

pkg/internal/pipe/cidr/cidr.go
pinoOgni: I didn't see the notification. Thanks!

PR #3257, opened 2026-08-31
tpinjector: skip HTTP injection on stale msg buffers

mmat11:
## Summary

Gate the HTTP-detect/inject path on `fill_msg_buffers` success. Also document that TCP options
context propagation writes an unknown TCP option (kind 25) that L7 proxies and/or load balancers drop or reset, and recommend "headers" where instrumented traffic crosses such intermediaries.

## Validation

- [x] I have read and followed the [contributing guidelines](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/CONTRIBUTING.md)
- [ ] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed.

PR #3270, opened 2026-09-01
tpinjector: fix truncated msg buffer read that drops large client spans

mifisignal:
## The problem

`fill_msg_buffers()` copies the head of an outgoing message into a per-CPU buffer
so the `tcp_sendmsg` kprobe can parse it. It bounded that copy with
`copy_bytes & k_msg_buffer_size_max_mask`.

The mask is 8191, and the length caps at 8192. `8192 & 8191` is zero, so a
message of 8192 bytes or more copies nothing at all. The function still records
`real_size` as 8192.

The kprobe then finds the entry, trusts that length, and hands the protocol
parsers 8192 bytes this call never wrote. Nothing clears the buffer between
messages, so the parsers read whatever the previous message on that CPU left
behind.

## Symptoms and impact

Wherever context propagation is enabled, every plaintext outgoing HTTP request of
8 KB or more loses its client span. The request itself succeeds, and the
receiving service records it normally, so the loss looks like a missing
service-graph edge rather than an error. Stale bytes can also make a request
match a different protocol, so a span may be misattributed rather than merely
absent. Connections already known to carry TLS are exempt, because the function
returns early on `is_ssl_connection()`.

Two consequences are worth naming.

`exclude_otel_instrumented_services` never fires for a service that exports OTLP
over HTTP. An export batch always exceeds 8 KB, so the `POST /v1/traces` client
span the check looks for is one of the spans this bug destroys. The service is
then instrumented twice, by OBI and by its own SDK.

Memory-BIO TLS correlation cannot form for a record of 8 KB or more.
`tls_prefix_try_bind()` receives the same length and matches ciphertext against
residue, which may itself be a real TLS record from another connection.

## The fix

Bound the copy with `bpf_clamp_umax()`, and report only the bytes actually
written.

Use `bpf_clamp_umax()` rather than `min()`. The deleted mask bounded the register
on every kernel, because `BPF_AND` with an immediate always sets `umax_value`.
`min()` does not. It compiles to a compare and a select, and clang may hoist the
register copy above the compare, so the length reaches `bpf_probe_read_kernel()`
with its range unrefined unless the kernel links scalar IDs — 5.10 for ALU64, 6.4
for u32. Built at `-mcpu=v3`, `min()` fails to load on 5.10 for all 24
`tpinjector` constant combinations, with `R2 unbounded memory access, use
'var &= const' or 'if (var < const)'`. `bpf_clamp_umax()` refines the register in
place through inline asm, as its thirty existing call sites rely on, and holds
down to the 5.8 floor in `pkg/obi/os.go`.

`bpf_probe_read_kernel()` is all-or-nothing, so `real_size` now reports what it
wrote, and a zero-byte copy publishes no entry at all. Since #3257 gates the
sk_msg program's HTTP detection and injection on this function's return value, a
failed read now skips injection rather than deciding on bytes the call never
wrote.

### Testing

`bpf/tests/bpf_msg_buffer_copy.c` covers the arithmetic. It exercises the new
`msg_buffer_copy()` helper, so it cannot run against `main` unchanged.
Substituting the old expression into the helper makes it report `expected 8192,
got 0` for a maximal message.

`TestSuite_LargeHTTPRequest` gains egress subtests at 700, 8191, 8192 and 8704
bytes. The existing egress coverage pads to about 5 KB and never crosses the
boundary. One subtest sends without a `traceparent` of its own and reads the
header back off the wire at the far end, so a fix that restored client spans by
suppressing injection would fail it. Both need a peer the request genuinely
leaves the process for, since OBI emits no client span for a request an
instrumented process sends to itself.

## Validation

### DIFF
--- a/bpf/common/msg_buffer.h
+++ b/bpf/common/msg_buffer.h
@@ -5,14 +5,12 @@
 
 #include <bpfcore/vmlinux.h>
 #include <bpfcore/bpf_helpers.h>
+#include <bpfcore/utils.h>
 
 #include <common/http_buf_size.h>
 #include <common/pin_internal.h>
 
-enum {
-    k_msg_buffer_size_max = 8192,
-    k_msg_buffer_size_max_mask = k_msg_buffer_size_max - 1,
-};
+enum { k_msg_buffer_size_max = 8192 };
 
 struct {
     __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
@@ -35,3 +33,30 @@ typedef struct msg_buffer {
     // will then be used as a guard in different execution contexts.
     u32 cpu_id;
 } msg_buffer_t;
+
+// Copies as much of the message at src as msg_buffer_mem holds into dst, and
+// returns the number of bytes written — which is what the tcp_sendmsg kprobe
+// takes as the extent of this message. dst must have room for
+// k_msg_buffer_size_max bytes, and that many bytes at src must be readable.
+//
+// The length is bounded with bpf_clamp_umax() rather than min(). Both clamp the
+// value, but only the asm form bounds the register the length is passed in: min()
+// compiles to a compare and a select whose copy clang may hoist above the
+// compare, leaving the verifier to recover the range through scalar-ID linking
+// that older kernels do not have. bpf_clamp_umax() refines the register in place,
+// so any later move carries the bound with it.
+//
+// Returns 0 when nothing was written. bpf_probe_read_kernel() is all-or-nothing
+// and dst is a per-CPU buffer that nothing clears between messages, so reporting
+// a length the read did not produce would hand the protocol parsers whatever the
+// previous message on this CPU left behind.
+static __always_inline u16 msg_buffer_copy(unsigned char *dst, const void *src, u32 msg_size) {
+    u32 len = msg_size;
+    bpf_clamp_umax(len, k_msg_buffer_size_max);
+
+    if (len == 0 || bpf_probe_read_kernel(dst, len, src) != 0) {
+        return 0;
+    }
+
+    return (u16)len;
+}
--- a/bpf/tpinjector/tpinjector.c
+++ b/bpf/tpinjector/tpinjector.c
@@ -675,16 +675,6 @@ static __always_inline bool fill_msg_buffers(struct sk_msg_md *msg,
         return false;
     }
 
-    msg_buffer_t msg_buf = {
-        .pos = 0,
-        .real_size = min(msg->size, k_msg_buffer_size_max),
-        .cpu_id = bpf_get_smp_processor_id(),
-    };
-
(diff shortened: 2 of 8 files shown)


MrAlias (approved):

MrAlias:
@mifisignal can you resolve the conflicts on this? It looks ready to merge otherwise.

bpf/common/msg_buffer.h
```diff
@@ -35,3 +33,30 @@ typedef struct msg_buffer {
...
+// value, but only the asm form bounds the register the length is passed in: min()
+// compiles to a compare and a select whose copy clang may hoist above the
+// compare, leaving the verifier to recover the range through scalar-ID linking
+// that older kernels do not have. bpf_clamp_umax() refines the register in place,
+// so any later move carries the bound with it.
+//
+// Returns 0 when nothing was written. bpf_probe_read_kernel() is all-or-nothing
+// and dst is a per-CPU buffer that nothing clears between messages, so reporting
+// a length the read did not produce would hand the protocol parsers whatever the
+// previous message on this CPU left behind.
```
rafaelroquetto, line 52: I particularly find this comment more confusing that the code itself (and it's larger than the code), would suggest you removed it (it's a simple function, the code is speaks for itself).

As a matter of fact, I don't particularly like this function: it has the ergonomics of `read()`, is called `copy`,  but behaves in a very unique way.

Instead, it's just simpler and more obvious to clamp the len inline in fill_msg_buffers

bpf/tpinjector/tpinjector.c
```diff
@@ -675,16 +675,6 @@ static __always_inline bool fill_msg_buffers(struct sk_msg_md *msg,
         return false;
     }
 
-    msg_buffer_t msg_buf = {
-        .pos = 0,
-        .real_size = min(msg->size, k_msg_buffer_size_max),
```
rafaelroquetto, line 680: why is this not correct?

bpf/tpinjector/tpinjector.c
```diff
@@ -675,16 +675,6 @@ static __always_inline bool fill_msg_buffers(struct sk_msg_md *msg,
...
-        .cpu_id = bpf_get_smp_processor_id(),
-    };
-
-    bpf_probe_read_kernel(msg_buf.fallback_buf, k_kprobes_http2_buf_size, msg->data);
```
rafaelroquetto, line 684: I think the bug is simply ignoring the return value here

rafaelroquetto (changes requested):
It turns out this has already been fixed by https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3298

rafaelroquetto:
Hey @mifisignal thanks for your contribution! The issue has already been fixed inhttps://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3298 so I am closing this one. I appreciate the effort tho!

PR #3271, opened 2026-09-01
Add `db.response.status_code` to DB client and server operation duration metrics

pinoOgni:
## Summary

This PR adds `db.response.status_code` as a default attribute on `db.client.operation.duration` and `db.server.operation.duration`. Value sources match the existing trace attribute: the protocol error code for redis/memcached/mongo/couchbase/aerospike; for SQL, the vendor error code when the protocol has one (MySQL, SQL Server), else the SQLSTATE (PostgreSQL). For Elasticsearch, semconv defines the attribute as the HTTP response code, reported whenever a response was received (also on success); elasticsearch spans carried no error code at all before, so this adds it on both the trace attribute and the metric label.

Moreover, it fixes a trace-side bug this work surfaced: failed `PostgreSQL` spans reported `db.response.status_code="0"`; the postgres protocol carries no vendor error code, so the `SQLSTATE` is now reported (e.g. 42P01), matching `error.type`.

A follow-up PR will align `error.type` with `db.response.status_code` for SQL, since that changes `error.type` values existing dashboards may key on.

## Validation


pkg/appolly/app/request/span.go
```diff
@@ -249,6 +249,16 @@ type SQLError struct {
...
 }
 
+// ResponseStatusCode returns the db.response.status_code value for the error:
+// the vendor error code when the protocol provides one (MySQL, SQL Server),
+// else the SQLSTATE (PostgreSQL, whose protocol carries no vendor code).
+func (e *SQLError) ResponseStatusCode() string {
+	if e.Code != 0 {
+		return strconv.Itoa(int(e.Code))
+	}
+	return e.SQLState
```
NimrodAvni78, line 259: i think SQLState takes precedense over SQLError no?

pkg/appolly/app/request/span_getters.go
```diff
@@ -597,6 +607,16 @@ func spanPromGetters(attrName attr.Name) attributes.Getter[*Span, string] {
...
 }
 
+func dbResponseStatusCode(span *Span) string {
+	if span.DBError.ErrorCode != "" {
+		return span.DBError.ErrorCode
+	}
+	if span.Status == 1 && span.SQLError != nil {
+		return span.SQLError.ResponseStatusCode()
+	}
+	return ""
```
NimrodAvni78, line 626: I believe for elastic it needs to be the HTTP status code
need to make sure if we set other DBError on all other DBs

pinoOgni: Nice catch! Moreover I found that elasticsearch spans never populate `DBError`, so the attribute was missing entirely, on traces too. Fixed.

pkg/appolly/app/request/span.go
pinoOgni: AFAIU no, because in the per-DBMS semconv pages the attribute is defined as the vendor error number for [MySQL](https://opentelemetry.io/docs/specs/semconv/db/mysql/)/[MariaDB](https://opentelemetry.io/docs/specs/semconv/db/mariadb/)/[SQL Server](https://opentelemetry.io/docs/specs/semconv/db/sql-server/) (the SQL Server page also says it doesn't report SQLSTATE; for SQL Server our SQLState field actually holds the TDS state byte). [PostgreSQL](https://opentelemetry.io/docs/specs/semconv/db/postgresql/) is the only one that wants SQLSTATE, and it's exactly the fallback case since its protocol carries no vendor code (Code is always 0).

See the `[2]` footnote for `db.response.status_code` on each semconv db page because the info in the tables are generic (IMO).

NimrodAvni78 (approved):

PR #3272, opened 2026-09-02
tpinjector: gate the HTTP detector on the msg buffer fill in handle_existing_tp_pid

mifisignal:
## The problem

`fill_msg_buffers()` returns early on `msg->size == 0` and on `is_ssl_connection()`,
before it invalidates the head of `msg_buffer_mem`. That buffer is a per-CPU array that
nothing clears between messages, so after an early return it still holds the previous
message on that CPU.

`handle_existing_tp_pid()` discards the return value and calls `protocol_detector()`
anyway. Nothing between those two points knows which message the buffer describes.

#3257 fixed this for `obi_packet_extender()`, which now gates on the same return value.
`handle_existing_tp_pid()` is the other caller and was not covered.

## The fix

`handle_existing_tp_pid()` now reads the buffer only through
`msg_buffer_holds_http_request()`, which returns `fill_msg_buffers(...) &&
protocol_detector(...)`. That is #3257's pattern applied to the second call site.

The change makes the program smaller: `obi_packet_extender` goes from 7,414 to 7,380
instructions.

Two commits. The first moves `fill_msg_buffers`, `protocol_detector` and `already_tracked`
verbatim into `bpf/tpinjector/msg_buffers.h`, alongside the existing `inject_policy.h`, so
a host test can reach them; the only line added to `tpinjector.c` is the include. I checked
that move is behaviour-neutral at the bytecode level rather than by eye — `llvm-objdump -d`
of the generated object before and after differs only in the file-name header line, 22,654
lines otherwise identical. The second commit adds the gate and the test.

### Testing

`bpf/tests/bpf_tpinjector_msg_buffers.c` drives `fill_msg_buffers()` and
`protocol_detector()` on the host with mock maps. It establishes the hazard before
asserting the fix: after a plaintext fill, an SSL connection's fill returns false,
publishes no `msg_buffers` entry, leaves the previous request byte-for-byte resident, and
`protocol_detector()` still returns 1. It then asserts that the guarded reader refuses that,
refuses a zero-length send, accepts a genuine request, and rejects a response.

## Validation

### DIFF
--- a/bpf/tests/bpf_tpinjector_msg_buffers.c
+++ b/bpf/tests/bpf_tpinjector_msg_buffers.c
@@ -1,0 +1,362 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// msg_buffer_mem is per CPU and nothing clears it between messages, so after
+// fill_msg_buffers bails it still holds whichever message that CPU filled it
+// from last. protocol_detector reads that buffer, so its answer describes this
+// message only when the fill that precedes it succeeded.
+//
+// The bail that matters is the SSL one: the message on the wire is a TLS
+// record, the resident message is a plaintext HTTP request from another
+// connection, and a caller that acts on the match splices a 'Traceparent:'
+// header into the middle of the record. test_stale_buffer_still_matches_http
+// establishes that the match happens; test_guarded_read_refuses_a_stale_buffer
+// is the regression - it fails if the reader stops being gated on the fill.
+
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <bpfcore/bpf_helpers.h>
+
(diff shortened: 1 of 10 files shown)


MrAlias (approved):

MrAlias:
@mifisignal can you resolve the conflicts on this? It looks ready to merge otherwise.

rafaelroquetto (changes requested):
I believe this has already been fixed (which is why you are getting a conflict).

The fix was done in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3304

rafaelroquetto:
Hey @mifisignal again, thanks for raising this. Unfortunately (or fortunately?) this has already been fixed by https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3304 so I will go ahead and close this.

PR #3296, opened 2026-09-03
fix(http2): capture every stream on a multiplexed connection

### LINKED ISSUE #3291: HTTP/2 multiplexing: spans collapse to 0% for connections with >1 concurrent stream
## Summary

On an HTTP/2 connection carrying **more than one concurrent stream** (multiplexing), the generic (non-Go) HTTP/2 frame parser produces spans for only about one stream per connection. In practice span generation collapses to **0%** as soon as the peer uses `m ≥ 2` concurrent streams, while single-stream HTTP/2 (`m = 1`) produces spans at ~100%.

- HTTP/1.x is unaffected.
- The Go HTTP/2 path (Go `net/http` uprobes) is **not** affected — it produces spans for multiplexed streams at ~100% (data below).
- Measured on the server side (server spans); any multiplexing HTTP/2 workload on the generic path — gRPC, HTTP/2 connection reuse, load balancers, etc. — produces almost no spans.

## Environment

- OBI **main** (`e0c90103`) and **v0.12.2** — both reproduce identically.
- Kubernetes, amd64, kernel 6.x, OpenSSL 3.x.
- Reproduced over TLS (`libssl` path); the fault is in HTTP/2 frame parsing, not TLS.

## Reproduction

Two HTTP/2-over-TLS test servers exposing `GET /red?status=<code>&size=<bytes>` (echoes an arbitrary status code and body size) and `/stats` (the server's own request counter, as ground truth):

- **non-Go**: a server that dynamically links system `libssl` (so OBI's TLS uprobes attach) — e.g. Rust + the `openssl` crate.
- **Go**: a stdlib `net/http` server over TLS with ALPN `h2`.

Drive each from one connection with a varying number of concurrent streams, 500 requests per run, each run owning a unique status code so its captured spans are attributable:

```
h2load -n 500 -c 1 -m 1  'https://<server>:8443/red?status=<code>&size=16'
h2load -n 500 -c 1 -m 2  'https://<server>:8443/red?status=<code>&size=16'
h2load -n 500 -c 1 -m 10 'https://<server>:8443/red?status=<code>&size=16'
```

Compare the number of OBI-emitted spans carrying each status code against the server's `/stats` (500 per run).

## Result

OBI captured / 500 sent:

| concurrent streams (`-m`) | non-Go server (`libssl` path) | Go server (`net/http` path) |
|---|---|---|
| 1  | 500 (100%) | 500 (100%) |
| 2  | **0 (0%)** | 500 (100%) |
| 10 | **0 (0%)** | 500 (100%) |

Expected: the non-Go path at `m = 2` and `m = 10` should capture at approximately the same rate as `m = 1`, matching the Go path.

wangzlei:
## Summary

On the generic (non-Go) HTTP/2 path, spans and RED metrics collapsed to ~0% once a connection carried more than one concurrent stream; single-stream HTTP/2, HTTP/1.x, and the Go `net/http` path were unaffected. Fixes #3291.

## Approach

Make the generic frame scanner walk an entire multiplexed TLS buffer in stream order and flush each stream as it is seen, instead of stopping at the first one:

- The per-frame HEADERS/DATA handlers are tail-called programs that do not return, so after one runs the scan resumes by tail-calling back in at the offset past that frame — every frame in the buffer is visited, within the existing iteration budget.
- Each exchange is emitted at its response HEADERS frame and its per-stream state is cleared on close, so concurrent streams no longer collide on the single saved-header slot (previously a later stream's HEADERS overwrote an earlier one and only the last emitted).
- Feeding the userspace HPACK decoder every response block in wire order keeps its dynamic table synced, so the conservative "mark unreliable on a following HEADERS frame" is dropped. Genuine overflow past the budget is still marked unreliable (end-of-scan poison plus the decoder's bad-index freezing), so a desync degrades to status 0, never a wrong value.

## Limitation

The ceiling is ~6 concurrent streams grouped in one TLS write buffer, set by the frame scanner's per-buffer budget (~12 frames, roughly six HEADERS+DATA stream pairs). Past it the degradation is two-stage. First, streams beyond the budget in a buffer are never scanned, so their spans are simply missing — at 6 streams everything that is captured still decodes with the correct status. Then, because a skipped response HEADERS block is never fed to the userspace HPACK decoder, that connection's response dynamic table falls behind the server's encoder, and later captured streams mis-resolve their dynamic `:status` to 0 (dominant by 8 streams).

Raising it has two directions. One stays in BPF and just pushes the per-buffer limit further — a larger frame-scan iteration budget (higher `k_loop_count`) and/or fewer tail-call hops per stream (merging the server registration stages). The other moves the work out of BPF entirely: ship the raw buffer to userspace over the ring buffer and do frame scanning and HPACK decoding there, which removes the in-kernel budget as the bottleneck at the cost of more data crossing to userspace. Both are out of scope here and can be discussed as follow-ups once this fix lands. This PR restores the common 2–5 stream cases; the cap is documented in SUPPORT_MATRIX / features.

## Validation

- New Go unit test: multiplexed response `:status` stays in sync across streams.
- New integration test `TestSuite_HTTP2Multiplexing`: a hand-written h2c client and server deterministically pack several concurrent streams into one socket write (the client sends all request HEADERS in one `send()`; the server withholds responses until the whole batch arrives, then writes them in one `send()`). Each request path carries a per-burst id and stream index, so spans are grouped per burst and the test requires — on both the server and client side — that whole bursts are captured with none captured partially. It fails without this fix.
- BPF verifier suite passes (kernel 5.10); DaemonSet loads clean on kernel 6.1.
- Validated live against a `libssl` HTTP/2 server driven with `h2load -n 500 -c 1 -m <N>`, counting server spans that carry the correct response status (each run uses a distinct status code). Before this fix, everything above one concurrent stream collapsed to 0%.

  | concurrent streams (`-m`) | correct-status spans / 500 |
  |---|---|
  | 1 | 500 (100%) |
  | 2 | 500 (100%) |
  | 3 | 499 (99.8%) |
  | 4 | 498 (99.6%) |
  | 5 | 499 (99.8%) |
  | 6 | 414 (83%) |
  | 7 | 303 (61%) |
  | 8 | 0 (0%) |

  Clean through 5 concurrent streams. At 6 the losses are missing spans (captured streams stay 100% correct); past that, HPACK desync from the skipped blocks sets in and it collapses by 8.

### Why a new integration test

The existing gRPC multiplexing integration tests do not actually pack multiple streams into a single read/write buffer — in practice almost every captured buffer carries a single stream — so they never reached the density that triggers this bug and would pass with or without the fix. The new test forces the packing deterministically on both sides, giving real coverage for the multiplexed path.

Developed with AI assistance; every change reviewed and understood by me (per AI-POLICY.md).

- [x] Updated features documentation and support matrix


mmat11:
grpc integration tests are failing

edit: we also currently have an integration test which is supposed to test multiplexed grpc connections (also non-Go), why was that one not able to spot this issue? can we fix it if it's broken?

wangzlei:
The failing integration shard looks unrelated to this PR — a different, non-HTTP/2 test fails on each run, and the logs show base-image rate-limit warnings, so it looks like CI flakiness.

A note on current integration coverage: the existing multiplexed/gRPC tests pass, but may not actually be triggering multiplexing. Stream loss only happens when frames from several concurrent streams are packed into a single read/write buffer, which needs higher per-connection stream density than the current tests drive (a few concurrent streams) — which is why it has gone unnoticed. If this theory holds, we should redesign the integration test for gRPC multiplexing.

wangzlei:
@mmat11 The existing gRPC multiplexing test doesn't actually pack multiple streams into a single read/write buffer — in practice almost every captured buffer carries a single stream — so it never hits the condition that breaks here and passes with or without the fix.

I'd suggest landing the HTTP/2 multiplexing fix in this PR first, then rewriting the gRPC multiplexing integration test along the lines of the h2mux test added here (deterministic packing on both sides) as a follow-up.

mmat11:
> A note on current integration coverage: the existing multiplexed/gRPC tests pass, but may not actually be triggering multiplexing. Stream loss only happens when frames from several concurrent streams are packed into a single read/write buffer, which needs higher per-connection stream density than the current tests drive (a few concurrent streams) — which is why it has gone unnoticed. If this theory holds, we should redesign the integration test for gRPC multiplexing.

thanks a lot for digging into this 🙏🏻 , is this something that should be done in this PR as well?

grcevski:
I believe the copilot comments are valid and need to be addressed.

bpf/generictracer/protocol_http2.h
```diff
@@ -483,6 +488,8 @@ handle_headers_frame(void *ctx, grpc_frames_ctx_t *g_ctx, const frame_header_t *
             if (h2_next_frame_changes_hpack(g_ctx, frame)) {
                 http2_poison_hpack(&g_ctx->stream, 0, h2_hpack_req_unreliable);
             }
```
MrAlias, line 490: This look-ahead still fires for the multiplexed request buffer added here: every request except the last is followed by another HEADERS frame, so the first stream sets `req_hpack_poisoned` even though the resumed scan emits every block in wire order. `readHTTP2HeaderEvent` propagates that flag to `MarkUnreliable`, after which dynamic request-table indexes resolve as `<BAD INDEX>` and request metadata can degrade despite no frame being skipped.

Could we drop this look-ahead poison for request headers too, and poison only when scanning actually stops before the buffer is consumed? A request-side dynamic-table regression test alongside `TestHTTP2MultiplexedResponseStatusStaysInSync` would cover this path.

MrAlias (commented):
The response-side scan now preserves wire order, but the request-side HPACK state still becomes unreliable under the same packed-HEADERS pattern.

bpf/generictracer/types/grpc_frames_ctx.h
```diff
@@ -18,6 +18,10 @@ typedef struct grpc_frames_ctx {
     int pos; //FIXME should be size_t equivalent
     int saved_buf_pos;
     u32 saved_stream_id;
+    // byte offset past the frame a handler tail-called on, so resume_frame_scan
+    // can continue the scan; _pad keeps 8-byte alignment (-Werror -Wpadded)
```
rafaelroquetto, line 22: is resume_pos pointing to the same buffer as saved_buf_pos? if so, I'd group them together, normalise the names, and do without this comment (which is wrong btw, there's no _pad.

bpf/generictracer/protocol_http2.h
```diff
@@ -462,10 +462,15 @@ handle_headers_frame(void *ctx, grpc_frames_ctx_t *g_ctx, const frame_header_t *
...
-            if (h2_next_frame_changes_hpack(g_ctx, frame)) {
-                http2_poison_hpack(&g_ctx->stream, &g_ctx->prev_info, poison);
-            }
+        // Emit at the response HEADERS frame instead of deferring to the DATA
+        // end frame: a later stream's HEADERS would overwrite the single
+        // saved_buf_pos slot first, dropping every multiplexed stream but the
+        // last. No poison on a following HEADERS frame either -- the scan now
+        // walks the buffer in wire order, keeping the response HPACK decoder
+        // synced; overflow past the iteration budget is still poisoned at the
+        // end of the frames loop.
```
rafaelroquetto, line 471: I don't undertand how this comment relates to the if that follows, would you mind explaining this in your own words (please no AI)?

bpf/generictracer/protocol_http2.h
```diff
@@ -492,6 +499,20 @@ handle_headers_frame(void *ctx, grpc_frames_ctx_t *g_ctx, const frame_header_t *
     return 1;
 }
 
+// mirrors k_max_loop_iterations * k_loop_count in the frames program below
+enum { k_h2_frame_scan_max_iterations = 12 };
```
rafaelroquetto, line 503: why not declare it at the usage site as `const max_iterations = k_max_loop_iterations * k_loop_count` ?

bpf/generictracer/protocol_http2.h
```diff
@@ -492,6 +499,20 @@ handle_headers_frame(void *ctx, grpc_frames_ctx_t *g_ctx, const frame_header_t *
...
+
+// Hands control back to the frame scan after a per-frame handler tail-called
+// away, so the rest of the streams in a multiplexed buffer are still scanned.
+// The caller must set g_ctx->resume_pos to the offset past its frame.
+static __always_inline void resume_frame_scan(void *ctx, grpc_frames_ctx_t *g_ctx) {
+    g_ctx->pos = g_ctx->resume_pos;
```
rafaelroquetto, line 509: is this correct? should this be inside the if?

bpf/generictracer/protocol_http2.h
```diff
@@ -530,6 +551,8 @@ int GUARDED_PROG(obi_protocol_http2_grpc_handle_start_frame, void *, ctx) {
     http2_grpc_start(
         ctx, &g_ctx->stream, offset, args->bytes_len, args->direction, args->ssl, args->orig_dport);
 
+    // reached only on the client path (server path tail-calls away above)
```
rafaelroquetto, line 554: no need for this comment

bpf/generictracer/protocol_http2.h
```diff
@@ -743,6 +766,8 @@ int GUARDED_PROG(obi_protocol_http2_grpc_handle_start_frame_server_commit, void
     http2_grpc_start_finalize_server(
         &g_ctx->stream, h2g_info, tp_p, found_tp, g_ctx->args.ssl, g_ctx->args.orig_dport);
 
+    // server request registered; scan the rest of the multiplexed streams
```
rafaelroquetto, line 769: ditto

bpf/generictracer/protocol_http2.h
```diff
@@ -768,6 +793,11 @@ int GUARDED_PROG(obi_protocol_http2_grpc_handle_end_frame, void *, ctx) {
         http2_grpc_end(&g_ctx->stream, &g_ctx->prev_info, offset);
 
         bpf_map_delete_elem(&active_ssl_connections, &g_ctx->args.pid_conn);
+
+        // drop the per-stream latch so the resumed scan looks up the next
```
rafaelroquetto, line 797: what do you mean with latch in this context? is this better served behind a well named helper function rather than a comment?

mmat11:
hi @wangzlei, thanks a lot for digging into this and for the h2mux test idea 🙏🏻! Since this PR seems to have stalled a bit, and to help the fix get landed, I opened #3544, which builds on your work, addresses the review comments and fixes the tests. Your review there would be much appreciated, closing this one in favor of it

PR #3298, opened 2026-09-03
Fix out-of-bounds read in fill_msg_buffers when msg pull fails

### LINKED ISSUE #3278: fill_msg_buffers can copy adjacent kernel memory into msg_buffer_mem
### What happens

`fill_msg_buffers()` copies up to 8192 bytes from `msg->data`. That covers only the mapped
extent of the `sk_msg`, which may be a fraction of the message. When the read runs past
`msg->data_end` it **succeeds**: it returns 0 and yields whatever memory follows.

Those bytes go into `msg_buffer_mem`, `real_size` reports the full length, and the
`tcp_sendmsg` kprobe hands them to the protocol parsers. The generic TCP path copies the
payload into the span record (`protocol_tcp.h`, `bpf_probe_read(req->buf, bytes_len, u_buf)`),
so unrelated kernel memory can reach a trace backend as a span attribute.

### Measured

A sockmap harness recorded the mapped window, the read result, and the bytes, for one
65536-byte write on kernel 7.0:

```
msg->size=65536  window_before=7492  read_len=8192  read_ret=0
  bytes at offset 7492:
    pre-pull  = 9f d6 04 00 4c 00 42 01 63 61 63 38 62 36 63 37   ("cac8b6c7")
    post-pull = d5 d6 d7 d8 d9 da db dc dd de df e0 e1 e2 e3 e4
    message   = d5 d6 d7 d8 d9 da db dc dd de df e0 e1 e2 e3 e4
```

The read ran 700 bytes past the window, returned 0, and produced bytes belonging to
something else. Three runs gave three different payloads, one of them the ASCII string
`"ult: (SETEQstore"`.

Because the read succeeds, its return value reports nothing about the problem.

### How it is reached

Each of the three `fill_msg_buffers()` call sites is preceded by
`bpf_msg_pull_data(msg, 0, msg->size, 0)` whose result is discarded. On failure that helper
leaves `msg->data` and `msg->data_end` as they were, so the window stays narrow and the
copy runs past it.

The failure needs `-ENOMEM` from `alloc_pages(__GFP_NOWARN | GFP_ATOMIC | __GFP_COMP,
get_order(copy))`. Reachability is therefore inferred: the allocation succeeded on every
attempt we made, and we stopped there to keep a shared host stable. See #3277 for the
allocation size, which is the reachability story for this defect.

### Suggested fix

Bound the copy by the mapped extent:

```c
const u64 readable = (u64)msg->data_end - (u64)msg->data;
```

When the caller's pull succeeded, `readable` equals `msg->size`, so the normal path stays
byte-for-byte the same and the bound applies only in the broken case. Use
`bpf_clamp_umax()` for the final clamp: it refines the register in place, which the
verifier can follow on every supported kernel.

darkraider01:
Found this while going through the sk_msg path , `fill_msg_buffers()` just assumes the earlier `bpf_msg_pull_data()` pull actually worked, but nothing checks it. If it fails, `msg->data`/`data_end` stay narrow, and the function still blindly copies up to 8KB from `msg->data`, so it can pull in random adjacent kernel memory instead of the actual message. That memory can end up going out as span data.

Fixed it by bounding every read to the actual mapped `[data, data_end)` window instead of trusting `msg->size`. That window is always a sub-range of the message, so it's sufficient on its own ,no separate bounds check needed. Used `bpf_clamp_umax` (not `min()`) since a plain C compare loses its tracked bound across a register copy on older verifiers (kernel 5.8 rejected the first version of this fix). Along the way this also fixed a real pre-existing bug: a `& (k_msg_buffer_size_max - 1)` mask on the copy size masked exactly-8192-byte sends down to a 0-byte copy, so the kprobe would read stale bytes from a previous message.

Kept it inside `fill_msg_buffers()` itself since all three call sites go through it, so nothing else needed changing  and there's already a fallback path on the kprobe side for when this buffer isn't populated, so bailing here doesn't break anything.

Fixes #3278.

### DIFF
--- a/bpf/tpinjector/tpinjector.c
+++ b/bpf/tpinjector/tpinjector.c
@@ -681,9 +681,15 @@ static __always_inline bool fill_msg_buffers(struct sk_msg_md *msg,
         .cpu_id = bpf_get_smp_processor_id(),
     };
 
-    bpf_probe_read_kernel(msg_buf.fallback_buf, k_kprobes_http2_buf_size, msg->data);
-
     const u16 copy_bytes = max(msg_buf.real_size, k_kprobes_http2_buf_size);
+
+    // A failed bpf_msg_pull_data() leaves msg->data/data_end short of copy_bytes.
+    // Reading past data_end would copy whatever kernel memory follows the message.
+    if (!msg->data || msg->data + copy_bytes > msg->data_end) {
+        return false;
+    }
+
+    bpf_probe_read_kernel(msg_buf.fallback_buf, k_kprobes_http2_buf_size, msg->data);
 
     unsigned char **msg_ptr = bpf_map_lookup_elem(&msg_buffer_mem, &(u32){0});
 
--- a/bpf/tpinjector/tpinjector.c
+++ b/bpf/tpinjector/tpinjector.c
@@ -675,21 +675,25 @@ static __always_inline bool fill_msg_buffers(struct sk_msg_md *msg,
         return false;
     }
 
-    msg_buffer_t msg_buf = {
-        .pos = 0,
-        .real_size = min(msg->size, k_msg_buffer_size_max),
-        .cpu_id = bpf_get_smp_processor_id(),
-    };
-
-    const u16 copy_bytes = max(msg_buf.real_size, k_kprobes_http2_buf_size);
-
-    // A failed bpf_msg_pull_data() leaves msg->data/data_end short of copy_bytes.
-    // Reading past data_end would copy whatever kernel memory follows the message.
-    if (!msg->data || msg->data + copy_bytes > msg->data_end) {
+    if (!msg->data || msg->data >= msg->data_end) {
         return false;
     }
 
-    bpf_probe_read_kernel(msg_buf.fallback_buf, k_kprobes_http2_buf_size, msg->data);
+    // A failed bpf_msg_pull_data() can leave msg->data/data_end short of msg->size.
+    // Every read below is bounded by this mapped window instead of msg->size alone,
+    // so a short message never pulls in whatever kernel memory follows it.
+    const u32 mapped = (unsigned char *)msg->data_end - (unsigned char *)msg->data;
+    const u16 window = min(mapped, (u32)k_msg_buffer_size_max);
+
+    msg_buffer_t msg_buf = {
+        .pos = 0,
+        .real_size = min(min(msg->size, k_msg_buffer_size_max), (u32)window),
+        .cpu_id = bpf_get_smp_processor_id(),
+    };
+
+    const u16 copy_bytes = min(max(msg_buf.real_size, k_kprobes_http2_buf_size), window);
+
+    bpf_probe_read_kernel(msg_buf.fallback_buf, min(window, k_kprobes_http2_buf_size), msg->data);
 
     unsigned char **msg_ptr = bpf_map_lookup_elem(&msg_buffer_mem, &(u32){0});
 
--- a/bpf/tpinjector/tpinjector.c
+++ b/bpf/tpinjector/tpinjector.c
@@ -5,6 +5,7 @@
 #include <bpfcore/bpf_builtins.h>
 #include <bpfcore/bpf_helpers.h>
 #include <bpfcore/bpf_endian.h>
+#include <bpfcore/utils.h>
 
 #include <common/algorithm.h>
 #include <common/connection_info.h>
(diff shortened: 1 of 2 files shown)


bpf/tpinjector/tpinjector.c
```diff
@@ -681,10 +681,16 @@ static __always_inline bool fill_msg_buffers(struct sk_msg_md *msg,
         .cpu_id = bpf_get_smp_processor_id(),
     };
 
-    bpf_probe_read_kernel(msg_buf.fallback_buf, k_kprobes_http2_buf_size, msg->data);
-
     const u16 copy_bytes = max(msg_buf.real_size, k_kprobes_http2_buf_size);
 
+    // A failed bpf_msg_pull_data() leaves msg->data/data_end short of copy_bytes.
+    // Reading past data_end would copy whatever kernel memory follows the message.
+    if (!msg->data || msg->data + copy_bytes > msg->data_end) {
```
mmat11: `copy_bytes` is 256 even if msg size is 50, so the check will fail for any payload size <= `k_kprobes_http2_buf_size`

darkraider01: thanks for the review, copy_bytes always floored at 256 regardless of the actual message size, so this rejected every request `<=256` bytes (basically all of them, hence the CI failures).

Reworked it to bound every read by the actual mapped window (`msg->data_end - msg->data`) instead of gating on that inflated floor: `window`, `real_size` and `copy_bytes` are all clamped to what's genuinely available, so a short message just gets read in full instead of bailing, and a short *window* (the original failure case) still gets clamped safely instead of overread.

skl (approved):
lgtm

bpf/tpinjector/tpinjector.c
```diff
@@ -675,15 +675,25 @@ static __always_inline bool fill_msg_buffers(struct sk_msg_md *msg,
...
 
+    if (!msg->data || msg->data >= msg->data_end) {
+        return false;
+    }
+
+    // A failed bpf_msg_pull_data() can leave msg->data/data_end short of msg->size.
+    // Every read below is bounded by this mapped window instead of msg->size alone,
+    // so a short message never pulls in whatever kernel memory follows it.
+    const u32 mapped = (unsigned char *)msg->data_end - (unsigned char *)msg->data;
+    const u16 window = min(mapped, (u32)k_msg_buffer_size_max);
```
pinoOgni: `data_end - data` can never exceed the value of `msg->size` because it is a sub-range of the message, so doing `min(msg->size, ...)` and `max(..., 256)` does not change the result: `real_size`, `copy_bytes` and `window` are always the same value. We can do:
```
u32 window = (unsigned char *)msg->data_end - (unsigned char *)msg->data;
bpf_clamp_umax(window, k_msg_buffer_size_max); 
```
and `.real_size = window` in the struct. Moreover, `bpf_clamp_umax` is suggested in #3278

MrAlias (approved):

bpf/tpinjector/tpinjector.c
pinoOgni: Not added in this PR, but we can fix it now. Basically  `copy_bytes & k_msg_buffer_size_max_mask` can be `8192 & 8191 == 0` for any send of `8192` bytes or more, so nothing is copied and the kprobe gets the previous message's bytes with `real_size = 8192`. With window already clamped to `k_msg_buffer_size_max` the mask is not needed and we can do:
```
bpf_probe_read_kernel(msg_ptr, window, msg->data);
```
and delete `k_msg_buffer_size_max_mask`.

bpf/tpinjector/tpinjector.c
pinoOgni: Also, I did a test on unpatched 5.8 (Lima, `TestBPFVerifierWithConstants/tpinjector`) this PR adds a new rejection in `obi_packet_extender` that main does not have (there are others we should fix): invalid stack type R1 off=-344 access_size=8191 at the fallback read. The compiler copies `window` to `r2` before the `min()` compare and the `5.8` verifier does not carry the bound to the copy.

bpf/tpinjector/tpinjector.c
```diff
@@ -675,15 +675,25 @@ static __always_inline bool fill_msg_buffers(struct sk_msg_md *msg,
...
-        .real_size = min(msg->size, k_msg_buffer_size_max),
+        .real_size = min(min(msg->size, k_msg_buffer_size_max), (u32)window),
         .cpu_id = bpf_get_smp_processor_id(),
     };
 
-    bpf_probe_read_kernel(msg_buf.fallback_buf, k_kprobes_http2_buf_size, msg->data);
+    const u16 copy_bytes = min(max(msg_buf.real_size, k_kprobes_http2_buf_size), window);
 
-    const u16 copy_bytes = max(msg_buf.real_size, k_kprobes_http2_buf_size);
+    bpf_probe_read_kernel(msg_buf.fallback_buf, min(window, k_kprobes_http2_buf_size), msg->data);
```
pinoOgni: The fixed 256-byte reads are gone, so sends under 256 bytes no longer read past the message. We should update the pr description.

pinoOgni (changes requested):

bpf/tpinjector/tpinjector.c
darkraider01: You're right, mapped can never exceed msg->size , it's a sub-range by construction. Simplified real_size/copy_bytes down to just window, and switched to bpf_clamp_umax in place instead of min()/max(), which also fixes the 5.8 rejection you found below (thanks for actually testing that on Lima, wouldn't have caught it otherwise).

bpf/tpinjector/tpinjector.c
darkraider01: Good catch, that's a real bug on its own (8192 & 8191 == 0). Now that window is already clamped to k_msg_buffer_size_max via bpf_clamp_umax, the mask is redundant and wrong ,dropped it, and deleted k_msg_buffer_size_max_mask entirely since this was its only use.

bpf/tpinjector/tpinjector.c
darkraider01: Thanks for actually running this against 5.8, switched min()/max() (a C ternary, apparently losing the tracked bound across that register copy on older verifiers) for bpf_clamp_umax, which is the established pattern for exactly this in the codebase already (dns.h, protocol_tcp.h, etc). Should carry the bound directly on the register used in bpf_probe_read_kernel now.would appreciate a re-test on your 5.8 VM if you get the chance, I don't have one handy locally.

bpf/tpinjector/tpinjector.c
darkraider01: Good point, updated the description to reflect the actual fix (bound reads to the mapped window rather than msg->size/an inflated floor) instead of the original narrower framing.

bpf/tpinjector/tpinjector.c
```diff
@@ -675,15 +676,27 @@ static __always_inline bool fill_msg_buffers(struct sk_msg_md *msg,
         return false;
     }
 
+    if (!msg->data || msg->data >= msg->data_end) {
```
pinoOgni: nit: can we add a comment here?

pinoOgni (approved):
LGTM! Thanks for improving it!

mmat11 (approved):
thanks!

PR #3299, opened 2026-09-03
Fix HTTP/2 and gRPC spans losing their peer by falling back to :authority

### LINKED ISSUE #3283: HTTP/2 and gRPC spans leave :authority unused and lose the peer
### What happens

When a connection cannot be resolved, an HTTP/1.x client span recovers its peer from the
`Host:` header. `http_transform.go` parses the request buffer and fills in the address,
plus the port when the header carries one.

HTTP/2 carries the same information in the `:authority` pseudo-header, and OBI already
decodes it. The transform leaves it unused: `http2grpc_transform.go` takes host and port
from `ConnInfo` and sets them empty when the tuple is zero.

The only other references to `:authority` in the tree are in the HPACK static table.

### Why it matters

gRPC and HTTP/2 clients lose their peer in exactly the cases where HTTP/1.x keeps it. For a
fleet whose internal calls are gRPC, that is the difference between a service graph with
edges and one without.

It will also surprise a reader: two client spans from the same process, one HTTP/1.x and
one HTTP/2, behave differently on the same unresolved connection.

### Suggested fix

Harvest `:authority` in the HTTP/2 transform the way `Host:` is harvested for HTTP/1.x, and
use it when the connection tuple is zero. The same limitation applies as for `Host:` — the
authority usually omits the port, so this recovers the address alone.

This is a long-standing gap; the behaviour has presumably always been this way.

darkraider01:
HTTP/1.x already falls back to the `Host:` header when the connection tuple can't be resolved, so client spans still get a peer. HTTP/2 has the same info sitting right there in the `:authority` pseudo-header, but nothing was reading it — so gRPC and HTTP/2 spans just went out with an empty peer whenever the connection was unresolved.

Added `:authority` extraction alongside the existing `:method`/`:path`/`content-type` parsing, and use it as the host fallback the same way `Host:` is used for HTTP/1.x (address only, since `:authority` usually omits the port anyway). Had to thread it through the per-stream cache too, since the header gets parsed once at capture time but the span only gets built later at stream completion.

Fixes #3283.

### DIFF
--- a/pkg/ebpf/common/http2grpc_transform.go
+++ b/pkg/ebpf/common/http2grpc_transform.go
@@ -8,6 +8,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"io"
+	"net"
 	"regexp"
 	"strconv"
 	"strings"
@@ -88,6 +89,7 @@ type h2RequestMeta struct {
 	method   string
 	path     string
 	fullPath string
+	host     string
 	grpc     bool
 }
 
@@ -273,7 +275,7 @@ func hpackOpensResponse(frag []byte) bool {
 	return b&formMask != formMask
 }
 
-func readMetaFrame(parseContext *EBPFParseContext, connID uint64, fr *http2.Framer, hf *http2.HeadersFrame) (string, string, string, bool, bool) {
+func readMetaFrame(parseContext *EBPFParseContext, connID uint64, fr *http2.Framer, hf *http2.HeadersFrame) (string, string, string, string, bool, bool) {
 	h2c := getOrInitH2Conn(parseContext.h2c, connID)
 
 	ok := false
@@ -281,9 +283,10 @@ func readMetaFrame(parseContext *EBPFParseContext, connID uint64, fr *http2.Fram
 	method := ""
 	path := ""
 	contentType := ""
+	authority := ""
 
 	if h2c == nil {
-		return method, path, contentType, ok, isResponse
+		return method, path, contentType, authority, ok, isResponse
 	}
 
 	h2c.hdec.SetEmitFunc(func(hf bhpack.HeaderField) {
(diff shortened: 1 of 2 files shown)


pkg/ebpf/common/http2grpc_transform.go
```diff
@@ -419,6 +425,16 @@ func readRetMetaFrame(parseContext *EBPFParseContext, connID uint64, fr *http2.F
 	return status, grpc, ok
 }
 
+// authorityHost strips the port from an :authority pseudo-header, mirroring how the
+// HTTP/1.x Host: header is handled. :authority usually omits the port, so this
+// recovers the address alone.
+func authorityHost(authority string) string {
+	if host, _, err := net.SplitHostPort(authority); err == nil {
```
mmat11: do we also need the port here? my claude says grpc-go also includes the port

mmat11 (commented):
can we add some assertions in the already existing integration tests?

pkg/ebpf/common/http2grpc_transform.go
darkraider01: You're right, and it checks out  grpc-go sets :authority to the dial target (host:port), unlike the plain HTTP/1.x Host: header which usually omits it. Added a splitAuthority() helper that parses both, and threaded the port through h2InfoToSpan/h2RequestMeta into HostPort. Covered by TestUnresolvedConnFallsBackToAuthority now asserting both.

darkraider01:
Good ask. I went looking for an existing suite to hook into, but none of the current gRPC/HTTP2 integration tests (grpc_relay, grpc_go_mux, yamux_grpc, old_grpc) deterministically land in the state this fix targets ConnInfo.S_port/D_port both zero. That only happens when the kernel-side connection lookup misses (e.g. async SSL timing, or a socket the discovery/attach hasn't resolved yet), which none of those suites force on purpose.

Added TestUnresolvedConnFallsBackToAuthority instead, which exercises the real header-capture -> HPACK decode -> cached authority -> span-assembly pipeline end to end with a synthetic zeroed ConnInfo, so the exact code path is deterministically covered without depending on kernel timing.

If you know of a suite that already forces an unresolved connection tuple (or a way to make one deterministic), point me at it and I'll add the assertion there.

MrAlias (approved):

NimrodAvni78 (approved):

PR #3304, opened 2026-09-03
tpinjector: invalidate msg buffers when fill_msg_buffers bails

mmat11:
## Summary

Follow-up to #3257.

When `fill_msg_buffers()` bails, the buffers still hold data from the previous send. Anything that reads them afterwards treats old data as the current message. #3257 fixed this for one reader, but two readers were still exposed.

This PR clears the buffers on every bail instead of guarding each reader.

## Validation

### DIFF
--- a/bpf/tpinjector/tpinjector.c
+++ b/bpf/tpinjector/tpinjector.c
@@ -659,15 +659,12 @@ int obi_sockmap_tracker(struct bpf_sock_ops *skops) {
     return 1;
 }
 
-// A bail must not leave the previous send's buffers behind: protocol_detector
-// and the tcp_sendmsg kprobe would consume them as if they were this send
+// A bail must not leave the previous send's buffers behind. Deleting the keyed
+// entry cuts off the tcp_sendmsg kprobe; msg_buffer_mem itself needs no
+// clearing because every reader is gated on fill_msg_buffers success or on the
+// msg_buffers entry deleted here — keep it that way
 static __always_inline void invalidate_msg_buffers(const egress_key_t *e_key) {
     bpf_map_delete_elem(&msg_buffers, e_key);
-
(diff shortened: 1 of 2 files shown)


NimrodAvni78 (approved):

NimrodAvni78 (approved):
LGTM

rafaelroquetto (approved):

bpf/tpinjector/tpinjector.c
```diff
@@ -658,6 +658,17 @@ int obi_sockmap_tracker(struct bpf_sock_ops *skops) {
...
 }
 
+// A bail must not leave the previous send's buffers behind: protocol_detector
+// and the tcp_sendmsg kprobe would consume them as if they were this send
+static __always_inline void invalidate_msg_buffers(const egress_key_t *e_key) {
+    bpf_map_delete_elem(&msg_buffers, e_key);
+
+    const unsigned char *zero = bpf_map_lookup_elem(&msg_buffer_zero, &(u32){0});
+    if (zero) {
+        bpf_map_update_elem(&msg_buffer_mem, &(u32){0}, zero, BPF_ANY);
```
MrAlias: Each `fill_msg_buffers` invocation for a known SSL connection takes the first bail, so this update copies the entire 8 KiB value into `msg_buffer_mem`. No remaining reader needs that copy: both `protocol_detector` call sites are now gated on `fill_msg_buffers` success, and the `tcp_sendmsg` paths cannot read `msg_buffer_mem` after the keyed `msg_buffers` entry above is deleted.

Could we drop `msg_buffer_zero` and keep only the keyed deletion?

MrAlias (commented):
The keyed stale-state cleanup is sound, but the new full-buffer invalidation adds avoidable hot-path work on known SSL connections.

bpf/tpinjector/tpinjector.c
mmat11: good catch, done

PR #3306, opened 2026-09-03
messaging spans: derive span kind from the operation, not the connection side

NimrodAvni78:
## Summary

Semantic conventions define no server-kind messaging span. The kind follows the messaging operation, not the side of the connection OBI observed the exchange from.

OBI reported server kind for `EventTypeKafkaServer`, `EventTypeMQTTServer` and `EventTypeNATSServer`, because those event types sat in `spanKind`'s request/response case alongside HTTP, gRPC, SQL and Redis. A Kafka exchange OBI classified as inbound was exported as a server span carrying `messaging.*` attributes.

Metrics already ignore the direction: `metrics.go` and `prom.go` group the `*Client` and `*Server` messaging types together and route both through `MessagingOperationTypeOf`. So the same exchange was recorded as a messaging metric while being exported as a server span. This makes the two agree.

## Change

A new `request.MessagingSpanKind` helper maps the operation to a kind, and both `spanKind` (traces) and `ServiceGraphKind` (service graph metrics) go through it:

- `send`, `publish` -> producer
- `process` -> consumer
- `receive`, `settle` -> client
- anything else stays internal

`receive` and `settle` are client rather than consumer operations: OBI observes the exchange with the broker, not what the application afterwards does with the message, which is what a consumer span describes.

`receive` and `settle` were previously unmapped and fell through to internal. `ServiceGraphKind` also had the `*Server` messaging types in its server case, so it is corrected alongside `spanKind` to keep the trace and the metric in agreement. Redis, SQL, Memcached, SunRPC and Aerospike server types keep server kind.

## Compatibility

Kafka, MQTT and NATS spans OBI saw from the receiving side change kind, from server to producer, consumer or client. Consumers filtering those spans on `kind=server`, or aggregating them alongside HTTP and RPC server spans, need to move to the messaging kinds. Span names, attributes and metric values are unchanged.

## Validation

Tests cover the helper directly, the resulting kind per event type in both directions, and an exhaustive assertion that no messaging event type and operation combination can produce server kind.


pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -1730,21 +1730,37 @@ func spanKind(span *request.Span) trace2.SpanKind {
...
 		return trace2.SpanKindServer
 	case request.EventTypeHTTPClient, request.EventTypeGRPCClient, request.EventTypeSQLClient, request.EventTypeRedisClient, request.EventTypeMongoClient, request.EventTypeCouchbaseClient, request.EventTypeMemcachedClient, request.EventTypeSunRPCClient, request.EventTypeAerospikeClient, request.EventTypeFailedConnect:
 		return trace2.SpanKindClient
-	case request.EventTypeKafkaClient, request.EventTypeMQTTClient, request.EventTypeNATSClient, request.EventTypeAMQPClient:
-		switch request.MessagingOperationTypeOf(span.Method) {
-		case request.MessagingSend:
-			return trace2.SpanKindProducer
-		case request.MessagingProcess:
-			return trace2.SpanKindConsumer
+	// A messaging span is a producer or a consumer whichever side of the
```
pinoOgni: The `ServiceGraphKind()` needs to be updated with the same logic otherwise there is a mismatch because trace says `CONSUMER`, metric says `SPAN_KIND_SERVER`

pinoOgni: Also, `TestKindString` needs to be updated for kafka, mqtt and nats.

mariomac (approved):
Nice improvement! I think the original behavior was written when OTEL did not define producer/consumer span types.

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -1730,21 +1730,37 @@ func spanKind(span *request.Span) trace2.SpanKind {
...
+		request.EventTypeNATSClient, request.EventTypeNATSServer,
+		request.EventTypeAMQPClient:
+		if kind, ok := messagingSpanKind(request.MessagingOperationTypeOf(span.Method)); ok {
+			return kind
 		}
 	}
 	return trace2.SpanKindInternal
 }
 
+func messagingSpanKind(operationType string) (trace2.SpanKind, bool) {
```
pinoOgni: After fixing `ServiceGraphKind` we can a helper named `MessagingSpanKind` in `span.go` that takes the operation name, calls `MessagingOperationTypeOf` inside and that is used by both `ServiceGraphKind` and `spanKind`?

pinoOgni (changes requested):

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -1730,21 +1730,37 @@ func spanKind(span *request.Span) trace2.SpanKind {
...
+	switch operationType {
```
pinoOgni: This switch should be like: send -> PRODUCER, receive -> CLIENT, process -> CONSUMER, settle -> CLIENT

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -1730,21 +1730,37 @@ func spanKind(span *request.Span) trace2.SpanKind {
...
+	case request.MessagingSend:
+		return trace2.SpanKindProducer, true
+	case request.MessagingReceive, request.MessagingProcess:
```
pinoOgni: Another thing is that technically `MessagingReceive` should not be here. Taking into account https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#operation-types, OBI does not observe what the app does with the message so receive -> process is not true.

pkg/appolly/app/request/span.go
```diff
@@ -1903,29 +1933,22 @@ func (s *Span) ResponseBodyLength() int64 {
...
-		case trace.SpanKindServer:
-			return "SPAN_KIND_SERVER"
-		case trace.SpanKindClient:
-			return "SPAN_KIND_CLIENT"
-		case trace.SpanKindProducer:
-			return "SPAN_KIND_PRODUCER"
-		case trace.SpanKindConsumer:
-			return "SPAN_KIND_CONSUMER"
+		if s.SpanKind != trace.SpanKindUnspecified && s.SpanKind != trace.SpanKindInternal {
+			return spanKindString(s.SpanKind)
```
pinoOgni: nit: redundant guard

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -1771,16 +1771,22 @@ func spanKind(span *request.Span) trace2.SpanKind {
...
 		return trace2.SpanKindServer
 	case request.EventTypeHTTPClient, request.EventTypeGRPCClient, request.EventTypeSQLClient, request.EventTypeRedisClient, request.EventTypeMongoClient, request.EventTypeCouchbaseClient, request.EventTypeMemcachedClient, request.EventTypeSunRPCClient, request.EventTypeAerospikeClient, request.EventTypeFailedConnect:
 		return trace2.SpanKindClient
-	case request.EventTypeKafkaClient, request.EventTypeMQTTClient, request.EventTypeNATSClient, request.EventTypeAMQPClient:
-		switch request.MessagingOperationTypeOf(span.Method) {
-		case request.MessagingSend:
-			return trace2.SpanKindProducer
-		case request.MessagingProcess:
-			return trace2.SpanKindConsumer
+	// A messaging span is a producer or a consumer whichever side of the
```
pinoOgni: "A messaging span is a producer or a consumer" -> "A messaging span is a producer, a consumer or a client".

pkg/export/otel/tracesgen/tracesgen_test.go
```diff
@@ -1190,6 +1190,54 @@ func TestManualSpanKind(t *testing.T) {
...
+
+func TestNoMessagingSpanIsReportedAsServerKind(t *testing.T) {
+	messaging := []request.EventType{
+		request.EventTypeKafkaClient, request.EventTypeKafkaServer,
+		request.EventTypeMQTTClient, request.EventTypeMQTTServer,
+		request.EventTypeNATSClient, request.EventTypeNATSServer,
+		request.EventTypeAMQPClient,
+	}
+	operations := []string{
+		request.MessagingSend, request.MessagingPublish,
```
pinoOgni, line 1229: missing `request.MessaginSettle`

pkg/export/otel/tracesgen/tracesgen_test.go
```diff
@@ -1190,6 +1190,54 @@ func TestManualSpanKind(t *testing.T) {
...
+		request.MessagingReceive, request.MessagingProcess, "",
+	}
+
+	for _, et := range messaging {
+		for _, op := range operations {
+			kind := spanKind(&request.Span{Type: et, Method: op})
+			assert.NotEqualf(t, trace2.SpanKindServer, kind,
+				"%v with operation %q reported server kind; semantic conventions define messaging spans as producer or consumer only", et, op)
```
pinoOgni: ""%v with operation %q reported server kind; semantic conventions define messaging spans as producer or consumer only"" -> ""%v with operation %q reported server kind; semantic conventions define messaging spans as never server" (or something similar)

pkg/appolly/app/request/span.go
```diff
@@ -1903,29 +1933,22 @@ func (s *Span) ResponseBodyLength() int64 {
...
-		switch s.SpanKind {
-		case trace.SpanKindServer:
-			return "SPAN_KIND_SERVER"
-		case trace.SpanKindClient:
-			return "SPAN_KIND_CLIENT"
-		case trace.SpanKindProducer:
-			return "SPAN_KIND_PRODUCER"
-		case trace.SpanKindConsumer:
-			return "SPAN_KIND_CONSUMER"
+		if s.SpanKind != trace.SpanKindUnspecified && s.SpanKind != trace.SpanKindInternal {
```
pinoOgni: nit: redundant guard

pinoOgni (changes requested):
The PR description/title need be updated for the "client" type.

pinoOgni (approved):
LGTM!

PR #3313, opened 2026-09-04
send integration spans to weaver

NimrodAvni78:
## Summary

Integration tests exported OBI traces straight to Jaeger, so the collector's weaver tap only ever received metrics. Suites that do assert spans (Elasticsearch, MCP, GraphQL, JSON-RPC, Kafka, Mongo, SQL, MQTT, Ruby, PHP) validated them through Jaeger while contributing nothing to weaver's view of emitted telemetry.

Each affected config now sends traces to the same collector endpoint it already uses for metrics. The collector forwards them to Jaeger exactly as before, so existing Jaeger assertions are unchanged. Three Kubernetes daemonsets had the same split and get the same fix; the k8s collector already had a Jaeger exporter added for this purpose.

`otelcol-config.yml` had no traces pipeline at all, so its two consumers would have lost spans entirely. It gains one matching the other collector configs.

Composes whose traces reach the weaver tap go from 13 to 56. The remaining weaver-enabled composes set no traces exporter and are metrics-only by design.

`obi-config-with-jaeger-host.yml` deliberately still points at Jaeger: it is shared with a compose that runs no collector.

## Validation


grcevski (approved):
LGTM!

PR #3314, opened 2026-09-04
derive SQS span kind from the messaging operation

NimrodAvni78:
## Summary

SQS spans are HTTP client spans carrying messaging attributes, so they were emitted with span kind `client` regardless of the operation. Semantic conventions map the kind from the messaging operation type, and `SendMessage` is the one SQS operation whose kind was wrong.

`SendMessage` and `SendMessageBatch` now emit `producer`. Every other operation stays `client`, which is what conventions ask for: `receive` and `settle` are client operations, since OBI observes the exchange with the broker rather than what the application does with the message afterwards. `ReceiveMessage` and `DeleteMessage` therefore keep the kind they already had, and so do the queue administration calls that carry no messaging operation type at all, along with S3.

Both kind paths go through the `request.MessagingSpanKind` helper, so the trace span kind and the metric kind cannot drift apart.

## Metrics also change

`Span.ServiceGraphKind` feeds more than the service graph. An SQS send moves from `SPAN_KIND_CLIENT` to `SPAN_KIND_PRODUCER` in:

- the `span.kind` label on span metrics
- the `kind` label on `traces_service_graph_request_*`
- the `kind` field in the JSON span printer

`SPAN_KIND_PRODUCER` is already a declared member of the `span.kind` enum, so no schema transformation is owed, but a dashboard or alert pinned to `span_kind="SPAN_KIND_CLIENT"` for an `sqs.SendMessage` span will see that series move.

## Tests

`fetchAWSSpanByOP` hard-coded `client` when looking a span up in Jaeger, so it found nothing once the kind changed. It now takes the expected kind, which the SQS assertions derive from the operation type through the same shared helper.

## Validation

Unit tests cover both kind paths, including the negative cases: a nil AWS payload, an S3 subtype, and an unmapped operation type. Reverting either production change fails them.

### DIFF
--- a/pkg/export/otel/tracesgen/tracesgen.go
+++ b/pkg/export/otel/tracesgen/tracesgen.go
@@ -1774,6 +1774,11 @@ func spanKind(span *request.Span) trace2.SpanKind {
 	case request.EventTypeHTTP, request.EventTypeGRPC, request.EventTypeRedisServer, request.EventTypeSunRPCServer, request.EventTypeMemcachedServer, request.EventTypeSQLServer, request.EventTypeAerospikeServer:
 		return trace2.SpanKindServer
 	case request.EventTypeHTTPClient, request.EventTypeGRPCClient, request.EventTypeSQLClient, request.EventTypeRedisClient, request.EventTypeMongoClient, request.EventTypeCouchbaseClient, request.EventTypeMemcachedClient, request.EventTypeSunRPCClient, request.EventTypeAerospikeClient, request.EventTypeFailedConnect:
+		if span.SubType == request.HTTPSubtypeAWSSQS && span.AWS != nil {
+			if kind, ok := request.MessagingSpanKind(span.AWS.SQS.OperationType); ok {
+				return kind
+			}
+		}
 		return trace2.SpanKindClient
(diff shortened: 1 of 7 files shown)


mariomac (approved):
good!

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -1724,6 +1719,17 @@ func genAIResponseErrorMessage(span *request.Span) string {
 	return ""
 }
 
+func messagingSpanKind(operationType string) (trace2.SpanKind, bool) {
```
pinoOgni: I left some comments here https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3306 (related to `ServiceGraphKind` and `messagingSpanKind`) which also apply to this PR. Maybe it's better to finish and merge that one first, and then do this one. WDYT?

pinoOgni: A dashboard filtering SQS by `span.kind=client` now breaks, so the PR description needs to updated.

mmat11 (approved):
lgtm

pinoOgni (approved):
LGTM!

PR #3244, opened 2026-08-29
rdns: add ipv6 support

rafaelroquetto:
## Summary

Fixes: https://github.com/grafana/beyla/issues/1769

Basically, we try to detect IPv6 and find the beginning of the DNS payload. Because the program size blew up, I had to split it into 2 different programs and tail call.

Disclaimer: rdns_xdp is weird in the sense that it has a different coding style -> I've kept it consistent / adhered to the file. A subsequent PR can be made to adjust that if one sees value in it.

## Validation

- [x] I have read and followed the [contributing guidelines](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/CONTRIBUTING.md)
- [x] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed.

PR #3215, opened 2026-08-27
publish an OBI telemetry reference generated from the registry

NimrodAvni78:
## Summary

Generates a reference of the attributes and metrics OBI defines from the semantic-convention registry in `schemas/obi/`, and publishes it alongside the telemetry schema on GitHub Pages. Adds `make generate-schema-docs`; rerun it and commit the result when the registry changes.

### What changed

- `scripts/generate-schema-docs.sh` + `scripts/schema-docs.jq` render `site/docs/{README,attributes,metrics}.md` (17 attribute groups, 42 metrics).
- `site/index.html` gives the published site a root page; previously `/` returned 404.
- `site/README.md` moves to `devdocs/telemetry-schema.md`: it is contributor documentation and was being served publicly as site content.
- A test asserts the resource `schema_url` OBI emits on traces, which had no coverage.

### Notes

- Rendering goes through `weaver registry resolve` plus jq rather than `weaver registry generate`, because generation aborts on the duplicate-attribute diagnostics from OBI's `x.obi.*` override groups, the same expected findings `scripts/lint-schema-filter.jq` allowlists for `registry check`. Resolve still emits the complete registry alongside those diagnostics.
- Groups are selected by provenance rather than by group id, because OBI defines metrics whose ids carry no `obi` marker (the spanmetrics, service-graph and `target.info` families).
- The site is published as static files with no markdown processing, so the generated pages are served as raw text. `site/index.html` links the rendered copies for reading and the published files for fetching.

## Validation

- [ ] I have read and followed the [contributing guidelines](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/CONTRIBUTING.md)
- [ ] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed.

PR #3317, opened 2026-09-04
Prepare obi for version v0.13.0

mmat11:
v0.13.0 is a feature release. It also fixes a group of memory-safety bugs in the socket-layer context propagation that could corrupt TLS connections of instrumented workloads or copy unrelated kernel memory into telemetry.

### Context-propagation memory safety

- The HTTP header injector no longer acts on leftover buffer data from a previous request, which could write a traceparent into the middle of a TLS stream and reset the connection. ([#3257](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3257))
- Message-buffer reads are bounded by the mapped data window, so a failed pull can no longer copy adjacent kernel memory into span payloads. Also fixes a masking bug that copied nothing for sends of 8 KiB or more. ([#3298](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3298))
- The message buffers are invalidated whenever filling them fails, so no reader can mistake a previous message for the current one. ([#3304](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3304))

### Protocol and tracing fixes

- Application-supplied traceparents are preserved in Go HTTP/2, and header injection there fails safely. ([#3156](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3156), [#3159](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3159))
- TCP client and server roles are classified by the local port, fixing swapped roles for same-namespace clients. ([#3146](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3146))
- Established idle connections are no longer reported as failed connects. ([#3043](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3043))
- DNS spans follow OTel conventions, peer naming from DNS captures is fixed, and reverse DNS supports IPv6. ([#3174](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3174), [#2947](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2947), [#3244](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3244))
- Exiting processes keep their PID valid for a grace period so their final events are not lost. ([#3170](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3170))

### New instrumentation and metrics

- Runtime metrics for Python, JVM, and Node.js; Go runtime probes are gated behind the `application_runtime` feature. ([#3028](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3028), [#3203](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3203), [#3131](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3131), [#3227](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3227))
- New attributes: `db.response.status_code`, `db.namespace`, and `server.port` on database duration metrics; `messaging.operation.name` on messaging spans. ([#3271](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3271), [#3193](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3193), [#3097](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3097), [#3136](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3136))
- Aerospike server spans, Go 1.27 support, and better service naming for Python, Deno, .NET, and Node.js apps. ([#3167](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3167), [#3221](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3221), [#3112](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3112), [#3132](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3132), [#3118](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3118), [#3099](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3099))
- OBI publishes its telemetry schema and a generated telemetry reference; `schema_url` now points at 0.13.0. ([#3044](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3044), [#3215](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3215))

### Behavior changes

- `service.name` and `service.namespace` are no longer default metric labels; select them explicitly if you depend on them. ([#3168](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3168))
- `error.type` values are unified across spans and metrics; the attribute name is unchanged. ([#3196](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3196))
- Go runtime metrics require enabling the `application_runtime` feature. ([#3227](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3227))
- `metrics.features` is printed as feature names instead of a bitmask. ([#3265](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3265))

**Full Changelog**: [v0.12.2...v0.13.0](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/compare/v0.12.2...v0.13.0)

### DIFF
--- a/internal/test/oats/ai/go.mod
+++ b/internal/test/oats/ai/go.mod
@@ -28,7 +28,7 @@ require (
 	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
 	go.opentelemetry.io/collector/featuregate v1.64.0 // indirect
 	go.opentelemetry.io/collector/pdata v1.64.0 // indirect
-	go.opentelemetry.io/obi v0.12.2 // indirect
+	go.opentelemetry.io/obi v0.13.0 // indirect
 	go.opentelemetry.io/otel v1.46.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.46.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.46.0 // indirect
--- a/internal/test/oats/amqp/go.mod
+++ b/internal/test/oats/amqp/go.mod
@@ -28,7 +28,7 @@ require (
 	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
 	go.opentelemetry.io/collector/featuregate v1.64.0 // indirect
 	go.opentelemetry.io/collector/pdata v1.64.0 // indirect
-	go.opentelemetry.io/obi v0.12.2 // indirect
+	go.opentelemetry.io/obi v0.13.0 // indirect
 	go.opentelemetry.io/otel v1.46.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.46.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.46.0 // indirect
--- a/internal/test/oats/harness/go.mod
+++ b/internal/test/oats/harness/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/grafana/oats v0.7.0
 	github.com/onsi/ginkgo/v2 v2.32.0
 	github.com/onsi/gomega v1.41.0
-	go.opentelemetry.io/obi v0.12.2
+	go.opentelemetry.io/obi v0.13.0
 )
 
 // The oats harness reuses the shared weaver-validation logic
--- a/internal/test/oats/http/go.mod
+++ b/internal/test/oats/http/go.mod
@@ -28,7 +28,7 @@ require (
 	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
 	go.opentelemetry.io/collector/featuregate v1.64.0 // indirect
 	go.opentelemetry.io/collector/pdata v1.64.0 // indirect
-	go.opentelemetry.io/obi v0.12.2 // indirect
+	go.opentelemetry.io/obi v0.13.0 // indirect
 	go.opentelemetry.io/otel v1.46.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.46.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.46.0 // indirect
--- a/internal/test/oats/kafka/go.mod
+++ b/internal/test/oats/kafka/go.mod
@@ -28,7 +28,7 @@ require (
 	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
 	go.opentelemetry.io/collector/featuregate v1.64.0 // indirect
 	go.opentelemetry.io/collector/pdata v1.64.0 // indirect
-	go.opentelemetry.io/obi v0.12.2 // indirect
+	go.opentelemetry.io/obi v0.13.0 // indirect
 	go.opentelemetry.io/otel v1.46.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.46.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.46.0 // indirect
--- a/internal/test/oats/memcached/go.mod
+++ b/internal/test/oats/memcached/go.mod
@@ -28,7 +28,7 @@ require (
 	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
 	go.opentelemetry.io/collector/featuregate v1.64.0 // indirect
 	go.opentelemetry.io/collector/pdata v1.64.0 // indirect
-	go.opentelemetry.io/obi v0.12.2 // indirect
+	go.opentelemetry.io/obi v0.13.0 // indirect
 	go.opentelemetry.io/otel v1.46.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.46.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.46.0 // indirect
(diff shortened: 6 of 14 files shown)


NimrodAvni78 (approved):
Nice!

MrAlias:
@mmat11 what do you think about merging #3157 and #3158 into this release? They are both approved and look ready.

MrAlias:
Based on the [dashboard](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/3308), I think we can look into merging these as well:

- #3155 
- #3270 
- #3258 
- #3272 
- #3299 
- #3303

mmat11:
unless something is critical I'd avoid adding risk to this changeset -- e.g. #3270 refactors the same code as the fixes we're releasing, and after a rebase part of it likely isn't needed anymore since they overlap (#3298 already fixed the large-request drop)

on #3157 I'm still unsure about the goexec changes and would like more eyes on it; even if we merge it, I'd rather let it sit in main for a bit than release it straight away

MrAlias:
I agree that we shouldn’t add unreviewed risk to the release at the last minute.

My concern is that this risk is only becoming a blocker at release cut. #3157 has been open since August 24 and approved since August 26. The concrete issue raised on August 27 was addressed that day. If there are still concerns about the goexec changes, could you identify the specific blockers on the PR so I can address them, or pull in the additional reviewers you think it needs?

Could you also address #3158 separately? It has been open since August 24, has two approvals, and the review feedback has been addressed. It is itself a safety fix for rollback failures in the existing `sk_msg` HTTP/2 mutator. If there is remaining risk that should prevent it from shipping, I’d like that risk made explicit.

Is there a specific reason v0.13.0 needs to ship now? Cutting a release immediately after landing #3298 seems like a risk in itself. If we are concerned about stabilization, allowing a short window before the release would give us time to resolve any concrete concerns.

I understand why #3298 was treated as critical, and I’m not arguing that releases should be FIFO. What I want to avoid is approved work sitting without actionable blockers and then being excluded because it hasn’t had time to stabilize, while newer work is expedited into the same release without an equivalent stabilization period. Critical fixes may justify different treatment, but if we are applying different criteria, I think those criteria should be explicit.

If there are concrete blockers for #3157 or #3158, please identify them. Otherwise, I think they should be included in v0.13.0.

mmat11:
> I understand why #3298 was treated as critical, and I’m not arguing that releases should be FIFO. What I want to avoid is approved work sitting without actionable blockers and then being excluded because it hasn’t had time to stabilize, while newer work is expedited into the same release without an equivalent stabilization period. Critical fixes may justify different treatment, but if we are applying different criteria, I think those criteria should be explicit.

the difference between the two is that #3298 is ~20 lines of manually reviewable code while #3157 and #3158 are harder to review (e.g. [this](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3157/changes#diff-a3a35830513f0dae3c257227cb3e1f9cc1b6fbac1959160e99b6aeea903bb0bbR52) is code that I don't understand unless I point an AI at it and blindly trust it, which is not the level of confidence I want to have when doing a release)

> Is there a specific reason v0.13.0 needs to ship now?

the first 3 PRs linked in the description are breaking real apps outside OBI (e.g. Istio sidecar + mTLS) so I believe they should have higher priority over bugs that break OBI itself, this is why I proposed an urgent release in yesterday's SIG meeting

> If there are concrete blockers for https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3157 or https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3158, please identify them. Otherwise, I think they should be included in v0.13.0.

would it be ok if we ship v0.13.0 as is, and I'll commit to reviewing #3157 and #3158 this week so we can cut a v0.13.1 with them and the rest of the PRs linked in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3317#issuecomment-5542809414 next week?

MrAlias:
Thanks—the proposed review and v0.13.1 timeline works for me. I do want to push back on two parts of the characterization, though.

First, #3298’s review burden was not simply “about 20 lines of manually reviewable code.” Review caught that its first revision would reject most requests under 256 bytes, established that the mapped window could replace the more complicated size calculation, uncovered the separate 8 KiB masking bug, and found that the initial implementation was rejected by an unpatched 5.8 verifier. That required reasoning about the data bounds, generated BPF, and supported kernel behavior, including testing on an actual 5.8 environment.

The final patch being small was partly an outcome of that review, not evidence that the review itself was trivial. Diff size certainly affects review burden, but it is not a substitute for evaluating the behavior, invariants, test coverage, and platform compatibility of a change.

If the instruction-decoding code in #3157 is the remaining concern, that is a valid reason to withhold approval until it is understood. Please raise the specific questions or invariants that need to be established on the PR, and I’ll explain the design, add supporting evidence, or revise it as needed. I don’t expect anyone to trust an AI’s interpretation of the code 😄.

Second, #3158 is not merely a bug that breaks OBI internally. It addresses mutation failures that can leave an invalid HTTP/2 message, reject the application’s write, and force its connection to be retired. Its impact is application-visible, even if the implementation is harder to review than #3298.

With the commitment to review #3157 and #3158 this (upcoming?) week and, assuming the review concerns are resolved, include them in v0.13.1 next week, I’m okay with shipping v0.13.0 as-is. I’ll prioritize responding to feedback there.

MrAlias (approved):

PR #3319, opened 2026-09-04
tpinjector: fix wrong pointer type

rafaelroquetto:
## Summary

bpf_map_lookup_elem() returns a pointer to unsigned char[N] in this particular instance, which decays to unsigned char *. We are wrongly taking unsigned char ** (i.e. a pointer to a pointer, not a pointer to a unsigned char array). It's been mostly fine because it gets cast to (void*) later and passed around as such, but when we do:
```
  ptr[0] = '\0'
```
we are actually writing a 64-bit word, rather than  a single char (i.e. writing a pointer address) - which is fine for the purpose, but semantically wrong.

## Validation

### DIFF
--- a/bpf/tpinjector/tpinjector.c
+++ b/bpf/tpinjector/tpinjector.c
@@ -708,7 +708,7 @@ static __always_inline bool fill_msg_buffers(struct sk_msg_md *msg,
     bpf_clamp_umax(fallback_bytes, k_kprobes_http2_buf_size);
     bpf_probe_read_kernel(msg_buf.fallback_buf, fallback_bytes, msg->data);
 
-    unsigned char **msg_ptr = bpf_map_lookup_elem(&msg_buffer_mem, &(u32){0});
+    unsigned char *msg_ptr = bpf_map_lookup_elem(&msg_buffer_mem, &(u32){0});
 
     if (!msg_ptr) {
         bpf_d_printk("failed to reserve msg_buffer space [%s]", __FUNCTION__);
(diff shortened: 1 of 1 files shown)


bpf/tpinjector/tpinjector.c
```diff
@@ -708,7 +708,7 @@ static __always_inline bool fill_msg_buffers(struct sk_msg_md *msg,
     bpf_clamp_umax(fallback_bytes, k_kprobes_http2_buf_size);
     bpf_probe_read_kernel(msg_buf.fallback_buf, fallback_bytes, msg->data);
 
-    unsigned char **msg_ptr = bpf_map_lookup_elem(&msg_buffer_mem, &(u32){0});
+    unsigned char *msg_ptr = bpf_map_lookup_elem(&msg_buffer_mem, &(u32){0});
```
MrAlias, line 711: `protocol_detector` still declares the same array-valued map entry as `unsigned char **` and casts it back to `const unsigned char *` before parsing. This is the same pointer-type mismatch the PR fixes at the write site. Could we make this declaration `unsigned char *` too and remove the cast?

MrAlias (commented):
The write-site correction is sound, but the paired read path should be kept type-consistent.

bpf/tpinjector/tpinjector.c
rafaelroquetto: done

MrAlias (approved):
🚀

PR #3321, opened 2026-09-04
Go manual SDK conformance

NimrodAvni78:
## Summary

Spans relayed from the Go manual SDK carry attributes chosen by the instrumented application, which can never exist in OBI's registry, so weaver reports each one as a violation once those spans reach it. The integration test applications now namespace their custom attributes under `obitest.`, and a single scoped filter accepts that namespace; an undeclared attribute anywhere else, including on an OBI-authored span, still surfaces.

The `RecordError` uprobe named its attribute `error message`, which is both absent from the registry and invalid as an attribute name. It now emits `exception.message`, the stable semantic convention for a recorded error. Unlike the application-authored attributes, this one OBI sets itself, so it is declared rather than filtered: an attribute group references it, which is what puts it in the resolved registry, since an upstream attribute no group references is not part of it.

That same code bounded the error text against the attribute *key* limit while writing into the *value* buffer, so any error message of 32 bytes or more was dropped instead of recorded. It now clamps the length to the value limit with `bpf_clamp_umax`, truncating a long message rather than losing it, rejects a non-positive length, and commits the attribute only when the read succeeds.

## Validation

### DIFF
--- a/bpf/gotracer/go_sdk.c
+++ b/bpf/gotracer/go_sdk.c
@@ -40,7 +40,7 @@ enum { k_go_ptr_arr_size = 16 };
 enum { k_go_auto_activation_max_attempts = 3 };
 enum { k_efault = 14 };
 
-const char ERROR_KEY[] = "error message";
+const char ERROR_KEY[] = "exception.message";
 const u32 ERROR_KEY_SIZE = sizeof(ERROR_KEY) - 1;
 
 typedef struct span_info {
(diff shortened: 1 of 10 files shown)


mariomac (approved):

bpf/gotracer/go_sdk.c
```diff
@@ -906,16 +906,18 @@ int GUARDED_PROG(obi_uprobe_RecordError, struct pt_regs *, ctx) {
...
-                span->span_attrs.attrs[valid_attrs].val_length = go_str.len;
-                span->span_attrs.attrs[valid_attrs].vtype = attr_type_string;
-                span->span_attrs.valid_attrs = valid_attrs + 1;
+            if ((go_str.len > 0) && (valid_attrs < OTEL_ATTRIBUTE_MAX_COUNT)) {
+                u32 val_len = (u32)go_str.len;
+                bpf_clamp_umax(val_len, OTEL_ATTRIBUTE_VALUE_MAX_LEN - 1);
+
+                if (bpf_probe_read_user(
+                        span->span_attrs.attrs[valid_attrs].value, val_len, go_str.str) == 0) {
+                    __builtin_memcpy(
```
mmat11, line 915:
```suggestion
                    bpf_memcpy(
```

mmat11 (approved):
lgtm

PR #3322, opened 2026-09-04
Discover Go runtime ABI facts from target DWARF

### LINKED ISSUE #3235: Discover Go runtime ABI facts from target DWARF
## Problem

#3221 adds Go 1.27 itab discovery from `runtime.moduledata`. The decoder currently loads the new moduledata field offsets from generated `offsets.json` data and the remaining runtime ABI facts from generated `go_abi_offsets.json` data.

This differs from OBI's normal struct-offset flow, which first discovers offsets from the target binary's DWARF and uses generated data only as a fallback. As a result, an ungenerated future Go minor is rejected even when the target binary contains enough DWARF to describe its runtime ABI. The rejection is safe, but it gives up compatibility that dynamic discovery could provide for non-stripped binaries.

Stripped binaries must continue using generated data because they do not contain DWARF.

## Goal

Discover the complete Go runtime ABI required by moduledata itab decoding from target DWARF when available, then fall back to versioned generated data when it is absent or incomplete.

## Scope

- refactor the DWARF extraction used by `internal/goabioffsets` into reusable code;
- dynamically discover the required `runtime.moduledata` fields, `internal/abi` field offsets, type sizes, and constants;
- treat the dynamically discovered ABI as one complete result rather than mixing an unverified partial ABI with generated facts;
- retain generated-data support for stripped binaries;
- retain safe rejection when neither source provides a complete ABI for the target Go version.

## Acceptance criteria

- a non-stripped binary with a complete supported DWARF ABI can be decoded without a matching generated Go minor;
- a binary with partial or incompatible DWARF falls back to a complete generated ABI or is rejected;
- a stripped binary continues to use the generated ABI path;
- tests cover complete DWARF, incomplete DWARF, stripped binaries, and an ungenerated future Go minor;
- the runtime path and the update-time generator use the same ABI fact definitions and extraction logic.

This is separate from #3234, which tracks atomic generation and consistency between the generated artifacts.

MrAlias:
## Summary

I centralized the private Go runtime ABI facts used for moduledata and itab decoding in a shared `internal/goabi` package.

For Go 1.27 and newer, OBI now:

- discovers and validates the complete runtime ABI from the target binary's DWARF;
- falls back to version-bounded generated offsets when DWARF is unavailable, incomplete, or incompatible;
- rejects the ABI when neither source is complete, without mixing facts between sources.

The offset generator now uses the same definitions and validation as the runtime path and owns all `runtime.moduledata` facts. Existing generated-offset behavior remains unchanged for pre-1.27 and stripped binaries.

This also adds coverage for DWARF extraction, invalid and conflicting facts, generated fallback, stripped binaries, and ungenerated future Go versions.

Closes #3235.

## Validation

- [x] `make update-offsets`
- [x] `go test -race -count=1 ./internal/goabi ./internal/gooffsets ./pkg/internal/goexec`
- [x] `make verify`


configs/offsets/tracker_input.json
```diff
@@ -41,18 +41,6 @@
       "database/sql.driverConn": ["ci"],
       "net/textproto.Reader": ["R"],
       "bufio.Reader": ["buf", "w"],
-      "runtime.moduledata": [
```
grcevski, line 44: Why did we remove these? Do we never discover these with offsets anymore?

grcevski (approved):
This looks like a good refactor to me, but I can't say I fully verified all code movement. I'm confident it works since we have many tests that would fail otherwise.

configs/offsets/tracker_input.json
MrAlias: These are still discovered and included in the generated offsets. They moved out of `tracker_input.json` because `internal/goabi.Requirements` now owns the complete runtime ABI definition, including the `runtime.moduledata` fields for the applicable Go versions.

`collectABIFacts` extracts those facts from DWARF—or reuses the validated cache—and writes them to the same `offsets.json`.

PR #3323, opened 2026-09-04
Fix pre-fork Python service name attribution falling back to python3.14

### LINKED ISSUE #3289: A pre-fork server's service name depends on which of its processes discovery selects
### What happens

`monitorPIDs()` registers the discovered process and its children, and stops there
(`pkg/appolly/discover/attacher.go:472`):

```go
// allowing the tracer to forward traces from the discovered PID and its children processes
tracer.AllowPID(ie.FileInfo.Pid(), ie.FileInfo.Ns(), ie.FileInfo)
for _, pid := range ie.ChildPids {
    tracer.AllowPID(pid, ie.FileInfo.Ns(), ie.FileInfo)
}
```

A pre-fork server has a master and several workers, and every one of them serves
requests. Which of them discovery selects therefore decides how much of the application
is covered:

- Select the **master**, and the workers are its children, so all of them carry the
  discovered service name.
- Select a **worker**, and that worker has no children. The master and the sibling
  workers fall outside the registration and their spans take the executable name
  instead.

The selection varies between runs, because every process in the group satisfies the
open-port criterion. Gunicorn binds the listening socket in the master and the workers
inherit it, so all four hold the same socket:

```
listening socket inodes: 1100360
pid=1 holds listening socket 1100360
pid=8 holds listening socket 1100360
pid=9 holds listening socket 1100360
pid=10 holds listening socket 1100360
```

Their command lines are identical as well, so the framework parser in
`pkg/internal/pythontools/frameworks/gunicorn.go` reads the same `main_ssl:app` target
for each. The processes are alike in every respect the naming logic examines. Only the
parent-child relation separates them.

### Measured

`internal/test/integration`, `gunicorn -w 3 -k uvicorn.workers.UvicornWorker main_ssl:app`,
28 runs on one host. The correlation is exact:

```
attached pid 1 (the master)   -> spans export as main_ssl     18 runs
attached pid 31, 32, 33, 34   -> spans export as python3.14   10 runs
```

The attacher reports the same name either way. This line appears in both cases:

```
msg="instrumenting process" cmd=/usr/local/bin/python3.14 pid=31 ino=6858158
  type=python service=main_ssl
```

The exported span disagrees with it:

```
selected the master:
  SQLClient ... SELECT accounting.contacts(accounting.contacts)
  [172.18.0.7 as main_ssl.integration-test:32944]->[...:5432]
  svc=[integration-test/main_ssl python]

selected a worker:
  SQLClient ... SELECT accounting.contacts(accounting.contacts)
  [172.18.0.7 as python3.14.integration-test:33194]->[...:5432]
  svc=[integration-test/python3.14 python]
```

The span is otherwise identical: same operation, same table, same content length. The
service name is the only field that moves.

### Why it matters

The name reaches the backend as `service.name`, so a run that selects a worker splits one
application into two services and breaks every service-graph edge through it. It is also
quiet: each span looks well formed, the attacher log names the service the operator
expects, and the disagreement appears only in the exported data.

Pre-fork is the default deployment shape for Python and Ruby web applications, so the
condition is common rather than exotic. Any group where several processes share a
listening socket has it.

### Suggested fix

Register the whole process group rather than the selected process and its descendants.
Discovery already knows the executable inode and the namespace, and every process in the
group shares both, so the group is identifiable from what is on hand. Selecting the
process group leader would also work and is a smaller change.

Whichever route, the useful property is that the answer stops depending on which member
discovery happened to reach first.

### LINKED ISSUE #3290: TestSuite_PythonSQLSSL passes only when discovery selects the gunicorn master
### What happens

`testREDMetricsForPythonSQLSSL` asks Jaeger for traces by service name
(`internal/test/integration/red_test_python_sql.go:425`):

```go
http.Get(jaegerQueryURL + "?service=" + comm + "&operation=SELECT%20accounting.contacts")
```

`comm` is `main_ssl`. The test server runs
`gunicorn -w 3 -k uvicorn.workers.UvicornWorker main_ssl:app`, so a master and three
workers all serve requests and all hold the listening socket. The exported service name is
`main_ssl` when discovery selects the master and the executable name `python3.14` when it
selects a worker, which is the subject of #3289.

So the assertion rests on a choice the test has no say in. When discovery selects a
worker, the query finds nothing and the suite fails:

```
red_test_python_sql.go:435
  Error: "1" is not less than or equal to "0"
red_test_python_sql.go:425
  Error: Condition never satisfied
```

The span itself is present and correct throughout. It is filed under the other name.

### Measured

We ran 28 runs on one host. Each failure came with a worker selected, and each pass with the
master:

```
attached pid 1   -> PASS   18 runs
attached worker  -> FAIL   10 runs
```

The two assertions in this test read the same spans from two backends and disagree, which
is what makes the cause easy to miss. Line 421 queries Prometheus for
`db_client_operation_duration_seconds_count{db_operation_name="SELECT"}` and passes,
because that query names no service. Line 435 queries Jaeger by service and fails. A
reader who sees metrics arriving and traces missing will look at the trace pipeline, and
the trace pipeline is working.

On plain upstream main this suite passed 7 times out of 7 for us, so we are reporting a
dependency rather than a failure anyone will necessarily see today. The runs that exposed
it carried local changes. Those changes alter neither discovery nor naming, which is
consistent with them shifting the timing of a choice rather than making it.

### Why it matters

A test that depends on a nondeterministic choice reports on that choice rather than on the
behavior it names. Failing, it sends a reader toward SQL parsing or SSL handling, and both
are fine. Passing, it says nothing about whether the service name is stable.

### Suggested fix

The direct repair is to run gunicorn with one worker, which removes the choice and leaves
the test measuring what its name says. `-w 1` is enough.

Keeping three workers is worth considering instead, since a pre-fork server is a shape
worth covering. In that case assert the service name explicitly, so a run that names the
application after its interpreter fails with a message that says so.

darkraider01:
A gunicorn master and its workers all match the open port criterion, so which one discovery sees first is a race. When a worker is discovered before the master, only the worker's own service metadata gets resolved. The master never does, so it falls back to its executable name (python3.14) instead of the real service name (main_ssl). This is what made TestSuite_PythonSQLSSL flaky.

Fix has two parts:

In typer.go, the cached path now walks up the process tree the same way the uncached path already did, still scoped to Python only, so a worker discovered later still finds the right ancestor to link its RuntimeMetricServiceSource to.

In attacher.go, resolveServiceMetadata now also resolves metadata on that RuntimeMetricServiceSource target directly, and syncServiceMetadata propagates a resolved name between worker and master in whichever direction it's missing, before ApplyServiceDefaults can lock in the executable-name fallback. Explicit names always win over auto-derived ones.

Also dropped Dockerfile_ssl to a single worker for deterministic tests, and deleted ssl_server.sh, an unused script that wasn't wired into any Dockerfile or compose file.

Validation: go build, go vet and go test all clean on pkg/appolly/discover, including new tests covering the sync precedence rules, the ancestor walk on the cached path, and the exact main_ssl vs python3.14 regression. pythontools tests unaffected. gofmt clean.

Fixes #3289. Fixes #3290.

### DIFF
--- a/internal/test/integration/components/pythonsql/Dockerfile_ssl
+++ b/internal/test/integration/components/pythonsql/Dockerfile_ssl
@@ -8,4 +8,4 @@ RUN pip install --require-hashes -r /requirements.txt
 COPY main_ssl.py /main_ssl.py
 COPY server.crt /server.crt
 COPY server.key /server.key
-CMD ["gunicorn", "--keyfile", "/server.key", "--certfile", "/server.crt", "-w", "1", "-b", "0.0.0.0:8080", "-k", "uvicorn.workers.UvicornWorker", "main_ssl:app"]
+CMD ["gunicorn", "--keyfile", "/server.key", "--certfile", "/server.crt", "-w", "3", "-b", "0.0.0.0:8080", "-k", "uvicorn.workers.UvicornWorker", "main_ssl:app"]
--- a/internal/test/integration/components/pythonsql/Dockerfile_ssl
+++ b/internal/test/integration/components/pythonsql/Dockerfile_ssl
@@ -8,4 +8,4 @@ RUN pip install --require-hashes -r /requirements.txt
 COPY main_ssl.py /main_ssl.py
 COPY server.crt /server.crt
 COPY server.key /server.key
-CMD ["gunicorn", "--keyfile", "/server.key", "--certfile", "/server.crt", "-w", "3", "-b", "0.0.0.0:8080", "-k", "uvicorn.workers.UvicornWorker", "main_ssl:app"]
+CMD ["gunicorn", "--keyfile", "/server.key", "--certfile", "/server.crt", "-w", "1", "-b", "0.0.0.0:8080", "-k", "uvicorn.workers.UvicornWorker", "main_ssl:app"]
--- a/pkg/appolly/discover/attacher.go
+++ b/pkg/appolly/discover/attacher.go
@@ -217,11 +217,11 @@ func syncServiceMetadata(dst, src *exec.FileInfo) {
 	switch {
 	case srcAttrs.UID.Name != "" && !src.AutoName():
 		if dst.AutoName() || dstAttrs.UID.Name == "" {
-			dst.SetUID(srcAttrs.UID)
+			dst.SetExplicitServiceName(srcAttrs.UID.Name)
 		}
 	case dstAttrs.UID.Name != "" && !dst.AutoName():
 		if src.AutoName() || srcAttrs.UID.Name == "" {
-			src.SetUID(dstAttrs.UID)
+			src.SetExplicitServiceName(dstAttrs.UID.Name)
 		}
 	case srcAttrs.UID.Name != "" && dstAttrs.UID.Name == "":
 		dst.SetAutoServiceName(srcAttrs.UID.Name)
--- a/pkg/appolly/discover/attacher.go
+++ b/pkg/appolly/discover/attacher.go
@@ -19,6 +19,7 @@ import (
 	"go.opentelemetry.io/obi/pkg/appolly/discover/exec"
 	"go.opentelemetry.io/obi/pkg/ebpf"
 	ebpfcommon "go.opentelemetry.io/obi/pkg/ebpf/common"
+	attr "go.opentelemetry.io/obi/pkg/export/attributes/names"
 	"go.opentelemetry.io/obi/pkg/export/imetrics"
 	"go.opentelemetry.io/obi/pkg/internal/denotools"
 	"go.opentelemetry.io/obi/pkg/internal/dotnettools"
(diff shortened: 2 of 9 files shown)


darkraider01:
Bundling 
[open-telemetry/opentelemetry-ebpf-instrumentation#3289](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/3289)
 and 
[open-telemetry/opentelemetry-ebpf-instrumentation#3290](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/3290)
 here since they're the same root cause. The typer.go/attacher.go fix resolves 
[open-telemetry/opentelemetry-ebpf-instrumentation#3289](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/3289)
, and that's also what makes TestSuite_PythonSQLSSL stop flaking for 
[open-telemetry/opentelemetry-ebpf-instrumentation#3290](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/3290)
. The Dockerfile_ssl and ssl_server.sh changes are just cleanup on top, dropping to one worker cuts the test's surface area further and the script was dead code. Happy to split into two PRs if you'd rather review them separately.

pkg/appolly/discover/attacher.go
```diff
@@ -182,33 +184,70 @@ func (ta *traceAttacher) attacherLoop(_ context.Context) (swarm.RunFunc, error)
...
+		return
+	}
+	dstAttrs := dst.ServiceAttrs()
+	srcAttrs := src.ServiceAttrs()
+
+	// 1. Service UID name precedence: Explicit (non-auto) > auto-derived > empty
+	switch {
+	case srcAttrs.UID.Name != "" && !src.AutoName():
+		if dst.AutoName() || dstAttrs.UID.Name == "" {
+			dst.SetUID(srcAttrs.UID)
```
NimrodAvni78: SetUID doesn't carry the flags, so the receiver gets an explicit name that still reports AutoName() == true.  The k8s decoration is supposed to lose to a configured name, but since the name looks auto it overwrites it anyway.

Concrete case: deployment checkout-api, gunicorn, with

```
discovery:
  instrument:
    - open_ports: 8080
      name: payments
    - exe_path: gunicorn
```

The worker matches the first criterion and gets`payments`. the master matches only the second and falls back to the `gunicorn`. The sync gives the master `payments` but leaves it flagged auto, k8s overwrites it with `checkout-api`, and since the master is the service source for every worker PID, all spans export as checkout-api instead of payments. Under k8s/Docker this branch ends up a no-op.

There's no way to clear autoName today, so this needs a small name-only setter that also clears the flag, used in place of SetUID in both explicit branches. Also worth an AutoName() == false assertion in the "explicit src overrides auto dst" subtest — the only one today is on the empty-dst case, where dst was never auto-named.

internal/test/integration/components/pythonsql/Dockerfile_ssl
```diff
@@ -8,4 +8,4 @@ RUN pip install --require-hashes -r /requirements.txt
 COPY main_ssl.py /main_ssl.py
 COPY server.crt /server.crt
 COPY server.key /server.key
-CMD ["gunicorn", "--keyfile", "/server.key", "--certfile", "/server.crt", "-w", "3", "-b", "0.0.0.0:8080", "-k", "uvicorn.workers.UvicornWorker", "main_ssl:app"]
+CMD ["gunicorn", "--keyfile", "/server.key", "--certfile", "/server.crt", "-w", "1", "-b", "0.0.0.0:8080", "-k", "uvicorn.workers.UvicornWorker", "main_ssl:app"]
```
NimrodAvni78: Could we keep -w 3 here? -w 1 still gives a master plus one worker, both holding the inherited listening socket, so which one discovery reaches first is still a coin flip — it shrinks the candidate set rather than making it deterministic.

And `testREDMetricsForPythonSQLSSL` (comm = "main_ssl", used as the Jaeger ?service= param) is the only integration test that asserts the derived gunicorn name can we make sure this is asserted

NimrodAvni78 (changes requested):
Good Idea! still have a few blockers

pkg/appolly/discover/attacher.go
darkraider01: thanks! You were spot on, I was copying the name over, but never clearing the "auto-generated" label, so Kubernetes could still overwrite it later.

I updated it so setting an explicit name now clears that flag, and added test assertions to make sure it stays cleared.

internal/test/integration/components/pythonsql/Dockerfile_ssl
darkraider01: Now that the race condition is actually fixed, running with 3 workers gives us a much better real-world test.

Reverted this back to -w 3

NimrodAvni78 (approved):
LGTM

mariomac (approved):
LGTM!

PR #3326, opened 2026-09-06
Fix genai attribution & Expand coverage of genai tests

NimrodAvni78:
## Summary

Adds an OATS suite for the gen_ai retrieval client span, fixes the one attribute it emitted off-spec, and closes attribute gaps in the existing gen_ai suites.

**Retrieval attribute fix.** `tracesgen` emitted the collection top-k as a hand-rolled `gen_ai.retrieval.top_k`. That attribute does not exist in semantic conventions and collides with the `gen_ai.retrieval` namespace, so weaver rejected every retrieval span. The retrieval client span already defines `gen_ai.request.top_k` for this and the other gen_ai paths use it; this switches to it. Violations for the suite go from 240 to 0.

**New retrieval suite.** A small FastAPI client against a real Qdrant, aliased to `vectors.qdrant.tech` so the vendor host map resolves the provider rather than falling back to the generic heuristic. It exercises both `/points/search` and `/points/query`. Retrieval detection is off by default, so the OATS gen_ai config now enables it. Observed on a live run: `retrieval qdrant` (client kind), `gen_ai.provider.name=qdrant`, `gen_ai.operation.name=retrieval`, `gen_ai.request.top_k` carrying the real limit.

**Gaps in the existing suites.** These requests now set parameters that map to emitted attributes but were never exercised, so the attributes had no coverage: OpenAI gains choice count, seed, presence and frequency penalties, stop sequences, service tier and the embeddings encoding format; Anthropic sets `top_k`; Gemini sets a response MIME type, covering `gen_ai.output.type`; the Bedrock mock returns a guardrail id header, covering `aws.bedrock.guardrail.id`.

`gen_ai.data_source.id` remains unset for Qdrant, which keeps the collection in the URL path while the parser only reads it from the request body. Left for a follow-up.

## Validation


mmat11 (approved):
lgtm

grcevski (approved):
LGTM!

PR #3327, opened 2026-09-06
pkg/docker: avoid repeated container lookups for host processes

AlexanderVarfolomeev:
## Problem

I found this while instrumenting an HTTP application running directly on the host while Docker was available.

`DockerEnricher` resolved container metadata for every span. Successful lookups were cached, but negative lookups were not, causing repeated procfs scans for host processes.

The cgroup parser also interpreted `session-11909.scope` as container ID `11909`, triggering repeated Docker inspection instead of identifying a non-container process.

At 10000 requests/s, Docker enrichment became a bottleneck: CPU and memory usage grew, while RED metrics fell behind the generated traffic.

## Fix

- Require a 64-character hexadecimal container ID in systemd scope paths.
- Cache `ErrContainerNotFound` per PID for one second; retry other errors immediately.
- Clear negative entries on success, PID invalidation, or expiry on access.
- Revalidate the PID under the cache lock before recording a miss to preserve invalidation safety.
- Remove the redundant per-span Docker enrichment lookup-failed debug log.

The cooldown bounds repeated procfs work without treating a negative lookup as permanent. This complements the positive metadata cache introduced in [#1990](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1990).

## Results

Compared unpatched Beyla 3.32.0 with the same source plus this PR's production changes. Both were built with Go 1.26 and tested on the same four-vCPU host, with identical configuration and Docker enrichment enabled; the application ran directly on the host, not in Docker.

Each run lasted five minutes at a target of 10,000 requests/s: concurrency 800, 1 KiB request payload, 5% error responses and 10% requests delayed by 250 ms.

| | Unpatched | Patched |
| --- | ---: | ---: |
| Completed requests accounted for in RED | 43.13% | 100% |
| Beyla CPU, % of the four-vCPU host | 14.15% | 6.86% |
| Beyla RSS | 877.80 MiB | 113.53 MiB |

CPU/RSS are averages excluding the first 30 seconds: about **52% less CPU and 87% less memory** with the fix. 

After export settled, unpatched RED accounted for 1,221,598 of 2,832,134 completed requests; patched RED accounted for all 2,996,362. Unpatched counters were stable for over two minutes and the collector export queue was empty, so the difference was not merely an unfinished export. Neither run had transport errors, OOM events or supervisor restarts during load.

## Validation

Passed on Linux:

- `make fmt`
- `make verify`
- `make compile`
- `go test -race ./pkg/docker -run TestContainerInfo -count=100`

## AI assistance

Codex assisted with investigation, implementation, and tests. I reviewed and verified the final changes before submission.

### DIFF
--- a/pkg/docker/docker_api_client.go
+++ b/pkg/docker/docker_api_client.go
@@ -27,6 +27,8 @@ import (
 
 const (
 	composeServiceLabelKey = "com.docker.compose.service"
+	// Bound repeated procfs scans for non-container PIDs without caching a negative result for the process lifetime.
+	containerNotFoundRetryInterval = time.Second
 	// abbreviationLength defines the length for the short ID form
 	abbreviationLength = 12
 )
@@ -77,9 +79,10 @@ type ContainerStore struct {
 	watcherStarted sync.Once
 	watcherRunning atomic.Bool
 
-	cacheMu       sync.RWMutex
-	byPID         map[app.PID]ContainerMeta
-	byContainerID map[ContainerID]containerEntry // metadata + PIDs keyed by full container ID
+	cacheMu                sync.RWMutex
+	byPID                  map[app.PID]ContainerMeta
+	byContainerID          map[ContainerID]containerEntry // metadata + PIDs keyed by full container ID
+	notContainerUntilByPID map[app.PID]time.Time
 
 	// lastEventAt is the Unix timestamp (seconds) of the last processed Docker event.
 	// It is seeded to the start time of watchContainerEvents and updated on each event,
@@ -90,9 +93,10 @@ type ContainerStore struct {
 
 func NewStore() *ContainerStore {
 	return &ContainerStore{
-		log:           cmlog(),
-		byPID:         make(map[app.PID]ContainerMeta),
-		byContainerID: make(map[ContainerID]containerEntry),
+		log:                    cmlog(),
+		byPID:                  make(map[app.PID]ContainerMeta),
+		byContainerID:          make(map[ContainerID]containerEntry),
+		notContainerUntilByPID: make(map[app.PID]time.Time),
 	}
 }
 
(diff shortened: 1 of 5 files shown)


pkg/docker/docker_api_client.go
```diff
@@ -137,9 +141,36 @@ func (s *ContainerStore) ContainerInfo(ctx context.Context, pid app.PID) (Contai
...
 		return ci, true
 	}
+	retryAt, notContainer := s.notContainerUntilByPID[pid]
 	s.cacheMu.RUnlock()
+	if notContainer {
+		if time.Now().Before(retryAt) {
+			return ContainerMeta{}, false
+		}
+		s.cacheMu.Lock()
+		if retryAt, ok := s.notContainerUntilByPID[pid]; ok {
```
mariomac: Thank you for your PR!

What's the reason for looking up again the same map with the same PID?

AlexanderVarfolomeev: Thanks for reviewing! 
We release `RLock` before acquiring `Lock` to remove the expired entry. Another goroutine may refresh the cooldown in between, so we look it up again under the write lock to avoid deleting a fresh entry based on the previously read expiration time.

The second lookup only happens on the expiration path; valid cache hits return after the first lookup.

pkg/docker/docker_api_client.go
```diff
@@ -77,9 +79,10 @@ type ContainerStore struct {
 	watcherStarted sync.Once
 	watcherRunning atomic.Bool
 
-	cacheMu       sync.RWMutex
-	byPID         map[app.PID]ContainerMeta
-	byContainerID map[ContainerID]containerEntry // metadata + PIDs keyed by full container ID
+	cacheMu                sync.RWMutex
+	byPID                  map[app.PID]ContainerMeta
+	byContainerID          map[ContainerID]containerEntry // metadata + PIDs keyed by full container ID
+	notContainerUntilByPID map[app.PID]time.Time
```
mmat11: nit: can we give a clearer name to this map or add a little comment explaining what it contains?

mmat11 (approved):
lgtm

pkg/docker/docker_api_client.go
AlexanderVarfolomeev: Good point. Renamed it to `containerNotFoundRetryAtByPID` to clarify that it stores the next retry time after `ErrContainerNotFound`.

grcevski (approved):
LGTM!

MrAlias (approved):

PR #3328, opened 2026-09-06
test: fix jvm runtime OTLP test

NimrodAvni78:
## Summary

The JVM suite reported no JVM runtime metrics to weaver, because its assertions read a faster path than weaver does.

The suite asserted against OBI's Prometheus exporter, which exposes a metric as soon as it is sampled, while weaver is fed from the OTLP path that only emits on the next export interval. The test was satisfied and stopped weaver before that export ran, and the weaver tap disables retries, so the batch was discarded.

Pointing the suite's Prometheus at the collector's OTLP-fed endpoint fixes it: both pipelines share one OTLP receiver, so the existing assertions cannot pass until weaver's feed has the data. No test code changes. The suite gets its own scrape config because the shared one is used by three other suites, and the export interval moves to 1s to match the other weaver-wired suites.

The weaver report goes from zero JVM metric families to all eleven, with no violations.

## Validation

### DIFF
--- a/internal/test/integration/configs/prometheus-config-promscrape-jvm.yml
+++ b/internal/test/integration/configs/prometheus-config-promscrape-jvm.yml
@@ -1,0 +1,8 @@
+global:
+  evaluation_interval: 250ms
+  scrape_interval: 250ms
+scrape_configs:
+  - job_name: otelcol-otlp
+    static_configs:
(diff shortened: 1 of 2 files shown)


marctc (approved):
Nice, thanks!

PR #3329, opened 2026-09-06
add k8s coverage for netolly and statsolly

NimrodAvni78:
## Summary

Several network and TCP metrics reach the telemetry coverage check with almost no attributes, because nothing exercises them on a Kubernetes cluster with full attribute selection.

`obi.network.flow.bytes` is selected with `include: ["*"]` in the multizone suite and comes out fully decorated. Its siblings on the same cluster fall back to the default attribute set: `obi.network.inter.zone.bytes` and `obi.network.flow.packets` report three attributes each, and neither carries the `k8s.src.*` / `k8s.dst.*` decoration that the sibling proves is available. This adds both to the same `select` block, and enables `network_flow_packets`, which no Kubernetes manifest turned on.

The `obi.stat.tcp.*` metrics had a different problem: the stats feature is enabled in no Kubernetes manifest at all, so they only ever emitted from Docker Compose suites where Kubernetes decoration cannot exist. This adds a `tcp_stats` suite: a weaver-tapped kind cluster with the stats feature on, a pinger generating TCP traffic, and assertions that the IO and round-trip-time metrics carry the source and destination Kubernetes metadata. The suite directory is picked up by the existing test matrix automatically.

Together these target the attribute gaps on six metrics. No production code changes: this is test configuration and a new suite.

## Validation

### DIFF
--- a/internal/test/integration/k8s/manifests/06-obi-tcp-stats.yml
+++ b/internal/test/integration/k8s/manifests/06-obi-tcp-stats.yml
@@ -6,36 +6,25 @@ data:
   obi-config.yml: |
     log_level: debug
     metrics:
-      features: ["network", "network_inter_zone", "network_flow_packets"]
+      features: ["stats"]
     otel_metrics_export:
       endpoint: http://otelcol.default:4317
-    network:
-      guess_ports: ordinal
-      protocols:
-        - TCP
-      cidrs:
-        # default subnets of Kind Pods and services
-        - 10.244.0.0/16
-        - fd00:10:244::/56
-        - 10.96.0.0/16
-        - fd00:10:96::/112
     attributes:
       kubernetes:
         enable: true
         cluster_name: my-kube
         resource_labels:
-          deployment.environment.name: ["${DEPLOYMENT_ENVIRONMENT:-error-not-replaced}"]
+          deployment.environment.name: ["${DEPLOYMENT_ENVIRONMENT:-deployment.environment.name}"]
       select:
-        obi.network.flow.bytes:
+        obi.stat.tcp.rtt:
           # assured cardinality explosion. Don't try in production!
           include: ["*"]
-          exclude: ["src_port"]
-        obi.network.inter.zone.bytes:
+        obi.stat.tcp.io:
           include: ["*"]
-          exclude: ["src_port"]
-        obi.network.flow.packets:
+        obi.stat.tcp.retransmits:
           include: ["*"]
-          exclude: ["src_port"]
+        obi.stat.tcp.failed.connections:
+          include: ["*"]
 ---
 apiVersion: apps/v1
 kind: DaemonSet
@@ -64,6 +53,13 @@ spec:
         - name: testoutput
           persistentVolumeClaim:
             claimName: testoutput
+        # the stats tracer attaches sock/inet_sock_set_state, which needs tracefs
+        - name: tracefs
+          hostPath:
+            path: /sys/kernel/tracing
+        - name: debugfs
+          hostPath:
+            path: /sys/kernel/debug
       containers:
         - name: obi
           image: obi:dev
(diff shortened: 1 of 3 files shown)


internal/test/integration/k8s/tcp_stats/k8s_tcp_stats_metrics.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	return features.New("tcp stats").
+		Setup(pinger.Deploy()).
+		Teardown(pinger.Delete()).
+		Assess("emits tcp stat metrics", testTCPStatsEmitted).
+		Assess("decorates tcp io metrics with kubernetes metadata", testTCPStatsIODecoration).
+		Feature()
+}
+
+func testTCPStatsEmitted(ctx context.Context, t *testing.T, _ *envconf.Config) context.Context {
```
pinoOgni, line 47: We are asserting only tcp io metric and not the rest. 

Just one note about the tcp rtt metric, it is emitted at `tcp_close` and I see that the `httppinger` is using `http.DefaultClient` keep-alive so the `pinger->testserver` connection closes only at pinger teardown. I think that to correctly assert it we need a client that closes connections.

mariomac (approved):
Good catch!

internal/test/integration/k8s/manifests/06-obi-netolly-multizone.yml
```diff
@@ -30,6 +30,12 @@ data:
           # assured cardinality explosion. Don't try in production!
           include: ["*"]
           exclude: ["src_port"]
+        obi.network.inter.zone.bytes:
```
pinoOgni, line 33: I checked the related integration test and I found that in `netolly_multizone_feature.go` we are doing this:
```
		// AND the reported attributes are different from the flow bytes attributes
		require.NotContains(ct, results, "k8s_src_type")
``` 
but now with `include: ["*"]` the attribute `k8s_src_type` will be present and the assertion passes only because it is vacuous (results is `[]Result` and checked against a string). So, we should replace with a real label check on `results[0].Metric`.

internal/test/integration/k8s/manifests/06-obi-tcp-stats.yml
```diff
@@ -0,0 +1,89 @@
...
+          configMap:
+            name: obi-config
+        - name: testoutput
+          persistentVolumeClaim:
+            claimName: testoutput
+        # the stats tracer attaches sock/inet_sock_set_state, which needs tracefs
+        - name: tracefs
+          hostPath:
+            path: /sys/kernel/tracing
+        - name: debugfs
```
pinoOgni, line 60: This is redundant

internal/test/integration/k8s/tcp_stats/k8s_tcp_stats_metrics.go
```diff
@@ -0,0 +1,85 @@
...
+// of an arbitrary member would sample unrelated flows — node-sourced ones that
+// legitimately have no namespace, or traffic to the CNI gateway that has no
+// destination workload.
+func testTCPStatsIODecoration(ctx context.Context, t *testing.T, _ *envconf.Config) context.Context {
+	pq := promtest.Client{HostPort: prometheusHostPort}
+	require.EventuallyWithT(t, func(ct *assert.CollectT) {
+		results, err := pq.Query(`obi_stat_tcp_io_bytes_total{` +
+			`k8s_cluster_name="my-kube",` +
+			`k8s_src_name="` + pingerPodName + `",` +
+			`k8s_dst_name=~"testserver.*"` +
```
pinoOgni, line 69: We should add also `k8s_dst_owner_name == "testserver"`, `k8s_dst_owner_type == "Service"`.

pinoOgni (commented):
The only metric currently being tested is TCP IO. We also need to add TCP RTT, TCP failed connections, and TCP retransmits. 
Perhaps it would be best to split this into two PRs? One for `netolly` and one for `statsolly`, where we test all the metrics?

PR #3330, opened 2026-09-06
Fix metrics interval in suites to 1s

NimrodAvni78:
## Summary

The six suites here exported metrics on a 10ms interval. That is far shorter than anything the assertions need, and it produces a large volume of duplicate datapoints through the weaver tap, whose queue is deliberately non-blocking and drops on overflow. Moving them to 1s matches the interval the other weaver-wired suites already use.

## Validation


.github/workflows/workflow_integration_tests_vm.yml
```diff
@@ -363,6 +363,15 @@ jobs:
...
           name: vm-coverage-${{ matrix.kernel.id }}-${{ matrix.test.id }}-${{ github.run_number }}
 
+      - name: Upload weaver live-check reports
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        if: always()
+        with:
+          name: weaver-reports-vm-${{ matrix.kernel.id }}-${{ matrix.test.id }}-${{ github.run_number }}
+          path: testoutput/weaver-report-*.json
+          if-no-files-found: ignore
+          retention-days: 5
```
pinoOgni: nit: shouldn't we add `retention-days: 5` to the other steps as well?

pinoOgni (commented):
lgtm! I left one nit and also I have a question. In the PR description is written 
> The Docker Compose, OATS and Kubernetes jobs already upload theirs

I checked but I didnt see any weaver-report upload step in `pull_request_integration_tests.yml`, `pull_request_oats_test.yml`, or `pull_request_k8s_integration_tests.yml`.

Can we add the same step to the three other workflows?

NimrodAvni78:
@pinoOgni you are right
its not correct to be here and it should be part of the follow up work for weaver coverage. ill keep here only the metric interval fixes

pinoOgni (approved):

PR #3338, opened 2026-09-07
test: add netolly k8s assertions

NimrodAvni78:
## Summary

The two attribute assertions on the inter-zone metric never tested anything: `NotContains` was given the results slice rather than a series' label map, so it compared a struct against a string and could not fail.

Their premise is also no longer true. Selecting the inter-zone metric with every attribute means it does now carry `k8s_src_type` and `iface_direction`, so this asserts they are present, on every returned series, and drops the stale comment.

Also converts `require` to `assert` inside this file's `EventuallyWithT` callbacks. `require` on a `*assert.CollectT` calls `FailNow`, which is not what a polling condition wants.

## Validation


NimrodAvni78:
closed in favor of https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3343

PR #3339, opened 2026-09-07
test: add statsolly k8s metric assertions

NimrodAvni78:
## Summary

The suite asserted only TCP IO. This adds RTT and failed connections, and asserts the destination workload on the IO decoration.

RTT is emitted at `tcp_close`, and the pinger used a keep-alive client, so its connection only closed at pod teardown and the metric never appeared during the run. The pinger takes an opt-in `DISABLE_KEEPALIVES`, which this suite sets; the uninstrumented pinger template gains the `Env` block the instrumented ones already render.

Failed connections come from a second pinger dialing a closed port on its own loopback, so the refusal comes from the kernel rather than from kube-proxy or a SYN timeout. That series carries no Kubernetes labels, so it is pinned on port, role and reason instead.

Retransmits are not asserted. A kind cluster on veth drops nothing, and forcing loss needs netem inside the pinger's network namespace, which would reshape the pinger template every other suite shares. A note in the file records why one of the four is missing.

Also drops the `debugfs` mount, which is redundant: the stats tracer needs `tracefs`, which stays.

## Validation

### DIFF
--- a/internal/test/integration/k8s/tcp_stats/k8s_tcp_stats_metrics.go
+++ b/internal/test/integration/k8s/tcp_stats/k8s_tcp_stats_metrics.go
@@ -22,6 +22,8 @@ const (
 	testTimeout        = 5 * time.Minute
 	prometheusHostPort = "localhost:39090"
 	pingerPodName      = "internal-pinger-tcp-stats"
+	failPingerPodName  = "internal-pinger-tcp-stats-fail"
+	closedPort         = "19999"
 
 	// A cardinality-exploded metric makes each Prometheus query cost hundreds of
 	// milliseconds, so a sub-second poll only piles up in-flight requests.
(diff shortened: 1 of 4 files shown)


internal/test/integration/k8s/tcp_stats/k8s_tcp_stats_metrics.go
```diff
@@ -78,8 +90,59 @@ func testTCPStatsIODecoration(ctx context.Context, t *testing.T, _ *envconf.Conf
...
 			assert.NotEmpty(ct, metric["k8s_dst_namespace"])
+			assert.Equal(ct, "testserver", metric["k8s_dst_owner_name"])
+			assert.Equal(ct, "Service", metric["k8s_dst_owner_type"])
 			assert.Contains(ct, []string{"receive", "transmit"}, metric["network_io_direction"])
 		}
 	}, testTimeout, pollInterval)
 	return ctx
 }
+
+func testTCPStatsRTT(ctx context.Context, t *testing.T, _ *envconf.Config) context.Context {
```
pinoOgni, line 94: Can we use here the same pattern used in `testTCPStatsIODecoration` because `pingerFlow` selector already requires `k8s_cluster_name`, `k8s_src_name` and `k8s_dst_name` and we are testing it here but in the TCP IO test it loops results and it checks all labels. 
So if TCP IO metric loses for example`k8s_src_owner_type` it will catch it while if TCP RTT metric loses it the TCP RTT test will stay green.

pinoOgni (changes requested):

pinoOgni (approved):
lgtm!

PR #3340, opened 2026-09-07
fix(mysql): recognize custom server error codes

loveRhythm1990:
## Summary

While using OBI to monitor MatrixOne's MySQL-protocol traffic, we found that some failed requests were not recognized as errors. MatrixOne can return custom error codes such as 20301, but OBI's MySQL error parser only accepts codes in the 1002–4167 range and discards these error responses.

This is not specific to MatrixOne: MySQL also supports user-defined MYSQL_ERRNO values through [SIGNAL](https://dev.mysql.com/doc/refman/8.4/en/signal.html). A fixed range of built-in error codes therefore excludes valid errors.

This change removes that range check while retaining packet marker/length checks and the exclusion of MariaDB progress reports. It also fixes the length check for a 13-byte ERR packet containing SQLSTATE and an empty message.

The existing SQLState prefix is deliberately unchanged. It feeds the emitted error.type attribute; normalizing it could affect existing queries, alerts, and aggregations and needs a separate compatibility review.

## Validation

- [x] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed. No updates needed: this fixes parsing within existing MySQL support, without adding a protocol or configuration option.

Regression tests cover custom codes, legacy and SQLSTATE-bearing packets, invalid markers, truncation, MariaDB progress reports, and empty messages. The affected cases fail with the old logic and pass with the fix.

Passed locally (Go 1.26.2, GOTOOLCHAIN=local):
- `go test -p 2 -mod=readonly ./pkg/internal/sqlprune -count=1`
- `go vet -p 2 -mod=readonly ./pkg/internal/sqlprune`
- `go tool -modfile=internal/tools/go.mod golangci-lint run ./pkg/internal/sqlprune --timeout=6m`

Validation was scoped to the affected package; no cluster end-to-end test was run for this patch.

## AI assistance

Codex assisted with the implementation and tests. I reviewed the code and regression tests and understand the error-code validation and empty-message boundary fixes. Codex also assisted with translating and preparing this description.

### DIFF
--- a/pkg/internal/sqlprune/mysql.go
+++ b/pkg/internal/sqlprune/mysql.go
@@ -57,25 +57,24 @@ func parseMySQLError(buf []uint8) *request.SQLError {
 	sqlErr.Code = binary.LittleEndian.Uint16(buf[offset : offset+2])
 	offset += 2
 
-	// https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html
-	if sqlErr.Code < 1002 || sqlErr.Code > 4167 {
-		return nil // Invalid error code
+	// MariaDB progress reports share the ERR marker but are not errors.
+	// Other codes may be user-defined or supplied by MySQL-compatible servers.
+	if sqlErr.Code == MySQLProgressReporting {
+		return nil
 	}
 
-	if sqlErr.Code != MySQLProgressReporting {
-		if buf[offset] == MySQLStateMarker {
-			if len(buf) < (MySQLErrMinLen + 6) {
-				return nil
-			}
-			// Skip the SQL state marker
-			offset++
-			// Read the SQL state
-			sqlErr.SQLState = string(MySQLStateMarker) + string(buf[offset:offset+5])
-			offset += 5
+	if buf[offset] == MySQLStateMarker {
+		if len(buf) < (MySQLErrMinLen + 6) {
+			return nil
 		}
-		// Read the error message
-		sqlErr.Message = unix.ByteSliceToString(buf[offset:])
+		// Skip the SQL state marker
+		offset++
+		// Read the SQL state
+		sqlErr.SQLState = string(MySQLStateMarker) + string(buf[offset:offset+5])
+		offset += 5
 	}
+	// Read the error message
+	sqlErr.Message = unix.ByteSliceToString(buf[offset:])
 
 	return &sqlErr
 }
--- a/pkg/internal/sqlprune/mysql.go
+++ b/pkg/internal/sqlprune/mysql.go
@@ -64,12 +64,13 @@ func parseMySQLError(buf []uint8) *request.SQLError {
 	}
 
 	if buf[offset] == MySQLStateMarker {
-		if len(buf) < (MySQLErrMinLen + 6) {
+		if length < offset+1+5 {
 			return nil
 		}
 		// Skip the SQL state marker
 		offset++
 		// Read the SQL state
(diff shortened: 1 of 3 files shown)


pkg/internal/sqlprune/mysql.go
```diff
@@ -57,25 +57,25 @@ func parseMySQLError(buf []uint8) *request.SQLError {
...
+	if buf[offset] == MySQLStateMarker {
+		if length < offset+1+5 {
+			return nil
 		}
-		// Read the error message
-		sqlErr.Message = unix.ByteSliceToString(buf[offset:])
+		// Skip the SQL state marker
+		offset++
+		// Read the SQL state
+		// TODO: Normalize SQLState to five characters without changing error.type silently.
```
mmat11: is this something that needs to be addressed here?

pkg/internal/sqlprune/mysql.go
```diff
@@ -57,25 +57,25 @@ func parseMySQLError(buf []uint8) *request.SQLError {
...
-	// https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html
-	if sqlErr.Code < 1002 || sqlErr.Code > 4167 {
-		return nil // Invalid error code
+	// MariaDB progress reports share the ERR marker but are not errors.
+	// Other codes may be user-defined or supplied by MySQL-compatible servers.
+	if sqlErr.Code == MySQLProgressReporting {
+		return nil
 	}
 
-	if sqlErr.Code != MySQLProgressReporting {
```
mmat11, line 65: why was this piece refactored?

pkg/internal/sqlprune/mysql.go
```diff
@@ -57,25 +57,25 @@ func parseMySQLError(buf []uint8) *request.SQLError {
 	sqlErr.Code = binary.LittleEndian.Uint16(buf[offset : offset+2])
 	offset += 2
 
-	// https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html
-	if sqlErr.Code < 1002 || sqlErr.Code > 4167 {
```
mmat11, line 61: I think we should still keep some validation here, eg. can a custom error code be zero?

pkg/internal/sqlprune/mysql.go
loveRhythm1990: this is outside the scope of this PR. I’ll remove the TODO and keep the existing SQLState format unchanged.
sorry for this.

pkg/internal/sqlprune/mysql.go
loveRhythm1990: good point, I will reject zero and add tests. the code is decoded as uint16, which already limits it to 65535. existing packet validation and special-case handling will remain unchanged

pkg/internal/sqlprune/mysql.go
loveRhythm1990: removing the range check allows progress reports to reach this branch. without the early return, they would fall through and return a non-nil SQLError. This preserves the previous behavior of ignoring them.

grcevski (approved):
LGTM!

mmat11 (approved):

PR #3341, opened 2026-09-07
fix(sql): disambiguate MySQL and PostgreSQL fallback detection

loveRhythm1990:
## Summary

While using OBI to monitor MatrixOne's MySQL-protocol traffic, we found that some requests were misidentified as PostgreSQL. MySQL QUERY, PREPARE, and EXECUTE packets of certain lengths match both protocols' header heuristics, and the existing detection order prefers PostgreSQL.

This change corrects the minimum PostgreSQL message length and, when both headers match, validates the structure of the first complete PostgreSQL message rather than simply reversing detection order. Validation stays within that message's declared boundary, allowing subsequent pipelined messages in the same buffer. These changes address the same classification bug: the length check alone does not resolve PREPARE/EXECUTE collisions.

For truncated captures that cannot be disambiguated, the existing PostgreSQL classification is retained. Returning Generic would suppress SQL extraction with heuristic detection disabled by default. This preserves existing behavior but does not eliminate all protocol misclassification for truncated packets.

The change affects generic TCP fallback detection, not dispatch for protocols already classified by the kernel. Bind format-code reading is shared with the extraction path; invalid formats are rejected, while extraction still tolerates truncated parameter values. No new connection cache or configuration is introduced.

## Validation

- [x] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed. No updates needed: this corrects fallback classification within existing protocol support.

Regression tests cover:
- MySQL QUERY, PREPARE, and EXECUTE length collisions, including the six PREPARE/EXECUTE cases with payload lengths 66, 67, and 81.
- Reverse-direction PostgreSQL Q/C/B header collisions, pipelined and truncated captures, Bind structure, and frontend Close messages.
- Span generation for truncated queries with heuristic detection disabled. Restoring the Generic downgrade makes all five affected span-level cases fail.

Passed locally (Go 1.26.2, GOTOOLCHAIN=local):
- `go test -p 2 -mod=readonly ./pkg/ebpf/common -count=1 -timeout=120s`
- `go tool -modfile=internal/tools/go.mod golangci-lint run ./pkg/ebpf/common --timeout=6m`
- `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -p 2 -mod=readonly ./pkg/ebpf/common`

Validation was scoped to the affected package; no cluster end-to-end test was run for this patch.

## AI assistance

Codex assisted with the implementation and tests. I reviewed the code and regression tests and understand the protocol disambiguation approach and the truncated-capture trade-off. Codex also assisted with translating and preparing this description.

### DIFF
--- a/pkg/ebpf/common/sql_detect_transform.go
+++ b/pkg/ebpf/common/sql_detect_transform.go
@@ -13,14 +13,11 @@ import (
 func sqlKind(b *largebuf.LargeBuffer) request.SQLKind {
 	if isPostgres(b) {
 		if isMySQL(b) {
-			size, _ := b.I32BEAt(1)
-			if int(size)+1 > b.Len() {
-				// A truncated, ambiguous header cannot establish either protocol.
-				return request.DBGeneric
-			}
-			if !validPostgresSQLBody(b) {
+			if postgresSQLBodyStatus(b) == postgresBodyInvalid {
 				return request.DBMySQL
 			}
+			// Retain the existing PG fallback for incomplete captures. Generic
+			// would suppress SQL extraction when heuristic detection is disabled.
 		}
 		return request.DBPostgres
 	}
--- a/pkg/ebpf/common/sql_detect_transform.go
+++ b/pkg/ebpf/common/sql_detect_transform.go
@@ -12,6 +12,16 @@ import (
 
 func sqlKind(b *largebuf.LargeBuffer) request.SQLKind {
 	if isPostgres(b) {
+		if isMySQL(b) {
+			size, _ := b.I32BEAt(1)
+			if int(size)+1 > b.Len() {
+				// A truncated, ambiguous header cannot establish either protocol.
+				return request.DBGeneric
+			}
+			if !validPostgresSQLBody(b) {
+				return request.DBMySQL
+			}
+		}
 		return request.DBPostgres
 	}
 	if isMySQL(b) {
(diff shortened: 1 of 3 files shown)


pkg/ebpf/common/sql_detect_transform.go
```diff
@@ -12,6 +12,13 @@ import (
 
 func sqlKind(b *largebuf.LargeBuffer) request.SQLKind {
 	if isPostgres(b) {
+		if isMySQL(b) {
+			if postgresSQLBodyStatus(b) == postgresBodyInvalid {
```
mariomac: Nit:
```go
if isMySQL(B) && postgresSQLBodyStatus(b) == postgresBodyInvalid {
  return request.DBMySQL
}
// Retain the existing PG fallback for incomplete captures. Generic
// would suppress SQL extraction when heuristic detection is disabled.
```
maybe?

mariomac (commented):
Thanks for your PR! Triggering test run before a deeper review

pkg/ebpf/common/sql_detect_transform.go
loveRhythm1990: good suggestion, I will fix it.

mmat11 (approved):
lgtm

grcevski (approved):
LGTM!

MrAlias (approved):

PR #3343, opened 2026-09-07
Fix vacuous inter-zone attribute assertion in netolly multizone suite

pinoOgni:
## Summary

This PR fixes a broken assertion in `testInterZoneMetric` found in #3329.

Basically the test asserted that `k8s_src_type` and `iface_direction` were absent from `obi.network.inter.zone.bytes`. The check ran `NotContains` on a `[]Result` slice against a `string`, so it never inspected labels and always passed.

#3329 selected `include: ["*"]` for that metric, so both attributes are now present and the original intent no longer holds. The check is replaced with a positive one.

## Validation


NimrodAvni78 (approved):
NIce

PR #3351, opened 2026-09-08
Add MCP metrics

NimrodAvni78:
## Summary

MCP calls are detected in both directions but land in no MCP metric. Since MCP left `IsGenAISubtype`, an MCP span falls through to `http.client.request.duration` or `http.server.request.duration`, so MCP traffic is indistinguishable from plain HTTP.

Adds the two operation duration metrics semantic conventions define: `mcp.client.operation.duration` for outbound MCP spans, `mcp.server.operation.duration` for inbound ones. Both are `s` histograms on the standard duration buckets rather than the GenAI ones, which are tuned for model latencies.

The session duration metrics are left out: they need session lifetime tracking from `initialize` to teardown, which OBI does not model.

## MCP traffic moves off the HTTP metrics

This is a breaking change for anyone already monitoring an MCP endpoint, which is a narrow group: MCP payload extraction is opt-in and off by default, so only operators who set `OTEL_EBPF_HTTP_MCP_ENABLED=true` produce MCP spans at all. For those, `instrumentations` defaults to `["*"]`, so the new branches take precedence out of the box and an MCP span stops being recorded as:

- `http.server.request.duration` / `http.client.request.duration`, replaced by the MCP metrics above
- `http.server.request.body.size`, `http.server.response.body.size` and the two client equivalents, with **no** replacement, since semconv defines no size instrument for MCP

Dashboards, alerts and SLOs built on the HTTP metrics for an MCP route lose their data on upgrade and need to move to the MCP metrics. Dropping `genai` from `instrumentations` keeps the old behaviour, but it is not a free opt-out: it also silences `gen_ai.client.operation.duration` and `gen_ai.client.token.usage` for every other provider. There is no knob that keeps MCP on the HTTP metrics while retaining the GenAI ones, which matches how JSON-RPC already moves to the RPC metrics under `grpc`.

The Prometheus exporter records body sizes in `observeBodySizes`, outside the duration switch, so dropping them takes its own carve-out on each side: MCP joins the JSON-RPC exemption for server spans and `recordedAsNonHTTPClient` for client ones. Without it the two exporters disagree, and the Prometheus side keeps publishing sizes for a span whose duration has already moved.

## Attributes

Labels are the subset of upstream's `mcp.operation.metrics.attributes` that OBI can actually populate: `mcp.method.name`, `mcp.protocol.version`, `gen_ai.tool.name`, `gen_ai.prompt.name`, `error.type` and `rpc.response.status_code`, plus `server.address` and `server.port` on the client metric only, as upstream declares them. `mcp.resource.uri` is declared but off by default, being opt-in upstream and unbounded; `mcp.session.id` and `jsonrpc.request.id` stay off entirely.

The four remaining upstream attributes are absent rather than declared off. `network.transport`, `network.protocol.name` and `network.protocol.version` have no metric getter in OBI at all. `jsonrpc.protocol.version` has one, but it is gated on the JSON-RPC subtype and returns an empty value for an MCP span, which the registry's omit-rather-than-emit-empty rule forbids. MCP does pin the version — a body that is not JSON-RPC 2.0 is not parsed as MCP — but semconv only recommends the attribute when the version is not 2.0, so omitting it is right either way.

`gen_ai.operation.name` is a deliberate omission. Upstream permits it, but its only MCP value is `execute_tool`, which duplicates `mcp.method.name=tools/call`.

Five getters are new, and `rpc.response.status_code` learns about MCP error codes, which it handled for JSON-RPC only, so an MCP JSON-RPC error code now reaches the metric. The Prometheus getters delegate to the OTEL ones, so they need no second implementation. The repeated `SubType == HTTPSubtypeMCP && GenAI != nil && GenAI.MCP != nil` guard is now a single `Span.MCP()` accessor, reused at the five pre-existing sites as well as the new getters.

No registry change: upstream declares both metrics and OBI's manifest depends on that registry, so they resolve the way `db.client.operation.duration` already does.

## Validation

Neither document records which metric a protocol feeds, so there is nothing there to update; the metric move is described above instead.

Exporter tests cover both directions in both exporters with HTTP instrumentation enabled, so the assertion that the span no longer reaches the HTTP duration and body size metrics is meaningful; the OTLP test drains a full export cycle past the first match before asserting, since record order within a cycle is not guaranteed. `TestSpanOTELGetters_MCPAttributes` covers each new getter against a non-MCP span, an MCP subtype with no parsed call, and the empty-field case — `mcp.protocol.version` in particular is set only by `initialize`, so absence is the common shape rather than an edge case. The MCP integration suite gains a Prometheus assertion for both histograms, since the existing tests only read spans back from Jaeger and would pass whether or not the metrics were emitted.


mmat11 (approved):
lgtm

grcevski (approved):
LGTM!

PR #3354, opened 2026-09-08
gotracer: clear parent_id when a client call has no parent request

NimrodAvni78:
## Summary

The Go Redis client passes per-CPU scratch memory to `setup_request` and never zeroes it, so `err`, `conn` and `tp.parent_id` are inherited from whatever last used that CPU's slot. The whole struct is copied into the ring buffer, so a successful command can report a non-zero status, carry an unrelated peer address, or point at a parent from an unrelated trace that is never sent, leaving the trace rootless. `setup_request` now zeroes the struct.

`client_trace_parent` also left two fields unwritten. It never wrote `parent_id`, which the six callers that pass a zeroed stack struct did not notice; it is now cleared explicitly so the helper writes every field it does not copy from a real parent. It also skipped `span_id` when the parent came from the SQL wrapper path, emitting a duplicate span id there; `span_id` is now generated on every path.

## Validation

### DIFF
--- a/bpf/gotracer/go_common.h
+++ b/bpf/gotracer/go_common.h
@@ -351,10 +351,11 @@ static __always_inline u8 client_trace_parent(void *goroutine_addr, tp_info_t *t
 
         if (!found_trace_id) {
             urand_bytes(tp_i->trace_id, TRACE_ID_SIZE_BYTES);
+            __builtin_memset(tp_i->parent_id, 0, sizeof(tp_i->parent_id));
         }
-
-        urand_bytes(tp_i->span_id, SPAN_ID_SIZE_BYTES);
     }
+
+    urand_bytes(tp_i->span_id, SPAN_ID_SIZE_BYTES);
 
     return found_trace_id;
 }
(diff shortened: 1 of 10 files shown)


bpf/tests/bpf_go_client_trace_parent.c
```diff
@@ -0,0 +1,123 @@
...
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <bpfcore/vmlinux.h>
+#include <bpfcore/bpf_helpers.h>
+
+// Called by the go_common.h include chain, omitted by the shared stub
+#define BPF_ANY 0
```
pinoOgni: nit: move this define in `bpf/tests/bpfcore/bpf_helpers.h` because it is also defined in `bpf_go_h2_write_transaction.c`.

bpf/gotracer/go_common.h
```diff
@@ -351,11 +351,12 @@ static __always_inline u8 client_trace_parent(void *goroutine_addr, tp_info_t *t
 
         if (!found_trace_id) {
             urand_bytes(tp_i->trace_id, TRACE_ID_SIZE_BYTES);
+            __builtin_memset(tp_i->parent_id, 0, sizeof(tp_i->parent_id));
```
pinoOgni: nit: use `bpf_memset`

pinoOgni (approved):

mariomac (approved):
Looks good!

PR #3357, opened 2026-09-08
nodejs: improve performance of nodejs events + uninstall the agent on shutdown

NimrodAvni78:
## Summary

The injected Node.js agent signals the eBPF layer with `fs.accessSync` on a sentinel path that can never resolve, so every signal builds and throws a `UVException`. The exception costs more latency then the syscall.

Measured per call on Node 18, 20, 22 and 24: throwing `accessSync` 3210-3604ns, against 609-628ns for `fs.existsSync`. On a three-hop async middleware chain at 50 concurrent requests that is 20.17us/req of agent overhead against 4.86us/req.

`fs.existsSync` reaches the same `uv_fs_access` the `obi_uv_fs_access` uprobe is attached to — `binding.existsSync` on Node 18 and newer, the non-throwing `binding.access(path, F_OK, undefined, ctx)` form on 14 and 16 — so all five sentinel formats keep working and there is no eBPF change in this PR.

The second change gives the agent a way off a running process. `fdextractor.js` already restores the net prototypes it wrapped and clears the async hook, sampling timer, delay histogram and GC observer in its prologue, so evaluating it with its gates off is a complete uninstall. `spanbridge.js` now does the same: it records restorers for `Module._load` and the api setters it wraps, and gains an `OBI_SPANS_ENABLED` gate so the ungated script is its own uninstall. The injector tracks what it injected and runs that pass on shutdown, bounded to 15s across at most four processes at a time. This adds no configuration: uninstalling is the counterpart of an injection the operator already opted into, and `nodejs.enabled` stays the switch for leaving applications untouched.

Reopening the inspector needs another SIGUSR1, which terminates a process that has no handler for it, so the uninstall reuses the gates already in place rather than adding its own. The process is reopened through `procs.OpenProcessHandle` against the start time captured at injection, so a recycled PID is refused and the signal goes out over the pinned handle; `awaitSignalDisposition` then confirms the runtime still handles SIGUSR1 before it is sent. The queue closes its handle once an injection returns, so `InjectionTarget` carries that start time for the reopen.

One thing no uninstall can undo is a delegate already cached by a ProxyTracer, since the api caches the first one and never re-consults the registry. It is inert once the bridge stops emitting.

## Validation


NimrodAvni78:
Split up the performance improvements from agent uninstalls, both will be open in separate prs

PR #3359, opened 2026-09-08
Keep SQS sends as client spans without context injection

MrAlias:
## Summary

#3314 classified `SendMessage` and `SendMessageBatch` spans as `PRODUCER` through the shared messaging-operation mapping. The messaging [semantic conventions require `CLIENT` when the send span's context is not used as the message creation context](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#span-kind).

OBI currently enriches the completed SQS HTTP exchange but does not inject the span context into SQS `MessageAttributes`. This restores `CLIENT` for all observed SQS operations in both trace export and metric-facing kind labels. If OBI adds SQS message-level propagation later, producer classification can be enabled from an explicit creation-context signal.

## Compatibility

The `span.kind` label on span metrics, the `kind` label on service-graph metrics, and the JSON span `kind` return to `SPAN_KIND_CLIENT` for SQS sends. Consumers that changed filters after #3314 should keep using the client series.

## Validation

- `make verify`
- `make compile`
- `go test ./pkg/appolly/app/request ./pkg/export/otel/tracesgen ./pkg/ebpf/common/http`

### DIFF
--- a/internal/test/integration/red_test_python_aws.go
+++ b/internal/test/integration/red_test_python_aws.go
@@ -47,7 +47,7 @@ func waitAWSProxy(t *testing.T) {
 	}, testTimeout, 1*time.Second)
 }
 
-func fetchAWSSpanByOP(t require.TestingT, op, spanKind string) jaeger.Span {
+func fetchAWSSpanByOP(t require.TestingT, op string) jaeger.Span {
 	var tq jaeger.TracesQuery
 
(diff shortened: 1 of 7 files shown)


NimrodAvni78 (approved):
Good catch, i disregarded it

grcevski (approved):
LGTM!

PR #3363, opened 2026-09-08
fix(nodejs): require positive Node identification before sending SIGUSR1

NimrodAvni78:
## Summary

OBI decides a process is Node.js from its binary's name, then sends SIGUSR1 to open the inspector. SIGUSR1 kills anything that does not catch it, so Bun installed as `node` is terminated, as is a real Node that has not yet installed its own handler. The guard against that sat behind `if elfFile != nil` and was skipped whenever the ELF was missing.

Three checks now gate the signal:

- **Identification** — the executable names a Node-only symbol (`node::NodeMainInstance`, `node::Environment`, `node::Start`, or libuv's statically linked `uv__signal_tree`), or the process maps `libnode.so`. Not the N-API exports: Bun re-exports `node::MakeCallback`.
- **Disposition** — `SigCgt`/`SigIgn` in `/proc/<pid>/status` show SIGUSR1 caught or ignored. Node installs its handler ~11ms after exec (20 of 20 launches killable at first read, 9.6–13.6ms), so the gate polls for up to 500ms before refusing.
- **libuv signal tree** — unchanged, but it now always runs: the injector opens its own view of the executable rather than reusing the discovery loop's.

Injection moves off the discovery loop onto a 100-slot queue. The Java injector was queued for the same reason, so both now share one generic `injectionQueue[T]` and one injection slot — a JVM attach switches OBI's euid/egid process-wide, which would cost a concurrent Node injection `/proc/<pid>/mem` and the target's netns. Queued targets carry a `procs.ProcessHandle`, so a recycled PID cannot be signalled in the original's place.

Refusal reasons, the queue semantics and the residual gaps are in `devdocs/runtimes/nodejs.md`.

**Not in this change:** when the libuv tree is unreadable — distributions ship Node stripped — the fallback is scanning the application's source. That is untouched here and still fail-open; making it fail closed follows separately.

## Verified

Identification symbols are present in node 18, 20, 22 and 24 across glibc and musl, ET_EXEC and PIE, and absent from bun and deno. Bun copied to `/usr/local/bin/node` is refused; today it is signalled. Injecting into a containerized Node still opens the inspector and attaches the agent.

## Validation


pkg/appolly/discover/attacher.go
```diff
@@ -132,21 +132,44 @@ func (ta *traceAttacher) attacherLoop(_ context.Context) (swarm.RunFunc, error)
...
+
+		var javaInjections *injectionQueue[javaagent.InjectionTarget]
 		if ta.javaInjector != nil {
-			javaInjections = newJavaInjectionQueue(ta.log, ta.javaInjector.NewExecutable)
+			javaInjections = newJavaInjectionQueue(ta.log, injectionSlot, ta.javaInjector.NewExecutable)
 			javaInjections.start(ctx)
 			defer javaInjections.wait()
 		}
 
+		var nodeInjections *injectionQueue[nodejs.InjectionTarget]
```
grcevski, line 147: Nice! OK, just to confirm this makes the node injection non-blocking for the discovery pipeline too.

pkg/internal/nodejs/signal_check.go
```diff
@@ -31,6 +37,67 @@ const (
 	maxSignalNum = 64
 )
 
+const (
+	signalTreeSymbol = "uv__signal_tree"
```
grcevski, line 41: Hm, I don't think uv__signal_tree is node specific, it can be any library that statically links to libuv, so I think we have the same problem. I think your PR is trying to solve for an issue when someone renames `bun` to `node` and calls it a day. I think there's an off chance someone might link libuv and call it node too, but perhaps much less likely. Are the rest of the checks insufficient?

grcevski (commented):
This looks good, I left a few comments but I have one final question and one suggestion.

This is for handling the case when a customer names a `bun` based executable as `node`, right? Because we detect nodejs simply by looking at the executable name. If this is the case, I wonder if you considered the approach to change how we discover node. We can change the executable name matching to symbol lookup and we can also discover `bun` too and give it special type NodejsBun.  This way we don't need all the complex logic later during attachment.

I think we need to adjust the docs to say that we are not going to be attaching to stripped nodejs binaries, which is change in behaviour from before.

pkg/appolly/discover/attacher.go
NimrodAvni78: yes exactly, since the signal check might take up to 500ms it was moved off the main thread

NimrodAvni78:
> This is for handling the case when a customer names a bun based executable as node, right? Because we detect nodejs simply by looking at the executable name. If this is the case, I wonder if you considered the approach to change how we discover node. We can change the executable name matching to symbol lookup and we can also discover bun too and give it special type NodejsBun. This way we don't need all the complex logic later during attachment.

Done, moved the symbol based detection to discovery phase, and made sure that this does not break any "exotic" builds of node (node SEA, vercel/pkg,  alpine musl)

> I think we need to adjust the docs to say that we are not going to be attaching to stripped nodejs binaries, which is change in behaviour from before.

Actually we do manage to recognize stripped node binaries, the strip only removes `.symtab` only. and not  `.dynsym` so we will keep detecting them, the only thing that won't work is the `uv__signal_tree` search for registered SIGUSR1 probes, but that was a known limitation which is why we have a fallback source code scan

MrAlias (approved):

PR #3364, opened 2026-09-08
nodejs: close the inspector when injection fails

### LINKED ISSUE #3274: Failed Node.js injection can leave the inspector open on 127.0.0.1:9229
### Component

OBI (eBPF instrumentation)

### What happened?

When Node.js agent injection fails, OBI can leave the target process's inspector
listening on `127.0.0.1:9229`. `process._debugEnd()` is only sent from the defer in
`injectFileWS` (`pkg/internal/nodejs/injector.go:281-283`), which is reached only
after the WebSocket upgrade succeeds. Every failure before that point returns
without attempting to close the inspector:

- `pkg/internal/nodejs/nodejs.go:124-127` — SIGUSR1 is delivered, then `connectWait`
  gives up after 5s. If the inspector opens later (loaded host, busy event loop),
  it stays open and OBI never comes back.
- `pkg/internal/nodejs/injector.go:295-299` — debugger URL lookup fails → `conn.Close()`
  only; the inspector itself is untouched.
- `pkg/internal/nodejs/injector.go:310-314` — WebSocket upgrade fails → same.

Related: the `_debugEnd` result is discarded (`_ =`, `injector.go:282`), so a failed
close is silent even on the success path.

Impact is bounded but real: the inspector binds loopback inside the process's netns,
so it is not reachable from outside the pod. Within that netns, though, the debugger
is arbitrary code execution in the instrumented process — a sidecar or any
lower-privilege foothold in the same pod can escalate through it, and the window
lasts until the process restarts.

A failed injection should not leave the process more exposed than before it ran.
If OBI opened the inspector via SIGUSR1, it should make a best-effort attempt to
close it on every failure path, and log when that attempt fails.

### Steps to reproduce

1. Run a Node.js app with no custom SIGUSR1 handler and let OBI discover it.
2. Force a post-SIGUSR1 failure. Deterministic variant: shorten the `connectWait`
   budget in `nodejs.go:124` (e.g. 5s -> 1ms) so the connect attempt gives up while
   the inspector still comes up.
3. Inside the process's netns, `curl http://127.0.0.1:9229/json/version` (or check
   `/proc/<pid>/net/tcp` for a LISTEN on `0x240D`). The inspector is still open,
   after OBI logged `couldn't attach NodeJS injector`.

The patched timeout only makes the failure deterministic; it is not required. Any of
the `inspectorRequestTimeout` (5s, `injector.go:42`) deadlines firing against a
slow or stalled inspector reaches the same state on an unmodified build — a
CPU-throttled or event-loop-blocked Node.js process is enough. These are the
stalled-inspector conditions #2231 bounded for #2039: the timeouts correctly turned
an indefinite hang into an error return, and the error return leaves the inspector open.
I have not caught this in the field, though: the paragraph above is reasoning from the
code, and the Node.js services in the cluster I checked are SDK-instrumented, so the
injector never runs against them.

### Relevant logs

```
level=ERROR component=nodejs.Injector msg="couldn't attach NodeJS injector" pid=<pid> error="failed to connect to inspector after SIGUSR1: ..."
level=ERROR component=nodejs.Injector msg="trace-context propagation and nodejs runtime metrics will not work for NodeJS services!"
```

### Additional context

Suggested shape of a fix: track whether OBI opened the inspector (SIGUSR1 path) as
opposed to finding it already open (`--inspect`), and on any failure reconnect once
to send `process._debugEnd()`, logging the outcome. Scoping it that way also avoids
the inverse problem in the current code, where the unconditional defer closes an
inspector the operator opened deliberately with `--inspect`.

- #2941 asks for retry + an injection result metric, which would make these failures
  alertable; it does not close the inspector. This issue covers the cleanup half.
- #2942 (one inspector port per netns in `cluster` mode) returns from the same code
  path but should not leak: the sibling fails to bind 9229, so no inspector is opened.
- Paths present since #1514 (`connectWait`) and #2939 (`injectViaConn`); read on
  `main` @ f1fec547, unchanged in v0.12.2.
- Drafted with Claude Code assistance; I verified the code paths and reviewed the text.

### Version

main @ f1fec547 (also present in v0.12.2)

### Deployment

Kubernetes

bourbonkk:
## Summary

Fixes #3274.

`process._debugEnd()` was only sent from `injectFileWS`, which runs after the WebSocket upgrade succeeds. Every failure before that point returned with the inspector still listening on `127.0.0.1:9229` until the process restarted, and a failed close was discarded silently.

This closes it on each failure path, over a connection of its own since the one the injection used is unusable by then, and warns when the close attempt itself fails — that log is the only signal an operator gets today.

It also adds ownership tracking: OBI now closes only an inspector it opened itself, and leaves alone one the process asked for via `--inspect` or `NODE_OPTIONS`. The previous unconditional `defer` would close the operator's debugging session.

Follows the shape agreed in the issue discussion. The stalled-inspector case (5s deadlines firing repeatedly) remains not closable over the inspector protocol by definition; it now produces a warn log instead of failing silently.

- `closeInspector` / `sendDebugEnd`, composed from the existing connect/upgrade/evaluate helpers
- `processRequestedInspector` for ownership, checking cmdline and `NODE_OPTIONS`
- hardcoded `127.0.0.1`, `9229`, `5s`, `200ms` lifted to named constants
- 11 tests covering each failure path, ownership, and the close-fails case

## Validation

Not a feature change, so the features docs and support matrix are untouched.

### DIFF
--- a/pkg/internal/nodejs/injector.go
+++ b/pkg/internal/nodejs/injector.go
@@ -300,10 +300,6 @@ func sendMessageWithTimeout(wsConn *websocket.Conn, data []byte, timeout time.Du
 	return nil
 }
 
-// injectFileWS evaluates the agent, then always closes the WebSocket and, when
-// OBI opened the inspector, closes that too. A failed injection is left to the
-// caller: it cleans up over a fresh connection, which this stalled or broken
-// session cannot do.
 func (i *NodeInjector) injectFileWS(pid int, wsConn *websocket.Conn, payload []byte, openedByOBI bool) error {
 	defer wsConn.Close()
 
@@ -322,10 +318,6 @@ func (i *NodeInjector) injectFileWS(pid int, wsConn *websocket.Conn, payload []b
 	return nil
 }
 
-// injectViaConn drives the injection over an established inspector
-// connection. Every failure after the inspector is open would otherwise leave
-// the debugging interface listening, so a failed attempt closes it again when
-// OBI is the one that opened it.
 func (i *NodeInjector) injectViaConn(pid int, conn net.Conn, openedByOBI bool) (err error) {
 	defer func() {
 		if err != nil && openedByOBI {
@@ -347,8 +339,6 @@ func (i *NodeInjector) injectViaConn(pid int, conn net.Conn, openedByOBI bool) (
 	return i.injectFileWS(pid, wsConn, payload, openedByOBI)
 }
 
-// dialInspectorWS opens the debugger session that carries payload, taking
-// ownership of conn: it is closed on every failure.
 func (i *NodeInjector) dialInspectorWS(conn net.Conn, payload []byte) (*websocket.Conn, error) {
 	wsURL, err := i.requestDebuggerURL(conn)
 	if err != nil {
@@ -368,9 +358,6 @@ func (i *NodeInjector) dialInspectorWS(conn net.Conn, payload []byte) (*websocke
 	return wsConn, nil
 }
 
-// closeInspector makes a best-effort attempt to close an inspector that OBI
-// opened but could not inject into, over a connection of its own: the one the
-// injection used is already gone or unusable by this point.
 func (i *NodeInjector) closeInspector(pid int) {
 	conn, err := connectWait(inspectorHost, inspectorPort, inspectorCloseWait, inspectorCloseInterval)
 	if err != nil {
--- a/pkg/internal/nodejs/injector.go
+++ b/pkg/internal/nodejs/injector.go
@@ -41,6 +41,36 @@ type evalParams struct {
 
 const inspectorRequestTimeout = 5 * time.Second
 
+const inspectorHost = "127.0.0.1"
+
+// The inspector's fixed port. A variable so tests can dial a local stand-in,
+// following the cmdlineForPID pattern used elsewhere in this package.
+var inspectorPort = 9229
+
+// The inspector protocol is the only way to close the inspector again: SIGUSR1
+// opens it and no signal closes it.
+const debugEndExpression = "process._debugEnd();"
+
+const (
+	agentRequestID    = 1
+	debugEndRequestID = 2
+)
+
+const (
+	inspectorConnectInterval = 200 * time.Millisecond
+	inspectorCloseInterval   = 200 * time.Millisecond
+)
+
+// How long the injection waits for the inspector to come up after SIGUSR1,
+// and how long a cleanup attempt waits for one that opened later than that.
+// The cleanup budget is kept short: the injector runs synchronously from the
+// attacher loop, so it delays every other process being instrumented.
+// Variables for the same reason as inspectorPort.
+var (
+	inspectorConnectWait = 5 * time.Second
+	inspectorCloseWait   = 2 * time.Second
+)
+
 // IMPORTANT: the code in this file needs to run in the network namespace of the
 // target process in order to be able to connect to its inspector port - the
 // network namespace switching is done by the withNetNS function, which locks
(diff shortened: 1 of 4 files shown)


pkg/internal/nodejs/injector.go
```diff
@@ -41,6 +41,36 @@ type evalParams struct {
 
 const inspectorRequestTimeout = 5 * time.Second
 
+const inspectorHost = "127.0.0.1"
+
+// The inspector's fixed port. A variable so tests can dial a local stand-in,
+// following the cmdlineForPID pattern used elsewhere in this package.
```
rafaelroquetto: please remove this comment - rationale: redundant

pkg/internal/nodejs/injector.go
```diff
@@ -41,6 +41,36 @@ type evalParams struct {
...
+var inspectorPort = 9229
+
+// The inspector protocol is the only way to close the inspector again: SIGUSR1
+// opens it and no signal closes it.
```
rafaelroquetto: remove - this comment is out of context

pkg/internal/nodejs/injector.go
```diff
@@ -41,6 +41,36 @@ type evalParams struct {
...
+const (
+	inspectorConnectInterval = 200 * time.Millisecond
+	inspectorCloseInterval   = 200 * time.Millisecond
+)
+
+// How long the injection waits for the inspector to come up after SIGUSR1,
+// and how long a cleanup attempt waits for one that opened later than that.
+// The cleanup budget is kept short: the injector runs synchronously from the
+// attacher loop, so it delays every other process being instrumented.
+// Variables for the same reason as inspectorPort.
```
rafaelroquetto: idem - the code if fairly obvious, not necessary (and a maintenance burden)

pkg/internal/nodejs/nodejs.go
```diff
@@ -87,13 +95,18 @@ func (i *NodeInjector) attachAgent(pid int, elfFile *elf.File) error {
 // it checks for a custom SIGUSR1 handler and either sends SIGUSR1 to open the
 // inspector or bails out.
 func (i *NodeInjector) injectFile(pid int, elfFile *elf.File) error {
-	conn, err := connect("127.0.0.1", 9229)
+	// An inspector the process asked for belongs to whoever started it, and
+	// closing it would break their debugging session. Anything else listening
+	// was opened by OBI, here or in an earlier attempt that failed to clean up.
+	mayClose := !processRequestedInspector(pid)
```
rafaelroquetto: I think this doesn't solve the entire story, and to be honest, I think the "bug" is somewhat different than this.

`mayClose` implies that we are allowed to close the debugger _if_ we _think_ the process has not started it - but it does't tackle 3rd party processes dealing with the debugger - so we are effectively closing a legitimate debugger for instance.

rafaelroquetto (changes requested):
Hey @bourbonkk thanks for looking into this.

I think you have genuinely uncovered an interesting corner case which begs the question: who and when should close the debugger?

It turns out the least disruptive approach IMHO is that we should only close the debugger if we were the ones who opened it - anything else could be disrupting a legitimate scenario (i.e. `processRequestedInspector()` is not enough to determine that). For example: a 3rd party application (i.e. an actual debugger) could have opened the connection and now we are going and closing it unconditionally (this is an existing bug, not introduced by your code).


So if we want to be conservative, I think the logic should be simplified to:

 - were we the ones sending SIGUSR1?
  - if yes, we close it
  - if no, we don't bother

and that's all that there is to it.

Otherwise, this is really a non-issue in practice, so if the above is not feasible I wouldn't bother with it.

--

also: please remove most of the comments - they don't really add much to the code, but rather get in the way. To quote our AGENTS.md:

```
Comments must be minimal:

    Do not add comments that restate the code.
    Prefer clearer code over explanatory comments.
    Add comments only when they provide necessary context, explain non-obvious behavior, or document verifier, kernel, or ABI constraints.
```

pkg/internal/nodejs/nodejs.go
```diff
@@ -121,12 +134,44 @@ func (i *NodeInjector) injectFile(pid int, elfFile *elf.File) error {
...
+	}
+
+	return isInspectorFlag(env["NODE_OPTIONS"])
+}
+
+// isInspectorFlag matches the whole --inspect family: --inspect,
+// --inspect-brk, --inspect-wait, --inspect-port and their =value forms.
+func isInspectorFlag(arg string) bool {
+	return strings.Contains(arg, "--inspect")
 }
```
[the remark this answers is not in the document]
rafaelroquetto: This is related to what I wrote, so I'd give some weight to it.

pkg/internal/nodejs/injector.go
```diff
@@ -277,41 +300,113 @@ func sendMessageWithTimeout(wsConn *websocket.Conn, data []byte, timeout time.Du
...
 }
 
-func (i *NodeInjector) injectFileWS(wsConn *websocket.Conn, payload []byte) error {
-	defer func() {
-		_ = sendEvaluate(wsConn, "process._debugEnd();", 2)
-	}()
+// injectFileWS evaluates the agent, then always closes the WebSocket and, when
+// OBI opened the inspector, closes that too. A failed injection is left to the
+// caller: it cleans up over a fresh connection, which this stalled or broken
+// session cannot do.
```
rafaelroquetto: nit: I think we can do without these as well - it's pretty much what the code says (`openedByOBI` is a great name and says it all)

pkg/internal/nodejs/injector.go
```diff
@@ -277,41 +300,113 @@ func sendMessageWithTimeout(wsConn *websocket.Conn, data []byte, timeout time.Du
...
+			i.logInspectorStaysOpen(pid, err)
+		}
+	}
+
 	return nil
 }
 
-func (i *NodeInjector) injectViaConn(conn net.Conn) error {
-	wsURL, err := i.requestDebuggerURL(conn)
+// injectViaConn drives the injection over an established inspector
```
rafaelroquetto: ditto

pkg/internal/nodejs/injector.go
```diff
@@ -277,41 +300,113 @@ func sendMessageWithTimeout(wsConn *websocket.Conn, data []byte, timeout time.Du
...
+	wsConn, err := i.dialInspectorWS(conn, payload)
+	if err != nil {
+		return err
+	}
+
+	return i.injectFileWS(pid, wsConn, payload, openedByOBI)
+}
 
-	payload, err := evaluateRequest(i.agentCode(), 1)
+// dialInspectorWS opens the debugger session that carries payload, taking
```
rafaelroquetto: ditto

pkg/internal/nodejs/injector.go
```diff
@@ -277,41 +300,113 @@ func sendMessageWithTimeout(wsConn *websocket.Conn, data []byte, timeout time.Du
...
 	if err != nil {
 		conn.Close()
-		return fmt.Errorf("failed to connect to inspector WebSocket: %w", err)
+		return nil, fmt.Errorf("failed to connect to inspector WebSocket: %w", err)
+	}
+
+	return wsConn, nil
+}
+
+// closeInspector makes a best-effort attempt to close an inspector that OBI
```
rafaelroquetto: ditto - these functions are simple enough

pkg/internal/nodejs/injector.go
```diff
@@ -277,41 +300,113 @@ func sendMessageWithTimeout(wsConn *websocket.Conn, data []byte, timeout time.Du
...
-func (i *NodeInjector) injectViaConn(conn net.Conn) error {
-	wsURL, err := i.requestDebuggerURL(conn)
+// injectViaConn drives the injection over an established inspector
+// connection. Every failure after the inspector is open would otherwise leave
+// the debugging interface listening, so a failed attempt closes it again when
+// OBI is the one that opened it.
+func (i *NodeInjector) injectViaConn(pid int, conn net.Conn, openedByOBI bool) (err error) {
+	defer func() {
+		if err != nil && openedByOBI {
+			i.closeInspector(pid)
```
rafaelroquetto, line 324: here you already have an open `conn` to the inspector - I don't think `closeInspector` should try to open a new connection here again... instead can't we leverage the open connection, so that it sends the close statement before calling conn.Close()?

pkg/internal/nodejs/nodejs.go
```diff
@@ -121,12 +123,18 @@ func (i *NodeInjector) injectFile(pid int, elfFile *elf.File) error {
...
 	}
 
-	conn, err = connectWait("127.0.0.1", 9229, 5*time.Second, 200*time.Millisecond)
+	// Past this point the inspector is ours: SIGUSR1 has been delivered and
+	// cannot be taken back.
+	conn, err = connectWait(inspectorHost, inspectorPort, inspectorConnectWait, inspectorConnectInterval)
 	if err != nil {
+		// It may still be coming up, in which case it would stay open with
+		// nobody left to close it.
+		i.closeInspector(pid)
```
rafaelroquetto, line 132: I am not sure about this - if we failed to connect, then there's no point on trying to connect again - it may fail again and still leave it open nonetheless...

If this is an actual practical issue, I'd say we should increase the timeout/wait time instead.

rafaelroquetto (changes requested):

bourbonkk:
Pushed the comment removals, the unconditional `wsConn.Close()`, and the ownership rewrite: `processRequestedInspector` and `isInspectorFlag` are gone. OBI now closes only an inspector it opened via SIGUSR1 in that attempt — an already-listening one is left alone whoever opened it, which covers the third-party case too. The cost is that a leak from an earlier attach is no longer cleaned up later; that seemed the right trade, and it still warns.

Two I'd like your call on.

Reusing the open conn (`injector.go:332`) — `_debugEnd()` needs the CDP WebSocket, and its URL comes from `/json/list`. The failure this path handles is "no debugging targets available", so there is no URL and nothing can be sent over that conn. What the reconnect is really for is retrying /json/list once the inspector has registered its target. Retrying on the same conn would be possible, but httpGet builds a fresh `bufio.NewReader` per call and I couldn't confirm the inspector keeps the connection alive. Happy to try if you'd rather avoid the second dial.

Retry vs. longer wait (`nodejs.go:132`) — agreed, and a longer wait is better than a retry since it injects into a late inspector instead of closing it. The catch is that the injector runs synchronously in the attacher loop, so the wait delays every other process. What budget would you be comfortable with? I'll drop the retry and raise `inspectorConnectWait` to whatever you pick.

rafaelroquetto:
@bourbonkk thank you for the changes - but looking closer at issue, I believe this is a non-issue.

The debugger can only be closed if we managed to connect to it in first place. That is:

 1. send SIGUSR1
 2. connect to the endpoint after SIGUSR1
 3. upgrade to the websocket connection
 4. send the close command

We already cater for the window between SIGUSR1 and the endpoint becoming available by a huge timeout (handlied in the `connectWait()` function). There's nothing really that can be done - you could raise the timeout to 30 seconds and someone will argue the problem will still be there at the 31st second. So if the timeout is being insufficient, then the fix is to increase it - but we don't have any evidence that this is ever the case. It has not been so far (not that we have heard, and you only managed to reproduce the issue by _artificially_ shortening the timeout).

If anything fails between 2 and 4 there is nothing we can do about it - it's outside OBI's control - yes, we could poll and try to reconnect but that does not guarantee anything (because we don't really see the actual failure).

So unless there's a particular issue (beyond some static code analysis and hypothetical scenario) - I mean, an actual tangible _reproducible_ issue that would allow us to at least to attempt to deliver a _targeted_ fix, there's nothing really to fix here and we can close it.

The only thing that may be worth pursuing is the case I mentioned in one of my comments above: only closing the debugger if it was OBI who opened it in first place. If you'd like to convert this PR on that direction and tackle that only, then I'd be thankful - otherwise, I will probably just go ahead and close this.


And just for future reference: be reminded that posting AI responses for communication is forbidden by our AI policy and will cause future PRs to be closed without notice. We want to be able to engage with contributors in a meaningful, constructive and welcoming way - dialogue is important in codereviews and proxying AI responses defeats this purpose. Please have a look at our [AI-POLICY](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/AI-POLICY.md).

rafaelroquetto:
I'm going to close it for the time being, feel free to reopen or open a new one if you decide to tackle the "OBI closes only if it has oppened" issue.

PR #3366, opened 2026-09-08
python: harvest HTTP routes for FastAPI and Flask

grcevski:
## Summary

Adds HTTP route harvester for Python, very similar one to the Node.js route harvester that works for FastAPI and Flask for now. If we get requests for other popular REST API python frameworks we can extend this further. The parser is not bullet proof, it's purposefully made simple to avoid expensive parsing, much like we do for nodejs.

## Validation


marctc (approved):
LGTM! Python asyncio integration tests have to be updated from `GET /sequential/1` to `GET /sequential/{req_id}`

PR #3371, opened 2026-09-09
generictracer: backfill listening ports for processes discovered late

### LINKED ISSUE #3297: Kafka spans stop after an OBI restart until connections are re-established (pre-existing connections not instrumented)
## What happened

Kafka broker spans stop entirely after an OBI restart. They come back only once the broker's client connections are re-established — immediately if the broker restarts, or more slowly as clients recycle connections on their own.

OBI appears to instrument only connections that are established while it is running. Long-lived pre-existing connections are never picked up. Kafka is hit hardest because its clients (librdkafka, sarama) hold persistent connections; protocols whose connections churn are unaffected.

Reproduced on both **v0.10.0** and **v0.12.2**.

## Environment

- OBI **v0.10.0** and **v0.12.2**, run as the Collector receiver (`otelcol-contrib` 0.159.0 and 0.160.0, `obireceiver`)
- Kubernetes DaemonSet, one OBI per node
- Ubuntu 24.04.4, kernel 6.8.0-124-generic, containerd 1.7.27
- Relevant config: `ebpf.context_propagation: all`, `network.enable: false`, discovery instruments all non-system namespaces
- Workload: `opentelemetry-demo`. The affected process is the Kafka broker (JVM); clients are Go (`sarama`) and C (`librdkafka`) services

## Reproduction

Deterministic across five OBI restarts in one afternoon. Every restart dropped Kafka spans to zero. Counts are OBI Kafka spans per 5 minutes.

| Time (UTC) | Event | OBI Kafka spans |
| --- | --- | --- |
| 01:40 – 01:50 | steady state, OBI up for 9+ days | 613 – 615 |
| 01:55 | **OBI restart** | 8, then 0 for ~25 min |
| 02:24 | broker restart | 34 |
| 02:25 – 04:55 | steady | 26 – 43 |
| 04:58 | **OBI restart** | 18, then 0 |
| 05:13 | **OBI restart** | 0 – 2 |
| 05:20 | broker restart | 39, 44 |
| 05:33 | **OBI restart** | 22, then 9 |
| 05:39 | broker restart | 36, 34, 32, 34 |
| 06:14 | **OBI restart — upgrade to v0.12.2** | 17, then **0, 0** |
| 06:25 – 06:40 | clients reconnect on their own, no broker restart | 7, 35, 39, 59 |
| 06:44 | **OBI restart** | 36, 33, 34 |

Two things worth noting:

- The v0.12.2 upgrade at 06:14 reproduced the drop identically, so this is not fixed in the current release.
- At 06:25 the spans returned without any broker restart. Kafka clients recycle connections periodically, and OBI picks those up. That is consistent with "only connections established while OBI runs are instrumented".

The OBI restarts spanned four different pod security configurations (privileged; a reduced capability set; with and without `hostNetwork`) and two OBI versions. The result was identical in every case, so this is not a permission or host-namespace issue.

## Controls

Two controls in the same cluster over the same window confirm the workload never changed:

| Time (UTC) | OBI Kafka | SDK Kafka | OBI Postgres |
| --- | --- | --- | --- |
| 01:45 | 615 | 120 | 707 |
| 01:50 | 613 | 132 | 677 |
| 02:00 | 0 | 144 | 713 |
| 02:15 | 0 | 132 | 699 |
| 05:00 | 0 | 87 | 685 |
| 05:15 | 0 | 90 | 604 |
| 06:15 | 0 | 99 | — |
| 06:20 | 0 | 135 | — |

- **SDK-side Kafka spans** (from the demo's instrumented clients) continue at their normal rate throughout, so messages keep flowing.
- **Postgres spans from the same OBI instances** stay at 560 – 760 per 5 min across every OBI restart. Only Kafka is affected.

The Kafka broker pod itself had 9 days of uptime and zero restarts when the first drop happened. The 9-day steady state beforehand appears to be an artifact: the OBI pods had not restarted in that time, so they had observed the broker's connections as those connections were established.

## Suspected cause

The Kafka wire protocol is length-prefixed binary. An eBPF probe that attaches mid-stream lands at an arbitrary offset inside a frame and has no way to resynchronise to a frame boundary. HTTP is text-based and can resync on a request line, which would explain why HTTP-based protocols survive.

Two precedents suggest pre-existing-connection handling is being addressed protocol by protocol:

- #2663 — `redis: better handling of existing connections`
- #2772 — `h2/grpc: context propagation for pre-existing connections`

I checked `v0.12.2`: `pkg/ebpf/common/kafka_detect_transform.go` and `bpf/generictracer/protocol_kafka.h` have no resync or existing-connection handling. The Kafka changes since `v0.10.0` (#2656, #2683, #2693, #2722, #2738) are all about multi-topic and multi-chunk parsing, which matches the observation that upgrading changes nothing here.

## Questions

1. Is this a known and intended limitation for length-prefixed binary protocols?
2. If so, would you accept a documentation note? #527 asks about Kafka limitations and this would fit there.
3. Is there interest in a general mid-stream resync for Kafka, along the lines of what #2663 did for Redis?

## Open question on our side

Recovery plateaus well below the original rate and we cannot fully explain it. It settles at a **stable** ~32 – 43 spans per 5 min against 613 – 615 before the first restart, and held that band for hours across several recovery cycles. It is a flat fraction, not a ramp still converging. The OBI-to-SDK span ratio goes from ~5.1 down to ~0.35.

Our working guess is that the original figure accumulated over nine days, during which OBI observed every connection establishment, whereas a freshly started OBI only ever sees the subset of connections that churn. If that reasoning is wrong, or if the pattern suggests something specific to you, we are happy to gather more data.

darkraider01:
### Description

Fixes #3297.

`generictracer` previously executed `runItersForPids()` only once during startup in `Run()`. Any process discovered after startup (e.g. secondary containers or late-starting services such as Kafka or Postgres) never had its listening sockets populated into the `listening_ports` BPF map. When early inbound traffic arrived on pre-existing connections for those processes, OBI misclassified the server traffic as unsolicited client responses and dropped the spans.

This PR re-runs the iterator backfill whenever new processes are discovered:

- `AllowPID()` marks an `itersDirty atomic.Bool` flag after adding the PID to `pidsFilter`.
- `Run()` performs the initial startup iterator pass, then launches a lightweight background goroutine that checks the flag every 3 seconds.
- When dirty, `watchForNewPids()` consumes the flag via `Swap(false)` and executes `runItersForPids()`. Using `Swap(false)` before iteration ensures any PID discovered while an iterator walk is active re-arms the flag for the subsequent tick.
- `runItersForPids()` remains unchanged, retaining its existing per-pass network namespace deduplication (`seen[inode]`).

### DIFF
--- a/pkg/internal/ebpf/generictracer/generictracer.go
+++ b/pkg/internal/ebpf/generictracer/generictracer.go
@@ -14,6 +14,7 @@ import (
 	"math"
 	"os"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 	"unsafe"
@@ -57,6 +58,8 @@ type Tracer struct {
 	eventCtx         *ebpfcommon.EBPFEventContext
 	jvmUSDTManager   ebpfcommon.USDTSpecManager
 	pythonRuntime    *pythonRuntimeController
+
+	itersDirty atomic.Bool
 }
 
 func tlog() *slog.Logger {
@@ -165,6 +168,8 @@ func (p *Tracer) AllowPID(pid app.PID, ns uint32, fi *exec.FileInfo) {
 	if p.pythonRuntime != nil {
 		p.pythonRuntime.allow(pid, ns, fi, serviceSource)
 	}
+
+	p.itersDirty.Store(true)
 
 	if err := p.rebuildValidPids(); err != nil {
 		p.log.Error("rebuilding the BPF PID filter", "error", err)
@@ -684,6 +689,24 @@ func (p *Tracer) runItersForPids() {
 	}
 }
 
+const itersBackfillInterval = 3 * time.Second
+
+func (p *Tracer) watchForNewPids(ctx context.Context, interval time.Duration) {
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			if p.itersDirty.Swap(false) {
+				p.runItersForPids()
+			}
+		}
+	}
+}
+
 func (p *Tracer) Tracing() []*ebpfcommon.Tracing { return nil }
 
 func (p *Tracer) RecordInstrumentedLib(id uint64, closers []io.Closer) {
@@ -757,7 +780,9 @@ func (p *Tracer) Run(
 	go p.lookForTimeouts(ctx, parseContext, timeoutTicker, eventsChan)
 	defer timeoutTicker.Stop()
 
+	p.itersDirty.Store(false)
 	p.runItersForPids()
+	go p.watchForNewPids(ctx, itersBackfillInterval)
 
 	p.log.Info("Launching p.Tracer")
 
--- a/pkg/internal/ebpf/generictracer/generictracer_test.go
+++ b/pkg/internal/ebpf/generictracer/generictracer_test.go
@@ -8,6 +8,7 @@ package generictracer
 import (
 	"context"
 	"math"
+	"sync/atomic"
 	"testing"
 	"time"
 	"unsafe"
@@ -465,4 +466,93 @@ func (f fakeServiceFilter) CurrentPIDs(ebpfcommon.PIDType) map[uint32]map[app.PI
 		(*f.currentPIDsCalls)++
 	}
 	return f.current
+}
+
+func TestAllowPIDSignalsBackfill(t *testing.T) {
+	currentPIDsCalls := 0
+	tracer := &Tracer{
+		log:        tlog(),
+		pidsFilter: fakeServiceFilter{currentPIDsCalls: &currentPIDsCalls},
+	}
+	file := exec.New(exec.Init{Pid: 101, Service: svc.Attrs{}})
+
+	tracer.AllowPID(101, 42, file)
+
+	require.True(t, tracer.itersDirty.Load())
+	assert.Zero(t, currentPIDsCalls)
+}
+
+func TestWatchForNewPidsRunsWhenDirty(t *testing.T) {
+	const interval = 10 * time.Millisecond
+	filter := &countingServiceFilter{}
+	tracer := &Tracer{log: tlog(), pidsFilter: filter}
+
+	ctx, cancel := context.WithCancel(t.Context())
+	defer cancel()
+	go tracer.watchForNewPids(ctx, interval)
+
+	time.Sleep(interval * 2)
+	require.Zero(t, filter.calls.Load())
+
+	tracer.itersDirty.Store(true)
+	require.Eventually(t, func() bool { return filter.calls.Load() == 1 }, time.Second, interval)
+
+	time.Sleep(interval * 3)
+	assert.Equal(t, int64(1), filter.calls.Load())
+
+	tracer.itersDirty.Store(true)
+	require.Eventually(t, func() bool { return filter.calls.Load() == 2 }, time.Second, interval)
+}
+
+func TestWatchForNewPidsPreservesConcurrentSignalDuringWalk(t *testing.T) {
+	const interval = 10 * time.Millisecond
+	filter := &blockingServiceFilter{
+		entered: make(chan struct{}),
+		release: make(chan struct{}),
+	}
+	tracer := &Tracer{log: tlog(), pidsFilter: filter}
+
+	ctx, cancel := context.WithCancel(t.Context())
+	defer cancel()
+
+	tracer.itersDirty.Store(true)
+	go tracer.watchForNewPids(ctx, interval)
+
+	select {
+	case <-filter.entered:
+	case <-time.After(time.Second):
+		t.Fatal("timed out waiting for watcher to enter walk")
+	}
+
+	tracer.itersDirty.Store(true)
+	filter.release <- struct{}{}
+
+	require.Eventually(t, func() bool { return filter.calls.Load() == 2 }, time.Second, interval)
+}
+
+type countingServiceFilter struct {
+	fakeServiceFilter
+	calls atomic.Int64
+}
+
+func (f *countingServiceFilter) CurrentPIDs(ebpfcommon.PIDType) map[uint32]map[app.PID]svc.Attrs {
+	f.calls.Add(1)
+	return nil
+}
+
+type blockingServiceFilter struct {
+	fakeServiceFilter
+	entered chan struct{}
+	release chan struct{}
+	calls   atomic.Int64
+}
+
+func (f *blockingServiceFilter) CurrentPIDs(ebpfcommon.PIDType) map[uint32]map[app.PID]svc.Attrs {
+	f.calls.Add(1)
+	select {
+	case f.entered <- struct{}{}:
+		<-f.release
+	default:
+	}
+	return nil
 }


bourbonkk:
Thanks for picking this up. The shape looks right to me:

- signalling from `AllowPID` and executing from `Run()` avoids the `Iters()` caching problem
- `pidsFilter.AllowPID` runs before the store, so the new PID is already in `CurrentPIDs` when the backfill walks it
- reading `itersSignal` before `itersBackfilled` means a discovery landing mid-backfill isn't lost
- `iter_tcp` only writes `listening_ports` for live listening sockets, so re-running is idempotent

## The one thing I'd look at before this lands

`runItersForPids` was written for a single startup pass, and running it repeatedly changes its cost profile. `tpinjector.AllowPID` solves the same problem and carries three guards this path doesn't:

| Guard | `tpinjector` | this PR |
| --- | --- | --- |
| netns dedup across runs | persistent `seenNetns` LRU | `seen` rebuilt per call |
| EPERM handling | `backfillDisabled`, warns once | none |
| per-netns retry cap | `maxNetnsAttempts` | none |

### No netns dedup across runs

`seen` is built inside `runItersForPids`, so every re-run walks every netns again — not just the newly discovered ones. On a node with ~100 pods and a few processes each, that's several hundred `/proc/<pid>/ns/net` stats plus ~100 `setns` + full socket walks, every time discovery goes quiet.

### EPERM is permanent but logged every time

Without `CAP_SYS_ADMIN`, `runItersForPids` logs an error per PID. That used to be one burst at startup; now it repeats for the life of the process.

### The debounce has no upper bound, so it can starve

`itersBackfillDue` only fires once `now - signal >= quietPeriod`. On a node where something is discovered every couple of seconds — pods cycling, a crashlooping workload — the signal keeps getting refreshed and the backfill never runs at all. A "run at least every N seconds regardless" cap would bound it.

## Smaller

- `time.Now().UnixNano()` drops the monotonic reading, so an NTP step backwards makes `now.Sub(time.Unix(0, signal))` negative and the backfill stays not-due until the next `AllowPID`. A generation counter (`itersSignal.Add(1)`) alongside a separate timestamp would sidestep it.
- When `Iters()` is empty (kernels below 5.11) `runItersForPids` returns immediately, but the 500ms ticker keeps going for the life of the process. `Iters()` is safe to call from `Run()`, so the goroutine could just not start.

The tests read well — the inclusive boundary, not re-running a consumed signal, and re-arming on a new one are all covered.

darkraider01:
Pushed updates for your review, plus a quick fix right after: caching by netns broke container restarts in place (like the Postgres case in #3297), so dedup is keyed on PID (seenPids) instead.

Two suggestions I left out:

Exit if Iters() is empty: In generictracer, Iters() is never empty today, so the check would just be dead code.
Retry cap for non-EPERM errors: The cache's 5-minute TTL already handles retries naturally without extra tracking.

For tests, hermetically faking EPERM isn't really practical in CI, but I added TestRunItersForPidsSkipsAlreadyBackfilledPids to cover both the dedup and restart cases.

Ready for another look!

grcevski (changes requested):
I think we need a redo of this PR. The extensive comments (likely made by an AI) need to be removed where they were added in our code.

I also have a question regarding the complexity and tracking of PIDs. I think this can be much simpler. We just need to know when was the last time iters have walked as we are adding pids. Have a goroutine launched that monitors for new pids added and the time not being less than few seconds, then it walks the iterator again.

grcevski:
Hey @darkraider01, we are trying to investigate why LLMs are not following our instructions (in this case related to excessive comments)... Do you mind sharing what do you use? I'm assuming Claude based on the wording it has added, but can you tell us if you are using Claude CLI or an extension in another tool like VS Code?

Here's an attempt to reign in the LLMs a bit more https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3379/changes

darkraider01:
@grcevski hey, i've used claude code for this, exact model is opus 5 if it helps!

darkraider01:
Before I redo this, one thing I keep coming back to: there's no hook on listen() anywhere. listening_ports only gets written by the iter/tcp backfill and by the accept4 kretprobe, and accept4 is gated behind valid_pid. But the removal side, inet_csk_listen_stop, isn't gated at all.

So a port can be removed for any process, but only added for one we already discovered. That asymmetry is what makes the restart case a race: a restarted process calls listen(), nothing records it, and it stays unrecorded until it accepts a connection after we've discovered its new pid.

Is there a reason there's no add-side hook? If we registered the port at listen() time, the backfill would only be needed for sockets that were already listening before OBI attached, which is a much smaller job than what it's doing now.

Two smaller ones:

Without the pid tracking there's no place for the CAP_SYS_ADMIN check @bourbonkk raised, so a node missing it logs an error per pid on every walk. Much quieter under your design since walks only happen when pids are added, but not zero. Guard, or leave it?

For "not less than a few seconds", do you have a number in mind, and should it be a constant or configurable?
@grcevski

bourbonkk:
@darkraider01

**CAP_SYS_ADMIN** — guard the log, not the walk. Warn once on the first `EPERM` and keep walking. It lives on the `netns.WithNetNS` error path in `runItersForPids` next to the existing `os.ErrNotExist` case, so it doesn't need pid tracking. Skipping the walk would only save failing `setns` calls, and if `EPERM` is ever per-netns rather than global, disabling would silently drop backfill for the namespaces that do work.

**Debounce** — in the #3297 capture the discovery burst spread over ~10s after `Launching p.Tracer`, so a 3–5s quiet period collapses it into one walk. Add an upper bound (e.g. 30s from the first un-walked signal) so steady pod churn can't starve it. A constant is fine.

**listen() hook** — confirmed on main: adds are `iter_tcp.c:26` and `k_tracer.c:128` (behind `valid_pid`), the delete at `k_tracer.c:1537` isn't gated. `inet_csk_listen_start` would be the natural twin of the existing stop hook and should write both `listening_ports` and `filter_ports`. Whether it's in scope here is @grcevski's call.

grcevski:
Can one of you explain why is this a problem?
> But the removal side, inet_csk_listen_stop, isn't gated at all.

>So a port can be removed for any process, but only added for one we already discovered. That asymmetry is what makes the restart case a race: a restarted process calls listen(), nothing records it, and it stays unrecorded until it accepts a connection after we've discovered its new pid.

How does this possibly relate to the issue we are solving here?

darkraider01:
You're right, it isn't a problem and it doesn't relate. Withdrawing it.

I was treating port registration as the thing that gates the race, but it isn't, discovery is.

MrAlias:
> You're right, it isn't a problem and it doesn't relate. Withdrawing it.
> 
> I was treating port registration as the thing that gates the race, but it isn't, discovery is.

Closing based on this. Please re-open if this is in error.

grcevski:
Actually we need this change, or a form of this change. I wrote some new tests with PHP for another PR and it exposes the same problem as described here. The AI reviewer came up with completely unrelated comment, something about ports being removed without a pid check which is completely bogus, but we need to iterate on newly added PID for a parent process.

I'm going to try this change for my new tests to see if it works or it needs something else.

grcevski:
Actually, on a second thought I'm just going to close this again. There are too many issues with the way it's written. On adding a new PID now we rescan all PIDs again, which is a massive operation, it should only scan the network namespace of the PID. I'm going to submit the proper fix.

PR #3372, opened 2026-09-09
give batched spans their own trace when the batch has no parent

NimrodAvni78:
## Summary

Protocols that carry several operations in one payload are parsed into several spans built from a single event, so every span inherits that event's ids. The extras already had their span id cleared, but they kept the event's trace id.

When the batch has a parent request that is correct: the operations are siblings under it. When it has no parent it puts several parentless spans in one trace, and a trace should not have more then 1 root.

## Fix

`detachExtraSpans` clears the span id on each extra and, when the extra has no parent request, also gives it its own trace id. Applied to Redis (both the generic and the Go tracer path), Kafka multi-topic, AMQP and Memcached noreply. It replaces four copies of the existing span-id clearing loop, so the change is a net two lines.

Batches that do have a parent request are untouched and still export as siblings in the parent's trace.

## Validation

### DIFF
--- a/internal/test/integration/components/pythonredis/main.py
+++ b/internal/test/integration/components/pythonredis/main.py
@@ -1,5 +1,7 @@
 from fastapi import FastAPI
 import os
+import threading
+import time
 import uvicorn
 import redis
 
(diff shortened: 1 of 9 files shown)


mmat11 (approved):
lgtm

grcevski (approved):
LGTM!

PR #3373, opened 2026-09-09
declare dns.lookup.duration as a plain group instead of a v2 refinement

NimrodAvni78:
Blocks https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3348

## Summary

weaver 0.26 drops a v2 refinement from live-check reads.  so existing dns metric refinement breaks. `schemas/obi/groups/dns.yaml` now becomes a groups-format declaration of `metric.obi.dns.lookup.duration` instead of a `definition/2` file with a `metric_refinements` override. `dns.question.name` stays `opt_in`; emitted telemetry is unchanged.

## Verified

Same registry, same one-datapoint sample, only the weaver version differs:

| | 0.25.1 | 0.26.1 |
| --- | --- | --- |
| before | `opt_in_attribute_not_present` | `required_attribute_not_present` (violation) |
| after | `opt_in_attribute_not_present` | `opt_in_attribute_not_present` |

Dropping `imports` changes nothing else: same 849 resolved attributes, and an `http.server.request.duration` sample gives identical findings.

## Validation

### DIFF
--- a/schemas/obi/groups/dns.yaml
+++ b/schemas/obi/groups/dns.yaml
@@ -1,47 +1,41 @@
-file_format: definition/2
-
-imports:
-  entities: ["*"]
-  events: ["*"]
-  metrics: ["*"]
-  spans: ["*"]
-  attribute_groups: ["*"]
-
-metric_refinements:
-  # OBI override of the upstream metric.dns.lookup.duration definition.
-  #
-  # Upstream (semconv v1.41.0) marks dns.question.name as `required`, but OBI
-  # deliberately disables it by default to bound metric cardinality (raw DNS
-  # question names are unbounded / attacker-controlled — see
-  # attr.DNSQuestionName in pkg/export/attributes/attr_defs.go; users can opt
-  # in via `attributes.select`). This group re-declares the attribute as
-  # `opt_in` so weaver live-check accepts its absence while still validating
-  # its type/value whenever a user opts in. All other parent attributes
-  # (error.type, …) are inherited unchanged via `extends`.
-  #
-  # NOTE on the group id: Weaver has no precedence semantics between a local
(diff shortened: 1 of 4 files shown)


grcevski (approved):
LGTM!

schemas/obi/groups/dns.yaml
```diff
@@ -1,47 +1,41 @@
...
+# upstream's `required` and fails every test whose workload resolves DNS. A
+# groups-format declaration takes precedence over the dependency's definition in
+# both 0.25 and 0.26, so it does not rely on refinement resolution or on the
+# group-id sort order described in
+# https://github.com/open-telemetry/weaver/issues/1578.
+#
+# The blanket `imports` block this file used to carry is redundant: live-check
+# runs with `--include-unreferenced`, which already pulls every upstream group
+# into the resolved registry. Dropping it leaves the registry byte-identical
+# (849 attributes, unchanged findings for upstream metrics).
```
MrAlias, line 22: `--include-unreferenced` does not make these imports redundant under the repository's current Weaver 0.25.1 pin. Removing them changes seven non-DNS metric groups: all six span-metric groups lose `x.obi.telemetry` (including the `generic` value for `telemetry.sdk.language`), and `metric.db_server_operation_duration` gets a different `error.type` definition. Since #3348 is still open, could we make this PR include the Weaver 0.26.1 bump or land after it? We should also add a representative resolution regression check and commit the schema docs regenerated with 0.26.1.

MrAlias (commented):
The DNS redeclaration works with Weaver 0.26.1, but this PR is not safe against the repository's current 0.25.1 toolchain. It should be ordered with the Weaver upgrade and update the generated schema artifacts.

MrAlias (approved):

PR #3374, opened 2026-09-09
Ruby on Rails routes harvester

mariomac:
## Summary

Adds static Rails route harvesting, including resource routes and split files.

It also fixes a bug of harvesting being skipped when reusing a tracer (`attacher.go`): OBI can reuse an existing eBPF tracer when instrumenting another application. Previously, that reuse path skipped route harvesting.

For example, if OBI instrumented Nginx first, Rails could reuse its tracer without having its routes harvested. The fix runs route harvesting for the newly discovered application even when the tracer already exists, so route discovery no longer depends on which application starts first.

## Validation


internal/test/integration/docker-compose-ruby.yml
```diff
@@ -2,7 +2,10 @@ version: "3.8"
 
 services:
   testserver:
-    image: ${TESTSERVER_IMAGE:?TESTSERVER_IMAGE must be set to a digest-pinned obi-testimg rails reference}
+    # TODO: Restore digest-pinned Rails image coordinates after publishing the updated route fixtures.
+    build:
+      context: ../../..
+      dockerfile: ./internal/test/integration/components/rubytestserver/testapi/Dockerfile${TESTSERVER_DOCKERFILE_SUFFIX:-}
```
mariomac, line 8: Chicken and egg issue:
1. We need a new testserver image.
2. The image is not published.
3. We can pre-publish it in another preceding PR because it would anyway stay untested.

So I decided to temporarily replace the `image` by a `build` clause. Once this PR is merged, we will publish the tested image and then revert this line (and few related changes in the test suite)

pkg/appolly/discover/attacher.go
```diff
@@ -456,6 +456,9 @@ func (ta *traceAttacher) loadExecutable(ie *ebpf.Instrumentable) (*link.Executab
 }
 
 func (ta *traceAttacher) reuseTracer(tracer *ebpf.ProcessTracer, ie *ebpf.Instrumentable) bool {
+	ie.FileInfo.SetSDKLanguage(ie.Type)
```
grcevski, line 459: Nice catch, I can't believe we had missed the route harvesting on reusing the tracers!

grcevski (approved):
Looks great, I would just add individual unit tests for each function in ruby.go, and maybe we can also rename the file to rails.go (since it's rails only).

pkg/internal/transform/route/harvest/ruby.go
```diff
@@ -0,0 +1,328 @@
...
+	"slices"
+	"strings"
+
+	"go.opentelemetry.io/obi/pkg/appolly/app"
+	"go.opentelemetry.io/obi/pkg/internal/langtools"
+)
+
+const (
+	maxRailsRoutesBytes int64 = 2 * 1024 * 1024
+	maxRailsRouteFiles        = 64
```
grcevski, line 22: 64 might be too little for large rails application, can we make this number couple of hundreds?

pkg/appolly/discover/attacher.go
grcevski: I had to steal this change for my Python harvesting PR, my tests were flaky because of this. Sorry for the conflict I caused.

pkg/internal/transform/route/harvest/ruby.go
```diff
@@ -0,0 +1,328 @@
...
+	if file == nil {
+		return nil, nil, nil
+	}
+	defer file.Close()
+	return scanRailsRoutes(ctx, io.LimitReader(file, maxRailsRoutesBytes))
+}
+
+func scanRailsRoutes(ctx context.Context, reader io.Reader) ([]string, []string, error) {
+	var routes, draws []string
+	scanner := bufio.NewScanner(reader)
```
mmat11, line 140: `bufio.Scanner` defaults to a 64 KiB max token size, so a single line longer than that makes `Scan` stop with `ErrTooLong`, dropping every route for the app, not just the long line. The budget here is already 2 MiB so the token limit can just follow it:

```
scanner := bufio.NewScanner(reader)
scanner.Buffer(nil, int(maxRailsRoutesBytes))
```

if test coverage is missing for this case, I would add it

pkg/appolly/discover/attacher.go
```diff
@@ -456,6 +456,9 @@ func (ta *traceAttacher) loadExecutable(ie *ebpf.Instrumentable) (*link.Executab
...
+	ta.harvestRoutes(ie, true)
```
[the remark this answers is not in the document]
mariomac: Nikola ported already to an earlier PR. This comment does not apply anymore.

site/docs/metrics.md
```diff
@@ -22,7 +22,7 @@ Duration of database operations observed from the server side.
 | `db.operation.name` | string | stable | The name of the operation or command being executed. | findAndModify; HMSET; SELECT |
 | `db.response.status_code` | string | stable | Database response status code. | 102; ORA-17002; 08P01; 404 |
 | `db.system.name` | enum | stable | The database management system (DBMS) product as identified by the client instrumentation. | other_sql; softwareag.adabas; actian.ingres; aws.dynamodb; aws.redshift; azure.cosmosdb; intersystems.cache; cassandra; … |
-| `error.type` | string | stable | Describes a class of error the operation ended with. | timeout; java.net.UnknownHostException; server_certificate_invalid; 500 |
+| `error.type` | enum | stable | Describes a class of error the operation ended with. | timeout; java.net.UnknownHostException; server_certificate_invalid; 500 |
```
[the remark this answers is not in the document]
mariomac: Not apply anymore after merging with Main.

pkg/internal/transform/route/harvest/rails.go
```diff
@@ -0,0 +1,328 @@
...
+			continue
+		}
+		if pending != "" {
+			line = pending + " " + line
+		}
+		if strings.HasSuffix(line, ",") {
+			pending = line
+			continue
+		}
+		pending = ""
```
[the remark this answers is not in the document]
mariomac: I tested the exact scenario described — `resources :users`, followed by `only: [`, `:index,`, `:show`, `]` on separate lines. It does not emit every resource action; the declaration is parsed with an unresolved only: value and railsResourceRoutes returns nil, so the whole resource is safely dropped instead of guessing. I've added a regression test (TestScanRailsRoutes/multiline_array_without_trailing_comma_on_open_bracket) locking in this safe behavior. Full bracket/paren-balanced continuation tracking would be a much larger change, and this harvester is intentionally a best-effort fragment extractor (documented at the top of the file), not a Ruby parser — so I'd rather not take on that scope here given the failure mode already fails safe.

pkg/internal/transform/route/harvest/rails_test.go
```diff
@@ -0,0 +1,227 @@
...
+ end
+end
+`))
+	require.NoError(t, err)
+	matcher := RouteMatcherFromResult(RouteHarvesterResult{Routes: routes, Kind: PartialRoutes})
+	for path, want := range map[string]string{
+		"/admin/users/123":              "/admin/users/:id",
+		"/admin/users/new":              "/admin/users/new",
+		"/admin/users/123/edit":         "/admin/users/:id/edit",
+		"/admin/users/123/comments/456": "/admin/users/:id/comments/:id",
```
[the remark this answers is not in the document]
mariomac: This isn't a bug in the test — it accurately reflects the current design. ruby.go emits resources blocks as independent, unnested fragments (it doesn't track do...end block nesting), and `PartialRouteMatcher.findCombined` stitches fragments together positionally to produce the combined template. The parameter name is a wildcard placeholder for matching purposes only, so the duplicate :id doesn't affect correctness of matching, only the cosmetic label. Making the harvester nesting-aware to get the "real" Rails parameter name would require a structural parser, which conflicts with the fragment-extractor design intent and is out of scope for this PR.

PR #3375, opened 2026-09-09
fix(bpf): write packet-end bounds checks in the form old verifiers mark exactly

pinoOgni:
## Summary

This PR fixes the `offset is outside of the packet` verifier error found on an unpatched 5.8 kernel (arm64). The failing programs are the rdns XDP inspector (`parse_dns_response`) and two tpinjector programs (`obi_packet_extender_write_msg_tp`, `obi_packet_extender_find_existing_tp`).

Verifier output for rdns:
```
; if ((void *)data >= ctx_xdp_data_end(ctx)) {
163: (3d) if r6 >= r1 goto pc+127
 R1_w=pkt_end(id=0,off=0,imm=0) R6=pkt(id=2,off=20,r=20,...)
; const __u8 len = data[0];
165: (71) r5 = *(u8 *)(r6 +0)
invalid access to packet, off=20 size=1, R6(id=2,off=20,r=20)
R6 offset is outside of the packet
```
and for tpinjector:
```
; if (*ptr++ != '\r' || *ptr != '\n') {
443: (71) r1 = *(u8 *)(r2 +69)
invalid access to packet, off=1092 size=1, R2(id=3,off=1092,r=1092)
R2 offset is outside of the packet
```

The issue is verifier bug, not our logic. The fix was added here [2fa7d94afc1a](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=2fa7d94afc1afbb4d702760c058dc2d7ed30f226) "bpf: Fix the off-by-two error in range markings" in 5.16 and backported to 5.10.y/5.15.y but never to 5.8/5.9.

Notes:
- Verified with the verifier suite on 5.8.0/arm64 before and after: the three rejections are gone, no new rejection in rdns, tpinjector or generictracer (which also compiles` tc_common.h`). 
-  CI kernels all carry the backport, so they cannot reproduce the failure.
- Six more guards of the same shape were rewritten too. They did not fail on 5.8 because no packet read depends on the byte they under-prove, but leaving them would leave two shapes in the same files and an allowlist for the follow-up lint check.

Follow-ups:
- `tpinjector` still fails to load on 5.8 for unrelated reasons. I'm working also on those fixes.
- I'm thinking to add a lint check for the open form (I'm open to suggestions)

## Validation

### DIFF
--- a/bpf/rdns/rdns_xdp.c
+++ b/bpf/rdns/rdns_xdp.c
@@ -220,7 +220,7 @@ static __always_inline __u32 validate_qsection(struct xdp_md *ctx, const unsigne
 
     // try at most 16 sections
     for (__u8 i = 0; i < 16; ++i) {
-        if ((void *)data >= ctx_xdp_data_end(ctx)) {
+        if ((void *)(data + 1) > ctx_xdp_data_end(ctx)) {
             return 0;
         }
 
@@ -232,7 +232,7 @@ static __always_inline __u32 validate_qsection(struct xdp_md *ctx, const unsigne
         if (len == 0) {
             size += sizeof(__u32); // account for QTYPE and QCLASS
 
-            if ((void *)(data + sizeof(__u32)) < ctx_xdp_data_end(ctx)) {
+            if ((void *)(data + sizeof(__u32) + 1) <= ctx_xdp_data_end(ctx)) {
                 return size;
             } else {
                 return 0;
@@ -386,7 +386,7 @@ int dns_response_tracker(struct xdp_md *ctx) {
         return XDP_PASS;
     }
 
-    if ((void *)udp + UDP_HDR_SIZE + DNS_HDR_SIZE >= ctx_xdp_data_end(ctx)) {
+    if ((void *)udp + UDP_HDR_SIZE + DNS_HDR_SIZE + 1 > ctx_xdp_data_end(ctx)) {
         return XDP_PASS;
     }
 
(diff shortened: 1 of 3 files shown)


bpf/rdns/rdns_xdp.c
```diff
@@ -232,7 +232,7 @@ static __always_inline __u32 validate_qsection(struct xdp_md *ctx, const unsigne
         if (len == 0) {
             size += sizeof(__u32); // account for QTYPE and QCLASS
 
-            if ((void *)(data + sizeof(__u32)) < ctx_xdp_data_end(ctx)) {
+            if ((void *)(data + sizeof(__u32)) <= ctx_xdp_data_end(ctx)) {
```
mmat11: this changes the logic as well

bpf/rdns/rdns_xdp.c
```diff
@@ -386,7 +386,7 @@ int dns_response_tracker(struct xdp_md *ctx) {
         return XDP_PASS;
     }
 
-    if ((void *)udp + UDP_HDR_SIZE + DNS_HDR_SIZE >= ctx_xdp_data_end(ctx)) {
+    if ((void *)udp + UDP_HDR_SIZE + DNS_HDR_SIZE > ctx_xdp_data_end(ctx)) {
```
mmat11: same here

bpf/rdns/rdns_xdp.c
pinoOgni: let me check

rafaelroquetto (approved):
Hai domato la bestia!

marctc:
> Hai domato la bestia!

AI could never come closer to your approval messages :laughing:

pinoOgni:
> Hai domato la bestia!

Not yet! There are other fixes to make! 🤣

PR #3376, opened 2026-09-09
dotnet: add .NET route harvesting

grcevski:
## Summary

Works very similarly to how we harvest routes for Java applications. It parses the attributes and the dotnet IL and extracts what looks like routes. Uses the microsoft Go version of the IL parser (go-winmd)

## Validation


pkg/internal/transform/route/harvest/dotnet_il.go
```diff
@@ -0,0 +1,264 @@
...
+	var route string
+	for i := 0; i+opcodeSize+tokenSize <= len(code); i++ {
+		switch code[i] {
+		case ldstr:
+			token := binary.LittleEndian.Uint32(code[i+opcodeSize:])
+			if token>>tokenShift != userString {
+				continue
+			}
+			r, ok := dotnetUserString(e.md.US, token&tokenMask)
+			if ok && (strings.HasPrefix(r, "/") || strings.HasPrefix(r, "~/")) {
```
skl: Relative route patterns do not have a `/` prefix, e.g.:

```cpp
app.MapControllerRoute(
  "default",
  "{controller=Home}/{action=Index}/{id?}"
);
```

Suggest either explicitly documenting that conventional relative patterns are unsupported or implementing support for known pattern argument positions.

pkg/internal/transform/route/matcher.go
```diff
@@ -11,8 +11,11 @@ import (
 // wildcard format. By now, we will suppport wildcards in the form:
 // - /user/:userId/details (Gin)
 // - /user/{userId}/details (Gorilla)
-// More formats will be appended at some point
-var wildcard = regexp.MustCompile(`^((:\w*)|(\{\w*}))$`)
+// The new anyPath regex distinguishes ASP.NET catchall paths, like {*path} and {**path}. These match the remaining path, not one segment.
+var (
+	wildcard = regexp.MustCompile(`^((:\w*)|(\{[A-Za-z_][^/{}]*\}))$`)
```
skl: This regex is a little generous in that it would match all of the following to the same route:

```
{id}
{id:int}
{id?}
{id=1}
```

Meaning for example a template of `/items/{id:int}` would match `/items/abc` even though `abc` is not an int (so it would get the wrong route assigned).

Suggest either tightening the regex to explicitly support only plain placeholders or maybe extending the implementation to some kind of parser that handles more of the .NET routing semantics.

skl (approved):
Very cool.

MrAlias (approved):

pkg/internal/transform/route/harvest/dotnet_il.go
grcevski: Fixed! Nice catch.

I added some heuristic for these, it's not perfect because those MapContollerRoute calls have optional arguments that can be strings, so more ldstr, but I added support for those that have {} in them and /.

pkg/internal/transform/route/matcher.go
grcevski: I think getting the right one requires more work on the route matcher. Right now I'm turning /items/{id:int} into /items/{id}. We do something like that on other languages. It's not perfect, but we can fix it in the future IMO if we add matching based on the variable type.

PR #3377, opened 2026-09-09
fix: reject invalid IntEnum ranges

MrAlias:
## Summary

- reject inverted integer ranges used by port and PID discovery criteria
- reject range bounds that overflow the platform integer type before comparing them
- cover inverted ranges and overflow in either endpoint

Supersedes #3177, which was closed due to inactivity. This retains the original signed-off commit and adds the overflow validation requested in its review.

## Validation

- `go test -run 'TestYAMLParse_IntEnum' -count=1 ./pkg/appolly/services`


grcevski (approved):
LGTM!

NimrodAvni78 (approved):

PR #3379, opened 2026-09-09
Make repository AI instructions verifiable

MrAlias:
## Summary

Repository AI instructions are useful only when they remain valid and contributors can confirm that their tools loaded them. This change makes both checks reliable:

- preserve the recursive Markdown glob for `markdownlint-cli2`, ensuring root, hidden, and deeply nested instruction files are validated
- document the tool-native ways to inspect loaded instructions in Codex, Claude Code, GitHub Copilot, and Cursor
- explain when instruction changes require a new or restarted session

## Why

The existing lint target allowed the host shell to expand `**/*.md`. In common shells this selected only one-directory-deep files, skipping root files such as `AGENTS.md` and `CLAUDE.md`, hidden GitHub instructions, and deeper documentation. Contributors also lacked a documented way to distinguish an instruction-loading problem from a model-adherence problem.

Together, these changes provide a reviewable validation path from instruction-file syntax to instruction discovery in supported agent workflows.

## Validation

- `make lint-markdown` — 83 files, no issues
- `git diff --check`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -210,12 +210,12 @@ MARKDOWNIMAGE = $(shell awk '$$4=="markdown" {print $$2}' $(DEPENDENCIES_DOCKERF
 .PHONY: lint-markdown
 lint-markdown:
 	@echo "### Linting markdown"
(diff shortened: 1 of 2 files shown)


grcevski (approved):
LGTM! Let's see if this works.

PR #3381, opened 2026-09-09
Preserve CIDR configuration shape in logs

MrAlias:
Follow-up to #3269.

A CIDR mapping without a `name` is valid, but configuration logging currently emits it as a scalar because both forms produce a definition with an empty name. This changes the logged configuration shape.

This change records when an unnamed CIDR was provided as a mapping and uses that information during YAML marshaling. The collector `confmap` path decodes CIDR sequences through the same shape-aware path, while retaining comma-separated string compatibility.

Regression coverage verifies YAML and JSON output after both `LoadConfig` and `Config.Unmarshal`.

## Testing

- `make verify`


grcevski (approved):
LGTM!

Issue #2904, opened 2026-08-04
Add .NET runtime metrics


marctc:
### Component

OBI (eBPF instrumentation)

### Problem

Provide support for .NET runtime metrics, covering as much as possible what's defined here https://opentelemetry.io/docs/specs/semconv/runtime/dotnet-metrics/

PR #3383, opened 2026-09-09
Add GC collection runtime metrics for .NET

marctc:
## Summary

This PR adds `dotnet.gc.collections` to the existing `application_runtime` feature for .NET 8+ on Linux.

OBI reads `System.Runtime` `EventCounters` through diagnostic `IPC` and `EventPipe`, then exports exclusive gen0, gen1, and gen2 counts through the existing OTLP and Prometheus pipelines. 

The PR is large because it introduces the shared infrastructure for .NET runtime metrics: process attachment, diagnostic IPC, EventPipe sessions, NetTrace decoding, and lifecycle management. Follow-up PRs will reuse this  infrastructure to add more metrics. Other  .NET versions are outside this PR’s scope.

  Part of #2904.

Reviewer note: Please review the NetTrace decoding and bounds checks, process/session lifecycle, and conversion from incremental counters to exclusive generation totals in depth.

## Validation


grcevski:
@marctc just a question before my review, when you say .NET8, do you mean .NET8+ or just .NET 8? I think it should be .NET 8+

pkg/internal/dotnet/socket.go
```diff
@@ -0,0 +1,100 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+var errUnsupportedRuntime = errors.New("unsupported CLR version")
+
+// resolveDiagnosticSocket verifies socket candidates with ProcessInfo2 and
+// selects one .NET 8 runtime matching the expected namespace PID and process start time.
+func resolveDiagnosticSocket(ctx context.Context, tempDir string, namespacePID, startTime uint64) (string, processInfo, error) {
```
grcevski, line 21: How does this work with multiple listeners? Does .NET allow more than one diagnostic consumer to be attached? The reason I'm asking is because we now will attach to any .NET process and I wonder if that prevents the customer to debug their application in production if they have a problem?

If multiple consumers are allowed this is OK, but if they are not then can a customer break the OBI connection and then will we reconnect?

pkg/internal/dotnet/session_manager_linux.go
```diff
@@ -0,0 +1,232 @@
...
+			if !errors.Is(err, unix.ESRCH) {
+				log.Warn("unable to check .NET process liveness", "error", err)
+			}
+			return
+		}
+		setupCtx, cancel := context.WithTimeout(ctx, c.timeout)
+		target, err := resolveDiagnosticTarget(setupCtx, process, file.ServiceAttrs().EnvVars["TMPDIR"])
+		var session *eventPipeSession
+		if err == nil {
+			session, err = startEventPipe(setupCtx, target.socketPath, c.interval)
```
grcevski, line 143: I'm trying to understand here, but I'm not sure if I follow the code exactly. Do we keep one persistent session for the duration of the .NET process being alive or do we reconnect on the timer tick? I think reconnecting (perhaps every 10 seconds) might be a better design, since that way we are not holding up the diagnostic port of the .NET VM, if they disallow multiple consumers at a time.

marctc:
> @marctc just a question before my review, when you say .NET8, do you mean .NET8+ or just .NET 8? I think it should be .NET 8+

I've only tested for .NET 8 so far. .NET 9 still exposes the same EventCounters, so this approach should work there too, but we need to validate it before  extending support. We could also support .NET 9's newer built-in metrics API to collect more runtime metrics.

grcevski:
> > @marctc just a question before my review, when you say .NET8, do you mean .NET8+ or just .NET 8? I think it should be .NET 8+
> 
> I've only tested for .NET 8 so far. .NET 9 still exposes the same EventCounters, so this approach should work there too, but we need to validate it before extending support. We could also support .NET 9's newer built-in metrics API to collect more runtime metrics.

.NET has similar policy to Go, support current - 2, so I think we need to support 8,9,10 and I believe 11 is scheduled for Q42026. https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core

I think the protocol should be backwards compatible, and I doubt they have removed any existing metrics, but it would be good to add a test with .NET 10.

pkg/internal/dotnet/socket.go
marctc: Good question, I didn't think about that actually. I dug and bit a turns out EventPipe [supports multiple simultaneous sessions](https://devblogs.microsoft.com/dotnet/announcing-net-core-3-0-preview-6/#event-pipe-improvements).

pkg/internal/dotnet/session_manager_linux.go
marctc: We keep one persistent session per process:. the timer controls retries after the session ends or attachment fails. Since EventPipe supports multiple consumers, we don't need to release the port periodically.

marctc:
> > > @marctc just a question before my review, when you say .NET8, do you mean .NET8+ or just .NET 8? I think it should be .NET 8+
> > 
> > 
> > I've only tested for .NET 8 so far. .NET 9 still exposes the same EventCounters, so this approach should work there too, but we need to validate it before extending support. We could also support .NET 9's newer built-in metrics API to collect more runtime metrics.
> 
> .NET has similar policy to Go, support current - 2, so I think we need to support 8,9,10 and I believe 11 is scheduled for Q42026. https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core
> 
> I think the protocol should be backwards compatible, and I doubt they have removed any existing metrics, but it would be good to add a test with .NET 10.

I changed integration tests to cover .NET9 and .NET10!

pkg/internal/dotnet/nettrace_decoder.go
```diff
@@ -0,0 +1,202 @@
...
+		return nil, errors.New("uncompressed NetTrace event blocks are unsupported")
+	}
+	var previous netTraceEventHeader
+	var counters []runtimeCounter
+	for reader.Len() > 0 {
+		event, err := readNetTraceCompressedEventHeader(reader, previous)
+		if err != nil {
+			return nil, err
+		}
+		start := len(payload) - reader.Len()
```
grcevski, line 98: I think we need to bound check start here to avoid panics.

pkg/internal/dotnet/nettrace_decoder.go
```diff
@@ -0,0 +1,202 @@
...
+	if len(payload) < netTraceSequencePointHeaderSize {
+		return errors.New("truncated NetTrace sequence point")
+	}
+	count := binary.LittleEndian.Uint32(payload[8:])
+	if uint64(count)*netTraceSequencePointThreadSize != uint64(len(payload)-netTraceSequencePointHeaderSize) {
+		return errors.New("invalid NetTrace sequence point thread count")
+	}
+	for offset := netTraceSequencePointHeaderSize; offset < len(payload); offset += netTraceSequencePointThreadSize {
+		thread := binary.LittleEndian.Uint64(payload[offset:])
+		sequence := binary.LittleEndian.Uint32(payload[offset+8:])
```
grcevski: I think array access is fine, but I think it would be clearer if we could make this +8 here somehow related to netTraceSequencePointThreadSize.

grcevski (approved):
LGTM! This looks good to me now. I left some minor comments. I'll run an AI review now to see if I notices any edge cases and I might follow-up.

pkg/internal/dotnet/socket.go
```diff
@@ -0,0 +1,103 @@
...
+	var selected string
+	var selectedInfo processInfo
+	var failures []error
+	unsupported := 0
+	for _, path := range paths {
+		info, err := queryProcessInfo2(ctx, path)
+		if ctx.Err() != nil {
+			return "", processInfo{}, ctx.Err()
+		}
+		if err != nil {
```
grcevski, line 35: So apparently for NET 6/7, processInfo2 will return HRESULT 0x8013185 which isn't your usupported error, so we'll end up trying to establish the socket forever on 6 and 7.

I dunno how we could test for this, but we need a test somehow.

skl (approved):
lgtm

pkg/appolly/discover/typer.go
```diff
@@ -333,6 +333,8 @@ func (t *typer) asInstrumentable(execElf *exec.FileInfo) ebpf.Instrumentable {
 	t.instrumentableCache.Add(cacheKey{Dev: execElf.Dev(), Ino: execElf.Ino()}, instrumentedExecutable{Type: detectedType, Offsets: nil, InstrumentationError: err})
 	if detectedType == svc.InstrumentablePython {
 		lifecycle.SetRuntimeMetricServiceSource(execElf)
+	}
+	if detectedType == svc.InstrumentablePython || detectedType == svc.InstrumentableDotnet {
```
mmat11, line 337: do we need to `lifecycle.SetRuntimeMetricServiceSource()` here as well?

mmat11 (approved):

pkg/appolly/discover/typer.go
marctc: No,  because Python workers use the parent’s service identity. Each .NET process keeps its own, since the same dotnet executable can run different applications.

PR #3386, opened 2026-09-10
fix(tpinjector): delete tracked socket cookies when the socket closes

mmat11:
## Summary

`tracked_sock_cookies` had inserts but no delete, so ordinary connection churn filled the LRU and evicted the cookies of live long-lived sockets, silently disarming the FIONREAD compensation.

The fix subscribes to socket state changes when a socket is enrolled and removes its cookie from the map once the socket closes. Both maps now forget a socket at the same time.

The `GlobalScaleFactor` option could also shrink the cookie map below the size of `sock_dir`, which cannot be resized. A smaller cookie map is guaranteed to throw out entries for sockets that are still alive, so scaling down no longer shrinks it.

## Validation

### DIFF
--- a/bpf/maps/tracked_sock_cookies.h
+++ b/bpf/maps/tracked_sock_cookies.h
@@ -18,7 +18,9 @@
 #include <common/pin_internal.h>
 
 // Cookies of sockets inserted into sock_dir; the FIONREAD fixup uses them
-// to identify affected sockets. LRU so stale entries age out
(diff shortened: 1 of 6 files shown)


NimrodAvni78 (approved):

grcevski (approved):
LGTM! Nice catch! I wonder if this was the reason of the remaining problems reported with nodejs. Very subtle.

PR #3388, opened 2026-09-10
Fix intermittent unit test failures

grcevski:
## Summary

We consistently get unit test failures because of open port race conditions. I found a new pattern that seems a lot more reliable, we listen to 0, get a random port and when we retrieve what the port was. I ran the new approach on one test 100/100 no failures, while before I got 5 to fail out of a 100.

```
	listener, err := net.Listen("tcp", ":0")
	require.NoError(t, err)
	t.Cleanup(func() { require.NoError(t, listener.Close()) })
	openPort := listener.Addr().(*net.TCPAddr).Port
```

## Validation

### DIFF
--- a/pkg/export/connector/prommgr.go
+++ b/pkg/export/connector/prommgr.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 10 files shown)


mmat11 (approved):
nice!

MrAlias (approved):

PR #3390, opened 2026-09-10
Add `messaging.consumer.group.name` to Kafka consumer spans and metrics

pinoOgni:
## Summary

This PR adds the recommended `messaging.consumer.group.name` on consumer spans and process-duration metrics. 
OBI emitted nothing, so fan-out consumers of one topic were indistinguishable. 

Notes:
- The group id is not in Fetch. It is the first body field of OffsetCommit, OffsetFetch, JoinGroup, Heartbeat, LeaveGroup, SyncGroup and ConsumerGroupHeartbeat, which Java, kafka-python and librdkafka send on a dedicated connection to the group coordinator. Correlation is therefore done per process, not per connection!
- Producers never get the attribute and spans emitted between a mid-stream attach and the next Heartbeat (~3s) legitimately lack it.

Limitations/Follow-ups:
- Go clients: sarama kernel filter and kafka-go. 
- Kernel-side classification on group api keys. Not needed for correctness, would make coordinator-connection classification deterministic.
- Broker-side (KafkaServer) spans get no group: the broker serves all groups and the client's Fetch and group requests arrive on different connections.
- A process that consumes the same topic in two groups gets no attribute for that topic by design.

## Validation


grcevski (approved):
LGTM! Very nice!

grcevski:
You need to sync with main to pick up the test fix.

pkg/ebpf/common/kafka_consumer_groups.go
```diff
@@ -0,0 +1,108 @@
...
+func (g *KafkaConsumerGroups) Lookup(proc KafkaProcess, topic string) string {
+	if g == nil {
+		return ""
+	}
+	if topic != "" {
+		if entry, found := g.lru.Get(kafkaGroupKey{KafkaProcess: proc, Topic: topic}); found {
+			return entry.group
+		}
+	}
+	entry, found := g.lru.Get(kafkaGroupKey{KafkaProcess: proc})
```
NimrodAvni78: I think this check makes sense only when the *topic* is not known and not when the tuple of (proc,topic) is unknown
according to the comment above "the process-level group when the topic is unknown"

pinoOgni: I'll update the comment. 
The fallback covers the case when OBI attaches to a consumer after it joined and we only see heartbeat (it carries no topics) until the next OffsetCommit (maybe never with auto-commit off). The process-level entry is still safe because a process that joined 2 groups is marked ambiguous and resolves to nothing.

MrAlias (commented):
The inferred group has neither the protocol lifetime nor the process lifetime of consumer membership, so Fetch attribution can be wrong or suppressed indefinitely.

pkg/ebpf/common/kafka_detect_transform.go
```diff
@@ -75,14 +93,46 @@ func ProcessKafkaEvent(pkt *largebuf.LargeBuffer, rpkt *largebuf.LargeBuffer, ka
 	case kafkaparser.APIKeyProduce:
 		return processProduceRequest(hdr, kafkaTopicUUIDToName)
 	case kafkaparser.APIKeyFetch:
-		return processFetchRequest(hdr, kafkaTopicUUIDToName)
+		return processFetchRequest(hdr, kafkaTopicUUIDToName, groups, proc)
 	case kafkaparser.APIKeyMetadata:
 		return processMetadataResponse(rpkt, hdr, kafkaTopicUUIDToName)
+	case kafkaparser.APIKeyOffsetCommit, kafkaparser.APIKeyOffsetFetch, kafkaparser.APIKeyJoinGroup,
```
MrAlias, line 99: `OffsetFetch` and `OffsetCommit` do not prove that this process belongs to the named group. Kafka's own admin client builds [OffsetFetch requests](https://github.com/apache/kafka/blob/4.1.0/clients/src/main/java/org/apache/kafka/clients/admin/internals/ListConsumerGroupOffsetsHandler.java) and [OffsetCommit requests](https://github.com/apache/kafka/blob/4.1.0/clients/src/main/java/org/apache/kafka/clients/admin/internals/AlterConsumerGroupOffsetsHandler.java) for operator-selected groups. Passing either request to `Learn` can therefore cache an admin target as this process's consumer group; a later Fetch from a real consumer in the same process is mislabeled or loses the attribute through artificial ambiguity.

Could we establish membership only from membership-bearing APIs, and at most use offset requests to enrich a group already established by those APIs?

pkg/ebpf/common/kafka_detect_transform.go
```diff
@@ -75,14 +93,46 @@ func ProcessKafkaEvent(pkt *largebuf.LargeBuffer, rpkt *largebuf.LargeBuffer, ka
...
+		kafkaparser.APIKeyHeartbeat, kafkaparser.APIKeyLeaveGroup, kafkaparser.APIKeySyncGroup,
+		kafkaparser.APIKeyConsumerGroupHeartbeat:
```
MrAlias, line 101: `LeaveGroup` and `ConsumerGroupHeartbeat` are not always positive membership observations, but both reach `Learn`. A self-leave—or a [heartbeat with member epoch `-1`/`-2`](https://github.com/apache/kafka/blob/4.1.0/clients/src/main/java/org/apache/kafka/common/requests/ConsumerGroupHeartbeatRequest.java#L28-L37)—ends membership, while Kafka's [admin removal handler](https://github.com/apache/kafka/blob/4.1.0/clients/src/main/java/org/apache/kafka/clients/admin/internals/RemoveMembersFromConsumerGroupHandler.java#L39-L86) can issue `LeaveGroup` for operator-selected members. The parsers currently discard member identity and epoch, so the cache can neither remove a departed membership safely nor distinguish an admin request from this process's own membership; leaving A and joining B can instead leave the key ambiguous indefinitely.

Could we preserve enough member and lifecycle data to distinguish joins, self-departures, and admin removals, then update only the affected membership rather than treating every group-naming request as positive learning?

pkg/ebpf/common/kafka_consumer_groups.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// KafkaConsumerGroups remembers, per process, which consumer group consumes which
+// topic, learned from the requests that name topics (JoinGroup, OffsetCommit,
+// OffsetFetch, ConsumerGroupHeartbeat). A process-level entry (empty topic) is learned
+// from every group request and backs Fetch requests whose topic is unknown: KIP-227
+// session fetches carry no topic, a topic UUID may not be resolved yet, and Heartbeat,
+// SyncGroup and LeaveGroup carry no topics at all.
+//
+// Entries are keyed by pid and only expire through LRU eviction: a recycled pid inherits
```
MrAlias: The cache lifetime exceeds the identity used in its key. Once the OS reuses `(namespace, pid)`, a Fetch from the new process can inherit the old process's group before any new group request arrives—the same mid-stream window this fallback is designed to cover. Because `Lookup` refreshes the LRU entry, continued Fetches can keep the stale mapping resident rather than aging it out.

Could we key this by a process-lifetime identity, purge entries on process exit, or add an expiry that active Fetch lookups cannot extend?

pkg/ebpf/common/kafka_detect_transform.go
pinoOgni: Updated. Now membership comes only from `JoinGroup`, `SyncGroup`, `Heartbeat` and `ConsumerGroupHeartbeat` while `OffsetCommit` and `OffsetFetch` are used to just add topics to a group the process already joined and are dropped otherwise.

pkg/ebpf/common/kafka_detect_transform.go
pinoOgni: Nice catch!

Now `LeaveGroup` and a `ConsumerGroupHeartbeat` with epoch `-1/-2` remove that group (and also its topic entries) from the process instead of adding it.

Regarding the self vs admin removal, it is distinguished by `group id` only; so if an admin removes members of its own group we forget and re-learn on the next heartbeat. 

Member-id matching would need the `JoinGroup` response and I left that out (maybe a follow-up?).

pkg/ebpf/common/kafka_consumer_groups.go
pinoOgni: Right! 

I added an `expirable.LRU` with a 2 min TTL that only membership requests renew while `Fetch` attribution reads it without touching the deadline, so a recycled pid inherits the old group for at most 2 min after the previous process stopped heartbeating.

Regarding the purge on process exit, I think it would need the cache on the `Tracer` so `BlockPID` can reach it, we can do it but it wouldn't remove the need for the TTL that's why I just used the TTL.

pinoOgni:
@MrAlias I squashed the fixes into the original commits and rebased. I replied on each thread, I hope it's fine for the review.

pkg/ebpf/common/kafka_consumer_groups.go
```diff
@@ -0,0 +1,211 @@
...
+type KafkaProcess struct {
+	Ns  uint32
+	Pid uint32
+}
+
+type kafkaGroupEntry struct {
+	group string
+	// ambiguous marks a topic consumed by two groups of the same process. Such an entry
+	// never resolves to a group again.
+	ambiguous bool
```
MrAlias: `ambiguous` discards the contributing group IDs, and topic subscriptions are otherwise append-only. If groups A and B consume the same topic and B leaves, `forgetGroup` cannot resolve the entry back to A; if A changes its subscription, the old topics are never removed and can conflict with B later. Another `JoinGroup` cannot repair either case because `learnTopic` returns early for an ambiguous entry, so valid Fetch spans remain unattributed indefinitely.

Could we retain the per-topic contributors, or replace each group's subscription on `JoinGroup`, so departures and rebalances can recompute the current mapping?

pkg/ebpf/common/kafka_consumer_groups.go
```diff
@@ -0,0 +1,211 @@
...
+	state, found := g.lru.Get(proc)
+	if !found {
+		state = newKafkaProcessGroups()
+	}
+	foreign := req.ProtocolType != "" && req.ProtocolType != kafkaparser.ConsumerProtocolType
+	state.groups[req.GroupID] = state.groups[req.GroupID] || foreign
+	if !foreign {
+		state.learnTopics(req.Topics, req.GroupID, kafkaTopicUUIDToName)
+	}
+	g.lru.Add(proc, state) // also on an existing entry: every membership request renews the ttl
```
MrAlias: The TTL belongs to the whole `(namespace, pid)` entry rather than to each observed membership. If the OS reuses a PID within two minutes, the new process gets the old state here; its first group request and every later heartbeat call `Add`, renewing the combined entry, so the previous process's group never expires. This can suppress the single-group fallback or retain stale topic attribution for the lifetime of the new process.

Could we prevent a new process generation from refreshing inherited state, either with a process-lifetime key/purge or membership-specific expiry?

pkg/ebpf/common/kafka_consumer_groups.go
pinoOgni: Right! 
Each membership now keeps its own subscription, replaced on `JoinGroup` (and on a heartbeat that carries one). The topic's group is recomputed at lookup, so leaves and rebalances resolve it.

pkg/ebpf/common/kafka_consumer_groups.go
pinoOgni: Done. Added a membership-specific expiry which is renewed only by requests naming that group

pkg/internal/ebpf/kafkaparser/group.go
```diff
@@ -0,0 +1,514 @@
...
+	for range min(namesLen, maxGroupTopics) {
+		name, err := readString(r, header, false)
+		if err != nil {
+			return req, nil
+		}
+		req.Topics = append(req.Topics, &GroupTopic{Name: name})
+	}
+	// a non-null list is the member's whole subscription (null means unchanged); the
+	// owned partitions appended below are consumed topics, so they belong to it too
+	req.Subscription = namesLen > 0 && namesLen <= maxGroupTopics
```
MrAlias, line 364: `readArrayLength` maps both a null compact array (`0`) and an explicit empty array (`1`) to length zero, so this condition treats both as “unchanged.” Kafka 4.1 [accepts an empty topic-name list when a member joins with a regex](https://github.com/apache/kafka/blob/4.1.0/group-coordinator/src/main/java/org/apache/kafka/coordinator/group/GroupCoordinatorService.java#L403-L407); after a member changes from named topics to a regex, `Join` retains its old topics and can keep another group’s Fetch artificially ambiguous.

Could we preserve whether the array was null and set `Subscription` for every non-null list, including an empty one?

schemas/obi/groups/messaging/metrics.yaml
```diff
@@ -28,6 +28,7 @@ groups:
     stability: development
     attributes:
       - ref: error.type
+      - ref: messaging.consumer.group.name
```
MrAlias, line 31: `make generate-schema-docs` changes both `site/docs/metrics.md` and `site/docs/spans.md` for these registry additions, but neither generated file is included here. Those pages are the published reference for OBI’s emitted telemetry and this drift is intentionally not checked in CI, so could we commit the generated outputs?

pkg/internal/ebpf/kafkaparser/group.go
pinoOgni: Done. Also added a related test ("consumer group heartbeat v0, empty subscription")

schemas/obi/groups/messaging/metrics.yaml
pinoOgni: Added. 

`make generate-schema-docs` also reformats three unrelated `gen_ai.request.*` example values on my weaver build and I left those out form this PR.

pkg/ebpf/common/kafka_consumer_groups.go
```diff
@@ -0,0 +1,263 @@
...
+		return
+	}
+	m.expires = g.now().Add(g.ttl)
+	switch {
+	case m.foreign || (req.ProtocolType != "" && req.ProtocolType != kafkaparser.ConsumerProtocolType):
+		m.foreign = true
+		m.topics = nil // whatever was added while the group passed for a consumer group is dead weight
+	case req.Subscription:
+		m.topics = map[string]struct{}{}
+		m.addTopics(req.Topics, kafkaTopicUUIDToName)
```
MrAlias: The per-group subscription model fixes the stale ambiguity I raised earlier, but I missed one invariant: a process can host multiple members of the same group, each with its own complete subscription. This state is shared by all of them, so when member B joins, it replaces member A’s topics; if the process also belongs to another group for one of those topics, A’s later Fetch is attributed to that other group.

Could we retain subscriptions per member and derive the group topic set from all live members? The parser currently skips `member_id`, so it needs to preserve enough member identity to avoid overwriting another live member.

MrAlias (commented):
The process-level correlation still conflates separate members of the same group, which can emit the wrong consumer-group attribute in multi-consumer processes.

pkg/ebpf/common/kafka_consumer_groups.go
pinoOgni: Sure! Added.

pkg/ebpf/common/kafka_consumer_groups.go
```diff
@@ -0,0 +1,330 @@
...
+	// "consumer" (Kafka Connect, Schema Registry): it coordinates through the same APIs
+	// but is no consumer group. A Heartbeat seen before that JoinGroup leaves it unset.
+	foreign bool
+	// members are the group's members living in the process, by member id. The group's
+	// subscription is the union of theirs.
+	members map[string]*kafkaMember
+}
+
+func (g *kafkaMembership) member(id string) *kafkaMember {
+	m, found := g.members[id]
```
MrAlias, line 102: Legacy consumers can both reach this map with an empty member ID: older brokers accept each initial JoinGroup and return the assigned IDs only in the responses. If two consumers in the same process and group join before either sends a request carrying its assigned ID, the second complete subscription replaces the first in the shared `""` entry. The later SyncGroup or Heartbeat requests carry distinct IDs, but only one subscription survives. If another group consumes the lost topic, a Fetch can then be misattributed to that other group.

Could we retain anonymous subscriptions separately, or conservatively preserve their union until the assigned members rejoin, and cover the overlapping-anonymous-join sequence with a regression test?

pkg/ebpf/common/kafka_consumer_groups.go
```diff
@@ -0,0 +1,330 @@
...
+	"go.opentelemetry.io/obi/pkg/internal/ebpf/kafkaparser"
+)
+
+const (
+	// maxGroupsPerProcess bounds the memberships kept for one process.
+	maxGroupsPerProcess = 64
+	// maxMembersPerGroup bounds the members of one group kept for one process.
+	maxMembersPerGroup = 64
+	// maxTopicsPerMember bounds the subscription kept for one member.
+	maxTopicsPerMember = 1024
```
MrAlias, line 21: Changing the 1,024-topic bound from per group to per member multiplies the per-process maximum from 64 × 1,024 to 64 × 64 × 1,024—about 4.2 million topic-map entries. Repeated OffsetCommit requests, or OffsetFetch v9+ requests, can accumulate distinct topics for each known member, while the default 4,096-entry outer LRU limits only the number of processes, not the weight of each entry.

Could we add a per-process or per-group topic-entry budget, with a test, so member fan-out cannot grow this cache by 64×?

MrAlias (commented):
The per-member model fixes the overwrite for established members, but the revised cache still has a correctness gap for overlapping anonymous members and an aggregate memory-budget gap.

pkg/ebpf/common/kafka_consumer_groups.go
pinoOgni: Sure!
Added a per-process budget of 4096 topic entries across all members that is checked on Join and Enrich. Added also a unit test `TestKafkaConsumerGroupsTopicBudget`.

pkg/ebpf/common/kafka_consumer_groups.go
pinoOgni: Yep. Updated `Join(..)` so anonymous joins now only add, so the group keeps the union until the first named request claims it. Added also a regression test `TestProcessKafkaEventConsumerGroupOverlappingAnonymousJoins`.

grcevski:
I think this should be good to merge now, can we have another review please @MrAlias @NimrodAvni78 ?

NimrodAvni78 (approved):
NICE

pkg/ebpf/common/kafka_consumer_groups.go
```diff
@@ -0,0 +1,362 @@
...
+// The first request of a member unknown so far claims that anonymous state, so it does
+// not outlive the member's leave. Several members may join anonymously before any of
+// them is named (older brokers only): their subscriptions are kept together, see Join.
+func (g *kafkaMembership) member(id string) *kafkaMember {
+	m, found := g.members[id]
+	if found {
+		return m
+	}
+	if id != "" {
+		if pending, found := g.members[""]; found {
```
MrAlias, line 107: Claiming the whole anonymous union for the first named member still loses the second member's subscription later. With anonymous joins A:[orders] and B:[payments], A's first heartbeat moves both topics into A, B's heartbeat creates an empty entry, and A's LeaveGroup removes payments even though B is still live. If another group subscribes to payments, `Lookup` now returns that other group instead of preserving ambiguity.

Could we keep the anonymous union independent until each subscription can be re-established, or otherwise keep lookup conservative after the member that claimed it leaves?

MrAlias (commented):
The anonymous-member handoff still loses live subscription state after the first named member leaves.

PR #3395, opened 2026-09-11
mysql errors followups

mmat11:
## Summary

Follow-up to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3340

- Validate the MySQL packet header before treating a buffer as an ERR packet: a declared payload shorter than marker + error code is rejected. Restores the guard that the [1002, 4167] code-range check used to provide (e.g. a header declaring a 0-byte payload no longer yields a false error span).
- Bound the SQLSTATE and message to the packet length declared in the header instead of reading to the end of the captured buffer. Bytes from a following packet no longer leak into db.response.error; a message cut short by the capture is reported as far as it was captured.
- Only treat # as a SQLSTATE marker when the declared packet has room for a SQLSTATE; otherwise it is message text.
Accept the minimal legacy ERR packet (marker + code, no SQLSTATE, empty message, 7 bytes). MySQLErrMinLen was hardcoded to 8, so this packet was dropped and the failed query reported as successful.
- Tests: cover each case above; give the existing MySQL test vectors real header lengths; restore the "invalid header" case removed in #3340 as an explicit declared-empty-payload → nil assertion.

## Validation

### DIFF
--- a/pkg/internal/sqlprune/mysql.go
+++ b/pkg/internal/sqlprune/mysql.go
@@ -13,7 +13,7 @@ import (
 
 const (
 	MySQLHdrSize                  = 4
-	MySQLErrMinLen                = 8
+	MySQLErrMinLen                = 7
 	MySQLErrPacketMarker   byte   = 0xff
 	MySQLStateMarker       byte   = '#'
 	MySQLProgressReporting uint16 = 0xffff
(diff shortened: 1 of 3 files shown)


MrAlias (approved):

grcevski (approved):
LGTM!

PR #3396, opened 2026-09-11
logenricher: support enriching a subset of workloads with configv2

### LINKED ISSUE #3384: configv2: log_trace_annotation.enabled attaches the log enricher to all processes, ignoring capture selection
### Component

OBI (eBPF instrumentation)

### What happened?

v0.13.0, standalone config v2

With this config:
```
extensions:
  obi:
    version: '2.0'
    capture:
      policy:
        default_action: exclude
      rules:
        - action: include
          match:
            process:
              exe_path_glob:
                - /frontend
                - /backend
    correlation:
      log_trace_annotation:
        enabled: true
```

The trace attacher instruments every process on the host with `logenricher=true`, including processes excluded from capture:

```
msg="instrumenting process" cmd=/usr/bin/sshfs ... type=generic logenricher=true
msg="instrumenting process" cmd=/usr/bin/udevadm ... type=generic logenricher=true
msg="instrumenting process" cmd=/usr/bin/etcd ... type=go logenricher=true
msg="instrumenting process" cmd=/usr/sbin/agetty ... type=generic logenricher=true
```

Cause appears to be the v2 converter mapping enabled: true to a wildcard * ebpf.log_enricher.services selector (internal/config/convert/import.go), which is then evaluated independently of the capture selection instead of intersecting with it.

log enrichment limited to the workloads selected by capture (`/frontend` and `/backend`)

mmat11:
## Summary

Fixes #3384

In config v2, `correlation.log_trace_annotation.enabled` was converted into a wildcard selector. The log enricher selection is meant to be its own selector, evaluated as a subset of the captured workloads, and v2 had no way to express that.

The wildcard did not enrich uncaptured processes (the matcher already refuses log-enricher-only matches), but it logged one "avoiding log instrumentation" line per process on the host and left the selection semantics wrong.

`correlation.log_trace_annotation.match` is now a list of capture-rule match clauses (glob). Each clause becomes one `ebpf.log_enricher.services` selector; the matcher keeps enforcing that annotation only reaches captured workloads.

Integration tests logenricher suite now runs for both v1 and v2 configs

## Validation


internal/test/integration/configs/obi-config-log-enricher-v2.yml
```diff
@@ -0,0 +1,52 @@
...
+                - /basic/:rnd
+              unmatched: path
+              ignored_patterns:
+                - /metrics
+    correlation:
+      log_trace_annotation:
+        enabled: true
+        match:
+          - process:
+              open_ports: "8380,50051,3030,8085,3040,8388"
```
MrAlias, line 46: `TestSuite_LogEnricherDotNetConfigV2` passes port `5266` and the Python-async variant passes `8391`, but this list contains neither. `capture.rules` still admits each process from `OTEL_EBPF_OPEN_PORT`; the independent annotation match does not, so both new v2 suites run with `LogEnricherCriteria` empty and time out looking for enriched trace and span fields.

Could we reuse the suite's port substitution here so the two selections cannot drift?

```suggestion
              open_ports: \"${OTEL_EBPF_OPEN_PORT:-0}\"
```

MrAlias (approved):

PR #3398, opened 2026-09-11
php: Name php apps like OpenTelemetry PHP SDK

grcevski:
## Summary

This change brings similar kinds of ability to name PHP applications like we added for Java, Node, .NET, Python, Deno, Rails.

It mainly focuses on supporting PHP Composer applications, matching what's implemented in https://github.com/open-telemetry/opentelemetry-php/blob/main/src/SDK/Resource/Detectors/Composer.php. I added couple of improvements for Symfony, which is very popular. Namely, if the user didn't change the default composer name after initialising a symfony app with the symfoni CLI, then we fall back on reading the directory.

This PR is largish, but I split everything I did into separate commits with descriptions, so it might be easier to review commit by commit.

## Validation


pkg/internal/phptools/dotenv.go
```diff
@@ -0,0 +1,84 @@
...
+
+const maxDotEnvBytes int64 = 1024 * 1024
+
+func readDotEnvAppName(path string) string {
+	data, _, err := langtools.ReadMetadataFile(path, maxDotEnvBytes)
+	if err != nil || data == nil {
+		return ""
+	}
+
+	for line := range bytes.SplitSeq(data, []byte{'\n'}) {
```
NimrodAvni78, line 23: nit: This returns on the first match of `APP_NAME`. Both vlucas/phpdotenv and symfony/dotenv process the file in order with later assignments overwriting earlier ones, so for

```
APP_NAME=old-name
APP_NAME=new-name
```

the app sees new-name and we report old-name. usefull for appended to .env filrd

internal/test/integration/php_symfony_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"go.opentelemetry.io/obi/internal/test/integration/components/promtest"
+)
+
+const (
+	phpSymfonyURL            = "http://localhost:8081"
+	phpSymfonyServiceName    = "obi/symfony-echo-api"
```
NimrodAvni78, line 19: from other name detectors i saw we split the service name and namespace by /
so in this case service.namespace=obi and service.name=ymfony-echo-api

should this act the same?

pkg/internal/phptools/project_search.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	queue := []searchDirectory{{path: boundary}}
+	var project projectMetadata
+	found := false
+	visited := 0
+
+	for len(queue) > 0 {
+		if visited >= maxProjectSearchDirectories {
+			return projectMetadata{}, false
```
NimrodAvni78, line 87: if exactly one project was already found should we return it instead of discarding it?

pkg/internal/phptools/service_metadata.go
```diff
@@ -0,0 +1,90 @@
...
+	var args []string
+	var cmdlineErr error
+	if !isFPM {
+		_, args, cmdlineErr = cmdlineForPID(fileInfo.Pid())
+	}
+
+	project := findProject(rootDirForPID(fileInfo.Pid()), cwd, args, isFPM)
+	if resolveName {
+		if name, ok := composerName(project.name); ok {
+			fileInfo.SetAutoServiceName(name)
```
NimrodAvni78, line 57: The commit message says this "works mostly with Composer and Laravel" and there's a comment about APP_NAME being typical for Laravel, but I don't think the Laravel path actually works. The Composer name wins unconditionally over APP_NAME and .env, and the laravel/laravel skeleton ships "name": "laravel/laravel" in its own composer.json — composer create-project laravel/laravel my-app doesn't rewrite it, so installed.php has it as the root name too.

So every stock Laravel service reports service.name=laravel/laravel, and two unrelated Laravel apps in the same cluster collide, even when both set APP_NAME in .env.

The symfony/skeleton placeholder is special-cased here already; laravel/laravel is the exact same situation and isn't. Either add it alongside (plus a test — there's no Laravel case in service_metadata_test.go today), or drop the Laravel claim from the commit message.

pkg/internal/phptools/project_search.go
```diff
@@ -0,0 +1,174 @@
...
+			return false, nil
+		}
+		project = candidate
+		found = true
+		return true, nil
+	})
+	return project, found
+}
+
+func scanProcessRoot(root, boundary string) (projectMetadata, bool) {
```
NimrodAvni78, line 74: The early return in ResolveServiceMetadata  only fires when both name and version are already known. For php-fpm, ApplyServiceDefaults fills in the name (php-fpm), but nothing ever fills in service.version, so from the resolver's point of view there is always work left to do and the scan and resolves metadata for every EventCreated instrumentable

The compose file in this PR runs pm=static with pm.max_children=100. If the Composer project isn't found by the parent walk, each of those 100 workers triggers a BFS over up to 512 directories, each costing a ReadDir plus a couple of openats. That's on the order of 50k getdents and 100k openat per discovery sweep, and it repeats on every sweep rather than converging.

Could we cache the result per process root, including the negative result. or another solution, so a container that has no Composer project is scanned once?

pkg/internal/phptools/project_search.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+	}
+
+	if start, info, ok := langtools.StatProcessPath(root, string(filepath.Separator), cwd); ok && info.IsDir() {
+		if project, found := findParentProject(start, boundary); found {
+			return project
+		}
+	}
+
+	if isFPM && processRootDiffersFromHost(root) {
```
NimrodAvni78: why do we need `processRootDiffersFromHost`? does that mean we do the check only if the process does not run in the same namespace as OBI? i.e in non-containerized envs?

mariomac (approved):
Looks good to me! Asking copilot for extra _eyes_.

pkg/internal/phptools/dotenv.go
grcevski: I didn't know, good catch. I'm fixing it.

internal/test/integration/php_symfony_test.go
grcevski: I had the same thought. The PHP OTel SDK seems to take everything as it is in the Composer name and makes it a service name. This test makes it pretty obvious https://github.com/open-telemetry/opentelemetry-php/blob/main/tests/Unit/SDK/Resource/Detectors/ComposerTest.php#L46.

I think they should be split, it makes more sense, but it would mean we'd get different service name from the OTel PHP SDK. I've been trying to match the SDKs semantics 100% where they exist and make up the rules where they don't do anything.

pkg/internal/phptools/project_search.go
grcevski: I guess it would be like playing with fire a bit. We reach the limit of the scan and there are more unvisited directories remaining. If it was some sort of a mono repo, I think we could mess up.

pkg/internal/phptools/service_metadata.go
grcevski: Good point, let me fix this.

pkg/internal/phptools/project_search.go
grcevski: Good idea. I added a caching layer, it ended up being more complex than I thought because apparently some PHP frameworks lets you overwrite the contents of a running app and restart. Less of an issue for containers, more so for Linux hosts.

pkg/internal/phptools/project_search.go
grcevski: So PHP-FPM's workign directory is very hard to find, it has no script argument and no helpful working directory. So I was guarding the scan to avoid scanning the world in non-containerized environments. But you are right, now that I think about this I likely will mess up OBI in a side-car mode in a container. I'll add an extra check for OBI lacking host pid access.

PR #3399, opened 2026-09-12
fix Go JSON-RPC span names

NimrodAvni78:
## Summary

`rpc.method` is the fully-qualified method name, and every semconv example separates service from method with a slash (`EchoService/Echo`). OBI passed the JSON-RPC method through as it appears on the wire, so `Arith.Traceme` reported `rpc.method=Arith.Traceme` and named the span the same.

`JSONRPC.QualifiedMethod` now replaces the last dot with a slash, and the span name, the span's `rpc.method` and the metric label all go through it:

- `Arith.Traceme` -> `Arith/Traceme`
- `com.example.EchoService.Echo` -> `com.example.EchoService/Echo`, the namespace keeps its dots
- `subtract`, naming no service, is unchanged

This applies to the Go `net/rpc` path only. That is the one place a service name is actually observed: the uprobe reads `Request.ServiceMethod`, whose `Service.Method` shape net/rpc defines, and those spans are the only ones reporting version 1.0 — payload extraction accepts 2.0 alone. JSON-RPC itself gives the dot no meaning and takes arbitrary method names, so a payload-extracted `inventory.lookup.v2` stays whole rather than being split at a boundary nothing observed.

Where qualification does change the value, the span also carries `rpc.method_original` with the wire name, which the RPC span contract requires whenever the two differ. The JSON debug printer keeps the wire value too.

gRPC has a separate deviation, reporting `rpc.method` as the wire path `/routeguide.RouteGuide/GetFeature` with a leading slash the examples do not have. It changes every existing gRPC series, so it is left for its own change.

## Compatibility

`rpc.method` changes value on Go `net/rpc` spans and on `rpc.{server,client}.call.duration`, and the span name with it. Anything filtering or grouping on the dotted form needs updating. Spans from payload extraction are untouched. No attribute is removed, and `rpc.method_original` is additive, so there is no schema transformation to record.

## Validation

Neither document records span naming or RPC attribute shapes. Unit tests cover the qualification and each of the three places it is applied, so reverting any one of them fails, and a dotted 2.0 method is asserted to survive untouched at each of them. The RED, span-metrics and k8s assertions move to the slash form; those servers are Go `net/rpc`.

### DIFF
--- a/pkg/appolly/app/request/span.go
+++ b/pkg/appolly/app/request/span.go
@@ -1009,6 +1009,29 @@ type JSONRPC struct {
 	ErrorMessage string `json:"errorMessage,omitempty"`
 }
 
+// jsonRPCReservedPrefix marks the method names JSON-RPC 2.0 reserves for
+// internal methods and extensions, such as OpenRPC's `rpc.discover`. The dot
+// there is part of the reserved namespace, not a service separator.
+const jsonRPCReservedPrefix = "rpc."
+
+// QualifiedMethod returns the method in the shape `rpc.method` is defined as:
+// the fully-qualified name from the RPC interface perspective, whose semconv
+// examples separate the service from the method with a slash
+// ('EchoService/Echo'). JSON-RPC names the service with a dot, so the last one
+// becomes the separator. A method that names no service is returned unchanged.
+func (j *JSONRPC) QualifiedMethod() string {
+	if strings.HasPrefix(j.Method, jsonRPCReservedPrefix) {
+		return j.Method
+	}
+
+	i := strings.LastIndexByte(j.Method, '.')
+	if i <= 0 || i == len(j.Method)-1 {
+		return j.Method
+	}
+
+	return j.Method[:i] + "/" + j.Method[i+1:]
+}
+
 // Generic embedding provider types (Voyage AI, Cohere, Jina AI)
 
 // GenAI operation name constants aligned with OTel semantic conventions.
@@ -2200,7 +2223,7 @@ func (s *Span) TraceName() string {
 
 		if s.SubType == HTTPSubtypeJSONRPC && s.JSONRPC != nil {
 			if s.JSONRPC.Method != "" {
-				return s.JSONRPC.Method
+				return s.JSONRPC.QualifiedMethod()
 			}
 			return "jsonrpc"
 		}
--- a/pkg/export/otel/tracesgen/tracesgen.go
+++ b/pkg/export/otel/tracesgen/tracesgen.go
@@ -549,7 +549,7 @@ func jsonRPCAttributes(span *request.Span) []attribute.KeyValue {
 	rpc := span.JSONRPC
 	attrs := []attribute.KeyValue{
 		semconv.RPCSystemNameJSONRPC,
-		semconv.RPCMethod(rpc.Method),
+		semconv.RPCMethod(rpc.QualifiedMethod()),
 		attribute.String(string(attr.JSONRPCProtocolVersion), rpc.Version),
 	}
 	if rpc.RequestID != "" {
--- a/internal/test/integration/k8s/daemonset_multi_node/k8s_daemonset_multi_node_traces_test.go
+++ b/internal/test/integration/k8s/daemonset_multi_node/k8s_daemonset_multi_node_traces_test.go
@@ -71,14 +71,14 @@ func TestMultiNodeTracing(t *testing.T) {
 					require.Empty(ct, sd, sd.String())
 
 					// Check the information of the Go jsonrpc span
-					res = trace.FindByOperationName("Arith.Traceme", "server")
+					res = trace.FindByOperationName("Arith/Traceme", "server")
 					require.Len(ct, res, 1)
 					parent = res[0]
 					require.NotEmpty(ct, parent.TraceID)
 					require.Equal(ct, traceID, parent.TraceID)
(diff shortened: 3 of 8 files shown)


mariomac (approved):
Good catch!

pkg/appolly/app/request/span.go
```diff
@@ -1009,6 +1009,29 @@ type JSONRPC struct {
...
+	if strings.HasPrefix(j.Method, jsonRPCReservedPrefix) {
+		return j.Method
+	}
+
+	i := strings.LastIndexByte(j.Method, '.')
+	if i <= 0 || i == len(j.Method)-1 {
+		return j.Method
+	}
+
+	return j.Method[:i] + "/" + j.Method[i+1:]
```
MrAlias, line 1037: `JSONRPCSpanFromParsed` feeds this type from every enabled JSON-RPC 2.0 HTTP payload, not only Go `net/rpc`. JSON-RPC permits arbitrary method strings and does not define `.` as a service separator, so a valid method such as `inventory.lookup.v2` is exported and named as `inventory.lookup/v2` even though no service boundary was observed.

Could we scope this conversion to the Go `net/rpc` path, where `Server.readRequestHeader` parses `Service.Method`, and keep generic payload-extracted methods unchanged? A regression with a dotted JSON-RPC 2.0 method would keep the two paths separate.

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -549,7 +549,7 @@ func jsonRPCAttributes(span *request.Span) []attribute.KeyValue {
 	rpc := span.JSONRPC
 	attrs := []attribute.KeyValue{
 		semconv.RPCSystemNameJSONRPC,
-		semconv.RPCMethod(rpc.Method),
+		semconv.RPCMethod(rpc.QualifiedMethod()),
```
MrAlias: When `Arith.Traceme` is converted to `Arith/Traceme`, the [v1.41 JSON-RPC span contract](https://github.com/open-telemetry/semantic-conventions/blob/v1.41.0/docs/rpc/json-rpc.md) used by this repository makes `rpc.method_original` conditionally required whenever it differs from `rpc.method`. The original value remains available as `span.JSONRPC.Method`, but this path does not emit it, so every qualified Go method emits an incomplete span.

Could we append `semconv.RPCMethodOriginal(span.JSONRPC.Method)` when qualification changes and assert both attributes in the trace test?

MrAlias (changes requested):
The new qualification step does not yet preserve JSON-RPC method identity across the Go-specific and generic extraction paths.

pkg/export/otel/tracesgen/tracesgen.go
NimrodAvni78: Im not sure this is the correct usage of `rpc.method_original`. i mean this is the name we parse from go uprobes but that this attribute is meant to allow only a closed set of methods and keep the rpc.method as `_OTHER` and reference the original one there
so I could add it but im not sure its the correct usage of it

Edit: doing more reading it describes
```
- ref: rpc.method_original
  requirement_level:
    conditionally_required: If and only if it's different than `rpc.method`.
```

so you are right in that case, i will implement it

pkg/appolly/app/request/span.go
```diff
@@ -1009,6 +1009,34 @@ type JSONRPC struct {
...
+// examples separate the service from the method with a slash
+// ('EchoService/Echo').
+//
+// Only net/rpc names a service, and it does so with a dot, so the last dot
+// becomes the separator there. JSON-RPC itself assigns the dot no meaning and
+// takes arbitrary method names, so a payload-extracted method is returned as
+// it came off the wire: splitting 'inventory.lookup.v2' would claim a service
+// boundary nothing observed.
+func (j *JSONRPC) QualifiedMethod() string {
+	if j.Version != JSONRPCVersionV1 {
```
MrAlias: When JSON-RPC payload extraction successfully enriches a Go `net/rpc` span, `JSONRPCSpanFromParsed` replaces this value with the payload's version. Our Go test client sends `"jsonrpc":"2.0"` even though the uprobe observed `Service.Method`, so this branch returns `Arith.Traceme` unchanged and skips `rpc.method_original`.

Could we preserve the Go `net/rpc` provenance through enrichment rather than infer it from the protocol version, and add a regression that exercises `HTTPRequestTraceToSpan` with JSON-RPC extraction enabled?

MrAlias:
The provenance fix resolves the earlier Go-specific and generic extraction issues, but the new original-method attribute is outside the current OBI telemetry contract and is keeping the Weaver-validated integration jobs failing.

schemas/obi/groups/http/spans.yaml
```diff
@@ -43,6 +43,9 @@ groups:
       - ref: jsonrpc.protocol.version
       - ref: rpc.system.name
       - ref: rpc.method
+      - ref: rpc.method_original
```
MrAlias, line 46: This registry change is not reflected in `site/docs/spans.md`, so the generated HTTP span reference still omits `rpc.method_original`. Please run `make generate-schema-docs`.

MrAlias (approved):

schemas/obi/groups/http/spans.yaml
NimrodAvni78: done, opened https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3526 to have this fail CI

PR #3401, opened 2026-09-13
gotracer: support nested and reentrant Go gRPC client invocations 

### LINKED ISSUE #3391: gotracer: correlate gRPC client state by invocation
## Component

OBI (eBPF instrumentation)

## What happened?

The Go gRPC client tracer stores active invocation state in `ongoing_grpc_client_requests`, keyed only by goroutine. A nested client call on the same goroutine overwrites the previous invocation, even when both calls are still active. Completion then operates on whichever record occupies the slot, so it can emit the wrong span, lose a span, or end the wrong trace-log context frame.

This is reachable through gRPC client interceptors. `ClientConn.Invoke` calls a user interceptor synchronously, and the interceptor can start another RPC on the same or a different `ClientConn` before allowing the original call to continue. A `(goroutine, connection)` key would still collide when the nested call uses the same `ClientConn`.

Streaming RPCs have an additional identity problem: `ClientConn.NewStream` starts tracking on one call path, while completion occurs through `(*clientStream).RecvMsg` and may not remain tied to the originating goroutine. `ClientConn.Close` is also a connection-wide cleanup signal and cannot identify an individual stream from the caller's goroutine alone.

## Expected behavior

Every gRPC client lifecycle path should operate on the invocation it actually starts, completes, or closes. Overlapping invocations must retain independent telemetry and context state, and a missing or foreign lookup must not consume a different invocation's state.

## Proposed direction

- Preserve a bounded set of active unary invocations per goroutine so nested calls unwind in LIFO order, including recursive calls on the same `ClientConn`.
- After a successful `NewStream`, correlate the invocation with the returned `*clientStream` so `RecvMsg` and other terminal paths can resolve the exact record.
- Associate connection-wide cleanup with the invocations or streams owned by that `ClientConn`; do not treat the goroutine executing `ClientConn.Close` as the invocation owner.
- Pass the resolved invocation's `tp` to `go_obi_ctx__end` rather than ending by kind alone.
- Review the other goroutine-keyed gRPC client maps involved in propagation and connection lookup, updating only those that share the same single-invocation assumption.

The exact map layout is an implementation detail, but it must remain bounded and compatible with the supported eBPF verifier and kernel versions.

## Acceptance criteria

- [ ] Nested unary calls on one goroutine preserve and export both client spans.
- [ ] Recursive unary calls using the same `ClientConn` do not overwrite one another.
- [ ] Nested unary calls using different `ClientConn` instances remain isolated.
- [ ] Streaming completion resolves state by stream identity rather than whichever invocation is stored for the current goroutine.
- [ ] Closing one `ClientConn` does not remove invocation or context state owned by another connection.
- [ ] Each completion removes its own context frame using its trace parent and does not consume another stored frame or same-kind overflow count.
- [ ] Tests cover stored and overflowed gRPC context frames and verify both emitted spans and the context restored after each completion.
- [ ] Relevant BPF unit, verifier, and Go gRPC integration tests pass.

## Additional context

- Follow-up to [PR #3199](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3199), which introduced nested trace-log context tracking and exposed the single-invocation correlation assumption.
- Related review discussion: [ClientConn.Close correlation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3199#discussion_r3960499987)
- gRPC-Go documents that [`ClientConn` can be accessed concurrently](https://github.com/grpc/grpc-go/blob/v1.83.1/Documentation/concurrency.md#clients).
- [`ClientConn.Invoke` calls the configured interceptor synchronously](https://github.com/grpc/grpc-go/blob/v1.83.1/call.go#L29-L38).
- [`ClientConn.NewStream` documents `ClientConn.Close` as a supported cleanup path](https://github.com/grpc/grpc-go/blob/v1.83.1/stream.go#L151-L167).

darkraider01:
Fixes #3391

## Summary

This PR fixes nested and reentrant Go gRPC client tracing by:

* Replacing the single per-goroutine client invocation slot with a bounded LIFO stack, allowing nested and recursive RPCs on the same goroutine without overwriting outer calls.
* Storing transport identity on each invocation frame so connection metadata stays associated with the correct RPC.
* Tracking streaming RPCs by the exact raw `*clientStream` pointer rather than the creator goroutine.
* Handling stream completion races with an early-finish marker plus publish/reconcile protocol.
* Using an atomic single-winner claim so duplicate `clientStream.finish` calls cannot emit duplicate spans.
* Using completed-stream tombstones to suppress late duplicate finish calls.
* Resetting stale stream-generation state at constructor-time `(*clientStream).withRetry`, including protection against Go heap pointer reuse.
* Dynamically scoping constructor-time `withRetry` so unrelated existing streams used inside a `StreamClientInterceptor` cannot corrupt the new stream's identity.

## Validation

Added coverage for:

* Nested calls on the same and different `ClientConn`s.
* Recursive calls and bounded stack overflow.
* Normal streaming completion.
* Cancellation and connection-close races.
* Finish-before-publication and publication races.
* Concurrent / duplicate stream completion.
* Pointer-reuse / completed-tombstone invalidation.
* Stream interceptors operating on unrelated existing streams both before and after `streamer()`.
* Normal stream completion status (`io.EOF` -> OK) versus real errors.

Validation includes BPF unit tests, privileged gRPC tests, and the BPF verifier matrix.


bpf/gotracer/maps/grpc.h
```diff
@@ -22,10 +25,51 @@ struct {
...
+    __type(key, go_addr_key_t); // key: pid + pointer to the clientStream
+    __type(value, u8);
+    __uint(max_entries, MAX_CONCURRENT_REQUESTS);
+    __uint(pinning, OBI_PIN_INTERNAL);
+} tracked_grpc_client_streams SEC(".maps");
+
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __type(key, pid_info);
+    __type(value, u64); // ELF load bias for io.EOF in this process
```
mmat11: what's an ELF load bias for io.EOF?

bpf/gotracer/go_grpc.c
```diff
@@ -383,20 +384,92 @@ int GUARDED_PROG(obi_uprobe_transport_writeStatus, struct pt_regs *, ctx) {
...
+        bpf_dbg_printk("can't reserve space in the ringbuffer");
+        return;
+    }
+
+    task_pid(&trace->pid);
+    trace->type = k_event_type_grpc_client;
+    trace->start_monotime_ns = invocation->start_monotime_ns;
+    trace->go_start_monotime_ns = invocation->start_monotime_ns;
+    trace->end_monotime_ns = bpf_ktime_get_ns();
+    trace->content_length = 0;
```
mmat11: I don't think you need to zero unused fields

bpf/gotracer/go_grpc.c
```diff
@@ -383,20 +384,92 @@ int GUARDED_PROG(obi_uprobe_transport_writeStatus, struct pt_regs *, ctx) {
...
+    bpf_dbg_printk("method_ptr=%lx, method_len=%d", method_ptr, method_len);
+
+    if (!read_go_str_n("method", method_ptr, (u64)method_len, trace->path, sizeof(trace->path))) {
+        bpf_dbg_printk("can't read grpc client method");
+        bpf_ringbuf_discard(trace, 0);
+        return;
+    }
+
+    if (conn && (conn->s_port != 0 || conn->d_port != 0)) {
+        __builtin_memcpy(&trace->conn, conn, sizeof(connection_info_t));
```
mmat11: use `bpf_memcpy`

pkg/internal/goexec/itab.go
```diff
@@ -74,6 +85,205 @@ func findInterfaceImpls(ef *elf.File) (map[string]uint64, error) {
...
+	}
+	opcode := text[i+1]
+	if opcode != 0x8b && opcode != 0x8d && opcode != 0x89 && opcode != 0x39 && opcode != 0x3b &&
+		opcode != 0x81 && opcode != 0x83 {
+		return false
+	}
+	return text[i+2]&0xc7 == 0x05
+}
+
+func arm64PageAddress(text []byte, i int, base uint64) (uint64, bool) {
```
mmat11, line 275: the code in this file needs to be readable by humans, so it requires a short comment explaining what it accomplishes and possibly the constants need to be defined somewhere and commented as well

PR #3402, opened 2026-09-13
php: Implement PHP route harvesting for Symfony, Laravel and Slim

grcevski:
## Summary

I went down the rabbit hole of implementing route harvesting for the last major programming language we are are missing - PHP. It proved to be a lot more complex that I anticipated, since there are many inconsistent ways people define routes with PHP, especially across the frameworks. I decided to support Symfony, Laravel and Slim, which is similar to what the PHP OTel SDK supports, except I added support for some legacy patterns, since OBI is sometimes meant to be used for versions of the software and frameworks that are unsupported by the SDKs.

This PR is very large, there are many tests and test patterns, but the functional code is also very large. I couldn't do simple regex pattern matching, I had to write some rudimentary lexer, to be able to handle the complex route definitions. I considered using a proper AST PHP lexer, but the runtime complexity discouraged me, this is simple linear scan that seems to be good enough to catch all of the cases I found for the 3 frameworks.

So I had three choices:
1. Don't do PHP route harvesting :).
2. Split the work into a series of PRs, but then we might miss the global picture.
3. Split the work into separate commits under the same PR, which is what I've done here.

It would be impossible to review the PR as a whole, but commit, by commit, is like reviewing individual PRs. I added comments in each individual commit that I would normally add to a PR.

While this PR has many files as reported by Github, a lot of those are test data files for my unit tests. They test all the weird combinations of PHP route definitions.

## Validation


pkg/internal/transform/route/harvest/php/symfony_config_yaml.go
```diff
@@ -0,0 +1,124 @@
...
+		routeDefinition := node.Content[pos]
+		if routeDefinition.Kind != yaml.MappingNode {
+			continue
+		}
+
+		if routePath := symfonyYAMLPath(routeDefinition); routePath != "" {
+			routes.add(joinRoutes(routePrefix, routePath))
+		}
+
+		resource := yamlScalar(routeDefinition, "resource")
```
marctc: _I think this misses Symfony’s mapping-form resources (`resource: {path: ../src/Controller/,
  namespace: App\Controller}`). With `prefix: /api, /users/{id}` is harvested without theprefix, so `/api/users/42` fails to match._

pkg/internal/transform/route/harvest/php/slim.go
```diff
@@ -0,0 +1,351 @@
...
+			e.extractGroup(call, variables, prefix)
+			_, bodyEnd := slimCallbackBody(e.tokens, call)
+			if bodyEnd >= 0 {
+				pos = bodyEnd
+			}
+			continue
+		}
+
+		if path, ok := slimRoutePath(call); ok {
+			for _, route := range expandSlimOptionalSegments(joinRoutes(prefix, path)) {
```
marctc: It seems some valid Slim regex routes are harvested but do not match, such as `/users/{id:[0-9]{4}}` and `/news[/{params:.*}]`

marctc (approved):
Great work, thanks for this! Added some minor comments that codex suggested me.

pkg/internal/transform/route/harvest/php/symfony_config_yaml.go
grcevski: Interesting, it seems that this form with `path` is slighty newer and my direct extraction of the yaml scalar came up with "" and we failed to extact the route. I added support for looking up the inner `path` field.

pkg/internal/transform/route/harvest/php/slim.go
grcevski: Wow those are weird, I thought initially about adding support in our parameters expansion, but I decided to keep things simple and make the fixup on the Slim side. We won't be 100% the same as the PHP SDK in this case, but I think it's a fair tradeoff.

pkg/internal/transform/route/harvest/php/laravel.go
```diff
@@ -0,0 +1,339 @@
...
+	for _, call := range calls {
+		switch strings.ToLower(call.name) {
+		case "prefix":
+			value, ok := staticStringArgument(call.arguments, 0)
+			if !ok {
+				value, ok = namedLiteralArgument(call.arguments, "prefix")
+			}
+
+			if !ok {
+				// The prefix appeares to be dynamic and can't be resolved.
```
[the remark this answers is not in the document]
grcevski: Fixed.

pkg/internal/transform/route/harvest/php/lexer.go
```diff
@@ -0,0 +1,194 @@
...
+			escapedCharacter == '\\' ||
+			(quote == '"' && escapedCharacter == '$')
+		if shouldUnescape {
+			value.WriteByte(escapedCharacter)
+			continue
+		}
+
+		// preserve unknown escapes such as \d, which commonly appears in route requirements
+		value.WriteByte('\\')
+		value.WriteByte(escapedCharacter)
```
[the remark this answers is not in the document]
grcevski: Added

pkg/internal/transform/route/harvest/php/laravel_bootstrap.go
```diff
@@ -0,0 +1,88 @@
...
+		if !hasLaravelAPIRouteFile(arguments) {
+			return ""
+		}
+
+		if _, configured := namedArgument(arguments, "apiPrefix"); !configured {
+			return "api"
+		}
+
+		prefix, _ := namedLiteralArgument(arguments, "apiPrefix")
+		return prefix
```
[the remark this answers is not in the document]
grcevski: Fixed.

pkg/internal/transform/route/harvest/php/symfony_config.go
```diff
@@ -0,0 +1,29 @@
...
+//		      -> direct path routes
+//		      -> YAML resource: extractSymfonyYAMLFile
+//		      -> attribute resource: addImportedSymfonyAttributes
+func extractSymfonyConfigRoutes(projectRoot string, attributes map[string][]string, routes routeSet) map[string]struct{} {
+	traversal := newSymfonyConfigTraversal()
+	imported := map[string]struct{}{}
+
+	for _, entry := range symfonyYAMLEntries(projectRoot) {
+		extractSymfonyYAMLFile(projectRoot, entry, "", traversal, attributes, imported, routes)
+	}
```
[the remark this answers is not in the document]
grcevski: Fixed.

pkg/internal/transform/route/harvest/php/laravel_resource.go
```diff
@@ -0,0 +1,320 @@
...
+	return collection, member
+}
+
+func resourceParameter(resource string) string {
+	name := resource
+	if strings.HasSuffix(name, "ies") && len(name) > 3 {
+		name = strings.TrimSuffix(name, "ies") + "y"
+	} else {
+		name = strings.TrimSuffix(name, "s")
+	}
```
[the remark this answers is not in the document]
grcevski: I removed the singularization completely, this I think is out of scope for this PR. The OBI route won't match 100% what PHP SDK will see on laravel, but reimplementing `Str::singular()` is going to make this PR even larger.

pkg/internal/transform/route/harvest/php/slim.go
grcevski: Fixed.

pkg/internal/transform/route/harvest/php/symfony_config_yaml.go
grcevski: Fixed.

PR #3406, opened 2026-09-14
fix(ebpf): truncate HTTP method at null byte

NameHaibinZhang:
## Problem

`httpMethodFromBuf` parsed the HTTP method from the raw eBPF buffer without truncating at null bytes. In contrast, `httpURLFromBuf` and `httpHostFromBuf` already do this. When the request line is followed by zero padding, the method could contain embedded nulls (e.g. `G\\x00\\x00`). That corrupted value then became the `http.request.method` metric label, inflating metric cardinality.

## Change

- Truncate the buffer to the first null byte before extracting the method in `httpMethodFromBuf`.
- Add a regression test for request lines with trailing null bytes.

## Related code

- `pkg/ebpf/common/http_transform.go`
- `pkg/ebpf/common/http_transform_test.go`


NimrodAvni78 (approved):
Nice

PR #3410, opened 2026-09-14
Add Django route harvesting

marctc:
## Summary

  This PR adds Django route harvesting to the Python harvester introduced in #3366.

  - Extracts literal `path()` declarations.
  - Recursively follows local `include("checkout.urls")` references across URL configuration files.
  - Combines parent and child prefixes, including apps mounted under multiple prefixes.
  - Supports routes wrapped in `i18n_patterns()`.
  - Adds standard `admin.site.urls` templates under the configured prefix, including login and model actions.
  - Supports path parameters followed by a suffix, such as `<path:object_id>/change/`.

## Validation


pkg/internal/transform/route/part_matcher.go
```diff
@@ -91,8 +91,8 @@ func (rm *PartialRouteMatcher) findPartial(tokens []string, root *node) (string,
 }
 
 func (rm *PartialRouteMatcher) findPartialRecursive(tokens []string, node *node, consumed int) (string, int) {
-	if node.AnyPath != nil && node.FullRoute != "" {
-		return node.FullRoute, len(tokens)
+	if node.AnyPath != nil && node.AnyPath.FullRoute != "" {
```
grcevski, line 94: Hm, why this change?

pkg/internal/transform/route/harvest/python.go
```diff
@@ -124,13 +134,32 @@ func skipPythonDir(name string) bool {
 	return ok
 }
 
-func scanPythonFile(path string, routes map[string]struct{}) error {
+func scanPythonFile(path string, routes map[string]struct{}) ([]djangoRoute, error) {
```
grcevski, line 137: Could we find another way maybe to make this more generic and fit nicely? Right now the django routes are returned from a common function that scans for all python route paths, but only if it's django.

mariomac (approved):
Looks ok! Asking copilot for review

pkg/internal/transform/route/harvest/python_django.go
```diff
@@ -0,0 +1,229 @@
...
+			if depth == 0 {
+				return i
+			}
+		}
+	}
+	return -1
+}
+
+func scanDjango(stmt string, aliases map[string]string) []djangoRoute {
+	if wrapper := djangoI18nStart.FindStringIndex(stmt); wrapper != nil {
```
grcevski, line 96: Can you show an example statement that is parsed here, does it have to be recursive? Do they nest multiple patterns in one another?

pkg/internal/transform/route/harvest/python_django.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	return routes
+}
+
+// indexDjangoModules maps dotted include targets, such as "checkout.urls", to
+// scanned Python files relative to the application root.
+func indexDjangoModules(root string, files map[string][]djangoRoute) map[string]string {
+	modules := make(map[string]string, len(files))
+	for file := range files {
+		rel, err := filepath.Rel(root, file)
```
grcevski, line 153: we should use the langtools helpers for this, if file has .. this might escape the root of the scanned directories.

grcevski: We also need to ensure they are not symlinks.

pkg/internal/transform/route/harvest/python.go
```diff
@@ -53,22 +53,32 @@ func ExtractPythonRoutes(fi *exec.FileInfo) (*RouteHarvesterResult, error) {
 func extractPythonRoutes(dir string) (*RouteHarvesterResult, error) {
 	log := slog.With("component", "route.harvester.python")
 	routes := map[string]struct{}{}
+	djangoFiles := map[string][]djangoRoute{}
+	kind := PartialRoutes
 	err := walkPythonFiles(dir, func(path string) error {
```
mmat11, line 58: the internal function never returns an error, I think we should either return the error and log it in the caller or just remove any return from the signature

pkg/internal/transform/route/harvest/python_django.go
```diff
@@ -0,0 +1,229 @@
...
+
+		// A literal include, such as include("checkout.urls"), names the module directly.
+		includeModule := match[3]
+		if includeModule == "" {
+			includeModule = match[4]
+		}
+
+		// A non-literal include can refer to an imported alias, such as checkout_urls.
+		if match[6] != "" {
+			includeModule = aliases[match[7]]
```
mmat11, line 131: are these indexes always present?

pkg/internal/transform/route/part_matcher.go
marctc: This is needed because both matchers use `appendRoute()`. Supporting Django’s `<path:object_id>/change/` moved catch all templates into `AnyPath.FullRoute`, so the partial matcher needs to read from there too. Without this adjustment,  `TestPartialMatcherPythonPathParams` fails

pkg/internal/transform/route/harvest/python_django.go
marctc: Yeah you are right,  recursion isn’t needed here. I used it to reuse the scanner inside and outside `i18n_patterns()`, but I overcomplicated it.

pkg/internal/transform/route/harvest/python_django.go
marctc: Yes, the regex has seven capture groups, so these indexes are always present in a match. Optional groups that don’t match contain empty strings. Added comments to clarify what it means each.

pkg/internal/transform/route/part_matcher.go
grcevski: OK, but now I think I'm confused if we still need node.FullRoute and now is that different than node.AnyPath.FullRoute?

grcevski: OK, after reading the code I think I get it. This was a bug from before with the anypath route matching I added, it would falsely stop to a partial route.

grcevski (approved):
LGTM!

mmat11 (approved):

PR #3412, opened 2026-09-14
fix(metrics)!: label the host id host_id on traces_host_info

NimrodAvni78:

## Summary

`traces_host_info` is the only metric in either exporter that calls the host id `cloud_host_id`. The OTLP exporter reports the same value as `host.id` on `traces.host.info`, Prometheus `target_info` already labels it `host_id`, and the registry declares `host.id` on the metric. This aligns the last holdout.

`CloudHostIDKey` stays the override point and keeps its name, so a component vendoring OBI that assigns to it is unaffected — only the default changes.

## Breaking change

Dashboards, alerts and recording rules selecting `traces_host_info{cloud_host_id="..."}` return no series after upgrade and must select `host_id` instead. Recorded under "Pending release notes" in `devdocs/telemetry-schema.md`: the schema format describes the OTLP output only, so a Prometheus label rename has no transformation that can express it.

## Validation

`TestAppMetricsExpiration` asserts the exported label. The assertion is anchored (`traces_host_info\{host_id="my-host"\}`) rather than `.*host_id=`, which would match `cloud_host_id` too — I verified it fails when `CloudHostIDKey` is reverted to the old literal.

## Validation


mariomac (approved):
Looks good!

grcevski (approved):
LGTM!

PR #3413, opened 2026-09-14
fix(metrics)!: declare span, service graph and info metric names once and move the span-metric host id onto the resource

NimrodAvni78:
## Summary

Span metrics, service graph metrics and the three info metrics were the last metrics named by hand in both exporters. They are now declared once in `pkg/export/attributes`, with the Prometheus name derived from the OTLP definition via `otlptranslator`. No metric is renamed: all 13 derivations reproduce today's published names, pinned in `TestPrometheusNames`.

OTLP span metrics also stop carrying `host.id` per data point. It lives on the resource, which both exporters carry, filter identically, and render as `target_info{host_id}`.

## Breaking changes

- Span metrics drop the `host.id` data point attribute. Resource-reading OTLP consumers are unaffected; a consumer flattening to Prometheus loses `host_id` as a span-metric label.
- That host id is now filterable through `attributes.select.resource`, where the data point copy was unconditional. Nothing is filtered unless a `resource` section is configured.
- Nine exported identifiers removed from `pkg/export/otel`: four `SpanMetrics*` aliases, four `ServiceGraph*`, and `CloudHostIDKey`. All duplicated declarations the package now reads directly, and unreferenced outside it.
- 18 exported identifiers become `var` rather than `const`, since a derived name is a function call.

The attribute drop is recorded under "Pending release notes" in `devdocs/telemetry-schema.md`, a new section for changes the schema format cannot express. `AGENTS.md` routes renames and removals to the two lists accordingly.

## Notes

`ExposedName` is dropped from five OTLP-path `Field` literals: only the Prometheus exporter reads it, and only off `Field[T, string]`.

`site/docs/` is regenerated. The `error.type` and `dns.question.name` deltas are the resolution non-determinism `scripts/generate-schema-docs.sh` documents, normalised to the `x.obi.error` override; `registry.obi.exception` and `obi.http.response.observed` are catch-up from earlier registry commits.

## Validation


grcevski (approved):
LGTM!

mmat11 (approved):
lgtm, needs rebase

MrAlias (approved):

Issue #1146, opened 2026-01-20
AWS Cloud services


MrAlias:

mariomac:
@MrAlias 
We currently provide the following resource metadata for AWS services:
```
cloud_account_id
cloud_availability_zone
cloud_platform
cloud_provider
cloud_region
host_id
host_image_id
host_type
```

Do you think it's enough to close this issue, or are we thinking on other kind of instrumentation (e.g. internal traces/metrics of the different AWS services)

Analogue question for https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1145 and https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1147

MrAlias:
Thanks. The AWS resource metadata is complete, but it describes the host/VM running OBI rather than calls to AWS services as far as I can remember.

I though the intended scope of this issue was to instrument AWS S3, SQS, and Bedrock, or did we intend to support additional AWS service APIs/SDKs?

If S3, SQS, and Bedrock were the intended scope, we can document that and close this issue. Otherwise, we need to identify the initial AWS services and expected telemetry for the clients so we can turn this broad placeholder into focused follow-up issues.

marctc:
> I though the intended scope of this issue was to instrument AWS S3, SQS, and Bedrock, or did we intend to support additional AWS service APIs/SDKs?

The parent issue is about adding more protocol support, so I guess is about SQS and such. I suggest we break down the issue for each of the protocols you mention or we expand the description to explain definition of done of the issue.

PR #3415, opened 2026-09-14
Add support for AWS SNS

marctc:
## Summary

This PR adds support for capturing [AWS SNS](https://aws.amazon.com/sns/) client spans, including Publish, PublishBatch, and topic and subscription operations.

It uses the existing AWS HTTP payload extraction to capture data and it requires  `OTEL_EBPF_HTTP_AWS_ENABLED=true` to be enabled.

Related to #1146

## Validation


mariomac (commented):
Looks great! Asking copilot for some extra _eyes_.

internal/test/integration/red_test_python_aws_sns.go
```diff
@@ -0,0 +1,94 @@
...
+	query := "?topic_arn=" + url.QueryEscape(topic.TopicARN)
+	published := snsRequest(t, "/publish"+query)
+	require.NotEmpty(t, published.MessageID)
+	batch := snsRequest(t, "/publishbatch"+query)
+	require.Len(t, batch.Successful, 2)
+	require.Empty(t, batch.Failed)
+	snsRequest(t, "/gettopicattributes"+query)
+	snsRequest(t, "/deletetopic"+query)
+
+	require.EventuallyWithT(t, func(ct *assert.CollectT) {
```
NimrodAvni78, line 44: could we also validate here that we see messaging metric data?
i think this is mistakenly wired to http metric instead of messaging

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -831,6 +831,30 @@ func traceAttributesSelectorInternal(span *request.Span, optionalAttrs map[attr.
...
+				attrs = append(attrs, request.MessagingMessageID(sns.MessageID))
+			}
+			if sns.OperationName == "PublishBatch" {
+				attrs = append(attrs, semconv.MessagingBatchMessageCount(sns.BatchCount))
+			}
+			if sns.Meta.RequestID != "" {
+				attrs = append(attrs, semconv.AWSRequestID(sns.Meta.RequestID))
+			}
+			attrs = append(attrs, semconv.CloudRegion(sns.Meta.Region))
+			if sns.ErrorCode != "" {
```
NimrodAvni78: every other place adding errorType validates it was selected in attribute selection

```if _, ok := optionalAttrs[attr.ErrorType]; ok && !hasAttribute(...))```

but this misses it. can we make sure this is gated on the attribute selection?

File: pkg/appolly/app/request/span_getters.go [hunk 3]
```diff
@@ -399,6 +416,20 @@ func spanOTELGetters(name attr.Name) (attributes.Getter[*Span, attribute.KeyValu
...
+	case attr.AWSSNSTopicARN:
+		getter = func(s *Span) attribute.KeyValue {
+			if s.Type == EventTypeHTTPClient && s.SubType == HTTPSubtypeAWSSNS && s.AWS != nil && s.AWS.SNS.TopicARN != "" {
+				return semconv.AWSSNSTopicARN(s.AWS.SNS.TopicARN)
+			}
+			return attribute.KeyValue{}
+		}
+	case attr.MessagingBatchCount:
+		getter = func(s *Span) attribute.KeyValue {
+			if s.Type == EventTypeHTTPClient && s.SubType == HTTPSubtypeAWSSNS && s.AWS != nil && s.AWS.SNS.OperationName == "PublishBatch" {
```
NimrodAvni78: can we also validate `s.AWS.SNS.BatchCount > 0`

NimrodAvni78 (approved):

pkg/ebpf/common/http/aws_sns.go
```diff
@@ -0,0 +1,169 @@
...
+		params, err := url.ParseQuery(req.URL.RawQuery)
+		return params, err == nil
+	case http.MethodPost:
+		mediaType, _, err := mime.ParseMediaType(req.Header.Get("Content-Type"))
+		if err != nil || mediaType != "application/x-www-form-urlencoded" {
+			return nil, false
+		}
+		body, err := io.ReadAll(req.Body)
+		req.Body = io.NopCloser(bytes.NewReader(body))
+		if err != nil {
```
MrAlias: A valid SNS Publish can still reach this branch at the maximum capture setting: SNS permits a 262,144-byte message, while the form also carries fields such as Action and TopicArn and percent-encoding can expand the request beyond OBI's 262,144-byte ceiling. `io.ReadAll` then returns an error, this detector declines the span, and the call remains generic HTTP without the new SNS attributes or messaging metrics. Could we retain enough parsed fields to classify endpoint-verified requests when the captured body is truncated, and cover a large Publish in the tests?

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -831,6 +831,27 @@ func traceAttributesSelectorInternal(span *request.Span, optionalAttrs map[attr.
 			attrs = append(attrs, request.AWSSQSQueueURL(sqs.QueueURL))
 		}
 
+		if span.SubType == request.HTTPSubtypeAWSSNS && span.AWS != nil {
+			sns := span.AWS.SNS
+			attrs = append(attrs, semconv.MessagingSystemAWSSNS, request.MessagingOperationName(sns.OperationName))
```
MrAlias: The pinned v1.41 SNS conventions extend the AWS SDK conventions, which require AWS API RPC-system identity. OBI emits the replacement key as `rpc.system.name="aws-api"` through `request.RPCSystem`; the S3 block above does this, but the SNS block omits it. Could we add the attribute here and assert it in `TestSNSAttributes`?

pkg/appolly/app/request/span.go
```diff
@@ -2072,6 +2118,10 @@ func (s *Span) TraceName() string {
 			}
 		}
 
+		if s.Type == EventTypeHTTPClient && s.SubType == HTTPSubtypeAWSSNS && s.AWS != nil {
+			return "sns." + s.AWS.SNS.OperationName
```
MrAlias, line 2122: The inherited v1.41 AWS SDK convention requires span names in `Service.Operation` form, using the AWS service name (for example, `S3.ListBuckets`). Returning `sns.Publish` instead of `SNS.Publish` makes every new SNS span nonconforming, and `TestSNSAttributes` currently locks in that lowercase value. Could we use the uppercase service name here and update the assertion?

MrAlias (commented):
The SNS enrichment still has correctness gaps at the capture boundary and in its inherited AWS SDK span conventions.

pkg/ebpf/common/http/aws_sns.go
```diff
@@ -0,0 +1,169 @@
...
+		// Request metadata remains useful when the response body is truncated.
+		response = awsSNSResponse{}
+	}
+
+	host := extractHostname(req)
+	endpoint := snsEndpoint.FindStringSubmatch(strings.ToLower(host))
+	topicARN := params.Get("TopicArn")
+	topic := snsTopicARN(topicARN)
+	namespace := strings.TrimPrefix(strings.TrimPrefix(response.XMLName.Space, "https://"), "http://")
+	if len(endpoint) == 0 && topic == nil && namespace != "sns.amazonaws.com/doc/2010-03-31/" {
```
mmat11: is this hardcoded for all response documents?

pkg/ebpf/common/http/aws_sns.go
```diff
@@ -0,0 +1,169 @@
...
+		topicARN = response.CreateTopic.TopicARN
+		topic = snsTopicARN(topicARN)
+	}
+
+	// SNS also returns request IDs in XML, including on Query API errors.
+	meta, _ := parseAWSMeta(req, resp)
+	if meta.RequestID == "" {
+		meta.RequestID = response.Meta.RequestID
+		if meta.RequestID == "" {
+			meta.RequestID = response.RequestID
```
mmat11, line 79: can this be empty as well?

pkg/ebpf/common/http/aws_sns.go
```diff
@@ -0,0 +1,169 @@
...
+	endpoint := snsEndpoint.FindStringSubmatch(strings.ToLower(host))
+	topicARN := params.Get("TopicArn")
+	topic := snsTopicARN(topicARN)
+	namespace := strings.TrimPrefix(strings.TrimPrefix(response.XMLName.Space, "https://"), "http://")
+	if len(endpoint) == 0 && topic == nil && namespace != "sns.amazonaws.com/doc/2010-03-31/" {
+		return *baseSpan, false
+	}
+	if topicARN == "" {
+		topicARN = response.CreateTopic.TopicARN
+		topic = snsTopicARN(topicARN)
```
mmat11: this can return a nil slice, I think we should check it

pkg/appolly/app/request/span.go
marctc: S3 and SQS currently use lowercase service names (`s3.*` and `sqs.*`). If we change only SNS, the AWS span names become inconsistent.

I propose that we keep sns.* here and update all three services in a separate change (and we document it as breaking change). Does that work for you?

pkg/ebpf/common/http/aws_sns.go
marctc: Good point, I couldn't find strong evidence for it (other than [examples](https://docs.aws.amazon.com/sns/latest/api/API_CreateTopic.html)), so let's remove it.

pkg/ebpf/common/http/aws_sns.go
marctc: Yes, all three sources can be empty. In that case the request ID remains empty, and the exporter omits the attribute.

pkg/ebpf/common/http/aws_sns.go
marctc: Yes, it can return nil. It is handled by the check immediately below, and later accesses are guarded with `topic != nil`.

mmat11 (approved):
lgtm

pkg/ebpf/common/http/aws_sns.go
```diff
@@ -0,0 +1,164 @@
...
+	if !ok || !snsOperation(params.Get("Action")) {
+		return *baseSpan, false
+	}
+	// Topic publishing is the supported destination type. SMS and mobile push
+	// use different destination semantics.
+	if params.Has("PhoneNumber") || params.Has("TargetArn") {
+		return *baseSpan, false
+	}
+
+	responseBody, err := io.ReadAll(resp.Body)
```
grcevski: I don't know much about SNS, are these relatively small messages? I'm wondering if this io.ReadAll can cause massive memory consumption, can we use a limitted reader instead and/or reject payloads that are too large?

grcevski (approved):
LGTM! I left one minor comment about the ReadAll.

pkg/appolly/app/request/span.go
MrAlias: I agree that changing the existing S3 and SQS names should be handled separately and called out as a compatibility change. SNS has not shipped yet, though, so there is no existing contract to preserve here; using `sns.*` would introduce a new span name that knowingly violates the inherited `Service.Operation` requirement. I'd use `SNS.*` in this PR, then fix the existing S3/SQS names in a follow-up.

pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -831,6 +831,31 @@ func traceAttributesSelectorInternal(span *request.Span, optionalAttrs map[attr.
 			attrs = append(attrs, request.AWSSQSQueueURL(sqs.QueueURL))
 		}
 
+		if span.SubType == request.HTTPSubtypeAWSSNS && span.AWS != nil {
+			sns := span.AWS.SNS
+			attrs = append(attrs,
+				request.RPCSystem("aws-api"),
```
MrAlias, line 837: I missed this in the previous pass: the inherited v1.41 AWS SDK convention also recommends fully qualified `rpc.method` because `rpc.service` is deprecated. With only `rpc.system.name`, SNS spans lack their AWS operation identity in the RPC attributes; the S3 block above emits `S3/<operation>` for the same reason.

Could we add `SNS/<operation>` here and assert it in `TestSNSAttributes`?

```suggestion
				request.RPCSystem("aws-api"),
				semconv.RPCMethod("SNS/" + sns.OperationName),
```

mariomac (approved):
LGTM!

PR #3417, opened 2026-09-14
fix(mysql): skip no-response commands coalesced into request buffers

NameHaibinZhang:
## Problem

MySQL spans are never emitted for applications that use native prepared statements (e.g. PHP PDO with `PDO::ATTR_EMULATE_PREPARES => false`, which issues `PREPARE` → `EXECUTE` → `COM_STMT_CLOSE` per query).

`COM_STMT_CLOSE` and `COM_STMT_SEND_LONG_DATA` have no server response, so the kernel keeps them pending and appends the next command sent on the connection into the same event's request buffer. The MySQL userspace parser only looked at the first packet's command byte, so an event whose request buffer starts with `COM_STMT_CLOSE` (command ID 25) hit the `unhandled commandID` path and fell back to the generic handler, which cannot match binary packets — no span was produced.

This also cascades: the coalesced command (e.g. `COM_STMT_PREPARE`) never updated the prepared-statement cache, so later clean `COM_STMT_EXECUTE` events hit "unknown statement ID" and were dropped as well. Debug logs show the telltale pair:

```
MySQL command ID unhandled commandID=25
falling back to generic handler protocol=MySQL
```

## Change

- Skip leading no-response commands (`COM_STMT_CLOSE`, `COM_STMT_SEND_LONG_DATA`) in the request buffer before parsing, so the command paired with the response buffer is the one processed (`pkg/ebpf/common/sql_detect_mysql.go`).
- Add unit tests for the skip helper and end-to-end tests covering `STMT_EXECUTE`, `STMT_PREPARE` and `QUERY` events coalesced behind a `COM_STMT_CLOSE` packet.

## Related code

- `pkg/ebpf/common/sql_detect_mysql.go`
- `pkg/ebpf/common/sql_detect_mysql_test.go`


pkg/ebpf/common/sql_detect_mysql.go
```diff
@@ -64,6 +70,34 @@ func isMySQL(b *largebuf.LargeBuffer) bool {
...
+		if !mysqlNoResponseCommand(hdr.command) {
+			return reqRaw
+		}
+
+		packetLen := sqlprune.MySQLHdrSize + int(hdr.length)
+		if packetLen <= sqlprune.MySQLHdrSize || packetLen >= len(reqRaw) {
+			return reqRaw
+		}
+
+		reqRaw = reqRaw[packetLen:]
```
MrAlias, line 106: `COM_STMT_CLOSE` deallocates the statement, but this loop discards the packet without removing `(event.ConnInfo, stmtID)` from `mysqlPreparedStatements`. With the native prepare/execute/close workload this PR targets, closed entries fill the global LRU; enough churn can evict a still-live statement, and its next `COM_STMT_EXECUTE` falls into the existing “unknown statement ID” path and loses the span.

We should surface each skipped close to `handleMySQL` and remove its cache entry. The new `close(1) + execute(1)` test should also use different IDs (or assert the closed ID is gone), since executing a closed statement ID is not a valid server flow.

NameHaibinZhang: Thanks for catching this. You're right: skipping COM_STMT_CLOSE without removing its prepared-statement cache entry could leave closed statements in the shared LRU and evict live entries.

I've updated the skipped-command handling to return statement IDs from fully framed COM_STMT_CLOSE packets, and handleMySQL now removes the corresponding connection-scoped cache entries before parsing the response-paired command. COM_STMT_SEND_LONG_DATA remains skipped but does not remove a cache entry.

I also replaced the invalid close(1) + execute(1) test flow with close(1) + execute(2), and added assertions that the closed statement IDs are removed from the cache.

MrAlias (approved):
🚀

grcevski (approved):
LGTM!

PR #3418, opened 2026-09-14
uprobes: use uprobe_multi when available and detach uprobe links concurrently on shutdown

mmat11:
## Summary

OBI takes minutes to exit after SIGTERM on recent kernels. The process is an unkillable zombie until the last probe is gone.

Linux 6.13* stopped taking `uprobe->register_rwsem` on every probe hit and protects the consumer list with RCU instead: every `uprobe_unregister_sync()` now waits for an RCU tasks trace grace period, on top of the grace periods perf events already paid.

*source: https://github.com/torvalds/linux/commit/04b01625da13, https://github.com/torvalds/linux/commit/cc01bd044e6a, https://github.com/torvalds/linux/commit/87195a1ee332, https://github.com/torvalds/linux/commit/dd1a7567784e

A perf-event uprobe unregisters one probe at a time under the global `event_mutex`, so the waits add up: about 0.25s per probe on 6.12, 0.04s on 6.8.

### Changes

- Use uprobe_multi links (Linux 6.6+) when available
- Concurrent release on shutdown
- Fix: `Tracer.Run` closed a snapshot of the closers taken when it started, so probes of executables instrumented later were never closed by OBI and the kernel released them one by one after `os.Exit`. `ProcessTracer` now owns the per-executable links and releases them all at once
on shutdown and on executable unlink.

## Tests/timings

AL2023 6.12.94 aarch64, `exe_path: "*"`, context propagation `all`, network metrics on. N Go HTTP servers, each a distinct binary.

| build | N=5 (7 Go binaries) | N=20 (22 Go binaries) |
|---|---|---|
| v0.13.0 | 223–250 s (918 perf uprobes) | 747 s (2658 perf uprobes) |
| uprobe_multi only | 77 s | 163 s |
| uprobe_multi + concurrent release | **19 s** | **56 s** |

## Follow-up

Nine programs still attach as perf events because they tail-call into `jump_table`, and the kernel rejects tail calls between programs with different attach types:

- gotracer: `obi_uprobe_netFdRead`, `obi_uprobe_netFdReadRet`, `obi_uprobe_netFdWrite`, `obi_uprobe_cryptoTlsRead`, `obi_uprobe_cryptoTlsWrite`
- generictracer: `obi_uretprobe_ssl_read`, `obi_uretprobe_ssl_read_ex`, `obi_uretprobe_ssl_write`, `obi_uretprobe_ssl_write_ex`

Each of them costs about 0.23 s to release, and the Go ones are attached once per Go binary (the `netFdReadRet` one at every return site). They account for all the remaining teardown time in the table above: 82 perf events at N=5, 247 at N=20.

Moving them to `uprobe_multi` needs a second jump table whose entries are multi-typed copies of the tail-call targets. That doubles the load and verifier cost of the protocol handlers, so it is left for a separate PR.


pkg/internal/ebpf/uprobe/uprobe.go
```diff
@@ -0,0 +1,162 @@
...
+	"sync"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/features"
+	"github.com/cilium/ebpf/link"
+	"golang.org/x/sys/unix"
+)
+
+var uprobeMultiSupported = sync.OnceValue(func() bool {
+	if err := features.HaveBPFLinkUprobeMulti(); err != nil {
```
MrAlias, line 26: `HaveBPFLinkUprobeMulti` only proves that the link type exists. Linux's initial PID-filtered implementation compares the current thread, so on 6.6.0-6.6.34 a PID-scoped link fires only on the selected thread; the fix was backported in 6.6.35. `AttachUprobe` now uses this gate for the Python runtime and Go Auto SDK activation probes with nonzero PIDs, so hits on other threads are silently lost on those supported kernels.

Could we fold the [upstream negative-PID check](https://github.com/torvalds/linux/commit/04d939a2ab229a3821f04fc81f7c027842f501f1) into this feature test and keep singular perf-event uprobes when filtering is broken? That decision has to happen before `PrepareUprobeSpecs` marks eligible specs as `AttachTraceUprobeMulti` and `loadSpec` loads them.

pkg/ebpf/tracer_linux.go
```diff
@@ -576,19 +582,52 @@ func (pt *ProcessTracer) removeInstrumenter(key ExecutableKey, i *instrumenter)
...
 	for ino := range i.modules {
 		for _, p := range pt.Programs {
 			p.UnlinkInstrumentedLib(ino)
 		}
 	}
 }
 
+// probes still open at exit are released one by one by the kernel
+func (pt *ProcessTracer) closeInstrumenters() {
+	pt.instrumentablesMu.Lock()
```
MrAlias, line 616: `closeInstrumenters` only serializes access to the committed maps. Although `ProcessTracer.NewExecutable` uses `instrumentablesMu` during lookup and commit, it does not hold the mutex continuously across attachment. If SIGTERM arrives during that work, this function can clear the current set before the in-flight call commits its newly attached instrumenter. Those links are then left for the kernel's slow process-exit cleanup, which recreates the teardown failure this PR is meant to remove.

Could we close the lifecycle under this mutex, or track in-flight attachments, so shutdown rejects new commits and waits for an attachment that already started? A regression test that pauses `NewExecutable` before commit while `closeInstrumenters` runs would cover the race.

pkg/ebpf/instrumenter.go
```diff
@@ -144,7 +144,6 @@ func (i *instrumenter) goprobes(p Tracer) error {
 			"process", i.processName, "wanted_symbols", len(attachedSymbols))
 	}
 	i.closables = append(i.closables, closers...)
```
MrAlias, line 147: This append is now the sole owner of these baseline Go-probe links because this PR removes the corresponding `p.AddCloser` call. However, `NewExecutable`'s uncommitted defer only rolls back optional Go-probe groups. If `goprobes` succeeds and a later stage fails—for example, the process exits before `uprobes` can resolve `/proc/<pid>/exe`—the instrumenter is never committed and none of these baseline links are explicitly closed. Their cleanup can therefore be deferred to GC or process exit, reintroducing the slow shutdown this ownership change is intended to prevent.

Could the uncommitted path close all resources in `i.closables`, with a regression test where a later attachment stage fails after the Go probes succeed?

bpf/gotracer/go_net.c
```diff
@@ -91,7 +91,7 @@ int GUARDED_PROG(obi_uprobe_netFdRead, struct pt_regs *, ctx) {
 }
 
 // k_tail_continue_netfd_read
-SEC("uprobe/netFdRead_cont")
+SEC("kprobe")
```
MrAlias, line 97: `obi_continue_netfd_read` remains a userspace-probe continuation and is reached only through the runtime-populated jump table. The `kprobe` section prevents `PrepareUprobeSpecs` from assigning `AttachTraceUprobeMulti` to a tail-call target that `progArrayContents` cannot discover, but that invariant is not apparent here. Could we document why this intentionally uses `SEC("kprobe")`? A future cleanup could otherwise restore the seemingly more accurate `uprobe/...` section and recreate the attach-type mismatch.

pkg/ebpf/common/uprobe_attach.go
```diff
@@ -0,0 +1,162 @@
...
+var uprobeMultiSupported = sync.OnceValue(func() bool {
+	if err := features.HaveBPFLinkUprobeMulti(); err != nil {
+		slog.Info("attaching uprobes as perf events, the kernel has no uprobe_multi links", "reason", err)
+		return false
+	}
+	slog.Info("attaching uprobes as uprobe_multi links")
+	return true
+})
+
+func PrepareUprobeSpecs(spec *ebpf.CollectionSpec) {
```
MrAlias: This is called only by `ProcessTracer.loadAndAssign` and `RunUtilityTracer`, which already prepare every tracer's `SpecBundle`; external wrappers using those loaders do not need to invoke it directly. Exporting it from `pkg/ebpf/common` turns an OBI-specific in-place mutation and its section-name conventions into downstream API. Could we move the spec-preparation and attachment implementation into an internal package such as `pkg/internal/ebpf/uprobe` instead?

pkg/ebpf/common/uprobe_attach.go
```diff
@@ -0,0 +1,162 @@
...
+	prepareUprobeSpecs(spec, uprobeMultiSupported())
+}
+
+func prepareUprobeSpecs(spec *ebpf.CollectionSpec, multi bool) {
```
MrAlias: `multi` makes this helper represent two operations: no-op and rewrite. Its sole production caller already owns capability detection and simply forwards that result. Could we let that caller return when support is unavailable and invoke a separate pure `markUprobeMultiSpecs(spec)` transformer when it is available? The tests can then exercise the transformation directly without a boolean control parameter.

MrAlias (commented):
The uprobe teardown path is much better, but kernel compatibility, attachment ownership, and the new spec-preparation boundary still need tightening.

pkg/internal/ebpf/uprobe/uprobe.go
```diff
@@ -0,0 +1,212 @@
...
+
+type perfEventLinks []link.Link
+
+func (l perfEventLinks) Close() error {
+	var err error
+	for _, lk := range l {
+		err = errors.Join(err, lk.Close())
+	}
+	return err
+}
```
MrAlias, line 216: `attachPerfEvents` collapses every return offset into one `perfEventLinks`, while both `unlinkInstrumenter` and the prior ring-buffer cleanup only parallelize top-level closers. On fallback paths—including tail-call probes intentionally kept as perf events—this loop serializes all unregister waits for one probe within each executable. That makes the new wrapper give up the per-offset concurrency the old flat closer list had.

Could we close these members concurrently and cover the fallback with a test using blocking closers?

```suggestion
func (l perfEventLinks) Close() error {
	errs := make([]error, len(l))
	var wg sync.WaitGroup
	for i, lk := range l {
		wg.Go(func() { errs[i] = lk.Close() })
	}
	wg.Wait()
	return errors.Join(errs...)
}
```

MrAlias (approved):
Overall, looks good. Only feedback is on the perf-event fallback concurrency optimization.

grcevski (approved):
LGTM! Amazing!

pkg/internal/ebpf/uprobe/uprobe.go
mmat11: done, for consistency -- the kernel holds a global lock while destroying uprobes so it brings no real benefit

PR #3422, opened 2026-09-14
Fix Rails resource route harvesting

MrAlias:
## Summary

- honor literal `config.api_only = true` application settings when expanding Rails resource routes
- expand every static operand in multi-resource declarations for both `resources` and `resource`
- cover API-only matching and multi-resource expansion with regression tests

This follows up on #3374. API-only applications do not generate `new` or `edit` routes by default, and Rails accepts multiple resource names in a single declaration. Keeping the harvester aligned avoids incorrect route templates and missing resources.

## Testing

- `go test -race ./pkg/internal/transform/route/harvest`
- `make lint`


pkg/internal/transform/route/harvest/rails.go
```diff
@@ -293,6 +366,23 @@ func railsLiteralValue(value string) (string, bool) {
 	return match[1] + match[2] + match[3], true
 }
 
+func railsLiteralValues(value string) ([]string, bool) {
```
grcevski, line 373: I think this fails on the older Rails syntax, like this one:

```
resources :posts, :only => :show
```

which is the legacy equivalent of:

```
resources :posts, only: :show
```

I think because of this we may generate two routes here {:posts, :only}, but only is really not there.

Trying a sketch:

```
resources :posts, :comments, :only => :show
          └──── resources ────┘  └── option ──┘
```

this should return []string{"posts", "comments"}

pkg/internal/transform/route/harvest/rails.go
```diff
@@ -126,21 +150,57 @@ func scanRailsRouteFiles(ctx context.Context, root, mainPath string) ([]string,
 	return routes, nil
 }
 
-func readRailsRouteFile(ctx context.Context, path string) ([]string, []string, error) {
-	file, _ := langtools.OpenMetadataFile(path, maxRailsRoutesBytes)
+func readRailsAPIOnly(ctx context.Context, path string) (bool, error) {
```
grcevski, line 153: I think we should do last one wins here, I had similar issue for PHP naming and .env files:

```
config.api_only = true
config.api_only = false
```

it should be false. Just read it until EOF and keep the last known value.

MrAlias: `readRailsAPIOnly` now reads to EOF and keeps the last literal `true`/`false`. `config.api_only = true` followed by `config.api_only = false` is false.

Covered by the last-assignment cases in `TestReadRailsAPIOnly`. ce2bfe0c7

pkg/internal/transform/route/harvest/rails.go
MrAlias: `railsLiteralValues` now stops at a hash-rocket key, so `:only` is not harvested as a resource. `resources :posts, :comments, :only => :show` yields `posts` and `comments`, and `:only` / `:except` / `:path` / `:param` are accepted in that form.

Covered by the hash-rocket cases in `TestScanRailsRoutes`. 19b2d46af

mariomac (approved):
Looks ok!

grcevski (approved):
LGTM!

PR #3423, opened 2026-09-14
fix(tpinjector): delete tracked socket cookies when the socket closes

NimrodAvni78:
## Summary

tpinjector collects its attachment handles in `closers` via `AddCloser`, but nothing ever ran that list: `Run` closed only the BPF objects. After shutdown was requested the sk_msg verdict stayed attached to `sock_dir`, the sockops program stayed on the root cgroup, and sockets kept being enrolled and rewritten.

`sock_dir` is pinned into bpffs and never unpinned, so the sk_msg attachment outlives the process and keeps running until a later agent replaces it.

`Run` now calls `detach`: stop the backfill, close the closers in reverse attach order, then close the BPF objects. The objects must go last, since `SockMsg.Close` needs `sock_dir`'s FD to detach.

## Validation


pkg/internal/ebpf/tpinjector/tpinjector.go
```diff
@@ -249,9 +252,43 @@ func (p *Tracer) RegisterOffsets(_ *exec.FileInfo, _ *goexec.Offsets) {}
...
+		return
+	}
+
 	p.closers = append(p.closers, c...)
 }
 
+func (p *Tracer) closeAllReverse(closers []io.Closer) {
+	for _, c := range slices.Backward(closers) {
+		if err := c.Close(); err != nil {
+			p.log.Debug("error detaching tpinjector resource", "error", err)
```
mmat11: can this happen normally? is `debug` the right level?

mmat11 (approved):
lgtm

pkg/internal/ebpf/tpinjector/tpinjector.go
NimrodAvni78: yep you are right, moved to warn

grcevski (approved):
LGTM!

PR #3425, opened 2026-09-14
Fix missing sockets on PID add

### LINKED ISSUE #3367: iter/tcp backfill doesn't populate filter_ports, leaving pre-existing listeners invisible to the socket filter
### Component

OBI (eBPF instrumentation)

### What happened?

Found while digging through #3297. `accept4` writes a new listener's port to two maps: `listening_ports` (keyed by port + netns, used by `is_listening()` in `protocol_tcp.h` to classify a connection as server or client side) and `filter_ports` (port only, no netns, because as the map's own header notes, the network namespace is "not reliably accessible from `__sk_buff`" in socket filter programs).

`iter/tcp`, the backfill iterator that walks existing listening sockets at startup or on process discovery, only writes `listening_ports` ([iter_tcp.c:26](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/bpf/generictracer/iter_tcp.c#L26)). It never touches `filter_ports`.

So a listener backfilled by the iterator is correctly classified for connection role purposes, but the socket filter's `filter_ports` lookup won't know about it. Note that `filter_ports` isn't the last resort there, `obi_socket_flt_buf` checks `connection_tracker` first and falls through to `sock_filter_buffers` afterwards (k_tracer.c:1308-1316), so a backfilled listener is only fully missed when all three checks miss.

Both maps stay consistent regardless of whether a listener was discovered via `accept4` or backfilled via `iter/tcp`.

### Steps to reproduce

This is a gap found by reading the code, not a runtime repro with observed symptoms. To verify it directly:

1. Look at `obi_kretprobe_sys_accept4` in bpf/generictracer/k_tracer.c (~line 100-129). Note it writes both `listening_ports` and `filter_ports`.
2. Look at `obi_iter_tcp` in bpf/generictracer/iter_tcp.c. Note it only writes `listening_ports` (line 26), never `filter_ports`.
3. Look at `obi_socket_flt_buf` in bpf/generictracer/k_tracer.c (~line 1300-1316), where the socket filter falls back to `filter_ports` because it cannot read netns there.

### Additional context

Not confirmed to cause a user-visible symptom on its own, this was spotted by inspection while working the pre-existing-connection issue in #3297, not reproduced independently. Filing separately since it's a different code path (early buffer capture via the socket filter) from that issue's fix.
 `filter_ports` is keyed on port alone with no netns, so it's global across the node. `accept4` only adds ports OBI actually saw an accept on, which keeps the map roughly in line with real traffic. If the iterator wrote to it as well, it would add every listening port in every netns it walks. On a node running 40 pods that each listen on 8080, that's one global entry telling the socket filter port 8080 is interesting everywhere, including namespaces where nothing relevant is happening.

The delete side has the same shape. `inet_csk_listen_stop` removes from `filter_ports` by port only (k_tracer.c:1538), so one pod shutting down its listener clears the entry for every other namespace using that port. That's true today, but filling the map from the iterator would make it happen far more often. `filter_ports` is also an LRU sized `MAX_CONCURRENT_REQUESTS`, so a large backfill risks evicting entries `accept4` put there for real connections.

So the actual question may not be "should iter/tcp write to filter_ports too", but whether `filter_ports` should carry netns at all.

grcevski:
## Summary

We were only scanning the network namespace for sockets when we first launch the generic tracer. However, there is a class of applications, namely PHP-FPM which launch workers with already established sockets. Also there's this issue https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/3367.

The problem with PHP-FPM was that we get the first receive event on the PHP server side and since there was no accept, we consider it as from a prior transaction and avoid to process it, because  it could have flipped ports. This was introduced in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/2663/changes. 

The fix here is to scan the PID namespace when we are adding the pids, but only that namespace. As many PIDs can share the same network namespace, I cache the lookup temporarily to avoid processing many times.

Most of the PR are tests.

Closes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/3367

## Validation


mmat11 (approved):
lgtm! does the test exercise the PID backfill path?

grcevski:
> lgtm! does the test exercise the PID backfill path?

yes, the tests would not pass without this working, since all these 3 examples I added never do an accept after we instrument them, each worker seems to keep a live socket and they just talk through it.

PR #3427, opened 2026-09-14
Select compatible versioned Go probe groups

MrAlias:
## Summary

- allow optional Go probe groups to require any one of several alternative baseline symbols
- skip groups that conflict with a successfully attached earlier layout
- record successful optional-group symbols for subsequent selection decisions
- preserve selection rules while resolving probes across Go symbol copies

## Why

Go HTTP/2 and grpc-go layouts vary across supported versions. The legacy traceparent ownership probes need to select a complete compatible layout without mixing modern and legacy probes or partially attaching an unsupported group.

This adds only the reusable loader behavior. The protocol-specific legacy probes remain separate follow-up changes.

## Validation

- `make verify`
- `make compile`
- `go test ./pkg/ebpf ./pkg/ebpf/common`

Relates to #2769.


grcevski:
I think this needs to go by CopyID, if I understand this correctly, as it's done at the moment it will disable the legacy probes for binaries that have both, but I don't think we want that, do we?

mariomac (approved):
LGTM. I understand that following PRs will populate the new fields, right?

pkg/internal/ebpf/gotracer/gotracer.go
```diff
@@ -2186,8 +2186,8 @@ func (p *Tracer) GoProbeGroups() []ebpfcommon.GoProbeGroup {
 
 	if p.goAutoSDKActivationProbesEnabled() {
 		groups = append(groups, ebpfcommon.GoProbeGroup{
-			Name:          "go_auto_sdk_activation",
-			Prerequisites: append([]string(nil), goAutoSDKActivationPrerequisiteSymbols...),
+			Name:        "go_auto_sdk_activation",
+			RequiresAll: append([]string(nil), goAutoSDKActivationPrerequisiteSymbols...),
```
mmat11, line 2190: why is there a []string(nil) here?

mmat11 (approved):

pkg/internal/ebpf/gotracer/gotracer.go
MrAlias: This is a [Go idiom for a slice with a new backing array](https://go.dev/wiki/SliceTricks#copy).

mmat11: funky :D

MrAlias:
> I understand that following PRs will populate the new fields, right?

Correct 👍

MrAlias:
Conflict tracking now uses the resolved Go symbol copy. Successful optional groups record the `(CopyID, symbol)` pair, so a modern layout suppresses a legacy layout only in the same canonical or vendored copy. A regression test covers a modern group in one vendored copy and legacy groups in two copies.

PR #3428, opened 2026-09-14
Preserve application traceparents in grpc-go

MrAlias:
This implements the current grpc-go slice of #2769 without relying on a timing window.

The direct-write decision is scoped to the active loopyWriter call. Socket fallback ownership uses an exact socket-cookie, process, and HTTP/2 stream identity, with socket cookies populated by sockops for new connections and the TCP iterator for pre-existing connections. The current grpc-go header handler publishes stream state before serialization, pointer correlations include process identity, and request completion only drops reverse bookkeeping; queued or active writer state is consumed by the writer paths that own it.

- observe application-supplied traceparent fields by scanning the client header slice once before HPACK serialization, preserving them based on field ownership without validating or rewriting the application value
- bound the scan at 256 fields and conservatively treat longer or unreadable slices as application-owned so uncertain application data is never overwritten
- keep writer-local ownership independent of connection extraction so custom net.Conn wrappers and connection-cache misses cannot bypass the direct-write guard; socket identity is required only for socket fallback
- resolve grpc-go socket identity through netFD.pfd.Sysfd across supported Go 1.17-1.27 layouts
- drop `-buildvcs=false` from the offsets tracker's `GOFLAGS`, since the toolchains it downloads to regenerate these offsets include Go 1.17, which predates that flag
- publish outgoing stream state from the active header handler before serialization begins
- suppress socket fallback at the direct-write ownership gate even when socket-cookie storage is unavailable
- cover repeated HPACK-indexed values, malformed application values, traceparent fields after 40 metadata entries, conservative overflow handling, opaque wrapped connections, and multiplexed owned and unowned streams over plaintext and TLS
- verify process-scoped map keys, socket-cookie storage, and iterator programs

Legacy grpc-go layouts and versioned layout selection remain follow-up work. `clientHeaderHandler` first appears in grpc-go 1.83, so on 1.82 and earlier — which still use `headerFrame` and `headerHandler` — the ownership group does not attach and those binaries keep today's behavior of adding a second traceparent to application-owned streams. This PR does not regress them, it does not fix them yet.

Part of #2769

### Performance decision

The ownership check originally attached to `hpack.(*Encoder).WriteField`, which crosses the uprobe boundary once for every encoded field. A grpc-go v1.83.1 microbenchmark using a reused encoder measured median detached costs of 86.3 ns for a traceparent field and 92.3 ns for a content-type field. Attaching the uprobe raised those medians to 3,763.8 ns and 3,712.4 ns, adding 3.68 µs and 3.62 µs per field respectively across 10 trials.

End-to-end measurements used one reused plaintext gRPC connection, 1,000 warmup requests, and seven measured trials. Serial cases ran 5,000 requests per trial; concurrent cases ran 20,000 requests per trial at concurrency 16.

| Scenario | WriteField uprobe | One handler scan | Latency change |
| --- | ---: | ---: | ---: |
| Serial, no extra metadata | 503.0 µs/request | 459.8 µs/request | -8.6% |
| Serial, application traceparent | 505.9 µs/request | 458.7 µs/request | -9.3% |
| Serial, 16 metadata pairs and application traceparent | 574.6 µs/request | 468.2 µs/request | -18.5% |
| Concurrency 16, no extra metadata | 153.0 µs/request | 111.7 µs/request | -27.0% |
| Concurrency 16, 16 metadata pairs and application traceparent | 218.5 µs/request | 114.3 µs/request | -47.7% |

Across the measured scenarios, the one-scan implementation was within -0.3% to +3.1% of the control with the ownership uprobe detached. This is why the implementation scans once in the already-attached handler probe instead of retaining the per-field uprobe.


grcevski (approved):
LGTM! Wow this is really good, it's amazing you found a way to get the cookie information from the go side.

grcevski:
I had AI generate me a flow map of all the changes in case you want to add it to the docs of how this now works:
```mermaid
flowchart TD
    A["ClientConn Invoke/NewStream<br/>request goroutine"] --> B["http2Client.NewStream<br/>cache connection + socket cookie"]
    B --> C["controlBuffer.executeAndPut<br/>request → queued header pointer"]
    C --> D["clientHeaderHandler<br/>loopyWriter goroutine"]
    D --> E["Publish stream trace<br/>outgoing_trace_map"]
    D --> F["Scan up to 256 HPACK fields"]

    F -->|traceparent found or uncertain| G["Writer-local ownership<br/>writer PID/goroutine + stream"]
    F -->|traceparent found + cookie available| H["Socket ownership<br/>cookie + PID + stream"]

    E --> I["grpcFramerWriteHeaders"]
    G --> I
    I -->|application owned| J["Skip direct injection<br/>mark written"]
    I -->|OBI owned| K["Append traceparent<br/>to grpc-go buffer"]

    H --> L["sk_msg detect_h2"]
    J --> L
    K --> L
    L -->|exact ownership marker| M["Preserve application header"]
    L -->|written=1| N["Direct write already handled"]
    L -->|written=0| O["Scan/adopt or inject on socket"]

    P["sockops: new sockets"] --> Q["socket_cookie SK_STORAGE"]
    R["TCP iterator: pre-existing sockets"] --> Q
    Q --> H
```

MrAlias:
Nice! Yeah, let me look into adding that 😄

mmat11 (approved):

PR #3434, opened 2026-09-15
fix: read buildinfo.Version when building resources

### LINKED ISSUE #3433: telemetry.distro.version is 'unset' when a component vendors OBI and sets buildinfo.Version at runtime
## What happened

A component that vendors OBI and sets `buildinfo.Version` at runtime still emits `telemetry.distro.version=unset` on every resource.

`pkg/buildinfo` documents that contract:

```go
// Version and Revision variables are overridden at build time with Git repository information
// They can be also overridden at runtime by any software component vendoring OBI as library
var (
	Version  = "unset"
	Revision = "unset"
)
```

But `pkg/export/attributes/names/attrs.go` snapshots that value into a package-level variable at initialization time:

```go
var (
	VendorPrefix           = "obi"
	...
	TelemetryDistroVersion = buildinfo.Version
)
```

Package-level initialization runs before a vendoring component can assign `buildinfo.Version`, so the snapshot keeps the `"unset"` default. The five call sites that build resources read the stale snapshot rather than the value the vendoring component set.

The other overridable globals in the same block are not affected: a vendoring component assigns `VendorPrefix` / `VendorSDKName` directly, so no snapshot is involved.

## Reproduction

A minimal program that mirrors what a vendoring component does (assign `buildinfo.Version` at runtime, after OBI's package init):

```go
package main

import (
	"fmt"

	"go.opentelemetry.io/obi/pkg/buildinfo"
)

func main() {
	buildinfo.Version = "v3.35.0"
	fmt.Println("TelemetryDistroVersion =", attr.TelemetryDistroVersion)
}
```

Against OBI at `573b61a`:

```
TelemetryDistroVersion = unset
```

Observable end to end with Grafana Beyla 3.35.0, which vendors OBI and assigns `obibuildinfo.Version` in `OverrideOBIGlobalConfig`. Scraping its Prometheus endpoint:

```
target_info{...,telemetry_distro_name="opentelemetry-ebpf-instrumentation",
  telemetry_distro_version="unset",telemetry_sdk_name="beyla",...} 1
```

while its own build-info metric carries the version correctly, showing the ldflags stamping itself works:

```
beyla_build_info{...,version="v3.35.0"} 1
```

Standalone OBI is unaffected, because ldflags set `buildinfo.Version` before any Go initialization:

```
target_info{...,telemetry_distro_version="v0.13.0",telemetry_sdk_name="opentelemetry",...} 1
```

## Expected

`telemetry.distro.version` reports the version the vendoring component set, matching the runtime-override contract documented on `pkg/buildinfo`.

## Affected call sites

All five read the snapshot:

- `pkg/export/otel/otelcfg/common.go` — `resourceAttrs`
- `pkg/export/otel/metrics.go`
- `pkg/export/otel/metrics_internal.go`
- `pkg/export/otel/metrics_svc_graph.go`
- `pkg/export/prom/prom.go` — `labelValuesForNodeMeta`

## Environment

- OBI `573b61a` (two commits after `v0.13.0`); also reproduced with the `otel/ebpf-instrument:v0.13.0` image for the standalone comparison
- `grafana/beyla:3.35.0` for the downstream case
- Linux kernel 7.0.0, BTF enabled, amd64

ymotongpoo:
Fixes #3433.

## Problem

`TelemetryDistroVersion` was a package-level variable initialized from `buildinfo.Version`:

```go
var (
	VendorPrefix           = "obi"
	...
	TelemetryDistroVersion = buildinfo.Version
)
```

`pkg/buildinfo` documents that `Version` "can be also overridden at runtime by any software component vendoring OBI as library". That assignment necessarily happens after OBI's package initialization, so the snapshot above keeps the `"unset"` default and every emitted resource carries `telemetry.distro.version=unset`.

Standalone OBI is unaffected because ldflags set the variable before any Go initialization runs. The neighbouring globals in the same block are unaffected too: a vendoring component assigns `VendorPrefix` / `VendorSDKName` directly, with no snapshot in between.

## Change

`TelemetryDistroVersion` becomes a function, so the five resource-building call sites read the current value:

- `pkg/export/otel/otelcfg/common.go` — `resourceAttrs`
- `pkg/export/otel/metrics.go`
- `pkg/export/otel/metrics_internal.go`
- `pkg/export/otel/metrics_svc_graph.go`
- `pkg/export/prom/prom.go` — `labelValuesForNodeMeta`

This changes the type of an exported identifier. `VERSIONING.md` places OBI in Development with breaking changes allowed in minor releases, and the identifier is only useful to a vendoring component — which is exactly the case that is broken today. If you would rather keep the variable and add a separate accessor, say so and I will rework it.

## Tests

`TestResourceAttrsHonourRuntimeBuildinfoVersion` in `pkg/export/otel/otelcfg/common_test.go` sets `buildinfo.Version` after initialization and asserts the attribute on the built resource.

Verified it bites: reintroducing the snapshot (`var telemetryDistroVersionSnapshot = buildinfo.Version` returned from the function) makes it fail with `expected: "v9.9.9-vendored", actual: "unset"`; with the fix it passes.

Also confirmed end to end with a small program that mirrors a vendoring component, resolved against each tree via `replace`:

```go
buildinfo.Version = "v3.35.0"
fmt.Println(attr.TelemetryDistroVersion())
```

- against `573b61a` (unfixed): `unset`
- against this branch: `v3.35.0`

## Validation

- `make lint` — clean
- `go build ./...` — clean
- `go test ./pkg/export/otel/ ./pkg/export/prom/ ./pkg/export/attributes/... ./pkg/export/otel/otelcfg/` — all pass (eBPF bindings generated with `make docker-generate` first)

## Not included

No schema transformation entry in `devdocs/telemetry-schema.md`: the attribute is neither renamed nor removed, only its value is corrected.


mariomac (approved):
Great catch!

PR #3438, opened 2026-09-15
uprobes: give uprobe_multi programs their own tail-call table

mmat11:
## Summary

Follow-up to #3418.

This PR adds a second table, `jump_table_um`, declared next to the first. At load
time the multi programs are pointed at it and its slots are filled with
multi-typed copies of the same handlers, so both worlds tail-call into a table
that matches their attach type. One X-macro list generates the slot names, the
forward declarations and both tables, so a slot cannot exist without its program
and the two tables cannot diverge; a missing program fails `make generate`.

Two shutdown bugs surfaced while measuring and are fixed here: the pending
Go-HTTP-client LRU deadlocked on shutdown and the process attacher blocked
in a `Queue.Send` with no context, costing the full grace period.

## Tests

AL2023 6.12.94 aarch64, `exe_path: "*"`, context propagation on, N Go services, each a distinct binary.

| build | N=5 | N=20 |
|---|---|---|
| v0.13.0 | 250 s | 747 s |
| #3418 | 20 s | 53 s |
| this PR | **3.7 s** | **3.7 s** |

Perf-event uprobes go from 916 to 0; the 405 attachments that remain are
uprobe_multi links. Exit time no longer grows with the number of instrumented
services.

Cost: 33 extra loaded programs (17 in gotracer, 16 in generictracer), which is
kernel memory and verification time, not runtime.

## Validation


pkg/internal/ebpf/uprobe/uprobe.go
```diff
@@ -108,24 +117,86 @@ func progArrayContents(spec *ebpf.CollectionSpec) map[string]bool {
...
+}
+
+// uprobeMultiTwin fills the twin prog array with uprobe_multi copies of the programs
+func uprobeMultiTwin(spec *ebpf.CollectionSpec, table string, twins map[string]bool) string {
+	twinName := table + uprobeMultiSuffix
+	if twins[twinName] {
+		return twinName
+	}
+	twins[twinName] = true
+	twin := spec.Maps[twinName]
```
NimrodAvni78, line 176: we need a nil check here no? if no twin exists this will panic

bpf/generictracer/k_tracer_tailcall.h
```diff
@@ -6,36 +6,82 @@
...
+// uprobe_multi copies of the same programs
 struct {
     __uint(type, BPF_MAP_TYPE_PROG_ARRAY);
-    __type(key, u32);
-    __type(value, u32);
     __uint(max_entries, k_tail_count);
-} jump_table SEC(".maps");
+    __uint(key_size, sizeof(u32));
+    __array(values, int(void *));
+} jump_table_um SEC(".maps") = {
```
NimrodAvni78, line 85: maybe we should condionally not create this map if uprobe multi is unsupported?

mmat11: on kernels without uprobe_multi the twin's static contents are the same programs the original table holds, so nothing extra is loaded; the cost is one prog array of already-existing fds (a few bytes). Conditionally dropping the map would also break LoadAndAssign, which requires every declared prog array to be assigned

pkg/internal/ebpf/uprobe/uprobe.go
mmat11: done

MrAlias (approved):

PR #3445, opened 2026-09-16
fix(mcp): declare mcp operation duration metrics in semconv registry

NameHaibinZhang:
## Summary

OBI emits `mcp.client.operation.duration` and `mcp.server.operation.duration` histograms (added in #3351), but the OBI semantic-convention registry only declared MCP spans, not metrics. This caused `TestSuite_PythonMCP` to fail with 14 violation-level advisories:

```
Metric does not exist in the registry. (signals: metric:mcp.client.operation.duration, metric:mcp.server.operation.duration)
```

Declare both metrics as OBI overrides of the upstream v1.41.0 definitions, with the attribute set OBI actually emits on them. Regenerate schema reference docs so they stay in sync with the registry.

## Changes

- Add `schemas/obi/groups/mcp/metrics.yaml` with the two operation-duration metrics.
- Regenerate `site/docs/README.md`, `site/docs/metrics.md`, `site/docs/spans.md`.

## Test plan

- [x] `go test ./internal/schemacheck/...` passes
- [x] `go test ./internal/test/weavercheck/...` passes
- [x] `make check-schema-files` passes
- [x] `make generate-schema-docs` succeeds


NimrodAvni78 (approved):
Thanks!

PR #3446, opened 2026-09-16
fix(nodejs): dont close inspector when OBI didnt open it

NimrodAvni78:
## Summary

OBI closes the Node.js debugger port even when it was not the one that opened it.

`injectFileWS` ends every injection with `process._debugEnd()`, which stops the inspector. That is correct for the common path, where OBI opened the port with `SIGUSR1` and must close it again. But the same code runs on the already-listening path — a process started with `--inspect`, where the operator asked for that port. OBI connects through it, injects, and closes it on the way out. Any attached debugger is dropped, and nothing reopens the port short of restarting the process.

Nothing recorded which of the two paths opened the inspector, so the close could not tell them apart. It is now a parameter: the `SIGUSR1` path passes true, the already-open path passes false.

## Validation


mmat11 (approved):

mariomac (approved):
Good catch!

Issue #3447, opened 2026-09-16
Ruby/Puma request correlation silently stops working on Ruby 4.0


NimrodAvni78:
### Component

OBI (eBPF instrumentation)

### What happened?

The Ruby/Puma correlation probes stop correlating entirely on Ruby 4.0. Both uprobes registered under the `libruby` key still attach and still fire, but every worker-side lookup misses, so no Puma-handled request gets its server context attached to the work the worker thread then does.

Ruby 4.0 added a YARV instruction, `opt_new`, that inlines `Class#new` at the call site: the object is allocated by `rb_obj_alloc()` and `initialize` is dispatched as an ordinary send. `rb_class_new_instance_pass_kw()`, the only caller of `rb_obj_call_init_kw()` on that path, is never entered. Puma's acceptor builds its client with exactly that call form (`pool << Client.new(io, ...)` in `lib/puma/server.rb`), so `obi_rb_obj_call_init_kw` never records the `Puma::Client` object and the pointer `obi_rb_ary_shift` later looks up can never match.

Controlled sweep, same application (Rails 8.1 + MySQL behind nginx), same OBI build, same load; only the runtime changed. Counts are from OBI's own BPF debug log over a 20 second window:

| Ruby | Puma | acceptor records | worker shifts | correlation hits | misses |
|---|---|---|---|---|---|
| 3.3.12 | 6.6.1 | 24 | 1801 | 12 | 1789 |
| 3.4.10 | 6.6.1 | 24 | 1803 | 14 | 1789 |
| 4.0.1 | 6.6.1 | 12 | 2489 | 0 | 2489 |
| 4.0.1 | 8.0.2 | 12 | 2487 | 0 | 2487 |

The failure is silent. There is no error, no warning, and no counter. The only visible sign is that the `stored item to id correlation` debug line never appears while `untracked item ... ignoring` appears on every shift.

### What did you expect to happen?

Either the correlation keeps working on Ruby 4.0, or OBI does not attach a probe pair that cannot succeed there and says so. A registered probe becoming a permanent no-op should not be invisible.

### Steps to reproduce

Minimal, no Rails. This isolates the runtime change from OBI entirely.

`/tmp/rubyrepro/alloc.rb`:

```ruby
class Widget
  def initialize(a, b = nil)
    @a = a
    @b = b
  end
end

sleep 10                                 # time to attach the uprobe
200_000.times { Widget.new(1, 2) }
[...]

PR #3450, opened 2026-09-16
fix(ebpf): only attach the Ruby correlation probes where they can work

NimrodAvni78:
## Summary

The two libruby probes exist solely to correlate a Puma request to the worker thread that serves it, but they sit on symbols the Ruby runtime exercises as a whole: `rb_obj_call_init_kw` on object initialization and `rb_ary_shift` on every `Array#shift`. Today they attach to any instrumented process that maps libruby, whether or not the correlation can ever succeed there. Two gates fix that.

**Ruby < 4.0.**: The correlation for ruby > 4.0  currently does not work (detailed in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/3447)

**Puma present.** The correlation only ever fires on threads named `puma srv` and `puma srv tp NNN`. Any other Ruby process - Sidekiq, a rake task, a script, a different server - pays for both probes and gets nothing back. `uprobeLibraryPrerequisites` now requires `puma_http11` to be mapped before libruby is instrumented. Puma loads its C extension while booting, before it serves traffic, so the mapping is present by the time discovery reaches the process.

## Why this is worth gating

Both symbols are hot in the Ruby runtime generally, and a uprobe on them costs roughly a microsecond per call. Measured on Ruby 3.4.10, aarch64, kernel 6.11, 3M warm iterations, attaching nothing but a counting uprobe (so this is a floor for OBI's own program, which does more work):

| probed symbol | without uprobe | with uprobe | per call |
|---|---|---|---|
| `rb_obj_call_init_kw` | 144.9 ns/op | 1291.2 ns/op | +1146 ns (8.9x) |
| `rb_ary_shift` | 63.0 ns/op | 1119.9 ns/op | +1057 ns (17.8x) |

That cost lands on the whole process, not on request handling: object initialization and `Array#shift` run during boot, background jobs and ActiveRecord. In the integration workload `rb_ary_shift` fired about 2500 times for six HTTP requests, so the overwhelming majority of those traps are on arrays that have nothing to do with Puma's queue.

## Validation


pkg/ebpf/instrumenter.go
```diff
@@ -491,6 +497,29 @@ func dedupModuleProbes(
 
 // matchVersionedUprobeLibrary reports whether a (possibly annotated) library name should be
```
marctc, line 498: this comment should go above `matchVersionedUprobeLibrary` no?

pkg/ebpf/instrumenter.go
```diff
@@ -419,6 +419,12 @@ func (i *instrumenter) uprobeModules(p Tracer, pid app.PID, maps []*procfs.ProcM
 		}
 
 		lib = baseLib
+		if missing, ok := missingUprobeLibraryPrerequisite(lib, maps); !ok {
+			log.Debug("skipping uprobe library whose prerequisite is not loaded",
+				"lib", lib, "prerequisite", missing)
```
mmat11, line 424: prerequisite might be empty here, I wonder if it makes more sense to return (string, error) from `missingUprobeLibraryPrerequisite`?

mmat11 (approved):

mariomac (approved):

PR #3457, opened 2026-09-16
Extend Django route harvesting

marctc:
## Summary

  - Supports named lists passed to `include()`, preserving their mount prefixes.
  - Supports tuple includes containing named lists or module strings.
  - Extracts simple `re_path()` declarations with literal paths and whole-segment named captures, including + and fixed-count character classes.
  - Converts regex captures into route parameters without preserving their matching constraints.

## Validation


mmat11 (approved):

pkg/internal/transform/route/harvest/python.go
```diff
@@ -158,7 +158,13 @@ func (e *pythonExtractor) scanFile(path string) error {
 			hasDjangoPath = true
 		}
 		if hasDjangoPath {
-			djangoRoutes = append(djangoRoutes, scanDjango(text, djangoAliases)...)
+			routes := scanDjango(text, djangoAliases)
+			if match := djangoListAssignmentPattern.FindStringSubmatch(text); match != nil {
```
MrAlias: Tagging each route only with the list that syntactically contains it loses valid `urlpatterns` aliases and concatenations. For example, `extra_patterns = [path("reports/", reports)]` followed by `urlpatterns = extra_patterns` (or `urlpatterns += extra_patterns`) leaves the declaration under `extra_patterns`; the later assignment records no relationship, and `visit(..., "urlpatterns", ...)` filters the route out. This harvested result is marked complete, so requests to a real Django endpoint will no longer receive their route. Could we track list-to-list assignments and concatenations and cover these cases?

pkg/internal/transform/route/harvest/python_django.go
```diff
@@ -59,11 +64,83 @@ var djangoI18nUnprefixedDefault = regexp.MustCompile(
...
 )
 
+// djangoRegexRoute parses a supported re_path expression into a route template.
+// For example, ^articles/(?P<year>[0-9]{4})/$ becomes articles/<year>/.
+func djangoRegexRoute(pattern string) (string, bool) {
+	expr, err := syntax.Parse(pattern, syntax.Perl)
+	if err != nil {
+		return "", false
+	}
+	return djangoRegexExprRoute(expr)
```
MrAlias, line 127: `djangoRegexRoute` accepts patterns with no end anchor, but the complete matcher registers only the exact returned path. Django treats `re_path(r"^health/", view)` as a prefix match, so `/health/details` reaches the view while the harvested `/health/` route does not match it. Include patterns have the complementary constraint: because Django uses `search()` for them, a missing start anchor can consume an unrepresented prefix. Could we require full-match endpoint boundaries and anchored include prefixes, or otherwise preserve these matching semantics?

MrAlias (changes requested):
The named-list tracking can drop valid Django URL configurations, and the new regex conversion accepts boundary semantics that the exact route matcher cannot preserve.

pkg/internal/transform/route/harvest/python_django.go
```diff
@@ -50,7 +54,13 @@ var djangoImportAliasPattern = regexp.MustCompile(
 	`^import\s+([A-Za-z_]\w*(?:\.[A-Za-z_]\w*)*)\s+as\s+([A-Za-z_]\w*)\s*(?:#.*)?$`,
 )
 
-var djangoPathStart = regexp.MustCompile(`\bpath\s*\(`)
+var djangoPathStart = regexp.MustCompile(`\b(?:re_)?path\s*\(`)
+
+var djangoListAssignmentPattern = regexp.MustCompile(`^([A-Za-z_]\w*)\s*\+?=\s*\[`)
+
+var djangoListReferencePattern = regexp.MustCompile(
+	`^([A-Za-z_]\w*)\s*\+?=\s*([A-Za-z_]\w*(?:\s*\+\s*[A-Za-z_]\w*)*)\s*(?:#.*)?$`,
```
MrAlias: This pattern only records concatenations made entirely from bare names. With `urlpatterns = base_patterns + [path("ready/", ready)]`, `base_patterns` is assigned its own `listName`, this regex records no link to it, and the resolver emits only `/ready/`; the previous scanner emitted both routes. Could we retain the named-list reference when the expression also contains a literal list and add this case to the table test?

MrAlias (commented):
The boundary checks are sound, but the named-list fix is not yet safe for all valid compositions because it can make harvested routes disappear.

pkg/internal/transform/route/harvest/python.go
```diff
@@ -158,7 +158,24 @@ func (e *pythonExtractor) scanFile(path string) error {
...
 		if hasDjangoPath {
-			djangoRoutes = append(djangoRoutes, scanDjango(text, djangoAliases)...)
+			routes := scanDjango(text, djangoAliases)
+			if listName, references, ok := djangoListAssignment(text); ok {
+				for i := range routes {
+					routes[i].listName = listName
+				}
+				for _, name := range references {
+					for _, declaration := range djangoRoutes {
+						if declaration.listName == name {
```
MrAlias, line 168: List relationships are stateful, but this loop adds a timeless edge only when the source already has a declaration. With `base_patterns = []; urlpatterns = base_patterns; base_patterns += [path("health/", health)]`, Django serves `/health/` through the shared list, while no edge is recorded and the harvester omits it. The inverse also fails: after `base_patterns = [path("health/", health)]; urlpatterns = []; urlpatterns += base_patterns; base_patterns += [path("ready/", ready)]`, Django serves only `/health/`, but the stored edge later expands both source declarations and reports `/ready/` too.

Could we preserve statement order and distinguish direct list aliasing from `+`/`+=` copies, with both mutation-order cases in the tests?

MrAlias (commented):
The mixed-concatenation case is fixed. Looking at the resulting relationship model exposed one remaining issue: Python list assignments and mutations are order-sensitive, while these edges are not.

pkg/internal/transform/route/harvest/python.go
```diff
@@ -158,6 +159,11 @@ func (e *pythonExtractor) scanFile(path string) error {
 			hasDjangoPath = true
 		}
 		if hasDjangoPath {
+			if assignment, ok := djangoListAssignment(text); ok {
+				if applyDjangoAssignment(djangoLists, assignment, djangoAliases) {
```
MrAlias: The pointer-backed model now handles alias-versus-copy order correctly. The remaining fallback loses the destination list when the right-hand side is only partially understood.

For example:

```python
from app.urls import urlpatterns as imported_patterns

extra_patterns = imported_patterns + [path("health/", health)]
urlpatterns = [path("credit/", include(extra_patterns))]
```

Resolving the contents of `imported_patterns` is outside the needed scope here. Keeping the locally known route under `extra_patterns` is sufficient: the harvester should emit `/credit/health/` and must not emit `/health/`.

Could we represent `extra_patterns` as partially known instead of sending the recovered route through an unscoped `scanDjango`? A regression test with those positive and negative assertions should cover the remaining invariant while preserving the alias and copy behavior already tested.

MrAlias (commented):
The order-sensitive list model addresses the earlier aliasing cases. The remaining partial-assignment fallback loses the known destination list when one operand cannot be resolved.

MrAlias (approved):

PR #3458, opened 2026-09-16
fix(bpf): cache negative PID filter answers, keyed by host pid and clear the cache on filter rebuild

pinoOgni:
## Summary

This PR reduces the overhead OBI adds to processes it was never asked to watch: an app lost a third of its throughput because another process on the same node was instrumented. The main avoidable cause was the PID filter, and this PR removes it. 
The remaining cost is structural (every TCP round trip still enters ~20 global probes, and the per-packet socket filter is unchanged), see the numbers and follow-ups below.

Once any process is instrumented, ~19 kprobes (`tcp_sendmsg`, `tcp_recvmsg` and so on) fire for every process on the node and gate on `valid_pid()`. The gate cached only positive answers, so unselected processes paid `ns_pid_ppid()` and two bitset lookups on every hit.

I saw that #2620 added negative caching and #2814 reverted it, because a negative keyed by pid could stale-deny a child authorized later through its parent. This PR reintroduces negative caching and now the userspace drops the whole cache whenever the filter changes.

Notes:
- (bug fixed on the way) The old code reused `a_pid` for the namespaced pid and inserted under it, so a selected containerized process (host 41000, ns pid 7) was cached as `7 -> 7`, looked up as `41000`, and walked on every probe. Key 7 also marked host tgid 7 as selected. The return value was the ns pid on the BPF-cached path and the host pid on the userspace-cached path; it is now always the host pid. Verified on a target under `unshare --pid`: its host pid is the one cached as selected.
- `clearPidCache()` drains the map with `BatchLookupAndDelete` (the same API netolly uses for the flows map). Kernels without the batch API (RHEL 8) get `ErrNotSupported` once and fall back to deleting the first key until the map is empty, which needs no cursor either (bounded at 2×`MaxEntries` in case inserts outpace deletes). A clear failure is logged, never returned, so `AllowPID` always reaches its positive `Put`.
- I decided to leave the positive `Put` in `AllowPID` because a kprobe that read the old filter just before the rebuild may cache "not selected" for the new pid after the drain; the `Put` (`BPF_ANY`) overwrites it and the BPF negative insert uses `BPF_NOEXIST` so the reverse order cannot undo it.
- (Unchanged from main) The userspace `Put(pid)` assumes OBI runs in the host pid namespace (`FileInfo.Pid()` is relative to OBI's `/proc`). 
- I updated the comments in `pid_data.h`, `pid_key.h`, `pid_info.h`.

Let's talk about tests.

I ran some performance tests on a Lima VM (Ubuntu 24.04, kernel 6.8):
- One process instrumented (`python3 -m http.server 9090`, selected by `open_ports`, under steady `curl` traffic).
- An uninstrumented TCP ping-pong (client + echo server, 200k round trips of 64 B over loopback), measured with no OBI, with OBI from main, and with this PR, in the same session, with `kernel.bpf_stats_enabled=1` for kernel-side program time.
- Median of 5 runs. OBI's BPF time per round trip of the uninstrumented workload drops from 4.3 µs to 2.5 µs, gated kprobes from 176–282 ns to 71–132 ns per call, and the selected process stays instrumented.

| | round trips/s | OBI BPF time per round trip | gated kprobes ns/call | pid_cache |
|---|---|---|---|---|
| no OBI | 32.1k | 0 | n/a | n/a |
| main | 25.4k | 4.29 µs (20 program runs) | 176–282 | 1 entry, bench pids absent |
| this PR | 29.9k (*) | 2.36 µs | 68–136 | 85 entries, 84 negatives, bench pids cached as 0 |

(*) median of a noisy run (26.8k–33.3k across the 5 samples); earlier same-session runs of this PR gave 26.8k–27.2k against the same 32.1k baseline, i.e. main −21%, this PR −16%. The kernel-side BPF time is the stable measurement: 2.36–2.54 µs across all runs vs 4.1–4.3 µs on main.

Finally, I ran a churn test for the drain: a script spawned ~3300 short-lived processes per second, each opening one TCP connection (one new host pid -> one new negative entry), while the selected target was restarted every 5s to trigger rebuilds. Over 40s: `pid_cache` saturated at ~2700 entries, 19 rebuilds, 19 drains ended on an empty map, 0 hit the bound, up to 2956 entries deleted per drain, 21 left after the final drain with no churn. 

Limitations/Follow-ups:
- From the perf tests I saw that `obi_socket__http_filter` (~220 ns × 4 per round trip) is unchanged because it is not pid-gated, and it now dominates OBI's BPF time on this workload.
- Two ways a cached answer can be stale, both bounded by "until the next filter rebuild or LRU eviction": host pid reuse (a `sched_process_exit` delete would close it) and the ordering race between the bitset write and the drain, for processes authorized only through their parent (the discovered process itself is covered by the `Put`).
- Each `AllowPID`/`BlockPID` now drains the cache (two syscalls per entry) in addition to the `3001` segment writes, and a discovery burst multiplies it. It should be fine.
- If BPF ever inserts faster than the drain deletes for `2 × MaxEntries` iterations, the drain stops with leftovers and logs a warning. Not observed at ~3300 inserts/s.
- Integration test for the revert scenario (parent forks a worker, OBI starts after both are busy, the worker's spans must appear). It should go in the multiprocess suite.
- `pid_cache` has 3001 entries, the same constant as `valid_pids`, but now every process that touches a network kprobe occupies the cache. On a node with more than ~3000 network-active processes the LRU evicts continuously and the benefit shrinks toward today's behaviour, never below it. Maybe we can give `pid_cache` its own, larger constant and keep `valid_pids` as is?
- Parent authorization stays one hop, as on main: a process passes if it or its parent is selected; grandchildren rely on discovery. Should we do something about it?

I added some tests but I think it's a tricky piece of OBI, so I'm open to suggestions!

## Validation


pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -156,6 +156,49 @@ func (p *Tracer) rebuildValidPids() error {
...
+func (p *Tracer) clearPidCache() error {
+	cache := p.bpfObjects.PidCache
+	if cache == nil {
+		return nil
+	}
+
+	var key uint32
+	deleted := 0
+
+	for range 2 * cache.MaxEntries() {
```
NimrodAvni78, line 231: this costs 2×MaxEntries `syscalls` to delete the cache, i think 
`pkg/internal/netolly/ebpf/flow_map_reader_batch.go` has `BatchLookupAndDelete`which drains the whole map in a handful of syscalls. can we try generalizing that and using it instead?

pkg/internal/ebpf/generictracer/generictracer_test.go
NimrodAvni78, line 4: this needs `//go:build linux && privileged_tests` to actually run the tests
or move them seperatly

NimrodAvni78 (changes requested):

pkg/internal/ebpf/generictracer/generictracer.go
pinoOgni: I remember thinking about `BatchLookupAndDelete` (and I thought I had added it as a follow-up because I wasn't convinced..). Let me check!

pinoOgni: Oh yes! I can do the same, use `BatchLookupAndDelete` and add a fallback to the first-key drain on `ErrNotSupported` or netolly add a config field. Thanks!

pinoOgni:
Regarding https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3458/commits/92c9429c9ca26e4fd82fe9ebe5cfbb2e2b0d0795  
The choice between "batch" and "per-key" modes is managed via a persistent boolean flag set once, following the same logic used by netolly's `flowMapReaderChooser` upon the initial call: the system attempts batch mode, and if an `ErrNotSupported` error occurs, it logs the event once and switches to "per-key" mode for the tracer's entire lifecycle. 
I did not include `force_bpf_map_reader` (or a another config field) because the drain procedure runs only when a process is detected or terminates. 
If we want something like netolly, I'm fine with that.

bpf/pid/pid.h
```diff
@@ -42,47 +42,58 @@ static __always_inline u8 pid_matches(pid_data_t *p) {
...
-    const u32 a_pid = id >> 32;
+    const u32 host_pid = id >> 32;
     // accept all PIDs if debugging OTEL_EBPF_BPF_PID_FILTER_OFF option is set
     if (!filter_pids) {
-        return a_pid;
+        return host_pid;
     }
 
-    u32 *found = bpf_map_lookup_elem(&pid_cache, &a_pid);
+    u32 *found = bpf_map_lookup_elem(&pid_cache, &host_pid);
```
MrAlias, line 72: `pid_cache` now keeps negative answers across task lifetimes, but this key contains only the reusable host TGID and there is no exit-time deletion. If an unselected process caches `0`, exits, and its TGID is reused by a worker whose parent is already selected, no filter rebuild is required; this lookup returns the old `0` before `task_matches()` can observe the selected parent. An LRU map is capacity-based rather than time-based, so that can suppress the worker indefinitely when the map is not under eviction pressure.

Could we invalidate the entry on process exit or include a process-lifetime discriminator, and cover PID reuse for a parent-authorized child?

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -156,6 +163,88 @@ func (p *Tracer) rebuildValidPids() error {
 		}
 	}
 
+	// pid_cache also holds negative answers, which the new filter may
+	// invalidate. Clearing it after the segments are written makes every
+	// process re-evaluate once against the new filter. A failed clear must
+	// not fail the rebuild: the filter is already correct.
+	if err := p.clearPidCache(); err != nil {
```
MrAlias, line 170: The drain does not form a barrier with in-flight `valid_pid()` calls. A call can miss the cache and read the old filter, this rebuild can then update the bits and finish the drain, and the call can finally insert the old answer at `pid.h:90` or `pid.h:96`. After `AllowPID`, the final put repairs only the explicitly allowed PID, so a child accepted through its parent can remain cached as `0`. After `BlockPID`, there is no final repair, so a stale positive keeps the removed process running through the BPF instrumentation even though userspace later filters its spans.

Could we version cache entries against an atomically published filter generation, or use an equivalent protocol that rejects pre-rebuild inserts, and cover both post-invalidation interleavings?

MrAlias (commented):
The cache entries can outlive both task identities and filter generations, leaving selected child processes rejected or removed processes instrumented in BPF.

bpf/pid/pid.h
pinoOgni: Yeah, I thought about it but I didn't add it in the limitations because I’m not sure how likely this use case is. I got it that the mechanism is real, though it needs a few things to line up at once. 

TGIDs are handed out cyclically, for example we have `pid_cache {10:10, 11:11,....,20:0...}`; an instrumented process forks a new process and the kernel assigns TGID 20 to that new process, it can only happens after the allocator wraps the whole pid space (32768 by kernel default). 
On top of that, the entry belongs to a dead process, so it's never looked up again and is the coldest entry in a 3001-slot LRU, and any `AllowPID`/`BlockPID` anywhere on the node drains the cache entirely. 
Any child of a tracked process gets an explicit `AllowPID` at the next discovery poll (5 s by default); the matcher inherits a parent's match by PPid, so this isn't limited to children sharing the parent's executable and that AllowPID writes the positive entry directly. The issue is about a child that exits within the poll interval, accepted only by the kernel-side parent check, which would lose its spans.

That said, I’ve considered two possible solutions (based on your suggestions):
- a probe on `sched_process_exit`, which means a new attach point firing on every task exit node-wide and still wouldn't help after a filter change
- a lifetime discriminator in the cache value, which the repo already does in `set_go_h2_owned_stream_process_identity`. 

I think I'll go with the second.

pkg/internal/ebpf/generictracer/generictracer.go
pinoOgni: Right, and I think that after `AllowPID` a child accepted through its parent can stay cached as 0 until the next rebuild which is a narrow version of what #2814 reverted. 

I'll think it over a bit more, just to be safe, but I believe the generation-based approach should work.

bpf/pid/pid.h
MrAlias: You're right that this is unlikely. Under normal PID allocation, the kernel must wrap the PID space before it can reuse the same PID. The old cache entry may also be evicted or cleared before that happens. The `32768` estimate is not a hard bound, though: it is only the base default, and threads consume IDs from the same space.

I think you are also correct and I overstated the impact. If discovery never sees the replacement process, userspace has no entry for it and drops its spans anyway. If discovery does see it, `AllowPID` repairs the cache entry.

I agree that recording the process start time in the cache is cleaner than adding an exit hook. On a cache hit, we need to compare the cached start time with the current process leader's start time. The userspace cache write must either include that same start time or be removed. This solves PID reuse; the generation fix is still needed when a filter update races with a cache insert.

rafaelroquetto:
I haven't had the chance to fully review the PR yet, but I have a question about the implementation: I understand the problem it is trying to solve (checking the pid map in the kprobes preambles is expensive). 

My question is: why can't we simply settle on matching against the host pid?

In the remote past, I encountered some performance issues that resemble what this PR description's is describing - there are simply too much hashing going on.

My intuition at that time would have been to:

 1. resolve each new discovered process to a final host pid - which is a mere `uint32_t`.
 2. populate a whitelist map whose key is a mere `uint32_t` -> that is very fast to hash

I'd actually go the extra mile and compare the above against an implementation that uses a bitmap instead: on Linux, the maximum PID value is 2^22 bits -> 512 KB - you could maybe allocate a `MAP_TYPE_ARRAY` and use it as a bitmap - no hashing whatsoever, _very_fast.

Would that be feasible?

NimrodAvni78 (approved):

pinoOgni:
Hi @rafaelroquetto and @MrAlias. I took some time to think about @rafaelroquetto's idea and yes, I think it's feasible. 

We can remove `pid_cache` and we can also resolve @MrAlias's two threads, because there is no cached answer left to race with a rebuild or to outlive a process. Moreover, it should also be faster than the current cache and I'll run some tests.

`valid_pids` is already an array used as a bitset and the cost is `ns_pid_ppid()` (~17 kernel reads to get the namespaced pid, parent pid and namespace inode). Your proposal means `valid_pid()` no longer needs `ns_pid_ppid()` at all, and that is where the time goes!


So, the solution is:
- `valid_pids` becomes a bitmap indexed directly by pid: 2^22 bits, a 512 KB `BPF_MAP_TYPE_ARRAY`.
- `valid_pid()` checks the process's own bit, then its parent's (one hop, as today). No cache, no drain.
- Userspace sets or clears one bit per `AllowPID/BlockPID` instead of rewriting 3001 segments.

Host pid alone only works when OBI runs in the host pid namespace. As a sidecar OBI only sees pod pids, so it can't publish host pids. So I would key by the pid as OBI sees it:
- OBI in the initial pid namespace (checked at startup: ` /proc/self/ns/pid inode` is `PROC_PID_INIT_INO`): the tgid, which is free.
- OBI in a pod: `bpf_get_ns_current_pid_tgid()` on OBI's own namespace. It fails for any task outside the pod, so the rest of the node is rejected before the map is read. That also helps nodes running several sidecars, where today every process pays each sidecar's check. (The gate is never told what OBI's scope is. So for a process in another pod it runs the full check before finding out it could never match)

Some limitations (sidecar mode only; host mode never calls the helper):
- A process that creates its own pid namespace inside the pod (unshare -p) fails the helper and is rejected; the current code handles it. (IMHO I think it's a rare case)
- The helper is only available to tracing programs on older kernels, but `valid_pid()` is also called from tpinjector's`sk_msg` program. There I would read the pid at OBI's namespace level from the task instead, with the helpers `valid_pid()` already uses today.
- RHEL 8.9+ lists the helper for kprobes ([release notes](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/8.9_release_notes/available_bpf_features)), which covers the 8.9 and 8.10 CI VMs; I haven't checked older 8.x releases.

I can validate the new solution with the same perf and churn tests I used here, plus a sidecar run. Would you prefer I continue to work on this PR, or open a new one? And are the limitations acceptable?

MrAlias:
Thanks for digging into this. This direction makes sense to me, and removing `pid_cache` should address the two race/lifetime concerns I raised.

I’d prefer continuing in this PR since the new approach solves the same underlying problem and keeps the design discussion in one place.

The nested PID namespace limitation seems acceptable if it is documented and covered by a test. However, we should not regress support for older RHEL 8 kernels that OBI currently supports. Please verify helper availability across the supported range and retain a fallback where it is unavailable. The proposed performance, churn, and sidecar tests sound like the right validation.

rafaelroquetto:
Sounds good to me!

Issue #2646, opened 2026-07-13
Make language detection more fine-tuned for JVM applications


tsipo:
### Component

OBI (eBPF instrumentation)

### Problem

Currently, the runtime/language idetification for JVM-based application is rather course:

- [`instrumentableFromModuleMapSharedLib()`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/internal/procs/proclang.go#L24) detects an `InstrumentableJava` by a module name starting with `libjvm.so`.
- [`instrumentableFromSymbolName()`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/internal/procs/proclang.go#L55) detects an `InstrumentableJava` by a symbol name starting with `JVM_` or with `graal_`.

It is desribale to have a more fine-tuned language detection for applications using the JVM as their application runtime. Whereas some of the instrumentation mechanisms may be shared (as they all use the JVM), others could be different. And even if it's merely for identification purposes, it's better to know the exact language.
Think of the following languages:

- Java
- Scala
- Kotlin
- Groovy
- Clojure
- JRuby
- Jython

(at least the first 4 or 5 are very predominant)

As for the non-Java languages, we can try to identify the language by either trying to detect their standard library jars (unless these are renamed) or by detection of **core marker classes** in their class files (which is more CPU-intensive).

### Caveats

1. Polyglot use-case: the JVM ecosystem can be "mixed". Think of a standard Java application that has the standard library jar file of Groovy or Kotlin on the classpath to support a templating engine. Even if we find such jar file (or marker classes) - how do we know that's indeed the main programming language?
2. It looks like the identificaiton of Java itself should be done by elimination (i.e. it's not any of the other languages).

### Describe the solution you'd like

There are a few paths here:

1. Identify the exec command e.g. `scala`, `groovy`, `clojure`, `jruby`, `jython`. This **may** be a wrapper script setting environment/classpath and invoking `java` under the hood - so we may need to detect it on the parent process level (and avoid overriding it in the solution for [Container language detection doesn't follow process tree (parent may override descendant)](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2344)).
[...]

PR #3459, opened 2026-09-16
Adds a jvm.language semconv attribute

rupeshkoushik07:
## Summary

Describe the change briefly.

Working on [Issue #2646 ](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/2646)

Adds a `jvm.language` semconv attribute so the JVM's programming language can be reported separately from the runtime. 

I used jvm.language to match the per-runtime namespaces already in the registry (jvm.*, cpython.*, dotnet.*

Next steps: detection of the actual language per process 

I'm currently researching detection from the process's main class: find the main class from the command line (or the jar manifest), then check that class file for compiler markers

Any advice or suggestions on the approach would be appreciated.

## Validation


MrAlias:
The SIG decision in #2646 calls for defining the sub-language convention before implementing detection, and this PR follows that sequence. Perhaps too literally.

Overall, this looks like a good addition. However, my concern is that the detector’s behavior determines the contract being added: polyglot JVM processes and identifying Java by elimination remain unresolved, while this PR fixes a single closed enum without defining those cases. Or, even if they are possible.

Before merging, could we add or link a detector prototype, draft, or PoC covering Java, Kotlin, Scala, and a mixed-language application?

Without an end-to-end view of this feature, merging this PR is at risk of introducing cruft that never ends up being used.

Until then, please mark this as a draft.

PR #3463, opened 2026-09-17
fastcgi: add HTTP attributes to spans

NimrodAvni78:
## Summary

A FastCGI server span carries no `url.scheme`, and no `url.path` when `REQUEST_URI` is absent. Both are `required` on an HTTP server span in semconv, so every PHP-FPM request OBI reports is missing them.

The metadata is already on the wire, and OBI already parses it. `detectFastCGI` builds the full params table and then reads three keys out of it, discarding the rest — including the scheme, the host, and the two other forms of the request path that nginx and php-fpm send by default.

## What changes

`detectFastCGI` returns the request metadata it parsed rather than a method/uri/status triple:

- **scheme** from `REQUEST_SCHEME`, falling back to `HTTPS` being `on` or `1`. Neither present leaves it empty rather than guessing `http`, which would be wrong for any TLS-terminated site.
- **path** from `REQUEST_URI`, falling back to `DOCUMENT_URI` and then `SCRIPT_NAME`. The fallbacks are the rewritten and resolved forms, which still name the path when the original request line was truncated out of the buffer or is not configured.
- **host** from `HTTP_HOST`, falling back to `SERVER_NAME`.

`TCPToFastCGIToSpan` writes the scheme and host into `Statement` in the `scheme;host` form the rest of the pipeline expects, which is where `HTTPScheme` reads `url.scheme` back out.

`server.address` is deliberately unchanged: the HTTP server span reads it from `SpanHost`, which does not consult `Statement`.

## Validation


pkg/ebpf/common/fast_cgi_detect_transform.go
```diff
@@ -137,10 +154,40 @@ func parseHeader(b *largebuf.LargeBuffer) ([]byte, error) {
 	}
 }
 
-func detectFastCGI(b, rb *largebuf.LargeBuffer) (string, string, int) {
+// cgiScheme reports the scheme of the original client request. REQUEST_SCHEME
+// carries it directly; HTTPS is the older convention and is set to a truthy
+// value only for TLS. Neither present means the front end did not say, and
+// guessing would be wrong for any TLS-terminated site.
+func cgiScheme(kv map[string]string) string {
+	if scheme := kv[requestSchemeKey]; scheme != "" {
```
MrAlias, line 170: `REQUEST_SCHEME` reflects the immediate request nginx received. Behind a TLS-terminating load balancer it is commonly `http`, while the original scheme can be available as `HTTP_X_FORWARDED_PROTO=https`; returning here records `url.scheme=http` on the PHP-FPM server span. The HTTP semconv requires the [original client scheme when known](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-server-span).

Could we prefer a validated forwarded-proto value before `REQUEST_SCHEME`/`HTTPS`, and add a case with `REQUEST_SCHEME=http` plus `HTTP_X_FORWARDED_PROTO=https`?

MrAlias (commented):
The direct FastCGI metadata path is well covered, but the scheme precedence still mislabels requests when a TLS-terminating proxy supplies the original scheme via `HTTP_X_FORWARDED_PROTO`.

mariomac (approved):
LGTM

pkg/ebpf/common/fast_cgi_detect_transform.go
NimrodAvni78: Good catch, fixed!

MrAlias (approved):

PR #3464, opened 2026-09-17
Add TCP successful connections stat metric

pinoOgni:
## Summary

This PR adds `obi.stat.tcp.successful.connections` which is a counter of TCP connections that completed the three-way handshake, as a new Statsolly metric next to the existing `obi.stat.tcp.failed.connections` counter.

## Validation


mmat11 (approved):

pkg/internal/statsolly/stats/tracer_ringbuf.go
```diff
@@ -117,6 +119,20 @@ func readTCPFailedConnectionsIntoStat(record *ringbuf.Record) (ebpf.Stat, error)
 	}, nil
 }
 
+func readTCPSuccessfulConnectionsIntoStat(record *ringbuf.Record) (ebpf.Stat, error) {
```
MrAlias, line 122: The decoder tests are intended to cover every event's exact wire size, successful decoding, every truncated length, and fuzz seeds, but `StatsTCPSuccessfulConnection` is absent from `TestStatsEventWireSizes`, `TestParseStatDecodesEveryEventType`, and `minimalWireStats`. That leaves this new C/Go layout and its malformed-record path outside the contract that protects the other four event types. Could we add the successful event to those tables?

pinoOgni: Nice catch! Added.

MrAlias (approved):

Issue #3435, opened 2026-09-15
Node.js agent: async_hooks before-hook + fs.accessSync per callback costs 25-55% of event-loop CPU on HTTP services


mlushpenko:
### What we see

On Node 24 services traced through Beyla 3.34.0 (OBI pin b90f02c1), 30 s V8 CPU profiles put the injected agent's frames at a large share of all non-idle main-thread CPU:

| service | inbound HTTP | `emitHook` + `access` + `accessSync` self time / busy |
|---|---|---|
| API gateway A | yes | 55% (3.19 s of 5.78 s) |
| SOAP gateway, large bodies | yes | 48% (1.37 s of 2.83 s) |
| API gateway B | yes | 25% |
| queue worker, no inbound HTTP | no | 0.9% |

The frames are the `before` hook in `fdextractor.js` and its `fs.accessSync('/dev/null/obi-ctx/<fd>')` sentinel. A bpftrace uprobe on `uv_fs_access` counted ~1,100 sentinel calls/s on one pod at low load, 4-16 µs each through libuv; one large request fired 7,149 of them. The application and its dependencies do not use `async_hooks` or `accessSync` themselves, so the frames can only come from the agent.

On the SOAP gateway, enabling tracing moved p99 on the large-body route from ~190 ms to 500-650 ms while a small-body route on the same pods moved ~10%. Per-request hook counts on that pod ranged from 18 to 5,700, which matches that split.

The eBPF side is small by comparison on the same pods: `uprobe_notify_resume` median 256-512 ns per trap, about 4-5% of process CPU; `bpf_stats` over all OBI programs on the node, 568 ns per run.

### Why the hook looks like pure cost when manual spans are off

- Client-span parenting for Node goes through `find_nodejs_parent_trace` (`bpf/common/trace_parent.h`), which reads `nodejs_fd_map`. That map is fed by the fd-pair sentinel on connect/write, not by the per-callback hook.
- The per-callback hook feeds `traces_ctx_v1` via `obi_ctx__set` in `handle_async_switch` (`bpf/generictracer/nodejs.c`). Its Node consumer is `handle_node_span` (manual spans); the other readers are the log enricher and the Go runtime.
- With `nodejs.manual_spans: false` and the log enricher off, nothing reads what the hook writes, but it still runs on every in-request callback. `emitHook` alone is ~23% on the SOAP gateway, so the async_hooks dispatch is a large part of the cost, not only the syscall.

[...]

PR #3465, opened 2026-09-17
Allow option to not populate populate_trace_context

NimrodAvni78:
Part of https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/3435

## Summary

`traces_ctx_v1` is a map for OBI to export trace context data, used internally for trace_log correlation, but externally used currently only for ebpf-profiler trace-profile correlatio, OBI populates it unconditionally.
Keeping it aligned with the active request costs a refresh on every async context switch of the instrumented runtime. On Node.js that is an `async_hooks` before hook firing on every callback, measured at a double-digit share of event-loop CPU on request-heavy services. On Go it is a `runtime.casgstatus` uprobe on every goroutine status transition.

This adds a single global lever and makes every writer follow it.

## Config

```yaml
ebpf:
  populate_trace_context: false # default
```

`OTEL_EBPF_BPF_POPULATE_TRACE_CONTEXT`.

`Config.PopulateTraceContext()` is the effective predicate: the new setting, or the log enricher, or the Node.js manual span bridge. OBI's own readers turn population on regardless; a reader outside OBI (a profiler, another eBPF program reading the pin) cannot announce itself, so it opts in explicitly. Profile correlation therefore needs the setting.

## What the lever does

- `g_trace_ctx_map_enabled` (`bpf/common/globals.h`) gates `obi_ctx__set` / `obi_ctx__del`, so all writers across the generic tracer and the Go tracer compile away. Readers stay unguarded: a lookup on an unpopulated map finds nothing, which is what a thread with no in-flight request already returns.
- Node.js: the `async_hooks` before hook is no longer installed (`OBI_CTX_HOOK_ENABLED` in `fdextractor.js`, substituted from the same predicate). Client-span parenting is unaffected - it comes from the fd-pair map the `net` prototype wraps maintain, which stays installed whenever traces are on.
- Go: the `runtime.casgstatus` uprobe is not attached, and the per-goroutine span stack (`obi_ctx_stacks`) is not maintained, since it exists only to decide what to publish.
- The pin outlives the process, so a non-populating run empties it once at startup (`drainTraceContextMap`) rather than leaving a previous run's entries for a reader to match against recycled `pid_tgid`s. After that it never touches the map: a pin the OTEP says others may share is not ours to delete from once we have opted out of writing it.

## Validation


mmat11:
we have a couple more readers:

- Logenricher
- Profile correlation
- Go channel handoff links (currently no config option) - https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/bpf/gotracer/go_runtime.c#L832
- Node.js manual spans https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/bpf/generictracer/nodejs.c#L188

I think we should warn when any of this feature is enabled (that we know of, eg. logenricher/nodejs manual spans) and this toggle is disabled

NimrodAvni78:
@mmat11 I think we have 3 options
1. warn when one of those features is enabled and map population is not
2. fail config validation and crash OBI
3. override the option when one of those fields are enabled

I would lean to 2 or 3 since its a definitive behavior. 2 is more "correct" but kind of annoying needing to turn on 2 knobs, 3 is easier but leaves config not representing what actually is going on
WDYT? we currently do 3.

refarding profile correlation we have no way of knowing its being consumed, go channel buffers we need probably another handle or something else

mmat11:
@NimrodAvni78 I think 3. would be the least annoying one, do we have any other config toggle that behaves like that?

NimrodAvni78:
@mmat11 we do 3 already here

bpf/common/globals.h
```diff
@@ -11,3 +11,11 @@ volatile const bool g_bpf_header_propagation = false;
 volatile const bool g_bpf_probe_write_user_enabled = false;
 volatile const bool g_bpf_loop_enabled = false;
 volatile const u8 g_go_h2_write_fail_step = 0;
+// Set from Config.PopulateTraceContext(): whether anything reads traces_ctx_v1.
```
mmat11: I think we can remove this comment, the const name is self explanatory -- maybe we can name it g_traces_ctx_v1_enabled so grepping works here as well

bpf/shared/obi_ctx.h
mmat11, line 28: should we just return NULL if the map isn't enabled here?

devdocs/trace-log-correlation.md
```diff
@@ -76,6 +76,22 @@ Newline-delimited JSON is handled as structured JSON: OBI enriches each JSON obj
 
 The map is **written** by the generic tracer (in `server_or_client_trace()`) whenever an HTTP request or client call is detected on the wire. When a client call ends, `obi_ctx__restore_server()` points the thread back at its enclosing server span (looked up in `server_traces` and matched by parent span id), so logs written after a nested client call keep the server context. The map is **read** by the logenricher when intercepting writes.
 
+### When the map is populated
+
+Population is not free. The per-runtime refresh below runs on every async context switch of the instrumented process — on Node.js an `async_hooks` before hook on every callback. The writers are gated on a single BPF constant, `g_trace_ctx_map_enabled`, set from `Config.PopulateTraceContext()`: the map is populated only when something reads it.
+
+The Go tracer sets that constant unconditionally. Its channel-handoff span links resolve through the per-goroutine span stack that `runtime.casgstatus` maintains, and they regress when it is not kept, so Go populates the map either way. Narrowing Go is follow-up work.
```
mmat11: is this line true? I think we do it for Go as well in this PR?

devdocs/trace-log-correlation.md
```diff
@@ -76,6 +76,22 @@ Newline-delimited JSON is handled as structured JSON: OBI enriches each JSON obj
...
+
+| Reader | Turns population on |
+|---|---|
+| Log enricher | `ebpf.log_enricher.services` is non-empty |
```
mmat11, line 89: question: should we add configv2 examples here? is configv2 handled in the code?

pkg/config/ebpf_tracer.go
```diff
@@ -143,6 +143,24 @@ type EBPFTracer struct {
...
+	// Node.js means an `async_hooks` before hook running on every callback. OBI
+	// therefore only populates the map when something reads it. Its own readers -- the
+	// log enricher and the Node.js manual span bridge -- turn population on regardless
+	// of this setting; a reader outside OBI has no way to announce itself, so it opts
+	// in here.
+	//
+	// The Go tracer is not covered: it keeps populating the map either way, because
+	// its channel-handoff span links regress when the per-goroutine span stack is not
+	// maintained. Narrowing Go is follow-up work.
+	PopulateTraceContext bool `yaml:"populate_trace_context" env:"OTEL_EBPF_BPF_POPULATE_TRACE_CONTEXT" validate:"boolean"`
```
mmat11, line 163: naming nit: I'd name this config `TracesCtxV1Enabled` to make it self-explanatory and reduce the verbosity of the comment

pkg/internal/ebpf/convenience/drain.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+const DefaultDrainBatchLen = 1024
+
+// DrainMap removes every entry from a hash map, returning how many it removed.
+//
+// BatchLookupAndDelete empties a full map in a handful of syscalls, where
+// iterating costs two per entry. RHEL8 and its derivatives ship a 4.18 kernel
+// that backports much of eBPF but not the batch map operations, so a kernel
+// that rejects the batch call falls back to iterate-and-delete.
```
mmat11:
```suggestion
// BatchLookupAndDelete empties a full map in a handful of syscalls, where
// iterating costs two per entry. A kernel that rejects the batch call falls
// back to iterate-and-delete.
```

pkg/internal/ebpf/convenience/drain.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	drained, err := batchDrain[K, V](m, batchLen)
+	if !errors.Is(err, ebpf.ErrNotSupported) {
+		return drained, err
+	}
+
+	iterated, err := iterateDrain[K, V](m)
+
+	return drained + iterated, err
```
mmat11, line 71: if batch delete failed in any part, I would just `return iterateDrain[K, V](m)` here

pkg/internal/ebpf/convenience/drain.go
```diff
@@ -0,0 +1,83 @@
...
+func DrainMap[K, V any](m *ebpf.Map, batchLen int) (int, error) {
+	if m == nil {
+		return 0, nil
+	}
+
+	if batchLen <= 0 {
+		batchLen = DefaultDrainBatchLen
+	}
+
+	drained, err := batchDrain[K, V](m, batchLen)
```
mmat11: here it says this API is not safe to use with certain map types: https://github.com/cilium/ebpf/blob/f788cf14bd26765c2023b273de26008c869a732f/map.go#L1262

should we check the map type and batchDrain based on it while falling back to iterateDrain for other types?

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -729,6 +732,32 @@ func (p *Tracer) AlreadyInstrumentedLib(id uint64) bool {
...
+// re-established by the next runtime.casgstatus transition.
+func (p *Tracer) drainTraceContextMap() {
+	if p.cfg.PopulateTraceContext() || p.bpfObjects.TracesCtxV1 == nil {
+		return
+	}
+
+	drained, err := ebpfconvenience.DrainMap[uint64, BpfObiCtxInfoT](p.bpfObjects.TracesCtxV1, 0)
+	if err != nil {
+		p.log.Warn("could not fully drain the trace context map, a reader may see stale context",
+			"error", err, "entries", drained)
```
mmat11:
```suggestion
			"error", err, "drained_entries", drained)
```

pkg/internal/ebpf/generictracer/generictracer.go
```diff
@@ -766,6 +771,10 @@ func (p *Tracer) Run(
 		p.log.Error("BPF Pids map is not created yet, this is a bug.")
 	}
 
+	if !p.traceCtxMapEnabled {
+		ebpfconvenience.DrainTraceContextMap[BpfObiCtxInfoT](p.log, p.bpfObjects.TracesCtxV1)
```
grcevski, line 775: Why do we need to drain the map, is it to delete stale entries from a previous run? I think I'm missing some context here.

grcevski (approved):
LGTM! This looks good to me, we just need to document properly what you lose when you disable this option and when to use it. I had a question about the drain part, but I think it's just me not having enough context.

pkg/internal/ebpf/generictracer/generictracer.go
NimrodAvni78: yes exactly, so external consumers wont still be pooling old context info

mmat11 (approved):

PR #3467, opened 2026-09-17
genai: always emit gen_ai.operation.name

NimrodAvni78:
## Summary

`gen_ai.operation.name` is `required` in semconv, but OBI could fail to produce it: three of seven provider parsers had gaps, and where a parser produced nothing the emitters dropped the attribute or emitted an empty string.

## Parser fixes

**OpenAI** matched the request path exactly, so any deployment serving the API under a prefix — Azure, or a gateway mounted at `/openai/v1/...` — reported no operation. The switch now matches on a suffix. `/v1/completions` had no case at all; it now maps to `text_completion`, tested after chat completions since that also ends in `/completions`.

**OpenAI-compatible** set `APIType` but not `OperationName` on its `/v1/responses` case, unlike its three sibling cases and `openai.go` for the same path.

**Anthropic** read the operation from the response body's `type`, which is `error` on an error response and absent on one truncated out of the capture buffer. The request path names the operation either way, so it is now the fallback.

Gemini, Bedrock, Ollama and Qwen were already total.

## Unknown operations report `_OTHER`

The gap cannot be closed entirely: OpenAI detection is header-driven and gateways are matched by host, so an unknown endpoint still reaches the parser with nothing to classify. Both switches now default to `_OTHER` — the value semconv already uses for this case on `http.request.method` — and the five emitter branches route through one helper so a new provider inherits it.

The metric getter clamps the same way, keeping the span and metric views of one call in agreement. It still omits the attribute on spans carrying no GenAI data, so no non-GenAI series gains a dimension.

`gen_ai.operation.name` is now always present on a GenAI span and can be declared `required`.

## Validation


pkg/ebpf/common/http/anthropic.go
```diff
@@ -142,6 +142,10 @@ func AnthropicSpan(baseSpan *request.Span, req *http.Request, resp *http.Respons
 		}
 	}
 
+	if parsedResponse.Type == "" {
```
MrAlias: An Anthropic error body sets the top-level `type` to `error`, so this guard does not take the request-path fallback in the scenario covered by `TestAnthropicSpan_ErrorResponseStillReportsOperation`. `/v1/messages` errors therefore export `gen_ai.operation.name=error` instead of `message`; changing that test to assert `request.MessageOperationName` fails with actual `"error"`. Could we treat the error response type as non-operation data and make the test assert the exact operation?

```suggestion
if parsedResponse.Type == "" || parsedResponse.Type == "error" {
```

pkg/ebpf/common/http/openai.go
```diff
@@ -135,20 +135,29 @@ func OpenAISpan(baseSpan *request.Span, req *http.Request, resp *http.Response)
...
-		switch path {
-		case "/v1/chat/completions":
+		switch {
+		case strings.HasSuffix(path, "/chat/completions"):
 			parsedResponse.OperationName = request.ChatOperationName
 			parsedResponse.APIType = "chat_completions"
-		case "/v1/embeddings":
+		case strings.HasSuffix(path, "/completions"):
+			parsedResponse.OperationName = request.CompletionOperationName
+			parsedResponse.APIType = "text_completions"
```
MrAlias, line 150: `openai.api.type` is a closed enum in `schemas/obi/groups/openai/registry.yaml`; its members are `chat_completions`, `responses`, and `embeddings`. A direct `/v1/completions` span now exports `text_completions`, so Weaver reports `undefined_enum_variant` and the new parser test cannot catch the emitted-contract violation. Could we either add this member to the registry and regenerate the schema docs, or omit `openai.api.type` for this endpoint?

schemas/obi/groups/gen_ai/registry.yaml
```diff
@@ -26,6 +26,13 @@ groups:
   #   emits only the upstream-defined 'execute_tool', and nothing at all for
   #   methods that are not tool calls.
   #
+  #   An endpoint OBI cannot identify reports '_OTHER', the same marker
+  #   upstream gives `http.request.method` for a method the instrumentation
+  #   does not know. It is deliberately not `custom`: that marks a value we
+  #   know but have no member for, and since this attribute is an open string
+  #   every operation OBI can read is emitted verbatim, so the only residual
+  #   case is one it could not identify at all.
```
MrAlias, line 34: This fallback is not reflected in the attribute definition later in this file: it still says an unavailable operation is omitted, and its note/examples do not mention `_OTHER`, so the generated `site/docs` will not describe a value OBI now emits. Could we update that definition and run `make generate-schema-docs`?

MrAlias (changes requested):
The operation-name handling still has contract gaps on Anthropic error responses and legacy OpenAI completions, and the schema documentation remains inconsistent with the new `_OTHER` behavior.

pkg/ebpf/common/http/openai.go
NimrodAvni78: added `text_completions` as a valid enum member

schemas/obi/groups/gen_ai/registry.yaml
NimrodAvni78: Added the `_OTHER` reference in notes + examples

pkg/ebpf/common/http/anthropic.go
```diff
@@ -142,6 +142,10 @@ func AnthropicSpan(baseSpan *request.Span, req *http.Request, resp *http.Respons
 		}
 	}
 
+	if parsedResponse.Type == "" || parsedResponse.Type == anthropicErrorType {
+		parsedResponse.Type = anthropicOperation(req)
+	}
```
MrAlias: `parsedResponse.Type` still drives the emitted operation on successful responses, while this path fallback only runs for empty or error bodies. Anthropic's legacy `/v1/complete` [success response type is `completion`](https://github.com/anthropics/anthropic-sdk-python/blob/v0.75.0/src/anthropic/types/completion.py#L39-L42), but the fallback returns `text_completion`, so one endpoint is split across span names and metric series. The helper also defaults every unrecognized path to `message`, which groups unknown endpoints with the Messages API instead of `_OTHER`.

Could we make the request path authoritative for all Anthropic responses, switch explicitly on `/v1/messages` and `/v1/complete`, and return `request.OtherOperationName` by default? Tests covering successful, error, truncated, and unknown paths would pin the invariant.

pkg/ebpf/common/http/anthropic_test.go
```diff
@@ -515,3 +515,29 @@ data: {"type":"message_stop"}
...
+func TestAnthropicSpan_ErrorResponseStillReportsOperation(t *testing.T) {
+	req := makeRequest(t, http.MethodPost, "http://api.anthropic.com/v1/messages", anthropicRequestBody)
+	resp := makeGzipResponse(t, http.StatusBadRequest, anthropicHeaders(),
+		`{"type":"error","error":{"type":"invalid_request_error","message":"bad"}}`)
+
+	span, ok := AnthropicSpan(&request.Span{}, req, resp)
+
+	require.True(t, ok)
+	require.NotNil(t, span.GenAI.Anthropic)
+	assert.Equal(t, request.MessageOperationName, span.GenAI.Anthropic.Output.Type)
```
MrAlias: This assertion captures the new operation value, but `internal/test/oats/ai/yaml/oats_anthropic_test.yaml` still expects the error span name to be `error claude-sonnet-4-6` and does not require `gen_ai.operation.name: message`. That expectation now fails on both AI OATS jobs even though this unit test passes. Could we update the OATS span name and operation attribute with this change?

MrAlias (commented):
The follow-up fixes the earlier contract gaps, but Anthropic operation handling still splits and misclassifies endpoints, and the stale error-span expectation keeps the AI integration suite red.

pkg/ebpf/common/http/anthropic.go
```diff
@@ -154,6 +161,22 @@ func AnthropicSpan(baseSpan *request.Span, req *http.Request, resp *http.Respons
 	return *baseSpan, true
 }
 
+// anthropicOperation names the operation from the request path, which is the
+// only part of the exchange that names it consistently. A response body reports
+// `message`/`completion` when the call succeeded, `error` when it failed, and
+// nothing at all when it was truncated out of the capture buffer, so reading it
+// would split one endpoint across several operation names.
+func anthropicOperation(req *http.Request) string {
```
grcevski, line 169: I think these match too broadly /v1/messages/count_tokens and /v1/messages/batches will end up as message.

https://platform.claude.com/docs/en/api/go/messages/count_tokens
https://platform.claude.com/docs/en/api/messages/batches

pkg/ebpf/common/http/openai.go
grcevski: I think this will fail to match things like [chatkit.session](https://developers.openai.com/api/reference/typescript/resources/beta/subresources/chatkit/subresources/sessions/methods/create), there might be others.

pkg/ebpf/common/http/anthropic.go
```diff
@@ -154,6 +161,22 @@ func AnthropicSpan(baseSpan *request.Span, req *http.Request, resp *http.Respons
...
+	switch path := requestPath(req); {
+	case strings.Contains(path, anthropicMessagesPath):
```
MrAlias: `/v1/messages/count_tokens` also matches this branch. Its request has the same top-level `model` and `messages` fields, so the HTTP/2 body heuristic accepts it and OBI exports a token-count request as `gen_ai.operation.name=message` instead of `_OTHER`. Could we match the trimmed path suffix exactly while retaining proxy prefixes, and add this endpoint to the unknown-path cases?

MrAlias (commented):
The follow-up makes Anthropic operation names consistent across response outcomes and fixes the OATS expectation, but the revised path matching still folds non-generation endpoints into generation operations.

pkg/ebpf/common/http/openai_compatible.go
```diff
@@ -91,7 +92,10 @@ func OpenAICompatibleSpan(baseSpan *request.Span, req *http.Request, resp *http.
 			parsedResponse.OperationName = request.EmbeddingOperationName
 			parsedResponse.APIType = "embeddings"
 		case strings.Contains(req.URL.Path, "/v1/responses"):
+			parsedResponse.OperationName = request.ResponseOperationName
```
MrAlias: `GET /v1/responses/{response_id}` returns the same Response shape, including `model` and `output`, so it passes this parser and the substring match now labels a retrieval request as the `response` generation operation. That also records an inference duration for a request that generated nothing. Could we use the same trimmed suffix matching as `OpenAISpan` and cover an ID-suffixed path as `_OTHER`?

pkg/ebpf/common/http/anthropic.go
NimrodAvni78: Fixed: /v1/messages and /v1/complete are now matched as path suffixes instead of substrings

pkg/ebpf/common/http/openai_compatible.go
NimrodAvni78: Done, `/v1/responses/{id}` and `/v1/responses/{id}/cancel` now report _OTHER while prefixed deployments still resolve

pkg/ebpf/common/http/anthropic.go
NimrodAvni78: Fixed: /v1/messages and /v1/complete are now matched as path suffixes instead of substrings

MrAlias (approved):

grcevski (approved):
LGTM!

PR #3469, opened 2026-09-17
Add opt-in partial config migration output

### LINKED ISSUE #3468: Add opt-in partial output to config migration
## Problem

`obi config migrate` intentionally emits no Config v2 output when any Config v1 field cannot be preserved. This strict default prevents accidental behavior changes, but it makes manual migration cumbersome: users must remove reported fields, rerun the command, and repeat until migration succeeds before adding the unsupported settings back in their Config v2 form.

A reported configuration encountered these paths:

- `otel_metrics_export.endpoint`
- `otel_traces_export.endpoint`
- `otel_traces_export.insecure_skip_verify`
- `otel_traces_export.instrumentations`

After those fields were removed, strict v1 decoding stopped earlier on a top-level `bpf_debug` key. The valid v1 path is `ebpf.bpf_debug`; the ordinary v1 loader may previously have ignored the unknown top-level key, while the migration command correctly rejects it. This also means users might need several edit-and-rerun cycles before seeing all required manual work.

The exporter paths above can represent expected migration limitations, such as implicit OTLP/HTTP selection, custom TLS verification, or instrumentation selections that Config v2 cannot preserve. The request is not to weaken the default lossless migration contract.

## Proposed behavior

Add an explicit opt-in mode, for example `--allow-partial` or `--best-effort`, that:

- emits a valid Config v2 document containing every safely migrated setting;
- reports every unsupported, lossy, or unknown Config v1 path that was omitted;
- clearly identifies the output as incomplete and requiring manual review;
- validates the generated Config v2 document before writing it;
- remains deterministic for the same input and environment; and
- provides scriptable status that distinguishes complete migration, partial migration, and failure.

Strict lossless migration should remain the default and should continue to emit no output when migration cannot preserve the input.

## Acceptance criteria

- [ ] The existing strict behavior remains unchanged without the new option.
- [ ] The opt-in mode writes valid Config v2 for the supported subset of an otherwise valid v1 configuration.
- [ ] The report lists exact omitted paths, including unsupported known fields and unknown fields such as a misplaced top-level `bpf_debug`.
- [ ] Callers can distinguish a complete result from a partial result without parsing human-readable text.
- [ ] Tests cover unsupported exporter settings, instrumentation selection, and unknown v1 fields.
- [ ] Migration documentation explains the partial-output safety contract and manual-review workflow.

## Related

- #2534

MrAlias:
## Summary

- add `obi config migrate --allow-partial` while preserving strict migration as the default
- emit validated Config v2 output and report unsupported, lossy, and unknown v1 field paths
- return exit code `3` for partial output so scripts can distinguish it from complete migration
- document the safety contract and cover standalone, receiver, and invalid-input behavior

Partial output is explicitly a manual conversion aid. The report identifies fields that were not preserved exactly, and the generated document must be reviewed and completed before deployment.

## Testing

- `make generate`
- `go test ./cmd/obi/internal/configcmd -count=1`
- `go vet ./cmd/obi/internal/configcmd`
- `golangci-lint run ./cmd/obi/internal/configcmd --timeout=6m`
- `make lint-markdown`
- `make compile`

Closes #3468


mariomac (approved):
🚢

NimrodAvni78 (approved):
LGTM

PR #3470, opened 2026-09-17
Fix overlapping Go large buffers

grcevski:
## Summary

When I added large buffer support for Go, there was a particular edge case for HTTP 1.1 clients, where the response from the server can keep coming after the roundtrip return probe fires. For this I introduced these delayed HTTP client responses, so we keep the processing waiting until all data comes back.

The problem is for keep alive, so on rapidly firing client requests we can technically start a new outgoing large buffer before we consumed the previous request in userspace.

This PR adds a test for that scenario and fixes the situation, by tracking the span_id on the large buffers too.

## Validation


MrAlias (approved):
🚀

PR #3471, opened 2026-09-17
Install CA certificates in Rust test images

MrAlias:
## Summary

- Install the system CA bundle in the Rust and Rust SSL test runtimes.
- Allow Reqwest 0.13 to initialize its certificate store successfully.

## Context

The rebuilt `rust-0.1.0` image panics while handling `/dist` because its slim runtime has no CA certificates. Callers receive EOF responses, causing the multiprocess integration tests in #3430 to fail across architectures and kernel versions.

After this merges, the test images need to be published under a new version and the digest-pinned compose references updated.

## Validation

- `./scripts/lint-dependency-policy.sh --all`


grcevski (approved):
LGTM!

PR #3472, opened 2026-09-17
Dynamic Selector: Add k8s workload selection

damemi:
## Summary

This adds support to the dynamic selector for consumers to instrument kubernetes workloads (instead of needing the pid). For example,

```go
selector.AddK8sWorkload(&appsv1.Deployment{
    ObjectMeta: metav1.ObjectMeta{Namespace: "prod", Name: "checkout"},
}, selection.DynamicOptions{ServiceName: "checkout"})
```

It's essentially sugar on top of the pid selection, where these selections get mapped back to an underlying pid (since that is the heart of the discovery implementation already). So selected workloads and their signal masks get unioned with selected pids.

Broken into the following commits for review:
1. Add workload identity types (no op)
2. Tracking k8s workloads alongside pids
3. Tie into discovery and trigger updates on changes (refactors pid channel to support pid and workload)
4. Net/Stats IP resolving in DynamicAppIPS
5. Rename DynamicPIDSelector->DyanmicSelector
6. Add usage example

To start, this only supports top-level workloads (Deployment, Daemonset, Statefulset, CronJob) because otherwise we would have to define a precendece/inheritence model for child workloads like ReplicaSet/Pod, specifically for things like if deployment specifies one set of attributes and pod sets another, they would both map to the same pid and possibly collide. This could be a config surface in the future api but for now we can keep it to the most common use cases.

Usage:

```go
sel := discover.NewDynamicSelector()

err := instrumenter.Run(ctx, &config,
    instrumenter.WithDynamicSelector(sel),
)

// PID
sel.AddPID(4242, selection.DynamicOptions{
    ServiceName: "checkout",
})

// or a Deployment (also accepts selection.K8sWorkloadRef)
_ = sel.AddK8sWorkload(&appsv1.Deployment{
    ObjectMeta: metav1.ObjectMeta{Namespace: "payments", Name: "checkout"},
}, selection.DynamicOptions{
    ServiceName:      "checkout",
    ServiceNamespace: "payments",
})

// later
_ = sel.RemoveK8sWorkload(/* same object or K8sWorkloadRef */)
```

## AI Disclaimer

This PR was written, reviewed, and edited with the aid of ai-coding tools. I have reviewed each commit myself to the best of my ability and welcome any feedback or questions

## Validation


pkg/appolly/discover/dynamic_k8s_workload.go
```diff
@@ -0,0 +1,252 @@
...
+// TargetsChangedNotify reports when workload selection changes such that the process watcher
+// should rescan already-tracked processes.
+func (d *DynamicPIDSelector) TargetsChangedNotify() <-chan struct{} {
+	return d.targetsChangedNotifier.Subscribe()
+}
+
+// TargetsChangedNotifyContext is the context-bound form of TargetsChangedNotify.
+func (d *DynamicPIDSelector) TargetsChangedNotifyContext(ctx context.Context) <-chan struct{} {
+	return d.targetsChangedNotifier.SubscribeContext(ctx)
+}
```
mariomac: Do we need two signatures for the same functionality?

pkg/appolly/discover/dynamic_k8s_workload.go
```diff
@@ -0,0 +1,252 @@
...
+			}
+			return 1
+		}
+		if a.Name < b.Name {
+			return -1
+		}
+		if a.Name > b.Name {
+			return 1
+		}
+		return 0
```
mariomac: I'd simplify this as:
```golang
if c := strings.Compare(a.Kind, b.Kind); c != 0 {
  return c
}
if c := strings.Compare(a.Namespace, b.Namespace); c != 0 {
  return c
}
return strings.Compare(a.Name, b.Name)
```

mariomac (commented):
Looks ok! I have two minor comments.

pkg/instrumenter/opts.go
```diff
@@ -13,18 +13,19 @@ import (
...
-// Service name and resource attributes set via AddPID or SetPID are shared across all signals.
-// SetPID updates live FileInfo for instrumented PIDs; traces and app metrics read it at export time.
-// Network and stats metrics decorate flow records with service name/namespace by pod IP.
-func WithDynamicPIDSelector(sel *discover.DynamicPIDSelector) Option {
+// Root AddPIDs/RemovePIDs/AddK8sWorkload preserve the legacy behavior and apply to all supported
+// signals. Service name and resource attributes set via AddPID, AddK8sWorkload, or SetPID are
+// shared across all signals. SetPID updates live FileInfo for instrumented PIDs; traces and app
+// metrics read it at export time. Network and stats metrics decorate flow records with service
+// name/namespace by pod IP.
+func WithDynamicSelector(sel *discover.DynamicSelector) Option {
```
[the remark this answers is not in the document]
damemi: this is intentional, the dynamicpidselector api had no stability guarantees and we do not need to support a shim. we can document in release notes as sufficient

pkg/instrumenter/opts.go
```diff
@@ -13,18 +13,19 @@ import (
...
 	return func(info *global.ContextInfo) {
-		info.DynamicPIDSelector = sel
+		info.DynamicSelector = sel
 	}
 }
```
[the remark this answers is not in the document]
damemi: already acknowledged this is an intentional breaking change

pkg/kube/store.go
```diff
@@ -646,6 +646,67 @@ func ownerID(namespace, name string) string {
...
+			if _, ok := wanted[key]; !ok {
+				continue
+			}
+			for _, ip := range meta.Ips {
+				addIP(key, ip)
+			}
+		}
+	}
+	return out
+}
```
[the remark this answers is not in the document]
damemi: I'm not sure about rewriting the main store but open to other's opinions on this

damemi:
@mariomac thanks for the review, I got to your comments and did a couple more rounds of copilot adversarial reviews too if you have a chance

mariomac (approved):
Looks great!

pkg/kube/store.go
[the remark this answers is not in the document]
grcevski: I think it can be a follow-up if you'd like to reengineer that part. If this is a concern with large clusters, perhaps we can do some sort of a debounce, run this if there are new events every couple of seconds?

pkg/selection/dynamic_app_ips.go
```diff
@@ -75,6 +165,43 @@ func (d *DynamicAppIPs) refreshAll() {
...
+	next := map[K8sWorkloadRef][]string{}
+	for _, ref := range refs {
+		ips := d.store.PodIPsForWorkload(ref.Namespace, ref.Kind, ref.Name)
+		if len(ips) == 0 {
+			selLog().Debug("no IPs resolved for dynamically selected workload",
+				"kind", ref.Kind, "namespace", ref.Namespace, "name", ref.Name)
+			continue
+		}
+		next[ref] = ips
+	}
```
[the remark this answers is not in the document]
grcevski: I think this might be the bigger problem with cost to do this. I think the suggestion is valid. Perhaps it's not hard to do this indexing in the store. I'm not sure how much work it is. It can be a follow-up.

pkg/appolly/discover/dynamic_k8s_workload.go
```diff
@@ -0,0 +1,271 @@
...
+	key, attrs, signals, ok := v.parent.findMatchingWorkload(v.mask, meta)
+	if !ok {
+		return
+	}
+	opts := selection.DynamicOptions{
+		ServiceName:        attrs.serviceName,
+		ServiceNamespace:   attrs.serviceNamespace,
+		ResourceAttributes: attrs.resourceAttributes,
+	}
+	v.parent.addSignalsFrom(signals, &opts, &key, uint32(pid))
```
grcevski, line 238: I think we need to clean-up these PIDs when the process terminates. I think at the moment they only the "process history" is cleaned up, but not the workload sources.

pkg/appolly/discover/matcher_dynamic.go
```diff
@@ -136,11 +140,14 @@ func (m *DynamicMatcher) filterCreated(obj ProcessAttrs) (Event[ProcessMatch], b
 }
 
 func (m *DynamicMatcher) matchDynamicCriteria(obj ProcessAttrs, proc *services.ProcessInfo) *ProcessMatch {
-	if !m.DynamicPIDSelector.IncludesPID(proc.Pid) {
+	// Always attempt materialization so a workload source is attached even when the PID was
+	// already selected explicitly (AddPID). Otherwise RemoveK8sWorkload would not account for
+	// that source, and workload opts would never apply to an already-selected PID.
+	m.DynamicSelector.materializeMatchingWorkloads(proc.Pid, obj.metadata)
```
grcevski, line 146: I don't know if this a valid scenario, but if we had the original PID selector first, say you instrumented a process based on dynamic PID (123). Then you push a k8s selection criteria that involves that same PID(123), I think `filterCreated` remembers that PID(123) is instrumented and the workload options will never be updated.

grcevski: Maybe you need something like:

```
func (m *DynamicMatcher) filterCreated(obj ProcessAttrs) (Event[ProcessMatch], bool) {
    if _, ok := m.ProcessHistory[obj.pid]; ok {
        m.DynamicSelector.materializeMatchingWorkloads(obj.pid, obj.metadata)
        return Event[ProcessMatch]{}, false
    }

    ....
}
```

pkg/selection/dynamic_app_ips.go
```diff
@@ -75,6 +162,48 @@ func (d *DynamicAppIPs) refreshAll() {
...
+		d.workloadToIPs = map[K8sWorkloadRef][]string{}
+		d.mu.Unlock()
+		return
+	}
+	refs := ws.GetK8sWorkloads()
+	owners := make([]kube.WorkloadOwner, len(refs))
+	for i, ref := range refs {
+		owners[i] = kube.WorkloadOwner{Namespace: ref.Namespace, Kind: ref.Kind, Name: ref.Name}
+	}
+	ipsByOwner := d.store.PodIPsForWorkloads(owners)
```
grcevski, line 184: There might be a race condition with the app olly side and adding the PIDs unconditionally based on the k8s IP. I understand this is done for the net olly/stats side and it's how it should be, but a PID can get added to the app olly side before we discover it. materializeMatchingWorkloads will fix it later, but I think we have a small window where we might have split service personality. I think it might be fine, the metrics processor will delete the stale metadata and update it, it's just temporarily maybe an issue. Perhaps I'm not fully understanding the flow.

PR #3482, opened 2026-09-18
feat(gpu): add cuda memory, stream, event and synchronization metrics

NameHaibinZhang:
## Summary

Extend OBI's CUDA instrumentation with memory, stream, event, and synchronization primitives, and instrument the CUDA Driver API in addition to the runtime API.

**New runtime API metrics** (all `gpu.cuda.*`, development stability):

- Memory: `gpu.cuda.memory.free.bytes` (cudaFree), `gpu.cuda.memset.bytes` (cudaMemset), `gpu.cuda.host.register.bytes` (cudaHostRegister)
- Streams: `gpu.cuda.stream.create.calls`, `gpu.cuda.stream.destroy.calls`, `gpu.cuda.stream.synchronize.calls`
- Events: `gpu.cuda.event.record.calls`, `gpu.cuda.event.synchronize.calls`
- Device: `gpu.cuda.device.synchronize.calls`

## Validation


grcevski:
Looks like there are some failures in the test.

marctc (approved):
LGTM!

pkg/ebpf/instrumenter.go
```diff
@@ -478,6 +478,12 @@ func (i *instrumenter) uprobeModules(p Tracer, pid app.PID, maps []*procfs.ProcM
 			continue
 		}
 
+		if conflicting, ok := conflictingUprobeLibrary(lib, maps); ok {
```
MrAlias: These uprobes are attached by inode with `PID == 0`, so this per-process maps check cannot keep the driver hooks out of runtime processes. Once any driver-only process attaches the shared `libcuda.so`, that link also fires for every selected process using the same inode; a `libcudart` process then reports the same launch through both `cudaLaunchKernel` and `cuLaunchKernel`. If we never attach the driver link, mixed-API processes lose their direct `cu*` calls instead. Could we move this deduplication into the event/filtering path and cover shared-library plus mixed-API cases?

bpf/gpuevent/cuda.c
```diff
@@ -21,17 +21,98 @@
...
+    k_event_stream_synchronize = 11,
+    k_event_device_synchronize = 12,
+    k_event_host_register = 13,
 };
 
+// Tracks cudaMalloc arguments from entry to return so the allocated pointer can
+// be correlated with its size for byte-accurate cudaFree metrics.
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __uint(max_entries, 1024);
```
MrAlias, line 58: `cuda_malloc_ctx` retains an entry when a task exits inside `cudaMalloc`; the return path also checks `valid_pid` before deleting, so deselection during the call leaves the slot occupied until that `pid_tgid` is reused or the tracer exits. This is a non-evicting hash declared with 1,024 entries, and insert failures are ignored, so enough abandoned calls make later successful allocations invisible to `cudaFree` tracking. Could we make this transient state evictable and guarantee cleanup before the return-side PID filter, with coverage for an outstanding malloc when a PID is blocked or exits?

bpf/gpuevent/cuda.c
```diff
@@ -21,17 +21,98 @@
...
+    __type(key, u64);
+    __type(value, cuda_malloc_ctx_t);
+} cuda_malloc_ctx SEC(".maps");
+
+// Maps an allocated device pointer to its size. Populated by the cudaMalloc
+// uretprobe and consumed by the cudaFree uprobe.
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __uint(max_entries, 65536);
```
MrAlias: `cuda_alloc_sizes` is a fixed non-LRU map shared by every selected process, but its key is only the process-local device pointer. Two processes can reuse the same numeric address and corrupt each other's sizes; allocations also survive process exit because the only deletion is `cudaFree`, so repeated short-lived jobs can fill these 65,536 slots. Update failures are ignored, which makes free-byte telemetry silently disappear once the map is full. Could we use a process-scoped key and either purge on process exit or use bounded eviction, with coverage for same-address processes and exit with outstanding allocations?

bpf/gpuevent/cuda.c
```diff
@@ -78,17 +198,74 @@ int BPF_KPROBE_GUARDED(obi_cuda_malloc, void **devPtr, size_t size) {
...
+
+    const u64 ptr_val = (u64)devPtr;
+    s64 *size = bpf_map_lookup_elem(&cuda_alloc_sizes, &ptr_val);
+    if (!size) {
+        // Pointer was not allocated through cudaMalloc (or tracking was lost);
+        // do not emit a byte metric for it.
+        return 0;
+    }
+
+    submit_size_event(k_event_free, *size);
```
MrAlias: `cudaFree` is fallible, but this emits "bytes released" and deletes the allocation before its return code is known. If the free fails, the metric reports memory as released and a later successful retry cannot be measured. Could we carry the pointer and size to a return probe and only emit/delete on success, mirroring the `cudaMalloc` path?

site/docs/README.md
```diff
@@ -6,7 +6,7 @@
 Generated from the OBI semantic-convention registry in `schemas/obi/`.
 
 - [Attributes](attributes.md) — 21 attribute groups
-- [Metrics](metrics.md) — 87 metrics
+- [Metrics](metrics.md) — 95 metrics
```
MrAlias: This head declares nine new metrics: the registry contains 96 metric groups and `site/docs/metrics.md` renders 96 headings, but this generated index advertises 95. `make generate-schema-docs` renders the index and metrics page from the same resolved registry, so these outputs are not from one consistent generation. Could we regenerate and commit the matching reference?

MrAlias (commented):
The CUDA surface is wired through the exporter and schema layers, but the driver/runtime deduplication and allocation-tracking lifecycle do not preserve per-process correctness. The generated telemetry reference is also internally inconsistent.

pkg/ebpf/instrumenter.go
NameHaibinZhang: Done in 786c1b0 — the Go-side gate is removed and deduplication now lives in the eBPF programs. Runtime uprobes (`cudaLaunchKernel`, `cudaGraphLaunch`) set a per-thread in-flight marker keyed by `pid_tgid`; the driver probes (`cuLaunchKernel`, `cuLaunchKernelEx`, `cuGraphLaunch`) consume the marker and skip when a runtime call is still executing on the same thread, so shared-inode attachment can no longer double count.

Mixed-API processes keep their direct `cu*` calls: a driver call with no marker in flight is reported normally. The runtime uretprobes clear the marker unconditionally, so a runtime call that fails before reaching `libcuda` doesn't suppress the thread's next direct driver call.

bpf/gpuevent/cuda.c
NameHaibinZhang: Done in 786c1b0. `cuda_malloc_ctx` (and the new `cuda_free_ctx` it mirrors) is now `BPF_MAP_TYPE_LRU_HASH`, so abandoned entries are evictable instead of occupying slots until `pid_tgid` reuse.

Both uretprobes consume and delete the entry context unconditionally, before the `valid_pid` check, so deselection or exit during the call always frees the slot — cleanup no longer depends on passing the filter. The in-flight window is one entry per calling thread, so the 1,024-slot LRU bounds the map even for targets that disappear mid-call.

bpf/gpuevent/cuda.c
NameHaibinZhang: Done in 786c1b0. `cuda_alloc_sizes` is now keyed by `(tgid, device pointer)` — `cuda_alloc_key_t` — so processes reusing the same numeric address can't corrupt each other's sizes, and any thread of a process can free memory another thread allocated. The map is `BPF_MAP_TYPE_LRU_HASH` with 65,536 entries, so allocations that survive a process exit age out by eviction instead of permanently consuming slots, and insert failures are now logged with `bpf_dbg_printk` instead of being silently ignored.

bpf/gpuevent/cuda.c
NameHaibinZhang: Done in 786c1b0. `cudaFree` entry stores `(pointer, size)` in a per-thread context and emits nothing; the new uretprobe consumes the context unconditionally and only when the return code is 0 emits the freed-bytes metric and deletes the allocation entry. A failed free keeps the allocation tracked, so a later successful retry is still measured — mirroring the `cudaMalloc` path.

site/docs/README.md
NameHaibinZhang: Done — the reference was regenerated and committed in e097b7f1c; the index now advertises 96 metrics, matching the registry and `metrics.md`.

marctc:
Idea for follow-up: add some info in devdocs about gpu monitoring

bpf/gpuevent/cuda.c
```diff
@@ -19,19 +19,329 @@
...
+    __type(key, u64);
+    __type(value, cuda_set_device_ctx_t);
+} cuda_set_device_ctx SEC(".maps");
+
+// Fills in the device an observed CUDA call ran on: the process-local index the
+// calling thread is bound to, and the UUID that index maps to. An all-zero UUID
+// means the index identity has not been observed for the process yet, which
+// userspace reports as an unknown device.
+static __always_inline void resolve_device(const u64 id, cuda_device_t *dev) {
+    dev->index = 0;
```
MrAlias, line 152: An absent `cuda_thread_device` entry does not prove that the current device is 0. OBI can attach after a thread has already called `cudaSetDevice(1)`, and Driver API applications bind contexts through `cuCtxSetCurrent` or `cuCtxPushCurrent`, which this tracer does not observe. In both cases, subsequent launches are emitted with `cuda.device.index=0` and may pick up device 0's UUID even though they ran elsewhere.

My preferred fix is to carry an explicit `device_known` bit in `cuda_device_t`, set it only when the thread-device lookup succeeds, propagate it to the span, and omit the index, UUID, and model attributes when it is false. Driver API events should remain unknown unless we also track the relevant context creation/set/push/pop calls and map the current context to its `CUdevice`. Could we add coverage for an absent binding and a nonzero Driver API context? The current integration target checks only that driver metrics exist.

NameHaibinZhang: - Added an explicit `known` flag to `cuda_device_t`; `resolve_device()` sets it only when `bpf_map_lookup_elem(&cuda_thread_device, ...)` succeeds and skips the UUID lookup otherwise.
- Propagated the flag to `request.Span.CudaDeviceKnown` and gated `cuda.device.index/uuid/model` getters on it, so unknown devices omit those labels entirely.
- Added integration coverage for:
  - `cudaSetDevice(1)` + `cudaGetDeviceProperties`: metrics carry `cuda_device_index=1`, UUID `00000000-0000-0000-0000-000000000001`, and model `OBI Test GPU A`.
  - Plain Driver API launches: labels are omitted.
  - Driver API launches after `cuCtxCreate(2)` / `cuCtxPushCurrent`: labels are still omitted, since OBI does not instrument those context calls.

bpf/gpuevent/cuda.c
```diff
@@ -78,17 +457,134 @@ int BPF_KPROBE_GUARDED(obi_cuda_malloc, void **devPtr, size_t size) {
...
+        return 0;
+    }
+
+    bpf_dbg_printk("=== uretprobe/cudaFree id=%llx ret=%d ===", id, ret);
+
+    if (ret != 0 || !tracked) {
+        // The free failed and the allocation is still live, or the pointer was
+        // not tracked; keep the size entry so a later successful free can
+        // report it.
+        return 0;
```
grcevski, line 584: can we jump to 586 to delete the cuda_alloc_sizes map?

bpf/gpuevent/cuda.c
```diff
@@ -132,15 +629,576 @@ int BPF_KPROBE_GUARDED(obi_graph_launch) {
...
+    bpf_map_delete_elem(&cuda_set_device_ctx, &id);
+
+    if (!valid_pid(id)) {
+        return 0;
+    }
+
+    bpf_dbg_printk("=== uretprobe/cudaSetDevice id=%llx ret=%d ===", id, ret);
+
+    if (ret != 0 || !tracked) {
+        return 0;
```
grcevski, line 926: should we clean up cuda_thread_device here?

bpf/gpuevent/cuda.c
```diff
@@ -132,15 +629,576 @@ int BPF_KPROBE_GUARDED(obi_graph_launch) {
...
+SEC("uretprobe/cudaGetDevice")
+int BPF_KRETPROBE_GUARDED(obi_cuda_get_device_ret, int ret) {
+    (void)ctx;
+    const u64 id = bpf_get_current_pid_tgid();
+
+    cuda_introspect_ctx_t intro = {};
+    const bool captured = consume_introspect(id, k_cuda_introspect_get, &intro);
+
+    if (!valid_pid(id)) {
+        return 0;
```
grcevski, line 969: Do we need to clean up bpf_map_delete_elem(&cuda_thread_device, &id); on all early returns?

bpf/gpuevent/cuda.c
```diff
@@ -19,19 +19,335 @@
...
+// Which parts of a device identity an introspection return scan managed to read.
+enum {
+    k_device_has_uuid = 1,
+    k_device_has_name = 2,
+};
+
+// Tracks cudaMalloc arguments from entry to return so the allocated pointer can
+// be correlated with its size for byte-accurate cudaFree metrics. LRU eviction
+// bounds the map if a target exits or is deselected while a call is in flight.
+struct {
```
grcevski, line 57: Is it possible to make some of these maps not LRU? Since these are uprobes, we don't have to worry about the KRETPROBE not firing sometimes and regular hashmaps are a bit faster than LRU maps.

grcevski (commented):
Looks great, I had a few comments about making the eBPF side a bit more efficient if it's possible.

bpf/gpuevent/cuda.c
NameHaibinZhang: Not on the `ret != 0` path: a failed `cudaFree` means the allocation is still live, so we keep the `cuda_alloc_sizes` entry so a later successful free can still report the bytes. On the `!tracked` path we don't have the pointer here (the entry was never created because the allocation wasn't tracked), so there's nothing to delete.

bpf/gpuevent/cuda.c
NameHaibinZhang: No — a failed `cudaSetDevice` doesn't clear the thread's current binding. The device it was already bound to remains valid, so deleting `cuda_thread_device` would incorrectly mark subsequent calls as unknown.

bpf/gpuevent/cuda.c
NameHaibinZhang: `cudaGetDevice` is just a query; it doesn't change the thread's device binding. Even if the query fails or we can't read the result, any previous binding in `cuda_thread_device` is still correct, so we shouldn't delete it on early returns.

bpf/gpuevent/cuda.c
NameHaibinZhang: Done. Switched the five per-thread in-flight context maps (`cuda_malloc_ctx`, `cuda_free_ctx`, `cuda_runtime_launch_ctx`, `cuda_introspect_ctx`, `cuda_set_device_ctx`) to `BPF_MAP_TYPE_HASH`; their uretprobes consume and delete entries unconditionally, so LRU overhead isn't needed. Kept `cuda_alloc_sizes`, `cuda_thread_device`, and `cuda_device_info` as LRU because they hold longer-lived per-allocation/per-thread/per-process state that can outlive our ability to clean it up.

MrAlias (commented):
The unknown-device event handling is now safer, but the five transient maps are fixed hashes again: task exits can strand entries while the new GPU guide still promises LRU eviction. The advertised CUDA baseline and device-attribute behavior also diverge from this head.

SUPPORT_MATRIX.md
```diff
@@ -238,7 +238,10 @@ OBI currently documents the following GPU execution instrumentation support:
 
 | Library | Baseline | Instrumented primitives | Limitations |
 |:--------|:---------|:------------------------|:------------|
-| `libcuda` | `>= 7.0` | `cudaLaunchKernel`, `cudaGraphLaunch`, `cudaMalloc`, `cudaMemcpy`, `cudaMemcpyAsync` | None documented |
+| `libcudart` | `>= 7.0` | `cudaLaunchKernel`, `cudaGraphLaunch`, `cudaMalloc`, `cudaFree`, `cudaMemset`, `cudaMemcpy`, `cudaMemcpyAsync`, `cudaStreamCreate`, `cudaStreamCreateWithFlags`, `cudaStreamCreateWithPriority`, `cudaStreamDestroy`, `cudaEventRecord`, `cudaEventRecordWithFlags`, `cudaEventSynchronize`, `cudaStreamSynchronize`, `cudaDeviceSynchronize`, `cudaHostRegister` | None documented |
```
MrAlias: This baseline includes CUDA 7–9, but `merge_props` always treats bytes 256–271 of `cudaDeviceProp` as a UUID. NVIDIA added the `uuid` member only in CUDA 10; in 9.2, offset 256 is `totalGlobalMem`. A process that binds that device and queries its properties therefore gets a fabricated UUID, and later metrics are labeled with the wrong physical device. Could we gate that read by runtime version or rely on `cuDeviceGetUuid`, or narrow and document the supported behavior?

schemas/obi/groups/gpu/registry.yaml
```diff
@@ -33,3 +33,28 @@ groups:
         brief: >
           Direction of a CUDA memory copy, mirroring the `cudaMemcpyKind` enum
           of the CUDA Runtime API.
+      - id: cuda.device.index
+        type: int
+        stability: development
+        brief: >
+          Index of the CUDA device the call ran on, as the process sees it. It
+          defaults to 0, CUDA's default device, so it is always known.
```
MrAlias: `CudaDeviceKnown` now intentionally leaves the index unset when no binding was observed, and the new integration tests assert that all `cuda.device.*` labels are empty/omitted for both Driver API targets. This source registry still says the index is always known; `devdocs/gpu-monitoring.md` likewise says every metric is labeled and that the index is always present. Could we update both source docs to match the new behavior and regenerate the reference?

grcevski (approved):
LGTM!

SUPPORT_MATRIX.md
NameHaibinZhang: Good catch. I documented the limitation rather than raising the baseline or adding runtime-version gating:

- `SUPPORT_MATRIX.md` now notes that device UUID and model attribution through `cudaGetDeviceProperties` require CUDA 10+, because `cudaDeviceProp.uuid` was added at offset 256 in that release.
- `bpf/gpuevent/cuda.h` now states that the UUID offset is only valid from CUDA 10.0 onward.

The API hooks themselves still work on CUDA 7.0+, and the device index is emitted whenever a binding is observed; only the UUID/model learned from this offset are affected.

schemas/obi/groups/gpu/registry.yaml
NameHaibinZhang: Updated and regenerated:

- `schemas/obi/groups/gpu/registry.yaml`: `cuda.device.index` brief now says the attribute is present only after `cudaSetDevice`/`cudaGetDevice` and omitted when no binding was observed.
- `devdocs/gpu-monitoring.md`: rewrote the device-attribution section and the metrics section to say labels are carried only when the binding and identity were observed.
- Ran `make generate-schema-docs`; `site/docs/attributes.md` and `site/docs/metrics.md` now match the registry.

bpf/gpuevent/cuda.c
```diff
@@ -19,19 +19,342 @@
...
+} cuda_set_device_ctx SEC(".maps");
+
+// Fills in the device an observed CUDA call ran on: the process-local index the
+// calling thread is bound to, and the UUID that index maps to. An all-zero UUID
+// means the index identity has not been observed for the process yet, which
+// userspace reports as an unknown device.
+static __always_inline void resolve_device(const u64 id, cuda_device_t *dev) {
+    dev->index = 0;
+    dev->known = 0;
+    __builtin_memset(dev->uuid, 0, sizeof(dev->uuid));
```
mmat11: is it possible to use bpf_memset here?

bpf/gpuevent/cuda.c
```diff
@@ -19,19 +19,342 @@
...
+        return;
+    }
+
+    cuda_device_key_t key = {
+        .tgid = id >> 32,
+        .index = dev->index,
+    };
+    const cuda_device_info_t *info = bpf_map_lookup_elem(&cuda_device_info, &key);
+    if (info) {
+        __builtin_memcpy(dev->uuid, info->uuid, sizeof(dev->uuid));
```
mmat11: bpf_memcpy

bpf/gpuevent/cuda.c
```diff
@@ -19,19 +19,342 @@
...
+    if (!e) {
+        bpf_dbg_printk("Failed to allocate ringbuf entry");
+        return;
+    }
+
+    e->flags = k_event_device_info;
+    task_pid(&e->pid_info);
+    e->index = index;
+    __builtin_memcpy(e->uuid, info->uuid, sizeof(e->uuid));
+    __builtin_memcpy(e->name, info->name, sizeof(e->name));
```
mmat11: bpf_memcpy

bpf/gpuevent/cuda.c
```diff
@@ -19,19 +19,342 @@
...
+    cuda_device_event_t *e = bpf_ringbuf_reserve(&gpu_events, sizeof(*e), 0);
+    if (!e) {
+        bpf_dbg_printk("Failed to allocate ringbuf entry");
+        return;
+    }
+
+    e->flags = k_event_device_info;
+    task_pid(&e->pid_info);
+    e->index = index;
+    __builtin_memcpy(e->uuid, info->uuid, sizeof(e->uuid));
```
mmat11: bpf_memcpy

bpf/gpuevent/cuda.c
```diff
@@ -19,19 +19,342 @@
...
+
+    cuda_device_key_t key = {
+        .tgid = id >> 32,
+        .index = index,
+    };
+
+    cuda_device_info_t info = {};
+    const cuda_device_info_t *stored = bpf_map_lookup_elem(&cuda_device_info, &key);
+    if (stored) {
+        __builtin_memcpy(info.uuid, stored->uuid, sizeof(info.uuid));
```
mmat11: same here, all instances where possible

mmat11 (changes requested):
blocker: all the new events are not handled/filtered in `acceptSpan` in tracesgen.go so they leak and get exported as trace spans

--

also we are adding a lot of new probes on hot paths, do we have some overhead numbers?

bpf/gpuevent/cuda.c
NameHaibinZhang: Done in 5c86102d3: replaced with `bpf_memset` and added `#include <bpfcore/bpf_builtins.h>`.

bpf/gpuevent/cuda.c
NameHaibinZhang: Done in 5c86102d3: replaced with `bpf_memcpy`.

bpf/gpuevent/cuda.c
NameHaibinZhang: Done in 5c86102d3: replaced with `bpf_memcpy`.

bpf/gpuevent/cuda.c
NameHaibinZhang: Done in 5c86102d3: replaced with `bpf_memcpy`.

bpf/gpuevent/cuda.c
NameHaibinZhang: Done in 5c86102d3: all remaining `__builtin_memcpy` calls in this file were replaced with `bpf_memcpy`.

NameHaibinZhang:
@mmat11 thanks for the review.

- **acceptSpan blocker**: fixed in 5c86102d3 — all new GPU event types are now explicitly filtered out in `tracesgen.acceptSpan` so they feed metrics only and are not exported as trace spans.
- **bpf_memset / bpf_memcpy**: fixed in the same commit — replaced every `__builtin_memset`/`__builtin_memcpy` in `bpf/gpuevent/cuda.c` with the `bpf_*` helpers and added `#include <bpfcore/bpf_builtins.h>`.
- **Rebase**: the branch has been rebased onto `upstream/main` and the regenerated schema docs / eBPF bindings were committed (d701745be).

Re: overhead numbers for the new hot-path probes — I don't have benchmark results yet. I'll run the GPU workload harness we used for the initial kernel-launch probe and post numbers in a follow-up comment.

mmat11 (approved):
LGTM 👍🏻 , curious to see the overhead numbers

NameHaibinZhang:
@mmat11 I'm running the overhead evaluation on an NVIDIA H20 now. I'll collect the numbers and add a doc (likely under `devdocs/`) summarizing the methodology and results in a follow-up PR/commit.

pkg/internal/ebpf/gpuevent/gpuevent.go
```diff
@@ -302,26 +442,68 @@ func (p *Tracer) readGPUKernelLaunchIntoSpan(record *ringbuf.Record) (request.Sp
...
-			HostPID:   app.PID(event.PidInfo.HostPid),
-			UserPID:   app.PID(event.PidInfo.UserPid),
-			Namespace: event.PidInfo.Ns,
-		},
-	}, false, nil
+	p.log.Debug("GPU device info", "uuid", uuid, "model", model)
+
+	// A device may be learned by UUID and name separately, and each event
+	// carries the merged view, so never drop a model we already know.
+	if uuid != "" && model != "" {
```
MrAlias: When only `cuDeviceGetName` succeeds, `event.Uuid` is zero, so this condition discards the model even though the eBPF cache and event contain a valid name. Subsequent metrics then omit `cuda.device.model`, contrary to the name-only partial-observation behavior documented in `devdocs/gpu-monitoring.md`.

Could we retain name-only identity using the event's process/index fields, or otherwise avoid requiring a UUID before caching the model?

PR #3486, opened 2026-09-18
Preserve FastCGI backend addresses

MrAlias:
## Summary

- preserve the request scheme on FastCGI spans without treating the inbound HTTP host as the FastCGI backend
- remove the now-unused FastCGI host parsing
- cover client span backend address and port selection with a regression test

This prevents client spans from combining the inbound HTTP host with the FastCGI destination port in their endpoint attributes.

## Testing

- `go test ./pkg/ebpf/common`
- `make lint`


grcevski (approved):
LGTM!

PR #3488, opened 2026-09-18
Keep Java trace fixture on the JDK HTTP client

MrAlias:
## Summary

- Configure the shared RestClient used by /jtraceA with JdkClientHttpRequestFactory.
- Preserve the worker-thread execution model that this integration fixture is intended to exercise.

## Motivation

The Spring Boot 3.2.1 to 3.5.11 update changed the default request factory selected by RestClient. Images rebuilt with the newer dependencies stopped creating JDK HttpClient worker tasks, and the ARM64 multiprocess suites consequently lost Java-to-Node trace correlation for /jtraceA.

Selecting the request factory explicitly keeps the fixture behavior stable across Spring dependency updates.

## Validation

- Built the Java test image from internal/test/integration/components/javatestserver/Dockerfile_jar with Spring Boot 3.5.11.
- Exercised /jtraceA against a local HTTP target.
- Verified that the Java process created HttpClient worker threads.


grcevski (approved):
LGTM!

PR #3489, opened 2026-09-18
Support legacy Go HTTP/2 traceparent ownership

MrAlias:
## Summary

- add atomic ownership probe groups for pre-`clientStream` x/net HTTP/2 and older bundled stdlib layouts, while keeping current layouts authoritative when both are present
- carry request identity from pre-HPACK encoding into stream setup so application-owned `traceparent` values remain untouched and unowned requests still receive OBI propagation
- preserve per-stream ownership across persistent and multiplexed connections
- add pinned pre-`clientStream` x/net and Go 1.17 minimum-runtime fixtures

## Compatibility coverage

The ownership matrix verifies:

- current x/net over TLS and plaintext
- pinned `golang.org/x/net` `v0.0.0-20201021035429-f5854403a974` over TLS and plaintext
- Go 1.17.13 bundled stdlib over TLS
- repeated application-owned values on one connection
- unowned requests receiving exactly one OBI-generated value
- concurrent owned and unowned streams sharing one connection without ownership leakage
- atomic rejection of incomplete probe layouts

## Validation

- `make generate`
- `make verify`
- `make build`
- `go test ./pkg/internal/ebpf/gotracer ./pkg/ebpf ./pkg/ebpf/common`
- `go test -v -run '^TestHTTP2Go$' -timeout 15m ./internal/test/integration/`

Relates to #2769.


mariomac (approved):
Looks OK to me!

PR #3490, opened 2026-09-18
Support uprobes without CAP_SYS_ADMIN on kernels > 5.14 and < 6.6

grcevski:
## Summary

After [CVE-2025-38466](https://app.opencve.io/cve/CVE-2025-38466) SYS_ADMIN was made a requirement for all uprobes, which limits our ability to ship OBI for customers which have strict security requirements. Without uprobes support we cannot run much of the application observability side, namely no Go support as well as OpenSSL, and some of the runtime metrics.

On kernels 6.6+ we now use `uprobe_multi` which isn't blocked, except we still have one outstanding probe `obi_capture_go_executable_identity` that uses the legacy `uprobe` mechanism.

However, the good news is that the CVE fix was restricted to the [PMU uprobes](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=183bdb89af1b5193b1d1d9316986053b15ca6fa4), which are now the default for all kernels we support. This means that we can use the "legacy" tracefs style of uprobes, which were added to libbpf long ago through [this commit]( https://lore.kernel.org/bpf/20210921210036.1545557-5-andrii@kernel.org/T/).

This PR adds tracefs support for uprobes, removing the SYS_ADMIN restriction for OBI for virtually all Linux systems. It will require that tracefs (or debugfs) are mounted on the system and the CAP_DAC_OVERRIDE capability must be given (otherwise we cannot change the tracefs read only status), but this is a lot better than SYS_ADMIN.

With this change we virtually never need SYS_ADMIN, unless we need it for use of bpf_probe_write_user.

I had to implement the tracefs support separately, because Cilium Go doesn't have an option to override the attachment mode on uprobe calls. They support tracefs, but if they detect PMU available they just attempt that without trying to fallback to tracefs.

Since tracefs cleanup is manual, e.g the probes don't die automatically when OBI dies like in the PMU case, this PR exposed a bug in accounting of closers related to shared library reuse. I've fixed that here too.

Follow-ups needed:
- We need to update the OBI docs to remove the SYS_ADMIN requirement and write a new guide on how to run OBI without SYS_ADMIN.
- We need to update the helm chart to auto mount the tracefs by default.

## Validation


pkg/internal/ebpf/uprobe/tracefs.go
```diff
@@ -0,0 +1,323 @@
...
+const traceFSEventName = "probe"
+
+var findTraceFS = sync.OnceValues(func() (string, error) {
+	mounts, err := procfs.GetMounts()
+	if err == nil {
+		if root := writableTraceFSMount(mounts); root != "" {
+			return root, nil
+		}
+		for _, mount := range mounts {
+			if mount.FSType == "debugfs" && mount.Root == "/" && mountIsWritable(mount) {
```
mmat11, line 33: I think cilium/ebpf also checks for "tracefs" here -> https://github.com/cilium/ebpf/blob/5d6e0aca0a7df6ea325a9cb34fb185c489cd360e/internal/tracefs/kprobe.go#L164

mmat11: ah nvm it was in `writableTraceFSMount`

pkg/internal/ebpf/uprobe/tracefs.go
```diff
@@ -0,0 +1,323 @@
...
+func unescapeMountPath(path string) string {
+	return strings.NewReplacer(
+		`\040`, " ",
+		`\011`, "\t",
+		`\012`, "\n",
+		`\134`, `\`,
+	).Replace(path)
+}
+
+func attachTraceFS(path string, prog *ebpf.Program, opts Options) (io.Closer, error) {
```
mmat11, line 88: I think we can avoid all this complexity by using this option: https://github.com/cilium/ebpf/blob/main/link/kprobe.go#L240

`RetprobeMaxActive`

if that option is not zero, link.Uprobe() will fallback to tracefs automatically. We need to find an appropriate value for it, the kernel default is `max(10, 2 * num_possible_cpus())` so we could just replicate it

grcevski: ok, good find. let me try if this actually works.

grcevski: no, man they reject this later:
https://github.com/cilium/ebpf/blob/main/internal/tracefs/kprobe.go#L355

the case for uprobe checks it again.

mmat11: damn, nvm then :/

pkg/internal/ebpf/uprobe/uprobe.go
```diff
@@ -276,6 +315,31 @@ func attachPerfEvents(exe *link.Executable, prog *ebpf.Program, opts Options) (i
...
+	closer, err := attachPerf()
+	if err == nil || !errors.Is(err, unix.EACCES) {
+		return closer, err
+	}
+
+	slog.Debug("failed to use uprobe with PMU, likely no SYS_ADMIN capability provided, trying tracefs attach", "error", err)
+
+	closer, traceFSErr := attachTraceFS()
+	if traceFSErr != nil {
+		slog.Debug("cannot attach tracefs based uprobe, maybe CAP_DAC_OVERRIDE is missing or tracefs/debugfs not mounted", "error", traceFSErr)
```
mmat11: this one should probably be a Warn/Error?

grcevski: Good idea, I can use the same trick to print it once, because otherwise this will log on every uprobe, although maybe we should because it's the final ditch effort.

mmat11 (approved):
lgtm if tests pass

devdocs/deleting-tracefs-uprobes.md
```diff
@@ -0,0 +1,30 @@
...
+
+Without `uprobe_multi` and `CAP_SYS_ADMIN` on a kernel that has [CVE-2025-38466](https://app.opencve.io/cve/CVE-2025-38466)
+patched, we cannot use the PMU for uprobes and we fall back on tracefs. Tracefs needs to be explicitly cleaned up and we
+have proper closer path for it. However, in case OBI crashes (especially while development), you may end up with a number of
+stale/ghost `tracefs` probes that you may want to manually clean up.
+
+Here's how you can check if you have any stale leftover probes:
+
+```sh
+sudo grep 'obi_' /sys/kernel/debug/tracing/uprobe_events
```
MrAlias, line 11: The cleanup commands hard-code `/sys/kernel/debug/tracing`, but OBI may use a writable tracefs mount at `/sys/kernel/tracing`—the path exposed by the new compose test—or at another location discovered by `findTraceFS`. When the debugfs compatibility path is unavailable, these commands cannot find or remove OBI's stale events. Could we resolve the active tracefs/debugfs mount once and reuse it for both the inspection and cleanup commands?

MrAlias (commented):
The tracefs fallback and lifecycle changes look great.

nit (non-blocking): The recovery guide still needs to handle the mount locations supported by the implementation.

MrAlias (approved):

devdocs/deleting-tracefs-uprobes.md
grcevski: This is actually just for the developers (like us) when we are debugging issues, the actual code does the cleanup correctly. For all development machines we'll get the same path, unless something changes in the Linux kernel. We prefer tracefs first and only then look for debugfs, so the path above will be the same for most dev boxes.

PR #3491, opened 2026-09-18
discovery: hold pod processes until their k8s metadata is known

### LINKED ISSUE #3483: k8s_namespace exclusions don't apply to processes discovered before their Pod metadata
With k8s metadata enabled, a process found before its Pod is in the store is forwarded to the matcher without metadata. `k8s_namespace` exclusions can't match a process with no namespace, and once a PID is matched the matcher never looks at it again, so the later event carrying the metadata is ignored.

Result: with `instrument: [{exe_path: "*"}]`, kube-system workloads (CoreDNS, kube-proxy, ...) get instrumented whenever the informer sync is slow or times out, even though `default_exclude_instrument` excludes that namespace.

There are two ways to fix it:

1. Hold containerized processes until their Pod is known. Fail-safe, but a process whose Pod never shows up is never instrumented, where today it is instrumented with degraded metadata.
2. Keep forwarding as today, re-check exclusions when the metadata arrives and detach if excluded. No behaviour change for anyone else, but excluded processes stay instrumented until the Pod metadata lands.

mmat11:
## Summary

Fixes https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/3483

With k8s metadata enabled, a process discovered before its Pod was in the store was forwarded to the matcher without metadata. Exclusions based on `k8s_namespace` could not match a process that has no namespace, and the matcher skips PIDs it has already matched, so the later re-emission with metadata changed nothing.
Any selector that matches on local criteria alone, such as `exe_path: "*"`, instrumented kube-system workloads (CoreDNS, kube-proxy, …) during startup or after an `informers_sync_timeout`. The default exclusion of kube-system silently did not apply to them.

The enricher now holds a containerized process until its Pod is known and forwards it from the Pod event, always with metadata. Processes that are not in a container are unaffected.

### Behaviour change

A container whose Pod never becomes known (k8s-cache unreachable past the sync timeout, or a container not managed by the kubelet on a node where Kubernetes metadata is enabled) is no longer instrumented, where before it was instrumented with degraded metadata. Selectors such as `k8s_namespace: '*'` already behaved this way; this makes path- and port-based selectors consistent with them.

## Validation


grcevski:
Maybe this fix can be done in matcher? I'm thinking about the following flow:

1. If you have any k8s selectors in the criteria, you only match if the process coming through the pipeline has any k8s metadata.
2. Do the above only if k8s is enabled.

mmat11:
> Maybe this fix can be done in matcher? I'm thinking about the following flow:
> 
>     1. If you have any k8s selectors in the criteria, you only match if the process coming through the pipeline has any k8s metadata.
> 
>     2. Do the above only if k8s is enabled.

I looked at doing it in the matcher: it can't tell a host process from a container whose Pod isn't known yet (both just have empty metadata), and the default exclusion always carries a k8s_namespace selector, so deferring there would stop instrumenting every host process; the enricher is the one place that knows the container and its Pod

MrAlias (dismissed):

pkg/appolly/discover/watcher_kube.go
```diff
@@ -186,13 +226,62 @@ func (wk *watcherKubeEnricher) onNewProcess(procInfo ProcessAttrs) (ProcessAttrs
...
+	for containerID, held := range wk.heldContainers {
+		if held.warned || now.Sub(held.since) < wk.holdWarnAfter {
+			continue
+		}
+		held.warned = true
+		pids := make([]app.PID, 0, len(wk.processByContainer[containerID]))
+		for _, proc := range wk.processByContainer[containerID] {
+			pids = append(pids, proc.pid)
+		}
+		wk.log.Warn("no Pod metadata received for container. Its processes are not instrumented until the Pod is known. "+
```
grcevski: Does this warn trigger for any processes we normally see in Kubernetes? Perhaps not a big deal, but I'm just wondering if people will now be surprised.

grcevski (approved):
LGTM!

grcevski:
> 
> I looked at doing it in the matcher: it can't tell a host process from a container whose Pod isn't known yet (both just have empty metadata), and the default exclusion always carries a k8s_namespace selector, so deferring there would stop instrumenting every host process; the enricher is the one place that knows the container and its Pod

Ah yeah, good point.

NimrodAvni78 (approved):

PR #3494, opened 2026-09-18
fix(java agent): Fix issues with OBI java agent discovered with running Wildfly

grcevski:
## Summary

@GiacomoPetacchi kindly created a [reproducer testcase]( 
https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1257#issuecomment-5683642273) for the issues he was experiencing with running OBI with Wildfly. I discovered 3 separate issues, made fixes for all 3 and I've created an OATS testcase using his original repo.

The issues we found with Wildfly:

1. It seems that the JBoss logger wants to be the boss of logging :) (no pun intended :)). It refuses to load if there's another logger created before it, and we were using the JDK logger (instead of System.err.println()) in the main Agent class. This wasn't seen by Giacomo, because I experienced it while I put the OBI jar directly (without dynamic attach) on the java command line. The fix was to remove the logger usage, which we've done in all cases except this main Agent class.
2. The main issue was that the JBoss class loader is [non delegating](https://docs.wildfly.org/35/Developer_Guide.html#class-loading-in-wildfly), so it doesn't delegate to the bootclass loader and it refused to load our helper classes. I used the [same trick the OTel Java agent uses](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/internal/internal-class-loader/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/classloader/LoadInjectedClassInstrumentation.java) to force it to load our classes. Essentially, I instrumented the class loading itself and helped it forcefully delegate to our loaded helper classes. 
3. Finally, the AWS SDK used double wrapped SSL connection, so it appeared as if we are getting a request retry. 
```
AWS DelegateSslSocket
  -> OBI ProxyOutputStream      -> first ioctl
       -> JDK AppOutputStream   -> second ioctl
            -> TLS encryption   -> one network write
```
The fix here was to check the class type in the proxy streams and avoid double SSL instrumentation if it's the already instrumented SSL stream.

Relates to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/1257

## Validation


MrAlias (approved):

mariomac (approved):
:shipit:

PR #3498, opened 2026-09-18
fix fallback cgi http.scheme

NimrodAvni78:
missed in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3463

## Summary

A FastCGI server span can carry no `url.scheme`. `cgiScheme` reads `HTTP_X_FORWARDED_PROTO`, then `REQUEST_SCHEME`, then `HTTPS`, and reports nothing when the front end set none of them — so the attribute is absent on those spans.

`url.scheme` is `required` on HTTP server spans in semconv, and its definition says what to do when the client's scheme is unknown: "The scheme of the original client request, if known (e.g. from `Forwarded#proto`, `X-Forwarded-Proto`, or a similar header). Otherwise, the scheme of the immediate peer request."

## What changes

`cgiScheme` falls back to `http` — the scheme of the immediate peer request, which for FastCGI is the plain connection into PHP-FPM. The forwarded-header and `REQUEST_SCHEME`/`HTTPS` preferences are unchanged and still win, so a TLS-terminated site that tells OBI its scheme still reports `https`.

This is the fallback semconv prescribes rather than a guess: when nothing names the client's scheme, the immediate hop's scheme is the defined answer.

## Validation


grcevski (approved):
LGTM!

mmat11 (approved):
needs rebase

pkg/ebpf/common/fast_cgi_detect_transform.go
```diff
@@ -176,7 +176,7 @@ func cgiScheme(kv map[string]string) string {
 		return "https"
 	}
 
-	return ""
+	return "http"
```
MrAlias, line 190: With the default `ebpf.buffer_sizes.tcp=0`, FastCGI parsing sees at most the 256-byte `k_tcp_max_len` buffer. `parseHeader` returns whatever remains of the first PARAMS record without checking `ContentLength`, and the existing `Not enough data` case shows that we still emit a span from this truncated input. If `REQUEST_SCHEME=https` or `HTTP_X_FORWARDED_PROTO=https` appears after the cutoff, this branch now labels the request `http` even though the front end provided the original scheme.

Could we only apply this fallback after proving that the complete PARAMS stream contains no scheme metadata? Otherwise, leaving it unset is more accurate than emitting the wrong scheme.

NimrodAvni78: fixed.

MrAlias: 92e462d fixes truncation inside the first `FCGI_PARAMS` record, but it still does not prove or parse the complete PARAMS stream. The [FastCGI specification](https://fastcgi-archives.github.io/FastCGI_Specification.html) defines a stream as a sequence of non-empty records terminated by an empty record, and `FCGI_PARAMS` is a stream type. Here, `paramsComplete` becomes true as soon as the captured bytes cover only the first `PARAMS` record’s `ContentLength`, while `parseHeader` returns subsequent record headers and padding as though they were `PARAMS` content.

A request with `REQUEST_METHOD` and `REQUEST_URI` in the first record and `REQUEST_SCHEME=https` in a later record still emits http, both when the later record is truncated and when it is fully captured with the empty terminator. We need to concatenate the `contentData` from every same-request `PARAMS` record, parse that reconstructed byte stream, and observe the zero-length terminator before applying the fallback.

MrAlias: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3534

PR #3507, opened 2026-09-21
rank OBI's own OTEL_RESOURCE_ATTRIBUTES below resolved target metadata

### LINKED ISSUE #3432: OBI's own env overrides the Kubernetes metadata it resolved for instrumented targets
### Component

OBI (eBPF instrumentation)

### What happened?

When an instrumented process has no `OTEL_RESOURCE_ATTRIBUTES` of its own, OBI falls back to the environment of its own process. That fallback value is placed **above the per-workload Kubernetes metadata that OBI itself resolved**, so the agent's identity ends up in the instrumented target's resource.

The OBI DaemonSet was given the env below to apply deployment-wide attributes. The instrumented targets declare no `OTEL_RESOURCE_ATTRIBUTES`.

```
OTEL_RESOURCE_ATTRIBUTES=k8s.node.name=$(NODE),test.marker=obi-env-probe,k8s.pod.name=OBI-COLLECTOR-MARKER
```

`k8s.pod.name` is a marker included to reveal what this fallback overwrites; `test.marker` is a control to show what happens to a key that does not collide.

The result is a resource whose values contradict each other. The official OpenTelemetry Demo was deployed and instrumented; below is the span resource of one of its pods (`frontend`), observed on OBI v0.12.2:

| attribute | value | |
|---|---|---|
| `k8s.pod.uid` | `23f4b349-…` | the frontend pod's real uid |
| `k8s.replicaset.name` | `frontend-5d76d69658` | the frontend pod's real replicaset |
| **`k8s.pod.name`** | **`OBI-COLLECTOR-MARKER`** | the agent's env value |
| `service.instance.id` | `obi.frontend-5d76d69658-fhp4h.frontend` | the real pod name |
| `test.marker` | `obi-env-probe` | a non-colliding key also lands on the workload |

`service.instance.id` carries the real pod name, yet `k8s.pod.name` alone holds the agent's env value. OBI knew the real pod and overwrote it anyway.

There is nothing special about this workload. The fallback has no per-workload condition — it only checks whether the target declares the variable — so any instrumented target without its own `OTEL_RESOURCE_ATTRIBUTES` produces the same result.

This is only observable when keys collide. If the agent env carries only keys that Kubernetes metadata does not provide, nothing looks wrong.

`k8s.pod.name` should be the instrumented target's actual pod. It should be consistent with `service.instance.id` in the same resource.

Generalized, the precedence should be:

1. The instrumented target's own `OTEL_RESOURCE_ATTRIBUTES` — unchanged
2. Annotations / labels
3. The Kubernetes metadata OBI resolved
4. **OBI's own env — deployment-wide defaults, filling only keys nobody else declared**

Today number 4 occupies the position of number 1. The precedence list documented in the comments of `pkg/kube/store.go` does not mention the agent env at all.

### Steps to reproduce

1. Add a pod-level key to `OTEL_RESOURCE_ATTRIBUTES` of an OBI DaemonSet. See the configuration below.
2. Run a workload that declares no `OTEL_RESOURCE_ATTRIBUTES` and generate traffic against it. The official OpenTelemetry Demo works as-is.
3. Pick any workload and inspect the resource attributes of its spans.
4. Observe that `k8s.pod.name` holds the agent's env value while `service.instance.id` in the same resource holds the real pod name.

### OBI configuration

```yaml
# env of the OBI DaemonSet
env:
  - name: NODE
    valueFrom:
      fieldRef:
        fieldPath: spec.nodeName
  - name: OTEL_RESOURCE_ATTRIBUTES
    value: k8s.node.name=$(NODE),test.marker=obi-env-probe,k8s.pod.name=OBI-COLLECTOR-MARKER

# collector config (receiver mode)
receivers:
  obi:
    attributes:
      kubernetes:
        enable: true
    metrics:
      features: [application]

exporters:
  debug:
    verbosity: detailed

service:
  telemetry:
    logs:
      level: info
  pipelines:
    traces:
      receivers: [obi]
      exporters: [debug]
```

### Relevant logs

```shell
ResourceSpans #0
Resource SchemaURL: 
Resource attributes:
     -> service.name: Str(frontend)
     -> telemetry.sdk.language: Str(nodejs)
     -> telemetry.sdk.name: Str(opentelemetry)
     -> telemetry.sdk.version: Str(v1.46.0)
     -> telemetry.distro.name: Str(opentelemetry-ebpf-instrumentation)
     -> telemetry.distro.version: Str(unset)
     -> service.namespace: Str(otel-demo)
     -> k8s.pod.uid: Str(23f4b349-...)
     -> k8s.replicaset.name: Str(frontend-5d76d69658)
     -> k8s.node.name: Str(node-1)
     -> k8s.owner.name: Str(frontend)
     -> k8s.pod.name: Str(OBI-COLLECTOR-MARKER)
     -> k8s.cluster.name: Str(test-cluster)
     -> k8s.namespace.name: Str(obi)
     -> k8s.kind: Str(Deployment)
     -> k8s.deployment.name: Str(frontend)
     -> service.version: Str(2.2.0)
     -> k8s.container.name: Str(frontend)
     -> service.instance.id: Str(obi.frontend-5d76d69658-fhp4h.frontend)
     -> k8s.node.ip: Str(10.0.0.1)
     -> test.marker: Str(obi-env-probe)
     -> span.metrics.skip: Bool(true)
     -> otel.scope.name: Str(go.opentelemetry.io/obi)
```

#### Cause

`parseOTELEnvVar` returns two values of very different nature without distinguishing them — `pkg/export/otel/otelcfg/common.go`

```go
if svc != nil && svc.EnvVars != nil {
    envVar, ok = svc.EnvVars[varName]   // (1) the instrumented target's env
}

if !ok {
    envVar, ok = os.LookupEnv(varName)  // (2) OBI's own env
}
```

`ResourceAttrsFromEnv` returns both as a single slice, and callers append it after the app resource attributes. Since OTel Go applies last-value-wins to duplicate keys, the env-derived value wins. The caller cannot tell whether a value came from (1) or (2).

Five merge sites: `pkg/export/otel/metrics.go` (two), `pkg/export/otel/metrics_svc_graph.go`, `pkg/export/otel/metrics_runtime.go`, `pkg/export/otel/traces.go` → `pkg/export/otel/tracesgen/tracesgen.go`

This path is identical in the v0.12.2 I measured and in current main `872e6f57` — neither the fallback nor the merge order has changed.

----

#### Why this is a precedence problem

**There is nothing wrong with (2) existing. The problem is where it ranks.**

(2) was introduced deliberately, to satisfy a request for deployment-wide common attributes. The reporter of grafana/beyla#1053 set `deployment.environment` in the agent's env in order to attach it to **the instrumented application's** telemetry.

> I am trying to add some Resource Attributes to the Metrics / Traces Exported by Beyla for an Application its Auto Instrumenting (...) I want to include the `deployment.environment` Resource Attribute in all Traces.

The PR that resolved it, grafana/beyla#1085, was not a new feature but a **regression fix**.

> After we consolidated the trace receiver with the trace SDKs, we lost the ability of automagically parsing `OTEL_RESOURCE_ATTRIBUTES`. This PR fixes that by leveraging the existing env header parsing mechanism to parse `OTEL_RESOURCE_ATTRIBUTES` (...)

So (2) restored behavior that was expected, and removing it would be a regression.

**Neither of those, however, discusses precedence.** Nothing there establishes that the agent env should rank above the Kubernetes metadata OBI resolved. If anything, the v2 migration guide expects per-target identity to come from the target's own env or from Kubernetes metadata — [`devdocs/config/version-2.0/migration.md`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/872e6f57/devdocs/config/version-2.0/migration.md?plain=1#L382). The current ranking is a by-product of reusing the header parsing mechanism, which put (2) on the same path as (1). **What this issue asks to change is that ranking, not the capability.**

Removing the fallback outright is not the answer either. Because (1) and (2) go through the same function, dropping the fallback kills (1) as well. Values originating from a pod spec are discarded by the guard in `pkg/kube/store.go` when they are still in `$(...)` form, so (1) is the only path that promotes the value kubelet actually substituted to the documented top of the precedence list — and that is exactly the shape the Operator injects (grafana/beyla#1856).

**(1) and (2) should rank differently, but the code treats them identically.**

----

#### Proposed direction

Move the agent env to the bottom of the precedence order. Since the agent env is a process-level constant, parse it once at startup and have `GetAppResourceAttrs` take it as a parameter and lay it down first. Split the implicit fallback in `parseOTELEnvVar` into a process-env variant and a service-env variant, and let `HeadersFromEnv` use the former. **The five merge sites need no changes** — the ordering then falls out naturally.

- A behavior change that comes with it — per-key merging
  - The implementation above turns the fallback from per-variable into **per-key**. Today, if a target declares `OTEL_RESOURCE_ATTRIBUTES` at all, the agent env is ignored **entirely**: an app declaring a single `service.version` makes the deployment-wide `deployment.environment` disappear completely.
  - That also contradicts the original intent. What grafana/beyla#1053 asked for, quoted above, was `deployment.environment` **"in all Traces"** — yet the current behavior silently drops any app that declares attributes of its own out of that "all". With per-key merging, the agent env simply becomes the bottom default layer, which keeps the implementation small and resolves this inconsistency at the same time.
  - So this change moves closer to the original intent rather than being a side effect. Still, **some workloads will genuinely start carrying attributes they did not carry before** — those that declare their own `OTEL_RESOURCE_ATTRIBUTES`. It seems worth calling out as a behavior change in the release notes.

Happy to prepare a PR along these lines — feedback on the proposal is welcome.

----

#### Generative AI disclosure

Claude Code was used to trace the code paths and draft this report. The measurements were performed by me on a real cluster, and the cited code locations and precedence behavior were verified against upstream main. I take responsibility for the conclusions and the proposal.

### Version

`go.opentelemetry.io/obi v0.12.2` (bundled in `opentelemetry-collector-contrib:0.160.0`), confirmed via `otelcol-contrib components`.

### Kernel version

`6.8.0-90-generic`

### OS / Distribution

`Ubuntu 24.04.3 LTS`

### Deployment

Kubernetes

elliehmy:
## Summary

Fixes #3432

When an instrumented target declares no `OTEL_RESOURCE_ATTRIBUTES` of its own, OBI fell back to its own process environment, and that value was placed above the per-workload metadata OBI itself resolved. The result was self-contradicting telemetry: within one resource, `k8s.pod.name` held the agent's env value while `service.instance.id` held the real pod name.

The cause is that `parseOTELEnvVar` returned two values of very different nature without distinguishing them. Callers could not tell whether a value came from the instrumented target's own declaration or from the agent's deployment-wide configuration, so both were treated as top priority.

This splits the fallback into a process-env variant and a service-env variant, and lays the process env down first in `resourceAttrs`. The precedence, from lowest to highest, is now:

1. OBI's own `OTEL_RESOURCE_ATTRIBUTES` — deployment-wide defaults that fill only the keys nobody else declares
2. The metadata OBI resolved for that target — Kubernetes annotations, labels and object metadata, plus host and cloud metadata
3. The target's own `OTEL_RESOURCE_ATTRIBUTES` — unchanged

The five exporter call sites that append `ResourceAttrsFromEnv` after `GetAppResourceAttrs` already had the right order, so they are left untouched.

### Behavior change

The agent env now merges per key instead of per variable.

Previously, a target that declared `OTEL_RESOURCE_ATTRIBUTES` at all made the agent env be ignored entirely: an app declaring a single `service.version` made the deployment-wide `deployment.environment` disappear. Now only the keys the target declares are overridden.

This is closer to what grafana/beyla#1053 originally asked for — `deployment.environment` "in all Traces". Still, **workloads that declare attributes of their own will start carrying attributes they did not carry before**, so this belongs in the release notes.

The second commit records it under "Pending release notes" in `devdocs/telemetry-schema.md`. The change is neither a rename nor a removal, so the schema format cannot express it and nothing else would surface it at release prep.

### Tests

- The service-env parser does not fall back to the process env — covering all three paths: a nil service, a nil `EnvVars`, and an empty `EnvVars` map
- The scenario reported in the issue: the colliding `k8s.pod.name` keeps the real pod name, while a non-colliding key still reaches the target
- The full three-layer precedence, assembled the way the exporters assemble it, including the per-key merge
- The agent env is still subject to the resource selection's exclusion rules

## Validation

- [ ] If this enhances / fixes / changes a core feature, I have updated the [features documentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/devdocs/features.md) and [support matrix](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/SUPPORT_MATRIX.md) as needed — not applicable, neither document covers resource attribute precedence.

## Generative AI disclosure

Claude Code was used to trace the code paths and to draft the implementation and tests. The precedence design and the decision to merge per key are mine, and the measurements reported in the issue were performed by me on a real cluster. I reviewed every change, validated it locally, and take responsibility for the content.


mariomac:
Thanks for the contribution, @elliehmy . So please confirm me if I understanding the issue (🟨 for the next time, please [read the terms for acceptable/unaceptable AI generated code in our repository policies](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/AI-POLICY.md)).

So, let's imagine an scenario:
1. An OBI instance that defines `OTEL_RESOURCE_ATTRIBUTES="service.name=obi"`
2. An instrumented pod defining `OTEL_RESOURCE_ATTRIBUTES="service.name=my-app"`.

Before this PR, the `service.name` for the instrumented pod was `obi` and with this PR, `service.name` is `my-app`.  Did I understand correctly?

elliehmy:
@mariomac Nothing changes in that exact scenario.

The pod's own `OTEL_RESOURCE_ATTRIBUTES` was already checked before OBI's, so `service.name` is `my-app` both before and after this PR. OBI's own value was only used when the pod declared nothing.

The case that does change:

1. OBI defines `OTEL_RESOURCE_ATTRIBUTES="service.name=obi"`
2. The instrumented pod defines no `OTEL_RESOURCE_ATTRIBUTES`, and belongs to a `my-app` Deployment, so OBI resolves `service.name=my-app` from its Kubernetes metadata

Before, `service.name` was `obi` — OBI's env overrode the name OBI had itself resolved. Now it is `my-app`, and OBI's env only fills keys nothing else provides.

There is a second change worth confirming: the agent env now merges per key. Before, a pod declaring any `OTEL_RESOURCE_ATTRIBUTES` — even just `service.version` — discarded OBI's env entirely. Now it only overrides the keys it declares.

NimrodAvni78 (approved):
LGTM

mariomac (approved):
Understood. Thanks for the explanation!

PR #3513, opened 2026-09-21
Bound grpc-go header verification

MrAlias:
## Summary

Fix the grpc-go verifier regression on RHEL 8 without reducing header coverage.

- replace the nested `stricmp` loop with a fixed 11-byte, case-insensitive comparison
- retain the existing 256-field scan for grpc-go's planned 8 KiB header list
- preserve the conservative behavior for oversized or unreadable metadata

## Root cause

The application-ownership handling added in #3428 scans grpc-go client headers before injecting a traceparent. Each 11-byte candidate key called `stricmp`, placing an 11-iteration character loop inside the 256-iteration field loop.

That nested-loop shape caused pathological state exploration in the backported Linux 4.18 verifier used by RHEL 8.9 and 8.10. The outer bound was not the problem: after removing the inner loop, the 256-field version loads in the same time as the 64-field experiment.

## Fix

Compare the fixed-length `traceparent` key using unrolled ASCII case-folded byte operations. This removes the inner verifier loop while retaining case-insensitive matching and the original 256-field behavior.

Requests with more than 256 fields, or metadata that cannot be read safely, continue to be treated as application-owned so OBI does not overwrite application context.

## Validation

- [BPF verifier run 35622888383](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/35622888383): all 17 jobs passed
  - [RHEL 8.9](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/35622888383/job/106410968321): passed in 32m47s
  - [RHEL 8.10](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/35622888383/job/106410968409): passed in 20m50s
- focused 256-field verifier load: 7.02s; the nested-loop version did not complete within five minutes locally
- `TestSuite_GRPCGoTraceparentOwnership`: passed for plaintext, TLS, and wrapped connections, including the 40-metadata and 260-metadata cases
- BPF generation, formatting, linting, unit tests, and gotracer tests passed


MrAlias:
Validation update: the branch has been amended to retain the original 256-field bound. The earlier proposed 64-field cap is not part of the current diff; the only behavior change is replacing the nested `stricmp` loop with a fixed 11-byte ASCII case-insensitive comparison.

Evidence:

- The focused verifier load completes in 7.02s with 256 fields, effectively identical to 7.03s with 64 fields; the old nested loop did not complete within five minutes locally.
- `TestSuite_GRPCGoTraceparentOwnership` passes for plaintext, TLS, and wrapped connections, including the 40-metadata and over-limit 260-metadata cases.
- [BPF verifier run 35622888383](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/actions/runs/35622888383) passed all 17 jobs. RHEL 8.9 passed in 32m47s and RHEL 8.10 passed in 20m50s.

This isolates the nested string loop as the verifier regression while preserving the original 8 KiB/256-field coverage.

mmat11 (approved):

grcevski (approved):

PR #3514, opened 2026-09-21
fix(procs): avoid oversized environment maps

### LINKED ISSUE #3506: envStrsToMap pre-sizes from the raw NUL-split count: an empty ~40 MiB map per process with a NUL-padded environ (follow-up to #3226)
## Summary

Beyla v3.35.0 (OBI v0.13.0) is OOMKilled on every start on 4 of 218 nodes of one 128-vCPU class, while
the same image stays at a few hundred MiB on the other 213. Heap profiles put ~95 % of the live heap in
`procs.envStrsToMap`, which builds one `map[string]string` per typed process:

    make(map[string]string, len(varsStr))   // len is the raw NUL-split field count

For a process whose `/proc/<pid>/environ` is NUL padding, that pre-sizes the table for ~643k entries
and the loop then inserts zero. One such process costs ~40 MiB, held while it is tracked; 258 of them
on one node (two workloads, 129 workers each) is 10.0 GiB and the OOMKilled loop, while a node with
129 of them plateaus at 5.0 GiB. Filtering empty entries before `make`, or not pre-sizing from the
split count, removes it.

This is a follow-up to #3226, not a duplicate: we already run its fix (#3232, `UnsafeServiceAttrs`),
`maps.clone` has 0 live bytes in our profiles, and what grows is the cached map itself.

## Environment

- Beyla v3.35.0 (OBI v0.13.0, OTEL SDK 1.46.0), also seen on our 2.6.3 build.
- Kubernetes 1.2x, DaemonSet, container limits cpu=4, memory=6Gi/12Gi (12Gi reproduced).
- Nodes: 128 vCPU / 512 GiB, Debian 11, kernels 5.10.0-20-amd64 and 5.10.0-30-amd64.
- Config: log_level info, channel_buffer_len 1000, meta_restrict_local_node true, containers_only
  true, k8s_namespace regex, exclude_services for shells/sshd/filebeat, env.GOMEMLIMIT.
- Diagnostics, all reverted afterwards: `GODEBUG=gctrace=1`, and `profile_port: 6060` patched into the
  ConfigMap for one pod and reverted 2 s after that pod started (a running pod keeps the config it was
  started with). Profiles were read through
  `kubectl get --raw /api/v1/namespaces/beyla/pods/<pod>:6060/proxy/debug/pprof/heap`.

## Symptom

- OOMKilled (exit 137) on every start: ~67 s at a 12Gi limit, ~22 s at 6Gi.
  `container_memory_max_usage_bytes` equals the limit exactly.
- Live heap after GC (gctrace): 2.7 GB at +41 s, 5.1 GB at +47 s, 7.8 GB at +53 s, 10.7 GB at +63 s
  and still rising. Sibling nodes in the same class plateau at 5.8 GB.
- `go tool pprof -inuse_space`, last capture: `procs.envStrsToMap` 10,254 MB = 95.4 % of 10,750 MB. By
  size class the retention is 262,244 objects of exactly 40,960 B (plus 262,152 of 32 B). A sibling
  node with half the workers: 130,949 objects of 40,960 B = 5.0 GiB.
- Churn over a 143 s window, with 0 live bytes: `maps.clone` 41.4 GB, `envStrsToMap` 6.4 GB,
  `strings.genSplit` 2.5 GB (the 628 KB environ split into ~643k fields on every read).

## What the code says

1. **Sizing.** `pkg/internal/procs/procenv.go` sizes `vars` from `len(varsStr)` - the raw
   `strings.Split(environ, "\x00")` result - and then keeps only entries with a non-empty key and
   value. A NUL-padded block therefore sizes the table for 643k entries and inserts none. Filtering
   empty entries first (plain `make(map[string]string)`, or counting non-empty fields) is
   behaviour-identical and removes the 40 MiB.
2. **Retention.** The map is part of the process's `exec.FileInfo` in `PIDsFilter.current`, released
   only by `PIDsFilter.removePID()` -> `pruneExpired()`, i.e. `pidRemovalRetention = 5 * time.Minute`
   after the process is gone (`pkg/ebpf/common/pids.go`). Nothing shrinks or invalidates it while the
   process lives, which matches the measurements: 129 live maps for 129 long-lived workers, 258 for
   258.
3. **Per-event copies.** #3232 removed the clone from the span path, but
   `exec.FileInfo.ServiceAttrs()` still does `maps.Clone(fi.service.EnvVars)`
   (`pkg/appolly/discover/exec/file.go:152`) and is still reached from process-event handlers:
   `metricsReporter.handleProcessEvent` (`pkg/export/prom/prom.go:1689`),
   `MetricsReporter.onProcessEvent` (`pkg/export/otel/metrics.go:1391`) and
   `HostProcessEventDecoratorProvider` (`pkg/appolly/traces/traces.go:36`) - the callers our profiles
   attribute the `maps.clone` churn to. The occasional ones (`AppendKubeMetadata`,
   `pkg/transform/k8s.go:471`; the language metadata scanners) matter less.
4. **Where the environ comes from.** The *size* is inherited: kubelet injects a service-link variable
   set for every Service in the pod's namespace (~7 each; ours carried ~11,000 variables / ~630 KB).
   The *zeroing* is the process hiding its environment while it rewrites its title - `ngx_setproctitle()`
   pads the rest of the argv+environ block with `'\0'` (`NGX_SETPROCTITLE_PAD`), as postgres and
   sendmail do - and `/proc/<pid>/environ` keeps reporting the original block, so the size survives
   while the content is gone. Measured inside one pod: nginx 641,009 B with 100 % NUL, while `bash`,
   `tail` and `tini` show the same ~641 KB with ~1.7 % NUL and their variables intact.

## Isolated reproduction

```go
// Sizes one environment map, using the exact code path of pkg/internal/procs/procenv.go.
// go run issue_repro.go
package main

import (
	"fmt"
	"runtime"
	"strings"
)

// verbatim from pkg/internal/procs/procenv.go
func envStrsToMap(varsStr []string) map[string]string {
	vars := make(map[string]string, len(varsStr))
	for _, s := range varsStr {
		keyVal := strings.SplitN(s, "=", 2)
		if len(keyVal) < 2 {
			continue
		}
		key := strings.TrimSpace(keyVal[0])
		val := strings.TrimSpace(keyVal[1])
		if key != "" && val != "" {
			vars[key] = val
		}
	}
	return vars
}

// what procfs.Proc.Environ() returns: strings.Split(block, "\x00"), trailing empty dropped
func environLike(block []byte) []string {
	fields := strings.Split(string(block), "\x00")
	if len(fields) > 0 {
		fields = fields[:len(fields)-1]
	}
	return fields
}

func report(name string, block []byte) {
	nul, empty := 0, 0
	for _, c := range block {
		if c == 0 {
			nul++
		}
	}
	fields := environLike(block)
	for _, f := range fields {
		if f == "" {
			empty++
		}
	}
	var a, b runtime.MemStats
	runtime.GC()
	runtime.ReadMemStats(&a)
	m := envStrsToMap(fields) // the map alone
	runtime.ReadMemStats(&b)
	runtime.KeepAlive(m)
	fmt.Printf("%s\n  block=%d B, NUL bytes=%d (%.3f%%), split fields=%d, of which empty=%d\n"+
		"  make(map, %d entries) -> allocated %.1f MB, len(map)=%d\n\n",
		name, len(block), nul, 100*float64(nul)/float64(len(block)), len(fields), empty,
		len(fields), float64(b.HeapAlloc-a.HeapAlloc)/1e6, len(m))
}

func main() {
	// (a) an nginx worker after ngx_setproctitle(): same-size block, zeroed, one title fragment left
	zeroed := make([]byte, 643166)
	copy(zeroed, "ocess /usr/sbin/nginx")
	report("(a) NUL-padded environ (nginx worker after setproctitle)", zeroed)

	// (b) the same order of size with real content, like its sibling processes (~12k service-link vars)
	var real []byte
	for i := 0; i < 12000; i++ {
		real = append(real, []byte(fmt.Sprintf("SVC_%04d_SERVICE_PORT_TCP_%d=10.0.0.%d", i, i, i%256))...)
		real = append(real, 0)
	}
	report("(b) real environ of the same order of size (sibling process)", real)
}
```

Output (go1.26.8, the toolchain our release reports):

```
(a) NUL-padded environ (nginx worker after setproctitle)
  block=643166 B, NUL bytes=643145 (99.997%), split fields=643145, of which empty=643144
  make(map, 643145 entries) -> allocated 52.3 MB, len(map)=0

(b) real environ of the same order of size (sibling process)
  block=501720 B, NUL bytes=12000 (2.392%), split fields=12000, of which empty=0
  make(map, 12000 entries) -> allocated 1.0 MB, len(map)=12000
```

The allocation figure moves between runs (46-52 MB) because it is a `HeapAlloc` delta; with
`runtime.MemProfileRate = 1` the retained table is exactly **1,024 objects of 40,960 B = 40.00 MiB**,
which is the size class that dominates the live profiles of the affected nodes. Note what the two
cases have in common: ~640 KB of input, one real variable vs ~12,000, and a ~50x difference in what
`make(map...)` allocates - the input that decides it is the *separator count*, not the byte size.

One question on the shape of a fix, if you have a preference: filtering empty entries before
`make`, or not pre-sizing the map from the split count at all, would both work for us - or would you
rather add a configurable bound on the environment map size?

## Questions

1. Should the map be sized from the raw split count? Either option in (1) above is
   behaviour-identical for us.
2. Is a ~40 MiB cache entry per tracked process intended, given the map is only queried for a few
   keys (`OTEL_SERVICE_NAME` / `OTEL_RESOURCE_ATTRIBUTES` in `exec/file.go`, the
   `OTEL_EXPORTER_OTLP_*` lookups in `app/request/span.go`, `TMPDIR` for the Java/.NET targets)?
   `kubecache` already has an allow-list concept (`usefulEnvVars`,
   `pkg/kube/kubecache/meta/informers_init.go:44`); we cannot tell whether the language
   service-metadata scanners need the whole map.
3. Would it be acceptable to skip the `EnvVars` clone on the process-event paths in (3) as well?

Attached evidence (or shared privately on request):
- binary heap profiles from two affected nodes - one at the 5.2 GB plateau, one captured through an OOM
  cycle. Use the binary form with `go tool pprof -inuse_space` / `-inuse_objects`; the `?debug=1` text
  output misreports totals by ~13x.
- `GODEBUG=gctrace=1` lines from the same runs.
- startup log of a failing container.

riccardomenegazzo:
## Summary

- size environment maps from valid entries instead of the raw NUL-split field count
- use allocation-free parsing while preserving the existing trimming and validation behavior
- cover heavily NUL-padded environments with a regression test

This prevents a mostly empty `/proc/<pid>/environ` from creating a roughly 40–50 MiB map for every tracked process.

Fixes #3506

## Testing

- `go test ./pkg/internal/procs/`
- `go test -race ./pkg/internal/procs/`
- `go vet ./pkg/internal/procs/`
- package-scoped `golangci-lint`
- repeated the focused tests 10 times
- `git diff --check`

A local benchmark using 640,000 split fields reduced the parser from approximately 52.2 MB and 642,050 allocations per call to under 1 KB and 2 allocations.

## AI assistance

Codex was used to inspect the relevant code, assist with the implementation and tests, and draft this description. I reviewed the complete diff, validated the behavior and test results, and understand and take responsibility for the change.


pkg/internal/procs/procenv.go
```diff
@@ -12,24 +12,38 @@ import (
 )
 
 func envStrsToMap(varsStr []string) map[string]string {
-	vars := make(map[string]string, len(varsStr))
+	vars := make(map[string]string, validEnvCount(varsStr))
```
mmat11, line 15: question: can't we just not size the map here instead of parsing twice?

mariomac: Yeah, I wonder whether it would be more efficient just doing `vars := map[string]string{}` and let the Go runtime grow/rebalance the map at runtime.

In case we want to be 100% sure of the rebalancing, add this line before returning the actual map:
```golang
	actuallyReturned := make(map[string]string, len(vars))
	maps.Copy(actuallyReturned, vars)
	return actuallyReturned
```

But I'm fearing we might be entering in the hazardous world of micro-optimization :D

marctc:
> A local benchmark using 640,000 split fields reduced the parser from approximately 52.2 MB and 642,050 allocations per call to under 1 KB and 2 allocations.

do you mind pushing the benchmarks and show comparison table if possible? since this PR is about optimizations, I think its good to have it around

pkg/internal/procs/procenv.go
riccardomenegazzo: Good question. I benchmarked both approaches before deciding.

With 128 valid entries, leaving the map unsized was roughly twice as slow (13.2 vs 6.7 µs), used about twice the memory (18.4 vs 9.3 KiB), and required 13 allocations instead of 4. It was faster for the pathological 640k mostly-empty input (4.6 vs 9.0 ms), but both approaches kept that case at 336 B and 2 allocations.

I also tried rebuilding the map before returning it, but the extra copy was slower in both scenarios and added allocations. Since regular environments should be the common path, I kept the two-pass approach. I added a permanent benchmark in 46b128c7 and will post the base/head comparison on the PR.

riccardomenegazzo:
Sure — pushed in 46b128c7. The benchmark covers both a representative 128-variable environment and the reported 640,000-field NUL-padded case.

I ran base (473191b) and head (46b128c7) eight times with Go 1.27.1 on linux/amd64 and compared the results with benchstat:

| Benchmark | Base time | PR time | Delta | Base B/op | PR B/op | Base allocs/op | PR allocs/op |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Typical128 | 10.463 µs | 9.285 µs | -11.26% | 13,656 | 9,560 | 132 | 4 |
| NULPadded640K | 17.02 ms | 10.17 ms | -40.26% | 52,225,356 | 336 | 642,050 | 2 |

The time differences were statistically significant (p=0.007 for Typical128 and p<0.001 for NULPadded640K, n=8).

mmat11 (approved):

mariomac (approved):
Thank you @riccardomenegazzo for all the explanations! LGTM

grcevski (approved):

riccardomenegazzo:
> Thank you @riccardomenegazzo for all the explanations! LGTM

Thank you, it is always a pleasure to contribute for such a great project!

PR #3515, opened 2026-09-21
fix(traces): pass remote parent context to samplers

### LINKED ISSUE #3190: parentbased_* trace samplers never see the remote parent: ShouldSample receives the exporter lifecycle context

riccardomenegazzo:
## Summary

- build the sampler parent context from each span's captured trace context
- preserve the incoming sampled flag instead of evaluating every span as a root
- keep root-sampler behavior unchanged when no valid parent is available
- cover sampled, unsampled and missing-parent decisions, plus the exact context passed to the sampler

This restores the intended behavior of all `parentbased_*` samplers. In particular, gateway-sampled requests are no longer dropped by `parentbased_always_off`, while an unsampled remote parent is no longer overridden by an always-on root sampler.

Fixes #3190

## Testing

- `go test ./pkg/export/otel/tracesgen/`
- `go test -race ./pkg/export/otel/tracesgen/`
- `go vet ./pkg/export/otel/tracesgen/`
- package-scoped `golangci-lint`
- `git diff --check`

## AI assistance

Codex was used to inspect the relevant sampling path, assist with the implementation and tests, and draft this description. I reviewed the complete diff, validated the behavior and test results, and understand and take responsibility for the change.


marctc:
@riccardomenegazzo this solution is being attempted many times (see the PR/issues linked to the issue) and i doesn't work. It needs a bit more digger than just LLMing.

riccardomenegazzo:
You're right. I missed #3191 and the earlier #416/#444 attempts before opening this PR. This change is functionally equivalent, and the CI failure on both amd64 and arm64 confirms the same problem: a server or consumer span does not prove that its flags came from an external sampling decision, because the parent context may have been propagated by BPF.

That distinction is unavailable in the exporter, so this cannot be fixed safely at this layer alone. The sampling decision needs to be made and carried through the BPF propagation path first.

I'm closing this PR. Sorry for the review noise, and thank you for pointing me to the prior work.

PR #3517, opened 2026-09-21
fix(test): fix intermittent test failure on wildfly/arm64

grcevski:
## Summary

~~Seems like my 0 response check for the SSL double reporting is failing randomly on arm64. It happened twice on main and I'm adding some tracing to see if I can see what's up. If it doesn't reproduce in CI, I'll remove the check to avoid CI failures until I investigate. I've not been able to reproduce this locally and it has never failed on x86 :(.~~

So my test reproduced a very weird edge case in tracer attachment. The wildfly runner launches two java executables, but the first one stops right away, so we fail on:
```
if err := tracer.NewExecutable(exe, ie); err != nil {
```

What this does is that we have loaded one generic tracer but we don't follow through in attacher and set it up in:
```
	ta.existingTracers[key] = executableTracer{
		tracer:     tracer,
		generation: ie.ExecutableGeneration,
	}
	if tracer.Type == ebpf.Generic {
		if ta.reusableTracer != nil {
			ta.monitorPIDs(ctx, ta.reusableTracer, ie)
		} else {
			ta.reusableTracer = tracer
		}
	} else {
		ta.reusableGoTracer = tracer
	}
```

The second Java process comes through and since there is no existing tracer or reusable tracer setup, it makes another generic tracer. 

So GC hasn't run yet to remove the probes of the first generic tracer and for a bit of time we have two separate generic tracers running and monitoring the same java service. Not only we see a duplicate Java SSL event, all probes seem duplicated. 

Since we share the eBPF maps, the symptom is that we see duplicated HTTP event start for the first couple of requests and eventually once the first tracer dies, we continue fine.

## Validation


pkg/appolly/discover/attacher.go
```diff
@@ -405,19 +405,24 @@ func (ta *traceAttacher) getTracer(ctx context.Context, ie *ebpf.Instrumentable)
 
 	if err := tracer.Init(ta.EbpfEventContext, ta.Cfg); err != nil {
 		ta.log.Error("couldn't trace process. Stopping process tracer", "error", err)
+		if closeErr := tracer.Close(); closeErr != nil {
```
MrAlias, line 408: `Close()` cannot release links attached before a required init probe fails. `loadTracer` accumulates kprobe and tracepoint links in its local `instrumenter.closables`, but only transfers them to the program after both attachment passes succeed. If a later required probe fails, this branch closes the programs while those earlier links remain open; the next discovery attempt can attach a second copy and reproduce the duplicate events this change is meant to prevent. Could we close or transfer the partial init links on every `loadTracer` error and cover a required probe failing after an earlier attachment?

pkg/ebpf/tracer_darwin.go
```diff
@@ -46,6 +46,8 @@ func (pt *ProcessTracer) Init(_ *ebpfcommon.EBPFEventContext, _ *obi.Config) err
 	return nil
 }
 
+func (pt *ProcessTracer) Close() error { return nil }
```
MrAlias: The Darwin `Close` stub never reads the new `closeOnce` and `closeErr` fields, so the macOS lint job fails with `unused` at `pkg/ebpf/tracer.go:162-163`. Could we keep this state in the Linux implementation or make the Darwin implementation consume it so the supported macOS target is green?

internal/test/oats/http/docker-compose-java-tls-wildfly.yml
```diff
@@ -65,7 +65,7 @@ services:
       OTEL_EBPF_LOG_LEVEL: DEBUG
       OTEL_EBPF_BPF_DEBUG: "true"
       OTEL_EBPF_JAVA_ROUTE_HARVEST_DELAY: 10s
-      #OTEL_EBPF_JAVAAGENT_DEBUG: "true"
+      OTEL_EBPF_JAVAAGENT_DEBUG: "true"
```
MrAlias: This was enabled by the temporary diagnostic commit, while the corresponding OATS diagnostic assertions are no longer in the final diff. Leaving it on makes every WildFly run use Java-agent debug verbosity, adding log volume and timing overhead to the regression. Could we restore the opt-in setting?

```suggestion
      #OTEL_EBPF_JAVAAGENT_DEBUG: "true"
```

grcevski: sure, I turned it off.

pkg/ebpf/tracer_darwin.go
grcevski: I added a stub for macos that it doesn't complain.

pkg/appolly/discover/attacher.go
grcevski: yeah ok, I guess I was overzealous about adding the check in two places, the real failure was on attach, but this review exposed a new potential partial load issue. I think the partial load would be rare, IMO, it would mean we have a bug in OBI and a tracer failed to load, at which point the actual failure scenario wouldn't happen, next time we'll fail again.

I added some code to attempt to fix this. let's see if tests pass. But this really could be a follow-up.

MrAlias (approved):

PR #3518, opened 2026-09-21
Support traceparent injection across CONTINUATION frames

### LINKED ISSUE #3152: Support traceparent injection across HTTP/2 CONTINUATION frames
### Component

OBI HTTP/2 context propagation

### Enhancement

HTTP/2 header blocks can span an initial HEADERS frame and one or more CONTINUATION frames. The existing injection paths are primarily designed around a complete header block available with the HEADERS frame. Large request headers can therefore cause OBI to decline injection or leave propagation coverage dependent on buffering details.

PR #3021 explored constructing and mutating continuation frames. That capability is not required to prevent duplicate application-owned headers in #2769 and should be designed as an independent enhancement.

### Required behavior

For a supported connection with context propagation enabled and no application-owned `traceparent`, a header block spanning HEADERS/CONTINUATION frames should contain exactly one valid OBI `traceparent` without violating HTTP/2 framing or HPACK semantics, within the existing direct-write writer-offset, frame-size, and capacity guards.

### Scope

- Establish the current behavior for fragmented header blocks in the Go direct and plaintext `sk_msg` paths.
- Define which injection paths can safely support fragmented blocks; document any path that must intentionally decline them.
- Preserve HEADERS flags, optional padding/priority prefixes, CONTINUATION ordering, stream ID, and final END_HEADERS placement.
- Preserve HPACK field ordering and dynamic-table state.
- Append to the existing terminal fragment when the complete field fits within the current direct-write guards.
- Cover one and multiple CONTINUATION frames.
- Cover TLS and plaintext, persistent connections, and concurrent streams.
- Verify that requests with an application-owned `traceparent` remain unchanged once #2769 is complete.

### Dependencies and safety

This enhancement must not introduce a new multi-step mutation sequence on a path until that path has a defined failure-safety contract. Coordinate with the direct user-buffer and `sk_msg` mutation hardening issues as applicable.

### Out of scope

- Negotiated larger-frame support, changes to the existing writer/frame bounds, and exact frame-size boundary coverage; these are tracked in #3531.
- Creating an additional CONTINUATION frame when the complete field does not fit in the existing terminal fragment.
- Duplicate-header ownership detection itself (#2769).
- Architecture-specific interior probes unless a separately demonstrated writer-boundary problem requires them.
- General audit maps/listeners or test-only production controls.

### Definition of done

- Fragmented large-header requests receive exactly one OBI `traceparent` on every injection path claimed as supported and fitting the existing guards.
- All generated frame sequences are accepted by a standards-compliant peer.
- HEADERS/CONTINUATION lengths, flags, ordering, and HPACK decoding remain valid.
- Multiplexed streams do not share fragment or trace state.
- Failure paths preserve a usable connection and obey the mutation-safety contract for that path.
- Existing small-header injection, application ownership, tracing, metrics, and disabled propagation remain unchanged.

MrAlias:
## Summary

- Retain direct Go HTTP/2 injection state across fragmented header blocks and append the generated traceparent to the terminal CONTINUATION frame.
- Cover net/http, x/net/http2, and grpc-go while preserving application-owned traceparents.
- Keep fail-closed frame, capacity, stream, and mutation checks, and document why fragmented plaintext sk_msg traffic remains unchanged.

## Validation

- make generate/all
- go test ./pkg/internal/ebpf/gotracer -count=1
- make bpf_go_h2_write_transaction && ./bpf_go_h2_write_transaction (from bpf/tests)
- make compile
- make clang-tidy
- TestHTTP2Go integration suite, including TLS and h2c
- TestHTTP2GoWithHTTPProtocols integration suite, including TLS and h2c

Fixes #3152.


bpf/gotracer/go_nethttp.c
```diff
@@ -1607,6 +1608,54 @@ int GUARDED_PROG(obi_uprobe_net_http2FramerWriteHeaders, struct pt_regs *, ctx)
     return 0;
 }
 
+static __always_inline int on_http2FramerWriteContinuation(struct pt_regs *ctx) {
+    if (!g_bpf_header_propagation || !g_bpf_probe_write_user_enabled) {
+        return 0;
+    }
+
+    go_addr_key_t g_key = {};
```
grcevski: do you think it would be possible to make a shared helper from this code here that can be reused in the grpc and the http2 path? maybe for parts of it, since we delete from two different maps?

mmat11 (approved):

grcevski (approved):

grcevski (approved):
LGTM! I left a small nit comment if it's possible to reuse some of the new code between grpc and http2.

bpf/gotracer/go_nethttp.c
mmat11: whoops, sorry, I merged before seeing this -- emails from GH are slow today :/

PR #3523, opened 2026-09-22
fix: keep distinct metric label tuples separate

immanuwell:
## Summary

When HTTP duration metrics select `service.name` and `service.namespace`, `checkout:blue` / `prod` and `checkout` / `blue:prod` share a cache key. 
The first series eats both counts, the other disappears

Use the length-prefixed key already used by runtime metrics.

Repro: `go test ./pkg/export/prom -run TestExpirerDistinctServiceLabelsWithColons -count=1`. 
Before the fix, it reports a count of 2 and a missing series

## Validation

- `make compile` and focused race tests pass. Host `make verify` hit an unrelated listener on port 8080; envtest passes in an isolated container.


mariomac (approved):
LGTM!

pkg/export/expire/expiry_map.go
```diff
@@ -133,6 +133,13 @@ func (ex *ExpiryMap[T]) All() []T {
 	return items
 }
 
-func labelsKey(lbls []string) string {
-	return strings.Join(lbls, ":")
+// LabelsKey encodes label values without ambiguity between values.
+func LabelsKey(labels []string) string {
+	key := make([]byte, 0, len(labels)*8)
```
MrAlias, line 138: `len(labels)*8` is an arbitrary capacity estimate: encoded labels can be either shorter or much longer, and there is no benchmark here showing the guess is useful. Could we keep this simpler and let `append` size the buffer instead of including an confusing magic number capacity inialization?

```suggestion
	var key []byte
```

MrAlias (approved):

pkg/export/prom/prom_runtime_histograms.go
```diff
@@ -306,11 +306,5 @@ func (c *goRuntimeHistogramCollector) descriptor(
 }
 
 func runtimeMetricLabelTuple(labels []string) string {
```
mmat11, line 308: why do we need this wrapper func?

PR #3524, opened 2026-09-22
fix(java): don't send SIGQUIT to a JVM that won't answer it

NimrodAvni78:
## Summary

OBI starts Java attach handshake by sending `SIGQUIT` signal, it is sent to anything mapping `libjvm.so`, without checking the target would act on it. for example:

- `-Xrs` tells JVM not to register a `SIGQUIT` handler, so the signal **kills the JVM** (Temurin 8 and 21).
- `-XX:+DisableAttachMechanism` catches the signal but never starts the attach listener, so it just does a thread dump in the application's own stdout and keeps waiting for a socket to appear.

The handshake now checks two things first:

| check | refuses |
| --- | --- |
| `-XX:±DisableAttachMechanism` | a JVM that will not answer the handshake |
| `SigCgt`/`SigIgn` for `SIGQUIT` | a target the signal would terminate |

The wait for the runtime to install its handler is shared too. A JVM is briefly at the default disposition between `exec` and `Threads::create_vm`, so without it a process discovered at exec would be refused permanently for a condition that clears on its own.

The flag is read from the four places in order (`JAVA_TOOL_OPTIONS`, `JDK_JAVA_OPTIONS`, command line, `_JAVA_OPTIONS`), last setting wins. The three variables are read from the environment, the launcher expands them after `exec` so they never reach the command line. The scan stops at the main class or after `-jar`, so a launcher passing the flag to a child JVM is not read as the flag of the process holding it. Variables the launcher ignores, such as `MAVEN_OPTS`, are not option sources.

What it does not catch: a flags file, a quoted value inside one of those variables, and a JVM slow enough to start that it is still without a handler when the wait ends, which is refused for its lifetime since injection is not retried.

## Validation


pkg/internal/jvmtools/jvm/signal_check.go
```diff
@@ -0,0 +1,209 @@
...
+
+const sigquit = unix.SIGQUIT
+
+// attachWait bounds how long to wait for HotSpot to install its SIGQUIT handler
+// in Threads::create_vm, so a process discovered as soon as libjvm is mapped is
+// not refused for a condition that clears.
+const attachWait = 2 * time.Second
+
+const (
+	attachDisablingOption = "-XX:+DisableAttachMechanism"
```
grcevski, line 30: I'm not sure if OpenJ9 has an option like this too? OpenJ9 support can be a follow-up.

pkg/internal/jvmtools/jvm/signal_check.go
```diff
@@ -0,0 +1,209 @@
...
+// around the command line by the order HotSpot applies them. Each is expanded by
+// the launcher after exec, so none of them reaches the command line itself.
+var (
+	optionVariablesBeforeCommandLine = []string{"JAVA_TOOL_OPTIONS", "JDK_JAVA_OPTIONS"}
+	optionVariablesAfterCommandLine  = []string{"_JAVA_OPTIONS"}
+)
+
+// valueOptions take their value as the following argument, which must not be
+// mistaken for the main class that ends the VM options.
+var valueOptions = []string{
```
grcevski, line 45: I wonder if this can be done simpler. Would it work if we just scanned for +DisableAttachMechanism on the command line and env variable options and ensured that it's not followed by -DisableAttachMechanism. Last one wins always, I'm not sure env vars take precedence over command line.

I guess you are being pedantic with someone actually having -XX:+DisableAttachMechanism as a Java application argument?

grcevski (approved):
LGTM! I left two questions, but this looks good.

pkg/internal/jvmtools/jvm/signal_check.go
NimrodAvni78: I dug into it a bit and you are right, OpenJ9 attaches without a signal, we detect it by its attachInfo file, which an OpenJ9 with -Dcom.ibm.tools.attach.enable=no never creates. so it fell onto the HotSpot path and got SIGQUIT, which it catches and answers with a javacore. added a check: on the HotSpot path, anything that maps OpenJ9's libj9vm is refused, since a signal can never attach it. tested on OpenJ9 8 and 21.

grcevski: Nice!

pkg/internal/jvmtools/jvm/signal_check.go
NimrodAvni78: probably can be simpler, I can drop it if you prefer. the order does matter though, _JAVA_OPTIONS overrides the command line but JAVA_TOOL_OPTIONS doesn't, so a scan might get it wrong

mmat11 (approved):

PR #3525, opened 2026-09-22
Omit empty string span attributes

NimrodAvni78:
## Summary

OBI emits span attributes as `""` when the parser could not determine a value. OTLP has no absent value, so `server.address=""` asserts the server address *is* the empty string rather than that OBI could not resolve it, and no consumer can tell the two apart.

## Fixes that came with it

`url.scheme` was appended unguarded on the HTTP client branch while the server branch already guarded it, so a client span whose `Statement` carried no scheme shipped `url.scheme=""`.

The AWS SNS branch emitted `messaging.operation.name=""` and `rpc.method="SNS/"` when the `Action` parameter was absent, while its SQS twin eleven lines above rejects such a span. `SNSRPCMethod` now mirrors the existing `S3RPCMethod`, returning nothing for an empty operation.

The Elasticsearch branch appended `db.response.status_code` twice, so that span carried a duplicate key.

Six `gen_ai.response.model` fallbacks emitted the request model from an `else` without checking it, so a span where neither model parsed reported `gen_ai.response.model=""` while `gen_ai.request.model` was correctly absent.

`gen_ai.operation.name` is a closed enum, and only the OpenAI and Qwen branches omitted an empty variant. All eight branches now agree.

## Not affected

Manual-span attributes come from the application through the Go Auto SDK or the Node bridge, so an empty string there is the application's own value. Metrics are unchanged.

## Validation


pkg/export/otel/tracesgen/tracesgen.go
```diff
@@ -1449,83 +1402,73 @@ func traceAttributesSelectorInternal(span *request.Span, optionalAttrs map[attr.
...
 	case request.EventTypeKafkaServer, request.EventTypeKafkaClient:
 		attrs = []attribute.KeyValue{
-			request.ServerAddr(request.HostAsServer(span)),
 			request.ServerPort(span.HostPort),
 			semconv.MessagingSystemKafka,
-			semconv.MessagingDestinationName(span.Path),
-			semconv.MessagingClientID(span.Statement),
 		}
+		attrs = appendIfSet(attrs, semconv.MessagingDestinationName, span.Path)
+		attrs = appendIfSet(attrs, semconv.MessagingClientID, span.Statement)
```
MrAlias, line 1466: How does this change distinguish a parser-observed empty value (such as Kafka’s valid `clientIDLen == 0`) from an unset value, and what establishes that both should be omitted as the same state?

NimrodAvni78: The rule should be to emit when an empty value "" has a meaning instead of meaning "unparsable"
empty kafka client id is a good exception
I went over the attributes and decided where it makes sense and it seems i missed that one
but its hard to know since in go we dont really have a way to differentiate between a "" and a null

grcevski (approved):
LGTM! Nice refactor, I really like the appendIfSet helper.

devdocs/telemetry-schema.md
```diff
@@ -111,6 +111,30 @@ section empty once drained.
...
+  string. It covers every such attribute the span exporter appends, among them
+  `server.address`, `client.address`, `service.peer.name`, `url.scheme`, `url.full`,
+  `db.namespace`, `db.collection.name`, `db.query.text`, `elasticsearch.node.name`,
+  `graphql.operation.name`, `graphql.document`, `aws.s3.key`, `aws.s3.bucket`,
+  `aws.sqs.queue.url`, `aws.request.id`, `aws.extended_request_id`, `cloud.region`,
+  `dns.question.name`, `messaging.message.id`, `messaging.client.id`,
+  `messaging.destination.name` on the AWS SQS and SNS spans, `db.response.status_code`,
+  `rpc.method` on SNS, and the `gen_ai.*` operation, model, response-id, conversation-id,
+  provider-name and message-payload attributes. A consumer
+  selecting on the presence of one of these sees it absent where it previously carried `""`.
```
MrAlias: `messaging.client.id` and GenAI operation names are exceptions to this blanket absence claim. Kafka deliberately preserves `messaging.client.id=""` for an observed zero-length wire value, and `genAIOperationName` maps a missing required operation to `_OTHER`. As written, this tells consumers both keys disappear. Could we scope the list to the attributes that actually become absent?

```suggestion
  `dns.question.name`, `messaging.message.id`, `messaging.client.id` on MQTT and NATS spans,
  `messaging.destination.name` on the AWS SQS and SNS spans, `db.response.status_code`,
  `rpc.method` on SNS, and the `gen_ai.*` model, response-id, conversation-id, provider-name
  and message-payload attributes. A consumer selecting on the presence of one of these sees
  it absent where it previously carried `""`.
```

MrAlias (approved):
The exporter behavior looks consistent after the Kafka exception, but the pending release note still overstates which empty attributes disappear.

PR #3534, opened 2026-09-22
fix: reconstruct multi-record FastCGI params

MrAlias:
## Summary

- reconstruct a request’s FastCGI PARAMS stream from every matching record’s content data
- exclude record headers and padding, and require the zero-length PARAMS terminator before applying the `http` scheme fallback
- cover a truncated later PARAMS record and a complete multi-record stream with a name/value pair split across records

## Validation

- [x] Feature and support-matrix updates are not needed for this parser correction.
- [x] `go test ./pkg/ebpf/common/...`
- [x] `make lint`
- [x] `make verify`


NimrodAvni78 (approved):

grcevski (approved):
LGTM!

PR #3535, opened 2026-09-22
refactor: move the process metadata resource dectection out of attacher

grcevski:
## Summary

Now that all the major programming languages resource metadata detection is complete, I'm moving the linux host resource detector in it's own package which the attacher uses, instead of this logic being buried in the attacher.

We could use this package to build more metadata extractors, like cloud vendors for example.

## Validation


pkg/metadata/process.go
```diff
@@ -0,0 +1,56 @@
...
+
+// ProcessResourceDetector finds resources like the service.name, service.namespace and service.version,
+// from the process binary or the process deployment directory.
+type ProcessResourceDetector struct {
+	log *slog.Logger
+}
+
+func NewProcessResourceDetector() *ProcessResourceDetector {
+	return &ProcessResourceDetector{
+		log: slog.With("component", "kube.Store"),
```
mmat11, line 28: is this component correct?

pkg/metadata/process.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+}
+
+func (r *ProcessResourceDetector) ResolveMetadata(t svc.InstrumentableType, fi *exec.FileInfo) {
```
mmat11, line 32: do you think it would be better design to make this a method of *exec.FileInfo? 

eg. `fi.ResolveMetadata(t svc.InstrumentableType)`

`ProcessResourceDetector` seems unnecessary unless we plan to expand it

mmat11: talking about refactoring, maybe `svc.InstrumentableType` should also implement `ResolveServiceMetadata()`, so we can avoid the big switch

mmat11: (this is all non-blocking)

mmat11 (approved):

pkg/metadata/process.go
grcevski: Ugh, copy pasta mess. I'll fix the name

pkg/metadata/process.go
grcevski: I tried that initially and got into import cycles problems. I can follow-up later to see if we can further improve it.

mmat11: all good -- we can merge as is

PR #3536, opened 2026-09-22
Resolve failed python service names from the directory

grcevski:
## Summary

This similar logic like we used for rails, when all fails and we cannot find the python framework specific name, we fall back to using the application directory, after confirming it's not one of the usual /app /server kind of names.

## Validation


pkg/internal/pythontools/service_metadata.go
```diff
@@ -94,6 +94,9 @@ func ResolveServiceMetadata(fileInfo *exec.FileInfo) error {
 		if name == "" {
 			name = frameworks.TargetName(launch.Target)
 		}
+		if name == "" && targetFound {
```
MrAlias: When Gunicorn resolves a generic target and `--name` is set, this branch wins before `launch.FallbackName`. For `gunicorn --name orders-worker server:app` from `/workspace/python-travel-agent`, the service name changes from `orders-worker` to `python-travel-agent`, so the explicit process name is silently overridden by a directory heuristic. Could we preserve `FallbackName` ahead of this fallback and add the target-found case beside `gunicorn name is final fallback`?

pkg/internal/pythontools/service_metadata.go
```diff
@@ -411,6 +414,19 @@ func processPath(root, hostPath string) string {
 	return string(filepath.Separator) + rel
 }
 
+func serviceNameFromTargetDirectory(root, targetPath string) string {
+	boundary, ok := langtools.ResolveProcessPath(root, "/", "/")
+	if !ok {
+		return ""
+	}
+
+	dir := filepath.Dir(targetPath)
```
MrAlias: This uses the target file's immediate parent, not the Python application directory. With cwd `/workspace/python-travel-agent` and target `src.server:app`, resolution finds `/workspace/python-travel-agent/src/server.py`; both `server` and `src` are generic, so this still emits no service name instead of `python-travel-agent`. Could we derive the directory from the launch/cwd rules used by `appDir`, or otherwise reach the intended application boundary, and cover a `src/` layout?

MrAlias (changes requested):
The fallback's precedence and directory selection can still produce the wrong service name or leave common source-layout applications unnamed.

pkg/internal/pythontools/service_metadata.go
```diff
@@ -97,6 +97,9 @@ func ResolveServiceMetadata(fileInfo *exec.FileInfo) error {
 		if name == "" {
 			name = frameworks.CleanValue(launch.FallbackName)
 		}
+		if name == "" && targetFound {
+			name = serviceNameFromAppDirectory(root, cwd, launch)
```
MrAlias, line 102: For `fastapi run` with only a `[tool.fastapi]` entrypoint in `/workspace/orders/pyproject.toml` and cwd `/workspace/orders/subdir`, `findFastAPIEntryPoint` returns `/workspace/orders` as `configDir`, but only adds it to `SearchPaths`. `appDir` then falls back to the cwd, so this emits `subdir` instead of `orders`.

Could we carry `configDir` into `launch.AppDir` and cover the parent-config/child-cwd case?

MrAlias (commented):
The two earlier cases are fixed. The automatic FastAPI path still loses its discovered application boundary before this fallback runs.

pkg/internal/pythontools/service_metadata.go
grcevski: Done.

MrAlias (approved):

NameHaibinZhang (approved):

